Issue #8, opened 2014-12-07
Shebang in Python files should use python2.7


HarveyHunt:
When trying to build seL4 on distros where Python 3 is the default python (such as Arch), the build fails when running python files as they are are python2 code, instead of python3.

The shebang should be:

```
#!/usr/bin/env python2.7
```

denibertovic:
/usr/bin/env python should use the default right? And if that's python3 it's going to fail, no?

lsf37:
Sorry, meant to reply to this way earlier: default python is a problem either way. There are systems where there are no `python2` or `python2.7` executables (mostly those that still have a python2 default, which still seems to be more widespread, even though that slowly is changing). This includes some of our internal standard dev platforms, so it would be pretty inconvenient for us to set that.

We use `/usr/bin/env python` precisely to enable use of virtualenv for those platforms that have a python3 default. 

So neither option will work for everyone and we have to make a choice. Currently, that is  `/usr/bin/env python` with python 2 documented as dependency in the build instructions. When things have shifted more towards python 3, we will probably change that default (and/or update the scripts as well).

PR #9, opened 2015-01-18
tools: *: use proper python2 shebang line

cyphar:
This patch updates the #! line in all of the python tools such that the
python version required for the tools (python2) is explcitly stated, so
that systems which have `python` symlinked to `python3` can still
compile the kernel without modification to the system or scripts.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/tools/bf_autocorres.py
+++ b/tools/bf_autocorres.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python2
+
 '''
 Functions related to generating AutoCorres-based definitions and proofs from
 the bitfield generator.
--- a/tools/lex.py
+++ b/tools/lex.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python2
 #
 # @TAG(OTHER_BSD)
 #
--- a/tools/yacc.py
+++ b/tools/yacc.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python2
 #
 # @TAG(OTHER_BSD)
 #
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 
 #
 # Copyright 2014, NICTA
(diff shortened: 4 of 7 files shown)


tools/bf_autocorres.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python2
+
```
auscompgeek: Uhh, correct me if I'm wrong, but I don't think this is supposed to have a shebang line.

tools/lex.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python2
```
auscompgeek: Same here.

tools/yacc.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python2
```
auscompgeek: ... and here.

tools/bf_autocorres.py
mbj: It is a very valid shebang line. Try to execute `/usr/bin/env python2` locally, and read: http://en.wikipedia.org/wiki/Shebang_%28Unix%29#Syntax

cyphar: I think he means that the file is not meant to be executed. The reason I use it here is so that it is semantically obvious that the file is a `python2` file from the first line.

mbj: Yeah I misread his comment. Sorry.

Smattr:
Sorry no one from seL4 has responded to this one yet. I guess it reflects that these files don't have a notional maintainer currently.

@auscompgeek is correct, that many of these files have no shebang because they're not intended to be executed. As for what shebang to use for the files that need one, this was discussed previously on #3 and #8. I believe the canonical way to handle a system where Python 3 is the default is with virtualenv, but I wouldn't be surprised if everyone internally is on systems that default to Python 2. Are you having issues using these files inside virtualenv?

PR #10, opened 2015-02-12
Experimental

Furao:
Adding support for the allwinner A20 processor, specifically on the CubieTruck. The allwinner A20 is a dual core ARM Cortex-A7 processor. The cortex-A7 is instruction equivalent to Cortex-A15, so I just said that the cpu was a Cortex-A15. For clarity purposes, this could be changed in the future.


AdrianDanis:
Is there a specific reason for this commit to be based on the experimental branch instead of master? Ideally for changes that do not affect verification we do them on master and then bring them over to experimental (I am aware that we do not say this anyway, so you have my apologies), just to try and avoid the two branches diverging unnecessarily. If its easy for you to rebase onto master that would be awesome, otherwise I'm happy to give it a crack, but I have no board to test it on.

Furao:
I saw a pull request on the experimental branch,  so I thought that was the
branch we needed to use. All I need to do is remove the last commit. I
actually first did my changes on the master branch. I will do that tomorrow.

Thanks.

Furao:
Rebased request to master as suggested.

PR #13, opened 2015-06-24
Update libsel4 to not have any dependency on libc/HAVE_LIBC.

winksaville:
This is accomplished by modifying several files:
 Kbuild, Kconfig, Makefile, syscalls.h, benchmark.h.

Porting code already implemented in the seL4:
 stdint.h, stdarg.h assert.[hc],
 machine_io.[hc] which becomes libsel4_io.[hc]

Porting sel4_crt0.S from libsel4platsupport which becomes libsel4_start.S

And creating a few new files:
 halt.[hc], stdbool.h, stddef.h,
 stdint.h which includes the sel4/arch/stdint.h, stddef.h

### DIFF
--- a/libsel4/Kbuild
+++ b/libsel4/Kbuild
@@ -9,4 +9,4 @@
 #
 
 libs-$(CONFIG_LIB_SEL4) += libsel4
-libsel4: $(libc) common
+libsel4: common
--- a/libsel4/Kconfig
+++ b/libsel4/Kconfig
@@ -10,7 +10,6 @@
 
 menuconfig LIB_SEL4
     bool "Build sel4 library"
-    depends on HAVE_LIBC
     default y
     help
         "Build the sel4 library"
--- a/libsel4/Makefile
+++ b/libsel4/Makefile
@@ -22,7 +22,12 @@ TARGETS          := libsel4.a
 
 # Source files required to build the target
 CFILES := \
-	$(patsubst $(SOURCE_DIR)/%,%,$(wildcard $(SOURCE_DIR)/src/*.c))
+	$(patsubst $(SOURCE_DIR)/%,%,$(wildcard $(SOURCE_DIR)/src/*.c)) \
+	$(patsubst $(SOURCE_DIR)/%,%,$(wildcard $(SOURCE_DIR)/src/${ARCH}/*.c))
+
+ASMFILES := \
+	$(patsubst $(SOURCE_DIR)/%,%,$(wildcard $(SOURCE_DIR)/src/*.S)) \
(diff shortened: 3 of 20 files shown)


winksaville:
This is not working with seL4test, I'm getting redefinitions, closing for now.

Smattr:
I realise you've closed this, but fair warning for when you resubmit that I would have straight up objected to this ;) My reasons can be boiled down to the following:

1. Work was done previously to constrain libsel4 to using the subset of C that our verification toolchain supports. These changes are a step backwards in this respect as they introduce things like builtin varargs support. This is a minor concern as it looks like the only instances of this stuff could be ifdefed out in a release build.
2. I don't believe it's practical or productive for us to maintain another implementation of `printf` inline. Yes, I'm aware we do this in the kernel, but there are fewer alternatives there.
3. Assertions are toggled with `DEBUG`. Perhaps this is a typo. Is this meant to be `NDEBUG`?
4. These changes are not suitable for a non-root task environment. E.g. the various halt instances (`halt`, return from `main`, etc.) are infinite loops. This DoSes everything else in the system if you are the highest priority thread.
5. There are various comments here that imply to me parts of this were lifted wholesale from the kernel. E.g. "Only base 10 and 16 supported for now. We want to avoid invoking the compiler's support libraries through doing arbitrary divisions." Things like this are not true in userspace. In fact, we link against libgcc. Then again, if you're purging libc dependencies, maybe you're also trying to purge GCC dependencies.
6. After these changes, the libsel4 files still include libc headers, e.g. stdint.h. Presumably this is meant to locate libsel4's copy of stdint.h, but in a project that also includes libc this will entirely depend on which library's headers get staged first. In fact in a parallel build you may even get a combination of incompatible headers from both libraries. The seL4 build system trusts libraries to namespace themselves (stage all their headers at "mylib/myheader.h") with only libc exempt from this. These changes make libsel4 and libc step on each other during staging.

This sounds like a bit of a laundry list of whingeing, so to clarify, I'm not opposed to the high level goal of this change. I just think it is significantly more involved than the currently proposed list of changes. Please let me know if anything I've said above doesn't make sense.

winksaville:
I've addressed some of the comments in my second pull request (https://github.com/seL4/seL4/pull/14). A comment I didn't address, but rather went much further, is I "lifted wholesale from the kernel" but in the case of libsel4_io I renamed the subroutines so the libc version can still be used. And the other header files I've prepended "libsel4_" on to the file name so they can't be accidentally used.

Anyway, I agree this is poor form and was also surprised when I couldn't' find seL4_BootInfo defined in the kernel and instead its "bi_t". So I'd suggest that a "new" library be created that is shared between the kernel and libsel4.

Lets continue the discussion on my newer change.

PR #14, opened 2015-06-24
Update libsel4 to not have any dependency on libc/HAVE_LIBC.

winksaville:
This is accomplished by modifying several files:
Kbuild, Kconfig, Makefile, syscalls.h, benchmark.h.

Porting code already implemented in the seL4:
 stdint.h becomes libsel4_stdint.h
 stdarg.h becomes libsel4_stdarg.h
 assert.[hc] becomes libsel4_assert.[hc]
 machine_io.[hc] which becomes libsel4_io.[hc]

Porting sel4_crt0.S from libsel4platsupport which becomes libsel4_start.S

And creating a few new files:
 libsel4_halt.[hc], libsel4_stdbool.h, libsel4_stddef.h,
 libsel4_stdint.h which includes the sel4/arch/stdint.h

### DIFF
--- a/libsel4/Kbuild
+++ b/libsel4/Kbuild
@@ -9,4 +9,4 @@
 #
 
 libs-$(CONFIG_LIB_SEL4) += libsel4
-libsel4: $(libc) common
+libsel4: common
--- a/libsel4/Kconfig
+++ b/libsel4/Kconfig
@@ -10,7 +10,6 @@
 
 menuconfig LIB_SEL4
     bool "Build sel4 library"
-    depends on HAVE_LIBC
     default y
     help
         "Build the sel4 library"
--- a/libsel4/Makefile
+++ b/libsel4/Makefile
@@ -22,7 +22,12 @@ TARGETS          := libsel4.a
 
 # Source files required to build the target
 CFILES := \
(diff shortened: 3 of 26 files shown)


AdrianDanis:
This change is taking libsel4 in the opposite direction to where I think it should be going. libsel4 is a *required* library for any piece of software on seL4. As such it should be being made more minimal, not more complicated. I am not against removing the dependency between libsel4 and libc, but I am definitely against pulling in more unneeded functionality.

libsel4 is meant to just be the syscall interface definitions. Anything that is on top of this should, ideally, be in a separate (and therefore optional) library. Currently there are things in libsel4 that violate this idea and could be moved out.

The question then is what is the best way to remove the dependency on libc. Should code in libsel4 simply not call `assert`? Should we define a different non colliding version, such as `libsel4_assert`? With your current change user code may use the `assert` defined in libsel4, or the `assert` defined in libc, depending on where/if they included the libc one, this should be avoided in my opinion.

heb1001:
How about allowing the user of libsel4 to register an assert_failed
callback and having an internal, private libsel4_assert which calls the
callback if it exists when an assert fails?

winksaville:
I'll start with changing the name of assert to libsel4_assert and I'll see the consequences of moving libsel4_assert and libsel4_printf outside libsel4.

PR #18, opened 2015-09-18
FAQ link is incorrect

DavidVorick:


wom-bat:
I've added a redirect on the sel4.systems site.  The link should work again now.

PR #23, opened 2016-05-26
fix Makefile

zhicheng:
fix Makefile delete source from incase-sensitive filesystem
fix CONFIG_DEBUG_BUILD not defined when DEBUG=y

Signed-off-by: Zhicheng Wei <zhicheng@opensourceforge.net>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,7 @@ DEFINES += ${CONFIG_DEFS:%=-D%}
 
 ifdef DEBUG
 DEFINES += -DDEBUG
+DEFINES += -DCONFIG_DEBUG_BUILD
 CFLAGS  += -ggdb -g3
 endif
 
@@ -581,7 +582,7 @@ autoconf.h: include/plat/${PLAT}/autoconf.h
 	@echo " [STRIP] $@"
 	$(Q)${STRIP} -o $@ $<
 
-%.o: %.s | ${DIRECTORIES}
+%.o: %.S | ${DIRECTORIES}
 	@echo " [AS] $@"
 	$(Q)${CC} ${ASFLAGS} -c $< -o $@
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -653,10 +654,6 @@ ${PROOFTHEORIES}: %_proofs.thy: %.pbf ${BF_GEN_PATH} ${STATICSOURCES} \
 # Preprocessed source files
 ###########################
 
-%.s: %.S ${GENHEADERS} ${STATICHEADERS} | ${DIRECTORIES}
-	@echo " [CPP] $@"
-	$(Q)${CPP} ${CPPFLAGS} -CC -E -o $@ $<
-
```
AdrianDanis: Why are you just deleting this build step?

Makefile
```diff
@@ -265,6 +265,7 @@ DEFINES += ${CONFIG_DEFS:%=-D%}
 
 ifdef DEBUG
 DEFINES += -DDEBUG
+DEFINES += -DCONFIG_DEBUG_BUILD
```
AdrianDanis: Can I ask what your use case is that you are wanting to do standalone debug kernel builds? Typically the only reason to build the kernel standalone is for verification, which is not a debug a build. Not saying we shouldn't fix this support, just curious as to your use case.

Makefile
zhicheng: in the incase-sensitive filesystem lowercase .s file and uppercase .S file are same file. this cause .s and .S become a empty file. and I don't see anywhere is using lowercase .s file. So I deleted. tell me if I was wrong.

Makefile
zhicheng: Trying porting to new board,thought this will be helpful.

Makefile
AdrianDanis: The lowercase .s was referenced in a previous rule (where you changed the .s to .S). The idea is that this build step runs the .S files through the C pre-processor, producing .s files that are passed to the assembler.

zhicheng:
```
%.o: %.S | ${DIRECTORIES}
        @echo " [AS] $@"
        $(Q)${CC} ${ASFLAGS} -c $< -o $@
```

if using gcc compiling .S file,It's will throught C pre-processor.
or
change .s to .S.s or something that don't overwrite the .S source code.

AdrianDanis: Yes, given we currently invoke the assembler through gcc the pre-processor is getting invoked anyway. There is a case then to be made for whether we still want to keep the explicit pre-process step or not, but that should be an explicit discussion, not just happen as an unwritten side effect of supporting case insensitive builds.

If you just changed the '.S' to something (like say '.s_pp' if you were to follow naming conventions elsewhere in the Makefile) then we would happily accept that change.

Not saying we wouldn't accept the change of removing a redundant pre-processing step, but it would be an actual discussion.

PR #24, opened 2016-05-30
fix Makefile

zhicheng:
fix Makefile delete source from incase-sensitive filesystem
fix CONFIG_DEBUG_BUILD not defined when DEBUG=y

Signed-off-by: Zhicheng Wei zhicheng@opensourceforge.net

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,6 +265,7 @@ DEFINES += ${CONFIG_DEFS:%=-D%}
 
 ifdef DEBUG
 DEFINES += -DDEBUG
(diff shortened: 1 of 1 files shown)


AdrianDanis:
This looks fine to me. I'll pull it internally and once all the regressions and tests are passing again it will get merged

AdrianDanis:
I should have checked this to begin with, but it looks like you haven't signed a CLA. Unfortunately you need to sign and send a CLA, see instructions [here](http://sel4.systems/Community/Contributing/), before we can accept your commits

PR #25, opened 2016-06-02
should enable print in debug build

zhicheng:
add CONFIG_PRINTING to DEBUG build
fix `Unnamed Structure and Union Fields' defined for old gcc (add -fms-extensions)
fix typo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -266,6 +266,8 @@ DEFINES += ${CONFIG_DEFS:%=-D%}
 ifdef DEBUG
 DEFINES += -DDEBUG
 DEFINES += -DCONFIG_DEBUG_BUILD
+DEFINES += -DCONFIG_PRINTING
+DEFINES += -DCONFIG_USER_STACK_TRACE_LENGTH=1
 CFLAGS  += -ggdb -g3
 endif
 
@@ -382,7 +384,7 @@ endif
 WARNINGS = all error strict-prototypes missing-prototypes nested-externs \
 	missing-declarations undef pointer-arith no-nonnull declaration-after-statement
 
-CFLAGS += --std=c99 -nostdlib -nostdinc -ffreestanding \
+CFLAGS += --std=c99 -nostdlib -nostdinc -ffreestanding -fms-extensions \
 	${WARNINGS:%=-W%} ${INCLUDES}
 LDFLAGS += -nostdlib -nostdinc
 LDFLAGS += -Wl,--build-id=none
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -382,7 +384,7 @@ endif
 WARNINGS = all error strict-prototypes missing-prototypes nested-externs \
 	missing-declarations undef pointer-arith no-nonnull declaration-after-statement
 
-CFLAGS += --std=c99 -nostdlib -nostdinc -ffreestanding \
+CFLAGS += --std=c99 -nostdlib -nostdinc -ffreestanding -fms-extensions \
```
AdrianDanis: Can you elaborate on exactly what this is fixing?

zhicheng: in file include/api/debug.h
```
typedef struct {
    debug_entry_type_t path;
    union {
        struct {
            word_t irq;
        };
        struct {
            word_t word;
        };
        struct {
            word_t number;
            word_t code;
        };
        struct {
            word_t fault_type;
        };
        struct {
            word_t syscall_no;
            word_t cap_type;
            word_t invocation_tag;
        };
    };
} debug_entry_t;
```
on my old gcc will got error
```
./include/api/debug.h:34:10: error: declaration does not declare anything
./include/api/debug.h:37:10: error: declaration does not declare anything
./include/api/debug.h:41:10: error: declaration does not declare anything
./include/api/debug.h:44:10: error: declaration does not declare anything
./include/api/debug.h:49:10: error: declaration does not declare anything
./include/api/debug.h:50:6: error: declaration does not declare anything
```

this is c11 feature, explain
https://gcc.gnu.org/onlinedocs/gcc/Unnamed-Fields.html

AdrianDanis: What is your plan going forward? The kernel sticks to relatively small subset of C, but the user libraries (including libsel4) assume very modern versions of gcc. Every architecture seL4 supports has an easy to acquire version of gcc that will build it. Instead of embarking on this journey of building with an old gcc on a case insensitive file system why do you not just build on another system (in a virtual machine for instance) where this will just work?

zhicheng: I was build with my newer gcc, but toolchain crashed (CONFIG_PRINTING relative I think).then I test my old gcc and got this issue,so I fixed.

if you don't like, just don't merge.

Thanks

AdrianDanis: I would prefer not to add additional flags. Perfectly happy with the other changes though

zhicheng: Sure,need me reopen a new pr?

AdrianDanis: You can rewrite this commit and force push to this PR, or open a new one, whichever is easier.

Whilst modifying the commit though it would be nice if you tried to conform with our normal commit styling though (do a git log through our commits), which is a short summary line, a blank line, then any necessary paragraphs giving more detail.

PR #26, opened 2016-06-02
add CONFIG_PRINTING to DEBUG build

zhicheng:
fix typo: src/plat/pc99/machine/acpi.c, ACIP should be ACPI


AdrianDanis:
I don't understand the line 'source and binary changed' in the commit message

zhicheng:
this patch will generate different binary. notify current user.

AdrianDanis:
Just about any of the commits in this repository would lead to a different binary. I think this line is confusing and unnecessary, could you please remove it?

AdrianDanis:
@zhicheng sorry not responding to this, github helpfully didn't notify me that you had added a commit. This will get merged once it gets through continuous integration

Issue #33, opened 2016-08-09
Idle thread runs with async aborts enabled


philipderrin:
The idle thread uses the hard-coded CPSR value `PMODE_IDLE`, which does not mask FIQs or async aborts. If an async abort is pending, the kernel will fault as soon as the system becomes idle.

I see this fault happening on i.MX6Q, presumably triggered by a bad device register access in either the loader or the kernel's serial or timer drivers. Depending on the platform it may also be possible to deliberately trigger it from a user thread that has device registers mapped.

This was fixed before by commit a5f61c7, but it looks like the fix was lost in a merge with the hyp support branch. Also, the `CPSR_IDLETHREAD` value added by that commit has not been updated for hyp; it should use `PMODE_IDLE` instead of `PMODE_SYSTEM`.

AdrianDanis:
Thanks for noticing @philipderrin. Will fix this up

jdub:
Just in case anyone hits this and has also stumbled upon the ["Wandaboard port"](https://sel4.systems/pipermail/devel/2015-March/000324.html) thread on the devel@ list… this looks like the same issue. The referenced fix was committed two days after that thread.

I brought those fixes back to the hyp-ed up 3.2.0 tag with this change:

```diff
diff --git a/include/arch/arm/arch/32/mode/machine/registerset.h b/include/arch/arm/arch/32/mode/machine/registerset.h
index 87abb58..75891f1 100644
--- a/include/arch/arm/arch/32/mode/machine/registerset.h
+++ b/include/arch/arm/arch/32/mode/machine/registerset.h
@@ -34,7 +34,7 @@
                              | CPSR_EXTRA_FLAGS   )

 #define CPSR_IDLETHREAD      ( PMASK_FIRQ         \
-                             | PMODE_SYSTEM       \
+                             | PMODE_IDLE         \
                              | CPSR_EXTRA_FLAGS   )

 /* Offsets within the user context, these need to match the order in
diff --git a/src/arch/arm/32/kernel/thread.c b/src/arch/arm/32/kernel/thread.c
index 44ccdb8..a9af577 100644
--- a/src/arch/arm/32/kernel/thread.c
+++ b/src/arch/arm/32/kernel/thread.c
@@ -26,7 +26,7 @@ Arch_switchToThread(tcb_t *tcb)
 BOOT_CODE void
 Arch_configureIdleThread(tcb_t *tcb)
 {
-    setRegister(tcb, CPSR, PMODE_IDLE);
[...]

PR #35, opened 2016-09-08
Mask async aborts for idle thread in debug kernels

jdub:
Fixes seL4/seL4#33

### DIFF
--- a/include/arch/arm/arch/32/mode/machine/registerset.h
+++ b/include/arch/arm/arch/32/mode/machine/registerset.h
@@ -35,6 +35,7 @@
 
 #define CPSR_IDLETHREAD      ( PMASK_FIRQ         \
                              | PMODE_IDLE         \
+                             | PMASK_ASYNC_ABORT  \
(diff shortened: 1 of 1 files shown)


jdub:
I suppose it's a bit cheeky to say this fixes #33, as it was already partially fixed in b727ede2c5c054277bf386566d4d966254883620. But hey. 😳

AdrianDanis:
Hi @jdub, thanks for the patch. As per the comments on #33 I think it would be much more preferable if this change was guarded by a config option, and not applied unilaterally to all debug builds.

jdub:
As it doesn't actually resolve any of the problems I raised in #33, is it worth adding the option? Is there a legitimate reason someone might want it? Otherwise, I'm happy to let it go.

AdrianDanis:
Certainly I'm happy to have the option added, but I cannot speak as to whether or not anyone else would actually want it.

jdub:
Ha, okay. My philosophy is that options should clear a very high bar, so if you can't think of one, I'm done. :smile:

PR #39, opened 2016-10-05
add stub Arch_userStackTrace for x86 PAE

ratmice:
When configuring with both CONFIG_PAE and CONFIG_PRINTING,
this needed to be declared (I didn't implement the function however, just added a stub)


AdrianDanis:
hi @ratmice thanks for the PRs

Before we can accept code we unfortunately need you to [sign a CLA](http://sel4.systems/Community/Contributing/)

I'm fine with this change to get PAE building again, but as you can see we don't really use PAE and so it is not well tested or maintained.

ratmice:
Hi @AdrianDanis, I sent the CLA last night, and got a vacation response, it should get to you guys eventually, nothing here that can't wait.

PR #43, opened 2016-10-26
Let x86_64 CallWithMRs

jdub:
* `syscall_stub_gen.py` now works on x86_64 without passing `--buffer`
* *All is well in the universe* with x86_64 in qemu if I remove `--buffer` from the `libsel4` Makefile


AdrianDanis:
I'm happy to accept this, since it does make the syscall stub generator work better, but I would prefer the commit message to be changed slightly. The second line gives the implication that `--buffer` is now safe to remove, which is sort of not true. If you just drop the second line and leave the first one then I'd be happy.

Sorry for being pedantic.

jdub:
Force push to amend the commit message.

PR #45, opened 2016-10-27
Retain register values when not using IPC buffer

jdub:
- passes the sabre/qemu test suite :-)
- this assumes we want to retain the same number of registers as we
  have input parameters, which on reflection makes no sense at all

### DIFF
--- a/libsel4/tools/syscall_stub_gen.py
+++ b/libsel4/tools/syscall_stub_gen.py
@@ -557,11 +557,10 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
     #
     # Setup variables we will need.
     #
-    if returning_struct:
-        result.append("\t%s result;" % return_type)
+    result.append("\t%s result;" % return_type)
     result.append("\tseL4_MessageInfo_t tag = seL4_MessageInfo_new(%s, 0, %d, %d);"  % (method_id, len(cap_expressions), len(input_expressions)))
     result.append("\tseL4_MessageInfo_t output_tag;")
-    for i in range(min(num_mrs, max(input_param_words, output_param_words))):
+    for i in range(num_mrs):
         result.append("\tseL4_Word mr%d;" % i)
     result.append("")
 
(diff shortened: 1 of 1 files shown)


jdub:
Is it okay if we chat about the changes here, rather than clogging up the list with my very specific questions?

In the process of updating this to retain all the real register values, I realise that none of these functions even *get* all of the values, because they're calling `seL4_CallWithMRs` like this:

```c
        output_tag = seL4_CallWithMRs(service, tag,
                &mr0, &mr1, seL4_Null, seL4_Null);
```

The parent function has two input parameters, so it only passes two pointers. We never even see the other two. I'm assuming this is part of the "errors forever lost" problem, because surely the number of error words for a function is unlikely to be related to the number of inputs?

So, the generator should:
- initialise `num_mrs` variables
- pass them *all* to CallWithMRs
- if there's an error, store *all* of them into the IPC buffer

Is that right?

AdrianDanis:
Yep, that all sounds about right to me

jdub:
Bit more to do, but is this approach correct and inoffensive so far?

AdrianDanis:
Looks good so far. I'll probably check the generated functions to be 100% sure though

jdub:
Here are [some examples](https://gist.github.com/jdub/40077cfffcf967a227685bf4880e240f) to save you the git juggling.

AdrianDanis:
So in the case where we are returning a `struct` I think that in the case of an error we should probably be putting MRs into the IPC buffer, and only in the case of success should we put the MRs into the struct.

jdub:
Something like this?

```c
LIBSEL4_INLINE seL4_ARM_Page_GetAddress_t
seL4_ARM_Page_GetAddress(seL4_ARM_Page service)
{
        seL4_ARM_Page_GetAddress_t result;
        seL4_MessageInfo_t tag = seL4_MessageInfo_new(ARMPageGetAddress, 0, 0, 0);
        seL4_MessageInfo_t output_tag;
        seL4_Word mr0 = 0;
        seL4_Word mr1 = 0;
        seL4_Word mr2 = 0;
        seL4_Word mr3 = 0;

        /* Perform the call, passing in-register arguments directly. */
        output_tag = seL4_CallWithMRs(service, tag,
                &mr0, &mr1, &mr2, &mr3);

        result.error = seL4_MessageInfo_get_label(output_tag);

        /* Unmarshal result or retain registers in IPC buffer. */
        if (result.error == seL4_NoError) {
            result.paddr = mr0;
        } else {
                seL4_SetMR(0, mr0);
                seL4_SetMR(1, mr1);
                seL4_SetMR(2, mr2);
                seL4_SetMR(3, mr3);
        }

        return result;
}
```

jdub:
I took a stab at this. How do you feel about [early returns](https://gist.github.com/jdub/d4548690dd82c58b116d3590fa46ed2e)?

jdub:
Looking good on ia32/qemu too.

jdub:
Added *WithMRs basic tests [and posted results](https://gist.github.com/jdub/8e950c3dff0c3941c60a08c0eae279e3). Will have to add something that checks errors, to properly test the *WithMRs path even if the suite is run with `--buffer` generated syscall stubs.

AdrianDanis:
Early return is fine.

This all looks really good. I'm going to run this on some internal projects that use errors (they are what triggered the `--buffer` change to begin with) and if they're good I'll merge it in.

Bit swamped right now though so might not get around to this till monday

jdub:
btw, this depends on #43 and #44 to build.

AdrianDanis:
Ran some tests with this and everything seems to work, so you can neaten the commit history now and we'll be good to go.

jdub:
Squashed!

AdrianDanis:
Awesome. This and all your other changes should appear tonight, unless our regressions are still failing

jdub:
Wow, sweet! Thanks for all your help. This was an enjoyable little achievement. :-)

PR #47, opened 2016-11-23
Optional inlining of functions and syscalls

jdub:
`CONFIG_LIB_SEL4_INLINE_INVOCATIONS=n` will now disable inlining for all external interfaces (except deprecated functions).

### DIFF
--- a/libsel4/Kconfig
+++ b/libsel4/Kconfig
@@ -16,15 +16,34 @@ menu "libsel4"
         help
             seL4 API library
 
+choice
+    prompt "Function attributes"
+    default LIB_SEL4_INLINE_INVOCATIONS
+
+config LIB_SEL4_DEFAULT_FUNCTION_ATTRIBUTES
+    bool "Default"
+    depends on LIB_SEL4
+    help
+        Verification friendly default configuration. syscalls will be inlined,
+        but generated functions will not.
+
 config LIB_SEL4_INLINE_INVOCATIONS
     bool "Inline generated syscall invocations"
     depends on LIB_SEL4
-    default y
     help
         When set to true will mark generated functions as 'inline', allowing
         them to be inlined by the callee user code. This may be undesirable
-        for verification, so setting to 'n' will forcively prevent the function
-        from being inlined
+        for verification, so setting to 'n' will forcibly prevent the function
+        from being inlined.
+
+config LIB_SEL4_PUBLIC_SYMBOLS
+    bool "Public symbols for all external interfaces"
+    depends on LIB_SEL4
+    help
+        When set to true will make all user facing functions available as
+        public symbols, which can be convenient for some language bindings.
+
+endchoice
 
 config LIB_SEL4_STUBS_USE_IPC_BUFFER_ONLY
     bool "use only IPC buffer for syscalls"
--- a/libsel4/include/sel4/macros.h
+++ b/libsel4/include/sel4/macros.h
@@ -33,8 +33,15 @@
 
 #ifdef CONFIG_LIB_SEL4_INLINE_INVOCATIONS
 #define LIBSEL4_INLINE static inline
+#define LIBSEL4_INLINE_FUNC static inline
+
+#elif CONFIG_LIB_SEL4_PUBLIC_SYMBOLS
+#define LIBSEL4_INLINE __attribute__((unused)) __attribute__((weak))
+#define LIBSEL4_INLINE_FUNC __attribute__((unused)) __attribute__((weak))
+
 #else
 #define LIBSEL4_INLINE __attribute__((noinline)) __attribute__((unused)) __attribute__((weak))
+#define LIBSEL4_INLINE_FUNC static inline
 #endif
 
 #define SEL4_DEPRECATED(x) __attribute__((deprecated(x)))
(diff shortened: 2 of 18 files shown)


jdub:
This is an RFC more than anything else. Makes life a little easier for some language bindings.

Does `libsel4/include/sel4/cap_rights.h` need some love too?

AdrianDanis:
The original intention of `LIBSEL4_INLINE` was to actually be able to result in having invocations (such as `seL4_TCB_Resume`) as separate functions, but with their contents all inline.

Not against preventing the inlining of all these functions, but it should be controlled through a different macro/configuration to preserve the old behaviour.

`cap_rights.h` is going to go away very soon (hopefully next week), so I wouldn't spend any time on it

jdub:
I'm not particularly enamoured with `LIBSEL4_INLINE_FUNC`, but that's easily changed.

Hmm, this should probably override `LIBSEL4_INLINE` such that syscalls aren't completely prevented from being inlined. In which case perhaps the configuration item could be called `LIB_SEL4_PUBLIC_SYMBOLS`.

AdrianDanis:
Yeah I like the configuration `LIB_SEL4_PUBLIC_SYMBOLS`. Haven't yet thought of anything better for `LIBSEL4_INLINE_FUNC` though. Probably it would have made more sense for `LIBSEL4_INLINE` to have originally been `LIBSEL4_INLINE_INVOCATION`, and then this naming would be easier.

Otherwise though I think this is fine. I'll let you clean up the commits before I pull it in.

jdub:
Status:
- Rebased, tidied up previous commits (new stuff is split out for easy review)
- Caught up with faults/mapping

Questions:
- `tools/bitfield_gen.py`… I think some of this needs to be public?
- None of `manual/` has been changed

AdrianDanis:
Seems fine to me. Not sure what you're referring to about `tools/bitfield_gen.py` though. It's fine to leave the manual unchanged.

jdub:
`tools/bitfield_gen.py` generates public interfaces, so I had to change its templates from `static inline` to `LIBSEL4_INLINE_FUNC`. There may be subtleties I missed.

Also, this breaks the build when it's not enabled -- `LIBSEL4_INLINE_FUNC` just needs a default in `macros.h`. Will add in a minute.

AdrianDanis:
Ahh I understand. Yes changing `bitfield_gen.py` like that makes sense. I suspect that you want to make sure that when compiling the kernel `LIBSEl4_INLINE_FUNC` just stays at `static inline`, and public interfaces only apply to the rest of the user level build.

jdub:
Ah, yuck. Good point. Oh, gross. OK.

jdub:
That ought to do it. Builds and tests fine, on and off.

AdrianDanis:
Yeh it's a bit yuck. What you have looks good though. I'll let you do round 2 of cleaning up the commits then

jdub:
The script wasn't so bad -- I thought it was going to take way longer. There was a non-Python, non-string-template generator for this stuff historically, wasn't there? Anything I can read about that?

AdrianDanis:
To my knowledge the bitfield generator has always been a python script, I'll see if I can dig around in some of the original repositories and see if it has a more checkered history.

libsel4/include/sel4/macros.h
```diff
@@ -37,6 +37,12 @@
 #define LIBSEL4_INLINE __attribute__((noinline)) __attribute__((unused)) __attribute__((weak))
 #endif
 
+#ifdef CONFIG_LIB_SEL4_PUBLIC_SYMBOLS
+#undef LIBSEL4_INLINE
+#define LIBSEL4_INLINE __attribute__((unused)) __attribute__((weak))
+#define LIBSEL4_INLINE_FUNC __attribute__((unused)) __attribute__((weak))
```
AdrianDanis: Should there be an `#else` branch defining `LIBSEL4_INLINE_FUNC` back to `static inline`?

jdub: Hmm, yeah, I guess it's falling back to nothing right now. I hadn't tested with both off! I might tidy this up with an `#elif`.

jdub:
How's that?

libsel4/include/sel4/macros.h
```diff
@@ -33,14 +33,15 @@
...
 #ifdef CONFIG_LIB_SEL4_INLINE_INVOCATIONS
 #define LIBSEL4_INLINE static inline
-#else
-#define LIBSEL4_INLINE __attribute__((noinline)) __attribute__((unused)) __attribute__((weak))
-#endif
+#define LIBSEL4_INLINE_FUNC static inline
 
-#ifdef CONFIG_LIB_SEL4_PUBLIC_SYMBOLS
-#undef LIBSEL4_INLINE
+#elif CONFIG_LIB_SEL4_PUBLIC_SYMBOLS
```
jdub: Uff, that means I have to come up with a prompt. And a "neither" option. *thinking*

jdub:
Kind of icky, but okay?

libsel4/Kconfig
```diff
@@ -16,24 +16,35 @@ menu "libsel4"
         help
             seL4 API library
 
+choice
+    prompt "Function attributes"
+    default LIB_SEL4_DEFAULT_FUNCTION_ATTRIBUTES
```
AdrianDanis: I think the default should be `LIB_SEL4_INLINE_INVOCATIONS`, to match that it was previously `default y`?

AdrianDanis:
Okay looks good. Let's go for round 3 of rebasing

jdub:
Go for it! Also, \*looks at time\* go home! 😄

PR #49, opened 2016-12-11
Enable SYSCALL (SCE) after entering long mode

jdub:
It appears to be somewhat more correct to flip SCE after entering long mode, as per other operating systems (including Linux).

c.f. [Linux 4.8's head_64.S](http://lxr.free-electrons.com/source/arch/x86/kernel/head_64.S?v=4.8#L210), which is executed in long mode.

### DIFF
--- a/src/arch/x86/64/head.S
+++ b/src/arch/x86/64/head.S
@@ -318,15 +318,16 @@ BEGIN_FUNC(common_init)
     andl $0x7fffffff, %eax
     movl %eax, %cr0
 
-#ifdef CONFIG_SYSCALL
-    call syscall_enable
-#endif
-
     call fsgsbase_enable
     /* Initialize boot PML4 and switch to long mode */
     call setup_pml4
     call enable_x64_mode
     lgdt _gdt64_ptr
+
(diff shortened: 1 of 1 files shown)


AdrianDanis:
Other than Linux initializing it here is there any indication or comments that it is actually more correct? I cannot find any indication in the Intel documentation that this is the case.

Not that I'm against moving it, just curious really.

jdub:
So (fully aware that I'm telling you what you already know), on Intel, SCE only has meaning in long mode, because there's no support for SYSCALL in protected mode. That gives me a subtle whiff of relative correctness about the order of operations, but then, it hasn't been a problem on target platforms to date.

That said, Xen* will cheerfully crashdump the guest at the `wrsmr` instruction if SCE is flipped in protected mode. At least on Intel. 😃 

\* "Aha!" I hear you say…

AdrianDanis:
Well your argument does make some degree of sense. I still feel I should be able to do SCE so that it is all ready and waiting for me when I enable long mode, but I guess I'm being picky. Thanks for pointing out the Xen case though

PR #51, opened 2016-12-14
Minor fixes from quick proof read of 4.0 manual

jdub:
Notes:

- In Faults 6.2.3, IA-32 appears under the table in the PDF… not sure how to fix.

- Is there something wrong with the `_service` descriptions in 10.3.x?

- Some of the functions in the manual are `LIBSEL4_INLINE_FUNC`, some are `static inline`.

- My understanding is that a VSpace capability represents the highest level paging structure, e.g. PML4 on x86_64. If that's correct, it'd be worth mentioning it. If not, clarifying the documentation. :-)

Hmm, but oh:

```c
/* Legacy code will have assumptions on the vspace root being a Page Directory
 * type, so for now we define one to the other */
#define seL4_CapInitThreadPD seL4_CapInitThreadVSpace
```

So, there's a second commit to make the manual reflect the current init cap list, but clearly my understanding of what's going on with VSpaces and paging structures is warped.

### DIFF
--- a/manual/parts/bootup.tex
+++ b/manual/parts/bootup.tex
@@ -38,7 +38,7 @@ The first 12 slots contain specific capabilities as listed in
       0x0 & \texttt{seL4\_CapNull}                & null \\
       0x1 & \texttt{seL4\_CapInitThreadTCB}       & initial thread's TCB \\
       0x2 & \texttt{seL4\_CapInitThreadCNode}     & initial thread's CNode \\
-      0x3 & \texttt{seL4\_CapInitThreadPD}        & initial thread's page directory \\
+      0x3 & \texttt{seL4\_CapInitThreadVSpace}    & initial thread's VSpace \\
       0x4 & \texttt{seL4\_CapIRQControl}          & global IRQ controller (see \autoref{sec:interrupts}) \\
       0x5 & \texttt{seL4\_CapASIDControl}         & global ASID controller (see \autoref{ch:vspace}) \\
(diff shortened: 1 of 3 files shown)


AdrianDanis:
The manual is getting out of date in places. There is some effort internally at the moment to start generating parts of the manual from the source code, which will fix the worst of the API/datastructure inconsistencies. And things like x86-64 just aren't in the manual at all, aside from where it overlaps with common ia32 stuff.

* The tables look fine to me. I see one table with "User exception outcome on the ARM architecture." under it and one with "User exception outcome on the IA-32 architecture". Am I looking at the right thing
* Doesn't seem right, will look into it
* This seems to be due to some invocations being generated from the source, and some not. Might have to preprocess before doxygen steps or something
* Yeah needs documentation of x86-64.

jdub:
Regarding 6.2.3:

- 6.2.3 section title
- para
- ARM
- table
- table
- IA-32 (should be before the table)
- 6.2.4 section title

jdub:
btw, did I figure it out correctly that the VSpace represents the highest level paging structure? In my experiments, that seems to be right, but the comment ahead of the PD -> VSpace define worries me.

AdrianDanis:
Okay I understand. I think those 'ARM' and 'IA32' heading should be dropped, since trying to force the table to appear inside them seems like it would lead to more annoying formatting.

Yep VSpace is meant to represent the 'vspace root', which is the highest level paging structure. The comment is trying to say that previously 'generic' code will have talked about `seL4_CapInitThreadPD` when they really just wanted to talk about the vspace root.

jdub:
Ah, got it. Glad I'm not bonkers! I'll see if I can come up with some appropriate verbiage to make that clear.

PR #64, opened 2017-06-19
Unused functions found by clang

mokshasoft:
Clang found some unused static functions. I think this touches verified code, but removing functions that aren't used is invisible to the proof? Currently I do not know how to update the proof, so if that is needed you could ignore these patches.

make compile-all passed all targets.

All tests passed running:
make mrproper ; make zynq_simulation_debug_xml_defconfig ; make ; make simulate-zynq

### DIFF
--- a/src/arch/arm/armv/armv7-a/cache.c
+++ b/src/arch/arm/armv/armv7-a/cache.c
@@ -10,11 +10,6 @@
 
 #include <arch/machine/hardware.h>
 
-static inline void invalidateByWSL(word_t wsl)
-{
-    asm volatile("mcr p15, 0, %0, c7, c6, 2" : : "r"(wsl));
-}
-
 static inline void cleanByWSL(word_t wsl)
 {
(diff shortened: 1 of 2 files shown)


src/arch/arm/armv/armv7-a/cache.c
```diff
@@ -10,11 +10,6 @@
 
 #include <arch/machine/hardware.h>
 
-static inline void invalidateByWSL(word_t wsl)
```
AdrianDanis: This is part of the cache interface and shouldn't be deleted just because it's not currently being used

mokshasoft: Is set_irq_active also considered part of an interface? In that case I'll close this pull request, otherwise I'll update this request or create a new one with only set_irq_active.

AdrianDanis: `set_irq_active` just looks like an internal helper that was never used, it can go

mokshasoft:
The commit for set_irq_active uploaded in pull request https://github.com/seL4/seL4/pull/65

Issue #67, opened 2017-08-29
x86 UEFI support by using Multiboot2 with GRUB2


alex-ab:

alex-ab:
I added the current commits as are for the moment.

@AdrianDanis: you meant that the ACPI RSDP information should not be specifically transferred via the bootinfo, instead by the extralen bootinfo mechanism, right ? Should just the ACPI RSDP specific part be copied or the whole multiboot2 data (which contains a copy of ACPI RSDP). I presume I would need a kind of identification for the multiboot2 copy in the seL4_BootInfoHeader ? (to differentiate just ACPI RSDP data or Multiboot2 data or/and Multiboot 1 data?)

AdrianDanis:
I don't think the whole multiboot2 data need be copied (or at least not as whole block). Some header for ACPI/RSDP information seems sufficient, that way it can be filled from any source, mbi2 or otherwise

alex-ab:
I switched to provide the ACPI RSDP via the extra bootinfo page. Please have a look at 55acca4db9416e7b9db1a4243b58be7d5b310b86

AdrianDanis:
@alex-ab Yep that is more what I was imagining with bootinfo

alex-ab:
Ok. Please make suggestions, if any,, to enhance the commit to make it fit for addition to your repository. From my side I'm in principle done for now.

AdrianDanis:
The main issue is the large duplication in the boot code. I don't see any obvious reason this couldn't be refactored

alex-ab:
Adjusted in 7e9694b59e33b7080945dab070a709fc5a99ef1a

AdrianDanis:
You have the choice of mbi1 or mbi2 driving the entire boot path, which I think is a little backwards, and has resulted in a bunch of stuff in a `try_load_boot_module` that doesn't seem to be in any way related to loading a boot module. Rather we just want to pull the information out of the differently structured mbi structures and keep on going. Given that we already put most things into the `boot_state` global I think it'd make more sense to have two versions for initializing that, then calling into the actual general booting in `try_boot_sys`.

alex-ab:
Adjusted in 27bc140865e82bbf3bb8d80c2a58b91d8ae27960.

AdrianDanis:
I was imagining a structure more like

```c
    if (multiboot_magic == MULTIBOOT_MAGIC) {
        result = try_parse_mbi1(mbi);
[...]

PR #71, opened 2017-09-27
x86: add multiboot2 support

alex-ab:
Issue #67


src/plat/pc99/machine/acpi.c
```diff
@@ -244,20 +215,34 @@ acpi_table_init(void* entry, enum acpi_type table_type)
...
 
-    if (acpi_rsdp == NULL) {
+    printf("ACPI: RSDP vaddr=%p\n", acpi_rsdp);
+
+    if (!acpi_rsdp || acpi_calc_checksum((char const *)acpi_rsdp, 20) != 0) {
         printf("BIOS: No ACPI support detected\n");
         return NULL;
     }
-    printf("ACPI: RSDP paddr=%p\n", acpi_rsdp);
-    acpi_rsdp = acpi_table_init(acpi_rsdp, ACPI_RSDP);
```
AdrianDanis: I'm trying to understand this change. It seems like `acpi_search_rsdp` is returning the mapped rsdp address, but that only gets called from a multiboot1 boot. When using the value from multiboot2 as far as I can tell `acpi_table_init` does not get called, and we are just using the raw pointer. It isn't clear to me why that is guaranteed to be mapped in?

AdrianDanis:
I have some minor nitpicks and changes I would like to make to this, both in the code as well as splitting up into multiple commits with extended messages. Given you have already done the bulk of the implementation thus far I am happy to do these and leave your name on the commits if you do not have time, otherwise I can do more detailed comments in this PR for you.

src/plat/pc99/machine/acpi.c
alex-ab: > I'm trying to understand this change. It seems like acpi_search_rsdp is returning the mapped rsdp address, but that only gets called from a multiboot1 boot.

Yes.

> When using the value from multiboot2 as far as I can tell acpi_table_init does not get called, and we are just using the raw pointer. It isn't clear to me why that is guaranteed to be mapped in?

Multiboot 2 provides a  copy of the ACPI RSDP data, directly embedded within the already mapped Multiboot 2 structure. So, the pointer points into the dense packed MBI2 structure. (This seems in general the main difference to MBI1, where all parts of MBI information may lie in arbitrary memory locations connected by pointers.)

alex-ab:
> Given you have already done the bulk of the implementation thus far I am happy to do these and leave your name on the commits if you do not have time.

I'm fine with that you take over, thank you.

src/plat/pc99/machine/acpi.c
AdrianDanis: Ahh I didn't pick this up from my quick skim of the multiboot2 spec. Thanks.

AdrianDanis:
Closing now that this has been added in 4241b2df6fb7cd182277d984324c0b3186f4588c

PR #74, opened 2017-10-20
ZynqMP Fixes

natestuder:
A couple of fixes for the recently added ZynqMP support based on testing of the port by I-A-I.


AdrianDanis:
@natestuder Changing the clock checks to a warning seems uncontroversial to me, but I will object to the autoconf.h change as building standalone with `make` is not supported, and is about to be stripped (in favor of a new standalone build style).

natestuder:
I've no problem with dropping the autoconf.h patch.  We only added it to be consistent with other platforms.

AdrianDanis:
No worries. Will pull this in then.

PR #77, opened 2017-11-20
x86: add framebuffer info of Multiboot 2 via boot_info

alex-ab:
In UEFI boot mode, the framebuffer information of Multiboot 2 contains the information (physical address of the framebuffer, resolution, color depth) of the already initialized framebuffer and can be used in user-mode to setup a simple framebuffer driver. The commits propagates the information through the kernel to the roottask, which then may delegate it further to the simple framebuffer driver (as used e.g. by Genode in UEFI boot case).


AdrianDanis:
Now that your other PR has finally gone through I'll pull this in, sorry for the delay

alex-ab:
Thanks, no problem, we have no time pressure ;-)

PR #81, opened 2018-03-12
Add thread name to debug info for SysDebugHalt and SysDebugSnapshot

mokshasoft:
This will make it easier to understand which tread that triggered these debugging syscalls. Similar to the print in kernel/src/kernel/faulthandler.c. I've run sel4test on config x64_simulation_debug_xml_defconfig with all tests passing.
```
Starting test 102: Test all tests ran
Test suite passed. 102 tests passed. 46 tests disabled.
All is well in the universe
```


AdrianDanis:
Looks fine, will pull it in. Thanks.

PR #82, opened 2018-03-26
Fix trivial comment typos.

waywardmonkeys:


AdrianDanis:
@waywardmonkeys Thanks for the fixes, and for sending in the CLA.
Looks like we should be more aggressive on spell checking.

AdrianDanis:
[Committed a rebased version](https://github.com/seL4/seL4/commit/b942a5044119ed75ad8f607618ca069a0d3f0aa9)

PR #91, opened 2018-05-10
Riscv32 Boot Fixes

natestuder:
Small changes to get sel4test to boot on riscv32.  Not completely sure if this is the right way to do this, but it works.


AdrianDanis:
@natestuder Thanks for the fixes. The changes look pretty good to me. Will merge these shortly, need to setup CI for rv32 internally first.

PR #92, opened 2018-05-21
riscv: Enable seL4 to run in Machine Mode (Highest Privilege)

heshamelmatary:
* This allows for future virtualization efforts (e.g. running Linux in S-Mode)
* User level is not affected by this change
* Kernel works in physical memory (doesn't use MMU or TLBs)
* No "global" kernel mappings in this mode
* User can use the entire virtual address space
* In the future, this allows the seL4 to use and/or export devices (like UART) without the need to trap to riscv-pk

### DIFF
--- a/include/arch/riscv/arch/encoding.h
+++ b/include/arch/riscv/arch/encoding.h
@@ -304,6 +304,9 @@
 #define rdcycle() read_csr(cycle)
 #define rdinstret() read_csr(instret)
 
+#define get_cycles() (uint64_t) (rdcycle() | (config_set(CONFIG_ARCH_RISCV_RV32)? (uint64_t) (read_csr(cycleh) << 32) : 0))
+#define get_time() (uint64_t) (rdtime() | (config_set(CONFIG_ARCH_RISCV_RV32)? (uint64_t) (read_csr(timeh) << 32) : 0))
+
 #endif
 
 #endif
--- a/include/arch/riscv/arch/encoding.h
+++ b/include/arch/riscv/arch/encoding.h
@@ -24,11 +24,23 @@
  * MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
  */
 
+/*
+ * Copyright (c) 2018, Hesham Almatary <Hesham.Almatary@cl.cam.ac.uk>
+ * All rights reserved.
+ *
+ * This software was was developed in part by SRI International and the University of
+ * Cambridge Computer Laboratory (Department of Computer Science and
+ * Technology) under DARPA contract HR0011-18-C-0016 ("ECATS"), as part of the
+ * DARPA SSITH research programme.
+ */
+
 /* This file is copied from RISC-V tools, it might change for
  * new spec releases.
  */
 #ifndef RISCV_CSR_ENCODING_H
 #define RISCV_CSR_ENCODING_H
+
+#include <config.h>
 
 #define MSTATUS_UIE         0x00000001
 #define MSTATUS_SIE         0x00000002
@@ -215,9 +227,59 @@
 #define RISCV_PGSHIFT 12
 #define RISCV_PGSIZE (1 << RISCV_PGSHIFT)
 
+/* Similar to ecall instruction (environment call), pre-fix registers according to
+ * the RISC-V mode seL4 is working in (Machine or Supervisor)
+ */
+#ifdef CONFIG_SEL4_RV_MACHINE
+#define PREFIX_ENV(token) m ## token
+#define ENV_STRING "m"
+#else
+#define PREFIX_ENV(token) s ## token
+#define ENV_STRING "s"
+#endif
+
+#define EPC_REG     PREFIX_ENV(epc)
+#define TVEC_REG    PREFIX_ENV(tvec)
+#define STATUS_REG  PREFIX_ENV(status)
+#define SCRATCH_REG PREFIX_ENV(scratch)
+#define BADADDR_REG PREFIX_ENV(badaddr)
+#define CAUSE_REG   PREFIX_ENV(cause)
+#define IE_REG      PREFIX_ENV(ie)
+#define IP_REG      PREFIX_ENV(ip)
+#define RET_INSN    PREFIX_ENV(ret)
+
 #ifndef __ASSEMBLER__
 
 #ifdef __GNUC__
+
+#define EEPC      STRINGIFY(EPC_REG)
+#define ECAUSE    STRINGIFY(CAUSE_REG)
+#define EBADADDR  STRINGIFY(BADADDR_REG)
+#define ESTATUS   STRINGIFY(STATUS_REG)
+#define ESCRATCH  STRINGIFY(SCRATCH_REG)
+#define ETVEC     STRINGIFY(TVEC_REG)
+#define EIE       STRINGIFY(IE_REG)
+#define EIP       STRINGIFY(IP_REG)
+#define ERET      STRINGIFY(RET_INSN)
+
+#define read_csr_env(reg) ({ unsigned long __tmp; \
+  asm volatile ("csrr %0, " ENV_STRING#reg : "=r"(__tmp)); \
+  __tmp; })
+
+#define clear_csr_env(reg, bit) ({ unsigned long __tmp; \
+  asm volatile ("csrrc %0, " ENV_STRING#reg ", %1" : "=r"(__tmp) : "rK"(bit)); \
+  __tmp; })
+
+#define write_csr_env(reg, val) ({ \
+  asm volatile ("csrw " ENV_STRING#reg ", %0" :: "rK"(val)); })
+
+#define swap_csr_env(reg, val) ({ unsigned long __tmp; \
+  asm volatile ("csrrw %0, " ENV_STRING#reg ", %1" : "=r"(__tmp) : "rK"(val)); \
+  __tmp; })
+
+#define set_csr_env(reg, bit) ({ unsigned long __tmp; \
+  asm volatile ("csrrs %0, " ENV_STRING#reg ", %1" : "=r"(__tmp) : "rK"(bit)); \
+  __tmp; })
 
 #define read_csr(reg) ({ unsigned long __tmp; \
   asm volatile ("csrr %0, " #reg : "=r"(__tmp)); \
--- a/include/arch/riscv/arch/sbi.h
+++ b/include/arch/riscv/arch/sbi.h
@@ -24,6 +24,16 @@
  * MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR MODIFICATIONS.
  */
 
+/*
+ * Copyright (c) 2018, Hesham Almatary <Hesham.Almatary@cl.cam.ac.uk>
+ * All rights reserved.
+ *
+ * This software was was developed in part by SRI International and the University of
+ * Cambridge Computer Laboratory (Department of Computer Science and
+ * Technology) under DARPA contract HR0011-18-C-0016 ("ECATS"), as part of the
+ * DARPA SSITH research programme.
+ */
+
 /* This file is copied from RISC-V tools/linux project, it might change for
  * new spec releases.
  */
@@ -31,6 +41,7 @@
 #ifndef _ASM_RISCV_SBI_H
 #define _ASM_RISCV_SBI_H
 
+#include <config.h>
 #include <stdint.h>
 
 #define SBI_SET_TIMER 0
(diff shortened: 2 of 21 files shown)


gernotheiser:
Nice, thanks!

I only last week said we needed a PoC implementation of this, to compare possible virtualisation approaches.

Gernot

On 22 May 2018, at 02:41, Hesham Almatary <notifications@github.com<mailto:notifications@github.com>> wrote:


  *   This allows for future virtualization efforts (e.g. running Linux in S-Mode)
  *   User level is not affected by this change
  *   Kernel works in physical memory (doesn't use MMU or TLBs)
  *   No "global" kernel mappings in this mode
  *   User can use the entire virtual address space
  *   In the future, this allows the seL4 to use and/or export devices (like UART) without the need to trap to riscv-pk

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/seL4/seL4/pull/92

Commit Summary

  *   riscv: Enable seL4 to run in Machine Mode (Highest Privilege)

File Changes

  *   M include/arch/riscv/arch/encoding.h<https://github.com/seL4/seL4/pull/92/files#diff-0> (68)
  *   M include/arch/riscv/arch/fastpath/fastpath.h<https://github.com/seL4/seL4/pull/92/files#diff-1> (8)
  *   M include/arch/riscv/arch/machine/registerset.h<https://github.com/seL4/seL4/pull/92/files#diff-2> (24)
  *   M include/arch/riscv/arch/sbi.h<https://github.com/seL4/seL4/pull/92/files#diff-3> (31)
  *   M include/plat/spike/plat/32/plat_mode/machine/hardware.h<https://github.com/seL4/seL4/pull/92/files#diff-4> (15)
  *   M include/plat/spike/plat/64/plat_mode/machine/hardware.h<https://github.com/seL4/seL4/pull/92/files#diff-5> (14)
  *   M include/plat/spike/plat/machine.h<https://github.com/seL4/seL4/pull/92/files#diff-6> (27)
  *   M include/plat/spike/plat/machine/hardware.h<https://github.com/seL4/seL4/pull/92/files#diff-7> (23)
  *   M src/arch/riscv/c_traps.c<https://github.com/seL4/seL4/pull/92/files#diff-8> (18)
  *   M src/arch/riscv/kernel/boot.c<https://github.com/seL4/seL4/pull/92/files#diff-9> (28)
  *   M src/arch/riscv/kernel/thread.c<https://github.com/seL4/seL4/pull/92/files#diff-10> (19)
  *   M src/arch/riscv/kernel/vspace.c<https://github.com/seL4/seL4/pull/92/files#diff-11> (24)
  *   M src/arch/riscv/machine/hardware.c<https://github.com/seL4/seL4/pull/92/files#diff-12> (12)
  *   M src/arch/riscv/machine/io.c<https://github.com/seL4/seL4/pull/92/files#diff-13> (25)
  *   M src/arch/riscv/machine/registerset.c<https://github.com/seL4/seL4/pull/92/files#diff-14> (12)
  *   M src/arch/riscv/traps.S<https://github.com/seL4/seL4/pull/92/files#diff-15> (17)
  *   M src/plat/spike/linker.lds<https://github.com/seL4/seL4/pull/92/files#diff-16> (23)
  *   M src/plat/spike/machine/hardware.c<https://github.com/seL4/seL4/pull/92/files#diff-17> (43)

Patch Links:

  *   https://github.com/seL4/seL4/pull/92.patch
  *   https://github.com/seL4/seL4/pull/92.diff

AdrianDanis:
@heshamelmatary My main dislike of this design is that seL4 is still 'calling' to some other M-mode run time to perform actions. One of the advantages to putting seL4 in M-mode would be to improve the verification story to not be passing control to unknown and unverified code. Not that I have anything against some form of support for running in M-mode, just saying that as it stands I'm not sure what the use of this is. The most compelling item for me is that it *may* have better performance, but this depends on the actual silicon implementations.

As for the changes themselves I'm first going to ask that you please break this up into a sensible series of commits for review.

Also I do not see a CLA from you. Have you sent one in?

heshamelmatary:
@AdrianDanis That's one of the future goals as I mentioned in the commit message: not to trap to riscv-pk. However, I'd do this for a real hardware but not for Spike (as a simulator platform). The reason not to do it for Spike (being the golden model of RISC-V) is to keep it as a reference RISC-V platform while avoiding maintaining platform changes on both Spike and QEMU. Spike and QEMU keep changing on the platform-level, riscv-pk is kept synced (by the maintainers) with them to add more devices/features (e.g. UART) and/or address platform backwards-incompatible changes (e.g. misa), so we get the good parts of riscv-pk while keeping the seL4-Spike platform portable and working as a reference try-it port, without having to worry much about tools, simulators and HW platform<->M-mode SW compatibility. 

That's not the case for silicon implementations (like HiFive Unleashed, which I'm working on), which I totally agree to get rid of riscv-pk for as the platform/board won't change, and performance would matter. That's mainly why I am submitting this PR besides thinking it might be beneficial to the community to build on or make use of.

I'll follow up on the CLA and break up commits

heshamelmatary:
I'm waiting for CLA approval from my department

src/arch/riscv/kernel/boot.c
```diff
@@ -275,14 +277,16 @@ try_init_kernel(
...
-    init_cpu();
+    if (!config_set(CONFIG_SEL4_RV_MACHINE)) {
+        map_kernel_window();
+    }
 
     /* initialize the platform */
     init_plat(dtb_reg);
 
+    /* initialise the CPU */
+    init_cpu();
```
AdrianDanis: It's not clear to me why moving `init_cpu` is required to not map the kernel window in M-mode

src/arch/riscv/kernel/boot.c
```diff
@@ -214,11 +214,13 @@ extern char trap_entry[];
 BOOT_CODE static void
 init_cpu(void)
 {
-
     /* Write trap entry address to stvec */
-    write_csr(stvec, trap_entry);
+    write_csr_env(tvec, trap_entry);
```
AdrianDanis: You had a previous commit that claimed "Replace S-mode instructions/registers with privilege-aware definitions", seems like this should have been part of that?

include/arch/riscv/arch/sbi.h
```diff
@@ -43,6 +54,25 @@
 #define SBI_REMOTE_SFENCE_VMA_ASID 7
 #define SBI_SHUTDOWN 8
 
+
+
+#ifdef CONFIG_SEL4_RV_MACHINE
+extern char trap_entry[];
+#define SBI_CALL(which, arg0, arg1, arg2) ({ 			\
+        unsigned long mtvec = read_csr(mtvec);                 \
+        write_csr(mtvec, 0x0000000080000004ul);                     \
```
AdrianDanis: Is this a spike specific physical address?

include/plat/spike/plat/machine/hardware.h
```diff
@@ -36,20 +46,33 @@
...
+#define BASE_OFFSET 0
+#define KERNEL_BASE_OFFSET 0
+#define PPTR_TOP PADDR_TOP
+#define PPTR_USER_TOP PADDR_TOP
+#else
 #define PPTR_TOP KERNEL_BASE
 #define PPTR_USER_TOP PPTR_BASE
 #define BASE_OFFSET (PPTR_BASE - PADDR_BASE)
+/* Translates from a physical address and a value in the kernel image */
+#define KERNEL_BASE_OFFSET (KERNEL_BASE - PADDR_LOAD)
```
AdrianDanis: I don't understand why this `#define` is special cased. Doing so implies that `KERNEL_BASE - PADDR_LOAD` is *not* going to be 0 in M-mode, but surely this can only be correct if `KERNEL_BASE` and `PADDR_LOAD` are the same value? Am I missing something in the memory layout for the kernel in M-mode?

src/plat/spike/machine/hardware.c
```diff
@@ -62,8 +73,7 @@ interrupt_t
 getActiveIRQ(void)
 {
 
-    uint64_t temp = 0;
-    asm volatile ("csrr %0, scause":"=r" (temp)::);
+    word_t temp = read_csr_env(cause);
```
AdrianDanis: You had a previous commit that "Replace S-mode instructions/registers with privilege-aware definitions", should this not also be in there?

src/plat/spike/machine/hardware.c
```diff
@@ -140,24 +140,35 @@ ackInterrupt(irq_t irq)
 }
 
 static inline uint64_t get_cycles(void)
-#if __riscv_xlen == 32
```
AdrianDanis: Not sure how this change is related to trapping to riscv-pk

src/plat/spike/machine/hardware.c
```diff
@@ -140,24 +140,35 @@ ackInterrupt(irq_t irq)
...
 {
+    UNUSED unsigned long mtvec;
+
+    if (config_set(CONFIG_SEL4_RV_MACHINE)) {
+        /* Trap to riscv-pk to get time */
+        mtvec = read_csr_env(tvec);
+        write_csr(mtvec, 0x80000004ul);
```
AdrianDanis: I find this a little confusing. On the one hand you `read_csr_env`, but then you explicitly write to mtvec, and then later on you put this value that you read explicitly into mtvec. I realize that all this code is guarded by `CONFIG_SEL4_RV_MACHINE`, but you should still be consistent if you use register abstractions or not

src/plat/spike/machine/hardware.c
```diff
@@ -140,24 +140,35 @@ ackInterrupt(irq_t irq)
...
+    }
+#ifdef CONFIG_ARCH_RISCV_RV32
     uint32_t nH, nL;
     __asm__ __volatile__ (
         "rdtimeh %0\n"
         "rdtime  %1\n"
         : "=r" (nH), "=r" (nL));
+    if (config_set(CONFIG_SEL4_RV_MACHINE)) {
+        write_csr(mtvec, mtvec);
+    }
```
AdrianDanis: I would suggest having the rv32 and rv64 arms of this code both produce a return value and then have a single rest of mtvec at the end of the function prior to the return

src/plat/spike/machine/hardware.c
```diff
@@ -140,24 +140,35 @@ ackInterrupt(irq_t irq)
 }
 
 static inline uint64_t get_cycles(void)
-#if __riscv_xlen == 32
 {
+    UNUSED unsigned long mtvec;
+
+    if (config_set(CONFIG_SEL4_RV_MACHINE)) {
+        /* Trap to riscv-pk to get time */
+        mtvec = read_csr_env(tvec);
```
AdrianDanis: nitpick: I would discourage calling local variables by the same name as a machine register as it makes things like `write_csr(mtvec, mtvec)` confusing as the pretense is that 'mtvec' is some kind of magical global symbol and `write_csr` is a function that will write to it. Obviously it is a macro and this code will work, but it's very jarring to read and breaks the illusion that these macros are trying to create. Also it meant I had to dance around language in writing these comments as saying "and later you write mtvec into mtvec" suddenly becomes a bit ambiguous

include/plat/spike/plat/machine/hardware.h
heshamelmatary: You are right, it's 0 in machine mode. I just added it to avoid confusion (and also for someone who's reading the code). Furthermore, the comment doesn't apply in machine mode as well, so I decided to make them explicitly separate.

src/plat/spike/machine/hardware.c
heshamelmatary: I separate commits/changes in the architecture (starting with "riscv/arch:"), from changes in the platform ("spike:")

include/arch/riscv/arch/sbi.h
heshamelmatary: Spike + riscv-pk. It's the value riscv-pk sets before going loading the payload. Yes, it's physical. I made a change to save riscv-pk's mtval on boot time. This should make seL4/spike plat more portable in case this address has ever changed by pk/spike (for any reason).

https://github.com/riscv/riscv-pk/blob/master/machine/mentry.S#L34

heshamelmatary:
@AdrianDanis I think I addressed all your comments that I didn't reply to. Please let me know if you've any more comments regarding the new changes

include/plat/spike/plat/machine/hardware.h
AdrianDanis: I think the comment does apply. Code in the kernel generally isn't M-mode specific and so does need to translate from a physical address to the kernel image virtual address, which is what the `KERNEL_BASE_OFFSET` macro does. The translation being the identify function in the case of M-mode does not make it not a translation.

Having separate definitions is more confusing as it forces the question of why do they need to defined separately.

Similarly I would define a correct value for `PPTR_BASE`, not just because it allows for a consistent `BASE_OFFSET` definition to be used, but also this is a value that has meaning and should be defined correctly. Just because it is currently only used in kernel window mapping code, which isn't being executed, doesn't mean it shouldn't be defined correctly.

include/arch/riscv/arch/encoding.h
```diff
@@ -304,6 +304,9 @@
 #define rdcycle() read_csr(cycle)
 #define rdinstret() read_csr(instret)
 
+#define get_cycles() (uint64_t) (rdcycle() | (config_set(CONFIG_ARCH_RISCV_RV32)? (uint64_t) (read_csr(cycleh) << 32) : 0))
+#define get_time() (uint64_t) (rdtime() | (config_set(CONFIG_ARCH_RISCV_RV32)? (uint64_t) (read_csr(timeh) << 32) : 0))
```
AdrianDanis: I don't understand why these are macros and not functions. This just seems to be making them unnecessarily cryptic. Also not convinced these are to do with encodings and probably don't belong in this header.

src/arch/riscv/model/statedata.c
```diff
@@ -28,3 +29,6 @@
 asid_pool_t *riscvKSASIDTable[BIT(asidHighBits)];
 
 pte_t kernel_pageTables[CONFIG_PT_LEVELS][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+
+/* Address of pk trap address */
+UNUSED word_t pk_trap_addr;
```
AdrianDanis: Not sure how `kernel_pageTables` above snuck through but the naming convention in statedata is `archKSStateName`, see `riscvKSASIDTable` further up and the definitions in all the other statedata

src/arch/riscv/kernel/boot.c
```diff
@@ -214,13 +214,15 @@ extern char trap_entry[];
 BOOT_CODE static void
 init_cpu(void)
 {
-    /* Write trap entry address to stvec */
-    write_csr_env(tvec, trap_entry);
```
AdrianDanis: I realize that it does not matter whether we write the trap entry address before or after we `activate_kernel_vspace` but reordering this with no justification in a commit that claims to only "Save the trap address" is very poor style

PR #94, opened 2018-07-22
Rocket-Chip support for RISCV 64-bit Spike

chrisguikema:
These commits add an option to build the 64-bit spike for the Rocket-Chip hardware (https://github.com/ucb-bar/fpga-images-zedboard/tree/7e0080ee9e29e399d80d74d9756f20cece7aeef0)

They add a config option for the cmake system, and update the load address to be closer to the RAM_BASE since the hardware only has 256MiB of space. This means that the address is no longer aligned on a 1GiB boundary, so the kernel had to be changed to support 2nd level mappings.

The vspace changes clean up the existing code by using an enum and defines, which make it a little more readable. Then when mapping, it checks to see if both the physical and virtual addresses are aligned on a 1GiB boundary. If so, the behavior doesn't change. However, if they aren't, then the system checks for a 2MiB boundary alignment, and uses the second level page table array that we created to fill in the gaps. So these changes should work when the load address isn't aligned on a 1GiB boundary, even when there is more than 1GiB of RAM available to the system.

There is also one of Hesham's commits in this pull request, since we determined the fence instruction was required to run on actual hardware, else thread's vspaces would get corrupted.

This was tested on the rocket-chip with the link to the open-source files above. We were able to run sel4test on the hardware with no issues provided that the corresponding elfloader (super important) and sel4test (fixed 3 failing tests) commits were included.


include/arch/riscv/arch/model/statedata.h
```diff
@@ -35,8 +37,47 @@ NODE_STATE_END(archNodeState);
...
 
+/* Kernel Page Tables */
 /* RISC-V has n-level page tables (depending on configured paging mode),
- * each of which is 4KiB, PTEs are of word_t size
-  */
-extern pte_t kernel_pageTables[CONFIG_PT_LEVELS][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+ *  As of version 1.10,
+ *  - level 1 are entries for 1GiB pages
+ *  - level 2 are entries for 2MiB pages
+ *  - level 3 are entries for 4KiB pages
```
heshamelmatary: This comment doesn't apply to the intended "general" n-level design of the kernel page tables that can work for Sv32, Sv39, Sv48 or future ones. Maybe move to a PT == 3 ```#ifdef```?

include/plat/spike/plat/machine/hardware.h
```diff
@@ -31,10 +31,15 @@
 #define PADDR_BASE 0x0lu
 #endif
 
+#ifndef CONFIG_BUILD_ROCKET_CHIP
```
heshamelmatary: Having this line (i.e. the _#ifndef_) priorities _CONFIG_BUILD_ROCKET_CHIP_ over any other, future, platforms. I'd suggest you _#ifdef_ for each specific platform. i.e.:

```
#ifdef CONFIG_BUILD_ROCKET_CHIP

#elif CONFIG_PLAT_SPIKE

#elif CONFIG_PLAT_XXX

#else 
#error "Not able to determine a platform
#endif
```

include/plat/spike/plat/machine/hardware.h
```diff
@@ -31,10 +31,15 @@
...
 /* This represents the physical address that the kernel image will be linked to. This needs to
  * be on a 1gb boundary as we currently require being able to creating a mapping to this address
  * as the largest frame size */
 #define PADDR_LOAD 0xC0000000lu
+#else
+/* The Rocket-Chip hardware only has 256MiB of Memory. */
```
heshamelmatary: The ```Rocket-Chip``` itself is a parameterised i.e. it's not restricted/bound to 256MiB of memory. For example, I am running the Rocket Chip on a board that has 1 GiB of memory. I'd suggest you rename the platform to your specific board that runs the Rocket Chip (e.g. ```PLAT_ROCKET_ZEDBOARD```). Otherwise, if you want to keep the config name as is, the Rocket Chip parameters should be also seL4 parameters to allow your code to run on other Rochet-Chip-based board.

[ditto](https://github.com/seL4/seL4/pull/94/files#r204326181)

src/arch/riscv/config.cmake
```diff
@@ -48,6 +50,12 @@ if (KernelSel4ArchRiscV32)
     set(KernelPTLevels 2 CACHE STRING "" FORCE)
 endif()
 
+config_option(KernelBuildRocketChip BUILD_ROCKET_CHIP "Build application to \
```
heshamelmatary: [ditto](https://github.com/seL4/seL4/pull/94/files#r204326181), please rename this config to your specific board

src/arch/riscv/kernel/vspace.c
```diff
@@ -71,52 +110,68 @@ map_kernel_window(void)
...
-        pptr += RISCV_GET_LVL_PGSIZE(1);
-        paddr += RISCV_GET_LVL_PGSIZE(1);
+
+#if CONFIG_PT_LEVELS == 3
+    word_t orig_indx_1GiB = RISCV_GET_PT_INDEX(pptr, PT_LEVEL_1);
+#endif
+
+    while (pptr < max_address && pptr >= KERNEL_BASE) {
+
+        /* if both phy and virt address are aligned on 1GiB boundary, use Level 1 mappings */
```
heshamelmatary: This comment only applies to Sv39

include/arch/riscv/arch/model/statedata.h
chrisguikema: I'll move the comment into the CONFIG_PT_LEVELS == 3 section immediately below this.

include/plat/spike/plat/machine/hardware.h
chrisguikema: Thats a good point about the #ifndef wrapper. The way we structured the ROCKET_CHIP option was to have it be an addition to the spike platform instead of its own unique platform, since so much of the code is identical. 

So in the gen_config.h file, CONFIG_PLAT_SPIKE is 1 as well as the CONFIG_BUILD_ROCKET_CHIP.

The way you have it could work, since the CONFIG_PLAT_SPIKE line wouldn't get processed, but this file would only get compiled if PLAT_SPIKE was selected.

I'll update it to use #ifdefs instead of #ifndefs

include/plat/spike/plat/machine/hardware.h
chrisguikema: will rename BUILD_ROCKET_CHIP to BUILD_ROCKET_CHIP_ZEDBOARD

src/arch/riscv/config.cmake
chrisguikema: will rename BUILD_ROCKET_CHIP to BUILD_ROCKET_CHIP_ZEDBOARD

src/arch/riscv/kernel/vspace.c
chrisguikema: I'll specify for Sv39, then add a line explaining that this checks for 2MiB alignment in Sv32.

src/arch/riscv/kernel/vspace.c
```diff
@@ -125,15 +129,51 @@ map_kernel_window(void)
...
-        assert(IS_ALIGNED(paddr, RISCV_GET_LVL_PGSIZE_BITS(PT_LEVEL_1)));
 
-        kernel_Level1_pageTable[RISCV_GET_PT_INDEX(pptr, PT_LEVEL_1)] = pte_next(paddr, true);
+#if CONFIG_PT_LEVELS == 3
+    word_t orig_indx_1GiB = RISCV_GET_PT_INDEX(pptr, PT_LEVEL_1);
+#endif
 
-        pptr += RISCV_GET_LVL_PGSIZE(PT_LEVEL_1);
-        paddr += RISCV_GET_LVL_PGSIZE(PT_LEVEL_1);
+    while (pptr < max_address && pptr >= KERNEL_BASE) {
```
kent-mcleod: I believe you can use `UINTPTR_MAX` here instead of max_address.

src/arch/riscv/kernel/vspace.c
```diff
@@ -63,6 +67,28 @@ RISCVGetReadFromVMRights(vm_rights_t vm_rights)
     return vm_rights != VMWriteOnly;
 }
 
+
+pte_t pte_next(word_t phys_addr, bool_t is_leaf)
```
kent-mcleod: If this helper is private to this file, can you declare it with `static` instead of having the function prototype?

include/arch/riscv/arch/model/statedata.h
```diff
@@ -35,8 +37,46 @@ NODE_STATE_END(archNodeState);
 
 extern asid_pool_t *riscvKSASIDTable[BIT(asidHighBits)];
 
+/* Kernel Page Tables */
+extern pte_t kernel_Level1_pageTable[BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
```
kent-mcleod: Note: I'm mostly just pattern matching from the arm/../statedata.h here:
This should have the VISIBLE attribute, and I don't think that ALIGN has any effect on the extern declaration.

src/arch/riscv/model/statedata.c
```diff
@@ -27,4 +29,8 @@
 /* The top level asid mapping table */
 asid_pool_t *riscvKSASIDTable[BIT(asidHighBits)];
 
-pte_t kernel_pageTables[CONFIG_PT_LEVELS][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+/* Kernel Page Tables */
+pte_t kernel_Level1_pageTable[BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+#if CONFIG_PT_LEVELS == 3
+pte_t kernel_Level2_Tables[NUM_2MiB_ENTRIES][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
```
kent-mcleod: These should be `ALIGN_BSS` instead of `ALIGN`

src/arch/riscv/config.cmake
```diff
@@ -48,6 +50,12 @@ if (KernelSel4ArchRiscV32)
     set(KernelPTLevels 2 CACHE STRING "" FORCE)
 endif()
 
+config_option(KernelBuildRocketChipZedboard BUILD_ROCKET_CHIP_ZEDBOARD "Build application to \
```
kent-mcleod: If it is specific to the Spike platform, this option can be moved to src/plat/spike/config.cmake.
Additionally, renaming it to something like KernelPlatformSpikeRocketChip will be more consistent with how other platforms and boards are named.

src/arch/riscv/kernel/vspace.c
chrisguikema: Awesome. I figured there was an easier way. I'll update that and remove the `max_address` calculation

include/arch/riscv/arch/model/statedata.h
chrisguikema: Good to know. I'll add the `VISIBLE` tag and remove the alignment in the declaration

src/arch/riscv/config.cmake
chrisguikema: Yeah, it is specific to the Spike platform. I'll move it to the suggested location and update the cmake option, leaving the `BUILD_ROCKET_CHIP_ZEDBOARD` output alone.

chrisguikema:
I'll look through the commit messages and fix the typos.

As for hesham's commit, it is necessary for it to run on the rocket-chip hardware. I understand about the CLA, since it was done on his own time. We didn't want to steal the commit right from him, especially since its a single line of code.

What are the other alternatives? We could make it our own commit and reference his work in the commit message?

pingerino:
@chrisguikema given Heshams commit is small you can make your own commit. AFAIK @heshamelmatary is working on getting a CLA signed.

pingerino:
@chrisguikema we have a `make style` target in the kernel. Can you please run this over your PR so that the commits follow the style guide?

src/arch/riscv/kernel/vspace.c
```diff
@@ -37,6 +39,13 @@
 #include <plat/machine/hardware.h>
 #include <kernel/stack.h>
 
+
+typedef enum PAGETABLE_LEVEL_E {
+        PT_LEVEL_1 = 1,
+        PT_LEVEL_2 = 2,
+        PT_LEVEL_3 = 3,
+} ptlevel_t;
+
```
lsf37: Can I ask what the intended use is for this enum? The PT design for RISC-V is generic in the number of levels, i.e. the only level one *should* need a symbolic name for is the top level, the rest should be computed. This is different to how it works on all other platforms in seL4, but since RISC-V is very regular, we're trying to take advantage of that in the (ongoing) verification. If at all possible, I'd very much like to remove the temptation for people to talk about specific levels elsewhere in the code.

I can see that this might be useful in platform-specific init code, but since PT_LEVEL_2 really is just 2, I think it'd be justified to not use an enum for it.

include/arch/riscv/arch/machine.h
```diff
@@ -117,6 +117,9 @@ static inline void setVSpaceRoot(paddr_t addr, asid_t asid)
                            asid,                         /* asid */
                            addr >> seL4_PageBits); /* PPN */
 
+    /* Make sure are reads/writes are ordered before writing sptbr*/
+    sfence();
```
heshamelmatary: Even though this is just one line, but I'm sure @chrisguikema is aware how vital it's to run on actual HW. @pingerino  I'm not sure when will I get the CLA, so I'm happy with @chrisguikema authoring it not to block this PR

src/arch/riscv/kernel/vspace.c
```diff
@@ -124,15 +124,51 @@ map_kernel_window(void)
...
+            }
+
+            kernel_Level2_Tables[level2_ref][indx_2MiB] = pte_next(paddr, true);
+            pptr += RISCV_GET_LVL_PGSIZE(PT_LEVEL_2);
+            paddr += RISCV_GET_LVL_PGSIZE(PT_LEVEL_2);
+        }
+#endif
+        else
+        {
+            assert("Kernel not properly aligned");
```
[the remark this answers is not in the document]
chrisguikema: good to know. I can update `assert` -> `fail`

src/arch/riscv/kernel/vspace.c
```diff
@@ -37,6 +39,7 @@
 #include <plat/machine/hardware.h>
 #include <kernel/stack.h>
 
+
```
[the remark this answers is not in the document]
chrisguikema: totally understandable about the whitespace only changes. We usually try to avoid those as well, This one must have slipped through the cracks. I'll make sure to remove it.

include/arch/riscv/arch/model/statedata.h
```diff
@@ -35,8 +37,46 @@ NODE_STATE_END(archNodeState);
...
+/* Kernel Page Tables */
+extern pte_t kernel_Level1_pageTable[BIT(PT_INDEX_BITS)] VISIBLE;
+
+#if CONFIG_PT_LEVELS == 3
+
 /* RISC-V has n-level page tables (depending on configured paging mode),
- * each of which is 4KiB, PTEs are of word_t size
-  */
-extern pte_t kernel_pageTables[CONFIG_PT_LEVELS][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+ *  As of version 1.10,
```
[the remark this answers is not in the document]
chrisguikema: I'll update the comment to specify the RISC-V Priv Spec

include/arch/riscv/arch/model/statedata.h
```diff
@@ -35,8 +37,46 @@ NODE_STATE_END(archNodeState);
...
+        +----------------+           [...]|    |    |    |    |   | |
+        |                |              | +-----------------------+ |
+        +----------------+              | +-----------------------+ |
+                      [NUM_2MiB_ENTRIES-1]|    |    |    |    |   | |
+                                        | +-----------------------+ |
+                                        +---------------------------+
+ */
+
+/* Kernel has Upper 2GiB of Virtual Memory. 512 Entries per PTE * 2MiB = 1GiB, so 2 of these are needed */
+#define NUM_2MiB_ENTRIES 2
```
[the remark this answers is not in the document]
chrisguikema: I'll update

src/arch/riscv/model/statedata.c
```diff
@@ -27,4 +29,8 @@
 /* The top level asid mapping table */
 asid_pool_t *riscvKSASIDTable[BIT(asidHighBits)];
 
-pte_t kernel_pageTables[CONFIG_PT_LEVELS][BIT(PT_INDEX_BITS)] ALIGN(BIT(seL4_PageTableBits));
+/* Kernel Page Tables */
+pte_t kernel_Level1_pageTable[BIT(PT_INDEX_BITS)] ALIGN_BSS(BIT(seL4_PageTableBits));
```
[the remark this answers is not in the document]
chrisguikema: i'll update everything to `kernel_level1_pageTable`

include/plat/spike/plat/machine/hardware.h
```diff
@@ -31,10 +31,15 @@
 #define PADDR_BASE 0x0lu
 #endif
 
+#ifdef CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD
```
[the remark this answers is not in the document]
chrisguikema: @heshamelmatary might have some better feedback than me, but really the only change needed to run on the rocket-chip was the address change, which required all of the mapping changes. And as hesham pointed out, if you're running a rocket-chip that has enough memory, you really wouldn't need any changes at all (except for the `fence` instruction)

include/arch/riscv/arch/machine.h
chrisguikema: @heshamelmatary Thank you for the feedback. I'll make the same commit and mention you in the commit message.

For future reference, you can run a single thread on actual hardware without this change. Once you add a second thread, I noticed that the second thread would run for one tick, and then the system would fault into the `idle` thread.

src/arch/riscv/kernel/vspace.c
```diff
@@ -40,6 +40,12 @@
 #include <kernel/stack.h>
 
 
+typedef enum PAGETABLE_LEVEL_E {
```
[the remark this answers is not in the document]
chrisguikema: Thanks for the information. Based on @lsf37's feedback, I don't think this `enum` is needed, since the code is already structured for the generic RISC-V page tables.

src/arch/riscv/kernel/vspace.c
chrisguikema: The purpose of the enum was to avoid the use of "magic numbers". However, we definitely aren't as familiar with how the RISC-V code is structured. We can definitely remove the use of this enum in the kernel code, since the code is already structured to be "SV Independent". 

I can especially see when compiling for 32-bit mode how `PT_LEVEL_1` would be confusing.

chrisguikema: Since you don't want levels discussed in the code, I also renamed the top level array to `kernel_toplevel_pageTable`

include/arch/riscv/arch/machine.h
heshamelmatary: Correct, I have debugged this issue for a while. To be more specific, it happens when having more than one thread with different ASIDs, after the first context switch. That's a TLB inconsistency issue. The riscv-priv manual suggests it *might* be needed before writing to ```satp``` register. I've had the impression that having an ```sfence``` after each ```[un]map``` and ```remap```  operation that should be sufficient, but that's not the case on Rocket

src/arch/riscv/kernel/vspace.c
heshamelmatary: nitpick: ```kernel_root_pageTable``` would also follow both riscv and seL4 conventions. ```level1``` is also fine as it exists for all Sv modes

chrisguikema: I updated to `kernel_toplevel_pageTable`

include/plat/spike/plat/machine/hardware.h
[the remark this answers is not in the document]
heshamelmatary: Yeah, I agree with @chrisguikema. I've done the same for SiFiveU. If there's a way in the kernel to make platforms share code or some sort of inheritance that would be great. In an ideal world, ```Spike``` platform, being a "reference" implementation, should work out of the box on actual HW if DTB is properly parsed and handled.

src/arch/riscv/kernel/vspace.c
chrisguikema: sorry @heshamelmatary, I didn't see that you had commented on this change. I'll update to `root_pageTable` as that makes the most sense.

chrisguikema:
@pingerino I've run the `make style` in the kernel. There were a few changes in the `vspace.c` file. I believe that I've addressed all of the comments.

chrisguikema:
Any update on this pull request?

chrisguikema:
So its been about two weeks since I've heard back from you guys on this pull request. Do you have any updates?

pingerino:
Hi @chrisguikema , I've just merged all of these internally, and the changes should come through after it passes internal CI.

Note that we updated the kernel PR a fair bit to adhere to our (not really documented) conventions in the boot code, specifically to make the code more static and less dynamic, and to make existing assumptions clearer. 

In some cases github might not detect the merge of the PR due to history rewrites, if so I'll point on here to the merged commits and close the PR manually. 

Thanks for your contributions.

pingerino:
This is now in master

PR #96, opened 2018-09-13
More RISC-V Changes

chrisguikema:
During the last pull request, there were some comments about the Rocket-Chip sharing a bunch of code with the Spike platform. We updated cmake a little bit to allow for different instances of Spike, since platforms will have hardware specific options, like the timebase clock used for the scheduling tick. 

The other big change is moving the `sfence` call after `sptbr` has been updated. This wasn't an issue while running seL4test on the zedboard, but it became an issue when developing other applications, since there would be VM faults on context switches. This order is also how Linux does it, so it is either an ISA issue or a Rocket-Chip issue.

### DIFF
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -117,13 +117,13 @@ static inline void setVSpaceRoot(paddr_t addr, asid_t asid)
                            asid,                         /* asid */
                            addr >> seL4_PageBits); /* PPN */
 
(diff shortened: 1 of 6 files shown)


chrisguikema:
We opened an issue on the RISC-V ISA Manual repository, and they answered here.

https://github.com/riscv/riscv-isa-manual/issues/226

chrisguikema:
Yeah, that is perfectly fine!

PR #98, opened 2018-09-14
zynqmp: add Ultra96 platform

axel-h:
Add Ultra96 platform, which is based on the zynqmp SoC

### DIFF
--- a/include/plat/zynqmp/plat/machine/devices.h
+++ b/include/plat/zynqmp/plat/machine/devices.h
@@ -27,7 +27,11 @@
 #define GIC_PL390_DISTRIBUTOR_PPTR  GIC_DISTRIBUTOR_PPTR
 
 /* These devices are used by the seL4 kernel. */
+#ifdef CONFIG_PLAT_ZYNQMP_ULTRA96
+#define UART_PADDR                  UART1_PADDR
+#else
 #define UART_PADDR                  UART0_PADDR
+#endif
 
 #define ACPU_GIC_PADDR              0xF9000000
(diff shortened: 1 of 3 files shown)


axel-h:
Is it ok to introduce CONFIG_PLAT_ZYNQMP_ULTRA96 or should there be only CONFIG_PLAT_ULTRA96? I've seen that for iMX6 there is just CONFIG_PLAT_SABRE and CONFIG_PLAT_WANDQ, but no CONFIG_PLAT_IMX6_WANDQ or CONFIG_PLAT_IMX6_SABRE. For iMX7/Sabre there seems to be a CONFIG_PLAT_IMX7_SABRE.

axel-h:
I'm also wondering if `autoconf.h` should be included in `devices.h` to guarantee that `CONFIG_PLAT_ZYNQMP_ULTRA96` is defined. But looking at `include/plat/imx6/plat/machine/devices.h` it checks `CONFIG_PLAT_SABRE` also without including `autoconf.h. But this could be something to fix then, give that every file should better include the header it depends on and not expect others to do this.

pingerino:
Hi @axel-h, our ARM platform naming conventions aren't great (and ARM devices themselves aren't helping). 

From your commits it looks like the only difference is the UART, so this is essentially the same platform with a tweak to the config. Your solution seems fine to me.

pingerino:
also @axel-h yes, all files should include `<autoconf.h>`. Given the way the kernel builds it can easily be forgotten and everything still works.

PR #111, opened 2019-01-10
tools: BSD compat for changed.sh

Smattr:
On e.g. FreeBSD, Bash lives at /usr/local/bin/bash. This change makes the script
find Bash wherever it may be hiding.

----

This script is quite handy outside the context of seL4 as well; thanks :) If you decide this change makes sense, it may be more consistent to change the shebang on all the shell scripts in a single commit. On the other hand, maybe it's better to avoid any indications that the build system runs on non-Linux.

### DIFF
--- a/tools/changed.sh
+++ b/tools/changed.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2017, Data61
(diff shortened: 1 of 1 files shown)


lsf37:
Merged in cf1428630e985ed8ba9999e4196b48b53c27e25b

PR #117, opened 2019-01-30
Fixed IPI VM bound notification bug

chrisguikema:
doRemoteOp1Arg data1 and cpu arguments were switched,
which triggered a conditional statement preventing
the VM from exiting and handling a bound notification.

This would only happen when a VM was running on its own core on x86.


include/arch/x86/arch/smp/ipi_inline.h
```diff
@@ -57,7 +57,7 @@ doRemoteClearCurrentVCPU(word_t cpu)
 static inline void
 doRemoteVMCheckBoundNotification(word_t cpu, tcb_t *tcb)
 {
-    doRemoteOp1Arg(IpiRemoteCall_VMCheckBoundNotification, cpu, (word_t)tcb);
+    doRemoteOp1Arg(IpiRemoteCall_VMCheckBoundNotification, (word_t)tcb, cpu);
```
pingerino: another great example of types in C

pingerino:
I've merged this internally and it will be pushed out once our tests are green. Thanks!!

PR #118, opened 2019-02-06
Keep bound notification in active state

chrisguikema:
Setting a bound notification to idle could
cause a second syscall to overwrite the notification
badge, preventing a VMM from routing a passthrough
interrupt properly.


kent-mcleod:
Hey @chrisguikema, @yyshen and I are about to look at this more closely, but are you able to provide more background about what the issue that this addresses is?

chrisguikema:
Hey Kent, our basic setup was a CAmkES-VM running on Core 1, with all the servers running on Core 0. A previous PR fixed the issue we were seeing with IPIs, allowing a VM to run on secondary cores. In order to test pass through devices on the VM, we tried passing through the USB host controller. The setup worked with the VM on Core 0, but not on Core 1.

The behavior we were seeing was basically: 
 - USB interrupt generated
 - USB interrupt passed to VM thread
 - Timer interrupt triggered from the Time Server
 - USB interrupt overwritten

We currently have this commit in our project, and we've been able to passthrough other devices (Ethernet, Sata, etc) to VMs running on secondary cores without issues.

axel-h:
what is the state of this?

lsf37:
Picking up this very old thread..

From a higher-level view the question here seem to be: should `VMCheckBoundNotification` properly receive the message on the notification (resetting the notification state as it does now) or should it just check and report the current value of the badge word? (as proposed here)

I can't say much on this from the verification side, because we don't have x86 virtualisation in the model, but if we did have a `check` operation in other configurations that did not actually receive, I don't think it would break anything in terms of invariants or internal consistency. It would be a new kind of operation, though.

So the question is really if this makes sense from an API perspective, which in turn mostly depends on what this is used for.

kent-mcleod:
> From a higher-level view the question here seem to be: should VMCheckBoundNotification properly receive the message on the notification (resetting the notification state as it does now) or should it just check and report the current value of the badge word? (as proposed here)
> So the question is really if this makes sense from an API perspective, which in turn mostly depends on what this is used for.

The current implementation is delivering a signal to a remote core that needs to call Arch_leaveVMAsyncTransfer() on that core instead of performing the operation remotely. Because the notification is delivered at this point, the Notification state needs to be set back to NtfnState_Idle as the badge value is copied into the thread's register and the thread is resumed.

I don't quite follow how the existing behavior is causing a passthrough interrupt to be dropped unless there was some other data race occurring.

The proposed change changes the behavior such that delivering a pending event badge from a notification bound to a VM thread that was in vm mode doesn't clear the pending event and so would be delivered multiple times which isn't the behavior we want.

Closing this PR, but feel free to open an issue if you still have an issue with this behavior.

PR #121, opened 2019-02-25
Arm GIC v3 support

kent-mcleod:
This is based off work contributed by @Jesse-Millwood and Dornerworks. The purpose of this PR is to agree on a version of the driver to merge.  The additional commits are some small stylistic fixups and some implementation policy changes based on concurrent work by @yyshen.

### DIFF
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -22,11 +22,6 @@
 /* Assume 8 cores */
 #define GICR_SIZE           (0x100000)
 
-/* Shift positions for GICD_SGIR register */
-#define GICD_SGIR_SGIINTID_SHIFT          0
-#define GICD_SGIR_CPUTARGETLIST_SHIFT     16
-#define GICD_SGIR_TARGETLISTFILTER_SHIFT  24
-
 #define GIC_DEADLINE_MS 2
 #define GIC_REG_WIDTH   32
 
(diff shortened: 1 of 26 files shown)


src/arch/arm/machine/gic_v3.c
```diff
@@ -276,6 +276,42 @@ cpu_iface_init(void)
...
+        icfgr |= (0b10 << bit);
+    } else {
+        icfgr &= ~(0b11 << bit);
+    }
+
+    if (is_ppi(irq)) {
+        gic_rdist_sgi_ppi_map[CURRENT_CPU_INDEX()]->icfgr1 = icfgr;
+    } else {
+        /* Update GICD_ICFGR<n>. Note that the interrupt should
+         * be disabled before chaning the field, and this function
```
Furao: "changing the field"

src/arch/arm/machine/gic_v3.c
```diff
@@ -12,7 +12,7 @@
  */
 #include <config.h>
 #include <types.h>
-#include <arch/machine/gic_500.h>
```
Furao: This is a good change, but shouldn't the gic_pl390 driver be changed as well to avoid confusion, since it also targets a spec instead of an implementation?

Furao:
My initial approve was an accident, and I can't seem to find a way to remove it, so please disregard it.

src/arch/arm/machine/gic_v3.c
kent-mcleod: I made the rename

kent-mcleod:
I'm not sure whether requesting the rereview worked.

Furao:
I didn't see a request to review again, but I was notified of both the comments you made.

PR #122, opened 2019-03-05
Initial i.MX8 support

kent-mcleod:
This adds support for the 64-bit i.MX8 Quad Max SOC. This port does
not have support for the messaging unit, which allows the cores to communicate
with the System Control Unit (SCU). Interaction with the SCU can be
done from user space or from the bootloader before branching to the
kernel. This means that clocks and peripherals must be powered up by
the bootloader or from userspace.
32-bit is not included here either because when booting with u-boot,
the Arm Trusted Firmware (ATF) needs to be 64-bit and is compiled into
u-boot so the developer may have to alter u-boot to load a 32-bit
image.

### DIFF
--- a/libsel4/sel4_plat_include/imx8qm/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/imx8qm/sel4/plat/api/constants.h
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2017, Data61
+ * Commonwealth Scientific and Industrial Research Organisation (CSIRO)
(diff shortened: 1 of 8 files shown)


axel-h:
https://github.com/seL4/seL4/issues/386 is blocked on this. This PR needs some rebase if this is to be merged to master. Otherwise just move this branch on top of the other other commit that that is no part of the mainline.

kent-mcleod:
Closing because of inactivity.

PR #125, opened 2019-04-01
seL4 kernel changes for zynqmp hypervisor extensions

chrisguikema:


chrisguikema:
Any update on this PR?

pingerino:
Hi @chrisguikema. @yyshen has it in his queue but is pretty busy at the moment. Is this blocking you? I can assign another engineer to work on it if so.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -26,7 +26,11 @@
 /* Allow native tasks to run at EL0, but restrict access */
 #define HCR_NATIVE ( HCR_COMMON | HCR_TGE | HCR_TVM | HCR_TTLB | HCR_DC \
                    | HCR_TAC | HCR_SWIO |  HCR_TSC | HCR_IMO | HCR_FMO | HCR_AMO)
+#ifdef CONFIG_PLAT_ZYNQMP
+#define HCR_VCPU   ( HCR_COMMON ) /* Allow zynqmp to use SMC calls */
```
pingerino: hi @chrisguikema I haven't put this commit as part of my updated 40-bit-pa support as I'm not sure it should go into master as is. Potentially we could hide it behind a config option? @kent-mcleod thoughts?

tools/dts/zynqmp.dts
```diff
@@ -843,34 +843,6 @@
 			};
 		};
 
-		pcie@fd0e0000 {
```
pingerino: we try to keep the device trees as close to original as possible and use overlays to modify them. As an example see https://github.com/seL4/seL4/blob/master/src/plat/imx6/overlay-sabre.dts.

include/arch/arm/arch/model/statedata.h
```diff
@@ -17,6 +17,10 @@
 #include <arch/machine/debug_conf.h>
 #include <mode/machine/registerset.h>
 
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
```
pingerino: since you raise this, this include has already been added (https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/model/statedata.h#L15)

pingerino:
Hi @chrisguikema, you've likely noticed I've changes from here into the new 40-bit PA PR which should be merged soon. I've commented on this commits I didn't incorporate with feedback. Thanks! This set of commits was very helpful for my work which also requires a 40-bit PA. I've tried to attribute commits appropriately but let me know if you have further comments.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
chrisguikema: Yeah, this should be hidden around a config option. Linux for the zynqmp uses a LOT of SMC calls to configure clocks and registers...obviously we don't want this behavior, but Linux can't run without it. I don't quite have the bandwidth for this, but I think the libsel4arm-vmm library will need to virtualize every SMC call. I think the config option is a good stop gap til that functionality exists.

tools/dts/zynqmp.dts
chrisguikema: So this may just be my system, but for whatever reason, the pcie device was generating devices with huge register addresses. So I just removed it for testing. 

Were you seeing that behavior at all?

include/arch/arm/arch/model/statedata.h
chrisguikema: ah cool!

chrisguikema:
@pingerino Yeah, I looked through your commits. They actually were pretty similar to the first set of commits that we open-sourced, but your version actually integrated with objects much better. As long as hypervisor extensions can be used on the cortex a53s, I'm happy. I felt as though you attributed commits properly.

chrisguikema:
I just added a couple commits for the last remaining issues on this PR. This is all thats required to run the zynqmp in hypervisor mode.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -26,7 +26,14 @@
 /* Allow native tasks to run at EL0, but restrict access */
 #define HCR_NATIVE ( HCR_COMMON | HCR_TGE | HCR_TVM | HCR_TTLB | HCR_DC \
                    | HCR_TAC | HCR_SWIO |  HCR_TSC | HCR_IMO | HCR_FMO | HCR_AMO)
+
+#ifdef CONFIG_ALLOW_SMC_CALLS
+/* This should only be a temporary fix. Guests should be able to call into EL3. These
+ * SMC calls should be virtualized in the VMM */
+#define HCR_VCPU   ( HCR_COMMON )
```
pingerino: I'm curious - what SMC calls is your guest using on the zynq? Does this commit help at all: https://github.com/SEL4PROJ/seL4_projects_libs/pull/8/commits/3d88add95ddb8f3432ace455bad57edff52518fd

chrisguikema: So I cherry-picked your commit, and while that does prevent any VMM issues from triggering, the guest does not work.

The zynqmp needs SMC calls for configuring clocks and other hardware resources. 

Here is the panic that I get from the guest:
```[    0.051622] Kernel panic - not syncing: zynqmp_plat_init power management API version error. Expected: v0.3 - Found: v65535.65535
[    0.051622]
[    0.064421] CPU: 0 PID: 1 Comm: swapper/0 Not tainted 4.9.0-xilinx-v2017.3 #2
[    0.071453] Hardware name: ZynqMP ZCU102 Rev1.0 (DT)
[    0.076344] Call trace:
[    0.078755] [<ffffff8008088138>] dump_backtrace+0x0/0x198
[    0.084069] [<ffffff80080882e4>] show_stack+0x14/0x20
[    0.089046] [<ffffff80083de514>] dump_stack+0x94/0xb8
[    0.094023] [<ffffff800812e9a8>] panic+0x114/0x25c
[    0.098741] [<ffffff8008c53aa8>] zynqmp_plat_init+0x140/0x168
[    0.104402] [<ffffff80080830b8>] do_one_initcall+0x38/0x128
[    0.109894] [<ffffff8008c30bd8>] kernel_init_freeable+0x84/0x1e0
[    0.115814] [<ffffff8008946548>] kernel_init+0x10/0x100
[    0.120961] [<ffffff8008082e80>] ret_from_fork+0x10/0x50
[    0.126198] ---[ end Kernel panic - not syncing: zynqmp_plat_init power management API version error. Expected: v0.3 - Found: v65535.65535
[    0.126198]```

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
pingerino: OK. Seems like this is a good solution for now.

As part of the work I'm doing at the moment I'll take a look at adding an `seL4_SMC_Control` cap to allow actual invocations.

axel-h:
What is the state if this PR by now, especially given https://sel4.discourse.group/t/rfc-9-add-new-capability-for-sel4-smc-forwarding/431

chrisguikema:
This PR can probably be closed. The RFC will allow specified threads to use SMC calls, which makes this PR pointless.

lsf37:
Good point. Closed in favour of the RFC.

PR #132, opened 2019-06-25
python2 --> python3

pingerino:
Update all scripts and build system to call python3, given python2's
upcoming doom. Use sys.maxsize instead of sys.maxint in one script
(maxint does not exist in python3).

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -1248,6 +1248,13 @@ def sign_extend_proof(high, base_bits, base_sign_extend):
         return ""
 
 
+def det_values(*dicts):
+    """Deterministically iterate over the values of each dict in `dicts`."""
+    def values(d):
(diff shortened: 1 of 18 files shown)


pingerino:
Looking into signing the CLA

pingerino:
CLA signed, this looks good to rebase cleanly and merge

tools/bitfield_gen.py
```diff
@@ -1248,6 +1248,13 @@ def sign_extend_proof(high, base_bits, base_sign_extend):
         return ""
 
 
+def det_values(*dicts):
```
pingerino: If this becomes a perf issue, consider using a `SortedDict` (we already have sorted containers as a python dep) so you don't need to sort every time. http://www.grantjenks.com/docs/sortedcontainers/sorteddict.html

PR #133, opened 2019-06-27
manual: Correct CNode slot size in bytes

canarysnort01:
It is 16 bytes on 32-bit systems and 32 bytes on 64-bit systems


kent-mcleod:
Thanks!

PR #134, opened 2019-07-01
arm: CONFIG_L1_CACHE_LINE_SIZE_BITS in cmake

pingerino:
Move definition of L1_CACHE_LINE_SIZE_BITS to cmake where it is set
according to the arm processor family.

This removes duplication in the hardware.h header files and makes adding
a new processor family require less lines changed.


include/arch/arm/arch/64/mode/kernel/thread.h
```diff
@@ -15,13 +15,25 @@ static inline word_t CONST sanitiseRegister(register_t reg, word_t v, bool_t arc
...
-                return v;
-            default:
-                break;
+            if (v & BIT(4)) {
+                switch (v & 0x3) {
+                case PMODE_USER:
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+                case PMODE_SUPERVISOR:
+#endif
+                    break;
```
yyshen: return v should be before the break?

PR #137, opened 2019-07-11
Minor changes, mostly for aarch64 hyp

pingerino:


configs/seL4Config.cmake
```diff
@@ -184,18 +184,25 @@ elseif(KernelArchArmV6)
 endif()
 if(KernelArmCortexA7)
     set(KernelArmCPU "cortex-a7" CACHE INTERNAL "")
+    config_set(KernelArmCacheLineSizeBits L1_CACHE_LINE_SIZE_BITS "6")
```
kent-mcleod: This needs to go in src/arch/arm/config.cmake

kent-mcleod: By needs I mean should.

pingerino: on it

include/arch/arm/arch/64/mode/kernel/thread.h
```diff
@@ -15,13 +15,25 @@ static inline word_t CONST sanitiseRegister(register_t reg, word_t v, bool_t arc
...
     if (reg == SPSR_EL1) {
         if (archInfo) {
-            switch (v & 0x1f) {
-            case PMODE_EL0t:
-            case PMODE_EL1t:
-            case PMODE_EL1h:
-                return v;
-            default:
-                break;
+            if (v & BIT(4)) {
```
kent-mcleod: This commit probably needs to be reordered until after we support 32-bit guests on aarch64

pingerino: removed this commit for now, I'll put it up later with other relevant changes

pingerino:
Rebased onto master and removed commits that are no longer relevant

PR #139, opened 2019-07-17
40 bit pa

pingerino:
This is an initial implementation of 40-bit PA support that is mature enough to merge into master. It is based on implementations by @chrisguikema  and @yyshen. I've tried my best to minimise invasive changes and make the API clean and less confusing.

I'm still in the process of testing this but I'd like initial feedback on the approach. One thing I'm not sure about is if all of the cap_page_global_directory_cap code should be #ifdef'd out? likely, as the compiler cannot tell it isn't being used. That could be done in a different PR though.


pingerino:
Added another commit which has some missing pieces I discovered in testing.

include/arch/arm/arch/64/mode/object/structures.h
```diff
@@ -54,7 +58,14 @@ typedef word_t vm_rights_t;
 
 #define VCPU_SIZE_BITS      seL4_VCPUBits
 
+#ifdef CONFIG_ARM_PA_SIZE_BITS_40
+/* For a 40 bit PA, EL1 and EL0 use a 3 level translation and skips the PGD */
```
chrisguikema: Quick question here. 

So generally, this is only true when the kernel is running at EL2, since VTCR_EL2 will be configured to start translations at Level 1. In most other cases, seL4 will run at EL1. These changes make any processor with the 40-bit PA range use a 3-level translation, regardless of kernel execution level. The ZUS+ could run seL4 before, so is there a reason for this setup?

pingerino: You're correct. I haven't experimented to make sure this builds in EL1. The intention of the configs is that `CONFIG_ARM_PA_SIZE_BITS_*` is unused when `CONFIG_ARM_HYPERVISOR_SUPPORT` is not enabled.

chrisguikema: Alright, so if thats the case, then my commit `fca033c` would need to be updated to only set that cmake variable when hypervisor support is enabled.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -106,10 +106,10 @@
                     Map an upper page directory
```
kent-mcleod: In 40 bit PA configurations, should this PUD object be not available in the API?

pingerino: Good point. Currently the kernel will `fail` if an invalid object is passed to `decodeARMMMUInvocation`.

pingerino: added a commit to resolve this

PR #140, opened 2019-07-22
boot: on arm, expose all pptrs as device memory

pingerino:
Some platforms have large amounts of devices which end up creating too
many untypeds to fit into boot info. Consequently we change the approach
on arm, and bring it in line with x86 to expose all non-kernel memory as
device untyped.

There's more to be done as a result of this PR: hardware_gen should stop generating the now-unused device list. You could also use the same approach on RISC-V, and it should be possible to make the x86 code more consistent with the arm code. However, I don't have time for this currently. My preference is to merge this now (assuming it doesn't break any platforms/projects), but am just putting it up to share/build-on.


kent-mcleod:
Thanks! This was incorporated into: https://github.com/seL4/seL4/commit/b5c56244f1394940fbb871616e6a80041fa1b038

PR #145, opened 2019-08-13
Gic v3 virtualisation

pingerino:


include/arch/arm/arch/machine/gic_v3.h
```diff
@@ -299,4 +333,209 @@ void ipi_send_target(irq_t irq, word_t cpuTargetList);
...
+}
+
+static inline virq_t gic_virq_pending_new(word_t group, word_t priority, word_t vid)
+{
+    return virq_virq_pending_new(0, group, priority, 0x200, vid);
+}
+
+static inline bool_t gic_valid_irq_idx(int irq_idx)
+{
+    return irq_idx < gic_vcpu_num_list_regs;
```
yyshen: given that irq_idx could be a result of | several bits, it could be larger than the gic_vcpu_num_list_regs?

src/arch/arm/object/vcpu.c
```diff
@@ -386,44 +385,19 @@ void VGICMaintenance(void)
...
-            irq_idx = ctzl(eisr1) + 32;
-        } else {
-            irq_idx = -1;
-        }
+        int irq_idx = get_vgic_irq_idx();
 
         /* the hardware should never give us an invalid index, but we don't
          * want to trust it that far */
-        if (irq_idx == -1  || irq_idx >= gic_vcpu_num_list_regs) {
+        if (irq_idx == GIC_INVALID_IRQ_IDX || !gic_valid_irq_idx(irq_idx)) {
```
yyshen: the gic_valid_irq_idx checkfor gicv3 could return incorrect result.

src/arch/arm/object/vcpu.c
```diff
@@ -447,6 +421,7 @@ void vcpu_init(vcpu_t *vcpu)
 #endif
     /* GICH VCPU interface control */
     vcpu->vgic.hcr = VGIC_HCR_EN;
+    vcpu->vgic.vmcr = 0xff000002; // TODO @yanyan wtf
```
yyshen: set default values for virtual EL1 PMR and enable group 1 virtual interrupts.

pingerino: Ok. Updated. Does this need to be only for the gicv3 or for both?

src/arch/arm/object/vcpu.c
```diff
@@ -566,6 +567,7 @@ exception_t invokeVCPUInjectIRQ(vcpu_t *vcpu, unsigned long index, virq_t virq)
 {
     if (likely(ARCH_NODE_STATE(armHSCurVCPU) == vcpu)) {
         set_gic_vcpu_ctrl_lr(index, virq);
+        vcpu->vgic.lr[index] = virq; // TODO @yanyan required?
```
pingerino: @yyshen

src/arch/arm/machine/gic_v3.c
```diff
@@ -130,10 +130,15 @@ static void gicv3_enable_sre(void)
     uint32_t val = 0;
 
     /* ICC_SRE_EL1 */
-    SYSTEM_READ_WORD(ICC_SRE_EL1, val);
-    val |= GICC_SRE_EL1_SRE;
+    SYSTEM_READ_WORD(ICC_SRE, val);
+    val |= GICC_SRE_SRE;
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        /* don't trap to el2 on access to ICC_SRE_EL1 */
+        /* TODO is this what we want permanently @yanyan?? */
```
pingerino: @yyshen

src/plat/imx8m-evk/overlay-imx8mq-evk.dts
```diff
@@ -26,4 +26,9 @@
 		status = "disabled";
 	};
 
+    guestvm@b0000000 {
```
kent-mcleod, line 29: I'm going to drop this change as a carve out isn't needed if this isn't going to be used in camkes-vm projects yet.

src/arch/arm/object/vcpu.c
```diff
@@ -227,6 +200,7 @@ void vcpu_init(vcpu_t *vcpu)
     /* Virtual Timer interface */
     vcpu->virtTimer.last_pcount = 0;
 #endif
+    vcpu->vgic.vmcr = VGIC_VMCR_VPMR | VGIC_VMCR_VENG1;
```
kent-mcleod, line 203: I'm also going to remove this initialization and leave these registers initialized to 0 as before.

lsf37:
The PR breaks the proofs for ARM_HYP in `VGICMaintenance()`, which looking at the diff seems to change quite a bit. I'll need to dig into the specs to understand what is going on there.

kent-mcleod:
> The PR breaks the proofs for ARM_HYP in `VGICMaintenance()`, which looking at the diff seems to change quite a bit. I'll need to dig into the specs to understand what is going on there.

Yea, the more I've looked through this PR while getting it ready to merge, the more I think we should just take the new one Benno has proposed.  This PR changes the semantics of VGICMaintenance() and the fault that's sent to user level quite a bit (it goes from placing a single interrupt in seL4_Fault_VGICMaintenance_new() to placing multiple interrupts that are bitwise-or'd together)

src/arch/arm/object/vcpu.c
```diff
@@ -430,7 +403,7 @@ exception_t decodeVCPUInjectIRQ(cap_t cap, unsigned int length, word_t *buffer)
         current_syscall_error.type = seL4_RangeError;
         return EXCEPTION_SYSCALL_ERROR;
     }
-    if (priority > 31) {
+    if (priority > GIC_MAX_PRIO) {
         current_syscall_error.type = seL4_RangeError;
         current_syscall_error.rangeErrorMin = 0;
         current_syscall_error.rangeErrorMax = 31;
```
lsf37, line 409: I guess `current_syscall_error.rangeErrorMax` should also become `GIC_MAX_PRIO` in that case.

src/arch/arm/machine/gic_v3.c
lsf37: @yyshen could you have another look at this part?

yyshen: I think this is needed to avoid trapping EL1 ICC_SRE access to EL2.

yyshen:
> The PR breaks the proofs for ARM_HYP in `VGICMaintenance()`, which looking at the diff seems to change quite a bit. I'll need to dig into the specs to understand what is going on there.
> 

A bit of history here. it was based on the observation that one maintenance interrupt can represent more than one interrupt EOIs on some hardware, and thus a loop is used to process all the EOIs. The GIC spec also says "• At least one List register entry has received an EOI request."
 
BTW, another issue is that a GICv3 maintenance interrupt can come when the VM is not running (at least on some hardware), and I think that breaks another assertion.

src/arch/arm/object/vcpu.c
yyshen: agreed.

yyshen:
https://github.com/seL4/seL4_projects_libs/tree/sylvain/gicv3 This is the corresponding VMM code for GICv3 emulation. I guess it is going to nowhere if no one takes it since the branch owner left? @lsf37 @kent-mcleod  That should work with new gicv3 PR as well, if the maintenance interrupt part is changed.

src/arch/arm/machine/gic_v3.c
kent-mcleod: If we don't trap access to this R/W register, then it would also need to be saved and restored during context switch which we don't do.  I'm more of a fan of what's in the other PR, this init code still uses ICC_SRE_EL1 to enable system registers, but doesn't make it accessible to EL1 without trapping.  This way EL1 can still access system registers and doesn't need to access ICC_SRE_EL1 which we don't make available to it.

kent-mcleod:
> https://github.com/seL4/seL4_projects_libs/tree/sylvain/gicv3 This is the corresponding VMM code for GICv3 emulation. I guess it is going to nowhere if no one takes it since the branch owner left? @lsf37 @kent-mcleod That should work with new gicv3 PR as well, if the maintenance interrupt part is changed.

Holding the kernel PR up until that PR is ready to merge is the main reason why the PRs have stayed open for the last 2 years ([context](https://github.com/seL4/seL4/pull/378#issuecomment-908220579)).  So we are planning to merge kernel support based on claims that it works with a VMM that isn't currently open sourced (https://github.com/seL4/seL4/pull/378#issuecomment-908153517)

kent-mcleod:
> A bit of history here. it was based on the observation that one maintenance interrupt can represent more than one interrupt EOIs on some hardware, and thus a loop is used to process all the EOIs. The GIC spec also says "• At least one List register entry has received an EOI request."

I believe that this can happen, but handling one IRQ per maintenance interrupt should still be acceptable as the maintenance IRQ is level triggered and will re-interrupt once it is reenabled after the first IRQ is processed.  If we want to optimize this and deliver multiple IRQs at once then we should make the change to GICv2 at the same time and update the verification as well.

kent-mcleod:
Closing this to show intent to merge https://github.com/seL4/seL4/pull/378.  But feel free to keep the discussion going here if necessary.

bennoleslie:
> > https://github.com/seL4/seL4_projects_libs/tree/sylvain/gicv3 This is the corresponding VMM code for GICv3 emulation. I guess it is going to nowhere if no one takes it since the branch owner left? @lsf37 @kent-mcleod That should work with new gicv3 PR as well, if the maintenance interrupt part is changed.
> 
> Holding the kernel PR up until that PR is ready to merge is the main reason why the PRs have stayed open for the last 2 years ([context](https://github.com/seL4/seL4/pull/378#issuecomment-908220579)). So we are planning to merge kernel support based on claims that it works with a VMM that isn't currently open sourced ([#378 (comment)](https://github.com/seL4/seL4/pull/378#issuecomment-908153517))

In somewhat chicken-egg problem, I want to release that VMM open source, but have held off until I have a workable platform on which to run it :)

yyshen:
> > A bit of history here. it was based on the observation that one maintenance interrupt can represent more than one interrupt EOIs on some hardware, and thus a loop is used to process all the EOIs. The GIC spec also says "• At least one List register entry has received an EOI request."
> 
> I believe that this can happen, but handling one IRQ per maintenance interrupt should still be acceptable as the maintenance IRQ is level triggered and will re-interrupt once it is reenabled after the first IRQ is processed. If we want to optimize this and deliver multiple IRQs at once then we should make the change to GICv2 at the same time and update the verification as well.

If I remember correctly, the maintenance interrupt is not disabled by the kernel. Thus, in theory, if you only process one EOI while there are several, once you return to user-mode, you will immediately get another maintenance interrupt before the first one is processed. Maybe the user VMM can still handle the case.

kent-mcleod:
> If I remember correctly, the maintenance interrupt is not disabled by the kernel. Thus, in theory, if you only process one EOI while there are several, once you return to user-mode, you will immediately get another maintenance interrupt before the first one is processed. Maybe the user VMM can still handle the case.

Yes you are right.  But because the VM thread has been made inactive, the VCPU will be disabled or a different one switched to which will mask the interrupts until the vgic is restored again.

yyshen:
> > If I remember correctly, the maintenance interrupt is not disabled by the kernel. Thus, in theory, if you only process one EOI while there are several, once you return to user-mode, you will immediately get another maintenance interrupt before the first one is processed. Maybe the user VMM can still handle the case.
> 
> Yes you are right.  But because the VM thread has been made inactive, the VCPU will be disabled or a different one switched to which will mask the interrupts until the vgic is restored again.

Disabling VCPU and thus VGIC does not mean an already pending interrupt will not fire.

kent-mcleod:
In which case it will trigger an interrupt in the kernel and be acknowledge after returning early here: https://github.com/seL4/seL4/blob/master/src/arch/arm/object/vcpu.c#L161

This is already needed to handle the case where the interrupt becomes pending just before a context switch and when it is handled there is already a different thread or without a vcpu. We rely on the interrupt being level triggered because we can acknowledge in the GIC and then when the VGIC LRs are reloaded it should trigger the interrupt again.  I don't believe that it would trigger an interrupt loop, but I don't have any hardware that has this behavior to test on.  

I do think that your idea of handling multiple IRQs in a single fault message is more efficient anyway.  I just think it should be changed in a future PR and also changed for GICv2 at the same time as it would be API breaking and require verification updates.

yyshen:
> In which case it will trigger an interrupt in the kernel and be acknowledge after returning early here: https://github.com/seL4/seL4/blob/master/src/arch/arm/object/vcpu.c#L161
> 
> This is already needed to handle the case where the interrupt becomes pending just before a context switch and when it is handled there is already a different thread or without a vcpu. We rely on the interrupt being level triggered because we can acknowledge in the GIC and then when the VGIC LRs are reloaded it should trigger the interrupt again. I don't believe that it would trigger an interrupt loop, but I don't have any hardware that has this behavior to test on.
> 
> I do think that your idea of handling multiple IRQs in a single fault message is more efficient anyway. I just think it should be changed in a future PR and also changed for GICv2 at the same time as it would be API breaking and require verification updates.

No problem. I just state some of the observations on some hardware since the assumption, maintenance interrupts only occur when the current thread has an associated VCPU, could be broken by the hardware.

PR #147, opened 2019-08-14
Multicore vmm

chrisguikema:
Just a couple commits that allow VMMs to run on secondary cores. These commits don't address the issue of getting the virtual timer irq to trigger (that needs to happen in userspace), but once the interrupt is turned on, then this allows the vms to function when running on secondary cores.

### DIFF
--- a/src/object/interrupt.c
+++ b/src/object/interrupt.c
@@ -41,7 +41,7 @@ exception_t decodeIRQControlInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
         irq_w = getSyscallArg(0, buffer);
-        irq = CORE_IRQ_TO_IDX(0, irq_w);
+        irq = CORE_IRQ_TO_IDX(getCurrentCPUIndex(), irq_w);
         index = getSyscallArg(1, buffer);
         depth = getSyscallArg(2, buffer);
 
--- a/include/arch/arm/arch/object/interrupt.h
+++ b/include/arch/arm/arch/object/interrupt.h
@@ -35,7 +35,7 @@ static inline void handleReservedIRQ(irq_t irq)
 #endif /* CONFIG_ARM_ENABLE_PMU_OVERFLOW_INTERRUPT */
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    if (irq == INTERRUPT_VGIC_MAINTENANCE) {
+    if (IDX_TO_IRQ(irq) == INTERRUPT_VGIC_MAINTENANCE) {
         VGICMaintenance();
         return;
     }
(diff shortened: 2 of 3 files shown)


pingerino:
Nice!

src/object/interrupt.c
```diff
@@ -41,7 +41,7 @@ exception_t decodeIRQControlInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
         irq_w = getSyscallArg(0, buffer);
-        irq = CORE_IRQ_TO_IDX(0, irq_w);
+        irq = CORE_IRQ_TO_IDX(getCurrentCPUIndex(), irq_w);
```
kent-mcleod: I don't agree with this change. The lack of a core argument should imply a core id of 0 instead of the core that the invocation is being called on.  You would also need to update the target core to the current CPU index in the case of SPIs.

kent-mcleod:
Are you able to remove Change-Id: * from your commit messages too please.  In the commit without a body message could you add more context (We style check the commit messages to try and keep them consistent against the policy laid out here: https://docs.sel4.systems/GitConventions#commit-messages).

src/object/interrupt.c
kent-mcleod: We discussed this internally when the initial changes were implemented. We also considered rejecting PPI irqs from being allocated using this invocation, but went with core id of 0 in all cases for backwards compatibility.

chrisguikema: I'm a bit confused here. For an SPI, `CORE_IRQ_TO_IDX` should always return the same value, regardless of core, which is not the case for a PPI, where the currently operating core is essential.

If PPIs are allowed to be passed down to userspace, then they should work for all cores. The way they are currently implemented, they only work for core 0, correct?

chrisguikema:
Yeah, I can remove the Change-ID. Thats for our internal code review system, but removing the tag shouldn't be difficult.

src/object/interrupt.c
Furao: @kent-mcleod So your suggested approach to support PPIs on other cores would require a change to the invocation to allow for the target core to be specified as an argument?

kent-mcleod: @chrisguikema, @Furao  by using this invocation you can only create handlers for PPIs on core 0.  If you want to create handlers for PPIs on other cores, then there is a newer invocation seL4_IRQControl_GetTriggerCore which allows a core to be specified.

kent-mcleod: Bump, I would like to merge the other changes from this PR but am happy to continue this discussion on a separate PR or under an issue?

Furao: thanks for the reminder. We will need to submit an issue because we could not use seL4_IRQControl_GetTriggerCore to fix the problem. @chrisguikema Can you update the PR to remove this change and then add that issue with the results we are seeing?

axel-h:
What is the state of this PR by now?

chrisguikema:
I'm fairly certain this is not needed anymore, since the virtual timer interrupt is now virtualized as an event

kent-mcleod:
> I'm fairly certain this is not needed anymore, since the virtual timer interrupt is now virtualized as an event

I think so too.

PR #149, opened 2019-08-23
armv8: expand generic timer user-level access

pingerino:
- allow the compare register to be set by user level (for both
  virtual and physical timer.
- allow the generic timer to be exported to EL1 as well when running
  seL4 as a hypervisor


nomadeel:
This was merged in f02e59c5.

PR #150, opened 2019-09-06
aarch64: allow access to memory below physBase

pingerino:
On aarch64 physBase is the constant that points to the bottom of
physical memory (RAM).

Prior to this change the kernel window was mapped directly to physBase,
which is usually not a 0 paddr. As a consequence the kernel could not
access any memory below physBase.

This change fixes this issue by mapping the start of the kernel window
to 0 in the physical address space.

- add new constant PADDR_LOAD, the location of the kernel image in the
physical address space.
- add new constant PADDR_BASE, the start of the physical address space
(0).
- add new constant KERNEL_ELF_BASE, the location of the kernel image in
kernel virtual memory.

A consequence of this change is that on aarch64, the kernelBase constant
now points to the start of the kernel window in virtual memory, but
*not* to the start of the kernel image as these are now different.


kent-mcleod:
Thanks for this.  Do you think that this change is significant enough to add to the CHANGES file?

pingerino:
Not really? CHANGES generally versions things that effect the user API.

kent-mcleod:
`As a consequence the kernel could not access any memory below physBase.` But I feel this does effectively affect the user API in a significant enough way?

src/arch/arm/common_arm.lds
```diff
@@ -18,13 +18,11 @@
 
 ENTRY(_start)
 
-KERNEL_BASE   = kernelBase;
-PHYS_BASE     = physBase;
-KERNEL_OFFSET = KERNEL_BASE - PHYS_BASE;
+KERNEL_OFFSET = KERNEL_ELF_BASE - PADDR_LOAD;
```
kent-mcleod: These definitions will also need to be created for aarch32?

PR #157, opened 2019-10-10
mcs: Donate scheduling context from bound ntfn

xurtis:
When a passive server receives a signal from a bound notification object
the SC from the bound notification object should be donated to the TCB
in order to handle the notification.

### DIFF
--- a/include/object/schedcontext.h
+++ b/include/object/schedcontext.h
@@ -74,6 +74,10 @@ void schedContext_bindNtfn(sched_context_t *sc, notification_t *ntfn);
 /* unbind scheduling context from a notification */
 void schedContext_unbindNtfn(sched_context_t *sc);
 
+/* Return a donated scheduling context to the notification to which it
+ * was bound if any is bound */
+void schedContext_maybeReturnToNtfn(sched_context_t *sc);
+
 time_t schedContext_updateConsumed(sched_context_t *sc);
 void schedContext_completeYieldTo(tcb_t *yielder);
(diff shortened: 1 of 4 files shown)


xurtis:
it looks like a bound notification object will only donate its SC if the notification object is not active when the notification is signalled and not when the TCB tries to complete an active signal when it performs a Recv

xurtis:
@ssrg-bamboo test

xurtis:
Affter looking at something @mbrcknl brought up regarding the way that SC donation is handled for notification objects I looked more into how it is handled for bound notification objects.

In your PhD you mentioned that when a thread blocks on an endpoint (although I think this should be any time it performs a Recv where it can accept a donated SC) it returns the SC from the bound notification. I couldn't find any code that implements this though; it seems the only time the SC from a notification object gets returned currently is on an explicit `seL4_Wait` on the notification from which the SC was donated.

I think we need to add some extra code in the `recv` paths for notifications and endpoints to return SCs donated from a notification.

xurtis:
Adding a test in sel4test should be fairly straightforward and I might get onto that today.

xurtis:
@pingerino can you take a look at the updated PR?

I've removed the code that only returns the donated SC when performing seL4_Wait on the notification object that donated the SC and replaced it with code that will return any SC donated from a notification object before initiating any potentially blocking receive operation.

xurtis:
A fix for this issue has already landed in master with #221

PR #158, opened 2019-10-23
output: use CR LF as line break and not LF CR

axel-h:


axel-h:
I've been wondering about this for some time, is there a reason why LF CR is used and all kernel logs contain an empty line between each line (on many systems). Or is this just a mistake and it should always have been CR LF actually?

kent-mcleod:
Nice find! I think it started out one way and then got copied across all the different serial drivers as they were added. I'm happy to get these merged. @ssrg-bamboo test

axel-h:
rebased to latest master

axel-h:
seems the bamboo bot does not act when you put the command into a comment

kent-mcleod:
I think it does, but the first time I made the commend I didn't format it properly and then edited it later. But I guess the bot already processed the message as read, @ssrg-bamboo `test`.

ssrg-bamboo:

PR #160, opened 2019-11-02
improve cmake scripts to support sub platforms

axel-h:
improve the cmake scripts to support sub platforms and avoid redundancy in all the platform scripts. By change, this also fixed a build breaker for imx7 then. Note that I have another patch on top of this that will create a generic imx8 platfom, but this requires a change in seL4/seL4_tools.git, so I will create another pull request then.


axel-h:
rebase to latest master, add changes to new ariane platform

axel-h:
Base on the information provided by @kent-mcleod  I'm dropping this pull request, as sub platforms will be handled differently.

PR #167, opened 2019-11-21
mcs: Prevent binding SCs to blocked TCBs

xurtis:
So as to ensure that all TCBs in queues are either faulted, passive, or
have schedulable SCs we cannot allow arbitrary SCs to be bound to TCBs
that are already in queues.

If a TCB is in a queue it should first be removed before having an SC
bound to it.

### DIFF
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -10,6 +10,8 @@
  * @TAG(DATA61_GPL)
  */
 #include <machine/timer.h>
+#include <kernel/sporadic.h>
+#include <object/structures.h>
 #include <object/schedcontext.h>
 
 static exception_t invokeSchedContext_UnbindObject(sched_context_t *sc, cap_t cap)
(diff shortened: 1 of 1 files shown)


xurtis:
@pingerino this is because there was an invariant identified where all TCBs in the blocked state are either bound to an SC that is ready and sufficient (such that they can be switched _to_ immediately when they become unblocked), not bound to an SC (so the can be switched to with SC donation), or they are blocked on send but not as part of a call.

In a related matter, there is an issue whereby `seL4_SchedControl_SetSchedParams` when called on a scheduling context bound to a TCB in the blocked state may lead to the SC becoming not-ready.

xurtis:
An alternative is not to assume that any thread transitioning from a blocked state to an unblocked state as the result of IPC is ready and sufficient.

pingerino:
@xurtis the alternative is not viable. We can't have blocked threads in queues wake up and not have budget, I trod this path, it led to recursion in the scheduler.

src/object/schedcontext.c
```diff
@@ -110,6 +112,18 @@ static exception_t decodeSchedContext_Bind(sched_context_t *sc, extra_caps_t ext
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+        switch (thread_state_get_tsType(TCB_PTR(cap_thread_cap_get_capTCBPtr(cap))->tcbState)) {
```
pingerino: This makes sense. If you are binding to a thread that is blocked to give it more time then it needs enough time to get out of the kernel.

src/object/schedcontext.c
```diff
@@ -110,6 +112,18 @@ static exception_t decodeSchedContext_Bind(sched_context_t *sc, extra_caps_t ext
...
+        case ThreadState_BlockedOnReceive:
+        case ThreadState_BlockedOnSend:
+        case ThreadState_BlockedOnNotification:
+        case ThreadState_BlockedOnReply:
```
pingerino: recommend a short handle for this (`isBlocked`)

xurtis: I'll need to come up with a different name, `isBlocked` already exists but is 'private' to thread.c and classifies slightly differently.

xurtis:
@pingerino are you sure the alternative isn't viable? I would expect it would be an added check in `schedule` that would revert to `chooseNewThread` if `ksCurrentSchedulerAction`/`candidate` is not ready or sufficient.

pingerino:
> @pingerino are you sure the alternative isn't viable? I would expect it would be an added check in `schedule` (perhaps an additional case on [`fastfail`](https://github.com/seL4/seL4/blob/master/src/kernel/thread.c#L386)) that would revert to `scheduleChooseNewThread` if `ksCurrentSchedulerAction`/`candidate` is not ready or sufficient.

Yes.

xurtis:
This will not be necessary with the proposed changes that allow unblock operations to result in a fault. This case will be more consistently handled by triggering a timeout or SC removed fault when a TCB becomes unblocked rather than by preventing the binding of the SC to the TCB.

Issue #168, opened 2019-12-01
Tutorial 'untyped' gave strange error


laokz:
Hi all,

I followed sel4-tutorial documents and tried 'untyped': 
```
../init --tut untyped --solution
```
The test succeeded. When I changed source line 25(the only I touched):
```
seL4_Word untyped_size_bits=13;
```
The last test failed:
```
     // allocate the whole child_untyped as endpoints
    seL4_Word num_eps = BIT(untyped_size_bits - seL4_EndpointBits);
	printf("untyped_size_bits:%d epbits:%d num_eps:%d\n", untyped_size_bits, seL4_EndpointBits, num_eps);
    error = seL4_Untyped_Retype(child_untyped, seL4_EndpointObject, 0, seL4_CapInitThreadCNode, 0, 0, child_tcb, num_eps);
    ZF_LOGF_IF(error != seL4_NoError, "Failed to create endpoints.");
```
```
<<seL4(CPU 0) [decodeUntypedInvocation/151 T0xffffff801fe09400 "rootserver" @4012b1>
main@main.c:76 [Cond failed: error != seL4_NoError]
	Failed to create endpoints.
```
In my box, the original untyped_size_bits was 12 and num_eps was 256. The last test was to try allocate 256 endpoint objects from 4K memory. I changed this to allocate 512 endpoint objects from 8K memory. The once whole allocation failed, but when I looped it 512 times with each only 1 object, it succeeded.

Is this a bug or what im wrong? My box is Debian bullseye x86_64, qemu v4.1.0, latest sel4. 

Regards.

maybe-sybr:
Amusingly I stumbled across this today as well.

<summary>Spoiler!<details>The solution expects you to use 12, which is double the size of a `TCBObject`, assuming that the other objects are not the same size. You can arrive at 13 legitimately if you don't make that assumption as instead take the largest bit size and multiply it by the number of objects to be retyped minus 1. That's what I chose to do, aiming to guarantee an exact or over-allocation in the child untyped object.</details></summary>

The behaviour is actually caused by the default `RETYPE_FAN_OUT_LIMIT` which is a hard maximum on the number of retyped objects you the kernel will allow you to construct in one call to `seL4_Untyped_Retype()`. That's set to 256 (2^8) and the final chunk of code in this example is attempting to retype the child into 512 (2^9 == 2^(13 - 4) where 13 is your child untyped object's bit size) endpoint objects. So it makes sense, but it's pretty annoying ):

[...]

PR #169, opened 2019-12-02
Use autoconf definition for `RetypeMaxObjects` def

maybe-sybr:
Relates to #168. The definition for `seL4_UntypedRetypeMaxObjects` lives
in the UAPI `types.h` but appears to have no link to the value actually
used by the kernel, which is configurable. This change sets the UAPI
definition to the generated definition from the kernel config steps and
defaults to the previous fixed value if, for some reason, the configured
definition is not available.

### DIFF
--- a/libsel4/include/sel4/types.h
+++ b/libsel4/include/sel4/types.h
@@ -31,7 +31,11 @@
 #include <sel4/shared_types.h>
 #include <sel4/mode/types.h>
 
-#define seL4_UntypedRetypeMaxObjects 256
+#ifdef CONFIG_RETYPE_FAN_OUT_LIMIT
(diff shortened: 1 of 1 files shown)


maybe-sybr:
I haven't signed the CLA before, may need to check with my employer but it should be fine.

kent-mcleod:
Thanks for this!  We should be able to merge it once a CLA is signed.

maybe-sybr:
I've sent the CLA back to the licensing address, let me know if it doesn't show up.

kent-mcleod:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

PR #172, opened 2019-12-13
mcs: Maintain IPC invariants

xurtis:
The following changes are made to ensure that invariants hold for IPC situations and to ensure that the receiver of an IPC does not need to trust a sender not to cause it to receive an IPC without an SC.

Senders of IPCs to passive servers are not given any control as to
whether SCs are donated so as to avoid the situation where a passive
server expects an SC but ends up inactive with no SC due to misplaced
trust in the sender.

The new semantics for scheduling context donation are simply that if the
sender has not timeout faulted, the receiver has no SC, and the sender
has an SC then the sender's SC will be donated to the receiver.

If the sender has an SC it is guaranteed to be ready on its core unless
it is a time-out fault (in which case the SC won't be donated to the
receiver).

This ensures a receiver will never time-out fault but that it may become
unschedulable if it ends up without an SC. To prevent this, trusted
software should not remove scheduling contexts from TCBs blocked on send
to endpoints serviced by passive servers.

Additionally, scheduling contexts must be configured and ready to execute on the core for which they are configured in order to be bound to a TCB or notification. This ensures that notification objects and threads blocked on endpoints are configured such that a receive operation will lead to the receiver having no SC or a ready SC but never a unready / timed-out SC.

The necessary updates to sel4test can be found in the sel4/sel4test#20 PR.

### DIFF
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -102,6 +102,12 @@ static exception_t decodeSchedContext_Bind(sched_context_t *sc, extra_caps_t ext
         return EXCEPTION_SYSCALL_ERROR;
     }
 
+    if (sc->scRefillMax == 0 || !refill_ready(sc) || !refill_sufficient(sc, 0)) {
+        userError("SchedContext_Bind: sched context requires re-configuration.");
+        current_syscall_error.type = seL4_IllegalOperation;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
     switch (cap_get_capType(cap)) {
     case cap_thread_cap:
         if (TCB_PTR(cap_thread_cap_get_capTCBPtr(cap))->tcbSchedContext != NULL) {
--- a/src/object/tcb.c
+++ b/src/object/tcb.c
@@ -1337,6 +1337,11 @@ exception_t decodeSetSchedParams(cap_t cap, word_t length, extra_caps_t excaps, 
             current_syscall_error.type = seL4_IllegalOperation;
             return EXCEPTION_SYSCALL_ERROR;
         }
+        if (sc->scRefillMax == 0 || !refill_ready(sc) || !refill_sufficient(sc, 0)) {
+            userError("TCB Configure: sched context requires re-configuration.");
+            current_syscall_error.type = seL4_IllegalOperation;
+            return EXCEPTION_SYSCALL_ERROR;
+        }
         break;
     case cap_null_cap:
         if (tcb == NODE_STATE(ksCurThread)) {
--- a/include/kernel/faulthandler.h
+++ b/include/kernel/faulthandler.h
@@ -21,7 +21,7 @@ static inline bool_t validTimeoutHandler(tcb_t *tptr)
 
 void handleTimeout(tcb_t *tptr);
 void handleNoFaultHandler(tcb_t *tptr);
-bool_t sendFaultIPC(tcb_t *tptr, cap_t handlerCap, bool_t can_donate);
+bool_t sendFaultIPC(tcb_t *tptr, cap_t handlerCap);
 #else
 exception_t sendFaultIPC(tcb_t *tptr);
 void handleDoubleFault(tcb_t *tptr, seL4_Fault_t ex1);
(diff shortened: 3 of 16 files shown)


src/object/schedcontext.c
```diff
@@ -102,6 +102,12 @@ static exception_t decodeSchedContext_Bind(sched_context_t *sc, extra_caps_t ext
         return EXCEPTION_SYSCALL_ERROR;
     }
 
+    if (sc->scRefillMax == 0 || !refill_ready(sc) || !refill_sufficient(sc, 0)) {
```
pingerino: I don't think this change should be merged. Servers cannot know the state of an SC when they invoke it for operations.

pingerino: I don't think this change should be merged. Servers cannot know the state of an SC when they invoke it for operations.

src/object/tcb.c
```diff
@@ -1337,6 +1337,11 @@ exception_t decodeSetSchedParams(cap_t cap, word_t length, extra_caps_t excaps,
             current_syscall_error.type = seL4_IllegalOperation;
             return EXCEPTION_SYSCALL_ERROR;
         }
+        if (sc->scRefillMax == 0 || !refill_ready(sc) || !refill_sufficient(sc, 0)) {
```
pingerino: as above

mbrcknl:
I think this works, with the following invariants:
- If a thread that is `BlockedOnSend` or `BlockedOnReply` has a scheduling context, then it must be active and either ready or timeout-faulted.
- If a thread that is `BlockedOnRecv` or `BlockedOnNtfn` has a scheduling context, then it must be active and ready.

None of these thread states require the thread to have a scheduling context. However, as you point out, if a passive server meets a sender that has no scheduling context, the server will go into limbo (runnable, but not queued anywhere). Our current invariants are happy enough with this limbo state, and you've given a reasonable explanation of why this makes sense from the system design perspective.

Here's my informal "proof":

This is a strictly weaker invariant than the one we've been trying to prove up to now, since we previously required the first case to also be "..., then it must be active and ready". So in all the places we've previously proved the stronger condition is established, we should get that the weaker one is established. There might be places where we used the stronger condition to establish something else, but I believe that is only during IPC, `cancelAllIPC` and `cancelBadgedSends`. The latter two only restart non-faulted threads, so they should be fine with the weaker condition.

In `sendIPC`, there are two main cases to consider: Firstly, if the sender blocks because there are no receivers waiting, then the invariant conditions must be satisfied by the thread performing the send. The two places we found that the old stronger conditions were not satisfied were `chargeBudget` and `doReplyTransfer`. In those cases, the sender of a timeout fault might have a scheduling context that is not ready. The new weaker invariant for threads `BlockedOnSend` allows for this. Secondly, if there is a receiver waiting, we need to ensure that if the receiver ends up with a scheduling context, then it must be ready for the call to `possibleSwitchTo`. If the receiver already has a scheduling context, then it must be ready according to the above invariant. If it received a donated context, then it can't have been from a timeout-faulted thread, since that case is explicitly excluded, and therefore must also be ready according to the invariant.

`receiveIPC` is now symmetric with `sendIPC`, so a similar argument applies.

In the case that a sender performs a call without donating a scheduling context, including any timeout fault, the sender ends up `BlockedOnRecv`. For this reason, the invariant condition for threads `BlockedOnRecv` must also be weakened. This is fine, because `doReplyTransfer` checks for readiness, and behaves accordingly.

src/object/schedcontext.c
mbrcknl: We need some way to ensure that threads that are blocked on IPC are ready if they have scheduling contexts (unless timeout faulted). Otherwise, a passive server might end up running on a non-ready SC. So I think we would either need this, or would need to cancel IPC on the thread being bound. From discussions with @xurtis, I understand that we at least want to be able to reconfigure scheduling contexts on threads that are `BlockedOnReply`, so cancelling might not be an option.

lsf37:
@xurtis if I remember correctly, the issue has been resolved in the meantime. Can the PR be closed?

xurtis:
We have a resolution (i.e. the concept of some sort of sched context removed fault that behaves like a timeout fault) that we intend to proceed with. That makes this PR wholly redundant in itself but is still yet to actually be implemented. A stop-gap implementation has been used in the meantime to get the proofs into a state where they pass for functional correctness.

PR #190, opened 2020-03-11
Add support for beaglebone

stevebob:
This adds the device tree spec for the beaglebone (original), and updates the build scripts to add the platform "am335x-bone".

The beaglebone is very similar to the beaglebone-black. All the devices used by the kernel are mapped to the same addresses with the same interrupt numbers. sel4test passes without updating any drivers in util_libs.

To use this, a corresponding change to seL4_tools is needed: https://github.com/seL4/seL4_tools/pull/37

I tested by running sel4test:
```
$ mkdir cbuild && cd cbuild
$ ../init-build.sh -DPLATFORM=am335x-bone -DAARCH32=1 -DCROSS_COMPILER_PREFIX=arm-linux-gnueabihf- -DFORCE_COLORED_OUTPUT=1
$ ninja
```
...and ran the result on my beaglebone. All the tests passed.

## CI
I'm pretty sure that an image produced by building for am335x-bone will run on a beaglebone-black, so it _shouldn't_ be hard to integrate this into CI. The inverse almost works for me. If I run sel4test built for am335x-boneblack on by beaglebone it crashes because it expects more memory than there is, but if I change src/plat/am335x/overlay-am335x-boneblack.dts such that only 256mb of memory is defined, it runs fine and passes all the tests.

### DIFF
--- a/src/plat/am335x/config.cmake
+++ b/src/plat/am335x/config.cmake
@@ -7,9 +7,9 @@
 cmake_minimum_required(VERSION 3.7.2)
 
 declare_platform(am335x KernelPlatformAM335X PLAT_AM335X KernelSel4ArchAarch32)
-set(c_configs PLAT_AM335X_BONEBLACK PLAT_AM335X_BONEBLUE)
-set(cmake_configs KernelPlatformAM335XBoneBlack KernelPlatformAM335XBoneBlue)
-set(plat_lists am335x-boneblack am335x-boneblue)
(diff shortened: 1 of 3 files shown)


kent-mcleod:
Hi @stevebob, thanks for contributing this.  We have a Beaglebone board already so should be able to add it to our CI. @branden-data61 will review your PRs. Are you also happy to update: https://docs.sel4.systems/Hardware/Beaglebone.html with information about this board?

stevebob:
@kent-mcleod happy to update the docs :)

axel-h:
What is the state here, any plan to merge this?

PR #191, opened 2020-03-16
Fix python error in code generation tool

stevebob:
`device.get_prop('compatible').strings` and `device.path` were both being passed as arguments to `join`, but `device.path` was supposed to be the second argument to `format`.


axel-h:
#196 got merged. Proposing to close this issue.

lsf37:
Indeed, #196 fixed this.

PR #194, opened 2020-05-02
riscv: add config option KernelRiscVBBL

axel-h:


axel-h:
see also https://github.com/seL4/seL4_tools/pull/39

yyshen:
Hi, @axel-h
Thanks for the PR. Could you explain why you added this config option and what the planned usage is. I do not see any other changes using this option.

axel-h:
Hi @yyshen,

currently seL4 on RISC-V architecture always creates an image with the ElfLoader packaged into a BBL image. However, there are RISC-V SOCs, where BBL is not used, because there are other bootloaders and SBI implementation. A SOC's customized build environment will not have a BBL repo at all then. Unfortunately, BBL usage is hard-code in the build system - so the build breaks unless the seL4 build system is patched - which mean forking. Or a dummy BBL repo is needed, which is not really a nice solution either.

Having this this config option solves this problem for 3rd party users, as BBL repo is not required then and the standard seL4 build system can still be used. At the moment in the official seL4 repos there is no RISC-V platform that uses this feature, both Ariane and HiFive ports currently seem to use the BBL. Not sure when we will be able to upstream our SOC's port and if we can publish the bootloader/SBI then. But I wonder, if this feature might also be useful for projects like OpenSBI, where you don't use BBL. 

Bottom line is, I thing hard-coding BBL usage just seems something to avoid.

yyshen:
> Hi @yyshen,
> 
> currently seL4 on RISC-V architecture always creates an image with the ElfLoader packaged into a BBL image. However, there are RISC-V SOCs, where BBL is not used, because there are other bootloaders and SBI implementation. A SOC's customized build environment will not have a BBL repo at all then. Unfortunately, BBL usage is hard-code in the build system - so the build breaks unless the seL4 build system is patched - which mean forking. Or a dummy BBL repo is needed, which is not really a nice solution either.
> 
> Having this this config option solves this problem for 3rd party users, as BBL repo is not required then and the standard seL4 build system can still be used. At the moment in the official seL4 repos there is no RISC-V platform that uses this feature, both Ariane and HiFive ports currently seem to use the BBL. Not sure when we will be able to upstream our SOC's port and if we can publish the bootloader/SBI then. But I wonder, if this feature might also be useful for projects like OpenSBI, where you don't use BBL.
> 
> Bottom line is, I thing hard-coding BBL usage just seems something to avoid.
> 
> EDIT: and this patch is meant to be integrated together with the patch [seL4/seL4_tools#39](https://github.com/seL4/seL4_tools/pull/39) then.

Thanks for the explanation. It looks like the seL4 kernel itself does not require this option, but the build system. Do you think the ELFLoader could be a better place for this option?

yyshen:
> > Hi @yyshen,
> > currently seL4 on RISC-V architecture always creates an image with the ElfLoader packaged into a BBL image. However, there are RISC-V SOCs, where BBL is not used, because there are other bootloaders and SBI implementation. A SOC's customized build environment will not have a BBL repo at all then. Unfortunately, BBL usage is hard-code in the build system - so the build breaks unless the seL4 build system is patched - which mean forking. Or a dummy BBL repo is needed, which is not really a nice solution either.
> > Having this this config option solves this problem for 3rd party users, as BBL repo is not required then and the standard seL4 build system can still be used. At the moment in the official seL4 repos there is no RISC-V platform that uses this feature, both Ariane and HiFive ports currently seem to use the BBL. Not sure when we will be able to upstream our SOC's port and if we can publish the bootloader/SBI then. But I wonder, if this feature might also be useful for projects like OpenSBI, where you don't use BBL.
> > Bottom line is, I thing hard-coding BBL usage just seems something to avoid.
> > EDIT: and this patch is meant to be integrated together with the patch [seL4/seL4_tools#39](https://github.com/seL4/seL4_tools/pull/39) then.
> 
> Thanks for the explanation. It looks like the seL4 kernel itself does not require this option, but the build system. Do you think the ELFLoader could be a better place for this option? Also, maybe RiscVFirmwareBBL is more meaningful. You could add RiscVFirmwareOpenSBI or RiscvFirmwareMyBlob for other machine-mode software.

axel-h:
@yyshen: good point, I'll remove this option then from the seL4 kernel project. However, using the ELFLoader also seem wrong, because it is also not really using this. it. Looking at https://github.com/seL4/seL4_tools/tree/master/cmake-tool these places might be an option
* helpers/risc-v.cmake (new file)
* helpers/application_settings.cmake (this has ELFLoader config stuff already)
* helpers/rootserver.cmake (keep things simple, this is also the place where we use the new config flag)

axel-h:
I'mm withdraw this patch, things are all in seL4/seL4_tools#39

axel-h:
@yyshen: concerning your suggestion about `RiscVFirmwareOpenSBI` or `RiscvFirmwareMyBlob`, I would postpone this until somebody is requesting this because there is another firmware. For now BBL seems the only thing available, so turning it on and off seem sufficient for all use cases
Our solution at the moment is using the seL4 build system as a sub-step in our build system, so the next stept can pick up the created images (like the ELFLoder) and process them further. I'm not not too keen to add more smartness (ie. complexity) to the seL4 build system.

yyshen:
> @yyshen: concerning your suggestion about `RiscVFirmwareOpenSBI` or `RiscvFirmwareMyBlob`, I would postpone this until somebody is requesting this because there is another firmware. For now BBL seems the only thing available, so turning it on and off seem sufficient for all use cases

Sorry I did not express it better. I meant these options RiscVFirmwareOpenSBI or RiscvFirmwareMyBlob could be added later if necessary.

> Our solution at the moment is using the seL4 build system as a sub-step in our build system, so the next stept can pick up the created images (like the ELFLoder) and process them further. I'm not not too keen to add more smartness (ie. complexity) to the seL4 build system.

yyshen:
> I'm withdrawing this patch, things are all in [seL4/seL4_tools#39](https://github.com/seL4/seL4_tools/pull/39)

OK. I will look at that PR. Thanks.

PR #198, opened 2020-05-07
Repair barriers in clh_lock_acquire

JonasOberhauser:
Strengthen the clh_lock_acquire to use release on the atomic_exchange that makes the node public. Otherwise (on ARM & RISCV), the store to the node value which sets its state to CLHState_Pending can become visible some time after the node is visible.
In that window of time, the next thread which attempts to acquire the lock will still see the old state (CLHState_Granted) and enters the critical section, leading to a mutual exclusion violation.
See also: https://gist.github.com/JonasOberhauser/6198862f0c47f68040369787f131248f which is a test written by Diogo Behrens that can reproduce the mutual exclusion violation on some ARM servers (Taishan 2280)

Signed-off-by: jonas <jonas.oberhauser@huawei.com>


JonasOberhauser:
Looks like I accidentally merged master into my repository, messing up this pull request. I'll create a new one.

PR #205, opened 2020-06-16
Repair barriers in try_arch_atomic_exchange

JonasOberhauser:
try_arch_atomic_exchange currently never acts as a release barrier.
The reason for this is that a release fence needs to be inserted before
the store it attempts to order with previous stores.
In the current implementation, a release barrier is only inserted after
the store, thus the exchange operation is never ordered w.r.t. previous
operations.
With the suggested fix, passing RELEASE or stronger as success memory order
will correctly order the exchange operation w.r.t. previous operations.

### DIFF
--- a/include/arch/arm/arch/model/smp.h
+++ b/include/arch/arm/arch/model/smp.h
@@ -22,6 +22,16 @@ static inline bool_t try_arch_atomic_exchange(void *ptr, void *new_val, void **p
     uint32_t atomic_status;
     void *temp;
 
+    /* On failure there is no store.
+        Thus we can ignore release in the failure memory order,
+        and place a release barrier only if the success memory order implies
+        such an ordering. */
+    if (success_memorder == __ATOMIC_RELEASE || success_memorder == __ATOMIC_ACQ_REL) {
+        __atomic_thread_fence(__ATOMIC_RELEASE);
+    } else if (success_memorder == __ATOMIC_SEQ_CST) {
+        __atomic_thread_fence(__ATOMIC_SEQ_CST);
+    }
+
     asm volatile(
         LD_EX "%[prev_output], [%[ptr_val]]             \n\t" /* ret = *ptr */
         ST_EX "%" OP_WIDTH "[atomic_var], %[new_val] , [%[ptr_val]] \n\t"  /* *ptr = new */
@@ -34,7 +44,11 @@ static inline bool_t try_arch_atomic_exchange(void *ptr, void *new_val, void **p
 
     /* Atomic operation success */
     if (likely(!atomic_status)) {
-        __atomic_thread_fence(success_memorder);
+        if (success_memorder == __ATOMIC_ACQUIRE || success_memorder == __ATOMIC_ACQ_REL) {
+            __atomic_thread_fence(__ATOMIC_ACQUIRE);
+        } else if (success_memorder == __ATOMIC_SEQ_CST) {
+            __atomic_thread_fence(__ATOMIC_SEQ_CST);
+        }
     } else {
         /* Atomic operation failure */
         __atomic_thread_fence(failure_memorder);


kent-mcleod:
Hi @JonasOberhauser , I asked for @yyshen to review this and the other variant.  Do you think it makes sense to attach the commit from https://github.com/seL4/seL4/pull/199 to these pull requests and treat them as a single PR or would you prefer to merge the earlier PR first and then consider these?  My interpretation is that one of these 2 new PRs are required to make the original PR change effective.  Is this correct?

JonasOberhauser:
This is partially correct. On ARM, #199 is not effective unless also #205 or #206 (or some other alternative) are implemented. However the bug fixed by #199 also exists on RISCV, on which `try_arch_atomic_exchange` correctly implements the barriers. There #205/#206 are not needed to make the change effective.

I definitely would prefer to add them separately as I consider them separate (although related) issues. But as always I'm open to suggestions.

yyshen:
Hi @JonasOberhauser, thanks for the PR. My understanding about weak memory model is limited, let me know if my understanding is incorrect.

> This is partially correct. On ARM, #199 is not effective unless also #205 or #206 (or some other alternative) are implemented. However the bug fixed by #199 also exists on RISCV, on which `try_arch_atomic_exchange` correctly implements the barriers. There #205/#206 are not needed to make the change effective.
RISCV with #199 is enough is because the atomic_exchange with ACQ_REL is translated to "fence iorw, ow" and "amoswap.d.aq". So, the previous stores are ordered before the amoswap.d.aq by the fence, and amoswap with acquire ensures the order of the following memory accesses. On Arm, as stated, the previous stores, especially "    big_kernel_lock.node_owners[cpu].node->value = CLHState_Pending;" is not ordered with the "strex" or "stxr". This is why the #205 or #206 is required.
> 
> I definitely would prefer to add them separately as I consider them separate (although related) issues. But as always I'm open to suggestions.

I think it is a good idea to pull the barriers out of the loop (#206), and the PR looks good to me. Given that the kinds of barriers we need before and after try_arch_atomic_exchange_rlx are clear, I am not sure we still need to pass the "memorder" around and use the if statements to select the barriers. Am I missing something?

Thanks!

JonasOberhauser:
> Hi @JonasOberhauser, thanks for the PR. My understanding about weak memory model is limited, let me know if my understanding is incorrect.
> 
> > This is partially correct. On ARM, #199 is not effective unless also #205 or #206 (or some other alternative) are implemented. However the bug fixed by #199 also exists on RISCV, on which `try_arch_atomic_exchange` correctly implements the barriers. There #205/#206 are not needed to make the change effective.
> 
> RISCV with #199 is enough is because the atomic_exchange with ACQ_REL is translated to "fence iorw, ow" and "amoswap.d.aq". So, the previous stores are ordered before the amoswap.d.aq by the fence, and amoswap with acquire ensures the order of the following memory accesses. On Arm, as stated, the previous stores, especially " big_kernel_lock.node_owners[cpu].node->value = CLHState_Pending;" is not ordered with the "strex" or "stxr". This is why the #205 or #206 is required.

In principle yes. A minor nitpick is that I hope the translation for the `atomic_exchange` would be different  : ) `amoswap.pc.aq.rel` or something like that. In our experience, fences are horribly inefficient, and it is better to use access-level barriers to generate the necessary ordering.

> 
> > I definitely would prefer to add them separately as I consider them separate (although related) issues. But as always I'm open to suggestions.
> 
> I think it is a good idea to pull the barriers out of the loop (#206), and the PR looks good to me. Given that the kinds of barriers we need before and after try_arch_atomic_exchange_rlx are clear, I am not sure we still need to pass the "memorder" around and use the if statements to select the barriers. Am I missing something?

You suggest to change the `sel4_atomic_exchange` by removing the memorder argument? I am ok with this change, but I'm not sure what the future plans for seL4 are. If you later want to have more fine-grained concurrency, maybe keeping a generic `sel4_atomic_exchange` around is a good idea.
My hope would be that with sufficient optimization levels, the compiler will inline the ifs and generate the same code?

yyshen:
> > Hi @JonasOberhauser, thanks for the PR. My understanding about weak memory model is limited, let me know if my understanding is incorrect.
> > > This is partially correct. On ARM, #199 is not effective unless also #205 or #206 (or some other alternative) are implemented. However the bug fixed by #199 also exists on RISCV, on which `try_arch_atomic_exchange` correctly implements the barriers. There #205/#206 are not needed to make the change effective.
> > 
> > 
> > RISCV with #199 is enough is because the atomic_exchange with ACQ_REL is translated to "fence iorw, ow" and "amoswap.d.aq". So, the previous stores are ordered before the amoswap.d.aq by the fence, and amoswap with acquire ensures the order of the following memory accesses. On Arm, as stated, the previous stores, especially " big_kernel_lock.node_owners[cpu].node->value = CLHState_Pending;" is not ordered with the "strex" or "stxr". This is why the #205 or #206 is required.
> 
> In principle yes. A minor nitpick is that I hope the translation for the `atomic_exchange` would be different : ) `amoswap.pc.aq.rel` or something like that. In our experience, fences are horribly inefficient, and it is better to use access-level barriers to generate the necessary ordering.

I tried with riscv64 gcc 9.2.1 and gcc8.3, it looks like the __atomic_exchange_n is implemented with "fence iorw,ow" and "amoswap.d.aq". The "fence iorw,ow" is used to order both IO memory accesses and physical memory accesses. Did I miss any compiler options for optimisation?

> 
> > > I definitely would prefer to add them separately as I consider them separate (although related) issues. But as always I'm open to suggestions.
> > 
> > 
> > I think it is a good idea to pull the barriers out of the loop (#206), and the PR looks good to me. Given that the kinds of barriers we need before and after try_arch_atomic_exchange_rlx are clear, I am not sure we still need to pass the "memorder" around and use the if statements to select the barriers. Am I missing something?
> 
> You suggest to change the `sel4_atomic_exchange` by removing the memorder argument? I am ok with this change, but I'm not sure what the future plans for seL4 are. If you later want to have more fine-grained concurrency, maybe keeping a generic `sel4_atomic_exchange` around is a good idea.
> My hope would be that with sufficient optimization levels, the compiler will inline the ifs and generate the same code?

Yes, the compiler should do the job. Also, removing the memorder argument can be done in another commit if needed.

kent-mcleod:
Closed due to merging of #206.

PR #215, opened 2020-07-02
Clear sctlr_el1.uci bit explicitly

yyshen:
Clear the sctlr_el1.uci bit to trap EL0 cache maintenance instruction to EL2. Previously, the we rely on the reset value, which is probably 0.


yyshen:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

yyshen:
@kent-mcleod any comments?

kent-mcleod:
This is inconsistent with `SCTLR_EL1_NATIVE` in vcpu.h which sets `sctlr_el1.uci` for native seL4 threads if hyp mode is enabled.  Do you remember why it is set in this case?

yyshen:
> This is inconsistent with `SCTLR_EL1_NATIVE` in vcpu.h which sets `sctlr_el1.uci` for native seL4 threads if hyp mode is enabled. Do you remember why it is set in this case?

Well, if my memory serves right, it is the leftover when I used cache clean instructions directly in the VMM when loading Linux images and doing some testing. You could remove that and run a test, it should not break existing VMM code.

axel-h:
what is the state of this PR?

Indanz:
We would like to use DMA on cached memory for performance reasons, and doing cache maintenance from EL0 would help here.

Perhaps add an (advanced) kernel build option to allow or disallow userspace cache maintenance instructions?

kent-mcleod:
> Perhaps add an (advanced) kernel build option to allow or disallow userspace cache maintenance instructions?

I think that this is appropriate. This config option would control the setting for native EL0 threads when the kernel is in EL1 or EL2.  VCPU threads would still have the default settings defined by `HCR_VCPU` which presently do not trap any cache operations. I think this is pragmatic while the kernel still doesn't have a model for managing cache maintenance authorities.

kent-mcleod:
I've put a newer version of this up: https://github.com/seL4/seL4/pull/1026

PR #216, opened 2020-07-16
New SC faulting semantics for MCS

xurtis:
Timeout faults shouldn't assume that the receiver of the fault has access to the SC so they no longer pass the consumed budget. This allows the SC to be passed back to a calling thread and fault in a manner that a handler that does have access to the SC can determine the amount of time consumed.

We check in all cases where a thread is resumed that may not have an active and ready SC that it does. If it does not we fault that thread.

A new fault is introduced for the case where a thread that was faulted as described above had no SC or an SC that was inactive. This fault can only be delivered to an active thread. If the timeout endpoint has a thread that is not active this fault will not be delivered and the thread will be left inactive.


xurtis:
I will add a commit that updates the manual to describe the details in full and the semantics of the changed faulting behaviour. I'll render out the PDF and add here once that is done.

include/kernel/faulthandler.h
```diff
@@ -7,13 +7,33 @@
...
+#include <kernel/sporadic.h>
+#endif
 
 #ifdef CONFIG_KERNEL_MCS
 static inline bool_t validTimeoutHandler(tcb_t *tptr)
 {
     return cap_get_capType(TCB_PTR_CTE_PTR(tptr, tcbTimeoutHandler)->cap) == cap_endpoint_cap;
 }
 
+static inline bool_t validInvalidSchedContextHandler(tcb_t *tptr)
```
pingerino: I don't understand the name of this function

src/kernel/thread.c
```diff
@@ -619,16 +619,16 @@ bool_t ensureSchedulable(tcb_t *thread, bool_t can_timeout)
...
+            current_fault = seL4_Fault_InvalidSchedContext_new(seL4_SchedContextMissing);
+            handleTimeout(thread);
         }
         return false;
     } else if (unlikely(!sc_active(sc))) {
         /* Not configured */
-        if (false) {
-            // TODO: Add not bound fault
-            // SCInvalid can only be sent to active threads
+        if (validInvalidSchedContextHandler(thread)) {
```
pingerino: where is ensureSchedulable called? I don't recall this function. When can you fault for not having an SC?

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
```diff
@@ -155,6 +155,34 @@ enum {
     seL4_TimeoutReply_Length,
     SEL4_FORCE_LONG_ENUM(seL4_TimeoutReply_Msg)
 } seL4_TimeoutReply_Msg;
+
```
pingerino, line 158: do you consider encoding this into the existing temporal fault rather than introducing a new fault type?

include/kernel/thread.h
```diff
@@ -291,5 +295,14 @@ void awaken(void);
 /* Place the thread bound to this scheduling context in the release queue
  * of periodic threads waiting for budget recharge */
 void postpone(sched_context_t *sc);
+/* Ensure that a thread can be placed in the scheduler with
+ * possibleSwitchTo, SCHED_ENQUEUE, or SCHED_APPEND. Faults the thread
+ * with a timeout fault or SC unbound fault if its SC would prevent it
+ * from being scheduled.
+ *
+ * If this returns true then the thread that was passed in is not
+ * modified and satisfies the precodnitions of possibleSwitchTo.
```
pingerino, line 304: nit: typo

xurtis:
I was trying to remove the assumption that the server that handles the fault would otherwise be able to modify the SC (rather than just the TCB). Everything else that can be done by the fault handler can be done via invocations on the TCB.

Having the consumed time reset when a fault is received means that the first thread that faults on an SC has its handler receive the time consumed by the SC but if that fault gets passed back along a call chain every other fault handler will receive 0.

I'm not entirely sure of the implications of either choice, a potential intermediary solution is to deliver the consumed time but not reset it to 0 when delivering the fault. It may even make sense to not reset it when accessed via the syscall.

xurtis:
Thanks for looking at this so soon. I'll look over the rest of your comments tomorrow morning

src/api/syscall.c
```diff
@@ -570,6 +570,13 @@ static void handleRecv(bool_t isBlocking)
         handleFault(NODE_STATE(ksCurThread));
         break;
     }
+
+#ifdef CONFIG_KERNEL_MCS
+    if (isRunnable(NODE_STATE(ksCurThread)) && !ensureSchedulable(NODE_STATE(ksCurThread), true)) {
```
pingerino, line 575: this is contrary to the initial design of the kernel where we *always* call rescheduleRequired when we do something that could require a reschedule, rather than checking at the end of every single system call. 

Also, I'm not sure you have enough state here to make this decision - how do you know at this point that ksCurThread is actually going to be scheduled? we could have already switched it to blocked? In which case, are you triggering a reschedule unnecessarily?

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
xurtis: Given they are represented much the same and are handled basically identically, that is an entirely good option. Would mean extending the existing temporal fault with the `reason`, but I think that makes a lot of sense.

src/kernel/thread.c
```diff
@@ -171,17 +171,11 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
...
-        } else {
-            if (validTimeoutHandler(receiver) && fault_type != seL4_Fault_Timeout) {
-                current_fault = seL4_Fault_Timeout_new(receiver->tcbSchedContext->scBadge);
-                handleTimeout(receiver);
-            } else {
-                postpone(receiver->tcbSchedContext);
-            }
-        }
+    if (isRunnable(receiver) && ensureSchedulable(receiver, fault_type != seL4_Fault_Timeout)) {
+        /* Presently we don't allow a thread to fault when replied to
```
pingerino, line 175: That's new. I had intended that faults should be able to propogate across replys so we can handle timeout faults in nested servers. What is the reasoning behind removing this?

src/kernel/thread.c
xurtis: `ensureSchedulable` is called basically before any call to `SCHED_ENQUEUE` (including via `possibleSwitchTo`) or `SCHED_APPEND` for a thread that is not the current thread or when the current thread may have changed or lost it's SC.

xurtis: Only exception is in `sendFaultIPC` which, to avoid recursion, only checks `isSchedulable` and will not fault an unblocked receiver for temporal reasons (although `sendFaultIPC` should not be called if there is a blocked receiver that would fault).

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
```diff
@@ -130,9 +130,6 @@ enum {
 #ifdef CONFIG_KERNEL_MCS
 enum {
     seL4_Timeout_Data,
-    /* consumed is 64 bits */
```
pingerino, line 133: "
The new model of time-out faults assumes that a time-out handler is
associated with thread (or at least the thread's time-out endpoint) and
does not assume that the time-out handler has any control over an SC.
"

Timeout faults were primarily designed for either user level scheduling or to handle client SC's running out of time in a shared server. In the latter case, in order to implement different escape strategies for the server, I needed to be able to  identify which SC had faulted. In the former case, I needed the consumed value to do efficient user level scheduling. 

Why is this being removed? How will you handle the rollback scenarios identified in my thesis? How much longer will user level scheduling take now that you have to make an extra invocation to get the consumed amount?

src/kernel/thread.c
pingerino: I did this initially and decided it was far too heavy weight, and better to work on a case-by-case basis. There are plenty of places where SCHED_ENQUEUE is called and we *know* we have a scheduling context.

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
pingerino: I highly recommend you do this. It will reduce the added code and total code size.

src/kernel/thread.c
xurtis: This doesn't prevent faults being propogated across replies, it prevents a thread that had time-out faulted to immediately time-out fault when the fault handler replies to the fault. The code that propogates time-out faults over replies is in `doReplyTransfer` I think.

xurtis: This doesn't remove any cases that would fault from the code it replaces.

xurtis: This comment could also be made... legible at all.

src/api/syscall.c
xurtis: This may have been an oversimplification of the common case at the end of `receiveIPC` and `receiveSignal`.

include/kernel/faulthandler.h
mbrcknl: It's like `validTimeoutHandler` for the new fault type. In any case, it goes away in the next commit.

src/kernel/thread.c
xurtis: I think all of those places made the assumption that threads blocked on IPC always had a ready and sufficient scheduling context which prevents configuring scheduling contexts without knowing what the IPC state of a thread is. I was under the impression we wanted to keep IPC state orthogonal to SC configuration.

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
xurtis: What was your reason for re-setting the consumed time in the SC when it is retrieved by syscall or fault?

The case that I am trying to preserve is having a passive server reply to a client such that it faults again and having the timeout fault handler for the client not observe a 0 consumed time in the fault.

It looks like you are trying to handle the case where the timer resolution is more coarse than 1us and so conversion to us may lead to overflow. Given that system time will eventually overflow both in ticks and in us, we could simply only guarantee that the system is guaranteed to behave correctly up to the lower of either bound.

Would it be reasonable to not reset the consumed time when providing it to user level and simply accept any overflow that might occur in the process?

xurtis:
I've updated the changes in response to some of the above discussion. I still need to do the documentation portion.

src/api/syscall.c
xurtis: On review, I think this is fine. If it is blocked, the `isRunnable` will return false. If it is runnable and (i.e. `ThreadState_Running`, `ThreadState_Restate`, or `ThreadState_RunningVM`) then it hasn't been blocked but it may have received a new scheduling context at this point or may have no scheduling context after donating the `ksCurSC` in the first phase (if there was one).

This is also only for `habdleRecv` rather than every syscall, so this is only on receive from an endpoint or notification.

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
pingerino: > What was your reason for re-setting the consumed time in the SC when it is retrieved by syscall or fault?

If you are using this info you are most likely trying to monitor the execution of the thread from user level. I couldn't see any other more efficient way to figure out what you had already read and get accurate accumluations over time. Look at the implementation of a user-level CFS scheduler. 

> The case that I am trying to preserve is having a passive server reply to a client such that it faults again and having the timeout fault handler for the client not observe a 0 consumed time in the fault.

I assumed the fault handlers were per server and cared about the time spent in the server. 

> It looks like you are trying to handle the case where the timer resolution is more coarse than 1us and so conversion to us may lead to overflow.

No. Although it does help. 

> Would it be reasonable to not reset the consumed time when providing it to user level and simply accept any overflow that might occur in the process?

No.

src/kernel/thread.c
pingerino: ok so - if the timeout fault handlers don't correctly handle the fault what happens now? we should fault again, IMO

src/api/syscall.c
pingerino: if its possible to remove branches then you should

pingerino:
> I was trying to remove the assumption that the server that handles the fault would otherwise be able to modify the SC (rather than just the TCB). Everything else that can be done by the fault handler can be done via invocations on the TCB.
> 

Invocations are slow. If possible the fault handler should be able to get all the info it needs from the fault message, and process all actions on the reply.

src/kernel/thread.c
pingerino: We have to keep them separate, since IPC state is mutable.

pingerino:
In future please try and keep PRs to a single feature change so they are digestible by time-poor reviewers.

> Timeout faults shouldn't assume that the receiver of the fault has access to the SC so they no longer pass the consumed
 > budget. This allows the SC to be passed back to a calling thread and fault in a manner that a handler that does have access  > the SC can determine the amount of time consumed.

In it's current state this change should not be made. Please give some examples and motivations for this change and we can see if there's another way to achieve what you want.

> We check in all cases where a thread is resumed that may not have an active and ready SC that it does. If it does not we fault > that thread.

I think this is OK, as long as we can still initialise a passive server and the timeout fault handling scenarios I implemented in sel4test and sel4bench still work.

> A new fault is introduced for the case where a thread that was faulted as described above had no SC or an SC that was  inactive. This fault can only be delivered to an active thread. If the timeout endpoint has a thread that is not active this fault will not be delivered and the thread will be left inactive.

As discussed, this shouldn't be a *new* fault but a variant of the existing timeout fault. If you can keep the code smaller, keep it smaller :)

xurtis:
I haven't had any time to write them and at this rate they may not get written until Thursday.

gernotheiser:
I’ll be off Thu-Mon

src/kernel/thread.c
xurtis: There are a handful of cases where the check isn't added where we were already doing all necessary checking (such as `yieldTo`) but I think they are in the minority.

I think to be more specific, this check is added when we try to reschedule a thread but have no prior knowledge of the SC with which it would be scheduled. This excludes, for example, the fastpath as that will always use the SC that was current at kernel entry.

src/kernel/thread.c
xurtis: Right, I wasn't sure what to do here. I'm also in favour of faulting again. Another suggestion was to use to allow the response to the fault to differentiate between resume with recurring fault or to resume and let the thread be postponed.

xurtis:
@pingerino I'll probably drop the changes to consumed, we can look at them again if it does turn out to be an issue.

src/object/schedcontext.c
```diff
@@ -353,10 +353,8 @@ time_t schedContext_updateConsumed(sched_context_t *sc)
 {
     ticks_t consumed = sc->scConsumed;
     if (consumed >= getMaxTicksToUs()) {
-        sc->scConsumed -= getMaxTicksToUs();
```
pingerino, line 356: Although this commit message states what the change does, and the implications of the change, It's not clear from the commit message *why* this change is being made. 

 The initial semantics were that when consumed is updated, it's cleared. The use cases I had in mind where that if you are using this info to monitor scheduling of multiple components, then that component needs to be the timeout handler of all components to reason correctly about scheduling, and thus gets all the information. For the examples I wrote this was sufficient.

What is the motivating example here?

src/object/endpoint.c
```diff
@@ -226,10 +226,10 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
             do_call = thread_state_ptr_get_blockingIPCIsCall(&sender->tcbState);
 
 #ifdef CONFIG_KERNEL_MCS
-            if (do_call ||
-                seL4_Fault_get_seL4_FaultType(sender->tcbFault) != seL4_Fault_NullFault) {
+            word_t faultType = seL4_Fault_get_seL4_FaultType(sender->tcbFault);
+            if (do_call || faultType != seL4_Fault_NullFault) {
                 if ((canGrant || canGrantReply) && replyPtr != NULL) {
-                    reply_push(sender, thread, replyPtr, sender->tcbSchedContext != NULL);
+                    reply_push(sender, thread, replyPtr, faultType != seL4_Fault_Timeout);
```
pingerino, line 232: I'm confused -- why remove `sender->tcbSchedContext != null`?

src/object/endpoint.c
```diff
@@ -226,10 +226,10 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
             do_call = thread_state_ptr_get_blockingIPCIsCall(&sender->tcbState);
 
 #ifdef CONFIG_KERNEL_MCS
-            if (do_call ||
-                seL4_Fault_get_seL4_FaultType(sender->tcbFault) != seL4_Fault_NullFault) {
+            word_t faultType = seL4_Fault_get_seL4_FaultType(sender->tcbFault);
```
pingerino, line 229: the original code here optimised by not reading the fault type at all if `do_call` was set. Please put this back.

include/kernel/faulthandler.h
```diff
@@ -7,11 +7,29 @@
 #pragma once
 
 #include <object.h>
+#ifdef CONFIG_KERNEL_MCS
```
pingerino, line 10: can this commit be split into smaller changes?

pingerino: also if its  possible to do this commit without restructuring the code so we can look at the semantics of the change that would help a lot. This commit changes the behaviour and refactors code at the same time. I don't have the state in my head, so the diffs are very hard to understand. 

this looks like it should be implemented in haskell first as well, since it changes the structure of sendIPC in ways that may not work with the haskell spec.

src/object/schedcontext.c
xurtis: I'm not sure if this is realistic, but the example I had in mind is where the fault first occurs in a shared resource server with a fault handler that only attempts to recover that server without modifying the donated SC (such as via some alternate recovery SC). In this case it could reply to the client that donated the SC, and the client's fault handler would receive the fault and still be given the appropriate consumed time if it was performing its own accounting.

src/object/endpoint.c
xurtis: `reply_push` performs that check internally. The fourth argument is only used in a condition where it combined with a logical and of the sender's SC not being `NULL`.

include/kernel/faulthandler.h
xurtis: Github does not make it clear which commit you are referring to.

pingerino: "mcs: Fault if scheduling unschedulable thread"

yeah I am finding the UI a bit confusing here

xurtis: I think this was a few different commits previously but they were confusing in their own right. I'll see if I can clear up the changes.

src/object/endpoint.c
xurtis: Is this also why `do_call` is not set by the call to `sendIPC` in `sendFaultIPC`, i.e. so `do_call` implies that the IPC is not the result of a fault?

src/api/syscall.c
xurtis: This branch is required on all invocations of `handleRecv` as it is always possible that this has occurred after the current thread donated its SC in the first phase. If it has, there is always the potential that it reaches the end of `handleRecv` without an SC, regardless of what kind of IPC object is used for the receive.

xurtis:
@pingerino I've removed the change relating to reporting consumed time back to the user as it is essentially separate and can be dealt with in future if it proves to be a problem. I think I have also fixed the other changes you requested.

In splitting the commit history to clarify that larger commit I have also elaborated on each of the component changes in their commit messages.

include/kernel/faulthandler.h
pingerino: thanks, this is now much easier to review

include/kernel/faulthandler.h
```diff
@@ -7,11 +7,29 @@
...
+#include <kernel/sporadic.h>
+#endif
 
 #ifdef CONFIG_KERNEL_MCS
 static inline bool_t validTimeoutHandler(tcb_t *tptr)
 {
-    return cap_get_capType(TCB_PTR_CTE_PTR(tptr, tcbTimeoutHandler)->cap) == cap_endpoint_cap;
+    cap_t handlerCap = TCB_PTR_CTE_PTR(tptr, tcbTimeoutHandler)->cap;
```
pingerino, line 17: I think this is ok. 

Food for thought: maybe we could still set the thread to a timeout faulted state, but not deliver the fault anywhere. this would allow timeout fault handlers to 'recover' from missed faults.

src/object/schedcontext.c
```diff
@@ -284,6 +284,12 @@ void schedContext_unbindTCB(sched_context_t *sc, tcb_t *tcb)
     tcbSchedDequeue(sc->scTcb);
     tcbReleaseRemove(sc->scTcb);
 
+    /* If the unbound TCB has a timeout handler, it is triggered to
```
pingerino, line 287: nit: the commit message on this one should be 'Timeout fault on SC unbind'. SC unbind itself doesn't have a timeout.

pingerino: also forcibly is a typo in the commit message

src/object/schedcontrol.c
```diff
@@ -71,7 +71,6 @@ static exception_t invokeSchedControl_Configure(sched_context_t *target, word_t
 #endif /* ENABLE_SMP_SUPPORT */
 
     if (target->scTcb && target->scRefillMax > 0) {
-        schedContext_resume(target);
```
pingerino, line 74: can you add an assert here to make sure it's true and document the assumption?

src/kernel/thread.c
pingerino: It should mirror what happens on a page fault - if the fault isn't handled, the thread will continue to fault

src/kernel/thread.c
```diff
@@ -616,6 +616,41 @@ void endTimeslice(bool_t can_timeout_fault)
         postpone(NODE_STATE(ksCurSC));
     }
 }
+
+bool_t ensureSchedulable(tcb_t *thread, bool_t can_timeout)
```
pingerino, line 614: this function is really big and branchy and not inlined and then called in a lot of places.

If possible it would be a lot better to create a set of specific functions that handle the exact cases we are dealing with inline rather than using a generic, branch predictor smashing function.

src/object/endpoint.c
pingerino: do_call implies that we are doing a call like operation which generates a reply. So Call or a fault.

src/kernel/thread.c
xurtis: There is only one case (the SC unbind) where we don't need to perform all of these checks together. In the SC unbind case there is always no SC. In every other case we have no static guarantee on whether or not the thread has an SC or whether that SC is configured or ready.

include/kernel/faulthandler.h
xurtis: We'd need to add some way of getting that state from the thread so we could check it later but that's entirely possible. I don't think a well configured system should be able to enter this state though.

src/object/endpoint.c
xurtis: In MCS, `do_call` is set to false in `sendFaultIPC` but not on master, so it does actually seem to imply that there isn't a fault.

src/kernel/thread.c
xurtis: I can probably move this function into a header and make it inline. The common case should be that all the top level checks are true.

src/kernel/thread.c
xurtis: What do you do if you want to resume it but have it be postponed?

src/kernel/thread.c
pingerino: - in SC bind you also know that  the scheduling context is set.
- you should write the function to be fastest in the good case

something like

```fault_type_t faultType = Fault_Null_Fault
    if (likely(sc != NULL && sc->scRefillMax == 0 && refill_ready(sc)) {
        return true;
    } 

    bool valid = validTimeoutHandler(thread)
    word_t badge = 0
    if (sc == NULL) {
        faultType = seL4_Timeout_NoSC;
    } else if (sc->scRefillMax == 0) {
        faultType = seL4_Timeout_Unconfigured
        badge = sc->scBadge
    } else { /* out of time */
       faultType =  seL4_Timeout_Exhausted;
       badge = sc->scBadge
    }

    if (canTimeout && validTimeoutHandler) {
        current_fault = seL4_Fault_Timeout_new(badge, faultType);
        handleTimeout(fault);
    } else if (faultType == seL4_TimeoutExhausted) {
        postpone(thread);
    }
    return false;
}'''

File: libsel4/include/sel4/constants.h [hunk 16]
```diff
@@ -83,6 +83,16 @@ typedef enum {
 /* the size of a single extra refill */
 #define seL4_RefillSizeBytes (2 * 8)
 
+/* Reasons for why a timeout fault occurred */
+enum {
+    /* The thread was scheduled without an SC */
```
pingerino, line 88: Are seL4_Timeout_NoSC and seL4_Timeout_Unconfigured valuable as seperate values? Could this just be seL4_Timeout_InvalidSC??

src/kernel/thread.c
pingerino: any thread with a timeout fault handler faults when it has no time. There's no exceptions. You could remove the fault handler I guess.

libsel4/include/sel4/constants.h
xurtis: One is plausible in the case that a client may have performed a send that refused to donate an SC, the other means that you are probably doing SC configuration in a questionable manner (but in a way that the kernel doesn't statically prevent) and is probably more useful as a debugging diagnostic than an actual fault case that would show up in a sane system.

src/kernel/thread.c
xurtis: I'm about to push a slightly different approach that should inline a bit better.

src/object/schedcontext.c
xurtis: Fixed this

src/object/schedcontrol.c
xurtis: Added this

libsel4/include/sel4/constants.h
pingerino: Lol, seL4 doesn't have debugging diagnostics! 
In all seriousness though, we do want to keep the code small, so in general we give out as little info as possible in a release kernel ...

src/kernel/thread.c
xurtis: Okay, if that's the rule then `ensureSchedulable` shouldn't need the second argument. I'll rebase and remove it.

pingerino: err except round robin threads. although that's a bit confusing. it doesn't really matter what the impl does as setting a timeout fault for an RR thread is not really valid, so we can do what we like as long as we don't break the kernel

libsel4/include/sel4/constants.h
xurtis: I honestly wasn't sure whether to present these differently. I figured that we distinguish the cases as the actual faulting circumstances are discernible even if no real-world system would actually care about the distinction (much like many hardware fault details I guess). It also means we can write tests that more directly show that the kernel behaves as we want in these circumstances, but that in itself is not a great motivator.

src/kernel/thread.c
xurtis: Alright, you might want to check over this commit again. The `can_timeout` argument is removed as the time exhaustion applies to all non-round-robin SCs under all circumstances (so a check is added for round-robin on the SC).

The outermost checks form the inlined function but are marked unlikely so as to have the same effect as the single likely condition you have above (i.e. push the body of each unlikely branch out of where the branch predictor will go) and the inner check is consolidated into a single function which gets called via an indirect branch.

src/object/schedcontrol.c
```diff
@@ -72,8 +72,10 @@ static exception_t invokeSchedControl_Configure(sched_context_t *target, word_t
     }
 #endif /* ENABLE_SMP_SUPPORT */
```
pingerino, line 73: "When reconfiguring an SC, refill_new and refill_update will always
produce a SC that is ready and so will never need to be postponed."

Does this mean that you can violate the bandwidth constraints on the SC by reconfiguring?

This was something the initial design tried to avoid, as there are valid cases where you want to adjust the scheduling parameters but still need to be able to to scheduling analysis.

pingerino:
Looking fairly solid at this point.

Are there test cases being added to sel4test for all the new fault conditions? I'd prefer not to approve until I have reviewed those as well.

src/object/schedcontrol.c
xurtis: Ah, this comment is incorrect. `refill_update` preserves readiness, so it can't make a ready SC become not-ready.

This means the assert is invalid and the SC may need to be returned to the release queue. I don't think `refill_update` should fault though. An alternative may be to not remove it from the release queue ore ready queues at all given that `refill_update` doesn't affect the time of the head refill.

 * [ ] Fix `refill_update` changes

xurtis: I replaced the commit with this one: https://github.com/seL4/seL4/pull/216/commits/133da34c4c9fdf7468912b3391f3046361f07b4e

This keeps the effect of the `schedContext_resume` but manually inlines it and removes any unnecessary checks (although any checks removed now get asserted).

src/object/schedcontrol.c
```diff
@@ -72,14 +72,20 @@ static exception_t invokeSchedControl_Configure(sched_context_t *target, word_t
...
 
-    if (target->scTcb && target->scRefillMax > 0) {
-        schedContext_resume(target);
-        if (SMP_TERNARY(core == CURRENT_CPU_INDEX(), true)) {
-            if (isRunnable(target->scTcb) && target->scTcb != NODE_STATE(ksCurThread)) {
-                possibleSwitchTo(target->scTcb);
+    assert(target->scRefillMax > 0);
```
pingerino, line 75: Love the asserts. They help verification work out what should be true and help us discovered during modification or before proofs if we're wrong.

include/kernel/faulthandler.h
pingerino, line 17: Can you do any optimisations here with the knowledge that the timeout handler cap is installed by the kernel and cannot be deleted? I'm thinking you can set a pointer in the TCB to the endpoint object then directly check it's state without a lookup if that pointer is set.

xurtis: I don't think so. If the cap gets destroyed via revocation then there is no way to determine that the cap is being used by the TCB as a timeout EP, so the TCB would end up with a stale pointer.

pingerino: Aha, but recovation only occurs when all caps to an object are deleted. Users cannot delete the cap in the TCB Cnode without deleting the TCB. So if the TCB exists, any cap in the Cnode is a cap to the object, which continues to exist.

xurtis: Except in the case of revocation, which allows copies of caps to be removed without needing direct access to the CNode in which they reside. Revoking the untyped for the EP or the original EP capability can removed it from the TCB, so we can't statically guarantee the pointer implies presence of the capability easily.

pingerino: So when you call finaliseCap you should be able to undo the link?

pingerino: Yes, I'm well aware what revoke does.

xurtis: When you finaliseCap you don't have a reference to the TCB.

pingerino: If you can get to the TCB CNode you can get the TCB pointer easily. I don't recall if you can do that from revoke but there's likely a way to do it. You could also consider a link between the TCP and EP objects, but I suspect that's not required. 

Anyway, you don't seem interested in pursuing this idea, and I'm done offering suggestions.

xurtis: I think you *can* do it, knowing that if you know the cap is specifically being used as the timeout endpoint of a TCB, you can determine where the CNode starts and where the TCB is based on knowing which slot it is in, but in the general case you can't determine the location of a CNode from a capability slot.

xurtis: I also think that while the overheads of the checks involved here should be optimised, that they can be addressed separately as it seems that moving the necessary checks elsewhere and guaranteeing the necessary invariants required to make them correct such that the check here is lightweight are going to add significant complexity in their own right.

pingerino: Once something is verified it's very hard to change. Also what are the impact of this change on the scheduling benchmarks that use timeout faults?

xurtis:
@pingerino this implementation faults in a large number of cases that cause issue for verification. Specifically, in `cancelAllIPC` this can result in faults being triggered for threads blocked on a deleted endpoint. Do you think it would make sense to make such threads inactive rather than to fault them? I suspect that this would mean we trust that the endpoint over which we are communicating won't be deleted in some sense. I suspect this would mostly be an issue if the original untyped were revoked.

lsf37:
Bit of additional detail:

> Specifically, in `cancelAllIPC` this can result in faults being triggered for threads blocked on a deleted endpoint. Do you think it would make sense to make such threads inactive rather than to fault them? 

The main thing we didn't spot before attempting the proof was that the change causes fault IPC to be delivered during cap deletion (via `cancelAllIPC`). From the API perspective, if you're deleting a large CNode full of endpoints, you might end up delivering thousands of fault IPCs in that delete, which doesn't sound good to me. It's an artificial example, but artificial can be set up by a malicious user.

> I suspect that this would mean we trust that the endpoint over which we are communicating won't be deleted in some sense. I suspect this would mostly be an issue if the original untyped were revoked.

Yes, I think so. From the trust perspective, if someone can delete the endpoint for a server, the server already has to trust that entity to some degree at least. So think it *should* be fine, but I might be missing something in the bigger picture.

pingerino:
@xurtis @lsf37 sounds fine to me. I'm a bit worried that this is being verified already and I haven't seen any tests?

xurtis:
I'm putting together tests now (hopefully have them finished and passing by the end of the day). I have been waiting for tests to be passing on all platforms and approved before indicating what we have is what will go into master.

pingerino:
Great, thanks for the clarification.

lsf37:
Yes, tests definitely have to come in first before we're starting in earnest. I'd call what we've done so far "exploratory verification" (mostly just spec updating on a branch), to flush out any obvious bits that we missed, and it looks like it paid off :-)

Ok, so it sounds like we have a way forward without IPC in delete.

xurtis:
@pingerino changes related to recent discussion: https://github.com/seL4/seL4/pull/216/commits/961ff159531779eef602c60b993dd479ea5e70a4

pingerino:
:+1: 

Are the manual changes coming in this PR too?

xurtis:
They should be but I seem to have lost track of them. I guess I know what I'm doing with the rest of the afternoon.

xurtis:
Added manual sections, rendered versions below.

steel 2020-08-07T1506 54
steel 2020-08-07T1507 11

The fully rendered manual with changes: [manual.pdf](https://github.com/seL4/seL4/files/5039426/manual.pdf).

xurtis:
All tests (including those from https://github.com/seL4/sel4test/pull/23) pass for all platforms!

manual/parts/threads.tex
```diff
@@ -559,19 +559,36 @@ \subsection{Debug Exception: Single-stepping}
 \subsection{Timeout Fault (MCS only)}
 \label{sec:timeout-fault}
 
-Timeout faults are raised when a thread consumes all of its budget and has a
-timeout fault handler that is not a null capability.  They allow a timeout
-exception handler to take some action to restore the thread, and deliver a
-message containing the scheduling context data word, as well as the amount of
-time consumed since the last timeout fault occured on this scheduling context,
-or since \apifunc{seL4\_SchedContext\_YieldTo}{schedcontext_yieldto} or
+Timeout faults are raised when a thread loses access to time with which
```
pingerino, line 582: Good start, however I recommend some changes for readability. 

* You've dropped the important and still relevant part of the first sentence: " has a
￼timeout fault handler that is not a null capability"
* Use a bulleted list for all of the reasons you may get a timeout fault or break into subsections . If this is too big and there are nuances,introduce the topic softly then go into detail on each one. this allows people to get the gist without reading, and also allows them to get to the section they want to know about very quickly. 

```
Timeout faults are raised when a thread attempts to execute but cannot as it has no access to time, *and* if the thread has a valid timeout fault handler. Timeout faults fall into three categories: budget exhaustion, missing scheduling contexts, or invalid scheduling contexts.
```
Note the above maps to the 3 fault types.

Now follow with three subheadings and details + examples of each type of fault.

xurtis: (Note for myself):

* [ ] Use list for fault types
* [ ] Add section describing *exhausted* SC reason w/ examples
* [ ] Add section describing *no* SC reason w/ examples
* [ ] Add section describing *unconfigured* SC reason w/ examples

lsf37:
@kent-mcleod @mktnk3 do you know which parts of this PR are yet to be merged? It looks like we have done most of it in separate parts, and I'm not sure where we are in that process.

Should we close this?

PR #217, opened 2020-07-16
MCS Systems changes

xurtis:
(This description probably requires more depth).

This is a consolidation of the changes we currently wish to verify and that have already been verified with some slight alterations.

There are some differences between this branch and what verification has actually proved. The primary difference is that this branch does not prevent binding SCs to TCBs in IPC queues as we anticipate that requirement will be dropped anyway.

Most other differences are presently in layout and structure to minimise the overall delta but are not fundamentally differences in what actually happens. The verification side will instead be updated to reflect this.

### DIFF
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -141,14 +141,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
 
 
 /* Charge `usage` to the current scheduling context.
- * This function should only be called only when charging `used` will deplete
- * the head refill, resulting in refill_sufficient failing.
  *
  * @param usage the amount of time to charge.
- * @param capacity the value returned by refill_capacity. At most call sites this
- * has already been calculated so pass the value in rather than calculating it again.
  */
-void refill_budget_check(ticks_t used, ticks_t capacity);
+void refill_budget_check(ticks_t used);
 
 /*
  * Charge a the current scheduling context `used` amount from its
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -34,6 +34,13 @@
 #define MIN_BUDGET_US (2u * getKernelWcetUs() * CONFIG_KERNEL_WCET_SCALE)
 #define MIN_BUDGET    (2u * getKernelWcetTicks() * CONFIG_KERNEL_WCET_SCALE)
 
+/* By ensuring that the total budget of a scheduling context is at least
+ * twice MIN_BUDGET, we can ensure that all refills are at least
+ * MIN_BUDGET and that it is always possible to fragment into at least 2
+ * refills. */
+#define MIN_SC_BUDGET_US (2 * MIN_BUDGET_US)
+#define MIN_SC_BUDGET (2 * MIN_BUDGET)
+
 /* Short hand for accessing refill queue items */
 #define REFILL_INDEX(sc, index) (((refill_t *) (SC_REF(sc) + sizeof(sched_context_t)))[index])
 #define REFILL_HEAD(sc) REFILL_INDEX((sc), (sc)->scRefillHead)
@@ -145,6 +152,12 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
  * @param usage the amount of time to charge.
  */
 void refill_budget_check(ticks_t used);
+
+/* Charge `usage` to the current scheduling context (round-robin).
+ *
+ * @param usage the amount of time to charge.
+ */
+void refill_budget_check_round_robin(ticks_t usage);
 
 /*
  * Charge a the current scheduling context `used` amount from its
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -37,7 +37,7 @@
 /* Short hand for accessing refill queue items */
 #define REFILL_INDEX(sc, index) (((refill_t *) (SC_REF(sc) + sizeof(sched_context_t)))[index])
 #define REFILL_HEAD(sc) REFILL_INDEX((sc), (sc)->scRefillHead)
-#define REFILL_TAIL(sc) REFILL_INDEX((sc), (sc)->scRefillTail)
+#define REFILL_TAIL(sc) REFILL_INDEX((sc), refill_tail_index(sc))
 
 
 /* Scheduling context objects consist of a sched_context_t at the start, followed by a
@@ -55,29 +55,49 @@ static inline word_t refill_absolute_max(cap_t sc_cap)
 /* @return the current amount of empty slots in the refill buffer */
 static inline word_t refill_size(sched_context_t *sc)
 {
-    if (sc->scRefillHead <= sc->scRefillTail) {
-        return (sc->scRefillTail - sc->scRefillHead + 1u);
-    }
-    return sc->scRefillTail + 1u + (sc->scRefillMax - sc->scRefillHead);
+    return sc->scRefillCount;
 }
 
 /* @return true if the circular buffer of refills is current full (all slots in the
  * buffer are currently being used */
 static inline bool_t refill_full(sched_context_t *sc)
 {
-    return refill_size(sc) == sc->scRefillMax;
+    return sc->scRefillCount == sc->scRefillMax;
+}
+
+static inline bool_t refill_empty(sched_context_t *sc)
+{
+    return sc->scRefillCount == 0;
+}
+
+static inline word_t refill_tail_index(sched_context_t *sc)
+{
+    assert(sc->scRefillHead <= sc->scRefillMax);
+    assert(sc->scRefillCount <= sc->scRefillMax);
+    assert(sc->scRefillCount >= 1);
+
+    word_t index = sc->scRefillHead + sc->scRefillCount - 1;
+
+    if (index >= sc->scRefillMax) {
+        index -= sc->scRefillMax;
+    }
+
+    assert(index < sc->scRefillMax);
+    return index;
 }
 
 /* @return true if the ciruclar buffer only contains 1 used slot */
 static inline bool_t refill_single(sched_context_t *sc)
 {
-    return sc->scRefillHead == sc->scRefillTail;
+    return sc->scRefillCount == 1;
 }
 
 /* Return the amount of budget this scheduling context
  * has available if usage is charged to it. */
 static inline ticks_t refill_capacity(sched_context_t *sc, ticks_t usage)
 {
+    assert(!refill_empty(sc));
+
     if (unlikely(usage > REFILL_HEAD(sc).rAmount)) {
         return 0;
     }
@@ -91,6 +111,7 @@ static inline ticks_t refill_capacity(sched_context_t *sc, ticks_t usage)
  */
 static inline bool_t refill_sufficient(sched_context_t *sc, ticks_t usage)
 {
+    assert(!refill_empty(sc));
     return refill_capacity(sc, usage) >= MIN_BUDGET;
 }
 
@@ -102,6 +123,7 @@ static inline bool_t refill_sufficient(sched_context_t *sc, ticks_t usage)
  */
 static inline bool_t refill_ready(sched_context_t *sc)
 {
+    assert(!refill_empty(sc));
     return REFILL_HEAD(sc).rTime <= (NODE_STATE_ON_CORE(ksCurTime, sc->scCore) + getKernelWcetTicks());
 }
 
--- a/include/object/schedcontext.h
+++ b/include/object/schedcontext.h
@@ -66,6 +66,10 @@ void schedContext_bindNtfn(sched_context_t *sc, notification_t *ntfn);
 /* unbind scheduling context from a notification */
 void schedContext_unbindNtfn(sched_context_t *sc);
 
+/* Return a donated scheduling context to the notification to which it
+ * was bound if any is bound */
+void schedContext_maybeReturnToNtfn(sched_context_t *sc, tcb_t *tcb);
+
 time_t schedContext_updateConsumed(sched_context_t *sc);
 void schedContext_completeYieldTo(tcb_t *yielder);
 void schedContext_cancelYieldTo(tcb_t *yielder);
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -308,6 +308,15 @@ void refill_split_check(ticks_t usage)
     REFILL_SANITY_END(sc);
 }
 
+
+static bool_t refill_unblock_check_mergable(sched_context_t *sc)
+{
+    ticks_t amount = REFILL_HEAD(sc).rAmount;
+    ticks_t tail = NODE_STATE_ON_CORE(ksCurTime, sc->scCore) + amount;
+    bool_t enough_time = REFILL_INDEX(sc, refill_next(sc, sc->scRefillHead)).rTime <= tail;
+    return !refill_single(sc) && enough_time;
+}
+
 void refill_unblock_check(sched_context_t *sc)
 {
 
@@ -323,15 +332,11 @@ void refill_unblock_check(sched_context_t *sc)
         NODE_STATE(ksReprogram) = true;
 
         /* merge available replenishments */
-        while (!refill_single(sc)) {
+        while (refill_unblock_check_mergable(sc)) {
             ticks_t amount = REFILL_HEAD(sc).rAmount;
-            if (REFILL_INDEX(sc, refill_next(sc, sc->scRefillHead)).rTime <= NODE_STATE_ON_CORE(ksCurTime, sc->scCore) + amount) {
-                refill_pop_head(sc);
-                REFILL_HEAD(sc).rAmount += amount;
-                REFILL_HEAD(sc).rTime = NODE_STATE_ON_CORE(ksCurTime, sc->scCore);
-            } else {
-                break;
-            }
+            refill_pop_head(sc);
+            REFILL_HEAD(sc).rAmount += amount;
+            REFILL_HEAD(sc).rTime = NODE_STATE_ON_CORE(ksCurTime, sc->scCore);
         }
 
         assert(refill_sufficient(sc, 0));
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -165,6 +165,28 @@ void refill_new(sched_context_t *sc, word_t max_refills, ticks_t budget, ticks_t
     REFILL_SANITY_CHECK(sc, budget);
 }
 
+static inline void schedule_used(sched_context_t *sc, refill_t new)
+{
+    if (refill_empty(sc)) {
+        assert(new.rAmount >= MIN_BUDGET);
+        refill_add_tail(sc, new);
+    } else {
+        /* We should never try to append overlapping refills to the tail */
+        assert(new.rTime >= REFILL_TAIL(sc).rTime + REFILL_TAIL(sc).rAmount);
+
+        /* schedule the used amount */
+        if (new.rAmount < MIN_BUDGET || refill_full(sc)) {
+            /* Merge with existing tail */
+            REFILL_TAIL(sc).rTime = new.rTime - REFILL_TAIL(sc).rAmount;
+            REFILL_TAIL(sc).rAmount += new.rAmount;
+        } else {
+            refill_add_tail(sc, new);
+        }
+    }
+
+    assert(!refill_empty(sc));
+}
+
 void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, word_t new_max_refills)
 {
 
@@ -199,32 +221,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
         refill_t new = { .rAmount = (new_budget - REFILL_HEAD(sc).rAmount),
                          .rTime = REFILL_HEAD(sc).rTime + new_period
                        };
-        refill_add_tail(sc, new);
+        schedule_used(sc, new);
     }
 
     REFILL_SANITY_CHECK(sc, new_budget);
-}
-
-static inline void schedule_used(sched_context_t *sc, refill_t new)
-{
-    if (refill_empty(sc)) {
-        assert(new.rAmount >= MIN_BUDGET);
-        refill_add_tail(sc, new);
-    } else {
-        /* We should never try to append overlapping refills to the tail */
-        assert(new.rTime >= REFILL_TAIL(sc).rTime + REFILL_TAIL(sc).rAmount);
-
-        /* schedule the used amount */
-        if (new.rAmount < MIN_BUDGET || refill_full(sc)) {
-            /* Merge with existing tail */
-            REFILL_TAIL(sc).rTime = new.rTime - REFILL_TAIL(sc).rAmount;
-            REFILL_TAIL(sc).rAmount += new.rAmount;
-        } else {
-            refill_add_tail(sc, new);
-        }
-    }
-
-    assert(!refill_empty(sc));
 }
 
 void refill_budget_check(ticks_t usage)
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -219,8 +219,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
         REFILL_HEAD(sc).rAmount = new_budget;
     } else {
         /* otherwise schedule the rest for the next period */
-        refill_t new = { .rAmount = (new_budget - REFILL_HEAD(sc).rAmount),
-                         .rTime = REFILL_HEAD(sc).rTime + new_period
+        ticks_t unused = new_budget - REFILL_HEAD(sc).rAmount;
+        ticks_t true_period = MAX(new_period, new_budget);
+        refill_t new = { .rAmount = unused,
+                         .rTime = REFILL_HEAD(sc).rTime + true_period - unused,
                        };
         schedule_used(sc, new);
     }
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -227,47 +227,27 @@ static inline void schedule_used(sched_context_t *sc, refill_t new)
     assert(!refill_empty(sc));
 }
 
-void refill_budget_check(ticks_t usage, ticks_t capacity)
+void refill_budget_check(ticks_t usage)
 {
     sched_context_t *sc = NODE_STATE(ksCurSC);
     /* this function should only be called when the sc is out of budget */
-    assert(capacity < MIN_BUDGET || refill_full(sc));
     assert(sc->scPeriod > 0);
     REFILL_SANITY_START(sc);
 
-    if (capacity == 0) {
-        while (REFILL_HEAD(sc).rAmount <= usage) {
-            /* exhaust and schedule replenishment */
-            usage -= REFILL_HEAD(sc).rAmount;
-            if (refill_single(sc)) {
-                /* update in place */
-                REFILL_HEAD(sc).rTime += sc->scPeriod;
-            } else {
-                refill_t old_head = refill_pop_head(sc);
-                old_head.rTime = old_head.rTime + sc->scPeriod;
-                schedule_used(sc, old_head);
-            }
-        }
-
-        /* budget overrun */
-        if (usage > 0) {
-            /* budget reduced when calculating capacity */
-            /* due to overrun delay next replenishment */
-            REFILL_HEAD(sc).rTime += usage;
-            /* merge front two replenishments if times overlap */
-            if (!refill_single(sc) &&
-                REFILL_HEAD(sc).rTime + REFILL_HEAD(sc).rAmount >=
-                REFILL_INDEX(sc, refill_next(sc, sc->scRefillHead)).rTime) {
-
-                refill_t refill = refill_pop_head(sc);
-                REFILL_HEAD(sc).rAmount += refill.rAmount;
-                REFILL_HEAD(sc).rTime = refill.rTime;
-            }
+    while (unlikely(REFILL_HEAD(sc).rAmount <= usage)) {
+        /* exhaust and schedule replenishment */
+        usage -= REFILL_HEAD(sc).rAmount;
+        if (refill_single(sc)) {
+            /* update in place */
+            REFILL_HEAD(sc).rTime += sc->scPeriod;
+        } else {
+            refill_t old_head = refill_pop_head(sc);
+            old_head.rTime = old_head.rTime + sc->scPeriod;
+            schedule_used(sc, old_head);
         }
     }
 
-    capacity = refill_capacity(sc, usage);
-    if (capacity > 0 && refill_ready(sc)) {
+    if (usage > 0) {
         refill_split_check(usage);
     }
 
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+#include <util.h>
 #include <types.h>
 #include <api/failures.h>
 #include <object/structures.h>
@@ -283,18 +284,12 @@ void refill_budget_check(ticks_t usage)
     assert(sc->scPeriod > 0);
     REFILL_SANITY_START(sc);
 
-    while (unlikely(REFILL_HEAD(sc).rAmount <= usage)) {
-        /* exhaust and schedule replenishment */
-        usage -= REFILL_HEAD(sc).rAmount;
-        if (refill_single(sc)) {
-            /* update in place */
-            REFILL_HEAD(sc).rTime += sc->scPeriod;
-        } else {
-            refill_t old_head = refill_pop_head(sc);
-            old_head.rTime = old_head.rTime + sc->scPeriod;
-            schedule_used(sc, old_head);
-        }
-    }
+    /* This implementation is simply the most simple operation that
+     * guarantees the post-condition. This is used as a stop-gap as the
+     * proof for integer overflow in the correct solution is difficult
+     * to prove. */
+
+    usage = MIN(usage, REFILL_HEAD(sc).rAmount);
 
     if (usage > 0) {
         refill_split_check(usage);
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -25,10 +25,6 @@
  * [x][t][][][][h][x][x]
  *
  * The queue has a minimum size of 1, so it is possible that h == t.
- *
- * The queue is implemented as head + tail rather than head + size as
- * we cannot use the mod operator on all architectures without accessing
- * the fpu or implementing divide.
  */
 
 /* return the index of the next item in the refill queue */
@@ -48,14 +44,14 @@ UNUSED static inline void print_index(sched_context_t *sc, word_t index)
 
 UNUSED static inline void refill_print(sched_context_t *sc)
 {
-    printf("Head %lu tail %lu\n", sc->scRefillHead, sc->scRefillTail);
-    word_t current = sc->scRefillHead;
-    /* always print the head */
-    print_index(sc, current);
+    printf("Head %lu length %lu\n", sc->scRefillHead, sc->scRefillCount);
 
-    while (current != sc->scRefillTail) {
-        current = refill_next(sc, current);
+    word_t current = sc->scRefillHead;
+    word_t seen = 0;
+    while (seen != sc->scRefillCount) {
         print_index(sc, current);
+        current = refill_next(sc, current);
+        seen += 1;
     }
 
 }
@@ -66,14 +62,16 @@ static UNUSED bool_t refill_ordered(sched_context_t *sc)
 {
     word_t current = sc->scRefillHead;
     word_t next = refill_next(sc, sc->scRefillHead);
+    word_t seen = 1;
 
-    while (current != sc->scRefillTail) {
+    while (seen != sc->scRefillCount) {
         if (!(REFILL_INDEX(sc, current).rTime <= REFILL_INDEX(sc, next).rTime)) {
             refill_print(sc);
             return false;
         }
         current = next;
         next = refill_next(sc, current);
+        seen += 1;
     }
 
     return true;
@@ -98,12 +96,14 @@ static UNUSED bool_t refill_ordered(sched_context_t *sc)
 /* compute the sum of a refill queue */
 static UNUSED ticks_t refill_sum(sched_context_t *sc)
 {
-    ticks_t sum = REFILL_HEAD(sc).rAmount;
+    ticks_t sum = 0;
     word_t current = sc->scRefillHead;
+    word_t seen = 0;
 
-    while (current != sc->scRefillTail) {
-        current = refill_next(sc, current);
+    while (seen < sc->scRefillCount) {
         sum += REFILL_INDEX(sc, current).rAmount;
+        current = refill_next(sc, current);
+        seen += 1;
     }
 
     return sum;
@@ -112,12 +112,13 @@ static UNUSED ticks_t refill_sum(sched_context_t *sc)
 /* pop head of refill queue */
 static inline refill_t refill_pop_head(sched_context_t *sc)
 {
-    /* queues cannot be smaller than 1 */
-    assert(!refill_single(sc));
+    /* queues cannot be empty */
+    assert(!refill_empty(sc));
 
     UNUSED word_t prev_size = refill_size(sc);
     refill_t refill = REFILL_HEAD(sc);
     sc->scRefillHead = refill_next(sc, sc->scRefillHead);
+    sc->scRefillCount -= 1;
 
     /* sanity */
     assert(prev_size == (refill_size(sc) + 1));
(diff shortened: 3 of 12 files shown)


src/object/schedcontrol.c
```diff
@@ -102,7 +102,9 @@ static exception_t decodeSchedControl_Configure(word_t length, cap_t cap, extra_
     }
 
     time_t budget_us = mode_parseTimeArg(0, buffer);
+    ticks_t budget = usToTicks(budget_us);
     time_t period_us = mode_parseTimeArg(TIME_ARG_SIZE, buffer);
+    ticks_t period = usToTicks(period_us);
```
pingerino, line 106: If I remember correctly I deliberately did the comparisons on us to avoid overflow.

include/kernel/sporadic.h
```diff
@@ -37,7 +37,7 @@
 /* Short hand for accessing refill queue items */
 #define REFILL_INDEX(sc, index) (((refill_t *) (SC_REF(sc) + sizeof(sched_context_t)))[index])
 #define REFILL_HEAD(sc) REFILL_INDEX((sc), (sc)->scRefillHead)
-#define REFILL_TAIL(sc) REFILL_INDEX((sc), (sc)->scRefillTail)
+#define REFILL_TAIL(sc) REFILL_INDEX((sc), refill_tail_index(sc))
```
pingerino, line 47: why rewrite this data structure? Is this easier for verification?

include/kernel/sporadic.h
```diff
@@ -141,14 +141,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget,
 
 
 /* Charge `usage` to the current scheduling context.
- * This function should only be called only when charging `used` will deplete
- * the head refill, resulting in refill_sufficient failing.
  *
  * @param usage the amount of time to charge.
- * @param capacity the value returned by refill_capacity. At most call sites this
```
pingerino, line 126: this comment is exactly why I passed capacity - to avoid recalculating it all the time

include/kernel/sporadic.h
```diff
@@ -141,14 +141,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget,
 
 
 /* Charge `usage` to the current scheduling context.
- * This function should only be called only when charging `used` will deplete
```
pingerino, line 122: is this no longer true??

src/kernel/sporadic.c
```diff
@@ -227,47 +227,27 @@ static inline void schedule_used(sched_context_t *sc, refill_t new)
...
-void refill_budget_check(ticks_t usage, ticks_t capacity)
+void refill_budget_check(ticks_t usage)
 {
     sched_context_t *sc = NODE_STATE(ksCurSC);
     /* this function should only be called when the sc is out of budget */
-    assert(capacity < MIN_BUDGET || refill_full(sc));
     assert(sc->scPeriod > 0);
     REFILL_SANITY_START(sc);
 
-    if (capacity == 0) {
```
pingerino, line 233: I don't know if this change is correct or not and I don't have time to go through it. Can you explain why this can be removed?

src/object/schedcontrol.c
xurtis: I see. I think this change was made because it was difficult to determine the bound in terms of ticks on the verification side. I will discuss with verification, I could also do the check on both.

include/kernel/sporadic.h
xurtis: I think at the time it helped us reason about what we were trying to verify and what the invariants were. There is a change in the implementation of refill split check that relies on the consistent state of an empty queue.

include/kernel/sporadic.h
xurtis: I think this still holds.

include/kernel/sporadic.h
xurtis: The only callsite I can find where the capacity needed to be calculated already anyway was in `checkBudget` where the case where it was needed was when an SC had exhausted all of its budget when attempting a syscall. I recall this being requested from verification to reduce complexity.

src/kernel/sporadic.c
xurtis: The initial loop will charge any complete refills as if they had been used which preserves future bandwidth in the presence of an overrun. The second part delays a refill that isn't completely used by the amount of time overrun. Delaying may work for the logic of sporadic server which does not attempt to preserve temporal isolation but only delaying will not preserve bandiwdth as it can still result in a subsequent period window with greater than proportional utilisation. Dealing with a partially consumed refill in an overrun by simply using `refill_split_check` will charge it in the exact same manner as if it had been consumed which should actually preserve future bandwidth in the event of overrun.

src/object/schedcontrol.c
```diff
@@ -28,9 +28,11 @@ static exception_t invokeSchedControl_Configure(sched_context_t *target, word_t
 #ifdef ENABLE_SMP_SUPPORT
             if (target->scCore == getCurrentCPUIndex()) {
 #endif /* ENABLE_SMP_SUPPORT */
-                if (checkBudget()) {
-                    commitTime();
-                }
+                /* This could potentially mutate state but if it returns
+                 * true no state was modified, thus removing it should
+                 * be the same. */
+                assert(checkBudget());
```
kent-mcleod, line 34: Is this in an assert because the if the current SC is performing this invocation then checkBudget must have returned true upon entry to the kernel?

include/kernel/sporadic.h
```diff
@@ -55,29 +55,49 @@ static inline word_t refill_absolute_max(cap_t sc_cap)
...
 
 /* @return true if the circular buffer of refills is current full (all slots in the
  * buffer are currently being used */
 static inline bool_t refill_full(sched_context_t *sc)
 {
-    return refill_size(sc) == sc->scRefillMax;
+    return sc->scRefillCount == sc->scRefillMax;
+}
+
+static inline bool_t refill_empty(sched_context_t *sc)
```
kent-mcleod, line 75: I don't understand what `refill_empty` signifies? Before this commit, the smallest size of the queue was when `sc->scRefillHead == sc->scRefillTail` signifying that the queue had a single element (assuming `scRefillMax > 0`).
Now it seems that `scRefillCount == 1` is the new way of checking this.  When would `refill_empty` return true and so how has `schedule_used` been changed to incorporate it (it removes `refill_single` and adds `refill_empty`)?

src/kernel/sporadic.c
pingerino: ok. Can you add this to a comment stating why it's different with reference to the paper? Might be a good idea to run this via Björn

include/kernel/sporadic.h
pingerino: Can you put the comment back then please

include/kernel/sporadic.h
pingerino: ok. It's pretty important to track why changes like this were made.

include/kernel/sporadic.h
pingerino: "At the time"? Is that still true? rewriting a data structure in the kernel that is tried and tested over multiple years should not be done trivially. I spent a good amount of time getting this right. Can someone from verification please comment and confirm? 

If the answer is 'well we verified this now' I'll be fairly disappointed.

src/object/schedcontrol.c
pingerino: The important thing to work out is what are the bounds you have now put on what these values can be. IIRC I chose these numbers to minimise restrictions on what the user could use and what we could check without overflowing. One way resulted in much more reduced values.

include/kernel/sporadic.h
kent-mcleod: It seems like `capacity == 0` was also used by `handleYield` to ensure that the full timeslice is consumed instead of just what has been consumed so far?

src/object/schedcontrol.c
pingerino, line 34: Have you run all of the multicore tests and benchmarks on this? I added this when doing multicore benchmarks as I saw problems. Could be redundant, could also be only relevant on multicore.

src/kernel/sporadic.c
```diff
@@ -323,15 +332,11 @@ void refill_unblock_check(sched_context_t *sc)
...
         /* merge available replenishments */
-        while (!refill_single(sc)) {
+        while (refill_unblock_check_mergable(sc)) {
             ticks_t amount = REFILL_HEAD(sc).rAmount;
-            if (REFILL_INDEX(sc, refill_next(sc, sc->scRefillHead)).rTime <= NODE_STATE_ON_CORE(ksCurTime, sc->scCore) + amount) {
-                refill_pop_head(sc);
-                REFILL_HEAD(sc).rAmount += amount;
-                REFILL_HEAD(sc).rTime = NODE_STATE_ON_CORE(ksCurTime, sc->scCore);
-            } else {
-                break;
```
pingerino, line 348: huh. Should this be added to the verification style guide for the kernel?

include/kernel/sporadic.h
```diff
@@ -34,6 +34,13 @@
 #define MIN_BUDGET_US (2u * getKernelWcetUs() * CONFIG_KERNEL_WCET_SCALE)
 #define MIN_BUDGET    (2u * getKernelWcetTicks() * CONFIG_KERNEL_WCET_SCALE)
 
+/* By ensuring that the total budget of a scheduling context is at least
```
pingerino, line 37: only reviewing based on commit message: how can you ensure the first split is always MIN_BUDGET? In the original design, I reasoned that the minimum refills is always 2 and the first split (tail) can be < MIN_BUDGET. Does this result in consuming the budget faster by forcing the initial split to be MIN_BUDGET? 
Note that MIN_BUDGET is based on kernel WCET which doesn't always get used (rarely, in fact) so we shouldn't force that to be minimal consumption.
The original invariant was that the HEAD must be min budget.

pingerino: Just picked up on the TWICE too. TWICE minbudget for a refill is quite large as a minimum consumed value.

src/kernel/sporadic.c
```diff
@@ -227,6 +222,51 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget,
     REFILL_SANITY_CHECK(sc, new_budget);
 }
 
+void refill_budget_check_round_robin(ticks_t usage)
```
pingerino, line 234: why is there a refill check for round robin? RR threads should be much simpler than this, where tail = consumed so far, head = what's left of this timeslice. Finish timeslice = tail = 0, head = budget.

src/kernel/thread.c
```diff
@@ -583,9 +583,7 @@ void chargeBudget(ticks_t consumed, bool_t canTimeoutFault, word_t core, bool_t
 {
 
     if (isRoundRobin(NODE_STATE_ON_CORE(ksCurSC, core))) {
-        assert(refill_size(NODE_STATE_ON_CORE(ksCurSC, core)) == MIN_REFILLS);
-        REFILL_HEAD(NODE_STATE_ON_CORE(ksCurSC, core)).rAmount += REFILL_TAIL(NODE_STATE_ON_CORE(ksCurSC, core)).rAmount;
-        REFILL_TAIL(NODE_STATE_ON_CORE(ksCurSC, core)).rAmount = 0;
+        refill_budget_check_round_robin(consumed);
```
pingerino, line 586: I don't understand this change. RR threads don't really have refills, we just abuse the space available for refills to account for RR threads. They have nothing to do with sporadic servers.

src/kernel/sporadic.c
```diff
@@ -150,7 +150,11 @@ void refill_new(sched_context_t *sc, word_t max_refills, ticks_t budget, ticks_t
     /* full budget available */
     REFILL_HEAD(sc).rAmount = budget;
     /* budget can be used from now */
-    REFILL_HEAD(sc).rTime = NODE_STATE_ON_CORE(ksCurTime, core);
+    if (period == 0) {
```
pingerino, line 154: does this change makes proof easier? I was trying to avoid adding more branches to the code - e.g if I could avoid logic like

```
if (rr) { do this } else { //sporadic do that}
```

for performance

src/kernel/sporadic.c
```diff
@@ -219,8 +219,10 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget,
         REFILL_HEAD(sc).rAmount = new_budget;
     } else {
         /* otherwise schedule the rest for the next period */
-        refill_t new = { .rAmount = (new_budget - REFILL_HEAD(sc).rAmount),
-                         .rTime = REFILL_HEAD(sc).rTime + new_period
+        ticks_t unused = new_budget - REFILL_HEAD(sc).rAmount;
```
pingerino, line 223: it was never the intention for round robin threads to have sporadic server refill logic applied to them.

IIRC this wasn't called on RR threads. But I could be remembering wrong.

src/object/notification.c
```diff
@@ -259,6 +259,9 @@ void completeSignal(notification_t *ntfnPtr, tcb_t *tcb)
         badge = notification_ptr_get_ntfnMsgIdentifier(ntfnPtr);
         setRegister(tcb, badgeRegister, badge);
         notification_ptr_set_state(ntfnPtr, NtfnState_Idle);
+#ifdef CONFIG_KERNEL_MCS
+        maybeDonateSchedContext(tcb, ntfnPtr);
```
pingerino, line 257: nice

include/object/schedcontext.h
```diff
@@ -66,6 +66,10 @@ void schedContext_bindNtfn(sched_context_t *sc, notification_t *ntfn);
 /* unbind scheduling context from a notification */
 void schedContext_unbindNtfn(sched_context_t *sc);
 
+/* Return a donated scheduling context to the notification to which it
```
pingerino, line 69: If I am reading the commit message correctly, this changes the semantics such that *any* blocking will return a notification SC to the notification?

The semantics I intended (even if I didfn't write it down correctly) were that blocking on the notification, or blocking on an endpoint when you are bound (anything that *ends the notification operation normally*) the SC would go back.  

What should not happen is that an unrelated blocking operation causes the SC to return. I need to be able to do blocking I/O via other notifications. Does this still work with this change?

src/kernel/sporadic.c
```diff
@@ -283,18 +284,12 @@ void refill_budget_check(ticks_t usage)
...
-        if (refill_single(sc)) {
-            /* update in place */
-            REFILL_HEAD(sc).rTime += sc->scPeriod;
-        } else {
-            refill_t old_head = refill_pop_head(sc);
-            old_head.rTime = old_head.rTime + sc->scPeriod;
-            schedule_used(sc, old_head);
-        }
-    }
+    /* This implementation is simply the most simple operation that
```
pingerino, line 287: whoa. What are the implications of this change? Are we going to fix it later?

You could likely rewrite the loop so it more obviously terminates, assuming that's the problem.

pingerino:
I've finished a pass of this. Note I don't have the time to look at the code in intricate detail. The descriptive commit message are really useful, thanks, and make the review possible.

include/kernel/sporadic.h
MitchellBuckley: scRefillTail and refill_tail_index don't exist in the abstract specification. The abstract model uses a boring old isabelle list which will later (soon) be refined against the concrete implementations in Haskell and C. So I'd be surprised if we requested this. Unless this change was made recently, I don't think it was us. @lsf37 may know more.

In either case "we verified this now" is certainly not the answer.

src/object/schedcontrol.c
xurtis: It seems that would depend on the timer in use and whether it was more or less coarse than a microsecond. The minimum of both may end up being the effective upper bound anyway.

include/kernel/sporadic.h
xurtis: This is what ended up being verified and it also passes the tests from the test suite.

include/kernel/sporadic.h
xurtis: Oh, I didn't realise it had been removed.

src/object/schedcontrol.c
xurtis: Yes

xurtis: I will have to look into that.

src/kernel/sporadic.c
xurtis: I think the proofs end up having to distinguish round-robin and spordaic almost entirely. This particular change may have just been to make it trivial to show that a round-robin thread was always ready by the logic of sporadic tasks.

src/kernel/sporadic.c
xurtis: The problem isn't that it terminates, termination is fairly straightforward. The issue is that charging overrun by charging time in advance can push refills into the future by `(overrun/budget)*period` and it's been difficult to determine under what circumstances that would lead to integer overflow. We need to identify an upper bound for the overrun amount (we impose the upper bound on `period` from your PR and the lower bound on `budget` results from the WCET).

src/kernel/sporadic.c
xurtis: It probably should. I also need to check with verification about what kinds of expressions can be used in conditions.

src/kernel/sporadic.c
xurtis: That wasn't clear to me. All of the checks for readiness are applied to round robin tasks. I also wanted to make sure that if a round-robin thread was pre-empted it didn't get an entire budget back when it was resumed. If it did it could potentially starve all other threads at the same prio.

src/kernel/sporadic.c
xurtis: I will look into simplifying RR independently of sporadic. I think because we had to prove a large number of invariants about all SCs we lost track of what was only particular to sporadic SCs.

include/object/schedcontext.h
xurtis: It does not, this was an overzealous change to resolve an issue where, after receiving an SC from a bound notification, the only way to return it was to recv directly on the bound notification. Kent has suggested a refinement where rather than returning on any IPC it returns on any receive, but this may still be too broad.

include/kernel/sporadic.h
xurtis: We somehow ended up with the invariant that all refills should be MIN_BUDGET. I can see why the invariant only needs to be on the head. I also only found the PR where you implement that last week.

This would consume time faster, but it does so by removing time at the end of the refill and delaying it. The effect would be that blocking (without donation) or being interrupted early would delay time at the end of the refill, although that time would still fall within the same period window.

src/object/schedcontrol.c
pingerino: I'm not sure this was practical for all of the systems we have seL4 ported to. Some of them are very high frequency.

include/kernel/sporadic.h
pingerino: @xurtis your answer seems contradictory to @MitchellBuckleys? 

In general it's not great rewrite code just because without a solid reason.

include/kernel/sporadic.h
pingerino: can you justify why consuming time faster is appropriate? I'd really like to see the impact in benchmarks before accepting this change.

src/kernel/sporadic.c
pingerino: All of the checks for readiness are applied to round robin tasks. --> yup, to minimise branches
 If it did it could potentially starve all other threads at the same prio. --> this is confusing. it should accumulate budget in the tail refill.

include/kernel/sporadic.h
xurtis: I wasn't saying that it was appropriate or justified, just clarifying what the behaviour there is.

include/kernel/sporadic.h
xurtis: `refill_split_check` changed to pop the head off of the refill queue when charging the head wouldn't leave a large enough refill to be larger than `MIN_BUDGET`. In this case, when the charged refill is added back to the end, the queue may now be empty.

include/kernel/sporadic.h
lsf37: We've verified the abstract level of this only at the moment (which doesn't have either count or tail as @MitchellBuckleys says), and modelled it in Haskell (which does model the pointers). Both versions (head/count and head/tail) are consistent and work as data structures on the level we've verified, but head/count can be empty whereas head/tail always has at least one element. Changing it back would be fairly low impact currently (about 1 work day).

I don't think we did the head/count change for verification, the reason was that some other change required the possibility of the list temporarily being empty on an active SC. Head/count is slightly neater for verification, but if the emptiness reason goes away, there is no problem verifying the head/tail version.

src/object/schedcontrol.c
kent-mcleod: I think that because the ticks_t is 64bits wide, even with a 1terahertz clock rate `getMaxUsToTicks()` is over 500,000 years in duration. I can't think of a way that this loses any precision unless the clock speed is significantly less than 1 MHz which I don't think is too common.  I may be wrong though as this makes my head hurt

include/kernel/sporadic.h
pingerino: Why does it need to be empty?

include/kernel/sporadic.h
kent-mcleod: Right, but in this commit when the queue has a single refill, `refill_split_check` is updating it in place and not popping off the head.  So there isn't a way for refill_empty to return true.

include/kernel/sporadic.h
lsf37: @xurtis, can you summarise? I thought I had seen it pop up somewhere, but I can't find it any more.

lsf37: Ah, it's here:
> refill_split_check changed to pop the head off of the refill queue when charging the head wouldn't leave a large enough refill to be larger than MIN_BUDGET. In this case, when the charged refill is added back to the end, the queue may now be empty.

include/object/schedcontext.h
kent-mcleod: My suggestion was the same as @pingerino, a receive on the notification object or any endpoint if the notification is bound.  Although this could still be too restrictive - it doesn't seem that the kernel can unambiguously know when the thread means to _end the notification operation normally_.  For example a server that wishes to rate limit the amount of messages it receives could wish to mostly block on the notification object for a periodic timer notification and then perform a seL4_Recv to get a message won't be able to ensure that it's SC doesn't go back to the notification.  It seems that returning the donated SC to the notification object should be a similar process to that of one donated from an endpoint - a reply object and a NBSendWait or NBSendRecv operation...

include/kernel/sporadic.h
pingerino: Why was that change made though? This was handled in the existing code

src/object/schedcontrol.c
pingerino: I'm sure running sel4test on all platforms supported by MCS will reveal why this needed to be done.

Note: I realise that there was a deadline to try and get all of the MCS changes through this week, but this really shouldn't be done until *all* tests pass on all platforms. Is that the case?

src/object/objecttype.c
```diff
@@ -717,6 +717,14 @@ exception_t decodeInvocation(word_t invLabel, word_t length,
                                    slot, excaps, call, buffer);
 
     case cap_domain_cap:
+#ifdef CONFIG_KERNEL_MCS
```
pingerino, line 720: I'm happy with this commit being merged

src/object/endpoint.c
```diff
@@ -226,7 +226,9 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
             if (do_call ||
                 seL4_Fault_get_seL4_FaultType(sender->tcbFault) != seL4_Fault_NullFault) {
                 if ((canGrant || canGrantReply) && replyPtr != NULL) {
-                    reply_push(sender, thread, replyPtr, sender->tcbSchedContext != NULL);
+                    bool_t canDonate = sender->tcbSchedContext != NULL
```
pingerino, line 229: this commit can also be merged, as long as it passes seL4test on all platforms. 

Ideally the benchmarks would run too as they stress test timeout faults more.

src/kernel/sporadic.c
pingerino: I'm struggling to understand the impact of this change on scheduling. I'm really not comfortable with it at all without taking the time to sit down and go through it line by line. Can someone explain in detail what changes? I need to know the exact implications.

src/object/objecttype.c
```diff
@@ -353,8 +353,10 @@ bool_t CONST sameRegionAs(cap_t cap_a, cap_t cap_b)
 #ifdef CONFIG_KERNEL_MCS
     case cap_sched_context_cap:
         if (cap_get_capType(cap_b) == cap_sched_context_cap) {
-            return cap_sched_context_cap_get_capSCPtr(cap_a) ==
-                   cap_sched_context_cap_get_capSCPtr(cap_b);
+            return (cap_sched_context_cap_get_capSCPtr(cap_a) ==
+                    cap_sched_context_cap_get_capSCPtr(cap_b)) &&
```
pingerino, line 357: this commit seems fine

src/object/notification.c
```diff
@@ -45,7 +45,14 @@ static inline void maybeDonateSchedContext(tcb_t *tcb, notification_t *ntfnPtr)
         sched_context_t *sc = SC_PTR(notification_ptr_get_ntfnSchedContext(ntfnPtr));
         if (sc != NULL && sc->scTcb == NULL) {
             schedContext_donate(sc, tcb);
-            refill_unblock_check(sc);
+            if (sc != NODE_STATE(ksCurSC)) {
```
pingerino, line 48: I think this change is fine to merge too

src/object/notification.c
```diff
@@ -50,16 +50,6 @@ static inline void maybeDonateSchedContext(tcb_t *tcb, notification_t *ntfnPtr)
         }
     }
 }
-
-static inline void maybeReturnSchedContext(notification_t *ntfnPtr, tcb_t *tcb)
-{
```
pingerino, line 55: I don't understand this change. The "previous commit" just adds an if? I'm confused.

Nit: it's nice if instead of 'previous commit' you use the hash, alhough this is a pain for rebasing.

pingerino: I thibnk github is confusing me - I'm not sure which commit "previous" is. But if its the other commit in order then this is commit is not relevant any more?

src/object/notification.c
pingerino: this commit can go in

src/kernel/sporadic.c
pingerino: @xurtis the semantics should be like so:

- if a round robin thread is preempted and it will be appended to the end of the scheduling queue, it should have it's budget completely refilled
- only if the round robin thread is preempted and is going to be placed at the head of its scheduling queue should it's budget not be completely refilled

src/kernel/sporadic.c
pingerino: "may have" ??  I'm not convinced about merging this commit

include/kernel/sporadic.h
pingerino: I'm not convinced this commit should be merged. If threads always consume MIN_BUDGET I think they will use up their budget unreasonably fast. Also the delta on the algorithm is really large, so I'd need to commit several hours to reviewing this commit properly.

src/kernel/sporadic.c
```diff
@@ -199,34 +221,12 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget,
         refill_t new = { .rAmount = (new_budget - REFILL_HEAD(sc).rAmount),
                          .rTime = REFILL_HEAD(sc).rTime + new_period
                        };
-        refill_add_tail(sc, new);
+        schedule_used(sc, new);
```
pingerino, line 228: what is the differefnce between schedule_used and refill_add_tail? I recall them being actually semantically different.

src/kernel/sporadic.c
pingerino: happy for this commit to go in assuming that the change is semantically equivalent to what it was before with the break

pingerino:
I've tagged the commits which I am comfortable with being merged.

In future these commits really should go up one by one and with everyone fresh on what is happening. 

I'm really, really uncertain about many of the changes to the sporadic server algorithm. It's very hard to get these algorithms correct and the implementation was modelled on published work (with schedulability tests & proofs) and tested extensively with benchmarks. seL4 Proofs and eyeballs are not going to guarantee the timing behaviour. 

Really they need to have serious scheduling analysis applied similar to what is seen in most scheduling papers which show that the algorithms result in the scheduling desired.

src/object/schedcontrol.c
kent-mcleod: After going through all of the platforms we currently support, there are a couple of platforms with clock speeds around 1MHz order-of-magnitude with the smallest being 1.2MHz for the hikey and 1MHz for the hifive.  Changing the MIN_BUDGET comparison to use ticks instead of cycles should still work as there is always a distinct cycle value for every us value (Only while the timer frequency is > 1MHz).  So if we keep the getMaxUsToTicks() comparison in us, and never have platforms with a clock < 1MHz frequency then this change shouldn't change the observable behavior of this function.

kent-mcleod: Although this doesn't seem to agree with this part of the commit message:
>  This is needed as the conversion is not guaranteed to preserve the properties of the checks due to possible truncation.

How does verification treat `usToTicks`?

src/object/schedcontrol.c
kent-mcleod: I looked back through the history for this and when the code was added, checkBudget was used to test whether to do a commitTime or chargeBudget based on if the SC had used up its budget or not and also applied to SCs currently running on separate cores:
```
        if (NODE_STATE_ON_CORE(ksCurSC, target->scCore) == target) {
            ticks_t capacity = refill_capacity(target, NODE_STATE_ON_CORE(ksConsumed, target->scCore));
            if (checkBudget()) {
                commitTime();
            } else {
                chargeBudget(capacity, NODE_STATE_ON_CORE(ksConsumed, target->scCore), false);
            }
        }
```
Later on the chargeBudget was moved into a branch that would only happen if the SC was on a different core:
```
        if (NODE_STATE_ON_CORE(ksCurSC, target->scCore) == target) {
#ifdef ENABLE_SMP_SUPPORT
            if (target->scCore == getCurrentCPUIndex()) {
                if (checkBudget()) {
                    commitTime();
                }
#ifdef ENABLE_SMP_SUPPORT
            } else {
                /* if its a remote core, manually charge the budget */
                ticks_t capacity = refill_capacity(target, NODE_STATE_ON_CORE(ksConsumed, target->scCore));

                chargeBudget(capacity, NODE_STATE_ON_CORE(ksConsumed, target->scCore), false, target->scCore, false);
                doReschedule(target->scCore);
            }
        }
```
Now that the other core is treated in a different condition, the `checkBudget()` in the case where the SC is on the current core is known to always return true due to the `checkBudget()` performed at the start of `handleSyscall()` and so treating it as an assert here now seems to be ok.

pingerino: In that case this commit should also be fine

src/object/notification.c
kent-mcleod: I left this commit out of https://github.com/seL4/seL4/pull/218 at the moment, because we still aren't sure about the next commit that returns the SC if the thread blocks.  My understanding of the current behavior:
- A notification will only donate its SC if a passive thread waits on it, if the notification is bound and the passive server waits on an endpoint, the notification will be delivered but the SC won't be donated.  This will cause the passive thread to be made runnable but not schedulable and it will no longer be in any queue.
- A donated notification SC will only be returned if the thread waits on the notification again.

My preference is that we change the donation behavior of bound notifications in the same PR as we change how they get returned.

lsf37:
>I'm really, really uncertain about many of the changes to the sporadic server algorithm. It's very hard to get these algorithms correct and the implementation was modelled on published work (with schedulability tests & proofs) and tested extensively with benchmarks. seL4 Proofs and eyeballs are not going to guarantee the timing behaviour.

You are completely right, the current proofs will not guarantee anything about the system-level timing behaviour. If there is time we'd really like to add some properties for that, but not for this deadline. 

For the loop specifically, I don't understand the underlying problems deeply enough to have an informed opinion. Ideally we'd re-run those tests and benchmarks that determined that the algorithm with loop was fine. I assume @xurtis has run everything we have in our test suite (@xurtis can you conform?), but it'd be nice if we could rebuild at least some of the stress testing from the original paper, and that may well be an option for the future (would certainly make me much more comfortable with changes).

The proof problems for the loop are real in the sense that if there are overflows we do know that the algorithm does not work for those cases, i.e. a malicious user-level component can configure values that break invariants the rest of the code relies on, which may lead to crashes in random other places (not the loop). So we need to do something. One option would be for the code to check for those overflow cases we can't prove and do something sensible when it happens. This may still screw with the behaviour of the objects in that operation in those corner cases, but hopefully in a slightly better way. It would also not change the behaviour of any existing systems that don't hit those cases, i.e. makes the situation slightly better overall even if it doesn't solve it.

This would then allow us to defer an actual solution to later.

src/object/notification.c
pingerino: The initial design was the opposite of the above point #1 - we want to be able to have passive servers which receive notifications.

so the semantics should be as follows:

* passive threads receives SCs any notification  notification objects if they receive a notification, regardless of whether this is via a bound endpoint or directly blocking on the notification
* If a passive thread _T_ is running on _SC_ which is received from notification _n_ then it should be returned to _n_ if _T_ blocks on its bound endpoint, or if _T_ blocks on _n_.

include/kernel/sporadic.h
xurtis: We had introduced invariants that (may not have been necessary or may be explicitly unwanted) that `refill_split_check` didn't satisfy, namely those preventing overlap of refills and the MIN_BUDGET requirement on all refills.

pingerino:
> > I'm really, really uncertain about many of the changes to the sporadic server algorithm. It's very hard to get these algorithms correct and the implementation was modelled on published work (with schedulability tests & proofs) and tested extensively with benchmarks. seL4 Proofs and eyeballs are not going to guarantee the timing behaviour.
> 
> You are completely right, the current proofs will not guarantee anything about the system-level timing behaviour. If there is time we'd really like to add some properties for that, but not for this deadline.
> 
> For the loop specifically, I don't understand the underlying problems deeply enough to have an informed opinion. Ideally we'd re-run those tests and benchmarks that determined that the algorithm with loop was fine. I assume @xurtis has run everything we have in our test suite (@xurtis can you conform?), but it'd be nice if we could rebuild at least some of the stress testing from the original paper, and that may well be an option for the future (would certainly make me much more comfortable with changes).
> 
> The proof problems for the loop are real in the sense that if there are overflows we do know that the algorithm does not work for those cases, i.e. a malicious user-level component can configure values that break invariants the rest of the code relies on, which may lead to crashes in random other places (not the loop). So we need to do something. One option would be for the code to check for those overflow cases we can't prove and do something sensible when it happens. This may still screw with the behaviour of the objects in that operation in those corner cases, but hopefully in a slightly better way. It would also not change the behaviour of any existing systems that don't hit those cases, i.e. makes the situation slightly better overall even if it doesn't solve it.
> 
> This would then allow us to defer an actual solution to later.

Can we bound the overflow? We know that the kernel timer mostly works to some degree of accuracy, so presumably threads cannot overrun by more than budget + accuracy.

src/object/notification.c
xurtis: But endpoints aren't bound to threads, notification s are. Are you suggesting that we add bound endpoints to threads?

include/object/schedcontext.h
xurtis: Currently, sending on an endpoint doesn't donate any SC. It would make sense for a send on a notification to return the SC in some (but not all) cases.

include/kernel/sporadic.h
pingerino: I'm not convinced that MIN_BUDGET should be the minimum for all refills, especially the tail. I think this will end up forcing threads to eat their budget too fast. Sure, the kernel doesn't provide a minimum bandwidth to threads, but going through the bandwidth too quickly reduces the utility.

src/kernel/sporadic.c
xurtis: I suspect a better place to start would be on the invariants and the model used in verification. From all the discussion that has occurred there seems to have been a misunderstanding about specifically what properties we want to guarantee and many of the changes were put in place to ensure properties that it seems we don't actually want.

src/kernel/sporadic.c
xurtis: `refill_add_tail` doesn't do any checking, it just adds the tail. `schedule_used` adds a tail but does some checking and moving of refills. It seemingly serves to satisfy some properties that I don't think I could find in the original paper.

pingerino: Right. In that case it was written to reduce branches and avoid checks we know we don't need.

xurtis:
I think I suggested at some point that the overflow would never be more than the maximum budget (which is a week and thus _very_ generous), at the time that looked like it wouldn't loose as more than a few bits of total system runtime (62-63 bits of runtime rather than 64).

src/kernel/sporadic.c
pingerino: If this isn't needed it would be better to avoid merging this one.

kent-mcleod:
Are we able to prioritize resolving the *all refills are at least MIN_BUDGET* change first? A lot of the changes in the PR are due to trying to enforce that and if we decide that we need to allow the tail to be < MIN_BUDGET then a lot of the changes go away?  Then we can resolve the issue with the overflow while loop?

xurtis:
> One option would be for the code to check for those overflow cases we can't prove and do something sensible when it happens.

A check in `refill_budget_check` along the lines of `SYSTEM_MAX_TIME - REFILL_HEAD(sc).rTime > sc->scPeriod  + REFILL_HEAD(sc).rAmount` might be enough to catch that case, wherein we can just place the refill at the latest point in time.

xurtis:
> Are we able to prioritize resolving the all refills are at least MIN_BUDGET change first? A lot of the changes in the PR are due to trying to enforce that and if we decide that we need to allow the tail to be < MIN_BUDGET then a lot of the changes go away?

I think that of the invariants we have, two need to change:

 * Rather than all refills having `MIN_BUDGET`, only the head refill needs to have `MIN_BUDGET`
 * We currently ensure that no refills overlap, but we likely only need to ensure that the head refill doesn't overlap with any refills (and maybe not even that)

If those changes do need to be made then most of (if not all of) the changes to the refill logic are unnecessary.

kent-mcleod:
Is there an example in the current implementation where refills end up overlapping? I can see how requiring a minimum refill size can lead to refills overlapping if subtracting more budget than what has been consumed causes the total budget to be consumed faster than the time is passing, but I can't easily find another example.

lsf37:
For context, these are the current main invariants for refills (rr = round robin, sp = sporadic):

```
definition rr_valid_refills :: "refill list ⇒ nat ⇒ ticks ⇒ bool" where
  "rr_valid_refills refills refill_max budget ≡
      (refills_sum refills = budget
      ∧ ordered_disjoint refills
      ∧ no_overflow refills
      ∧ (MIN_BUDGET ≤ r_amount (refills ! 0))
      ∧ (length refills = 2 ⟶ MIN_BUDGET ≤ r_amount (refills ! 1))
      ∧ window refills budget
      ∧ 0 < length refills
      ∧ length refills ≤ refill_max
      ∧ MIN_SC_BUDGET ≤ budget
      ∧ r_time (hd refills) = 0
      ∧ refill_max = MIN_REFILLS
      ∧ budget ≤ MAX_SC_PERIOD)"

definition sp_valid_refills :: "refill list ⇒ nat ⇒ ticks ⇒ ticks ⇒ bool" where
  "sp_valid_refills refills refill_max period budget ≡
      (refills_sum refills = budget
      ∧ ordered_disjoint refills
      ∧ no_overflow refills
      ∧ (∀refill ∈ set refills. MIN_BUDGET ≤ r_amount refill)
      ∧ window refills period
      ∧ 0 < length refills
      ∧ length refills ≤ refill_max
      ∧ MIN_SC_BUDGET ≤ budget
      ∧ budget ≤ period
      ∧ MIN_REFILLS ≤ refill_max
      ∧ period ≤ MAX_SC_PERIOD)"
```

lsf37:
So you're saying we don't need the `disjoint` part of the `ordered_disjoint` and `(∀refill ∈ set refills. MIN_BUDGET ≤ r_amount refill)`.

@mktnk3 @MitchellBuckley do you know why we have added these invariants?

xurtis:
> Is there an example in the current implementation where refills end up overlapping?

I believe the current `master` implementation of `refill_budget_check` can produce an overlapping refill when the head refill is split and the remainder merged with the subsequent refill (if the two refills after the head refill are less then the remainder apart then they end up overlapping).

src/object/schedcontrol.c
kent-mcleod: After more discussion internally, we believe that verifying the check in microseconds is unlikely to be an issue.  So this change isn't required for now.

kent-mcleod:
The work in this PR was merged in PRs: https://github.com/seL4/seL4/pull/218 and https://github.com/seL4/seL4/pull/221.

PR #222, opened 2020-07-25
riscv: tweak section alignment

sorear:
Fixes the `size(1)` issue described in seL4/docs#59.  Tested on `-DPLATFORM=spike` simulation.

PR #224, opened 2020-07-25
riscv gp fixes

sorear:
Tested via `-DPLATFORM=spike` build and `./simulate`.  (saves 386 of the 14683 instructions in a master release build)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,7 +164,15 @@ if(KernelArchARM)
     endif()
 endif()
 if(KernelArchRiscV)
-    KernelCommonFlags(-mcmodel=medany)
+    # Group "small" data objects together in a small-data section so they can
+    # be referenced using gp-relative addressing.  The exact value of
+    # small-data-limit is not crucial but it should be lowered if .small
+    # exceeds 4KiB.
+    #
+    # Without -fno-common uninitialized variables are generated as *COM*
+    # symbols with no section information.  -fno-common is the default on
+    # gcc 10+ (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678).
+    KernelCommonFlags(-mcmodel=medany -fno-common -msmall-data-limit=1024)
 endif()
 KernelCommonFlags(-fno-pic -fno-pie)
 add_compile_options(
--- a/src/arch/riscv/common_riscv.lds
+++ b/src/arch/riscv/common_riscv.lds
@@ -43,12 +43,13 @@ SECTIONS
 
     /* Start of data section */
     _sdata = .;
-    .sdata : {
+    .small : {
+        /* Small data that should be accessed relative to gp.  ld has trouble
+           with the relaxation if they are not in a single section. */
         __global_pointer$ = . + 0x800;
-        *(.sdata*)
-    }
-    .srodata : {
         *(.srodata*)
+        *(.sdata*)
+        *(.sbss)
     }
 
     .rodata . : AT(ADDR(.rodata) - KERNEL_OFFSET)
@@ -65,7 +66,6 @@ SECTIONS
     .bss . : AT(ADDR(.bss) - KERNEL_OFFSET)
     {
         *(.bss)
-        *(.sbss)
 
         /* 4k breakpoint stack */
         _breakpoint_stack_bottom = .;
--- a/src/arch/riscv/traps.S
+++ b/src/arch/riscv/traps.S
@@ -79,7 +79,10 @@ trap_entry:
   csrr s0, scause
   STORE s0, (31*REGBYTES)(t0)
 
+.option push
+.option norelax
   la gp, __global_pointer$
+.option pop
 
 #ifdef ENABLE_SMP_SUPPORT
   /* save the user sp */


CMakeLists.txt
```diff
@@ -164,7 +164,7 @@ if(KernelArchARM)
     endif()
 endif()
 if(KernelArchRiscV)
-    KernelCommonFlags(-mcmodel=medany)
+    KernelCommonFlags(-mcmodel=medany -msmall-data-limit=64 -fno-common)
```
yyshen, line 167: -fno-common should be the default setting for gcc?

sorear: Unclear if that's the upsteam default, but if I don't have that I get a large number of `*COM*` symbols in `kernel/CMakeFiles/kernel.elf.dir/kernel_all.c.obj` and they ultimately get allocated in `.bss`, when using the `sel4-riscv` image

yyshen:
@ssrg-bamboo `test`

ssrg-bamboo:

CMakeLists.txt
yyshen: Thanks for the info.

ssrg-bamboo:

CMakeLists.txt
kent-mcleod: Are you able to add a comment explaining how without `-fno-common` many symbols that would otherwise end up in the small sections (.sbss, .srodata and .sdata) get left in the default sections (.bss, .rodata and .data) so that it's clear that both options are dependent on each other?

kent-mcleod, line 167: Doing some of my own experiments, I found that the small-data-limit could be increased to 1024 before the small section exceeded 0x1000 in size and the final instruction count ended up at 14640 down from 15062.  What are your thoughts about this?

src/arch/riscv/common_riscv.lds
```diff
@@ -43,12 +43,11 @@ SECTIONS
...
     /* Start of data section */
     _sdata = .;
-    .sdata : {
-        __global_pointer$ = . + 0x800;
-        *(.sdata*)
-    }
-    .srodata : {
+    .small : {
         *(.srodata*)
+        __global_pointer$ = .;
```
kent-mcleod, line 48: My understanding is that if we wanted to maximize the number of symbols that can have their addressing relaxed then the `__global_pointer` symbol should be placed where the most symbols are +/- 0x800 from its definition.  Looking at `kernel_all.c.obj`, .srodata is only 0x210 in size, so it seems that something like the following gets better coverage?
```
    .small :  AT(ADDR(.small) - KERNEL_OFFSET)
    { 
        __global_pointer$ = . + 0x800;
        *(.srodata*)
        *(.sdata*)
        *(.sbss)
    }
```

sorear:
Updated to address comments.

Re. measurement: I have been using `objdump -d FILE | grep $':\t' | wc -l`.  This fails to properly handle long wrapped instructions on x86, and also counts three spurious instructions because .boot.rodata is being merged into a code section (see #222).

branden-data61:
@kent-mcleod: Any thoughts on the bump sorear did?

Also, to anyone interested: is an upstream binutils bug report in order?  It's maintained by the former Cygnus team, now part of Red Hat (now part of IBM).  Their tracker is at https://sourceware.org/bugzilla/ .

sorear:
Which part of this are we considering a binutils bug?  The generation of COM symbols is a gcc thing and AFAIK it's a compatiblity feature for pre-ANSI C.

branden-data61:
@sorear: My mistake regarding the object file. :)

Could you rebase with master, please?

axel-h:
@sorear: could you rebase this and resolve the conflict, so we can merge this eventually. Thanks.

CMakeLists.txt
```diff
@@ -164,7 +164,15 @@ if(KernelArchARM)
...
-    KernelCommonFlags(-mcmodel=medany)
+    # Group "small" data objects together in a small-data section so they can
+    # be referenced using gp-relative addressing.  The exact value of
+    # small-data-limit is not crucial but it should be lowered if .small
+    # exceeds 4KiB.
+    #
+    # Without -fno-common uninitialized variables are generated as *COM*
+    # symbols with no section information.  -fno-common is the default on
+    # gcc 10+ (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678).
+    KernelCommonFlags(-mcmodel=medany -fno-common -msmall-data-limit=1024)
```
axel-h, line 175: We have `-fno-common` in the `add_compile_options()` list now also. Should we remove it there and the have this unconditionally now:
```
KernelCommonFlags(-fno-pic -fno-pie -fno-common)
```

axel-h:
I've done some cleanup and then rebased your change at https://github.com/axel-h/seL4/commits/riscv_gp_fixes

src/arch/riscv/common_riscv.lds
```diff
@@ -43,12 +43,13 @@ SECTIONS
...
     _sdata = .;
-    .sdata : {
+    .small : {
+        /* Small data that should be accessed relative to gp.  ld has trouble
+           with the relaxation if they are not in a single section. */
         __global_pointer$ = . + 0x800;
-        *(.sdata*)
-    }
-    .srodata : {
         *(.srodata*)
```
axel-h, line 50: Is it intended that `.srodata {...}` is dropped and `*(.srodata*)` is part of `.small` now?

kent-mcleod:
Updated PR in https://github.com/seL4/seL4/pull/850

PR #225, opened 2020-07-26
riscv-motivated general bitfield optimizations

sorear:
- [x] Replace the current idiom for sign extending a bitfield with a new one that gcc and clang can optimize to an arithmetic shift.  It's rather long and I pulled it out to an inline function, which means it appears in the libsel4 API (namespaced).  Saves about 9.8% of static instructions and 10.4% of text on riscv64 by itself.
- [x] Elide masks in `_new` functions when they are redundant with the range check.  Very marginal by itself, mostly a prerequisite for
- [x] Add a `widen` flag to the bitfield generator which sign or zero extends a field into adjacent padding, and assumes that the padding is always set that way.  Most useful for `field_high`s, especially ones like mdbNext which can now be accessed just by loading a word and masking it with -4.  Saves 1.4% riscv64 text and instructions _right now_, I believe it is limited because:
- [ ] Move `capType` to the bottom of the first word on 64-bit systems.  This allows most pointers in capabilities to become widens.  Savings TBD.
- [ ] Fix all fastpaths after the above.
- [ ] Write documentation for the bitfield generator, which is pretty opaque especially if you don't have a build environment set up yet and can't look at the output.
- [ ] Test all other architectures.
- [ ] Evaluate performance of the changes, individually and combined, on other architectures.  (I expect this to benefit x86 and be close to a wash on Arm which has bitfield instructions, but I don't know this yet)
- [ ] Try to adjust proofs


mbrcknl:
I'm not the best person to comment on the systems programming aspects of this, but it is certainly exciting to see that *someone* is interested in it! :-)

If you need help with the "Try to adjust proofs" item, please reach out to me, @Xaphiosis or @lsf37. We're quite busy this year, so I can't make too many promises, but we'll do our best to help.

lsf37:
This looks like pretty cool stuff, nice to see someone getting into the bitfield generator again. My main concern is the proof impact of the new changes. If it is contained within the generator itself, then it should be doable, but if we haven't fully kept to the bitfield API and instead looked under the hood into the implementation in a few places, then it might get painful. I don't particularly remember any such spots (apart from zero-ing), but it's possible. The only real way to find out is to try.

In terms of performance impact, the main thing to check are how the fastpath benchmarks are affected (time may go up even if length/size goes down, I dimly remember some of the bitfield layout arrangements being made specifically for fastpath access. @kent-mcleod might remember).

In any case, if you'd like input or help, as Matt said, please do ping us.  It'd be very nice to squeeze out a little more performance with this.

sorear:
This has gotten longer than I anticipated and I'll probably break it up into several PRs when "done", but keeping everything together for now makes it easier to evaluate interactions.

re. proof impacts: I agree.  There is a natural split here between:

1. local changes to the bitfield generator that do not change functional behavior (even when passed invalid blocks) and do not change the in-memory representation of blocks
2. reordering bitfields will have impacts on places where blocks are modified without using the accessors.  There are quite a few of these in the fastpaths, but it's hard to get a complete list of them; there are a lot of functions that use "words" when accessing hardware-defined blocks, but those were never candidates for reordering in the first place.
3. widening bitfields doesn't change the in-memory representation of validly constructed blocks or affect fastpath code, but does weaken the postconditions of accessors used on blocks constructed in userspace or by hardware.  The hardware blocks are mostly obvious, I'm less confident that there are no userspace-shared blocks other than the ones in shared_types.bf.

If getting all three in is too onerous to prove, just getting the first would be a good win.

What would people think of including the name of the block or union in the name of the words field, so that you can search the code for e.g. `cap_words` to find anywhere the cap abstraction is broken?  Using a better IDE would help but I'd still have to separately check a bunch of different build configurations.

@lsf37 What do you mean by "zero-ing"?  Implicit initialization of kernel objects by Retype?

re. benchmarks: I don't currently have one of every device on the seL4 benchmarks, so I'm not sure what the best way to handle that is.

re. workload: That's fine, unfortunately I'm currently blocking on a hardware upgrade so I'm not ready.

lsf37:
> @lsf37 What do you mean by "zero-ing"? Implicit initialization of kernel objects by Retype?

Yes, `resetUntypedCap` uses `memzero` to clear memory and so that Retype can assume it's all zeroes, which then has the corresponding meaning for the bitfields. This shouldn't really cause any problems.

lsf37:
> What would people think of including the name of the block or union in the name of the words field, so that you can search the code for e.g. cap_words to find anywhere the cap abstraction is broken? Using a better IDE would help but I'd still have to separately check a bunch of different build configurations.

The string `.words[` in `.c` and `.h` files should be a good indicator of where the abstraction is broken in general. I'd hope that there are not too many of them, or at least not where it's relevant.

lsf37:
> re. benchmarks: I don't currently have one of every device on the seL4 benchmarks, so I'm not sure what the best way to handle that is.

@nomadeel would you be able to help with that?

sorear:
I've cleaned up and given meaningful commit messages for everything up through "Minimize masking in generators".

> Yes, resetUntypedCap uses memzero to clear memory so that Retype can assume it's all zeroes, which then has the corresponding meaning for the bitfields. This shouldn't really cause any problems.

Indeed, none of the changes I've considered changes the all-zero representation.

> The string .words[ in .c and .h files should be a good indicator of where the abstraction is broken in general. I'd hope that there are not too many of them, or at least not where it's relevant.

There are 267 references to `.words` or `->words` in `src`, `include`, and `libsel4` in `.c` and `.h` files, including https://github.com/seL4/seL4/blob/b3ed77311fe89af5f1fbbebdfc28edcb6d5160ac/src/object/objecttype.c#L417 which breaks the abstraction without a bracket.

lsf37:
It looks like there is a bunch of work here that would still be useful, even though verification impact could be fairly high. @sorear are you planning on working more on this?

If not I would propose to close this PR, but extract the sign-extend change and see if that can be verified in isolation.

PR #229, opened 2020-07-31
riscv/smp: Add missing include

sorear:
With this and seL4/seL4_tools#41, I can build riscv32 and riscv64 `-DSMP=ON -DPLATFORM=spike -DSIMULATION=ON` and run the subset of non-timer tests with `./simulate -M virt -b 'qemu-system-riscvNN -smp 4'`, although there are a few "Spurious interrupt" messages.

### DIFF
--- a/include/arch/riscv/arch/32/mode/machine.h
+++ b/include/arch/riscv/arch/32/mode/machine.h
@@ -8,3 +8,4 @@
 
 /* Place holder for 32-bit machine header */
 
+#include <arch/model/smp.h>
--- a/include/arch/riscv/arch/64/mode/machine.h
+++ b/include/arch/riscv/arch/64/mode/machine.h
@@ -8,3 +8,4 @@
 
 /* Place holder for 64-bit machine header */
 
+#include <arch/model/smp.h>


kent-mcleod:
@ssrg-bamboo `test`, Thanks!

ssrg-bamboo:

ssrg-bamboo:

xurtis:
@sorear can you rebase this PR against `master`?

sorear:
(I am somewhat confused by the workflow here.  Am I expected to be doing anything other than just rebasing?  Does it need to be rebased just to run the CI bot, or for other reasons?  Are we trying to do something like rust-lang's [homu](https://bors.tech/homu-io/)?)

xurtis:
The process is a little complicated at the moment. We have an internal bitbucket repository on which a much more extensive set of CI tests are run including tests on several hundred configurations on actual hardware. When we merge into `master` on the internal bitbucket all of those tests are run and, if they all pass, all changed repositories get pushed back out to GitHub.

If a new push comes out to github with commits matching those in open PRs, those PRs then get marked as merged. If we rebase internally the commits change and the GitHub doesn't pick up on the fact that the changes have been included in master.

It's a rough process at the moment but we are investigating ways of simplifying and de-duplicating where possible but doing so is a large amount of work that involves a lot of specialised components.

xurtis:
>  Are we trying to do something like rust-lang's homu

Yes, but the step that merges into master is manual and there are additional duplication steps.

sorear:
I see.  Will be nice if that can be automated, somehow. (It appears to work for Rust because they use merge commits.  The "Allow edits from maintainers" feature might help?)

xurtis:
Most of the complexity currently comes from the duplication of the repositories. If we could the fast-forward and merge on the github side I suspect that would be less of an issue.

lsf37:
> Most of the complexity currently comes from the duplication of the repositories. If we could the fast-forward and merge on the github side I suspect that would be less of an issue.

Yes, it's quite annoying for our workflow that github doesn't have a ff-only merge option. Rebase and ff-only merge could probably be supported a bit better with a github app if somebody is keen to write one. When I looked last, I saw one somewhere that used to do that, but it's now unmaintained.

sorear:
What's a "ff-only merge option"?

Would this conversation make sense anywhere other than the comments of an unrelated issue?

lsf37:
> What's a "ff-only merge option"?

fast-forward merge only, i.e. a merge that doesn't need a merge commit, so that you get a linear history supporting a rebase workflow. Github does have a rebase-on-merge button, but that one destroys the commit hashes so it's hard to sync between systems. 
 
> Would this conversation make sense anywhere other than the comments of an unrelated issue?

Yes, it's getting a bit off topic. We're aware of the issue and working on it, so we probably don't really need to open another issue for it here, unless you think there is need for discussion. Longer term, I think we'll want to move the whole development to just github without the need to cycle between two systems. There's just a lot of infrastructure to move with it, so we have this somewhat annoying incremental in-between thing for now.

xurtis:
@branden-data61 this is ready to be enqueued and merged on our end

xurtis:
@sorear you'll need to remove that merge commit that I added when @branden-data61 gets you to rebase again.

xurtis:
@ssrg-bamboo 'test'

xurtis:
@ssrg-bamboo `test`

ssrg-bamboo:

PR #236, opened 2020-08-11
CI: Add a PR check to do a stand-alone kernel compilation 

LukeMondy:
On a PR, compile all verified configs of seL4.

I put it in its own yml file, so we can better sort through the results, using the GitHub GUI.

### DIFF
--- a/.github/workflows/compilation-checks.yml
+++ b/.github/workflows/compilation-checks.yml
@@ -1,0 +1,31 @@
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Compilation actions to run on pull requests
+
+name: Compilation checks
+
+on: [pull_request]
+
+jobs:
+  standalone_kernel:
+    name: Compile kernel
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ARM, ARM_HYP, RISCV64, X64]
+        compiler: [gcc, llvm]
+        python: [py2, py3]
+        exclude:
+          # llvm RISCV64 compilation is not currently supported
+          - arch: RISCV64
+            compiler: llvm
+    steps:
+    - uses: actions/checkout@v2
+    - uses: seL4/ci-actions/standalone-kernel@master
+      with:
+        ARCH: ${{ matrix.arch }}
+        COMPILER: ${{ matrix.compiler }}
+        PYTHON: ${{ matrix.python }}
--- a/.github/workflows/compilation-checks.yml
+++ b/.github/workflows/compilation-checks.yml
@@ -6,7 +6,11 @@
 
 name: Compilation checks
 
-on: [pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 
 jobs:
   standalone_kernel:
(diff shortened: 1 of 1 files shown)


.github/workflows/compilation-checks.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Compilation actions to run on pull requests
+
+name: Compilation checks
+
+on: [pull_request]
```
lsf37, line 9:
```suggestion
on:
  push:
    branches:
      - master
  pull_request:
```

I'd run these on push to master as well. Theoretically it should already be implied that this always works on master if we check on every PR, but it's nice to be able to look at just the branch and directly see the CI status.

.github/workflows/compilation-checks.yml
```diff
@@ -0,0 +1,31 @@
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Compilation actions to run on pull requests
+
+name: Compilation checks
```
lsf37, line 7: Can we find a shorter name? Maybe just "Compile". The shorter the name, the easier it is to see what's happening in the "Checks" section of a PR, esp with matrix builds that tend to generate long combination names.

LukeMondy: Yep, `Compile` sounds good.

Looking to the future though, the number of checks for `sel4test` would be >>50. So perhaps we should isolate them, meaning this one would be `Kernel compile`, and the future seL4test one would be `seL4test compile`.

lsf37: We could also call the workflow `Kernel`, and the job `compile`. Then it should come out as something like `Kernel / compile (ARM, llvm, py3)`.

lsf37: But happy for you to decide either way.

lsf37:
Hm. The DCO check complains about commits made via the `suggestion` interface. Happy to for you to just set it to `pass` for this one (trivial change).

LukeMondy:
I'm not actually logged on to my work VM, so I can't rebase it now (or make the name changes).
Happy for you to do it now, or I'll do it tomorrow (so it passes the DCO, and with the suggested name changes)

lsf37:
I'll give it a go

.github/workflows/compilation-checks.yml
```diff
@@ -0,0 +1,35 @@
...
```
lsf37, line 7:
```suggestion
name: Kernel
```

.github/workflows/compilation-checks.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  standalone_kernel:
+    name: Compile kernel
```
lsf37, line 17:
```suggestion
    name: compile
```

lsf37:
Squashed the commits -- I hope this worked ;)

lsf37:
So this didn't work quite as nicely as I expected. The squash showed me as the commit author (I thought), hence I added Luke as co-authored-by, but it turns out Luke is there now twice. And the commit message is confusing, because I missed the subject line of the interface. I should do a proper manual rebase/squash next time -- sorry about the borked commit.

PR #237, opened 2020-08-11
riscv: Clear reservation state on slowpath exit

sorear:
For those wondering, this code was correct when it was written as trap-return instructions were originally specified to clear reservations.  riscv/riscv-isa-manual@03a5e722 broke the link between the privileged architecture and the memory model as part of RVWMO pre-ratification improvements.

### DIFF
--- a/src/arch/riscv/c_traps.c
+++ b/src/arch/riscv/c_traps.c
@@ -45,6 +45,9 @@ void VISIBLE NORETURN restore_user_context(void)
         LOAD_S "  gp, (2*%[REGSIZE])(t0)  \n"
         /* skip tp */
         /* skip x5/t0 */
+        /* no-op store conditional to clear monitor state */
+        /* this may succeed in implementations with very large reservations, but the saved ra is dead */
+        "sc.w zero, zero, (t0)\n"
         LOAD_S "  t2, (6*%[REGSIZE])(t0)  \n"
         LOAD_S "  s0, (7*%[REGSIZE])(t0)  \n"
         LOAD_S "  s1, (8*%[REGSIZE])(t0)  \n"


xurtis:
To clarify, the following change in behaviour is why the code you have introduced correctly clears any reservation (as now only one reservation may be outstanding in a hart), yes?

> In the standard A extension, an SC must fail if there is another SC (to any address) between the LR and the SC in program order.  In other words, multiple outstanding reservations are not permitted.

sorear:
That's the relevant clause but I'm not sure "only one reservation" is the best way to think about it.  Under the old rules you could succeed all SC if there was a LR more recent than the last xRET and fail all SC if there was a xRET more recent than the last LR.

xurtis:
At any rate, I'm happy that this change is necessary here (and not on the fastpath). This also looks like is should still be compatible with any implementation that hasn't incorporated those changes.

xurtis:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

xurtis:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

kent-mcleod:
@sorear are you able to explain what you mean by `This is not necessary on the fastpath for functional correctness since
    user threads do not perform ipc in the middle of lr/sc sequences.`? I think that this does need to be added to the fastpath as the verification has to prove that the fastpath is equivalent to some path through the normal slowpath and so should have a similar operation.  At the moment the riscv fastpath is exccluded by the proofs and so they haven't checked this.

kent-mcleod:
I added https://github.com/seL4/seL4/pull/276 as a candidate for re-syncing the fastpath with the slowpath due to my above comment.

PR #243, opened 2020-08-27
Don't clobber msginfo register after invocation

canarysnort01:
**PR Purpose**: Ready to merge.
**Testing performed**: Manual inspection of registers from userspace on x86_64.

The messageinfo register is set to 0 after processing a kernel object
invocation if the invocation set the thread to restart.

This is a problem if the kernel object invocation had also set the
message info register.

The existing convention in this case is for the kernel invocation to set
the thread state to running right before it returns to avoid the
problematic code path.

However, there are some invocations that do not follow this convention
and their message info register gets clobbered. The affected kernel
object invocations are:

  - ARMPageGetAddress
  - RISCVPageGetAddress
  - X86PageGetAddress
  - X86VCPUReadVMCS
  - SchedContextConsumed
  - SchedContextYield

This commit fixes these.

Signed-off-by: Jimmy Brush <code@jimmah.com>


mbrcknl:
Thanks for the contribution!

This looks to me like a good change, but it will need a proof update. That should be reasonably straightforward, but we'll need to fit it into the schedule.

kent-mcleod (approved):
Nice fix!

axel-h:
any plans to merge this?

lsf37:
I'll have a look at the proof update for this. The spec has the same behaviour and basically demands that the message is clobbered, which is not really what we want.

It'd be nice to express this a bit better in the spec, but there is no clear general formal way to say "this information is more wanted than that information", so the only way is to explicitly say per syscall what the user should see in the end.

lsf37:
(rebased and removed trailing whitespace)

lsf37:
Turns out this is a proper stuff-up in the spec. When the `PageGetAddress` call was added, the reply convention in the spec wasn't used, but it sets message registers manually, promptly getting things wrong. Well, wrong-ish. It is correctly modelling the wrong behaviour in the C code, but instead of doing that, the C should have been changed, of course.

Xaphiosis:
The Haskell code (`l4v/spec/haskell/src/SEL4/Syscall.lhs`) describes the intended protocol:

`While the system call is running, the thread's state is set to "Restart", so any preemption will cause the system call to restart at user level when the thread resumes. If it is still set to "Restart" when the operation completes, it is reset to "Running" so the thread resumes at the next instruction. Also, if this is a call, then a reply message is generated when the thread is restarted.`

lsf37:
> The Haskell code (`l4v/spec/haskell/src/SEL4/Syscall.lhs`) describes the intended protocol:
> 
> `While the system call is running, the thread's state is set to "Restart", so any preemption will cause the system call to restart at user level when the thread resumes. If it is still set to "Restart" when the operation completes, it is reset to "Running" so the thread resumes at the next instruction. Also, if this is a call, then a reply message is generated when the thread is restarted.`

We should add that comment to the C source, so that protocol is more obvious to people. Will see if I can find a good spot.

canarysnort01:
There are a couple more instances of this problem from a recent commit:

- ARMSIDGetFault
- ARMCBGetFault

In addition, I have found some more problematic code paths that I missed originally because in addition to not changing their thread state to running, they do not attempt to set the message info register at all:

- TCBGetBreakpoint
- TCBConfigureSingleStepping
- X86PageDirectoryGetStatusBits
- X86VCPUWriteVMCS

I've converted this to a draft issue while I work on fixing these.

lsf37:
Thanks for that. I think we should find a more central place for that comment, because the way it is now it looks like these operations are special in some way, when they were basically just not following the protocol all operations are supposed to follow.

It might be hard to find a good place for that, will have a look.

I still need to go back to the proof update for `PageGetAddress`. It is trivial (and makes things much nicer) in the specification levels, but a bit annoying on the C proof side, which is were I got interrupted.

So TODO for me:
- [ ] finish proof update
- [ ] find good place to comment return protocol (then potentially point to that)

kent-mcleod:
Closing as this PR is superseded by https://github.com/seL4/seL4/pull/846

Issue #270, opened 2020-10-21
zynqmp dts irqs


chrisguikema:
In the `ethernet` nodes of the zynqmp device tree, the `interrupts` block contains duplicate interrupts.

```interrupts = < 0x00 0x3b 0x04 0x00 0x3b 0x04 >;```

 This causes an issue when using the `seL4DTBHardwareThreadless` templates, as the `irq_set` value contains two interrupts, causing the following error:

```[323/373] Generating capdl_spec.c
FAILED: capdl_spec.c 
cd /host/build && /host/build/capDL-tool/parse-capDL --code capdl_spec.c --code-dynamic-alloc --object-sizes=/host/build/capdl/object_sizes/object_sizes.yaml /host/build/Ethdriver.cdl
parse-capDL: IRQ already mapped: 95
```

This can be fixed by removing the duplicate values in the interrupt blocks.

Would you like me to make a PR for this?

axel-h:
If this is simply a bug, then that sound like a reasonable thing. If there is more behind this duplicate entry, because the device tree comes from Linux and this is used for something, then understanding the details might be worth it.

chrisguikema:
Right. Looking at other places with a zynqmp device tree (u-boot, for example) they have two identical interrupts for the gem devices as well. Can you think of any reason why this would be the case?

chrisguikema:
Looking at the TRM, there are two interrupts associated with the GEM devices
![image](https://user-images.githubusercontent.com/21088707/96883079-f9becb00-144d-11eb-9179-18c41da2ac41.png)

Perhaps this is a case of reserving a spot for the wakeup interrupts?

axel-h:
Seems even in the git history of the official XILINX repos it has been there from the begin on, see https://github.com/Xilinx/device-tree-xlnx/commit/a8c0d0ce863129b029f3d5e6b30a8e45ff9d0cb3 and https://github.com/torvalds/linux/commit/5d1b79d2b2639bb71238e76d73431fada9e38553

I'm not really an expert on linux drivers, but I can do some guessing now. Looking at https://github.com/Xilinx/linux-xlnx/blob/master/drivers/net/ethernet/xilinx/xilinx_axienet_main.c, it uses set of function like platform_get_irq() and there is some code about RX and TX interrupt. Maybe that is the explanation, we see support for different interrupts for RX and TX here, but practically they use the same interrupt.

You guys don't happen to have a contact at Xilinx that could help here?

[...]

PR #271, opened 2020-10-22
zynqmp: update ethernet interrupts in overlay

chrisguikema:
The Linux drivers use duplicate interrupts in the device tree; however,
seL4 drivers do not. The duplicate interrupts cause issues while using
templates in CAmkES applications. This commit adds an overlay which
overwrites the duplicate ethernet interrupts.

### DIFF
--- a/src/plat/zynqmp/overlay-zynqmp.dts
+++ b/src/plat/zynqmp/overlay-zynqmp.dts
@@ -5,6 +5,27 @@
  */
 
 / {
+	amba {
+		/* These blocks remove the duplicate interrupt which causes issues
+		 * in the low-level drivers
+		 */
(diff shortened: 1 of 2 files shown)


axel-h:
Please add the sign-off message to the commit

kent-mcleod (approved):
I think longer term we should update the tool to handle these edge cases better.

chrisguikema:
I agree. But this is a quick fix until someone has time to look into all of the edge cases.

axel-h:
And if you look at the discussion of https://github.com/seL4/seL4/issues/270, this might not even be a bug, but it's also a question which device trees we want to use and if we are willing to accept the implications everywhere then also.

PR #281, opened 2020-11-11
use include guards for auto generated files

axel-h:
This avoids warnings that the files are included multiple times. Currently such warnings are shown for:
- camkes-tool/libsel4camkes/gen_config/sel4camkes/gen_config.h
- kernel/gen_config/kernel/gen_config.h
- libsel4/gen_config/sel4/gen_config.h

### DIFF
--- a/configs/seL4Config.cmake
+++ b/configs/seL4Config.cmake
@@ -14,8 +14,9 @@ set(bf_declarations "")
 
 include(${CMAKE_CURRENT_LIST_DIR}/../tools/internal.cmake)
 include(${CMAKE_CURRENT_LIST_DIR}/../tools/helpers.cmake)
-# Create and set all of the Kernel config options that can
-# be derived from the seL4 arch which is one of the following:
+
+# Create and set all of the Kernel config options that can be derived from the
+# seL4 arch which is one of the following:
 # aarch32, aarch64, arm_hyp, riscv32, riscv64, x86_64, ia32
 # This macro is intended to be called from within a platform config.
 macro(declare_seL4_arch sel4_arch)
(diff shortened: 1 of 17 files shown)


kent-mcleod:
What compiler option is causing these warnings? These files should be safe to include multiple times as a preprocessor definition is allowed to be redefined to the same value (https://gcc.gnu.org/onlinedocs/cpp/Undefining-and-Redefining-Macros.html)

axel-h:
We are basically building with all warnings enabled, so it may also issue warning about such things that are just "not nice", but still legal. I wonder, are you against using these include guards, because they may hide some potential errors? Otherwise I think it's good practice to have these.

kent-mcleod:
> I wonder, are you against using these include guards, because they may hide some potential errors

I'm curious about which warning flag causes these errors as I thought that redefining macros were "safe" so would be interested in reading the documentation for the flags.

I'm happy with adding the include guards.  I can't think of any situation where they would hide potential errors.  In the recent case of `CONFIG_*` variables being redefined, using `#pragma once` wouldn't have hidden it.  Maybe if we started adding conditional defines then it could become an issue, but as far as I know, these files are all only lists of `#define` statements.

kent-mcleod (approved):

oliver-wm (approved):

oliver-wm:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

bennoleslie (approved):

axel-h:
can we merge this? Otherwise, what is blocking this? The automated check fail, but this seem unrelated to me.

axel-h:
Could you merge this, please?

oliver-wm:
Sorry! Will merge tomorrow!

axel-h:
Note that this branch is sitting on top of https://github.com/seL4/seL4/pull/314

oliver-wm:
@ssrg-bamboo `test`

ssrg-bamboo:

PR #298, opened 2021-02-07
Add kernel log for debug tracing

xurtis:
This extends the existing benchmarking log and logging API for debugging purposes. It provides a general set of tools for describing and logging kernel events of varying size to a memory frame visible at user-level.

The intent for this PR is to enable new logging events to be added to trace kernel behavior in debugging situations in a manner that allows multiple different kinds of events to be traced together and to make it easy to add log events to the kernel source while debugging it without having to fully determine a new messaging format each time.

Rather than adding a new log buffer and a new set of syscalls, this PR adds a kernel configuration option that allows the benchmark log buffer to be enabled and configured in non-benchmark debug configurations and a messaging format with variable-length messages to the logs.

The nature of the messaging format prohibits existing messaging formats in the log buffer from being used at the same time, so this PR also implements a set of entry tracking messages that can be used along with any other added messages in the newer message format.

Much like the existing log message formats, the new format with messages of varying size only ever requires the kernel to write to the log region itself and always writes in a linear fashion through the log region so should demonstrate similar minimal use of the cache, although the number of cache lines accessed for each message write may be greater due to the lack of cache-line alignment on messages.

The impact of the new message format on system performance has not been checked and so logging in performance-sensitive cases should continue to use the existing benchmark logging APIs. It may well be that this new log format has sufficiently low overhead that it replaces the existing log buffer message formats, although removing those formats should be considered in a separate PR.

This change does not resolve the existing issues of the kernel log buffer for SMP configurations whereby all cores access the same log buffer region and the same global internal kernel state for tracking the kernel's write position in the buffer. The easiest solution for that problem would be to maintain a distinct cursor for each core and to divide the region into a set of smaller per-core logs. It would then require user-level to reset each core's log separately by some change to the syscall API. This fix is out of scope of this PR but could be implemented and proposed in a separate PR.

A complimenting set of tools will be added to seL4_utils for managing and exporting the log-buffer at user-level for offline analysis. seL4/seL4_libs#25 demonstrates how this would be used at user level.

### DIFF
--- a/include/kernel/traps.h
+++ b/include/kernel/traps.h
@@ -10,6 +10,7 @@
 #include <util.h>
 #include <arch/kernel/traps.h>
 #include <smp/lock.h>
+#include <log.h>
 
 /* This C function should be the first thing called from C after entry from
  * assembly. It provides a single place to do any entry work that is not
@@ -20,6 +21,8 @@ static inline void c_entry_hook(void)
 #if defined(CONFIG_BENCHMARK_TRACK_KERNEL_ENTRIES) || defined(CONFIG_BENCHMARK_TRACK_UTILISATION)
     ksEnter = timestamp();
 #endif
+
+    debugLog(Entry);
 }
 
 /* This C function should be the last thing called from C before exiting
@@ -40,6 +43,8 @@ static inline void c_exit_hook(void)
         NODE_STATE(benchmark_kernel_time) += exit - ksEnter;
     }
 #endif /* CONFIG_BENCHMARK_TRACK_UTILISATION */
+
+    debugLog(Exit);
 
     arch_c_exit_hook();
 }
--- a/libsel4/include/sel4/log.h
+++ b/libsel4/include/sel4/log.h
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <sel4/simple_types.h>
+#include <sel4/constants.h>
+#include <sel4/sel4_arch/constants.h>
+#include <sel4/benchmark_track_types.h>
+
+/*
+ * Log event implementation
+ * ========================
+ *
+ * Each event is a sequence of words in the log buffer. The first word
+ * of an event is the seL4_LogEvent which describes the length in words
+ * of the event and the event type. There is also an additional
+ * component to the seL4_LogEvent with use dependend on the event type.
+ *
+ * For an event of type t, there will be an event identifier
+ * seL4_Log_TypeId(t), and a struct representing the event
+ * seL4_Log_Type(t).
+ *
+ * The event can be logged from within the kernel using
+ * seL4_Log(t)(event_args...). seL4_Log expressions are automatically
+ * removed if logging is disabled.
+ */
+
+/* The event header word which appears at the start of every event */
+typedef struct {
+    /* Type of event logged */
+    seL4_Word type: 8;
+    /* The remaining 24 or 56 bits of the header word can be used for
+     * the event itself */
+#if seL4_WordSizeBits == 2
+    seL4_Word data: 24;
+#elif seL4_WordSizeBits == 3
+    seL4_Word data: 56;
+#else
+#error "Unsupported word seL4_WordSizeBits"
+#endif
+} seL4_LogEvent;
+
+/* The name of the type identifier for a given log event type */
+#define seL4_Log_TypeId(event) seL4_Log_TypeId_ ## event
+
+/* The name of the stuct type for a given log event type */
+#define seL4_Log_Type(event) seL4_Log_Struct_ ## event
+
+/* Cast a log event to a particular log type */
+#define seL4_Log_Cast(event) (seL4_Log_Type(event) *)
+
+/* Get the length of an event structure as a number of words */
+#define seL4_Log_Length(event) (((sizeof(seL4_Log_Type(event)) - 1ul) >> seL4_WordSizeBits) + 1ul)
+
+/*
+ * Definition of log events
+ * ========================
+ *
+ * Each log event must have a type identifier in the enum below, a
+ * struct describing the layout of the message (with the first field
+ * being the seL4_LogEvent).
+ *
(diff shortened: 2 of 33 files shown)


src/api/syscall.c
```diff
@@ -188,20 +188,8 @@ exception_t handleUnknownSyscall(word_t w)
...
-            arch_clean_invalidate_L1_caches(getRegister(thread, msgInfoRegister));
-        } else {
-            arch_clean_invalidate_caches();
-        }
-#else
-        arch_clean_invalidate_caches();
-#endif
-        return EXCEPTION_NONE;
-    } else if (w == SysBenchmarkResetLog) {
+#if defined(CONFIG_ENABLE_BENCHMARKS) || defined(CONFIG_KERNEL_LOG_BUFFER)
```
pingerino, line 192: This looks incorrect to me. In config.cmake above, the constant is `KERNEL_DEBUG_LOG_BUFFER` and here it's `KERNEL_LOG_BUFFER`. IMO the latter is better if you have to choose, it's not just for debugging.

pingerino (commented):
Its a bit unclear to me what is happening here. Are you effectively introducing a new mode for the kernel log buffer that is debugging? One that is incompatible with other modes of the log buffer?

src/api/syscall.c
pingerino: Or are these two different buffers? I'm confused

libsel4/include/sel4/log.h
```diff
@@ -0,0 +1,186 @@
...
+    seL4_Word data: 24;
+#elif seL4_WordSizeBits == 3
+    seL4_Word data: 56;
+#else
+#error "Unsupported word seL4_WordSizeBits"
+#endif
+} seL4_LogEvent;
+
+/* The name of the type identifier for a given log event type */
+#define seL4_Log_TypeId(event) seL4_Log_TypeId_ ## event
```
pingerino, line 48: I might be a  bit rusty but I think the style guide still expects macros to be all caps?

include/kernel/traps.h
```diff
@@ -20,6 +21,8 @@ static inline void c_entry_hook(void)
 #if defined(CONFIG_BENCHMARK_TRACK_KERNEL_ENTRIES) || defined(CONFIG_BENCHMARK_TRACK_UTILISATION)
     ksEnter = timestamp();
 #endif
+
+    debugLog(Entry);
```
xurtis, line 25: This approach of having `debugLog` be an empty `define` for builds not configured with the log still generates whitespace that triggers failure in the PR preprocess checks. I'm not sure whether it's the best approach.

src/api/syscall.c
xurtis: It should be the same buffer, I think you're correct and that's using the wrong define, also that `KERNEL_LOG_BUFFER` makes more sense given it's the same buffer either way.

xurtis: The only difference should be the message format in the buffer and the external structure used to track & iterate through messages inside and outside the kernel.

pingerino:
Can you please provide more context? AFAICT this PR does the following:

* Adds a new type of log buffer for use debugging, which is more flexible as it doesn't care about performance. 
* Duplicates (?) the kernel entry tracking with this new debug log buffer.

IMO any debug tracing *should* be fairly performance sensitive as it is likely going to be used a lot for debugging, and debugging a system with different timing and cache behaviour is not fun as behaviour can diverge too much from the non-debug system.

Also, how does this work for multicore?

libsel4/include/sel4/log.h
xurtis: > Use `SCREAMING_SNAKE_CASE` for, and only for, preprocessor symbols and values of enumeration type.

Yup it does. Should make this look a bit less magical too.

* [ ] `SCREAMING_SNAKE_CASE` for preprocessor symbols

xurtis:
I've updated the PR description above and added a complimentary user-level PR to seL4_Libs for further context. As for current use cases, I've been using this to get detailed tracing of scheduler decisions to get a more precise trace of the MCS scheduler behavior, I also intend to use it for tracing kernel behavior in scenarios where the existing logging features don't provide sufficient detail on scheduling behavior.

I didn't initially implement this with the intent of it being particularly performance sensitive, however I have revised it to hopefully be as minimal an overhead to performance as possible. The addition of the entry tracing here was simply because I wanted to still get that trace at the same time as also getting other logging events, and the existing trace logging could only be used if it was the only message format in the log buffer.

I can run some numbers and see what the overhead is; the current implementation may be similarly low overhead to the existing message formats and, as it uses the same mechanism for the log buffer itself, it also benefits from the mapping and cache configurations used for the existing benchmark log buffer where platforms support specifying write-through caching in the mapping (risc-v being the notable exception).

src/api/syscall.c
xurtis: * [ ] `KERNEL_LOG_BUFFER` for configuration definition

lsf37:
What do we do with this one? @xurtis I assume you're not going to be able to finish it.

The PR as such seems close, and could be completed, but I'm unclear on how needed the functionality is. If nobody is exercising this code, it'd be better to not merge and close the PR instead. It'd still be available for re-opening.

Any opinions?

axel-h:
Actually, I very interested in this feature and keep advertising this to colleagues. There are just not enough resources to work on this at the moment.

lsf37:
Ok, you be happy to be assigned to this and the PR being kept on draft until it is done?

axel-h:
@lsf37: yes.

axel-h:
I've rebased this at https://github.com/axel-h/seL4/tree/kernel-debug-log

src/api/syscall.c
axel-h: We should call the new feature `CONFIG_KERNEL_EVENT_TRACING` to avoid naming confusion.

Following the logic of other automatically generated config settings (`CONFIG_ENABLE_BENCHMARKS`, `CONFIG_ENABLE_SMP_SUPPORT`), the name `CONFIG_KERNEL_LOG_BUFFER` could become `CONFIG_ENABLE_KERNEL_LOG_BUFFER`. However, that would change a lot of places, so I'm not sure how much backing this gets.

lsf37:
@axel-h I'm close this PR so you can make a new one from your branch when the work is done.

PR #301, opened 2021-02-14
headers: avoid clashes with the std header names

axel-h:
Avoid using names that are used by the std headers. This could end up an an include quirk one day. It seems unlikely that this could happen, unless one does heavy cherry picking with a custom build system. But we don't know what people do with seL4, so  minimizing the "problem surface" seem good practice. 
Note that util.h is not a std header, but the name seems very generic also, so it becomes a bit more specific now


axel-h:
Seem the failing CI/Links check failure is unrelated:
```
Setting up
README.md
	ERROR	https://lists.sel4.systems/postorius/lists/announce.sel4.systems
		X509: certificate has expired or is not yet valid
	ERROR	https://lists.sel4.systems/postorius/lists/devel.sel4.systems
		X509: certificate has expired or is not yet valid
```
The certificate from lists.sel4.systems has exposed, it says "Expires On: Sunday, February 14, 2021 at 1:27:28 AM"

wom-bat:
> Seems the failing CI/Links check failure is unrelated:
> 
> ```
> Setting up
> README.md
> 	ERROR	https://lists.sel4.systems/postorius/lists/announce.sel4.systems
> 		X509: certificate has expired or is not yet valid
> 	ERROR	https://lists.sel4.systems/postorius/lists/devel.sel4.systems
> 		X509: certificate has expired or is not yet valid
> ```
> 
> The certificate from lists.sel4.systems has expired, it says "Expires On: Sunday, February 14, 2021 at 1:27:28 AM"

Fixed this morning.  I thought I'd set up auto-update on all the certs but missed one.

kent-mcleod:
A counter argument is that following consistent naming conventions of header files makes it easier for people to find where common definitions are located and also indicates what definitions are expected to be found in those header files.  I would prefer to not have to remember the way that these header files are named slightly differently.  If kernel files are compiled without the -nostd associated flags then I would already expect there to be more errors than accidentally pulling in incorrect headers.  I agree with your comment that it's good to be fault-tolerant because you can't imagine all the ways that something will be used, but I think if someone is taking the kernel's private header files into a different project then it's up to them to spend more effort avoiding any filename collisions?

axel-h:
Fair point. I've seen that in musllibc we also have multiple `stdint.h` files, one at the main include level and then several target specific ones. Seems the kernel just follows the same pattern. Once you get used to this, it's understandable and seems common practice. My personal preference would still be, that when searching for a include file, there should be as few hits as possible to there is less to chase.
But it turned out the issue I'm currently hunting is not cause bay any include quirk here, so I'll close this PR for now.

kent-mcleod:
Ok.  I actually thought you were talking about naming collisions in unrelated libraries such as `/usr/include/stdint.h`. For same-named headers in the same project, eg `#include<stdint.h>` and `#include<mode/stdint.h>` I would still consider these to be unique names due to the mode-specific namespace of the second header. Maybe having `#include <mode_stdint.h>` is better, but I think `#include <mode/mode_stdint.h>` has a redundant component. I think if we wanted to have unique names for each header file, then we should also remove the subdirectories at the same time and apply the rename across the entire project. As this would be a large change, there would need to be a stronger motivation.

PR #307, opened 2021-02-17
pc99: Add CONFIG_PC99_TSC_FREQUENCY config option

kent-mcleod:
This option allows a static definition of the TSC frequency. This allows
a kernel to use a known value if the platform doesn't provide access to
an appropriate MSR for reading it during startup. If this config is set,
the kernel won't try and estimate the TSC frequency by running the PIT
for 200ms during startup.

### DIFF
--- a/src/plat/pc99/machine/hardware.c
+++ b/src/plat/pc99/machine/hardware.c
@@ -76,6 +76,10 @@ BOOT_CODE static inline uint32_t measure_tsc_khz(void)
 BOOT_CODE uint32_t tsc_init(void)
 {
 
+#if CONFIG_PC99_TSC_FREQUENCY > 0
+    return CONFIG_PC99_TSC_FREQUENCY / 1000000;
+#endif
+
     x86_cpu_identity_t *model_info = x86_cpuid_get_model_info();
     uint32_t valid_models[] = {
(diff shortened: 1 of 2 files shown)


src/plat/pc99/machine/hardware.c
```diff
@@ -76,6 +76,10 @@ BOOT_CODE static inline uint32_t measure_tsc_khz(void)
 BOOT_CODE uint32_t tsc_init(void)
 {
 
+#if CONFIG_PC99_TSC_FREQUENCY > 0
+    return CONFIG_PC99_TSC_FREQUENCY / 1000000;
```
yyshen, line 80: worth saying that the CONFIG_PC99_TSC_FREQUENCY is in the unit of HZ, and the return value is MHz.

yyshen (approved):

src/plat/pc99/machine/hardware.c
kent-mcleod: I updated the config help string to say that the unit is Hz

lsf37:
The preprocess failure is unrelated to this PR (the new manifest hasn't made it out yet).

lsf37 (approved):
Good to merge.

PR #309, opened 2021-02-22
Add i.MX6 Nitrogen6_SoloX board

axel-h:
Patch is part of a series of patches to add support for the i.MX6 Nitrogen6_SoloX board

* https://github.com/seL4/camkes-tool/pull/57
* https://github.com/seL4/seL4_tools/pull/53
* https://github.com/seL4/util_libs/pull/69

In addition to `CONFIG_PLAT_NITROGEN6SX` this also adds `CONFIG_PLAT_IMX6DQ` and `CONFIG_PLAT_IMX6SX` internally depending on the i.MX6 board target. This allows differentiating between the SOCs  i.MX Double/Quad and i.MX SoloX, which makes much more sense in some cases than using a board selector like `CONFIG_PLAT_NITROGEN6SX`, just because this happens to be a board using this SOC.

### DIFF
--- a/src/plat/imx6/config.cmake
+++ b/src/plat/imx6/config.cmake
@@ -1,5 +1,6 @@
 #
 # Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2020, HENSOLDT Cyber GmbH
 #
 # SPDX-License-Identifier: GPL-2.0-only
 #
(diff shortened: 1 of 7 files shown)


axel-h:
sel4test log:
[sel4test.log](https://github.com/seL4/seL4/files/6024775/sel4test.log)

src/plat/imx6/overlay-nitrogen6sx.dts
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2020, Hensoldt Cyber
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
```
axel-h, line 4: Do you prefer this to be under GPL or is BSD acceptable here also?

lsf37: It'd be nice to avoid another BSD variant, i.e. BSD-2-Clause instead of BSD-3-Clause would be good.

Dual-licensed would be optimal for people who worry about having to think about different licenses (I don't see any dependencies on GPL code here, so BSD-2-Clause would be fine, but people worry about all kinds of things and if it's easy to do, why not).

axel-h: There is https://github.com/seL4/seL4/blob/master/LICENSES/BSD-3-Clause.txt, so using 'BSD-3-Clause' should be fully covered already. Unless there is a plan to remove this?

axel-h:
Is anything blocking this from a merge?

src/plat/imx6/config.cmake
```diff
@@ -8,26 +9,31 @@ cmake_minimum_required(VERSION 3.7.2)
 
 declare_platform(imx6 KernelPlatImx6 PLAT_IMX6 KernelSel4ArchAarch32)
 
-set(c_configs PLAT_SABRE PLAT_WANDQ)
-set(cmake_configs KernelPlatformSabre KernelPlatformWandQ)
-set(plat_lists sabre wandq)
-foreach(config IN LISTS cmake_configs)
-    unset(${config} CACHE)
```
kent-mcleod, line 15: This loop was here in order to be able to switch KernelArmPlatfrom in the build config editor without the previous config option remaining in the cache.  For example if you first select sabre, and then later re-select to wandq, the KernelPlatformSabre and KernelPlatImx6dq options need to be unset otherwise they will remain in the cache and can cause false positive platform checks to occur.

axel-h: Good to know. That is a workflow we are not using, we always have distinct workspaces (build directories) for each configuration. Plus the workspace is always wiped when a change in the build parameters is detected. We found that this is the the best way, as switching on the same workspace yields surprises sometimes. 
If you want to keep this switching feature, I have to rework some things also in other PR. But I have another branch that changes `declare_platform()` and I might be able to keep this working then. Haven made it a PR yet,see https://github.com/axel-h/seL4/commits/patch-axel-13

kent-mcleod: I would like to keep this behavior.  ( I did find myself thinking the other day that maybe the kernel build could get a large refactor to not use the cache at all which would avoid these state effects).

axel-h: @kent-mcleod I've change it that the variables are cleared

kent-mcleod (approved):

axel-h:
The build failure seem to come from the change of https://github.com/seL4/seL4/pull/323 that is merged now. Is there still a reference file that needs to be updated so this can pass?

lsf37:
I think the build failure is now gone. Let me double check @ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

PR #256, opened 2020-09-25
Synchronise master kernel and abstract model

xurtis:
This change outlines the changes to the master branch of the kernel that were used in the process of verifying the MCS implementation.

Some of these commits are still being actively worked on for verification; 7cc034b is one that has been discussed before but has not yet been included.

3ae3ee1673d24ab73f51f15ce198bd70c3cb9fae and 754cbc9 are needed given the invariants in the master kernel but the changes in #216 will make these checks unnecessary and should result in their eventual removal.

At some point there was a change to the representation used for the refill queue such that changed it from a pair of head and tail to a pair of head and the number of elements. This was done as changes (that have since been abandoned) were introduced that would lead to the queue being made empty during `refill_budget_check`. The current changes do not include the changes to the representation of the list as there is no longer the need to represent an empty list during operations on scheduling contexts.

A side effect of these changes having been made is that the Haskell specification currently uses the head+ count model and, as such, so do the refine proofs from the abstract model down to the Haskell specification. Resolving this requires either making the kernel consistent with the Haskell or the Haskell and refine proofs consistent with the kernel.

PR #312, opened 2021-02-26
Sub PR2 (Synchronise master kernel and abstract model)

kent-mcleod:
This PR is a reduced variant of the larger Synchronise master kernel and abstract model PR from #256.

It is expected that this PR is: "allowing scheduling contexts to accumulate budget or not, and keep the default behavior matching the current implementation while also supporting real sporadic servers for those who need that and are happy to abide by the constraints required for that to work with untrusted threads who can make calls to lower priority threads. This provides a new api option, not currently used. We can look at deprecating the original if we find no use for it."

Based on:

  799dfc4: mcs: schedule_used merge cases
  eb0c552: mcs: Use single function to charge budget to SC
  9b962c0: mcs: Refills are disjoint
  75bad2b: mcs: refill_unblock_check on thread activations
  12843fe: mcs: SC must be active for refill_unblock_check
  3c2231a: Don't unblock current SC on self-restart
  5598dc8: Only check and unblock SC once for notifications
  5927801: mcs: Expand comments describing preemtion cases
  e0140e7: squash! Only check and unblock SC once for notifications
  ae254fb: Set constant-bandwidth per SC
  a28c688: trivial: Update CHANGES for sporadic server option
  986c68d: Alias old SchedControl Configure API to new API
  757f720: mcs: Add sporadic flag to SchedControl_Configure

And a new commit extracted from:
    1465add: mcs: add function to check SC is released
    2d73678: mcs: Prevent binding SCs to blocked TCBs

### DIFF
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -150,7 +150,7 @@ static inline bool_t sc_released(sched_context_t *sc)
  */
 static inline bool_t sc_sporadic(sched_context_t *sc)
 {
-    return sc != NULL && !sc->scConstantBandwidth;
+    return sc != NULL && sc->scSporadic;
 }
 
 /*
@@ -160,7 +160,7 @@ static inline bool_t sc_sporadic(sched_context_t *sc)
  */
 static inline bool_t sc_constant_bandwidth(sched_context_t *sc)
 {
-    return sc->scConstantBandwidth;
+    return !sc->scSporadic;
 }
 
 /* Create a new refill in a non-active sc */
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -119,6 +119,31 @@ static inline bool_t refill_ready(sched_context_t *sc)
     return refill_head(sc)->rTime <= (NODE_STATE_ON_CORE(ksCurTime, sc->scCore) + getKernelWcetTicks());
 }
 
+/*
+ * Return true if an SC has been successfully configured with parameters
+ * that allow for a thread to run.
+ */
+static inline bool_t sc_active(sched_context_t *sc)
+{
+    return sc->scRefillMax > 0;
+}
+
+/*
+ * Return true if a SC has been 'released', if its head refill is
+ * sufficient and is in the past.
+ */
+static inline bool_t sc_released(sched_context_t *sc)
+{
+    if (sc_active(sc)) {
+        /* All refills must all be greater than MIN_BUDGET so this
+         * should be true for all active SCs */
+        assert(refill_sufficient(sc, 0));
+        return refill_ready(sc);
+    } else {
+        return false;
+    }
+}
+
 /* Create a new refill in a non-active sc */
 #ifdef ENABLE_SMP_SUPPORT
 void refill_new(sched_context_t *sc, word_t max_refills, ticks_t budget, ticks_t period, word_t core);
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -144,6 +144,25 @@ static inline bool_t sc_released(sched_context_t *sc)
     }
 }
 
+/*
+ * Return true if a SC's available refills should be delayed at the
+ * point the associated thread becomes runnable (sporadic server).
+ */
+static inline bool_t sc_sporadic(sched_context_t *sc)
+{
+    return sc != NULL && !sc->scConstantBandwidth;
+}
+
+/*
+ * Return true if a SC's available refills should be delayed at the
+ * point the associated thread becomes the current thread (constant
+ * bandwidth).
+ */
+static inline bool_t sc_constant_bandwidth(sched_context_t *sc)
+{
+    return sc->scConstantBandwidth;
+}
+
 /* Create a new refill in a non-active sc */
 #ifdef ENABLE_SMP_SUPPORT
 void refill_new(sched_context_t *sc, word_t max_refills, ticks_t budget, ticks_t period, word_t core);
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -166,14 +166,6 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
 void refill_budget_check(ticks_t used);
 
 /*
- * Charge a the current scheduling context `used` amount from its
- * current refill. This will split the refill, leaving whatever is
- * left over at the head of the refill. This is only called when charging
- * `used` will not deplete the head refill.
- */
-void refill_split_check(ticks_t used);
-
-/*
  * This is called when a thread is eligible to start running: it
  * iterates through the refills queue and merges any
  * refills that overlap.
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -64,11 +64,15 @@ UNUSED static inline void refill_print(sched_context_t *sc)
 /* check a refill queue is ordered correctly */
 static UNUSED bool_t refill_ordered(sched_context_t *sc)
 {
+    if (isRoundRobin(sc)) {
+        return true;
+    }
+
     word_t current = sc->scRefillHead;
     word_t next = refill_next(sc, sc->scRefillHead);
 
     while (current != sc->scRefillTail) {
-        if (!(refill_index(sc, current)->rTime <= refill_index(sc, next)->rTime)) {
+        if (!(refill_index(sc, current)->rTime + refill_index(sc, current)->rAmount <= refill_index(sc, next)->rTime)) {
             refill_print(sc);
             return false;
         }
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -210,15 +210,16 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
 
 static inline void schedule_used(sched_context_t *sc, refill_t new)
 {
-    /* schedule the used amount */
-    if (new.rAmount < MIN_BUDGET && !refill_single(sc)) {
-        /* used amount is to small - merge with last and delay */
+    if (unlikely(refill_tail(sc)->rTime + refill_tail(sc)->rAmount >= new.rTime)) {
+        /* Merge overlapping refill */
         refill_tail(sc)->rAmount += new.rAmount;
-        refill_tail(sc)->rTime = MAX(new.rTime, refill_tail(sc)->rTime);
-    } else if (new.rTime <= refill_tail(sc)->rTime) {
-        refill_tail(sc)->rAmount += new.rAmount;
-    } else {
+    } else if (likely(!refill_full(sc))) {
+        /* Add tail normally */
         refill_add_tail(sc, new);
+    } else {
+        /* Delay existing tail to merge */
+        refill_tail(sc)->rTime = new.rTime - refill_tail(sc)->rAmount;
+        refill_tail(sc)->rAmount += new.rAmount;
     }
 }
 
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -223,18 +223,6 @@ static inline void schedule_used(sched_context_t *sc, refill_t new)
     }
 }
 
-static inline void ensure_sufficient_head(sched_context_t *sc)
-{
-    /* ensure the refill head is sufficient, such that when we wake in awaken,
-     * there is enough budget to run */
-    while (refill_head(sc)->rAmount < MIN_BUDGET || refill_full(sc)) {
-        refill_t refill = refill_pop_head(sc);
-        refill_head(sc)->rAmount += refill.rAmount;
-        /* this loop is guaranteed to terminate as the sum of
-         * rAmount in a refill must be >= MIN_BUDGET */
-    }
-}
-
 static bool_t refill_head_overlapping(sched_context_t *sc)
 {
     if (!refill_single(sc)) {
@@ -249,94 +237,52 @@ static bool_t refill_head_overlapping(sched_context_t *sc)
 void refill_budget_check(ticks_t usage)
 {
     sched_context_t *sc = NODE_STATE(ksCurSC);
-    /* this function should only be called when the sc is out of budget */
-    ticks_t capacity = refill_capacity(NODE_STATE(ksCurSC), usage);
-    assert(capacity < MIN_BUDGET || refill_full(sc));
-    assert(sc->scPeriod > 0);
+    assert(!isRoundRobin(sc));
     REFILL_SANITY_START(sc);
 
-    if (capacity == 0) {
-        while (refill_head(sc)->rAmount <= usage) {
-            /* exhaust and schedule replenishment */
-            usage -= refill_head(sc)->rAmount;
-            if (refill_single(sc)) {
-                /* update in place */
-                refill_head(sc)->rTime += sc->scPeriod;
-            } else {
-                refill_t old_head = refill_pop_head(sc);
-                old_head.rTime = old_head.rTime + sc->scPeriod;
-                schedule_used(sc, old_head);
-            }
-        }
+    /* We assume that the estimates on worst-case kernel overheads are
+     * an upper bound and that overrun cannot occur. */
+#ifdef CONFIG_DEBUG_BUILD
+    if (usage > refill_head(sc)->rAmount) {
+        ticks_t overrun = usage - refill_head(sc)->rAmount;
+        userError(
+            "SC (%p) overran budget by %lluus (%llu ticks)",
+            sc,
+            ticksToUs(overrun),
+            overrun
+        );
+    }
+#endif
+    usage = MIN(usage, refill_head(sc)->rAmount);
 
-        /* budget overrun */
-        if (usage > 0) {
-            /* budget reduced when calculating capacity */
-            /* due to overrun delay next replenishment */
-            refill_head(sc)->rTime += usage;
-            /* merge front two replenishments if times overlap */
-            if (!refill_single(sc) &&
-                refill_head(sc)->rTime + refill_head(sc)->rAmount >=
-                refill_index(sc, refill_next(sc, sc->scRefillHead))->rTime) {
+    /* Charge the usage to the head refill */
+    if (usage > 0) {
+        refill_t used = (refill_t) {
+            .rAmount = usage,
+            .rTime = refill_head(sc)->rTime + sc->scPeriod,
+        };
 
-                refill_t refill = refill_pop_head(sc);
-                refill_head(sc)->rAmount += refill.rAmount;
-                refill_head(sc)->rTime = refill.rTime;
-            }
-        }
+        refill_head(sc)->rAmount -= usage;
+        /* We need to keep the head refill no more than a period before
+         * the start of the tail refill. */
+        refill_head(sc)->rTime += usage;
+        schedule_used(sc, used);
     }
 
-    capacity = refill_capacity(sc, usage);
-    if (capacity > 0 && refill_ready(sc)) {
-        refill_split_check(usage);
+    /* Ensure the head refill has the minimum budget */
+    while (refill_head(sc)->rAmount < MIN_BUDGET) {
+        refill_t head = refill_pop_head(sc);
+        refill_head(sc)->rAmount += head.rAmount;
+        /* Delay head to ensure the subsequent refill doesn't end any
+         * later (rather than simply combining refills). */
+        refill_head(sc)->rTime -= head.rAmount;
     }
 
-    ensure_sufficient_head(sc);
-
-    REFILL_SANITY_END(sc);
-}
-
-void refill_split_check(ticks_t usage)
-{
-    sched_context_t *sc = NODE_STATE(ksCurSC);
-    /* invalid to call this on a NULL sc */
-    assert(sc != NULL);
-    /* something is seriously wrong if this is called and no
-     * time has been used */
-    assert(usage > 0);
-    assert(usage <= refill_head(sc)->rAmount);
-    assert(sc->scPeriod > 0);
-
-    REFILL_SANITY_START(sc);
-
-    /* first deal with the remaining budget of the current replenishment */
-    ticks_t remnant = refill_head(sc)->rAmount - usage;
-
-    /* set up a new replenishment structure */
-    refill_t new = (refill_t) {
-        .rAmount = usage, .rTime = refill_head(sc)->rTime + sc->scPeriod
-    };
-
-    if (refill_size(sc) == sc->scRefillMax || remnant < MIN_BUDGET) {
-        /* merge remnant with next replenishment - either it's too small
-         * or we're out of space */
-        if (refill_single(sc)) {
-            /* update inplace */
-            new.rAmount += remnant;
-            *refill_head(sc) = new;
-        } else {
-            refill_pop_head(sc);
-            refill_head(sc)->rAmount += remnant;
-            schedule_used(sc, new);
-            ensure_sufficient_head(sc);
-        }
-        assert(refill_ordered(sc));
-    } else {
-        /* leave remnant as reduced replenishment */
-        assert(remnant >= MIN_BUDGET);
-        /* split the head refill  */
-        refill_head(sc)->rAmount = remnant;
-        schedule_used(sc, new);
+    /* Ensure that the refills are now not disjoint */
+    while (unlikely(refill_head_overlapping(sc))) {
+        refill_t head = refill_pop_head(sc);
+        refill_head(sc)->rAmount += head.rAmount;
+        refill_head(sc)->rTime = head.rTime;
     }
 
     REFILL_SANITY_END(sc);
--- a/src/kernel/sporadic.c
+++ b/src/kernel/sporadic.c
@@ -215,7 +215,18 @@ void refill_update(sched_context_t *sc, ticks_t new_period, ticks_t new_budget, 
 static inline void schedule_used(sched_context_t *sc, refill_t new)
 {
     if (unlikely(refill_tail(sc)->rTime + refill_tail(sc)->rAmount >= new.rTime)) {
-        /* Merge overlapping refill */
+        /* Merge overlapping or adjacent refill.
+         *
+         * refill_update can produce a tail refill that will overlap
+         * with new refills when time is charged to the head refill.
+         *
+         * Preemption will cause the head refill to be partially
+         * charged. When the head refill is again later charged the
+         * additionally charged amount will be added where the new
+         * refill ended such that they are merged here. This ensures
+         * that (beyond a refill being split as it is charged
+         * incrementally) a refill split is only caused by a thread
+         * blocking. */
         refill_tail(sc)->rAmount += new.rAmount;
     } else if (likely(!refill_full(sc))) {
         /* Add tail normally */
@@ -268,7 +279,15 @@ void refill_budget_check(ticks_t usage)
 
         refill_head(sc)->rAmount -= usage;
         /* We need to keep the head refill no more than a period before
-         * the start of the tail refill. */
+         * the start of the tail refill. This ensures that new refills
+         * are never added before the tail refill (breaking the ordered
+         * invariant). This code actually keeps the head refill no more
+         * than a period before the end of the tail refill (which is
+         * stronger than necessary) but is what is used in the current
+         * proofs. In combination with the merging behaviour of
+         * schedule_used, the following will still ensure that
+         * incremental charging of a refill across preemptions only
+         * produces a single new refill one period in the future. */
         refill_head(sc)->rTime += usage;
         schedule_used(sc, used);
     }
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -91,7 +91,7 @@ void restart(tcb_t *target)
         cancelIPC(target);
 #ifdef CONFIG_KERNEL_MCS
         setThreadState(target, ThreadState_Restart);
-        if (target->tcbSchedContext != NULL && sc_active(target->tcbSchedContext)
+        if (sc_sporadic(target->tcbSchedContext) && sc_active(target->tcbSchedContext)
             && target->tcbSchedContext != NODE_STATE(ksCurSC)) {
             refill_unblock_check(target->tcbSchedContext);
         }
@@ -142,7 +142,7 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
     assert(thread_state_get_replyObject(receiver->tcbState) == REPLY_REF(0));
     assert(reply->replyTCB == NULL);
 
-    if (receiver->tcbSchedContext && sc_active(receiver->tcbSchedContext)
+    if (sc_sporadic(receiver->tcbSchedContext) && sc_active(receiver->tcbSchedContext)
         && receiver->tcbSchedContext != NODE_STATE(ksCurSC)) {
         refill_unblock_check(receiver->tcbSchedContext);
     }
@@ -327,6 +327,9 @@ static void switchSchedContext(void)
         NODE_STATE(ksReprogram) = true;
         assert(refill_ready(NODE_STATE(ksCurThread->tcbSchedContext)));
         assert(refill_sufficient(NODE_STATE(ksCurThread->tcbSchedContext), 0));
+        if (sc_constant_bandwidth(NODE_STATE(ksCurThread)->tcbSchedContext)) {
+            refill_unblock_check(NODE_STATE(ksCurThread)->tcbSchedContext);
+        }
     }
 
     if (NODE_STATE(ksReprogram)) {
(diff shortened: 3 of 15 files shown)


include/kernel/sporadic.h
```diff
@@ -119,6 +119,50 @@ static inline bool_t refill_ready(sched_context_t *sc)
...
+}
+
+/*
+ * Return true if a SC's available refills should be delayed at the
+ * point the associated thread becomes the current thread (constant
+ * bandwidth).
+ */
+static inline bool_t sc_constant_bandwidth(sched_context_t *sc)
+{
+    return !sc->scSporadic;
```
kent-mcleod, line 163: @xurtis Should this also have an sc != NULL part?

src/object/schedcontext.c
```diff
@@ -278,6 +277,9 @@ void schedContext_bindTCB(sched_context_t *sc, tcb_t *tcb)
 
     SMP_COND_STATEMENT(migrateTCB(tcb, sc->scCore));
 
+    if (sc_active(sc)) {
```
kent-mcleod, line 280: This should have `sc_sporadic(tcb->tcbSchedContext)` I think ?

src/object/schedcontext.c
```diff
@@ -172,7 +172,6 @@ static exception_t invokeSchedContext_YieldTo(sched_context_t *sc, word_t *buffe
 
     bool_t return_now = true;
     if (isSchedulable(sc->scTcb)) {
-        refill_unblock_check(sc);
```
kent-mcleod, line 175: Does this need to be added back with !sc_sporadic(tcb->tcbSchedContext) around it?

src/object/notification.c
```diff
@@ -259,6 +278,16 @@ void completeSignal(notification_t *ntfnPtr, tcb_t *tcb)
         notification_ptr_set_state(ntfnPtr, NtfnState_Idle);
 #ifdef CONFIG_KERNEL_MCS
         maybeDonateSchedContext(tcb, ntfnPtr);
+#endif
+#ifdef CONFIG_KERNEL_MCS
+        if (tcb->tcbSchedContext != NULL && sc_active(tcb->tcbSchedContext)) {
```
kent-mcleod, line 283: Does this block also need an sc_sporadic() block?

include/kernel/sporadic.h
xurtis: It should probably be renamed, but this is never used in a situation where the SC may be NULL (because we're about to start executing with it). The other is always used in a situation where the SC may be null because it is called when a TCB is ublocked and the TCB may have had its SC removed at that stage.

src/object/notification.c
xurtis: I think you're right

src/object/schedcontext.c
xurtis: I don't think so. If the SC is sporadic then we don't unblock it as this isn't actually an activation and if the SC is continuous bandwidth limited then the unblock will happen in the switch to anyway.

xurtis: This was redundant for non sporadic and conflicted with sporadic (which is why it was not problematic before this PR)

src/object/schedcontext.c
xurtis: I think so. In general, Not guarding any of the `refill_unblock_check` that need to happen for a sporadic SC wont cause issue if the SC isn't sporadic, it just means that they're redundant.

src/kernel/sporadic.c
```diff
@@ -64,11 +64,15 @@ UNUSED static inline void refill_print(sched_context_t *sc)
 /* check a refill queue is ordered correctly */
 static UNUSED bool_t refill_ordered(sched_context_t *sc)
 {
+    if (isRoundRobin(sc)) {
```
kent-mcleod, line 67: I'm confused as to why this is here? RoundRobin threads don't have refill queues and aren't passed into any of the refill queue functions that call refill_ordered as far as I know.  Would it be more accurate for this to be an assert?

src/kernel/sporadic.c
```diff
@@ -248,94 +252,60 @@ static bool_t refill_head_overlapping(sched_context_t *sc)
...
-        }
-        assert(refill_ordered(sc));
-    } else {
-        /* leave remnant as reduced replenishment */
-        assert(remnant >= MIN_BUDGET);
-        /* split the head refill  */
-        refill_head(sc)->rAmount = remnant;
-        schedule_used(sc, new);
+    /* Ensure that the refills are now not disjoint */
+    while (unlikely(refill_head_overlapping(sc))) {
```
kent-mcleod, line 305: When would this condition become true? From what I can tell, any other operation leaves the queue without any overlapping refills and any changes to the queue earlier in this function also leaves the queue without overlapping entries.

mbrcknl: For binary verification, it would be very helpful if we could avoid this loop. :-)

src/kernel/sporadic.c
```diff
@@ -248,94 +252,60 @@ static bool_t refill_head_overlapping(sched_context_t *sc)
...
-    if (refill_size(sc) == sc->scRefillMax || remnant < MIN_BUDGET) {
-        /* merge remnant with next replenishment - either it's too small
-         * or we're out of space */
-        if (refill_single(sc)) {
-            /* update inplace */
-            new.rAmount += remnant;
-            *refill_head(sc) = new;
-        } else {
-            refill_pop_head(sc);
-            refill_head(sc)->rAmount += remnant;
```
kent-mcleod, line 328: I think this branch is necessary to handle the case when the refill queue is full (https://github.com/seL4/seL4/blob/master/include/kernel/thread.h#L258).  The part that is being removed would correctly result in refill_ready being false upon return from refill_budget_check inside chargeBudget.  This way when chargeBudget calls endTimeslice (https://github.com/seL4/seL4/blob/master/src/kernel/thread.c#L598) the thread doesn't get added to the release queue when it should.  (It takes the roundrobin path if there's no timeout fault handler).  Then it will keep looping through these steps upon any kernel entry until it uses up its available time I think.

src/kernel/thread.c
```diff
@@ -137,6 +141,11 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
     reply_remove(reply, receiver);
     assert(thread_state_get_replyObject(receiver->tcbState) == REPLY_REF(0));
     assert(reply->replyTCB == NULL);
+
+    if (sc_sporadic(receiver->tcbSchedContext) && sc_active(receiver->tcbSchedContext)
+        && receiver->tcbSchedContext != NODE_STATE(ksCurSC)) {
```
kent-mcleod, line 146: I think this line would need to have `NODE_STATE_ON_CORE(ksCurSC, receiver
->tcbSchedContext->scCore)` here to cover the scenario where the reply object is blocked on by a thread on a different core, and a different thread on the same core is currently executing with the SC held in that core's ksCurSC?

src/object/notification.c
```diff
@@ -103,6 +101,16 @@ void sendSignal(notification_t *ntfnPtr, word_t badge)
                     MCS_DO_IF_SC(tcb, ntfnPtr, {
                         possibleSwitchTo(tcb);
                     })
+#ifdef CONFIG_KERNEL_MCS
+                    if (tcb->tcbSchedContext != NULL && sc_active(tcb->tcbSchedContext)) {
+                        sched_context_t *sc = SC_PTR(notification_ptr_get_ntfnSchedContext(ntfnPtr));
+                        if (tcb->tcbSchedContext == sc) {
+                            /* Only unblock if the SC was donated from the
```
kent-mcleod, line 108: If the thread receiving the signal is currently in the ThreadState_RunningVM state then doesn't it have to already have an SC? So there wouldn't be any donated SC in this case?

kent-mcleod:
Here is another draft PR that's a subset of this PR that just contains changes relating to new sporadic server types: https://github.com/kent-mcleod/seL4/pull/5.
- I added a commit of proposed changes to squash at the top: https://github.com/kent-mcleod/seL4/pull/5/commits/8dadcc347bafe9bad4285868715277c0a150e0c3
- And added a new commit at the bottom: https://github.com/kent-mcleod/seL4/pull/5/commits/10895166309f1bdb62c1201fc5852da6f8f71f48

I also think that all of the new refill_unblock_check calls make it pretty hard to reason about when an SC is becomes activated from a usability perspective.  If there's a simple mental model for when this should happen then I would expect there to be a much smaller set of locations where refill_unblock_check is placed, such as `possibleSwitchTo`, but I understand that this may not be feasible to change at the moment.

src/kernel/sporadic.c
mbrcknl: I'm probably missing something, but what would prevent `refill_unblock_check` being called on a round-robin thread?

kent-mcleod: Nothing but `refill_unblock_check` already has this at the top:
```
    if (isRoundRobin(sc)) {
        /* nothing to do */
        return;
    }
```

mbrcknl: Thanks, that's what I missed. :-)

src/kernel/sporadic.c
mbrcknl: I also can't see how this loop would ever be executed. I can't see any code that ever creates overlapping refills.

mbrcknl: (at least not in this PR).

src/kernel/sporadic.c
mbrcknl: I don't fully understand everything here yet, so please bear with me...

Are you saying that when a non-round-robin thread enters the kernel with `refill_full` true, it should *always* end up `postpone`d? I can see that this is the case in what's currently on `master`, due to the branch you pointed out here. However, I think this is only the case if we assume that next refill does not overlap with the current one, and AFAICT, the code currently on `master` does *not* ensure non-overlapping refills. That's because there are a number of places where code on `master` adds to `rAmount` without adjusting `rTime` or checking whether the bigger refill overlaps with the next one.

If I understand what you're saying, the problem with the code in this PR is that non-round-robin threads that enter the kernel with enough time left to still be sufficient after usage is charged, but with `refill_full` true, will follow the round robin path. That is, the used portion will be taken from the front of the head refill and added to the end of the tail refill. And since the head remnant is still sufficient, the thread will go straight back into the ready queue.

Is that what you're saying? I can't tell whether this contradicts the last few words of the comment in `checkBudget` ("the rest of the budget is forfeit").

Another thing I noticed about this PR's version of `refill_budget_check`: When a thread enters the kernel with `refill_full` true, but where the head refill is *not* sufficient after the usage is charged, then after `refill_budget_check` returns, `refill_full` is no longer true for the current SC. That's because `schedule_used` is called first, which adds the used amount to the tail refill, and then the first `while` loop combines the remnant (which is still present as the head refill) with subsequent refills until it meets `MIN_BUDGET`. Is it intended that this situation always leads to the SC being *not* `refill_full`?

src/kernel/sporadic.c
```diff
@@ -248,94 +252,60 @@ static bool_t refill_head_overlapping(sched_context_t *sc)
...
     }
-
-    capacity = refill_capacity(sc, usage);
-    if (capacity > 0 && refill_ready(sc)) {
-        refill_split_check(usage);
+#endif
+    usage = MIN(usage, refill_head(sc)->rAmount);
+
+    /* Charge the usage to the head refill */
+    if (usage > 0) {
```
mbrcknl, line 274: IIUC, `usage` could only be zero if we arrive here via `checkBudget` where the current SC is not round-robin but is `refill_full`. Is that right? (If `ksConsumed` was zero, then we should have had `capacity >= MIN_BUDGET` in `checkBudget`.)

src/kernel/sporadic.c
mbrcknl: Thinking some more about that `refill_full` condition in `checkBudget`... Is it really the case that any interrupt should bump the current thread if it's SC is `refill_full`?

pingerino: @mbrcknl I think so.

The logic is as follows:

We don't know at the start of the kernel if we are going to have to split the budget or not. Assume we do have to split it (meaning we're switching scheduling contexts). Now if there is not enough space for the split to happen, we have to consume the whole head refill. Now we don't have enough time left in the SC because we had to consume the whole head refill (could't split it, no space) and it might need to be put in the release queue, but we were supposed to do that at the start of the kernel. 

There might be a way to wrangle this in a way that doesn't make a thread with plenty of time in its head refill lose it, involving more precarious charging, but I think you either lose what's at the head (hence bumping the thread) or push back the next replenishment. Either way you lose.

include/kernel/sporadic.h
pingerino: it should have an assert if that is the case

src/kernel/sporadic.c
pingerino: I intentionally tried to keep any round robin logic out of sporadic.c

src/kernel/sporadic.c
mbrcknl: Right, I think I see. The code in this PR appears to take a third approach, which is to push back the tail refill, so the used portion goes to the next period, and the existing tail then gets pushed back to adjoin it. But as you say, you still *eventually* lose when you get to what should have been that tail refill.

mbrcknl: Wait, maybe I didn't quite register everything you said, @pingerino... Why do we need to decide at kernel entry whether we're going to split the head refill? It's too late to do it in `switchSchedContext`?

I'm hoping it's better late than never to be asking these questions. :-)

src/kernel/sporadic.c
kent-mcleod: I think `usage` could only be 0 if `ksConsumed` is still 0 after `updateTimestamp` is called which would indicate getCurrentTime() hasn't returned a value that is strictly larger than a previous value it returned.
So usage > 0 could almost be an assert here I believe.

mbrcknl: Right, I understand that the `usage` here comes from `ksConsumed` after `updateTimestamp`. I was trying to understand what else must be true to have a zero `usage` here, since I don't think our axiomatisation of time allows us to prove that `getCurrentTime` has advanced. My reasoning was 

In particular, I was looking at `checkBudget`, which calls `chargeBudget` if `capacity < MIN_BUDGET` or if a non-round-robin thread is `refill_full`. My thinking was the `capacity < MIN_BUDGET` can only be true if `ksConsumed` is indeed non-zero, since we maintain the invariant that the head refill is at least `MIN_BUDGET`. And therefore, to have zero `usage` in `refill_budget_check`, we must be a non-round-robin thread which is `refill_full`.

However, there are other calls to `chargeBudget` that I hadn't taken into account, so that reasoning might not hold.

kent-mcleod: I think what you're claiming holds for the other calls to chargeBudget also.  They all either get called with ksConsumed or the head `rAmount` in the case of yield.

kent-mcleod:
> Do we have a clear idea of what questions need to be resolved before this (or something like it) can be merged? Since I've not caught up on the discussion on the original "synchronise" PR, I don't feel qualified to hit the "approve" button, but I'm keen to know what we can target for binary verification.

I think that the draft PR in https://github.com/seL4/seL4/pull/312#issuecomment-790190122 can be merged after only small cosmetic changes, and the leftovers in this PR would be all of the changes in sporadic.c which I believe still has unresolved questions.

kent-mcleod:
> I'm a bit confused by that comment. Were you saying earlier that you think there might be a simpler way of doing something like refill_unblock_check that wouldn't require so many call sites, but that we shouldn't try to figure that out now? (And hence should merge a change with the many call sites.)

Well I'm reviewing the changes under the assumption that the key difference in the new sporadic model is that refill_unblock_check is called differently.  Instead of being called whenever an SC is switched to it is called whenever a thread becomes unblocked and not after being preempted. Based on that assumption I would have thought that the new calls to refill_unblock_check would be called whenever:
- a thread became unblocked (so placed in possibleSwitchTo) unless the SC was being donated from the current thread to the new thread (so in possibleSwitchTo if the new SC != the ksCurSC) or,
- in any cases where a thread that's already active is switching SCs (On the reply receive paths).
This would likely be easier to check than all of the places where it is inserted now.
Ordinarily I would prefer that this would be figured out now but if there isn't sufficient time to make the change then I would at least like the PR to have a record of the discussion. 

> Are you able to make those cosmetic changes to kent-mcleod#5, and perhaps retarget to seL4/seL4?

I was looking for someone else to yay/nay the suggested changes first.

> Are you able to summarise what those questions are? Is it mainly the refill_full issue?

At the moment it seems to be:
- the refill_full issue (https://github.com/seL4/seL4/pull/312#discussion_r586106844) where the new change seems to reduce the amount of bandwidth a constant-bandwidth thread will have contrary to the goal stated here (https://github.com/seL4/seL4/pull/256#issuecomment-786310363). 
- the removal of overrun handling.
- Whether the new while loop in refill_budget_check is actually necessary.  (https://github.com/seL4/seL4/pull/312#discussion_r585245753)
- Whether roundRobin threads can end up in refill_ordered (https://github.com/seL4/seL4/pull/312#discussion_r585205462).

I'm trying to qualify what impact the changes to sporadic.c will have on existing uses as well as review the new functionality at the source level.  My goal isn't to block any of these changes, just to split up the overall PR so that the changes are applied as independently as possible so that it's easier to track in the future and also resolve any unanticipated issues.

src/kernel/sporadic.c
xurtis: There was a point at which the implementation I was using also applied to round-robin threads, and I think the constraints generally still apply to them although what I ended up with still never passed them into the sproadic.c functions.

src/kernel/sporadic.c
xurtis: The implementaion should now delay the tail if there isn't enough unused refills in the queue for a split, the only reason the thread wouldn't still be schedulable when the kernel exits is if the head would have less the MIN_BUDGET left after charging any consumed time.

src/kernel/sporadic.c
xurtis: I'm not sure as to why this ended up back in here. I also don't think there's any way for overlapping refills to occur.

src/kernel/thread.c
xurtis: This was only trying to exclude the case that the current SC was donated back to the receiving thread. If the SC on the thread isn't the current SC for the current core, it can't be the current SC on any core because it must have been bound to the thread that was blocked on the reply object.

xurtis: Ah, I see your point though. In which case your suggested change would be appropriate.

src/object/notification.c
xurtis: I don't think there is any guarantee for that. It's possible that the thread was running, had a bound SC removed and then this was signalled such that the SC from the bound notification was then bound to the VM.

src/kernel/sporadic.c
pingerino: Meta-comment: Part of the reason this PR is taking so long is there are lots of changes, and the commit messages are not explicitly stating what is happening, making it very difficult to review sufficiently. This is why what @kent-mcleod  has done is so valuable, but even in the split PR's, we are discovering new changes which we weren't aware of.

@xurtis can you explain why you thought it better to delay the tail? Does this deviate from the published algorithms? Really this should be a separate commit in itself, as it changes the scheduling behaviour. I don't recall why I chose one or the other.

@mbrcknl  we need to know at the start of the kernel if we have enough budget to do the operation we're in the kernel for, otherwise we can't do the operation and we need to simulate a timer tick. 

Example: If it's an enqueue into an IPC thread, we have to have more than MIN_BUDGET when we wake the thread up on the next kernel entry. If we check for the split at the end, we may end up in a case where we have to undo what we did, effectively looping in the kernel as we did an operation and now have to simulate the timer tick at the end.

src/object/notification.c
kent-mcleod: I think we need an isSporadic check here too anyway

xurtis:
> the refill_full issue (#312 (comment)) where the new change seems to reduce the amount of bandwidth a constant-bandwidth thread will have contrary to the goal stated here (#256 (comment)).

Increasing the bandiwidth on the non-sporadic tasks wasn't a goal, just a possible side effect of the changes. The point I wanted to make there was that if the bandwidth _is_ increased, it isn't increased beyond the point of violating the sliding window constraint.

The change does increase the bandwidth at some instants though, rather than delaying the head to merge with a subsequent refill when there are insufficient refill slots for a split, the split is performed anyway and, in the case that there aren't sufficient refill slots, the tail is delayed and merged.

> the removal of overrun handling.

If the kernel WCET and interrupt delivery latency are actually proper over-estimates, then overrun shouldn't occur. The defaults stop being accurate when you enable kernel debug printing though.

If we want to assume that a correct configuration of the kernel will still need to deal with overrun, then the proofs could be set back by months as the proofs need to show the bound for integer overflow in that case which would either require some fixed upper-bound on expected overrun or for the substantially more complex proof of showing that the maximum possible overrun that could be produced by the system wouldn't then cause so much time to be accounted for in that loop that it places any of the refills far enough into the future to overflow. It would also do horrible things for kernel WCET.

My argument here is that we can determine a legitimate overestimate for kernel WCET and timer latency such that overflow never occurs and that the kernel no longer provides as strict timing guarantees where that isn't performed correctly. If we do add the logic to account for overrun, it could also kill our ability to have a bounded kernel WCET and introduce a large verification complexity.

> Whether the new while loop in refill_budget_check is actually necessary. (#312 (comment))

It doesn't appear to be and I am fairly sure the changes I made intentionally avoided ever producing overlapping refills so I'm unsure as to why it was introduced. It may have been that it was added to master in the process and absorbed through the rebasing process.

> Whether roundRobin threads can end up in refill_ordered (#312 (comment)).

Currently, they should not ever be passed to code in `sporadic.c`.

pingerino:
> If the kernel WCET and interrupt delivery latency are actually proper over-estimates, then overrun shouldn't occur. The defaults stop being accurate when you enable kernel debug printing though.

This is not true.

The kernel WCET is a guess. We don't have a kernel WCET at all for the RT kernel, and it may be completely impractical to use the real value (not sure what it is). 

Timer IRQ accuracy on these platforms is also unknown, as is the WCET from getting an IRQ and it arriving in the kernel. 
You are effetively stating that seL4 MCS cannot be used on any platform without a WCET measurement which seems impractical given the state of the art in WCET. 

> My argument here is that we can determine a legitimate overestimate for kernel WCET and timer latency such that overflow never occurs and that the kernel no longer provides as strict timing guarantees where that isn't performed correctly.

How? this isn't possible on *any* of our current platforms, except maybe the KZM. What is the current 'WCET' estimate for the kernel?

Even more so, which hardware platforms guarantee that timer interrupts are precise? Are there *any* that are not microcontrollers? What about under an interrupt storm?

xurtis:
> What is the current 'WCET' estimate for the kernel?

I think the current configuration on all platforms is 10ms.

pingerino:
> > What is the current 'WCET' estimate for the kernel?
> 
> I think the current configuration on all platforms is 10ms.

That is the guess number which seemed to work in testing and could be used if we have overrun handling.
What is the actual WCET of the kernel for any platform?

xurtis:
I think I understand your point and my argument is flawed.

If we do need to account for overrun, the original implementation that repeatedly charged individual refills has proven to be far too unwieldy to incorporate into the proofs. I've spoken with @michaelmcinerney who, along with @MitchellBuckley, was working to implement the proofs for that loop and ran into some difficulty.

To return to the discussion that originally put me onto working on MCS, there was an idea that the entire budget could be scheduled in the future all at once and at the time I didn't really understand the context well enough to determine how to do this without losing the principle from the original mechanism used to do this.

The following is an attempt at at an implementation that doesn't lose the underlying principle from the paper regarding the POSIX fixes that should hopefully indicate a much easier upper bound on current time and that could be more tractably included in the proofs:

```c
// Deal with overrun by delaying all refills to be merged with tail
if (usage < REFILL_HEAD(sc).rAmount) {
    refill_t tail = REFILL_TAIL(sc);
    tail.rTime -= sc->scBudget - tail.tAmount;
    tail.rAmount = sc->scBudget;
    REFILL_HEAD(sc) = tail;
    sc->scRefillCount = 1;
}

// Charge the entire tail repeatedly until usage can be charged
// to a single refill. The two following blocks should be equivalent.
#ifdef CAN_INT64_DIVIDE
REFILL_HEAD(sc).rTime += (usage / sc->scBudget) * sc->scPeriod;
usage = usage % sc->scBudget;
#else
while (usage > REFILL_HEAD(sc).rAmount) {
    REFILL_HEAD(sc).rTime += sc->scPeriod;
    usage -= REFILL_HEAD(sc).rAmount;
}
#endif
```

mbrcknl:
> I'm reviewing the changes under the assumption that the key difference in the new sporadic model is that refill_unblock_check is called differently. Instead of being called whenever an SC is switched to it is called whenever a thread becomes unblocked and not after being preempted. Based on that assumption I would have thought that the new calls to refill_unblock_check would be called whenever:
> 
>     * a thread became unblocked (so placed in possibleSwitchTo) unless the SC was being donated from the current thread to the new thread (so in possibleSwitchTo if the new SC != the ksCurSC) or,
> 
>     * in any cases where a thread that's already active is switching SCs (On the reply receive paths).
>       This would likely be easier to check than all of the places where it is inserted now.
>       Ordinarily I would prefer that this would be figured out now but if there isn't sufficient time to make the change then I would at least like the PR to have a record of the discussion.

Thanks, that makes sense. I would guess that consolidating the calls to `refill_unblock_check` to a smaller number of sites would be good for binary verification. It's hard to guess what the impact would be on the other proofs, since it would depend on what else would need to be refactored.

> At the moment it seems to be:
>   ...

Great summary, thanks!

mbrcknl:
@xurtis I guess `if (usage < REFILL_HEAD(sc).rAmount)` should be `if (REFILL_HEAD(sc).rAmount < usage)`, i.e. flip the comparison?

xurtis:
> @xurtis I guess `if (usage < REFILL_HEAD(sc).rAmount)` should be `if (REFILL_HEAD(sc).rAmount < usage)`, i.e. flip the comparison?

Yes, condition should be the same as the loop condition

kent-mcleod:
> `usage = usage % sc->scBudget;`

What happens to the remaining usage afterwards?  Would it be charged by the remainder of the function as usual?

mbrcknl:
> > `usage = usage % sc->scBudget;`
> 
> What happens to the remaining usage afterwards? Would it be charged by the remainder of the function as usual?

It would be helpful to see @xurtis' proposal in context. I'm wondering if it should actually look something like this:

```
// Deal with overrun by delaying all refills to be merged with tail
if (usage > REFILL_HEAD(sc).rAmount) {
    refill_t tail = REFILL_TAIL(sc);
    tail.rTime -= sc->scBudget - tail.tAmount;
    tail.rAmount = sc->scBudget;
    REFILL_HEAD(sc) = tail;
    sc->scRefillCount = 1;

  // Charge the entire tail repeatedly until usage can be charged
  // to a single refill. The two following blocks should be equivalent.
#ifdef CAN_INT64_DIVIDE
  REFILL_HEAD(sc).rTime += (usage / sc->scBudget) * sc->scPeriod;
  usage = usage % sc->scBudget;
#else
  while (usage > REFILL_HEAD(sc).rAmount) {
      REFILL_HEAD(sc).rTime += sc->scPeriod;
      usage -= REFILL_HEAD(sc).rAmount;
  }
#endif
  return;
}
// Normal (non-overrun) handling follows...
```

kent-mcleod:
I think @xurtis's proposal is good, but it introduces a potentially large discontinuity where (usage - REFILL_HEAD(sc).rAmount) goes from 0 to 1.  For a 1 tick overrun all refills get merged into the tail which can be up to a full period delay before any usage is charged.  This represents overcharging the SC by almost a full scBudget for a single tick overrun in addition to the overrun amount.

If the division approach makes it easier to verify, then what about a function that loops over the refill queue only once and produces the same result as the original loop?:
```
    if (usage > REFILL_HEAD(sc).rAmount) {
        ticks_t usage_remainder = usage % sc->scBudget;
        ticks_t usage_periods = usage / sc->scBudget;

        word_t tail_index = sc->scRefillTail;
        word_t index_pointer = sc->scRefillHead;
        // Process every refill in the queue exactly once.
        // Every refill gets it's rTime advanced by scPeriod * usage_periods.
        // While usage_remainder is >= the head refill it gets delayed by an additional period (usage_periods + 1)
        // and moved to the back of the queue. 

        word_t seen_tail = false;
        while (seen_tail == false) {
            // Loop condition ends when we get to the tail_index from before we started
            if (index_pointer == tail_index) {
                seen_tail = true;
            }

            // If usage_remainder is still greater than full refill amounts, move these refills
            // to the back of the queue by popping and adding tails.
            // Everytime the first branch is taken index_pointer should match scRefillHead.
            if (refill_index(sc, index_pointer).rAmount <= usage_remainder) {
                assert(sc->scRefillHead == index_pointer);
                refill_t head = refill_pop(sc);
                usage_remainder -= head.rAmount;
                head.rTime += sc->scPeriod * (usage_periods+1);
                refill_add_tail(sc, head);
            } else {
                refill_index(sc, index_pointer).rTime += scPeriod * usage_periods;
            }
            index_pointer = refill_next(sc, index_pointer);
        }
        usage = usage_remainder;
        assert(usage < REFILL_HEAD(sc).rAmount);
    }
    // remainder of refill_split_check functionality follows
```

xurtis:
An alternative option suggested by @lsf37 that is closer to the existing verification work is to remove integer overflow at runtime in the problematic loop. I'd like to get input from @michaelmcinerney on this and ask if it, or something similar to it, could ease the issues in verification.

Can you confirm @michaelmcinerney that the issues with this loop were mostly to determine the maximum `ksCurTime`?.

```c
/* Never push refills into the future */
while (
    refill_head(sc)->rAmount  <= usage &&
    /* max_time */ - refill_head(sc)->rTime >= sc->scPeriod
) {
    usage -= refill_head(sc)->rAmount;
    ticks_t new_time = refill_head(sc)->rTime;
    if (/* max_time */ - new_time < sc->scPeriod) {

    if (refill_single(sc)) {
        refill_head(sc)->rTime += sc->scPeriod;
    } else {
        refill_t old_head = refill_pop_head(sc);
        old_head.rTime = old_head.rTime + sc->scPeriod;
        schedule_used(sc, old_head);
    }
}

if (refill_head(sc)->rAmount  <= usage) {
    /*
     * We've reached a point where we can no longer correctly account for
     * time, so we now just need to exit the kernel without breaking invariants.
     *
     * At this point, trying to advance any other refills in this SC would lead to
     * overflow, so we just don't do it.
     */
    return;
}
```

kent-mcleod:
@xurtis  should getMaxTicksToUs be getMaxUsToTicks or just UINT64_MAX as our units are ticks and not miccroseconds.   Also is this line an unintended addition: `    if (/* max_time */ - new_time < sc->scPeriod) {`?

xurtis:
> @xurtis should getMaxTicksToUs be getMaxUsToTicks or just UINT64_MAX as our units are ticks and not miccroseconds. Also is this line an unintended addition: ` if (/* max_time */ - new_time < sc->scPeriod) {`?

Sorry, the line is an addition. The values in this case are in ticks, not us, so the overflow would occur either when they hit UINT64_MAX or the maximum value that ticks can get to before overflowing when converted to US.

kent-mcleod:
But the function your example calls, `getMaxTicksToUs()`, returns a value that's in US and not ticks?

xurtis:
> But the function your example calls, `getMaxTicksToUs()`, returns a value that's in US and not ticks?

`getMaxTicksToUs` should be the largest value that can be passed to `ticksToUs` without overflowing.

kent-mcleod:
Shouldn't it be `getMaxUsToTicks`? Looking at https://github.com/seL4/seL4/blob/master/include/arch/riscv/arch/machine/timer.h#L42 getMaxTicksToUs is taking max ticks and converting that to a us value isn't it?  So it would need to be converted back to ticks before being subtracted by refill_head(sc)->rTime (which is in ticks)

xurtis:
> What happens to the remaining usage afterwards? Would it be charged by the remainder of the function as usual?

This would come before the existing charging logic that partially charges the head.

xurtis:
I think you're right. The definitions of the functions there are a little confusing, but `getMaxTicksToUs` does seem to be the upper bound in us, and `getMaxUsToTicks` seems to produce upper bounds in ticks.

xurtis:
@kent-mcleod @pingerino @lsf37 @michaelmcinerney 

Can I get you to check out https://github.com/kent-mcleod/seL4/pull/7 which is a PR onto this branch? It should resolve the issue of accounting for overrun in a manner that is easy to incorporate into verification. @michaelmcinerney seems to already be making good progress with this.

xurtis:
I believe that #348 is intended to replace this. If so, should we close this PR @kent-mcleod ?

kent-mcleod:
> I believe that #348 is intended to replace this. If so, should we close this PR @kent-mcleod ?

Yup

PR #320, opened 2021-03-17
manual: Correct description of CNode addressing

nspin:
A CPointer addressing a CNode is accompanied by a depth limit. The manual describes the depth limit as the length of a big-endian prefix of the CPointer, whereas the specification and implementation treat the depth limit as the length of a big-endian suffix.

This commit corrects the manual to match the specification.

[1] https://github.com/seL4/l4v/blob/dd6529f520a25328469fc44cfd54f3b29ffd75cc/spec/abstract/CSpace_A.thy#L243
[2] https://github.com/seL4/l4v/blob/dd6529f520a25328469fc44cfd54f3b29ffd75cc/spec/haskell/src/SEL4/Kernel/CSpace.lhs#L128
[3] https://github.com/seL4/seL4/blob/d6b83163b55c20ad146620fe82e261221d321316/src/kernel/cspace.c#L126

### DIFF
--- a/manual/parts/cspace.tex
+++ b/manual/parts/cspace.tex
@@ -397,10 +397,12 @@ and addressing a range of capability slots.
 
 Recall that the translation algorithm described above will traverse
 \obj{CNode} capabilities while there are address bits remaining to be
-translated. Therefore, in order to address a \obj{CNode} capability,
-the user must supply not only a capability address but also specify
-the maximum number of bits of the capability address that are to be
-translated, called the \emph{depth limit}.
+translated. Therefore, in order to address a capability which may be
+a \obj{CNode} capability, the user must supply not only a capability
+address but also specify the maximum number of bits of the capability
+address that are to be translated, called the \emph{depth limit}.
+When a CPointer is paired with depth limit \textit{depth}, only its
+\textit{depth} least significant bits are used in translation.
 
 Certain methods, such as
 \apifunc{seL4\_Untyped\_Retype}{untyped_retype}, require the user to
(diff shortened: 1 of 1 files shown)


manual/parts/cspace.tex
```diff
@@ -425,11 +427,12 @@ \subsection{Addressing Capabilities}
...
 \begin{description}
 \item[Cap A.] The first CNode has a 4-bit guard set to 0x0, and an
-  8-bit radix. Cap A resides in slot 0x60 so it may be referred to by
-  any address of the form 0x060xxxxx (where xxxxx is any number,
-  because the translation process terminates after translating the
-  first 12 bits of the address). For simplicity, we usually adopt the
-  address 0x06000000.
+  8-bit radix. Cap A resides in slot 0x60 so, provided that it is not
+  a \obj{CNode} capability, it may be referred to by any address of
+  the form 0x060\textit{aaaaa} (where \textit{aaaaa} is any sequence
```
axel-h, line 432: Maybe we should stick with using "xxxxx" instead of "aaaaa", because 'a' is a valid hex char and that could become confusing even if there is an explanation given immediately after. In the example below it could say then "adopt the address 0x06000000 instead of e.g. 0x060abcde"

manual/parts/cspace.tex
```diff
@@ -451,15 +454,15 @@ \subsection{Addressing Capabilities}
   the user must supply not only a capability address but also specify
   the depth limit, which is the maximum number of bits to be
   translated.  L2 CNode Cap resides at offset 0x0F of the first CNode,
-  which has a 4-bit guard of 0x0.  Hence, its address is 0x00F00000,
-  with a depth limit of 12 bits.
+  which has a 4-bit guard of 0x0.  Hence, it may be referred to by any
+  address of the form 0x\textit{aaaaa}00F with a depth limit of 12
```
axel-h, line 458: same here, better use "xxxxx"

manual/parts/cspace.tex
```diff
@@ -451,15 +454,15 @@ \subsection{Addressing Capabilities}
...
+  bits, where \textit{aaaaa} is any sequence of 5 hexadecimal digits.
 
 \item[L3 CNode Cap.] This capability resides at index 0x00 of the
   second CNode, which is reached by the L2 CNode Cap. The second CNode
-  has a 4-bit guard of 0x0. Hence, the capability's address is
-  0x00F00000 with a depth limit of 24 bits. Note that the addresses of
-  the L2 and L3 CNode Caps are the same, but that their depth limits
-  are different.
+  has a 4-bit guard of 0x0. Hence, the capability may be referred to
+  by any address of the form 0x\textit{aa}00F000 with a depth limit of
```
axel-h, line 464: same here, use "xx" instead of "aa"

manual/parts/cspace.tex
nspin: I agree that `a` isn't the best choice because it is a valid hexadecimal digit. As you noted below, `x` is confusing in one of the other cases because the `x`s directly follow the `0x` prefix. I tried italicizing the placeholder `x`s, but the typeset result looked bad. `n`, which you suggest below, seems like the best choice to me. I've changed each instance to use `n`.

axel-h:
Looks good to me. This verbose explanation should have been there years ago when I tried to understand the black magic here.

axel-h (approved):

lsf37 (approved):
Thanks for that, this is much clearer.

lsf37:
@oliver-wm can you do the merge, please?

oliver-wm:
@ssrg-bamboo `test`

ssrg-bamboo:

PR #321, opened 2021-03-19
trivial: Make the branch easier to understand.

baohaojun:
It's impossible for "n_bits < levelBits", because a few lines above, it
has already judged this case and would have returned with error:

        if (unlikely(levelBits > n_bits)) {
            current_lookup_fault =
                lookup_fault_depth_mismatch_new(levelBits, n_bits);
            ret.status = EXCEPTION_LOOKUP_FAULT;
            return ret;
        }

Signed-off-by: Bao Haojun <baohaojun@lixiang.com>

### DIFF
--- a/src/kernel/cspace.c
+++ b/src/kernel/cspace.c
@@ -171,7 +171,7 @@ resolveAddressBits_ret_t resolveAddressBits(cap_t nodeCap, cptr_t capptr, word_t
         offset = (capptr >> (n_bits - levelBits)) & MASK(radixBits);
         slot = CTE_PTR(cap_cnode_cap_get_capCNodePtr(nodeCap)) + offset;
 
-        if (likely(n_bits <= levelBits)) {
+        if (likely(n_bits == levelBits)) {
             ret.status = EXCEPTION_NONE;
             ret.slot = slot;
             ret.bitsRemaining = 0;


mbrcknl:
I do like this change, but it might affect the proofs (hence the "Preprocess" test failures). I'll start a run of the proofs to see what breaks...

mbrcknl:
For those with Data61-internal access, proofs are checking at [L4V-GHTESTBOARD5-1](https://bamboo.ts.data61.csiro.au/browse/L4V-GHTESTBOARD5-1).

mbrcknl:
The proofs have indeed failed. This is not particularly surprising, but it means that someone would need to explain to the theorem prover that these conditions are equivalent in this context.

axel-h:
@baohaojun: an intermediate code improvement here could be adding a comment with your findings and the remark that this can't be changes in the code at the moment without a proof update.

lsf37:
I'll have a look at that proof update, should be trivial.

lsf37:
I've rebased the commit and and a verification PR for it will be incoming shortly.

lsf37 (approved):
This can be merged as soon as the proofs on seL4/l4v#400 pass.

The change actually brings the code more in line with the specs, which already use `==` here.

PR #322, opened 2021-03-22
Fastpath for signalling a notification on MCS kernel

shanekadish:
# Summary
Added a fastpath for signalling a notification on the MCS kernel. Benchmarks run with [sel4bench](https://github.com/seL4/sel4bench) indicate a speedup of signal operations by up to 76% in non-context-switching cases, and up to 56% in context-switching cases

---

# Results

Platform: sabre lite
* quad-core ARM Cortex A9 processor, running at 1GHz
* 1GB of DDR3 RAM @ 532Mhz

## Without Signal Fastpath
```
One way IPC microbenchmarks:
Mean: 314
StdDev: 1
Samples: 16

Signal overhead:
Mean: 7
StdDev: 0
Samples: 100

Signal to high prio thread:
Mean: 1048
StdDev: 170
Samples: 100

Signal to low prio thread:
Mean: 604
StdDev: 6
Samples: 100

Hardware null_syscall thread:
Mean: 121
StdDev: 0
Samples: 100
```

## With Signal Fastpath
```
One way IPC microbenchmarks:
Mean: 315
StdDev: 1
Samples: 16

Signal overhead:
Mean: 7
StdDev: 0
Samples: 100

Signal to high prio thread:
Mean: 485
StdDev: 298
Samples: 100

Signal to low prio thread:
Mean: 136
StdDev: 0
Samples: 100

Hardware null_syscall thread:
Mean: 121
StdDev: 0
Samples: 100
```

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -255,13 +255,11 @@ config_string(
     UNQUOTE
 )
 
-
 config_string(
     KernelSignalFastpath SIGNAL_FASTPATH
     "Enable notification signal fastpath"
     DEFAULT OFF
     DEPENDS "KernelIsMCS" UNDEF_DISABLED
-    DEPENDS "${KernelNumDomains} EQUAL 1"
 )
 
 find_file(
--- a/config.cmake
+++ b/config.cmake
@@ -248,6 +248,7 @@ config_string(
     UNQUOTE
 )
 config_option(KernelFastpath FASTPATH "Enable IPC fastpath" DEFAULT ON)
+config_option(KernelSignalFastpath SIGNAL_FASTPATH "Enable notification signal fastpath" DEFAULT OFF)
 
 config_string(
     KernelNumDomains NUM_DOMAINS "The number of scheduler domains in the system"
--- a/config.cmake
+++ b/config.cmake
@@ -248,12 +248,20 @@ config_string(
     UNQUOTE
 )
 config_option(KernelFastpath FASTPATH "Enable IPC fastpath" DEFAULT ON)
-config_option(KernelSignalFastpath SIGNAL_FASTPATH "Enable notification signal fastpath" DEFAULT OFF)
 
 config_string(
     KernelNumDomains NUM_DOMAINS "The number of scheduler domains in the system"
     DEFAULT 1
     UNQUOTE
+)
+
+
+config_string(
+    KernelSignalFastpath SIGNAL_FASTPATH
+    "Enable notification signal fastpath"
+    DEFAULT OFF
+    DEPENDS "KernelIsMCS" UNDEF_DISABLED
+    DEPENDS "${KernelNumDomains} EQUAL 1"
 )
 
 find_file(
--- a/include/fastpath/fastpath.h
+++ b/include/fastpath/fastpath.h
@@ -6,6 +6,11 @@
 
 #pragma once
 
+#ifdef CONFIG_KERNEL_MCS
+#include <object/reply.h>
+#include <object/notification.h>
+#endif
+
 /* Fastpath cap lookup.  Returns a null_cap on failure. */
 static inline cap_t FORCE_INLINE lookup_fp(cap_t cap, cptr_t cptr)
 {
@@ -113,6 +118,97 @@ static inline reply_t *thread_state_get_replyObject_np(thread_state_t ts)
 #endif
 }
 #endif
+
+static inline void cancelSignal_fp(tcb_t *tptr, notification_t *ntfnPtr)
+{
+    tcb_queue_t ntfn_queue;
+
+    /* Haskell error "cancelSignal: notification object must be in a waiting" state */
+    assert(notification_ptr_get_state(ntfnPtr) == NtfnState_Waiting);
+
+    ntfn_queue.head = (tcb_t *)notification_ptr_get_ntfnQueue_head(ntfnPtr);
+    ntfn_queue.end = (tcb_t *)notification_ptr_get_ntfnQueue_tail(ntfnPtr);
+
+    /* Dequeue TCB */
+    ntfn_queue = tcbEPDequeue(tptr, ntfn_queue);
+    notification_ptr_set_ntfnQueue_head(ntfnPtr, (word_t)ntfn_queue.head);
+    notification_ptr_set_ntfnQueue_tail(ntfnPtr, (word_t)ntfn_queue.end);
+
+    /* set the thread state to idle if the queue is empty */
+    if (!ntfn_queue.head) {
+        notification_ptr_set_state(ntfnPtr, NtfnState_Idle);
+    }
+
+    /* Make thread inactive */
+    thread_state_ptr_set_tsType_np(&tptr->tcbState, ThreadState_Inactive);
+}
+
+static inline void cancelIPC_fp(tcb_t *tptr)
+{
+    thread_state_t *state = &tptr->tcbState;
+
+#ifdef CONFIG_KERNEL_MCS
+    /* cancel ipc cancels all faults */
+    seL4_Fault_NullFault_ptr_new(&tptr->tcbFault);
+#endif
+
+    switch (thread_state_ptr_get_tsType(state)) {
+    case ThreadState_BlockedOnSend:
+    case ThreadState_BlockedOnReceive: {
+        /* blockedIPCCancel state */
+        endpoint_t *epptr;
+
+        epptr = EP_PTR(thread_state_ptr_get_blockingObject(state));
+
+        /* Haskell error "blockedIPCCancel: endpoint must not be idle" */
+        assert(endpoint_ptr_get_state(epptr) != EPState_Idle);
+
+        /* Dequeue TCB */
+        endpoint_ptr_set_epQueue_head_np(epptr, TCB_REF(tptr->tcbEPNext));
+        if (unlikely(tptr->tcbEPNext)) {
+            tptr->tcbEPNext->tcbEPPrev = NULL;
+        } else {
+            endpoint_ptr_mset_epQueue_tail_state(epptr, 0, EPState_Idle);
+        }
+
+#ifdef CONFIG_KERNEL_MCS
+        reply_t *reply = REPLY_PTR(thread_state_get_replyObject(tptr->tcbState));
+        if (reply != NULL) {
+            reply_unlink(reply, tptr);
+        }
+#endif
+        thread_state_ptr_set_tsType_np(&tptr->tcbState, ThreadState_Inactive);
+        break;
+    }
+
+    case ThreadState_BlockedOnNotification:
+        cancelSignal_fp(tptr,
+                     NTFN_PTR(thread_state_ptr_get_blockingObject(state)));
+        break;
+
+    case ThreadState_BlockedOnReply: {
+#ifdef CONFIG_KERNEL_MCS
+        reply_remove_tcb(tptr);
+#else
+        cte_t *slot, *callerCap;
+
+        tptr->tcbFault = seL4_Fault_NullFault_new();
+
+        /* Get the reply cap slot */
+        slot = TCB_PTR_CTE_PTR(tptr, tcbReply);
+
+        callerCap = CTE_PTR(mdb_node_get_mdbNext(slot->cteMDBNode));
+        if (callerCap) {
+            /** GHOSTUPD: "(True,
+                gs_set_assn cteDeleteOne_'proc (ucast cap_reply_cap))" */
+            cteDeleteOne(callerCap);
+        }
+#endif
+
+        break;
+    }
+    }
+}
 
 #include <arch/fastpath/fastpath.h>
 
--- a/include/object/notification.h
+++ b/include/object/notification.h
@@ -8,6 +8,7 @@
 
 #include <types.h>
 #include <object/structures.h>
+#include <object/schedcontext.h>
 
 void sendSignal(notification_t *ntfnPtr, word_t badge);
 void receiveSignal(tcb_t *thread, cap_t cap, bool_t isBlocking);
@@ -36,4 +37,31 @@ static inline void maybeReturnSchedContext(notification_t *ntfnPtr, tcb_t *tcb)
 }
 #endif
 
+#ifdef CONFIG_KERNEL_MCS
+static inline void maybeDonateSchedContext(tcb_t *tcb, notification_t *ntfnPtr)
+{
+    if (tcb->tcbSchedContext == NULL) {
+        sched_context_t *sc = SC_PTR(notification_ptr_get_ntfnSchedContext(ntfnPtr));
+        if (sc != NULL && sc->scTcb == NULL) {
+            schedContext_donate(sc, tcb);
+            if (sc != NODE_STATE(ksCurSC)) {
+                /* refill_unblock_check should not be called on the
+                 * current SC as it is already running. The current SC
+                 * may have been bound to a notificaiton object if the
+                 * current thread was deleted in a long-running deletion
+                 * that became preempted. */
+                refill_unblock_check(sc);
+            }
+            schedContext_resume(sc);
+        }
+    }
+}
 
+static inline void ntfn_set_active(notification_t *ntfnPtr, word_t badge)
+{
+    notification_ptr_set_state(ntfnPtr, NtfnState_Active);
+    notification_ptr_set_ntfnMsgIdentifier(ntfnPtr, badge);
+}
+
+
+#endif
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -22,8 +22,19 @@ static inline
 FORCE_INLINE
 #endif
 #endif
-void NORETURN fastpath_signal(cap_t cap)
+void NORETURN fastpath_signal(word_t cptr)
 {
+    /* We land up in here on every SYSCALL_SEND, so we need to check that we are
+     * actually signalling a notification */
+
+    /* Lookup the cap */
+    cap_t cap = lookup_fp(TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbCTable)->cap, cptr);
+
+    /* Check it's a notification */
+    if (!cap_capType_equals(cap, cap_notification_cap)) {
+        slowpath(SysSend);
+    }
+
     notification_t *ntfnPtr = NTFN_PTR(cap_notification_cap_get_capNtfnPtr(cap));
     uint32_t ntfnState = notification_ptr_get_state(ntfnPtr);
     word_t badge = cap_notification_cap_get_capNtfnBadge(cap);
@@ -39,6 +50,7 @@ void NORETURN fastpath_signal(cap_t cap)
             /* TODO: If bound thread is higher prio, switch to it directly (via switchToThread, rather than schedule). */
             tcb_t *tcb = (tcb_t *)notification_ptr_get_ntfnBoundTCB(ntfnPtr);
             /* Check if we are bound and that thread is waiting for a message */
+            /* Note that the benchmarks currently do not measure signalling a bound thread, so this path isn't taken in sel4bench */
             if (tcb) {
                 if (thread_state_ptr_get_tsType(&tcb->tcbState) == ThreadState_BlockedOnReceive) {
                     /* Send and start thread running */
@@ -113,6 +125,8 @@ void NORETURN fastpath_signal(cap_t cap)
             setThreadState(dest, ThreadState_Running);
             setRegister(dest, badgeRegister, badge);
 
+            /* It looks like the threads in sel4bench do have scheduling contexts,
+               so this path isn't taken in sel4bench */
             if (dest->tcbSchedContext == NULL) {
                 sched_context_t *sc = SC_PTR(notification_ptr_get_ntfnSchedContext(ntfnPtr));
                 if (sc != NULL && sc->scTcb == NULL) {
@@ -129,6 +143,7 @@ void NORETURN fastpath_signal(cap_t cap)
                 }
             }
 
+            /* It looks like this path is taken by the scheduler benchmarks */
             if (isSchedulable(dest)) {
                 possibleSwitchTo(dest);
             }
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -15,6 +15,131 @@
 #endif
 #include <benchmark/benchmark_utilisation.h>
 
+/* TODO: Add some checks from fastpath_call (e.g. valid VTable) */
+#ifdef CONFIG_ARCH_ARM
+static inline
+#ifndef CONFIG_ARCH_ARM_V6
+FORCE_INLINE
+#endif
+#endif
+void NORETURN fastpath_signal(cap_t cap)
+{
+    notification_t *ntfnPtr = NTFN_PTR(cap_notification_cap_get_capNtfnPtr(cap));
+    uint32_t ntfnState = notification_ptr_get_state(ntfnPtr);
+    word_t badge = cap_notification_cap_get_capNtfnBadge(cap);
+    switch (ntfnState) {
+        case NtfnState_Active: {
+            word_t badge2 = notification_ptr_get_ntfnMsgIdentifier(ntfnPtr);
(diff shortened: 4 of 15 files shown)


src/fastpath/fastpath.c
```diff
@@ -15,6 +15,268 @@
...
+    /* Lookup the cap */
+    cap_t cap = lookup_fp(TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbCTable)->cap, cptr);
+
+    /* Check it's a notification */
+    if (!cap_capType_equals(cap, cap_notification_cap)) {
+        slowpath(SysSend);
+    }
+
+    /* Check that we are allowed to signal this notification */
+    if (!unlikely(cap_notification_cap_get_capNtfnCanSend(cap))) {
```
oliver-wm, line 52: I think you would have to check the generated assembly to see if it's faster but it might be better to put all the unlikey blocks together.

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,268 @@
...
+    /* Get fault type. */
+    fault_type = seL4_Fault_get_seL4_FaultType(NODE_STATE(ksCurThread)->tcbFault);
+
+    /* Check there's no saved fault */
+    if (unlikely(fault_type != seL4_Fault_NullFault)) {
+        slowpath(SysSend);
+    }
+
+    /* We land up in here on every SYSCALL_SEND, so we need to check that we are
+     * actually signalling a notification */
```
oliver-wm, line 41: Can this be checked in the assembly before the call to this function?

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,268 @@
...
+
+    /* Lookup the cap */
+    cap_t cap = lookup_fp(TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbCTable)->cap, cptr);
```
oliver-wm, line 44: Possibly move this to top to avoid pipeline stalls after line 33.

oliver-wm: https://sjmulder.nl/dl/pdf/unsorted/2012%20-%20Blackham%20&%20Heiser%20-%20Correct,%20Fast,%20Maintainable,%20Choose%20Any%20Three.pdf

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,268 @@
...
+            notification_ptr_set_ntfnMsgIdentifier(ntfnPtr, badge2);
+            restore_user_context();
+        }
+        case NtfnState_Idle: {
+            /* TODO: If bound thread is higher prio, we should switch to it
+             * directly via switchToThread_fp; fastpath_restore; We shouldn't
+             * need to invoke the scheduler. */
+            tcb_t *tcb = (tcb_t *)notification_ptr_get_ntfnBoundTCB(ntfnPtr);
+            /* Check if we are bound and that thread is waiting for a message */
+            if (tcb) {
```
oliver-wm, line 78: What values can this take?

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,268 @@
...
+                if (thread_state_ptr_get_tsType(&tcb->tcbState) == ThreadState_BlockedOnReceive) {
+                    /* Send and start thread running */
+                    cancelIPC(tcb);
+                    setThreadState(tcb, ThreadState_Running);
```
oliver-wm, line 82: Can fastpath inline versions of these be defined in fastpath.h ?

oliver-wm (commented):
Is this only for MCS?

oliver-wm:
That's a start, I can continue review after you finish the TODO section. It's looking good :)

src/fastpath/fastpath.c
shanekadish: As in, can we check if we're signalling a notification (rather than sending a message to an endpoint) in the assembly?

As far as I know the way that we distinguish between whether a syscall is signalling a notification or sending to an endpoint is by checking the cap type (via `cap_capType_equals(cap, cap_notification_cap)`), since both syscalls use the `SYSCALL_SEND` code.

I think if we added a separate code for signalling a notification (e.g. `SYSCALL_SIGNAL`), then we could check in the assembly code whether we're signalling a notification. Would that be feasible?

oliver-wm: I don't think so, that would require changes to the syscall interface. Would need to discuss with others on that. @xurtis thoughts?

src/fastpath/fastpath.c
```diff
@@ -8,13 +8,242 @@
...
+    word_t fault_type;
+    dom_t dom;
+
+    /* Get fault type. */
+    fault_type = seL4_Fault_get_seL4_FaultType(NODE_STATE(ksCurThread)->tcbFault);
+
+    /* Check there's no saved fault */
+    if (unlikely(fault_type != seL4_Fault_NullFault)) {
+        slowpath(SysSend);
+    }
```
shanekadish, line 39: @xurtis do we need to check for saved faults in the signal fastpath?

src/fastpath/fastpath.c
oliver-wm: I think it's worth exploring adding a separate sycall code. @Xaphiosis can you comment on what effect that would that have proof wise?

shanekadish:
> Is this only for MCS?

Yep.

src/fastpath/fastpath.c
shanekadish: I'm now using the [`thread_state_ptr_set_tsType_np`](https://github.com/seL4/seL4/blob/master/include/fastpath/fastpath.h#L55-L58) defined in fastpath.h to set the thread state. Do you suggest that I also define an inline version of `cancelIPC` in fastpath.h?

src/fastpath/fastpath.c
shanekadish: As in what values can `notification_ptr_get_ntfnBoundTCB` return? I presume either `0` or a valid `tcb_t` address. Some other usages of `notification_ptr_get_ntfnBoundTCB` seem to use it the same way e.g. [here](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L80-L83) and [here](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L277-L281).

src/fastpath/fastpath.c
lsf37: That does not sound like a good idea -- not so much for the proof (although it would lead to duplication), but more because it would break the general concept that everything is a `send` to a cap.

Don't we have the same problem in the fast path for normal endpoints? What do we do there?

Xaphiosis: This might require a bit more thinking before even hitting verification. seL4 seems to have a design goal of not having many system calls, and selecting its actions by cap type instead. You'd have to introduce the new system call changing the syscall datatype, then move decodes and performs around... it probably will be a bunch of non-interesting work, spec shuffling, and result in some proof duplication. Then there's updating the security property proofs, capDL, and all the other proofs on the other verified architectures. I don't expect anything *exciting* to come up, but I've been wrong before.
For doing it on master, there might be no one available to deal with it in the near future.
As this looks MCS only, the people on that verification project would no doubt want to have a say. @corlewis is handling the verification team and can say more.

It would be nice to look at a prototype and diff in C that does what people actually want and evaluate verification cost from that before embarking on proof adventures.

Usual warning for fastpaths and changes: the most common issue we find during verification is that people think their fastpath does what the slow path does in the same circumstances, but it turns out to not be the case.

oliver-wm: Thanks, that gives something to think on.

shanekadish: > Don't we have the same problem in the fast path for normal endpoints? What do we do there?

I tried to use the same approach as what we do in [`fastpath_call`](https://github.com/seL4/seL4/blob/master/src/fastpath/fastpath.c#L51-L57), which is to lookup the cap and check that it's an endpoint (but in this case we check that it's a notification).

src/fastpath/fastpath.c
shanekadish: Super interesting paper :-). I've lifted the cap lookup so we now do it immediately getting the fault type.

src/fastpath/fastpath.c
shanekadish: Cool, done.

src/fastpath/fastpath.c
oliver-wm: Yeah I think so.

include/object/notification.h
```diff
@@ -36,4 +37,31 @@ static inline void maybeReturnSchedContext(notification_t *ntfnPtr, tcb_t *tcb)
 }
 #endif
 
+#ifdef CONFIG_KERNEL_MCS
```
oliver-wm, line 40: Need to add the `&& CONFIG_FASTPATH`

gernotheiser:
I think we should avoid doing this sort of thing for a single architecture only. While the code is in that arch-dependent part, the logic is arch-independent, and so is the need for this fastpath

lsf37:
> I think we should avoid doing this sort of thing for a single architecture only. While the code is in that arch-dependent part, the logic is arch-independent, and so is the need for this fastpath

You still have to start somewhere..

src/fastpath/fastpath.c
xurtis: I don't think so. Can you explain why we might?

include/object/notification.h
shanekadish: Is this referring to the definition of `maybeDonateSchedulingContext`? Isn't this used for the slowpath too though? I just moved [this definition from `notification.c`](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L41-L51) (where it didn't have `CONFIG_FASTPATH`) to `notification.h`.

src/fastpath/fastpath.c
shanekadish: I'm not super sure but I'm thinking that maybe we shouldn't check for saved faults here so that we don't unnecessarily delay the (possibly higher priority) waiter due to a fault from the signaller. If there are saved faults for the signaller we can handle them when the signaller is next scheduled to run. That's just my guess though. What do you think?

shanekadish:
> I think we should avoid doing this sort of thing for a single architecture only. While the code is in that arch-dependent part, the logic is arch-independent, and so is the need for this fastpath

Sure. I've now added what I think are the necessary signal fastpath hooks for x86, risc-v, and arm 64-bit architectures. I'm not able to test it out because I no longer have access to the CSIRO VPN and so can't access machine queue. (I've been using a borrowed sabre lite). @oliver-wm has kindly offered to test it out on the other architectures.

src/fastpath/fastpath.c
shanekadish: Cool, I've had a go at implementing a [`cancelIPC_fp`](https://github.com/shanekadish/seL4/blob/signal-fastpath/include/fastpath/fastpath.h#L146-L211) and [`cancelSignal_fp`](https://github.com/shanekadish/seL4/blob/signal-fastpath/include/fastpath/fastpath.h#L122-L144) in fastpath.h. Let me know what you think.

src/fastpath/fastpath.c
shanekadish: Let me know if this is okay to mark as resolved.

src/fastpath/fastpath.c
shanekadish: Let me know if this is okay to mark as resolved.

src/fastpath/fastpath.c
shanekadish: Let me know if this is okay to mark as resolved.

oliver-wm:
Hey @nomadeel would you be able to do a boopit on this pr?

oliver-wm:
@ssrg-bamboo `test`

ssrg-bamboo:

ssrg-bamboo:

include/object/notification.h
lsf37: If you've just moved it, it shouldn't have `CONFIG_FASTPATH`, yes.

config.cmake
```diff
@@ -248,6 +248,7 @@ config_string(
     UNQUOTE
 )
 config_option(KernelFastpath FASTPATH "Enable IPC fastpath" DEFAULT ON)
+config_option(KernelSignalFastpath SIGNAL_FASTPATH "Enable notification signal fastpath" DEFAULT OFF)
```
lsf37, line 251: If this is only tested with MCS, the config option needs to depend on MCS, otherwise people will think it is safe to use.

lsf37 (changes requested):
Before we merge this: please make sure only tested configurations are enabled. This probably works fine without MCS -- if that is confirmed by tests added to sel4test, it's fine to leave as is here (same for SMP), but if not, we should guard against people enabling this by accident until it is tested.

shanekadish:
> Before we merge this: please make sure only tested configurations are enabled. This probably works fine without MCS -- if that is confirmed by tests added to sel4test, it's fine to leave as is here (same for SMP), but if not, we should guard against people enabling this by accident until it is tested.

Currently this doesn't build without MCS. This is because we rely on functions like [`maybeDonateSchedContext`](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L41-L51) to (conditionally) perform the scheduling context donation in the signal fastpath, which is only defined for MCS. These functions were guarded with `ifdef CONFIG_KERNEL_MCS` before my changes and I'm not sure exactly how the logic would change for a non-MCS kernel. Same story with SMP. I've now made the `KernelSignalFastpath` option depend on `KernelIsMCS` and `${KernelNumDomains} EQUAL 1`.

config.cmake
shanekadish: Sure, done.

lsf37:
> Currently this doesn't build without MCS. This is because we rely on functions like [`maybeDonateSchedContext`](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L41-L51) to (conditionally) perform the scheduling context donation in the signal fastpath, which is only defined for MCS. These functions were guarded with `ifdef CONFIG_KERNEL_MCS` before my changes and I'm not sure exactly how the logic would change for a non-MCS kernel. Same story with SMP. I've now made the `KernelSignalFastpath` option depend on `KernelIsMCS` and `${KernelNumDomains} EQUAL 1`.

Cool. Is `${KernelNumDomains} EQUAL 1` strictly required? If this should go through verification at some point, we'll probably need `${KernelNumDomains}` unrestricted.

gernotheiser:
and it should really be tested on SMP. Not supporting non-MCS is fine

pingerino:
How does this relate to the fastpath for irqs I added? (never merged) -- https://github.com/pingerino/seL4/blob/phd/src/fastpath/fastpath.c#L421

gernotheiser:
Didn't even know that existed, would have been the next thing on my list. An IRQ fastpath would be highly desirable

pingerino:
@gernotheiser my branch has both. The signal fastpath is for when we don't context switch.

gernotheiser:
Right. @shanekadish's fastpath also works for signal to high prio (i.e. context switching)

pingerino:
@shanekadish can you comment on which scenarios you have fastpathed? If you are changing scheduling contexts does that mean you have fastpathed the sporadic server logic?

src/fastpath/fastpath.c
```diff
@@ -6,15 +6,269 @@
...
+            ksKernelEntry.is_fastpath = true;
+#endif
+
+        thread_state_ptr_set_tsType_np(&dest->tcbState, ThreadState_Running);
+        setRegister(dest, badgeRegister, badge);
+        if (ksCurThread->tcbPriority < dest->tcbPriority) {
+            /* switch to waiter immediately */
+            SCHED_ENQUEUE_CURRENT_TCB;
+            switchToThread_fp(dest, cap_pd, stored_hw_asid);
+            NODE_STATE(ksCurSC) = NODE_STATE(ksCurThread)->tcbSchedContext;
```
pingerino, line 260: where does the scheduling context get charged?

include/fastpath/fastpath.h
```diff
@@ -114,5 +119,96 @@ static inline reply_t *thread_state_get_replyObject_np(thread_state_t ts)
 }
 #endif
 
+static inline void cancelSignal_fp(tcb_t *tptr, notification_t *ntfnPtr)
```
pingerino, line 122: how different are these to the existing cancel functions? Is it really worth the duplication?

shanekadish:
> > Currently this doesn't build without MCS. This is because we rely on functions like [`maybeDonateSchedContext`](https://github.com/seL4/seL4/blob/master/src/object/notification.c#L41-L51) to (conditionally) perform the scheduling context donation in the signal fastpath, which is only defined for MCS. These functions were guarded with `ifdef CONFIG_KERNEL_MCS` before my changes and I'm not sure exactly how the logic would change for a non-MCS kernel. Same story with SMP. I've now made the `KernelSignalFastpath` option depend on `KernelIsMCS` and `${KernelNumDomains} EQUAL 1`.
> 
> Cool. Is `${KernelNumDomains} EQUAL 1` strictly required? If this should go through verification at some point, we'll probably need `${KernelNumDomains}` unrestricted.

No, not anymore. There were previously some build errors when building for SMP but I've fixed them now.

src/fastpath/fastpath.c
shanekadish: I'm not sure if/where it does currently. From what I understood this SC donation is similar to [how we do it in `fastpath_call`](https://github.com/seL4/seL4/blob/master/src/fastpath/fastpath.c#L184-L187), isn't it? Have I misunderstood something?

include/fastpath/fastpath.h
shanekadish: Not very different from what I understand. @oliver-wm suggested that I implement them [here](https://github.com/seL4/seL4/pull/322#discussion_r602050907).

src/fastpath/fastpath.c
shanekadish: Hm, might this happen in the call to [`maybeDonateSchedulingContext`](https://github.com/seL4/seL4/pull/322/files#diff-931125ed3a85619bdb93c0e1a4f69f62af97b5c815d986d0270d1cfb5a5209e5R188).

pingerino: When we donate a scheduling context, no charging happens, as we are not changing the scheduling context.

however, if we change scheduling context, we need to charge the previous one for the time that has passed (or we'll lose it). 

So far we don't fastpath anything that doesn't donate a scheduling context. I am not sure what the intent of this fastpath is. It needs to be some cut down version of the slowpath, likely the most simple form of charging a scheduling context (I *think* this is what I did in the irq fastpath on the branch I linked, but I don't have time to check now). 

Do you have a high level statement of what this fastpath should be fastpathing? Ideally its a small subset of things that happen on the slowpath.

include/fastpath/fastpath.h
pingerino: i recommend inlining the functions and adding `likely` to any branches. This will favour the fastpath. 

The only case for duplication is when you want to select a subset of the behaviour from a function that is too big to inline. 

perhaps I should write a 'so  you want to add a fastpath' guide :)

pingerino:
What is the critera for merging this PR? Is it going to be 'verified? verifiable? somewhat correct?'

We need to add some tests to make sure this fastpath is indeed a subset of slowpath behaviour. As it stands it will not charge scheduling contexts properly - it should be possible to write a test for this.

src/fastpath/fastpath.c
pingerino: My above statement is wrong. Even if we do `maybeDonateSchedulingContext` this donates from the notification object to the thread - we're still changing the current SC. So we have to charge.

My recommendation is you find the fastest path through switching scheduling context and fastpath that and only that.

lsf37:
> What is the critera for merging this PR? Is it going to be 'verified? verifiable? somewhat correct?'

From my perspective, to be merged, it needs to be at least verifiable and as correct as we can make it, i.e. behaviour must be equivalent to the slow path as far as we can tell without full verification. 

> We need to add some tests to make sure this fastpath is indeed a subset of slowpath behaviour. As it stands it will not charge scheduling contexts properly - it should be possible to write a test for this.

Yes, a test case like that would indeed be very good.

Also anything else we can think of that shows the behaviour is the same as the slow path. In some ways this would be an ideal candidate for fuzzing or at least random testing if we can set it up, because we have a fully executable correctness criterion.

pingerino:
> Also anything else we can think of that shows the behaviour is the same as the slow path. In some ways this would be an ideal candidate for fuzzing or at least random testing if we can set it up, because we have a fully executable correctness criterion.

As an aside, the fastpath is supposed to be a valid execution of the slowpath - couldn't we do something cool where we use the compiler to make sure the AST of the proposed fastpath is a subset of the slowpath? I am not a compiler person so maybe that is an insane but would be a really cool way to evaluate proposed fastpaths.

lsf37:
> > Cool. Is `${KernelNumDomains} EQUAL 1` strictly required? If this should go through verification at some point, we'll probably need `${KernelNumDomains}` unrestricted.
> 
> No, not anymore. There were previously some build errors when building for SMP but I've fixed them now.

SMP is less the issue for this one (but good if it works). `${KernelNumDomains}` is the number of security domains for the domain scheduler, i.e that can be > 1 even with SMP off.

The signal fast path would mostly make sense within a security domain, but the code must still compile for `${KernelNumDomains} > 1` and be equivalent to the slow path if we want to verify it. It'd be Ok to bail out to slow path on `${KernelNumDomains} > 1` generally, but it might be faster even for the `${KernelNumDomains} = 1` case to not have to check at runtime and "just work", and only bail out when it crosses domains (like in the IPC fast path).

It looks like you already have code for that (e.g. things like `dom = maxDom ? ksCurDomain : 0`), so it's mostly a question of making sure it works.

lsf37:
> > Also anything else we can think of that shows the behaviour is the same as the slow path. In some ways this would be an ideal candidate for fuzzing or at least random testing if we can set it up, because we have a fully executable correctness criterion.
> 
> As an aside, the fastpath is supposed to be a valid execution of the slowpath - couldn't we do something cool where we use the compiler to make sure the AST of the proposed fastpath is a subset of the slowpath? I am not a compiler person so maybe that is an insane but would be a really cool way to evaluate proposed fastpaths.

Matching sub-ASTs is probably too strong. Only the behaviour needs to be a real subset, the ASTs are allowed to be different, and they mostly will be, even if we partially evaluate the slow-path to remove the non-fastpath cases. It's close-ish, but not close enough for an automated check. The IPC fastpath proof has to talk a lot about re-orderings of instructions and clever "don't need to check this, because already follows from other things" that a pure AST comparison can't figure out.

ssrg-bamboo:

src/fastpath/fastpath.c
shanekadish: > Do you have a high level statement of what this fastpath should be fastpathing? Ideally its a small subset of things that happen on the slowpath.

At a high level, the intention of this fastpath is to fastpath signalling a lower priority thread, and signalling a higher (or equal) priority thread in cases where the signaller and signalled thread are in the same scheduling domain. You may find [the thesis report](https://drive.google.com/file/d/143Quc9KlXeYT6p6k2ask22zCkj-c9VTv/view?usp=sharing) helpful for a bit more context (in particular the flowchart at Figure 4.4, though now I realise that the 'Donate scheduling context from signaller to waiter' stage should rather say 'Donate scheduling context from _notification_ to waiter').

For my understanding; are you saying that in cases where we switch to the signalled thread, we need to charge the signaller's SC for the time elapsed? Based off your comment and a quick dig through the code it looks like [`SCHED_ENQUEUE_CURRENT_TCB`](https://github.com/shanekadish/seL4/blob/signal-fastpath/src/fastpath/fastpath.c#L258) does _not_ handle this charging logic, right? So we need to explicitly charge the signaller's SC before/after invoking that?

It's not clear to me where this charging is happening in the signal slowpath. Is it done in something downstream of [`possibleSwitchTo(dest)`](https://github.com/shanekadish/seL4/blob/signal-fastpath/src/object/notification.c#L116-L118)?

ssrg-bamboo:

lsf37:
Closing in favour of #793

PR #326, opened 2021-03-31
fastpath: some risc-v fastpath improvements

oliver-wm:
Inline fastpath functions and call out to them from traps.S.

### DIFF
--- a/include/arch/riscv/arch/fastpath/fastpath.h
+++ b/include/arch/riscv/arch/fastpath/fastpath.h
@@ -20,9 +20,11 @@
 void slowpath(syscall_t syscall)
 NORETURN;
 
+static inline
 void fastpath_call(word_t cptr, word_t r_msgInfo)
 NORETURN;
 
+static inline
 #ifdef CONFIG_KERNEL_MCS
 void fastpath_reply_recv(word_t cptr, word_t r_msgInfo, word_t reply)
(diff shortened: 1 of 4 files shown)


xurtis (approved):

oliver-wm:
@lsf37 ok with you from the proof perspective?

Xaphiosis:
Hard to tell at a glance. Obviously the riscv fastpath isn't currently verified.
Is this doing any checks other than "if user invoked reply_recv or call, invoke the relevant fastpath functions, which will fall over to slow path if conditions aren't right"?
It seems `CONFIG_FASTPATH` usage has disappeared.

If I were to go out on a limb, I think verification can handle it. Is this much different what happens on other platforms? I'm missing all the context on why this change is happening, so my questions might be a bit silly.

yyshen (approved):

oliver-wm:
Ah thanks @Xaphiosis I'll add the CONFIG_FASTPATH usage. It's not doing any other checks and I'm not sure it needs to. They will be handled from slowpath if conditions are not met.

lsf37:
A verification testboard should show up in the checks in the next few minutes.

lsf37:
Testboard passed, good to merge.

lsf37 (approved):

PR #331, opened 2021-04-04
CMake: rework declare_platform()

axel-h:
This is still work in progress, as it does not work for all platforms yet. But I'd be interested in some feedback, if this can be a way to simplify adding platforms and board in the future. It does more things internally, so there is some freedom to also refactor this in the future and do not expose it everywhere. There a lot less that must be done in actual platform CMake files. And there might more that can be removed in the future.


kent-mcleod:
@axel-h, are these changes motivated by finding that when adding new platforms it is hard to know what configuration settings are needed and that it is pretty easy to make errors when providing a new configuration?  Thus, the benefit of having one function to call is that you can expect the function to handle all of the configuration provided that you give it the correct parameters?

I'm not convinced that having one large function that attempts to handle all variability of different platform configurations is going to make it easier to add and manage platforms.  Having a single interface to call for declaring a platform would make it easier in the optimal case, but in situations where the implementation of this function causes a configuration that isn't appropriate for the new platform then the problem then I feel it would become much more complicated than it is now.  

In this case the person adding the new platform has to look into the implementation and figure out how to get the configuration they want while still preserving the correct configurations for all other platforms using the interface.  I would expect that this leads to lots of flags in the interface specifying different behavior options and ends up being less clear than what is there currently.  

I think the way that we currently factor out common configurations into utility functions such as declare_platform, declare_seL4_arch, declare_default_headers and add_sources already tries to make it easier to add and validate a platform configuration.  However over time new options or workarounds get added and these then would need to be factored into their own util functions once they start crowding out the rest of the config options.  

These configs are examples of platforms that still have pretty simple config files that are easy to understand:
- https://github.com/seL4/seL4/blob/b77c9b205193470b1802591d5ac8141d7c093b56/src/plat/hifive/config.cmake
- https://github.com/seL4/seL4/blob/b77c9b205193470b1802591d5ac8141d7c093b56/src/plat/fvp/config.cmake
- https://github.com/seL4/seL4/blob/b77c9b205193470b1802591d5ac8141d7c093b56/src/plat/tx2/config.cmake
These configs are getting too messy and unstructured 
- https://github.com/seL4/seL4/blob/b77c9b205193470b1802591d5ac8141d7c093b56/src/plat/exynos5/config.cmake
- https://github.com/seL4/seL4/blob/b77c9b205193470b1802591d5ac8141d7c093b56/src/plat/imx6/config.cmake

Instead of re-architecting how platform configuration is declared, here are some ways that the existing configurations could be cleared up a bit:
- KernelArmMach could be dropped from the kernel configuration and instead be provided by some helper in seL4_tools like with `correct_platform_strings`.  The common mach groupings only make sense for libraries like libplatsupport that are changed in lock-step with the kernel and can handle having arbitrary mach definitions.  EG why does exynos have a mach grouping but rpi doesn't?
- Board level subconfigurations within a single platform could also be removed.  If you are adding new imx6 boards then it could be a good idea to add them as new KernelPlatforms rather than sub platforms of imx6.
- Options like KernelArmCortexA9, and KernelArchArmV7a and several inputs to declare_default_headers could be queried from the device tree and automatically set.

axel-h:
Thanks for the feedback!

- Concerning RasPi, a kind of "mach" grouping is still on the to-do list actually. RasPi4 is (unfortunately) just a clone of RasPi at the moment.  I have some local patches for the platsupport libs that try to reduce the code duplication, but have not find the time to fully test them. And for the rest, it's on the to-do list to unify RasPi3 and Raspi3 as much as possible. 
- I'm looking for a way to make `KernelARMPlatform` more generic, as we need something like this on RISC-V also. But having a `KernelRISCVPlatform` seems odd, as it's basically the same concept on both architectures. Maybe I should try to refactor this first, so we have a  generic 'KernelArchPlatform' variable.
- I think there is a certain `Architecture -> SOC -> Board` chain, we just lack the board support so it gets more visible. There is something for imx6 and zynqmp, but as you said, imx6 a bit if a mess. This `KernelArmMach`somehow, but it could also become a list of "libs", so thing can be a bit more fine grained. That is a step towards using the device tree. I agree, you can get a lot from the device trees, but I wonder how long it will take if somebody implements this. My proposal could be an intermediate step to enforce more alignment for the things that can be aligned. It also should not replace things (I can improve thing there a bit), it just makes them easier to standard cases.
- Currently one can pick the `CAMKE_VAR` and `C_DEFINE` freely, my proposal prepares enforcing a naming scheme there so this can be set automatically and there is no room for error. All is based on the platform name you pick. 
- I also tried to enforce a naming scheme for the DTS file. So this can also be picked by default unless overwritten. 
- basically the Idea is really to enforce more defaults that tend to work across all existing platforms. There are a few tough corner cases, but I think there is a way to handle them, because they just need a bit of care. Or they remain real corner cases that need a lot of insight knowledge anyway. 
- I was hoping to eventually be able to drop `correct_platform_strings()`, as this feel a bit redundant. The information could be obtained from the kernel configs also. The other more radical option I'm considering is moving configs largely out of the kernel. What could remain there is just what the kernel really needs. But there is no longer any decision making there or even defining supported platforms. This would happen in a generic place e.g. where `correct_platform_strings()` currently is. So one could use this without a huge dependency on the kernel. It would just say to the kernel build "I need kernel config X".

kent-mcleod:
> Concerning RasPi, a kind of "mach" grouping is still on the to-do list actually. RasPi4 is (unfortunately) just a clone of RasPi at the moment. I have some local patches for the platsupport libs that try to reduce the code duplication, but have not find the time to fully test them. And for the rest, it's on the to-do list to unify RasPi3 and Raspi3 as much as possible.

If we linked in platsupport drivers based on compatibility strings in the device tree wouldn't this remove the need for "mach" grouping? Every platform that is part of the same "mach" group are going to have some devices that are common and some that aren't common with platforms in the same match.  The intersection of all devices that are common shrinks once more platforms get added to the same grouping. This quickly becomes harder to maintain as the two goals of avoiding breaking changes and easily adding new platforms without code duplication start fighting against each other and you get https://github.com/seL4/util_libs/blob/master/libplatsupport/CMakeLists.txt#L79 where the abstraction stops being useful.

> I think there is a certain Architecture -> SOC -> Board chain, we just lack the board support so it gets more visible. 

There's also revision number or board version and different hardware configurations (ie 2GB ram or 4GB ram versions) and then different firmware configurations (eg the RPI allowing the GPU to carve out different sizes of Ram for exclusive access). Additionally, a board may have more than one SOC or an SOC have more than one core with multiple options for seL4 to run on.  When we claim that seL4 is verified on a particular platform and a version + configuration of the kernel has passing proofs this is usually referring to a specific instance of all of these things and changing any of them may lead to no longer meeting the proof assumptions.  A device tree should hopefully capture all of this variability (currently we even already try and select which CPU that the kernel boots on https://github.com/seL4/seL4/blob/master/src/plat/exynos5/overlay-exynos5410.dts#L10) and so I wish that a KernelPlatform should refer to a particular unique device tree.  This also means that when you have a list of passing tests for a particular KernelPlatform it avoids mistakes when assuming that the kernel would still correctly run on that platform but under a different hardware revision.

In practice each KernelPlatform has a range of different hardware configurations that it is expected to run on and we don't do a good job of describing this.  Some platforms have unspecified set of build-time configuration arguments that change the hardware description that the kernel is being built for (like am335x, imx6 or exynos5 which take KernelArmPlatform to specify the board and changes the device tree).  Other platforms would require changing the device trees or sources manually to handle different hardware configurations (you would have to do this when changing the memory that the GPU uses on RPI for example).

> Currently one can pick the CAMKE_VAR and C_DEFINE freely, my proposal prepares enforcing a naming scheme there so this can be set automatically and there is no room for error. All is based on the platform name you pick.

We have this different naming scenario for all config options that are defined in CMake and also generated in a C header. It ended up like this because the C config names end up being part of the libsel4 API and enforcing a consistent mapping would cause a lot of breaking changes to existing config options that were created without a consistent naming scheme.  However it seems like a good idea to update the `config_*` cmake macros to support defining config options with a single name and automatically picking a C translated name and then using this for new config definitions.

> The other more radical option I'm considering is moving configs largely out of the kernel. 

I think being able to provide out of tree configurations that get included here (https://github.com/seL4/seL4/blob/master/configs/seL4Config.cmake#L195) is a good idea and I've wanted to do it in the past.  While it would be pretty easy to do technically, I'm not sure how it would work with the seL4 Foundation licensing and trademark requirements as this would make it harder to assess how the kernel sources have been modified.

PR #333, opened 2021-04-04
support PRIu64 and SEL4_PRIu_word in kernel

axel-h:
support PRIu64 and SEL4_PRIu_word in kernel and use it

### DIFF
--- a/include/basic_types.h
+++ b/include/basic_types.h
@@ -9,6 +9,20 @@
 #include <stdint.h>
 #include <arch/types.h>
 
+/* arch/types.h is supposed to define word_t and _seL4_word_fmt */
+#ifndef _seL4_word_fmt
+#error "missing _seL4_word_fmt"
+#endif
+
+#define _macro_str_concat_helper2(x)    #x
+#define _macro_str_concat_helper1(x,y)  _macro_str_concat_helper2(x ## y)
+#define _macro_str_concat(x,y)          _macro_str_concat_helper1(x,y)
+
+#define SEL4_PRIu_word  _macro_str_concat(_seL4_word_fmt, u)
+#define SEL4_PRIx_word  _macro_str_concat(_seL4_word_fmt, x)
+#define SEL4_PRI_word   SEL4_PRIu_word
+
+
 enum _bool {
     false = 0,
     true  = 1
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -2440,7 +2440,7 @@ void kernelDataAbort(word_t pc)
            getFAR(), getDFSR());
     halt();
 }
-#endif
+#endif /* CONFIG_DEBUG_BUILD */
 
 #ifdef CONFIG_PRINTING
 typedef struct readWordFromVSpace_ret {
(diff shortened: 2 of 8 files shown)


include/basic_types.h
```diff
@@ -9,6 +9,20 @@
 #include <stdint.h>
 #include <arch/types.h>
 
+/* arch/types.h is supposed to define word_t and _seL4_word_fmt */
+#ifndef _seL4_word_fmt
```
kent-mcleod, line 13: Have you tried just including <sel4/basic_types.h> from kernel source files?  libsel4 include directories are in the kernel include path.  Some of them aren't supported for inclusion in the kernel, but basic_types.h may work.

axel-h: ok, will give this a try since nobody has raised an objection in the last 4 days.

bennoleslie: This is probably not the best place to discuss, however, is include seL4 library header files in the kernel a good practise?

When first looking at the kernel I found it surprising, and made things more difficult to understand and debug. I guess is it something we really want to continue encouraging or eventually try to phase out?

axel-h: I don't really have an opinion here, but so far this seems pretty much separated. Obviously, merging this PR without change would be the easiest thing - and then discuss further steps. I've seen you also did some refactoring in your branches on the repo, which may eventually have some impact on this discussion?

bennoleslie: Yeah, the comment wasn't intended to say this shouldn't be merged, just to spark thinking.

include/basic_types.h
```diff
@@ -9,6 +9,20 @@
...
+#erro "missing _seL4_word_fmt"
```
bennoleslie, line 14: Should be `#error`

src/arch/arm/64/kernel/vspace.c
```diff
@@ -2426,27 +2426,21 @@ void kernelDataAbort(word_t pc) VISIBLE;
...
-    printf("Faulting instruction: 0x%lx\n", (unsigned long)pc);
-    printf("FAR: 0x%lx ESR (DFSR): 0x%x\n", (unsigned long)far, (unsigned int)dfsr);
-
+    printf("Faulting instruction: 0x%"SEL4_PRIx_word"\n", pc);
+    printf("FAR: 0x%"SEL4_PRIx_word" ESR (DFSR): 0x%"SEL4_PRIx_word"\n",
+           getFAR(), getDFSR());
     halt();
 }
-#endif
+#endif /* CONFIG_DEBUG_BUILD */
```
bennoleslie, line 2443: I'm not sure the overall style (or overall policy on updating style in the middle of other fixes).

I don't have a strong feeling either way as long as it is consistent.

bennoleslie (commented):
Overall looks good to me. Some very minor comments.

src/arch/arm/64/kernel/vspace.c
axel-h: moved to separate commit

axel-h:
Could this be merged? And the remaining comments are a follow-up action on later then.

kent-mcleod (approved):

PR #337, opened 2021-04-06
Rename MAX_BUDGET to MAX_PERIOD

xurtis:
As this variable bounds both the period and the budget and the period itself bounds the budget, the name for this variable would be more appropriately named 'MAX_PERIOD'.

@michaelmcinerney originally requested this change.


config.cmake
```diff
@@ -454,7 +454,7 @@ config_string(
 )
 
 config_string(
-    KernelStaticMaxBudgetUs KERNEL_STATIC_MAX_BUDGET_US
+    KernelStaticMaxPeriodUs KERNEL_STATIC_MAX_PERIOD_US
```
pingerino, line 457: the effort to impact ration of this change seems very poor. But if it's already done...

pingerino (approved):

michaelmcinerney (approved):

config.cmake
xurtis: Micheal suggested it from the side of verification, I normally wouldn't suggest a change of this kind at this point.

kent-mcleod (approved):
This is effectively an API breaking change if the configuration API is considered part of the API?

PR #338, opened 2021-04-06
Current thread must not yield to multiple threads

xurtis:
We must ensure that the current thread is not already yielding to another thread when it makes a YieldTo invocation. This should never happen as we should always cancel any YieldTo before returning to user-level, however we cannot prove this invariant at this time.

@MitchellBuckley and @michaelmcinerney may have some input on the need for this change.

### DIFF
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -228,6 +228,16 @@ static exception_t decodeSchedContext_YieldTo(sched_context_t *sc, word_t *buffe
         return EXCEPTION_SYSCALL_ERROR;
     }
 
+    // This should not be possible as the currently running thread
+    // should never have a non-null yieldTo, however verifying this
+    // invariant is being left to future work.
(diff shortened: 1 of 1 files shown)


src/object/schedcontext.c
```diff
@@ -229,6 +229,15 @@ static exception_t decodeSchedContext_YieldTo(sched_context_t *sc, word_t *buffe
         return EXCEPTION_SYSCALL_ERROR;
     }
 
+    if (NODE_STATE(ksCurThread)->tcbYieldTo != NULL) {
+        // This should not be possible as the currently running thread
```
pingerino, line 233: can you also add a assert in case it does happen? Since we can't prove it currently we want to know if we hit it in a test, in a way that fails more obviously than a user error message.

pingerino (changes requested):

src/object/schedcontext.c
xurtis: Done

kent-mcleod (approved):

PR #339, opened 2021-04-06
mcs: removed donator from release queue

xurtis:
If one thread replies on behalf of the a thread to which an SC is donated and that thread is in the release queue, that thread must be removed from the release queue when the SC is returned.

### DIFF
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -333,6 +333,7 @@ void schedContext_donate(sched_context_t *sc, tcb_t *to)
     if (from) {
         SMP_COND_STATEMENT(remoteTCBStall(from));
         tcbSchedDequeue(from);
+        tcbReleaseRemove(from);
         from->tcbSchedContext = NULL;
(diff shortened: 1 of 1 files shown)


src/object/schedcontext.c
```diff
@@ -333,6 +333,7 @@ void schedContext_donate(sched_context_t *sc, tcb_t *to)
     if (from) {
         SMP_COND_STATEMENT(remoteTCBStall(from));
         tcbSchedDequeue(from);
+        tcbReleaseRemove(from);
```
pingerino, line 336: Is this another 'we can't prove it's impossible' scenario? Or is there a code path that can get us here?

xurtis: If a passive thread receives a donated SC and then ends up in the release queue, it's possible for a different thread to perform a send on the reply object at the head of the call stack. This would donate the SC back up the call stack, but without this change the TCB that initially received the SC would end up in the release queue without an SC attached.

I believe this was a genuine bug caught by verification.

kent-mcleod (approved):

pingerino (approved):

PR #341, opened 2021-04-06
mcs: move maybeReturnSc to end of receiveSignal

xurtis:
This should not change the behavior of this function in any way but makes the order of operations more similar to other locaitons in the code in a manner that eases verification.

Original PR by @michaelmcinerney

### DIFF
--- a/src/object/notification.c
+++ b/src/object/notification.c
@@ -210,9 +210,6 @@ void receiveSignal(tcb_t *thread, cap_t cap, bool_t isBlocking)
                                         ThreadState_BlockedOnNotification);
             thread_state_ptr_set_blockingObject(&thread->tcbState,
                                                 NTFN_REF(ntfnPtr));
-#ifdef CONFIG_KERNEL_MCS
-            maybeReturnSchedContext(ntfnPtr, thread);
-#endif
             scheduleTCB(thread);
 
             /* Enqueue TCB */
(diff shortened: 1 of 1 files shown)


src/object/notification.c
```diff
@@ -187,6 +184,10 @@ void receiveSignal(tcb_t *thread, cap_t cap, bool_t isBlocking)
 
             notification_ptr_set_state(ntfnPtr, NtfnState_Waiting);
             ntfn_ptr_set_queue(ntfnPtr, ntfn_queue);
+
+#ifdef CONFIG_KERNEL_MCS
+            maybeReturnSchedContext(ntfnPtr, thread);
+#endif
```
xurtis, line 190: I think the need for the change is the fact that this can move an SC and its easier to show that the invariants still hold after the updates to the notification queue.

src/object/notification.c
```diff
@@ -187,6 +184,10 @@ void receiveSignal(tcb_t *thread, cap_t cap, bool_t isBlocking)
 
             notification_ptr_set_state(ntfnPtr, NtfnState_Waiting);
             ntfn_ptr_set_queue(ntfnPtr, ntfn_queue);
+
+#ifdef CONFIG_KERNEL_MCS
+            maybeReturnSchedContext(ntfnPtr, thread);
```
pingerino, line 189: This should alter the behaviour of  the call to `scheduleTCB`? I don't see how that is ok.

xurtis: Regardless of whether `scheduleTCB` is called before or after `maybeReturnSchedContext`, the call to `scheduleTCB` within `maybeReturnSchedContext` will always return false as the thread is always `ThreadState_BlockedOnNotification` at that point. As such, this particular call to `scheduleTCB` doesn't depend on whether the thread in question has a bound SC or the state of any bound SC.

pingerino (approved):

kent-mcleod (approved):

PR #344, opened 2021-04-08
risc-v: reorganise traps.S

oliver-wm:
Move syscall and fastpath checks after interrupt and exceptions.
Exceptions were being interpreted as null-syscalls.
Remove duplication for return address.

### DIFF
--- a/src/arch/riscv/traps.S
+++ b/src/arch/riscv/traps.S
@@ -98,49 +98,35 @@ trap_entry:
   csrr x1,  sepc
   STORE   x1, (33*REGBYTES)(t0)
 
-#ifdef CONFIG_FASTPATH
-  li t3, SYSCALL_REPLY_RECV
-  beq a7, t3, fp_reply_recv
-
-  li t3, SYSCALL_CALL
-  beq a7, t3, fp_call
-#endif
-
   /* Check if it's an interrupt */
   bltz s0, interrupt
 
-  li   s4, 8       /* ratified priv has value 8 for ecall exception */
+  /* ratified priv has value 8 for ecall from U-mode exception */
+  li   s4, 8
   bne  s0, s4, exception
 
-syscall:
+handle_syscall:
   /* Set the return address to sepc + 4 in the case of a system/environment call */
   addi x1, x1, 4
   /* Save NextIP */
   STORE   x1, (34*REGBYTES)(t0)
 
+#ifdef CONFIG_FASTPATH
+  li t3, SYSCALL_CALL
+  beq a7, t3, c_handle_fastpath_call
+
+  li t3, SYSCALL_REPLY_RECV
+#ifdef CONFIG_KERNEL_MCS
+  /* move reply to 3rd argument */
+  mv a2, a6
+#endif
+  beq a7, t3, c_handle_fastpath_reply_recv
+#endif
+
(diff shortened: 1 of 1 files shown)


oliver-wm:
This fixes the failing simulation and hardware builds on sel4test

src/arch/riscv/traps.S
```diff
@@ -98,50 +98,43 @@ trap_entry:
...
-
-  li t3, SYSCALL_CALL
-  beq a7, t3, fp_call
-#endif
-
   /* Check if it's an interrupt */
   bltz s0, interrupt
 
-  li   s4, 8       /* ratified priv has value 8 for ecall exception */
+  /* ratified priv has value 8 for ecall exception */
```
axel-h, line 104: maybe be more precise and say: "for 'ecall from U-mode' exception"

src/arch/riscv/traps.S
```diff
@@ -98,50 +98,43 @@ trap_entry:
...
+  li   s4, 8
   bne  s0, s4, exception
 
-syscall:
+  /* If it's not an exception or an interrupt it's a syscall */
```
axel-h, line 108: That comment is a bit misleading actually, better simply say "handle syscall" now.

src/arch/riscv/traps.S
```diff
@@ -98,50 +98,43 @@ trap_entry:
...
+syscall:
+
+  /* move syscall number to 3rd argument */
   mv a2, a7
 
   j c_handle_syscall
 
 fp_reply_recv:
 
-  addi x1, x1, 4
```
axel-h, line 127: What is the commit policy actually concerning trivial changes, should removing this be a separate commit?

axel-h:
change looks ok to me.

src/arch/riscv/traps.S
```diff
@@ -98,50 +98,43 @@ trap_entry:
...
+#endif
+
+syscall:
+
+  /* move syscall number to 3rd argument */
   mv a2, a7
 
   j c_handle_syscall
 
 fp_reply_recv:
```
yyshen, line 129: why not combine this to the code above, just as the c_handle_fastpath_call?

src/arch/riscv/traps.S
```diff
@@ -98,50 +98,43 @@ trap_entry:
...
-  addi x1, x1, 4
-  /* Save NextIP */
-  STORE   x1, (34*REGBYTES)(t0)
-
 #ifdef CONFIG_KERNEL_MCS
+  /* move reply to 3rd argument */
   mv a2, a6
 #endif
 
   j c_handle_fastpath_reply_recv
```
yyshen, line 136: will this cause a compiler error when FASTPATH is not enabled?

oliver-wm: Yes, you're right.

src/arch/riscv/traps.S
oliver-wm: I'll do that.

oliver-wm:
@yyshen happy with it?

xurtis (approved):

yyshen:
> @yyshen happy with it?

Yes, I was about to click the approve button.

oliver-wm:
Thanks!

PR #350, opened 2021-04-16
Merge memory regions before checking for reserved areas

bvirfollet:
As mentioned [here](https://sel4.discourse.group/t/elfloader-memory-mapping-boundaries), contiguous regions can lead to memory fragmentation. Merging those before checking for reserved area would prevent it.
Signed-off-by: Bertrand Virfollet <bvirfollet@silicom.fr>

### DIFF
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -26,6 +26,38 @@ def get_memory_regions(tree: FdtParser):
     return regions
 
 
+def merge_memory_regions(regions: Set[Region]) -> Set[Region]:
+    ''' Check all region and merge adjacent ones '''
+    all_regions = []
+    idx = 0
+    for region in regions:
+        all_regions.append(dict(idx=idx, region=region, right_adj=None, left_adj=None))
+        idx += 1
+
+    # Find all right contiguous regions
+    for dreg in all_regions:
+        for dnreg in all_regions[dreg['idx']+1:]:
+            if dreg['region'].owner == dnreg['region'].owner:
+                if dnreg['region'].base == dreg['region'].base + dreg['region'].size:
+                    dreg['right_adj'] = dnreg
+                    dnreg['left_adj'] = dreg
+                elif dreg['region'].base == dnreg['region'].base + dnreg['region'].size:
+                    dnreg['right_adj'] = dreg
+                    dreg['left_adj'] = dnreg
+
+    # Find all the left-most contiguous regions
+    contiguous_regions = set()
+    for reg in all_regions:
+        if reg['left_adj'] is None:
+            size = reg['region'].size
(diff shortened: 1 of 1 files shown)


tools/hardware/utils/memory.py
```diff
@@ -26,6 +26,38 @@ def visitor(node: WrappedNode):
     return regions
 
 
+def merge_memory_regions(regions: Set[Region]) -> Set[Region]:
+    ''' Check all region and merge adjacent ones '''
+    all_regions = []
+    idx = 0
+    for region in regions:
```
oliver-wm, line 33: prefer enumerate here

tools/hardware/utils/memory.py
```diff
@@ -26,6 +26,38 @@ def visitor(node: WrappedNode):
...
+        all_regions.append(dict(idx=idx, region=region, right_adj=None, left_adj=None))
+        idx += 1
+
+    # Find all right contiguous regions
+    for dreg in all_regions:
+        for dnreg in all_regions[dreg['idx']+1:]:
```
oliver-wm, line 39: This assume that all_regions is sorted?

tools/hardware/utils/memory.py
bvirfollet: What do you mean? Get everything done in this loop?

oliver-wm:
```suggestion
    for idx, region in enumerate(regions):
```

oliver-wm: Then you don't need to worry about incrementing or declaring the idx variable

bvirfollet: Yeap, sorry I'm a bit old school on this :)

tools/hardware/utils/memory.py
bvirfollet: It seems important to keep separated the device memory regions and untyped memory regions. All memory regions are inspected, but contiguity is considered in respect to region owner. And as a result, a memory region spanning over 2 buses would not get merged.

kent-mcleod (approved):

nomadeel (approved):

nomadeel:
Looks good, but I had a look at the commits and there's a merge commit. If possible, could you remove that commit and use rebase instead? So for your case, `git rebase origin/master` where origin is upstream.

lsf37:
If the merge commit is from GitHub's `update branch` button, it will disappear if you do a "Rebase and merge".

lsf37:
What is failing and will prevent the merge is `gitlint`, though (the title of the commit messages needs to be < 50 characters)

oliver-wm (approved):

lsf37:
Merged via #562

PR #358, opened 2021-05-04
SELFOUR-442: only lookup IPC buffer if required

xurtis:
Prior to this change the kernel would always lookup the IPC
buffer on the slowpath, even though this is only required
for ipcs and invocations larger than n_msgRegsiters or
with extra caps.

This change increases slowpath performance by 5% on ia32
and 6% on imx6.

Moving PR for [SELFOUR-442](https://sel4.atlassian.net/browse/SELFOUR-442) to GitHub


include/object/tcb.h
```diff
@@ -217,7 +217,7 @@ exception_t invokeTCB_NotificationControl(tcb_t *tcb, notification_t *ntfnPtr);
 cptr_t PURE getExtraCPtr(word_t *bufferPtr, word_t i);
 void setExtraBadge(word_t *bufferPtr, word_t badge, word_t i);
 
-exception_t lookupExtraCaps(tcb_t *thread, word_t *bufferPtr, seL4_MessageInfo_t info);
+exception_t lookupExtraCaps(tcb_t *thread, word_t *bufferPtr, word_t length);
```
zaklogician, line 220: Naming sanity check, purely for pragmatics:

I'm right that `length` indeed happens to be the length of `bufferPtr`, right? Reading the loop starting at  `src/object/tcb.c:395`, this seems to be the case.

 I'd find it jarring to have a buffer and length argument, when length is not buffer's length.

src/api/syscall.c
```diff
@@ -396,20 +396,26 @@ static exception_t handleInvocation(bool_t isCall, bool_t isBlocking)
...
-    buffer = lookupIPCBuffer(false, thread);
-
-    status = lookupExtraCaps(thread, buffer, info);
-
-    if (unlikely(status != EXCEPTION_NONE)) {
-        userError("Lookup of extra caps failed.");
-        if (isBlocking) {
-            handleFault(thread);
+    buffer = NULL;
+    length = seL4_MessageInfo_get_length(info);
```
zaklogician, line 400: No such thing as a zero-length buffer, right? If length is zero, the pointer is NULL.
Otherwise, `src/api/syscall.c:419` semantics would change.

zaklogician:
I haven't seen CRefine in a good while, so if I'm asked to comment on how this impacts verification, I can't give a high-certainty answer. What I could do: compare with existing code that I know is fine as far as verification is concerned. The new code doesn't seem to do anything too unusual, so I'd say it's probably good to go.

See my comment about naming, and my possibly dumb question in `src/api/syscall.c`.

lsf37:
If it's semantics-preserving, which seems to be the intention, then I don't think this would take very long.

Maybe 1-2 full-time weeks for someone with CRefine experience. Two possible approaches: prove that it's an optimisation, i.e. only change CRefine, or reflect it up into ASpec. Second option is easer, but larger. Might well be faster.

You could also make the same cut (optimisation vs pulling it up) on the Haskell level, which might be a bit easier. If you go for optimisation, I'd probably recommend that.

include/object/tcb.h
pingerino: Nope, it's the length of the message, which determines if we need the ipc buffer or not.

oliver-wm (approved):

zaklogician (approved):

lsf37 (changes requested):
This PR is failing simulation tests for breakpoints on x86

PR #362, opened 2021-05-14
risc-v: fix register bug

oliver-wm:
An aditional two parameters for the dtb were added to the init_kernel
function call in the elfloader. The corresponding _start entry
in the kernel was never updated.

### DIFF
--- a/src/arch/riscv/head.S
+++ b/src/arch/riscv/head.S
@@ -17,8 +17,8 @@
 .extern trap_entry
 
 /*
- * When SMP is enabled, the elfloader passes the hart ID in a4
- * and logical core ID in a5.
+ * When SMP is enabled, the elfloader passes the hart ID in a6
+ * and logical core ID in a7.
  */
 _start:
   fence.i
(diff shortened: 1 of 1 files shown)


src/arch/riscv/head.S
```diff
@@ -17,8 +17,8 @@
 .extern trap_entry
 
 /*
- * When SMP is enabled, the elfloader passes the hart ID in a4
- * and logical core ID in a5.
+ * When SMP is enabled, the elfloader passes the hart ID in a6
```
yyshen, line 20: You mention one additional is added; why do we skip registers from a4,a5 to a6, a7. Also, the commit title is kind of misleading since this is an omission when changing the elfloader I think.

xurtis (approved):

src/arch/riscv/head.S
oliver-wm: Sorry it was two. `(paddr_t) dtb` and  `dtb_size`

PR #363, opened 2021-05-17
Some python improvments

oliver-wm:
Add memoization to umm.py.
Consolidate logic for risc-v and arm memory base.

### DIFF
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: GPL-2.0-only
 #
+from typing import List, Set
+from hardware.memory import Region
 
 
 class Config:
@@ -29,6 +31,15 @@ class Config:
         ''' Get page size in bits for mapping devices for this arch '''
         return self.get_page_bits()
 
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' Given a set of regions, sort them and align the first so that the
+        ELF loader will be able to load the kernel into it. Will return the
+        aligned memory region list, a set of any regions of memory that were
+        aligned out and the physBase value that the kernel will use. memory
+        region list, a set of any regions of memory that were aligned out and
+        the physBase value that the kernel will use. '''
+        pass
+
 
 class ARMConfig(Config):
     ''' Config class for ARM '''
@@ -38,6 +49,20 @@ class ARMConfig(Config):
     def get_kernel_phys_align(self) -> int:
         ''' on ARM the ELF loader expects to be able to map a supersection page to load the kernel. '''
         return self.SUPERSECTION_BITS
+
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' Arm wants physBase to be the physical load address of the kernel. '''
+        ret = sorted(regions)
+        extra_reserved = set()
+
+        new = ret[0].align_base(self.get_kernel_phys_align())
+        resv = Region(ret[0].base, new.base - ret[0].base, None)
+        extra_reserved.add(resv)
+        ret[0] = new
+
+        physBase = ret[0].base
+
+        return ret, extra_reserved, physBase
 
 
 class RISCVConfig(Config):
@@ -49,6 +74,21 @@ class RISCVConfig(Config):
         ''' on RISC-V OpenSBI is loaded at the start
         of physical memory. Mark it as unavailable. '''
         return self.MEGA_PAGE_SIZE
+
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' Currently the RISC-V port expects physBase to be the address that the
+        bootloader is loaded at. To be generalised in the future. '''
+        ret = sorted(regions)
+        extra_reserved = set()
+
+        physBase = ret[0].base
+
+        resv = Region(ret[0].base, self.get_bootloader_reserve(), None)
+        extra_reserved.add(resv)
+        ret[0].base += self.get_bootloader_reserve()
+        ret[0].size -= self.get_bootloader_reserve()
+
+        return ret, extra_reserved, physBase
 
     def get_device_page_bits(self) -> int:
         ''' Get page size in bits for mapping devices for this arch '''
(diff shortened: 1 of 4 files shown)


oliver-wm:
This may or may not be needed, just had them in a working directory and thought it could be useful.

tools/hardware/utils/memory.py
```diff
@@ -61,11 +61,6 @@ def align_memory(regions: Set[Region], config: Config) -> List[Region]:
     ret = sorted(regions)
     extra_reserved = set()
 
-    if config.arch == 'riscv':
-        # RISC-V is special: it expects physBase to be
-        # the address that the bootloader is loaded at.
-        physBase = ret[0].base
```
kent-mcleod, line 67: How come this doesn't need to be set before the following two if statements anymore? It looks like it still needs separate treatment compared to Arm.

oliver-wm: Not sure this was from a month or so ago. It does look like those ifs are needed, will remove.

oliver-wm: I've refactored align_memory into the config class. I think this makes it much clearer as to what is going on.

axel-h: According to @kent-mcleod, this ist kind of RISC-V hack that https://github.com/seL4/seL4/pull/365 will resolve then.

lsf37 (approved):
This now looks fine to me (style format still needs to be run). Any concerns merging this @kent-mcleod ?

lsf37:
(style fix and rebased)

axel-h:
I'm fin with these changes if they have been tested and work. I have https://github.com/seL4/seL4/pull/365 with more cleanup that I will rebase on top of this then and test that this works with out RISC-V changes

tools/hardware/irq.py
```diff
@@ -185,6 +185,7 @@ def parse_irq(self, child, data):
     'qcom,msm-qgic2': ArmGic,
     'ti,am33xx-intc': RawIrqController,
     'ti,omap3-intc': RawIrqController,
+    'riscv,plic0': RawIrqController,
```
axel-h, line 188: We could make this a separate commit, because it is really separate from the other changes

tools/umm.py
```diff
@@ -76,9 +76,19 @@ def base_name(x):
     return x[1]
 
 # This assumes that membership is a DAG which is the case in C
-# We could memoize, doesn't seem worth it ...
 
 
+def memoize(func):
```
axel-h, line 81: So is this worth it now or not?  And when we add this, the comment "# This assumes that membership is a DAG which is the case in C" should be moved down where `paths_to_type` is declared.

lsf37: My initial thought was it doesn't hurt, but you are right, memoization is complete overkill here, the runtime of the entire program is already insignificant, and it just adds unneeded complexity. So, I'll remove it and the comment.

tools/hardware/config.py
```diff
@@ -29,6 +31,12 @@ def get_device_page_bits(self) -> int:
         ''' Get page size in bits for mapping devices for this arch '''
         return self.get_page_bits()
 
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' Given a set of regions, sort them and align the first so that the ELF loader will be able to load the kernel into it. Will return the aligned
```
axel-h, line 35: Break the line to improve readbility:
```suggestion
        ''' Given a set of regions, sort them and align the first so that the 
        ELF loader will be able to load the kernel into it. Will return the
        aligned memory region list, a set of any regions of memory that were
        aligned out and the physBase value that the kernel will use. '''
```

tools/hardware/config.py
```diff
@@ -39,6 +47,20 @@ def get_kernel_phys_align(self) -> int:
         ''' on ARM the ELF loader expects to be able to map a supersection page to load the kernel. '''
         return self.SUPERSECTION_BITS
 
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' Arm wants physBase to be the physical load address of the kernel. '''
```
axel-h, line 51: "Arm" -> "ARM"?
```suggestion
        ''' ARM wants physBase to be the physical load address of the kernel. '''
```

tools/hardware/config.py
```diff
@@ -50,6 +72,20 @@ def get_bootloader_reserve(self) -> int:
         of physical memory. Mark it as unavailable. '''
         return self.MEGA_PAGE_SIZE
 
+    def align_memory(self, regions: Set[Region]) -> List[Region]:
+        ''' RISC-V expects physBase to be the address that the bootloader is loaded at. '''
```
axel-h, line 76: I have some doubt that this is really what RISC-V wants. This just seem the be something that is currently there due to the history of the RISC-V port - and it feels wrong that we make assumption about the bootloader. If you plan to merge this PR now, it is at least helpful bring some light on the current implementation. But the comment could say this is not the intended long term solution because this creates a nasty dependency on the botloader and how big it is. The other PR (https://github.com/seL4/seL4/pull/365) I'm working at will remove all this code again and use DTB overlays to reserve bootloader memory and align things with how it's done on ARM.

tools/hardware/config.py
lsf37: Arm is the correct spelling these days.

tools/hardware/config.py
lsf37: Happy to add that

tools/hardware/config.py
lsf37: It's annoying that they changed it. I'm not really keen on replacing it everywhere in a PR like this, using the old one seems wrong, and using the new one for new stuff looks inconsistent. Using the latter now and at some point doing a PR that systematically replaces ARM with Arm is probably the way to go.

kent-mcleod: We normally choose to update it if the code is being touched.  This is the same guidelines we follow for style changes too.

lsf37: That makes sense

PR #364, opened 2021-05-17
ia32: Fix linker errors with capDL printing

nomadeel:
This is a quick bug-fix to fix a red build in CI.

Although the functions aren't implemented properly, this is a stop gap
issue to prevent the linker from raising problems about missing
references to various platform-specific capDL printing functions.

Signed-off-by: Damon Lee <Damon.Lee@data61.csiro.au>

### DIFF
--- a/src/arch/x86/32/machine/capdl.c
+++ b/src/arch/x86/32/machine/capdl.c
@@ -1,6 +1,6 @@
 /*
  * Copyright 2021, Axel Heider <axelheider@gmx.de>
- * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
  *
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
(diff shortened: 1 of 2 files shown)


oliver-wm (approved):

src/arch/x86/32/machine/capdl.c
```diff
@@ -12,6 +12,46 @@
...
 #include <arch/machine/capdl.h>
 
+word_t get_tcb_sp(tcb_t *tcb)
+{
+    return tcb->tcbArch.tcbContext.registers[ESP];
+}
+
+void print_ipc_buffer_slot(tcb_t *tcb)
+{
+#ifdef CONFIG_PRINTING
```
axel-h, line 22: is this define really necessary? `printf()` should be defined to nothing if `CONFIG_PRINTING this is not set. And in one of my other PR the feedback was that in this cases there should be no define guard around this then.

lsf37: I think Axel is right, this is already outside the verification build so can't disturb anything there, and `printf` itself is `#ifdef`'d, so we should drop all the `#ifdef CONFIG_PRINTING` in this patch.

nomadeel: Sounds good.

nomadeel:
Dropped all the `#ifdef CONFIG_PRINTING`s in the previous commit.

axel-h (approved):

nomadeel:
Got approval to merge without passing all checks from discussion outside here with @lsf37.

lsf37:
> Got approval to merge without passing all checks from discussion outside here with @lsf37.

Something funky is going on with the preprocess check in general, I'm looking into it.

PR #366, opened 2021-05-18
trivial: Fix semicolon error and missing symbol

nomadeel:


oliver-wm (approved):

nomadeel:
Similar to #364, we're ignoring the preprocess checks for now.

PR #368, opened 2021-05-26
pc99: Remove PT_PHDR segment from linker script

kent-mcleod:
This is to work around an issue in a variant of syslinux that treats a
PHDR segment as distinct from a PT_LOAD segment and rejects an ELF if
the two segments are overlapping.


nomadeel (approved):

nomadeel:
The links check is irrelevant to this, so I think we're good to merge.

PR #369, opened 2021-05-27
ARM64: Add option to ignore SErrors

Indanz:
By default any SError interrupt will halt the kernel. SErrors
may be caused by e.g. writes to read-only device registers or
uncorrectable ECC errors.

When this option is enabled SErrors will be ignored.

Enable this option for the TX2 platform.


Indanz:
I get a probably erroneous "Error: Action style failed." message for config.cmake, assuming it complaints about the lack of indentation: Any white space before the start of line continuations shows up in the actual help text as white space will and mess up the CMake help text formatting for this option. So in my opinion all the other help texts are badly formatted, as they all show messed up text in ccmake.

src/arch/arm/config.cmake
```diff
@@ -179,6 +179,17 @@ config_option(
     DEFAULT_DISABLED OFF
 )
 
+config_option(
+    KernelAArch64PassSError AARCH64_PASS_SERROR
```
sylGauthier, line 183: There is a trailing white space here.

sylGauthier (changes requested):
Hi Indanz,

I'm happy to approve the PR, just a few minor things:

- There is a trailing white space in the config.cmake file.
- The commit title is too long.
- Some of the lines in the commit message are too long (> 72 characters).

src/arch/arm/config.cmake
Indanz: I tried fixing the other complaints as well, amended the commit and pushed it. (In hindsight I should have created a branch for this pull request.) This means the help text will be messed up like all the others, but that should be fixed for all help texts instead of just one.

Is there a way to do these style checks locally before commit instead of finding out about it afterwards?

axel-h: You could run the script from https://github.com/seL4/seL4_tools/blob/master/misc/style.sh

lsf37: Yes, that's the one. I tend to use https://github.com/seL4/seL4_tools/blob/master/misc/style-all.sh to get all files so I have to think less about which files I changed, but the effect is the same.

Indanz: Thanks, calling seL4_tools/misc/style-all.sh from my seL4's root directory works. I also copied .gitlint and ran "gitlint install-hook" to get pre-commit message checking.

yyshen (commented):
I guess a random fault handler does not necessarily have the knowledge of handling a system error, which needs platform-specific information. Could we instead report such system errors to a dedicated fault handler which has the knowledge of handling serrors for a platform?

kent-mcleod:
> I guess a random fault handler does not necessarily have the knowledge of handling a system error, which needs platform-specific information. Could we instead report such system errors to a dedicated fault handler which has the knowledge of handling serrors for a platform?

Yea.  I believe that this patch is intended to be an interim measure until something like you describe can be designed and implemented.

kent-mcleod:
I think you would need to also ensure that the idle thread is never switched to as there would be no fault handler for the kernel to deliver an SError fault to.

Indanz:
> I think you would need to also ensure that the idle thread is never switched to as there would be no fault handler for the kernel to deliver an SError fault to.

This, together with the missing SErrors when the current task runs our of budget are big enough problems that a more complex solution is necessary. Fixing the edge-triggered interrupt problem is straightforward, but I have no idea how to solve the idle thread problem.

Current patch is still better than just halting, but not quite good enough. I'll try to route SError to a fake (per core) IRQ and see how that goes.

kent-mcleod:
>  I have no idea how to solve the idle thread problem.

If you have any user level components that are trusted to the same level as the kernel (which I assume you do), then you could create a TCB with the lowest priority level and assign a fault handler.

Indanz:
For now I'm going for the simplest solution of ignoring SErrors like TX2 does.

So newest commit is trivial: Instead of ignoring SErrors only for TX2, make it a config option instead.

yyshen:
> For now I'm going for the simplest solution of ignoring SErrors like TX2 does.
> 
> So newest commit is trivial: Instead of ignoring SErrors only for TX2, make it a config option instead.

Ignoring and SError might not be a good idea. When an serror happens, the correctness of system state could be compromised. Thus, if the kernel or a user-mode handler does not how to recover from the serror, halting the system with some debug messages is an acceptable approach.

lsf37:
> Ignoring an SError might not be a good idea. When an serror happens, the correctness of system state could be compromised. Thus, if the kernel or a user-mode handler does not how to recover from the serror, halting the system with some debug messages is an acceptable approach.

You are right, but I don't think either approach is good for all cases. Halting the entire system based on user actions is not acceptable for a whole class of systems (where denial of service is catastrophic). Ignoring errors is also not acceptable for a whole class of systems (where the SError could leave the system in undefined state). 

There are probably systems for which neither works, but making the behaviour configurable, will at least increase the chances of it being acceptable for a particular system until we have worked out a better overall approach. It does look like this will need at least a tweak in kernel mechanism to provide something.

kent-mcleod (approved):

Indanz:
> Ignoring an SError might not be a good idea. When an serror happens, the correctness of system state could be compromised. 

An SError is nothing more than an asynchronous system error, it is not special in that regard. It is only special because it is implementation defined and badly documented what can cause it.

The possible SErrors for our implementation are:

https://developer.arm.com/documentation/100236/0002/register-descriptions/aarch64-system-registers/exception-syndrome-register--el2?lang=en

The only documented error in the SoC datasheet is SLVERR, which will be raised for uncorrectable ECC errors on SDRAM. However, those also have a separate IRQ, so we don't need to catch them via SError. The same is true for double bit errors in L2.

DECERR is slightly documented in the GIC datasheet and can be caused by wiring something up wrongly to an AMBA bus, or accessing unused device memory.

nREI is RAM error, but our controller gives SLVERR and doesn't mention nREI.

And then we have the catch-all nSEI/VSEI: Vague undocumented system errors.

Considering neither our SoC datasheet, nor our CPU technical reference manual documents raising those errors, they must come from peripherals connected to the GIC. In which case the chance of them actually compromising the system state is very low, as the important exception, ECC errors, is handled separately.

In practise writing to some read-only device registers causes SLVERR too. Reading from some write-only registers causes a synchronous fault. Doing the same in a guest VM does not seem to raise SErrors in seL4.

> Thus, if the kernel or a user-mode handler does not how to recover from the serror, halting the system with some debug messages is an acceptable approach.

Agreed that logging a debug message, and in our case warning the user the system had an error would be best, but halting the system for known harmless SErrors is unacceptable. Otherwise it is not safe to give any device memory to user space, as they can halt the kernel.

My previous patch would be the most practical solution achieving this, but it forces all fault handlers to check for SErrors and gives no guarantees. It also has those pesky corner cases which makes re-using existing fault handling code harder. Having a separate SError notification mechanism (e.g. as an IRQ or special system fault handler on e.g. the idle task) would lose the current task information and add more complexity and code to the kernel, which I'm reluctant to do for something as obscure as this.

If a more comprehensive solution is wanted I'm willing to put more time into it, but in the meantime it would be nice to make TX2's hack a generic config option so there is an intermediate solution.

kent-mcleod:
@yyshen are you happy for this to be merged?

yyshen:
> @yyshen are you happy for this to be merged?

My happiness is irrelevant here ;-). Ignoring known **harmless** SErrors is reasonable; the system builders know what they are doing. Maybe just put a warning on the ignore option to remind it could be dangerous.

yyshen:
> > Ignoring an SError might not be a good idea. When an serror happens, the correctness of system state could be compromised.
> 
> An SError is nothing more than an asynchronous system error, it is not special in that regard. It is only special because it is implementation defined and badly documented what can cause it.
> 
> The possible SErrors for our implementation are:
> 
> https://developer.arm.com/documentation/100236/0002/register-descriptions/aarch64-system-registers/exception-syndrome-register--el2?lang=en
> 
> The only documented error in the SoC datasheet is SLVERR, which will be raised for uncorrectable ECC errors on SDRAM. However, those also have a separate IRQ, so we don't need to catch them via SError. The same is true for double bit errors in L2.
> 
> DECERR is slightly documented in the GIC datasheet and can be caused by wiring something up wrongly to an AMBA bus, or accessing unused device memory.
> 
> nREI is RAM error, but our controller gives SLVERR and doesn't mention nREI.
> 
> And then we have the catch-all nSEI/VSEI: Vague undocumented system errors.
> 
> Considering neither our SoC datasheet, nor our CPU technical reference manual documents raising those errors, they must come from peripherals connected to the GIC. In which case the chance of them actually compromising the system state is very low, as the important exception, ECC errors, is handled separately.
> 
> In practise writing to some read-only device registers causes SLVERR too. Reading from some write-only registers causes a synchronous fault. Doing the same in a guest VM does not seem to raise SErrors in seL4.
> 
> > Thus, if the kernel or a user-mode handler does not how to recover from the serror, halting the system with some debug messages is an acceptable approach.
> 
> Agreed that logging a debug message, and in our case warning the user the system had an error would be best, but halting the system for known harmless SErrors is unacceptable. Otherwise it is not safe to give any device memory to user space, as they can halt the kernel.
> 
> My previous patch would be the most practical solution achieving this, but it forces all fault handlers to check for SErrors and gives no guarantees. It also has those pesky corner cases which makes re-using existing fault handling code harder. Having a separate SError notification mechanism (e.g. as an IRQ or special system fault handler on e.g. the idle task) would lose the current task information and add more complexity and code to the kernel, which I'm reluctant to do for something as obscure as this.
> 
> If a more comprehensive solution is wanted I'm willing to put more time into it, but in the meantime it would be nice to make TX2's hack a generic config option so there is an intermediate solution.

Thanks for the detailed explanation. Indeed, the docs about SErrors are vague. So it is probably best to leave builders of a specific platform to decide how to handle them: ignore, user-mode handler, or halt.

Indanz:
> > @yyshen are you happy for this to be merged?
> 
> My happiness is irrelevant here ;-). Ignoring known **harmless** SErrors is reasonable; the system builders know what they are doing.

For the record, I'm not very happy with all of this either. The biggest problem is that we can't distinguish known harmless errors from unknown harmful ones. Best we could do is guess based on physical fault address.

But even then it's unknown if rebooting the system will cause more damage than trying to continue with an error. If the SError is not so bad that the system can function to notify the user about the problem, it's probably in a good enough shape to do most of its normal functions too, as mostly likely only one process is affected. If that process is important enough it will have multiple fail-safe measures and be restarted on failure, and that will also cause any device it handles being reset.

> Maybe just put a warning on the ignore option to remind it could be dangerous.

It's marked as advanced already, or do you mean the description? If you have any suggestions I'm willing to change the text.

But not ignoring SErrors can also be dangerous, depending on the use case.

yyshen:
> > > @yyshen are you happy for this to be merged?
> > 
> > 
> > My happiness is irrelevant here ;-). Ignoring known **harmless** SErrors is reasonable; the system builders know what they are doing.
> 
> For the record, I'm not very happy with all of this either. The biggest problem is that we can't distinguish known harmless errors from unknown harmful ones. Best we could do is guess based on physical fault address.
> 
> But even then it's unknown if rebooting the system will cause more damage than trying to continue with an error. If the SError is not so bad that the system can function to notify the user about the problem, it's probably in a good enough shape to do most of its normal functions too, as mostly likely only one process is affected. If that process is important enough it will have multiple fail-safe measures and be restarted on failure, and that will also cause any device it handles being reset.
> 
> > Maybe just put a warning on the ignore option to remind it could be dangerous.
> 
> It's marked as advanced already, or do you mean the description? If you have any suggestions I'm willing to change the text.
> 
> But not ignoring SErrors can also be dangerous, depending on the use case.

Oh, it is already marked as "advanced", so that is probably enough. Thanks for the contribution.

nomadeel:
Merging this so that this can go into the release. Dismissing @sylGauthier on behalf of him as the problems have been fixed.

Issue #347, opened 2021-04-13
KernelPaddrUserTop set incorrectly for ARM 64-bit platforms


bennoleslie:
A full discussion on the topic is available on discourse: https://sel4.discourse.group/t/device-region-untypeds/328

Given the current implementation for ARM 64-bit expects KernelPaddrUserTop to be exclusive, and changes will not impact verification I propose a very quick and easy fix to the config script in the short term.

This can be followed by a fix to the source to treat KernelPaddrUserTop to be inclusive (and ensures consistency across platforms) as possible.

lsf37:
👍

PR #376, opened 2021-06-01
Set KernelPaddrUserTop correctly on ARM platforms

bennoleslie:
On ARM platforms the code assumes the top phys addr is
exclusive, not inclusive (as is the case on other platforms).

While it would be good to make this consistent across platforms
that is a more invasive change. This simple fix makes the
confir correct for ARM. A more comprehensive consistency fix
can be applied at a later time.


lsf37 (approved):
From the verification side this looks good (and they are all on aarch64, so invisible to the current 32 bit proofs anyway).

kent-mcleod:
Resolves https://github.com/seL4/seL4/issues/347

kent-mcleod (approved):

PR #380, opened 2021-06-04
Update util.h

blackqbit:
delete duplicates : #define FASTCALL     __attribute__((fastcall))

Signed-off-by: blackqbit <blackqbit@gmail.com>


axel-h:
Thanks for the contribution. Please add the DCO to the commit message for this to be merged

kent-mcleod:
@lsf37 does this need a DCO if it is only removing code?

lsf37:
Since it's only one line, I'd be happy to accept it without DCO.

For removals, I guess it depends on whether it's a significant change to existing content. I.e. if you remove a whole file, because it's not used anywhere, I don't think that would be a significant change, but if you cleverly remove some lines from an algorithm to form a new algorithm, then the new algorithm might be new derived work in its own right, and would require a DCO. Does this make sense?

kent-mcleod (approved):

PR #383, opened 2021-06-09
risc-v: adapt to OpenSBI post v0.9

axel-h:
- risc-v/hifive: adapt to post-v0.9 OpenSBI
- riscv: provide wrapper for new (extended) SBI interface

Test with seL4/sel4test-manifest#13

It seems `sifive/fu540` is no longer a valid OpenSBI platform, all this is generic now? Does anybody have access to a hifive or polarfire to test what works here? Seems there is QEMU support also, has anybody tested this?

### DIFF
--- a/src/plat/hifive/config.cmake
+++ b/src/plat/hifive/config.cmake
@@ -13,7 +13,7 @@ if(KernelPlatformHifive)
     declare_seL4_arch(riscv64)
     config_set(KernelRiscVPlatform RISCV_PLAT "hifive")
     config_set(KernelPlatformFirstHartID FIRST_HART_ID 1)
-    config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "sifive/fu540")
+    config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "generic")
     list(APPEND KernelDTSList "tools/dts/hifive.dts")
     list(APPEND KernelDTSList "src/plat/hifive/overlay-hifive.dts")
     declare_default_headers(


yyshen (commented):
Hello @axel-h could you add a bit context in your commit message? What does "proper" mean? My guess is "generic" means "OpenSBI"?

axel-h:
> Hello @axel-h could you add a bit context in your commit message? What does "proper" mean? My guess is "generic" means "OpenSBI"?

@yyshen We are trying to get it compiling and running everywhere. Seem the OpenSBI build system wants the parameter PLATFORM="generic" now instead of "sifive/fu540". I'm trying to clarify details here.

axel-h:
Do not merge this unless there is a plan to update OpenSBI to post-v0.9. We are currently using the `master` here, but so far have not found a really strong need to use it. The log output is a bit nicer and it has some cutting enge features useful for development purposes.

axel-h:
deleted branch by accident

kent-mcleod:
> Do not merge this unless there is a plan to update OpenSBI to post-v0.9. We are currently using the `master` here, but so far have not found a really strong need to use it. The log output is a bit nicer and it has some cutting enge features useful for development purposes.

If you still don't want this to be merged, are you able to convert the PR to a draft for now?

axel-h:
The PR is a draft now. Once OpenSBI releases a new version or there is a good reason to pick an intermediate version we can merge this.

axel-h:
https://github.com/seL4/sel4test-manifest/pull/13 should be merged first.

axel-h:
added a commit that make the kernel header file use the new new SBI interface, which is an extension of the legacy interface. Eventually we should start to use this then for all RISC-V platforms that support this.

axel-h:
Seems using the new SBI interface requires a update in the proofs also, as we have two registers with return values now. So I will remove this commit from here again and make this a separate PR.

axel-h:
The HiFive still seems to start OpenSBI v0.8 and not v0.9, shouldn't the 'Test with seL4/sel4test-manifest#13' make CI pick this up?

lsf37:
It should, let me have a look. It might not work for manifest revisions, actually, only the repos in the manifests.

lsf37:
Yes, I think that's it. This'll be a bit tricky to implement, because it uses the manifest to get information about what those PRs mean and what repos it should update. Might have to just add a special case for a PR on the manifest itself.

axel-h:
The failure in Hifive SMP/MCS run might be due to the invalid OpenSBI configuration, as this uses the v0.9 setting "generic" now instead of the v0.8 setting "sifive/fu540".

lsf37:
That is likely, yes.

axel-h:
Note that OpenSBI v1.0 has been released now, so using this version should become the long term goal now. Updating to 0.9 could still be an intermediate step.

kent-mcleod:
This change also works on 0.8.  Looking at the commit you linked that removes `sifive/fu540` it appears that hifive has worked on generic platform for a while.

kent-mcleod (approved):

PR #378, opened 2021-06-02
Add support for GICv3 virtualization on aarch64

bennoleslie:
This adds sufficient kernel support for the GICv3 interrupt controller
to be used in a virtualization context on aarch64.

This set of changes has some limitations, however it is still an
improvement on the status quo.

Limitations:

1: This only provides support for aarch64. Anyone wanting support
for aarch32 + GICv3 + virtualization would need to add additional
code.

2: This code only supports 32 priority levels. Support for more
than 32 priority requires changing the get/set_gic_vcpu_ctrl_apr
interface. This is feasible, but requires a more invasive set of
changes. 32 priority levels has been shown to be sufficient in
practise.

Impacts on verification:

This set of changes should only impact Aarch64 Hypervisor
configurations. This is not yet verified so should not have
an impact on verification.

Level of testing:

This has been tested on an iMX8QXP based board. Testing
has at this point in time been limited to a single virtual
machine.

Note: support for this board is not yet upstrea, but is
currently being prepared.

Explanation of changes:

Ideally a new config item would not be required and this
could be driven purely by DTS and hardware.yml configuration.
However, the structures.bf requires changes. This can only
deal with config.h header files, not other more complex
header files. As such it was necessary to introduce a config
item which can be used for this purpose.

The appropriate platforms (as determined by examination of
DTS files) have been updated with the appropriate config
setting. This config setting only has any relevance if
hypervisor mode is already enabled, so should not cause
any difficulty for existing code or configuration.

Note: No testing has been performed on the updated
platforms.

There may be alternative factorings of this, which could
be considered in future work.

Signed-off-by: Ben Leslie <benno@brkawy.com>

PR #387, opened 2021-06-11
Fix gic_dist_map->res9 size.

Indanz:
Binary compatible with old code because of padding before the 64-bit iroutern field.

Without the lucky padding the offset to iroutern would be wrong.


axel-h (approved):
Change seem sane. Maybe we should add some static asserts after the definition to check that some offset are sane. Something like:
```
// check that some_field is really at offset 0x42
_Static_Assert( 0x42 == (uintptr_t)&((some_reg_struct_t *)0)->some_field );
```

Indanz:
I added the static assert check for iroutern as that is the last field.

I also checked the other structures in that header file the same way, but those are a lot simpler, so I didn't commit those checks.

axel-h:
Nice! Could Can you break the line at 80 chars, to improve code readability?

axel-h:
Seems the style checker is still unhappy with the indention.

kent-mcleod (approved):
Nice find!

Indanz:
I found it when reviewing #378. I didn't realise Ben didn't add that table and just checked all offsets just in case.

lsf37:
>Something like:
> 
> ```
> // check that some_field is really at offset 0x42
> _Static_Assert( 0x42 == (uintptr_t)&((some_reg_struct_t *)0)->some_field );
> ```

For posterity: please do not use `_Static_Assert` in the kernel code, it's not C99 and not in the verification subset. Instead use `compile_assert`, which maps to `_Static_Assert` when that is available and can be handled properly by the C parser in the verification tool chain (see also #1048).

PR #390, opened 2021-06-15
manual: apply seL4 Foundation branding

axel-h:


manual/manual.tex
```diff
@@ -114,12 +115,29 @@
...
+Matthias~Daum,
+Michael~von~Tessin,
+Peter~Chubb,
+Simon~Winwood,
+Thomas~Sewell,
+Timothy~Bourke and
+Toby~Murray.
+\\
+\\
+All authors and contributors can be contacted via the seL4 Foundation.
```
lsf37, line 140: Would rephrase that, we probably collectively have a private email address for everyone in here, but I don't think everyone in the list wants to be contactable via the foundation.

Maybe instead we should add a link to discourse/mailing list for contact?

manual/manual.tex
```diff
@@ -103,7 +103,8 @@
 
   \vfill
 
-  \copyright~{\commityear} General Dynamics C4 Systems.\\
+  \copyright~{2014} General Dynamics C4 Systems.\\
+  \copyright~{\commityear} seL4 Foundation.\\
```
lsf37, line 107: This is not quite right. The copyright year does not novate automatically with the last commit (must be substantial changes), and the copyright of the manual is not with the foundation.

There have been many substantial non-GD contributions in the past, so we could say something more generic like
```suggestion
  \copyright\ the seL4 authors and contributors.\\
```

manual/manual.tex
```diff
@@ -103,7 +103,8 @@
...
 
   \textsc{All rights reserved}.
```
lsf37, line 109: We should probably update this to the license the manual is provided under. Ideally this would be CC-BY-SA-4.0, but to re-release under that license we'd need the consent of all copyright holders, which is probably too hard by now. So I guess it has to remain GPL-2.0-only as defined in the SPDX header.

We could say:
```suggestion
  \textsc{All rights reserved}. You may use this manual and its sources under the terms of the GPL license, version 2.
```

manual/manual.tex
```diff
@@ -34,7 +34,7 @@
 \chead{}
 \rhead[\fancyplain{}{\sl\leftmark}]{\fancyplain{}{\sl\thepage}}
 \lfoot[\fancyplain{\sl\thepage}{}]{}
-\cfoot{\ifDraft\textsf{Data61 Confidential}\fi}
+\cfoot{\ifDraft\textsf{seL4 Foundation Confidential}\fi}
```
lsf37, line 37: We can probably drop the confidential part now, since any updates will be developed publicly anyway. I.e. I'd propose:
```suggestion
\cfoot{}
```

lsf37 (changes requested):
This is an excellent idea, thanks for doing that!

There is an actual foundation latex style (which Gernot wrote) in the whitepaper repo. We should eventually use that one, and I'm happy to do the conversion, but I propose we still merge this PR first after the minor comments below.

manual/manual.tex
axel-h: I don't have any preference what to put there and I'm open to suggestions. Basically I just removed the sentence about TS e-mail, because I was assuming this is no longer really correct. I'll change this to discourse and the mailing list then.

manual/manual.tex
axel-h: * will add your suggestion.
* Can you confirm 2014 for GD is ok. Should this list a starting year also? Should we add 2020 as starting year for the seL4 Foundation?
* Should we add a license note, can be takle "Distributed under the Creative Commons Attribution-ShareAlike 4.0 International (CC BY-SA 4.0) License." that is used in the seL4 whitepaper. 
* Should we add the statement "seL4 is a trademark of LF Projects, LLC.", as the whitepaper says?

manual/manual.tex
axel-h: is says `\ifDraft` there and having drafts may be handy for all non-release builds. So use "seL4 Foundation DARFT" instead?

manual/manual.tex
lsf37: - 2014 is correct for GD, but I'd leave out both GD and the Foundation, actually, and only do the generic part, since they would both be authors/contributors
- unfortunately can't do CC-BY-SA-4.0 (see other comment below)
- yes, the trademark sentence is definitely a good idea

manual/manual.tex
lsf37: IIRC, when the draft option is on, each page will have a huge DRAFT watermark in the background, so I think it's Ok to leave the footer empty.

axel-h:
I've added your suggestions and trademark references.

manual/manual.tex
```diff
@@ -103,9 +103,19 @@
...
-  \textsc{All rights reserved}.
+  \textsc{All rights reserved}.\\
+  \\
+  \texttt{SPDX-License-Identifier: GPL-2.0-only}\\
+  You may use this manual and its sources under the terms of the GPL license,
+  version 2.\\
+
+  seL4\textregistered\ is a trademark of LF Projects, LLC.\\
+  Arm\textregistered\ is a registered trademark of Arm Limited.\\
+  RISC-V\textregistered\ is a registered trademark of RISC-V International.\\
```
lsf37, line 117: 👍

lsf37 (approved):

axel-h:
ok, the license check is happy now also, I've put the SPDX statement visible in the document on its own line.

axel-h:
and finally updating the contributor list, while I'm at it.

lsf37:
Nice! Happy for you to merge when you're happy with it.

mbrcknl (approved):

PR #394, opened 2021-06-16
CMake: don't enforce std channels

axel-h:
Enforcing the std channels seems to conflict with Jenkins CI, we don't see any error messages then. I remember that we had a discussion here before that on some other systems this enforced redirectin is needed - but this make me wonder why the std channels do not work there and if this is the issue to be resolved then.


kent-mcleod (approved):
When these were originally added it was because the default behavior didn't seem to connect stdin, stdout and stderr to the commands spawned by execute_process.  Running some more tests now it appears that the streams are being connected as expected.  So I'm happy to approve these changes.

PR #395, opened 2021-06-17
manual: apply foundation branding

lsf37:
- applies the seL4 foundation latex style
- switch to natbib style for references (foundation default)
- use breakable space for author names
- small tweaks to use the style smoothly

We might consider using more colour for the API reference at some point, but I'll stop here for now.

(This is a follow-on from Axel's PR #390)


lsf37:
So it's complaining about trailing whitespace in the PDF, which is a new one. If people are otherwise happy, I'd just ignore that. Or should we investigate why `git diff --check` does not recognise the PDF as binary?

manual/manual.tex
```diff
@@ -125,32 +120,31 @@
   \begin{abstract}
 % This list of contributors is based on the hg log. If you make commits please
 % add your name in alphabetical order.
-The primary authors of this document are Matthew~Grosvenor and Adam~Walker,
+The primary authors of this document are Matthew Grosvenor and Adam Walker,
 with contributions from
-Adrian~Danis,
```
axel-h, line 130: I added this on purpose, intention was really to keep the names together, especial for people with more than two words in a name, So I'm curious why you remove it again.

axel-h (approved):

gernotheiser:
Looks good, thanks for taking the initiative, @axel-h

manual/manual.tex
```diff
@@ -82,18 +74,18 @@
...
-
 \begin{document}
 
-  \title{seL4 Reference Manual\\Version \version}
+  \title{seL4\\ Reference Manual}
+  \subtitle{Version \version}
 
   \author{seL4 Foundation}
-  \AuthorEmail{https://sel4.systems}
+  \email{https://sel4.systems}
```
axel-h, line 88: `\url{}` ?

manual/parts/intro.tex
```diff
@@ -15,7 +15,7 @@ \chapter{\label{ch:intro}Introduction}
 spaces, threads, and inter-process communication (IPC). The small number
 of services provided by seL4 directly translates to a small
 implementation of approximately $8700$ lines of C code. This has allowed
-the ARMv6 version of the kernel to be formally proven in the Isabelle/HOL 
+the ARMv6 version of the kernel to be formally proved in the Isabelle/HOL
```
axel-h, line 18: nitpicking: change does not fit to commit message ;)

manual/parts/threads.tex
```diff
@@ -458,8 +458,8 @@ \subsection{Debug Exception: Breakpoints and Watchpoints}
 
 \begin{description}
 \item[seL4\_NumHWBreakpoints]: Defines the total number of hardware break
-registers available, of all types available on the hardware platform. On the ARM
-Cortex A7 for example, there are 6 exclusive instruction breakpoint registers,
+registers available, of all types available on the hardware platform. On the Arm
+Cortex~A7 for example, there are 6 exclusive instruction breakpoint registers,
```
axel-h, line 462: Arm uses a dash "Cortex-A7" in the naming. In the `amsmath` package there is `\nobreakdash` for this.

manual/parts/threads.tex
```diff
@@ -632,7 +632,7 @@ \section{Virtualisation}
 as storage for this state. For simplicity we refer to this virtualised higher privileged level as
 'guest mode'. \obj{VCPU}s are bound in a one-to-one relationship with a \obj{TCB} in order
```
axel-h, line 633: nitpicking: also does not match commt message.

manual/parts/intro.tex
gernotheiser: ;-)

axel-h:
I added a few minor remarks, but these don't block merging.

manual/manual.tex
lsf37: Right, I was wondering where it came from. Forcing LaTeX to keep all names together makes it impossible for it to actually lay out the paragraph correctly -- most lines go into the margin, and for some it made LaTeX try to hyphenate the name, which looked even worse.

It's fine for first/last names to not be on the same line, you'll find that pretty much everywhere. I've left `Le~Sueur` in, because even I found that strange on seperate lines.

We could change the whole thing to print each name on a separate line, which would also give it a bit more prominence.

gernotheiser:
Building all kernel versions for a doc commit seems a bit overkill ;-) Fortunately the manual doesn't change often without kernel changes...

manual/manual.tex
lsf37: The `\email` command is from the `sel4.sty` package. Usually you want an email there, but it just renders it as url anyway, so I think that's fine.

manual/parts/threads.tex
lsf37: I'd prefer to fix these separate and not side-track the PR even more..

manual/manual.tex
gernotheiser: I agree that preventing first-last name separation is not needed. However, the overfull lines are prevented by having a \sloppy right after the \documentclass command (almost always the right thing unless you do a lot of layout copy-editing)

PR #399, opened 2021-06-19
Set KernelPaddrUserTop correctly on RISC-V platforms

axel-h:
The top phys addr is exclusive, not inclusive. On 64-bit platforms there is a trivial way to fix this as long as the whole address space is never used. For 32-bit platforms the fix will be more complicated, so this is postponed.

See also https://github.com/seL4/seL4/pull/376 where this got fixed on ARM platforms.


axel-h:
will recreate this one, there is a braach name conflict.

PR #402, opened 2021-06-20
cleanup CMake ARM config

axel-h:


configs/seL4Config.cmake
```diff
@@ -184,71 +187,98 @@ unset(KernelArmMachFeatureModifiers CACHE)
...
-elseif(KernelArm1136JF_S)
-    set(KernelArmCPU "arm1136jf-s" CACHE INTERNAL "")
-endif()
 if(KernelArchARM)
+
+    # Check ARM Cortex-A cores setup. While different Cortex-A version can be
+    # enabled, 'KernelArmCPU' can only be set to one. So we take the first one
+    # from the list, the current ordering ins simply ascending without any
+    # deeper meanding
+    set(first_match "")
```
kent-mcleod, line 226: In my opinion this new rewrite of the above code isn't easier to read/maintain/understand because it's got more complex control flow than what it's replacing.  Are you looking at trying to provide error messages for if the platform/config.cmake files make incorrect CPU/ARCH settings? Another way that we could achieve this is by inverting the logic, and requiring the platform files to set the string variants of KernelArmCPU and KernelArmArmV and then handling the boolean options the same way as with KernalSel4Arch.  

Ultimately, it would be better to set the CPU and March from the device tree.  I looked at doing this, but currently hardware_gen.py takes KernelPaddrUserTop as an argument, and KernelPaddrUserTop is set based on the ArmCPU value.  So it would require a bit more effort to implement.

PR #408, opened 2021-06-21
cmake: Add seL4Config.cmake include CMakeLists.txt

kent-mcleod:
seL4Config.cmake is responsible for generating a valid
CMAKE_TOOLCHAIN_FILE and setting up platform config options at the start
of the build. The CMAKE_TOOLCHAIN_FILE variable has to be set before the
first cmake `project()` function is processed to take effect.
Previously this file was required to be imported in a CMake script
before the kernel's CMakeLists.txt could be processed. This prevented
the main CMakeLists.txt file from being used without an additional
configuration file:
cmake -G Ninja -C ../configs/ARM_verified.cmake ../

Now it is possible to do:
cmake -G Ninja -DKernelPlatform=imx6 -DKernelARMPlatform=sabre ../

This should make it easier to invoke CMake for building kernel
configurations from other build environments.

Because this file is now imported in the Kernel's CMakeLists.txt
context, there is no longer a requirement to save all the intermediate
settings into the cache and then read them out again.


kent-mcleod:
@ssrg-bamboo `test`

configs/seL4Config.cmake
```diff
@@ -185,7 +156,6 @@ unset(KernelArmCPU CACHE)
 unset(KernelArmArmV CACHE)
 
 # Blacklist platforms without MCS support
```
axel-h, line 158: maybe this comment should also be removed. Or updated to say that platform must explicitly turn this off now.

kent-mcleod: I think your other PR that updates the comment should be sufficient.  I haven't changed the behavior of KernelPlatformSupportsMCS in a way that makes the comment wrong. 
KernelPlatformSupportsMCS still works the same way:
- It's default to ON, but platforms that don't support MCS set it to off.
What's changed is that the result isn't written into the cache which was requiring the entry to be deleted each time the script reruns.

ssrg-bamboo:

lsf37 (approved):

ssrg-bamboo:

lsf37:
@kent-mcleod This is independent of this PR, but the style check wants the `.cmake` files to be not executable. Is there a reason why they were set executable in the first place?

lsf37:
Ok, I found 297d2b63da39ea26b3c437a15247fcc6508d8598, which explains the `+x`. I'm happy to ignore the style check on this one.

lsf37:
Cancelling the hardware test before merge where it will run again anyway (it has passed before the branch update).

PR #409, opened 2021-06-21
Assume seperate distinct kernel ELF mapping on all platforms

xurtis:
This updates the code such that it assumes that the kernel ELF is not accessed via the physical memory window and instead has a dedicated region into which it is mapped. This has the most noticeable impact for aarch32 which has a verified configuration that *does* access the kernel via the physical memory window.

This will allow configurations with the kernel imaging API to function correctly but will not allow the API to be configured without a dedicated mapping window.

### DIFF
--- a/include/machine.h
+++ b/include/machine.h
@@ -26,6 +26,15 @@ static inline paddr_t CONST addrFromPPtr(void *pptr)
     return (paddr_t)pptr - PPTR_BASE_OFFSET;
 }
 
+/* When obtaining a physical address from a reference to an address from
+ * the kernel ELF mapping, this function must be used. */
+static inline paddr_t CONST addrFromKPPtr(void *pptr)
+{
+    assert((paddr_t)(pptr) >= KERNEL_ELF_BASE);
+    assert((paddr_t)(pptr) <= KERNEL_ELF_TOP);
+    return (paddr_t)pptr - KERNEL_ELF_BASE_OFFSET;
+}
+
 static inline region_t CONST paddr_to_pptr_reg(p_region_t p_reg)
 {
     return (region_t) {
@@ -41,6 +50,7 @@ static inline p_region_t CONST pptr_to_paddr_reg(region_t reg)
 }
 
 #define paddr_to_pptr ptrFromPAddr
-#define pptr_to_paddr(x) addrFromPPtr(x)
+#define pptr_to_paddr addrFromPPtr
+#define kpptr_to_paddr addrFromKPPtr
 
 #include <mode/machine.h>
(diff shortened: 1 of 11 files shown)


xurtis:
@ssrg-bamboo `proof`

xurtis:
@ssrg-bamboo `test`

ssrg-bamboo:
Hello, I'm a bot! I've set up a proof testboard for this PR [here](https://github.com/seL4/gh-testboard/blob/xurtis-use-kernel-translation-409-052926/manifest.xml), and results should show up as `pending` in status checks in a few minutes.

include/machine.h
```diff
@@ -26,6 +26,15 @@ static inline paddr_t CONST addrFromPPtr(void *pptr)
     return (paddr_t)pptr - PPTR_BASE_OFFSET;
 }
 
+/* When obtaining a physical address from a reference to address from the
```
axel-h, line 29:
```suggestion
/* When obtaining a physical address from a reference to an address from the
```

ssrg-bamboo:

Hello, I'm a bot! I've set up a proof testboard for this PR [here](https://github.com/seL4/gh-testboard/blob/xurtis-use-kernel-translation-409-085235/manifest.xml), and results should show up as `pending` in status checks in a few minutes.

include/machine.h
```diff
@@ -26,6 +26,15 @@ static inline paddr_t CONST addrFromPPtr(void *pptr)
...
+ * kernel ELF mapping, this function must be used. */
+static inline paddr_t CONST addrFromKPPtr(void *pptr)
```
axel-h, line 31: add a "p" in the name (`paddrFromKPPtr()`) to clearly imply this is a physical address? That make the code that uses this easier to understand.

include/machine.h
```diff
@@ -41,6 +50,7 @@ static inline p_region_t CONST pptr_to_paddr_reg(region_t reg)
 }
 
 #define paddr_to_pptr ptrFromPAddr
-#define pptr_to_paddr(x) addrFromPPtr(x)
+#define pptr_to_paddr addrFromPPtr
```
axel-h, line 53: Sould use explicit parameter names here.
```suggestion
#define pptr_to_paddr(x) addrFromPPtr(x)
```
Or maybe use one name only for everything to reduce confusion.

include/machine.h
xurtis: Whilst I agree, I don't want to make an edit that would touch that many locations here.

include/machine.h
xurtis: The change here was to remove the reference to parameters and only update the token of the function name. Once again, I don't want to introduce an edit here that changes all the locations that these are called.

mbrcknl (approved):
Proofs are done and checked, so this is good to go from the verification perspective.

sylGauthier (approved):
Looks good to me

PR #410, opened 2021-06-21
CMake: Add support for `cmake install <dir>` cmd

kent-mcleod:
When using CMake to only build the kernel, an install target is now
provided to copy important outputs into an installation directory.

Currently only the following files are installed:
- ./bin/kernel.elf: Location of kernel.elf binary
- ./kernel_Config/include: The include root for the kernel_Config
  interface library
- ./libsel4/include: The include root for libsel4
- ./libsel4/src: The c source files for the libsel4 library
- ./libsel4/sel4_Config/include: The include root for the sel4_Config
  interface library
- ./libsel4/autoconf/include: The include root for the autoconf interface
  library

To build and install this project to an installation directory should
now only require the following from a clean build directory:
```
cmake -G Ninja -DCMAKE_INSTALL_PREFIX=<path-to-install> -DOption=val <src-dir>;
cmake --build .;
cmake --install .;
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -583,3 +583,49 @@ target_link_libraries(kernel.elf PRIVATE kernel_Config kernel_autoconf)
 set_property(TARGET kernel.elf APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-T ${linker_lds_path} ")
 set_target_properties(kernel.elf PROPERTIES LINK_DEPENDS "${linker_lds_path}")
 add_dependencies(kernel.elf circular_includes)
+
+# The following commands setup the install target for copying generated files and
+# compilation outputs to an install location: CMAKE_INSTALL_PREFIX.
+# CMAKE_INSTALL_PREFIX can be set on the cmake command line.
+#
+# The current installation outputs are:
+# - ${CMAKE_INSTALL_PREFIX}/bin/kernel.elf: Location of kernel.elf binary
+# - ${CMAKE_INSTALL_PREFIX}/libsel4/include: The include root for libsel4
+# - ${CMAKE_INSTALL_PREFIX}/libsel4/src: The c source files for the libsel4 library
+#
+# The install target is only created if this is the top level project.
+# We don't currently support creating install targets if the kernel is
+# imported in another project.
+if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
+
+    # Import libsel4 to get access to generation targets
+    add_subdirectory(libsel4)
+    # Add a default target that builds kernel.elf and generates all libsel4 headers
+    add_custom_target(single-project ALL DEPENDS sel4_generated kernel.elf)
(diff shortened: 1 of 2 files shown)


kent-mcleod:
This partially depends on https://github.com/seL4/seL4/pull/408 to enable configuration without requiring an additional `-C <src-dir>/configs/seL4Config.cmake`

lsf37 (approved):
Looks good. @kent-mcleod could you please fix up the gitlint failure (one line too long in the body), then I think we can merge it.

kent-mcleod:
> Looks good. @kent-mcleod could you please fix up the gitlint failure (one line too long in the body), then I think we can merge it.

There was some discussion on Jira in relation to this issue that I have been sitting on and should still flesh out before merging.

lsf37:
Ah, missed that. Which issue is it? Same as #411?

bennoleslie:
This looks like a good start to me. What is the rationale to put all the different include into different directories.

I have a Makefile, that I'd like to look something like this:

```
CFLAGS := -nostdlib -ffreestanding -O3 -I$(SEL4_INSTALL)/libsel4/include
```

but right now I have it looking like this:

```
CFLAGS := -nostdlib -ffreestanding -O3 -I$(SEL4_INSTALL)/libsel4/include -I$(SEL4_INSTALL)/libsel4/autoconf/include -I$(SEL4_INSTALL)/libsel4/sel4_Config/include -I$(SEL4_INSTALL)/kernel_Config/include
```

What's the motivation for splitting these into different include roots?

In my `sel4-core` branch I had effectively merged all these into a single include directory and did not have any negative issues (at least none that I could determine, but I could easily be missing something).

axel-h:
Does this also provide the dts and dts-overlay files?

bennoleslie:
> Does this also provide the dts and dts-overlay files?

It doesn't. I think #411 is probably the best place to explain the need for DTS files.

bennoleslie (approved):

kent-mcleod:
> This looks like a good start to me. What is the rationale to put all the different include into different directories.

There are a lot of libraries that have an include dependency on kernel_config (or autoconf) headers, but explicitly not on libsel4.  So adding those headers into the libsel4 include dir would break that isolation.  I've gone back and forth on this for a while, and I know that using different CMake install command parameters can be used to control this behavior, so I think for the option with no special include arguments it's probably best to just put everything in the same include directory.

kent-mcleod:
https://github.com/seL4/util_libs/blob/master/libutils/CMakeLists.txt#L35 is an example of a library that uses sel4_autoconf but nothing from libsel4.  Every library in util_libs is the same.  Same with all the libraries in https://github.com/seL4/projects_libs/.

bennoleslie:
I don't strongly care one way or the other regarding same or different directories for headers. For my purposes I move them all into a single location with post processing. I'd be very happy seeing this merged as-is. I'm sure once it is merged and people start using it and have more experience there will be additional changes requested.

bennoleslie (approved):

PR #412, opened 2021-06-21
github: CI action to lint seL4 IDL files

lsf37:
This will lint everything called `sel4*.xml` in `libsel4` and should trigger only on changes to those files.

Closes seL4/ci-actions#70


lsf37:
Have tested triggering these with one failed and one successful lint, seems to be working as intended.

kent-mcleod (approved):

PR #421, opened 2021-06-28
risc-v: don't add 8 bytes space at end of bss

axel-h:
Is there any explanation what this is for?

### DIFF
--- a/src/arch/riscv/common_riscv.lds
+++ b/src/arch/riscv/common_riscv.lds
@@ -1,6 +1,7 @@
 /*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright 2015, 2016 Hesham Almatary <heshamelmatary@gmail.com>
+ * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -80,7 +81,6 @@ SECTIONS
 
         /* large data such as the globals frame and global PD */
         *(.bss.aligned)
-        . = . + 8;
     }
 
     . = ALIGN(4K);


axel-h:
Can we merge this is nobody has an explanation why/where this is needed?

yyshen:
> Can we merge this is nobody has an explanation why/where this is needed?

@axel-h Can you observe any differences between the files with/without the change?

axel-h:
Without these 8 byte things also work for us. Furthermore, the memory layout looks nicer, because no page is wasted. The 8 byte go into a new page actually and then the rest of the page is filled with padding. I could not come up with any explanation what the 8 bytes are for, maybe this is some copy/paste from somewhere else, e.g. a HFIT hack for QEMU/Spike that is now handles in the SBI.

lsf37:
`git blame` shows this is from @heshamelmatary's original port still, which was for Spike.

If he doesn't remember any reason for it, I think it's fine to remove. @kent-mcleod do you have any memory of it?

lsf37:
(I removed the bug label, because it doesn't seem to break anything with or without. Not sure what category it should go into, maybe we need a new one for things like these).

axel-h:
Ture, it's not really a bug. But it "breaks" a nice&clean memory layout, which initially made me look into what is going on here and if this tries do reserve another a page for something.

lsf37:
I'd be happy with "Enhancement". It's not a new feature, but it cleans up something and makes it nicer. Maybe we should have a "Cleanup" label.

kent-mcleod:
Sometimes these small offsets are to get the linker to do something specific.  This one (https://github.com/seL4/seL4/blob/master/src/plat/pc99/linker.lds#L48) was the only way to get the linker to not place that segment at the end of all of the other ones.

Considering that there is no comment addressing why this was originally added and it's causing an extra page to be allocated, it is an argument for removing it.  

I went and dug some further and I believe I know what it is for.  Looking at the .bss definition of binutils linker scripts:
```
  .bss            :
  {
   *(.dynbss)
   *(.bss .bss.* .gnu.linkonce.b.*)
   *(COMMON)
   /* Align here to ensure that the .bss section occupies space up to
      _end.  Align after .bss to ensure correct alignment even if the
      .bss section disappears because there are no input sections.
      FIXME: Why do we need it? When there is no .bss section, we do not
      pad the .data section.  */
   . = ALIGN(. != 0 ? 32 / 8 : 1);
  }
  _bss_end__ = .; __bss_end__ = .;
  . = ALIGN(32 / 8);
  . = SEGMENT_START("ldata-segment", .);
  . = ALIGN(32 / 8);
  __end__ = .;
  _end = .; PROVIDE (end = .);
```
The last statement in the .bss segment is there to ensure that the _end symbol is the end of the segment that the .bss section lies in.  Otherwise the segment could end a page earlier and the page before _end isn't loaded by the loader.  

In the kernel's case, there's actually a missing `*(COMMON)` statement that is inserted automatically, after the `. = . + 8;` that is being removed.  With gcc versions earlier than 10, global variables defined but not initialized get put in the COMMON section and would end up at the end of the .bss.  It's only when upgrading to gcc-10 that the compiler changed the default behavior to place these symbols inside .bss instead of COMMON. So in gcc-10 the _ki_end is one page higher than it needs to be but on earlier versions of gcc it isn't.  

In any case, we just need to ensure that ki_end is > than any kernel memory so that the backing memory doesn't get turned into untypeds. 


As an aside, @axel-h, your explanation below is something that I would argue should be going into the commit message for this change.  Otherwise in the future when looking back, we get the same situation where we can't figure out why a change was made.  

> Without these 8 byte things also work for us. Furthermore, the memory layout looks nicer, because no page is wasted. The 8 byte go into a new page actually and then the rest of the page is filled with padding. I could not come up with any explanation what the 8 bytes are for, maybe this is some copy/paste from somewhere else, e.g. a HTIF hack for QEMU/Spike that is now handles in the SBI.

The commit style we try and enforced is described here: https://docs.sel4.systems/processes/git-conventions.html#commit-messages.
In particular the following is pretty important:

> Follow this with a blank line, then a summary of why the changes are required, and if non-trivial, how they are required.

Unfortunately, when we switched to using DCO sign-offs, the gitlint check stopped enforcing this as it treat's the DCO sign-off line as a commit message body.  In reality it should be failing if any commit message doesn't have an explanation of why the change was required.

lsf37:
Cool, thanks for digging deeper into this.

We could try to hack gitlint to get the old behaviour back, it's been bugging me for a while that it takes the sign-off as body. We probably don't want to enforce this on every repo, but the kernel repo specifically would be one where we do want it.

On danger of side tracking: are there useful properties that we want of linker script outcomes? I.e. more things like you mentioned `ki_end is > than any kernel memory`. If we could formalise these and the semantics of linker scripts, we could actually get some useful checks out of that.

axel-h:
- fixed the commit comment. 
- for the missing COMMON, is https://github.com/axel-h/seL4/commit/a7473d9c12a655555116cde0a7b9f7f73ae3d3b6 what should be there?

kent-mcleod:
> for the missing COMMON, is [axel-h@a7473d9](https://github.com/axel-h/seL4/commit/a7473d9c12a655555116cde0a7b9f7f73ae3d3b6) what should be there?

Yes

heshamelmatary:
@lsf37 I'm afraid I don't remember the reason why this was added in the first place. I don't see a good reason to keep it now.

lsf37:
> @lsf37 I'm afraid I don't remember the reason why this was added in the first place. I don't see a good reason to keep it now.

No worries, if you don't remember anything specific either, I think Kent's interpretation is probably right.

axel-h:
Can we merge this?

lsf37 (approved):

ratmice:
> If we could formalise these and the semantics of linker scripts, we could actually get some useful checks out of that. (Or in fact, potentially just running the linker script and checking that the properties are true)

Might already know about it, but the following project/paper does formalize some semantics of linker scripts in Isabelle/HOL I believe.

https://github.com/rems-project/linksem 
https://www.cl.cam.ac.uk/~pes20/rems/papers/oopsla-elf-linking-2016.pdf

lsf37:
We currently still don't have a RISC-V board in the test, but I checked it manually on `hifive` and things work as expected

lsf37:
> Might already know about it, but the following project/paper does formalize some semantics of linker scripts in Isabelle/HOL I believe.
> 
> https://github.com/rems-project/linksem
> https://www.cl.cam.ac.uk/~pes20/rems/papers/oopsla-elf-linking-2016.pdf

Nice one. I was thinking of a paper by the AWS formal reasoning group on linker script verification, but have so far failed to find it again. Should be from about the same time, but my sense of time might be skewed a bit there.

axel-h:
Created https://github.com/seL4/seL4/issues/506 for the linker script formalization.

PR #422, opened 2021-06-28
boot/arm: pass DTB as pptr region

axel-h:


src/arch/arm/kernel/boot.c
```diff
@@ -41,19 +41,26 @@ BOOT_BSS static volatile int node_boot_lock;
...
+BOOT_BSS static region_t res_reg[MAX_RESERVED];
 
-BOOT_CODE static void arch_init_freemem(p_region_t ui_p_reg, region_t dtb_reg, v_region_t it_v_reg,
+BOOT_CODE static void arch_init_freemem(p_region_t ui_p_reg, region_t dtb_reg,
+                                        v_region_t it_v_reg,
                                         word_t extra_bi_size_bits)
 {
-    reserved[0].start = KERNEL_ELF_BASE;
-    reserved[0].end = (pptr_t)ki_end;
+    unsigned int index = 0;
```
pingerino, line 50: what principles are you applying to rewrite this code? What's the point here?

src/arch/riscv/kernel/boot.c
```diff
@@ -273,12 +273,12 @@ static BOOT_CODE bool_t try_init_kernel(
     seL4_BootInfoHeader header;
     if (dtb_reg.start) {
         header.id = SEL4_BOOTINFO_HEADER_FDT;
-        header.len = extra_bi_size;
+        header.len = sizeof(header) + dtb_size;
```
pingerino, line 284: why recalculate this?

axel-h: It's recalculated here to avoid implicit dependencies, and this is also what the ARM code does. There could be more in `extra_bi_size` than just this, so here I ensure that we just do exactly what we need. Note also, that this is part or a larger patch set, there is more to come where this is wrapper into helper function then that are shared code between ARM and RISC-V. The change here allows better tracing the changes.

src/arch/arm/kernel/boot.c
axel-h: This is supposed to make it more easy to understand an more compact. And it aligns things between ARM and RISC-V. This patches are from a bigger patch set that I'm upstreaming in smaller pieces. I can drop the last commit from this PR and then include it in a later PR, where this function disappears completely

pingerino: Is the intent to unify code between RISC-V and ARM and then remove this duplication after unifying? If so that makes sense, and I appreciate the effort and the upstreaming. 

Context is really important to help reviewers understand what is going on and will help the review process go smoothly. It would help if that was in the commit messages, because as it stands the comment just says 'clean up the code' which looks at face value like it's being rewritten for no good reason. Note that this policy is described in the guide for pull requests: https://docs.sel4.systems/processes/code-review.html.

I'm very against any unnecessary code churn (which without context, looks like what this is), because:
    - it takes up reviewer time
    - especially when it appears that most of the hardware tests are currently broken (so we don't know the changes don't introduce bugs) *and* its in unverified code (so we *really* don't know).

axel-h: I see you point and have remove the last commit from this PR.

axel-h:
Could this get merged now?

pingerino:
@axel-h the preprocess checks are failing, so you'll need someone from verification to approve I think, cc @lsf37

lsf37:
It looks good, just double-checking:

@ssrg-bamboo `proof`

lsf37 (dismissed):

axel-h:
I had to reworked this patch, since it is actually part of a bigger patch set. I wonder, is there a reason why all the reserved/available memory region processing happens on virtual addresses and not on physical addresses? On my bigger change set it turns out that tings are much easier to understand and less conversion forth and back are needed if physical regions are used and only at the end the virtual region are calculated where necessary. It looks to me that doing this on virtual region has historical reason mainly.

axel-h:
Can we merge this?

axel-h:
Passing physical DTB location to `arch_init_freemem()` on RISC-V now also. This is helpful for further patches that depend on this PR and also aligns the code with what is done on ARM.

axel-h:
I'l look into the rv64 failure, seems there is a quirk in the parameters passed on RISC-V.

axel-h:
RV64 works again, I will create separates PR to clean up the the types quirks, because this starts in the ElfLoader already.

axel-h:
The link check failed for `README.md` with:
```
	ERROR	https://doi.org/10.5281/zenodo.591727
		Dialing to the given TCP address timed out
```

I think this is just an "internet hiccup", but I wonder about the linke there and have created issue https://github.com/seL4/seL4/issues/512 as a follow up

axel-h:
rebased

axel-h:
now builds on top of https://github.com/seL4/seL4/pull/551

axel-h:
can we merge this?

axel-h:
Can we merge this now?

src/arch/arm/kernel/boot.c
```diff
@@ -326,25 +325,6 @@ static BOOT_CODE bool_t try_init_kernel(
     bi_frame_vptr = ipcbuf_vptr + BIT(PAGE_BITS);
     extra_bi_frame_vptr = bi_frame_vptr + BIT(PAGE_BITS);
 
-    /* If no DTB was provided, skip allocating extra bootinfo */
-    p_region_t dtb_p_reg = {
-        dtb_addr_start, ROUND_UP(dtb_addr_end, PAGE_BITS)
```
kent-mcleod, line 331: Rounding up the dtb_addr_end to page alignment looks to be unnecessary here.  This region of memory is only tracked by the kernel in order to give it to user level in an extra boot info field.  The extra boot info datastructure is a list of variable length blocks, with a seL4_BootInfoHeader at the start followed by a variable length of data. No guarantees are made about either the start of the bootinfo header, or the start of the data that comes after it being page aligned.  If alignment is required, then there is an explicit SEL4_BOOTINFO_HEADER_PADDING type for adding an entry that is used to add padding so that the next entry can have a certain alignment.  Rounding the dtb size up to a PAGE almost guarantees that the bootinfo extra size is going to use an additional page of memory because of `extra_bi_size = sizeof(seL4_BootInfoHeader) + (dtb_reg.end - dtb_reg.start);` further along.

src/arch/arm/kernel/boot.c
kent-mcleod: > This is supposed to make it more easy to understand an more compact. And it aligns things between ARM and RISC-V. This patches are from a bigger patch set that I'm upstreaming in smaller pieces. I can drop the last commit from this PR and then include it in a later PR, where this function disappears completely

Can you split this single commit into the part that changes Arm and the part that updates RISC-V to be the same as Arm.  The motivations for changing the Arm implementation are different from the motivations to make RISC-V the same as Arm and doing it all in the same change makes it harder to follow what changes are new for Arm and what changes are new for RISC-V

src/arch/arm/kernel/boot.c
kent-mcleod: Also it's odd that the memory for the DTB is marked as reserved when new memory is allocated for the DTB to be copied into and the original memory is never released. The original DTB region should ideally be free'd and  given back to userlevel as untypeds like how the boot code and data segments are.


(Both of these things are thoughts for future changes as they are extra to any clean up you are trying to do here.)

axel-h: Maybe reserving the area is intended for the situation where a DTB is passed from a previous bootloader transparently by the ELF Loder (instead of using the DTB from the CPIO archive). Since the ELF loader does not tell us if it provided the DTB or just passes one on, reserving this area seems not a bad idea. We an add a comment here with this explanation. In the long adding some adding some flags to the boot interface about the DTB origin might be an option.

src/arch/arm/kernel/boot.c
axel-h: I've split up the commits and remove the parameter name change.

src/arch/arm/kernel/boot.c
axel-h: Or maybe this it just intended to protect the DTB from getting overwritten accidentally by the kerne memory initialization when it needs to allocate memory for internal purposes. Since the DTB from the ELF Loader is likely in an area that is considered free memory, this might happen. What we could do here is recycle the memory like we already do with the kernel boot memory,

PR #423, opened 2021-06-29
github: run C parser on push/PR

lsf37:
This action runs the verification C parser on the main kernel
configurations to check that the code is in the C verification subset.


lsf37:
One more step towards replacing Bamboo.

nomadeel (approved):

PR #424, opened 2021-06-29
meta: add vulnerability disclosure policy

mbrcknl:
This adds a simple vulnerability disclosure policy, loosely based on:
- Disclose.io materials.
- The OWASP Vulnerability Disclosure Cheat Sheet.

I didn't include the typical safe harbour provisions, since I guess that might need legal advice and/or board approval.

Once we agree on the content, I'll make similar pull requests in whichever other repositories we want this to cover.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,99 @@
+<!--
+     Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Vulnerability disclosure policy
+
+Security is a core value of the seL4 Foundation. If you believe you have
+found a security vulnerability in seL4, we ask you to work with us to
+resolve it according to principles of responsible disclosure. This
+policy outlines what we ask of you, and what you can expect of us.
+
+## Scope
+
+This policy currently applies to the most recent released versions, and
+the heads of the default branches of the software in the following seL4
+Foundation repositories:
+- https://github.com/seL4/seL4
+- https://github.com/seL4/capdl
+- https://github.com/seL4/camkes-tool
+
+Note that these repositories include various code generation tools, for
+example the seL4 [bitfield generator], the [capDL-tool] and the
+[camkes-tool]. This policy applies to the code these tools ultimately
+generate, but not to the code generators themselves. In other words,
+it's the code that ends up running on seL4-based systems that we care
+about.
+
+[bitfield generator]: https://github.com/seL4/seL4/blob/master/tools/bitfield_gen.py
+[capDL-tool]: https://github.com/seL4/capdl/tree/master/capDL-tool
+[camkes-tool]: https://github.com/seL4/camkes-tool
+
+## What we ask of you
+
+- Report any vulnerabilities you discover as soon as you can, using the
+  official channel below.
+- Provide enough detail to allow us to understand the vulnerability and
+  its impact.
+- Be patient with us if we have questions.
+- Allow us reasonable time to fix the vulnerability before you disclose
+  it publicly.
+- Be mindful that updating the proofs for a fix might take longer than
+  you imagine.
+- Avoid violating the privacy of others, disrupting the normal operation
+  of our systems, or destroying data.
+
+## What you can expect of us
+
+- We will respond to your report within 14 days, and usually sooner.
+- We will work with you to understand and reproduce the vulnerability
+  you are reporting.
+- We will try to reach agreement on an appropriate timeframe for fixing
+  the vulnerability. We will usually aim for 90 days, but sometimes we
+  will need significantly longer to complete difficult proofs.
+- We will work to fix the vulnerability in a timely manner, and will
+  keep you informed of our progress.
+- When we have developed a fix, we will publicly acknowledge the
+  vulnerability.
+- If you agree, we will publicly acknowledge your role in finding and
+  helping us fix the vulnerability.
+- We will protect your privacy, and will not disclose your personally
+  identifying information to third parties without your explicit
+  permission.
+
+Unfortunately, we are not currently able to offer bug bounties.
+
+## Official channel
+
+Please send vulnerability reports by email to security@seL4.systems,
+optionally encrypted using our security officer's [OpenPGP key].
+
+[OpenPGP key]: https://seL4.systems/security.asc
+
+## FAQ
+
+### How can formally verified software contain security vulnerabilities?
+
+seL4 supports many configurations, and only some of these are verified.
+For unverified configurations, we want to know about vulnerabilities, so
+we can provide the level of support that you would expect from any
+well-engineered software.
+
+Even verified configurations contain some code that is assumed to do the
+right thing, including boot code, assembly stubs, and cache management.
+If that code does the wrong thing, there might be a security
+vulnerability in a verified configuration, and we want to know about
+that.
+
+The proofs also make a number of assumptions about the way hardware
+works. If our assumptions are not valid for real hardware, there might
+be a security vulnerability, and we want to know so we can try to fix
+our assumptions. Some assumptions might not be possible to fix. For
+example, the assumption that memory is incorruptible may be violated by
+Rowhammer, but it is quite fundamental to the proofs.
+
+For more information about what the proofs mean, see the [seL4 FAQ].
+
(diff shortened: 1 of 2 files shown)


SECURITY.md
```diff
@@ -0,0 +1,86 @@
...
+resolve it according to principles of responsible disclosure. This
+policy outlines what we ask of you, and what you can expect of us.
+
+## Scope
+
+This policy currently applies to the most recent released versions, and
+the heads of the default branches of the software in the following seL4
+Foundation repositories:
+- https://github.com/seL4/seL4
+- https://github.com/seL4/seL4
```
axel-h, line 20: https://github.com/seL4/capdl ?

SECURITY.md
```diff
@@ -0,0 +1,86 @@
...
+- https://github.com/seL4/seL4
```
axel-h, line 21: https://github.com/seL4/camkes-tool ?

SECURITY.md
```diff
@@ -0,0 +1,86 @@
...
+
+## What we ask of you
+
+- Report any vulnerabilities you discover as soon as you can, using the
+  official channel below.
+- Provide enough detail to allow us to understand the vulnerability and
+  its impact.
+- Be patient with us if we have questions.
+- Allow us reasonable time to fix the vulnerability before you disclose
```
axel-h, line 30: Google Project Zero gives 90 day. Should we give a time frame here also?

SECURITY.md
```diff
@@ -0,0 +1,86 @@
...
+  identifying information to third parties without your explicit
+  permission.
+
+Unfortunately, we are not currently able to offer bug bounties.
+
+## Official channel
+
+Please send vulnerability reports by email to security@seL4.systems,
+optionally encrypted using the OpenPGP key at
+https://seL4.systems/security.asc.
```
axel-h, line 60: this gives me a 404 Not Found

SECURITY.md
lsf37: I don't think we should commit to a time to fix. We do commit to a time to respond below.

SECURITY.md
lsf37: The key still has to be set up. I agree with the inline URL, that would be a bit nicer.

lsf37 (approved):
Nice! If you could fix up the bits Axel mentioned, I think this is ready to merge already.

I'll set up the gpg key in the meantime.

lsf37:
In terms of which repo: seL4, camkes and capDL are the main ones, I think. We can add more later if the need arises.

We might also want to have this text on docs.sel4.systems somewhere. Ideally not replicated, but included, which might be possible. That should be a separate step, though.

SECURITY.md
lsf37: But maybe we should add a time frame of what we think is reasonable. 90 days might be way too short for a proof update, but it would be Ok for most code updates.

pingerino:
For Camkes and capdl, we should be clear that the vulnerabilities are for specific artifacts - e.g generated code - not the code generators, which aren't trusted, and are likely full of issues.

june-andronick (approved):
Very nice. Looks good to me. Thanks for doing this.

mbrcknl:
> In terms of which repo: seL4, camkes and capDL are the main ones, I think. We can add more later if the need arises.

Whoops, those were the ones I meant to include, but I guess I just pasted the seL4 URL three times.

Should that be `camkes`, or `camkes-tool`?

> For Camkes and capdl, we should be clear that the vulnerabilities are for specific artifacts - e.g generated code - not the code generators, which aren't trusted, and are likely full of issues.

I'll try to draft something.

> We might also want to have this text on docs.sel4.systems somewhere. Ideally not replicated, but included, which might be possible. That should be a separate step, though.

Agreed.

There's also [security.txt](https://securitytxt.org/), which gives a standard way to point to a policy. After this is merged, it might good to drop the following text in a file at https://seL4.systems/.well-known/security.txt, possibly signed (with a clear-text signature):

```
Contact: mailto:security@seL4.systems
Expires: 2022-07-01T00:00:00.000Z
Encryption: https://seL4.systems/security.asc
Canonical: https://seL4.systems/.well-known/security.txt
Policy: https://github.com/seL4/seL4/blob/master/SECURITY.md
```

Though if we do that, we'll have to remember to bump the expiry every year.

SECURITY.md
mbrcknl: @lsf37, by "inline URL", I guess you mean as @axel-h shows above?

lsf37:
> Should that be `camkes`, or `camkes-tool`?

`camkes-tool`, which is naturally where camkes mostly lives :-) (we really have some confusing naming things going on)
 
> > For Camkes and capdl, we should be clear that the vulnerabilities are for specific artifacts - e.g generated code - not the code generators, which aren't trusted, and are likely full of issues.
> 
> I'll try to draft something.

:+1:

> There's also [security.txt](https://securitytxt.org/), which gives a standard way to point to a policy. After this is merged, it might good to drop the following text in a file at https://seL4.systems/.well-known/security.txt, possibly signed (with a clear-text signature):
> 
> ```
> Contact: mailto:security@seL4.systems
> Expires: 2022-07-01T00:00:00.000Z
> Encryption: https://seL4.systems/security.asc
> Canonical: https://seL4.systems/.well-known/security.txt
> Policy: https://github.com/seL4/seL4/blob/master/SECURITY.md
> ```
> 
> Though if we do that, we'll have to remember to bump the expiry every year.

Nice idea. Is the expiry required to be only one year? We can set it to a bit longer (maybe not too long), and have a reminder job ping us when it's getting close to expiry.

SECURITY.md
lsf37: yes

SECURITY.md
mbrcknl: I've added a bit to say we will usually aim for 90 days, but sometimes will need significantly longer.

mbrcknl:
> For Camkes and capdl, we should be clear that the vulnerabilities are for specific artifacts - e.g generated code - not the code generators, which aren't trusted, and are likely full of issues.

@pingerino @lsf37 Can you take a quick look at the bit I added? Does it make sense?

mbrcknl:
> > Though if we do that, we'll have to remember to bump the expiry every year.
> 
> Nice idea. Is the expiry required to be only one year? We can set it to a bit longer (maybe not too long), and have a reminder job ping us when it's getting close to expiry.

It's required to have *some* expiry, and it's recommended to have it no longer than a year. If we expect the URLs to be stable, we could go for a longer expiry, say 3 or 5 years. In any case, it can be done separately to this.

The only things we'd really need in place before merging this are the PGP key and the mailing list.

@wom-bat is security@seL4.systems a suitable choice of email address?

SECURITY.md
```diff
@@ -0,0 +1,99 @@
...
+
+Unfortunately, we are not currently able to offer bug bounties.
+
+## Official channel
+
+Please send vulnerability reports by email to security@seL4.systems,
+optionally encrypted using our security officer's [OpenPGP key].
+
+[OpenPGP key]: https://seL4.systems/security.asc.
+
```
wom-bat, line 74: Make sure this key gets installed before publishing this change.

wom-bat (approved):

SECURITY.md
lsf37: working on it

SECURITY.md
lsf37: That looks good to me!

Xaphiosis:
Looks good. Thank you.

SECURITY.md
```diff
@@ -0,0 +1,86 @@
...
+- Be mindful that updating the proofs for a fix might take longer than
+  you imagine.
+- Avoid violating the privacy of others, disrupting the normal operation
+  of our systems, or destroying data.
+
+## What you can expect of us
+
+- We will respond to your report within 14 days, and usually sooner.
+- We will work with you to understand and reproduce the vulnerability
+  you  are reporting.
```
yyshen, line 52: two spaces after "you"?

mbrcknl: Thanks! Dodgy MacBook keyboard.

SECURITY.md
lsf37: should now be online

mbrcknl: Yep, I've been periodically refreshing, and just noticed. :-)

lsf37:
Ok, the key is up and the only thing failing now is the link to itself, which is fine. Ready to merge from my side.

PR #425, opened 2021-06-29
ARM/RISC-V: boot documentation improvements and cleanup

axel-h:
* remove unused references
* don't return to asm code during boot

### DIFF
--- a/src/arch/arm/32/head.S
+++ b/src/arch/arm/32/head.S
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014, General Dynamics C4 Systems
+ * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -160,10 +161,25 @@ BEGIN_FUNC(_start)
 
     /* Put the DTB address back on the new stack for init_kernel. */
     push {r7, r8}
-    /* Call bootstrapping implemented in C */
+
+    /* Call bootstrapping implemented in C with parameters:
+     *   r0: user image physical start address
+     *   r1: user image physical end address
+     *   r2: physical/virtual offset
+     *   r3: user image virtual entry address
+     *   sp[0]: DTB physical address (0 if there is none)
+     *   sp[1]: DTB size (0 if there is none)
+     */
     blx init_kernel
 
-    /* Restore the initial thread */
+    /* Restore the initial thread. Note that the function restore_user_context()
+     * could technically also be called at the end of init_kernel() directly,
+     * there is no need to return to the assembly code here at all. However, for
+     * verification things are a lot easier when init_kernel() is a normal C
+     * function that returns. The function restore_user_context() is not a
+     * normal C function and thus handled specially in verification, it does
+     * highly architecture specific things to exit to user mode.
+     */
     b restore_user_context
 
 END_FUNC(_start)
--- a/src/arch/arm/64/head.S
+++ b/src/arch/arm/64/head.S
@@ -1,5 +1,6 @@
 /*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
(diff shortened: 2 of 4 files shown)


axel-h:
Can we merge this or is there a reason to leave this in the asm code?

src/arch/arm/32/head.S
```diff
@@ -160,10 +160,15 @@ BEGIN_FUNC(_start)
...
 
-    /* Restore the initial thread */
-    b restore_user_context
+    /* jump to init_kernel() with parameters:
+     * r0: user image physical start addres
+     * r1: user image physical end address
+     * r2: physical/virtual offset
+     * r3: user image virtual entry address
+     * sp[0]: DTB physical address (0 if there is none)
+     * sp[-1]: DTB size (0 if there is none)
```
kent-mcleod, line 170: If this is supposed to be C array syntax, shouldn't the index here be 1 as the stack is decremented on in the push (stmdb sp!) instruction?

axel-h: indeed, that would be more clear. I was wondering what is the best way to describe this.

src/arch/arm/kernel/boot.c
```diff
@@ -626,5 +626,6 @@ BOOT_CODE VISIBLE void init_kernel(
 #endif
     schedule();
     activateThread();
+    restore_user_context();
+    UNREACHABLE();
```
lsf37, line 630: This is probably not a good idea for verification. `restore_user_context()` doesn't have a control flow that is expressible in C, so if we are ever going to verify kernel initialisation, we'll have to remove it again, or at least make it separate from the main `init_kernel` function.

Generally `restore_user_context()` is assumed and dealt with in the machine interface. The only time we do deal with it in C is in the fast path exit, but the fast path is already a very special case, and every new special case for something like this (proof assumptions) is a risk of getting something wrong.

This would have to be a blocker for right now, because kernel init verification is not in the "very soon" project bucket at the moment, but it would have to be changed again when we get there.

lsf37:
> Can we merge this or is there a reason to leave this in the asm code?

Sorry, took me a while to take a deeper look. Yes, the reason we return to asm here is that this makes `init_kernel` a "normal" C function, so verification can treat it as such and leave the assumption part in asm where we already have to do that anyway.

We're not currently verifying `init_kernel`, but we still want to do that at some later point, so we should avoid things we know will definitely break verifcation.

axel-h:
Wouldn't the same apply for the C code in `src/arch/riscv/c_traps.c` then also: `c_handle_interrupt(), `c_handle_exception()`, `slowpath()`?
I can leave this as it was, but then add a comment instead that explains why things are done the way they are.

lsf37:
> Wouldn't the same apply for the C code in `src/arch/riscv/c_traps.c` then also: `c_handle_interrupt()`, `c_handle_exception()`, `slowpath()`?
> I can leave this as it was, but then add a comment instead that explains why things are done the way they are.

Yes, all of `traps.c` is already behind that machine interface assumption, i.e. it's modelled manually outside of C.

axel-h:
Removed the code changes and adding comments instead that explain the design choices

src/arch/arm/32/head.S
```diff
@@ -160,10 +160,25 @@ BEGIN_FUNC(_start)
...
+     */
     blx init_kernel
 
-    /* Restore the initial thread */
+    /* Restore the initial thread. Note that the function restore_user_context()
+     * could technically also be called at the end of init_kernel() directly,
+     * there is no need to return to the assembly code here at all. However, for
+     * verification things are a lot easier when init_kernel() is a normal C
+     * function that returns. The function restore_user_context() is not a
+     * normal C function and handed specially in verification, as it does very
```
lsf37, line 179: small typo:

```suggestion
     * normal C function and handled specially in verification, as it does very
```

src/arch/arm/64/head.S
```diff
@@ -113,9 +113,24 @@ BEGIN_FUNC(_start)
...
+     */
     bl      init_kernel
 
-    /* Restore the initial thread */
+    /* Restore the initial thread. Note that the function restore_user_context()
+     * could technically also be called at the end of init_kernel() directly,
+     * there is no need to return to the assembly code here at all. However, for
+     * verification things are a lot easier when init_kernel() is a normal C
+     * function that returns. The function restore_user_context() is not a
+     * normal C function and handed specially in verification, as it does very
```
lsf37, line 131: (same as above)

```suggestion
     * normal C function and handled specially in verification, as it does very
```

src/arch/riscv/head.S
```diff
@@ -39,7 +38,25 @@ _start:
...
+   *    a7/x17: core ID (SMP only)
+   */
   jal init_kernel
 
+  /* Restore the initial thread. Note that the function restore_user_context()
+   * could technically also be called at the end of init_kernel() directly,
+   * there is no need to return to the assembly code here at all. However, for
+   * verification things are a lot easier when init_kernel() is a normal C
+   * function that returns. The function restore_user_context() is not a
+   * normal C function and handed specially in verification, as it does very
```
lsf37, line 58: (same as above)
```suggestion
   * normal C function and handled specially in verification, as it does very
```

lsf37 (approved):
I'd be happy to merge this now. (A small typo in the comments to fix up still)

axel-h:
typos fixed.

PR #438, opened 2021-07-03
boot: print reserved and available memory regions

axel-h:
Print reserved and available memory regions during boot, this makes porting and debugging a bit easier.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -717,9 +717,12 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
I would be happy with this.

axel-h:
The ugly thing here actually is, that we list the virtual address regions that are not too useful - but still better than nothing. I have a patch set that reworks this ro uses phys regions there, that will be more helpful for debugging then.

lsf37:
You have a lot of patches in your queue, pretty productive :-) I think this is an improvement in any case, so will merge this first.

PR #440, opened 2021-07-04
boot: make insert_region() static

axel-h:
`insert_region()` is not use outside of `boot.c`.

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -41,7 +41,6 @@ bool_t init_freemem(word_t n_available, const p_region_t *available,
                     word_t n_reserved, const region_t *reserved,
                     v_region_t it_v_reg, word_t extra_bi_size_bits);
 bool_t reserve_region(p_region_t reg);
-bool_t insert_region(region_t reg);
 void write_slot(slot_ptr_t slot_ptr, cap_t cap);
 cap_t create_root_cnode(void);
 bool_t provide_cap(cap_t root_cnode_cap, cap_t cap);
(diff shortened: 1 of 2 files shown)


lsf37:
@ssrg-bamboo `proof`

lsf37 (approved):
Looks fine from my side. Have kicked of the proofs just to make sure, but `static` should be fine for the parser, IIRC.

ssrg-bamboo:
Hello, I'm a bot! I've set up a proof testboard for this PR [here](https://github.com/seL4/gh-testboard/blob/axel-h-patch-axel-12-440-012223/manifest.xml), and results should show up as `pending` in status checks in a few minutes.

axel-h:
The Preprocess failures are expected, because this PR adds the "static" keyword.
```
-__attribute__((__section__(".boot.text"))) bool_t insert_region(region_t reg)
+__attribute__((__section__(".boot.text"))) static bool_t insert_region(region_t reg)
```

axel-h:
@lsf37: seem the proofs fail due to resource shortage (`An error occurred (MaxSpotInstanceCountExceeded) when calling the RunInstances operation: Max spot instance count exceeded`), as you mentioned in https://github.com/seL4/seL4/pull/457#issuecomment-881044260

axel-h:
I had to extend this patch, `create_rootserver_objects()` is also not used outside of `boot.c` and thus it can be `static` also. Removing the definition from `boot.h` in this PR will comes handy for some follow-up patches where the signature changes to return a `bool_t` in case of an error - but that leaves `boot.h unmodified then.

axel-h:
Can we merge this?

lsf37:
There is now a small conflict with the other boot refactor. If you resolve and rebase, I'm happy to merge this one.

axel-h:
Rebased and fixed the conflict. Also improved the comment about `create_rootserver_objects()` a bit.

lsf37:
This now does appear to break the proof on ARM, which is really strange. I'll have a closer look.

lsf37:
It appears the proof test started before the verification manifest had updated itself to the proof for the latest seL4 master. Rerunning..

PR #441, opened 2021-07-04
boot: use SEL4_PRIx_word instead of 'lx'

axel-h:


mbrcknl:
Where is `SEL4_PRIx_word` defined?

axel-h:
@mbrcknl: It's in `include/basic_types.h` and `libsel4/include/sel4/simple_types.h`. It's still an open point if using one common location would be possible.

mbrcknl:
> @mbrcknl: It's in `include/basic_types.h` and `libsel4/include/sel4/simple_types.h`.

Oh, no wonder my recursive `grep` didn't find it... I was in the `l4v` directory! :-)

> It's still an open point if using one common location would be possible.

I guess we don't try to solve that here.

mbrcknl (approved):

axel-h:
- rebased
- commit comment is even more verbose now.

lsf37 (approved):

PR #443, opened 2021-07-04
create empty array for kernel_devices

axel-h:
Do not set `kernel_devices` to `NULL` but create an empty array, so the `ARRAY_SIZE()` macro can be used. This avoids any needs for special handling and the HiFive/Spike hack on RISC-V can be removed.


axel-h:
Seem this is another thing the C parser does not like ...?

lsf37:
I'm going by the description only, but an empty (zero length) array would indeed be something the C parser does not support.

axel-h:
@lsf37: the CI action does not print any output from the C parser - is there any that might help understanding what exactly it is complaining about?

kent-mcleod:
> Seem this is another thing the C parser does not like ...?

It's also because when following the strict C standard, zero length array declarations are not allowed: https://en.cppreference.com/w/c/language/array (See the notes section at the bottom)

Setting the array to null is usually how we've previously encoded zero length.  It does make it necessary to carry around functions for getting the size of the array, but it's not possible to use a sizeof macro for computing array size if it is possible that the array could be empty.

lsf37:
> @lsf37: the CI action does not print any output from the C parser - is there any that might help understanding what exactly it is complaining about?

It's the C parser itself in this case -- it just silently fails, which is not great. The action passes on any output it gets from the parser, and usually there is at least a hint. I'll put up an issue in l4v for this case (0-length array).

axel-h:
Looks like the C parser is happy with this solution now. What happens on RISC-V/HiFive still seems a bit odd to me, but this is a separate issue then.

axel-h:
After look a bit into the history of the kernel code I wonder if `kernel_device_regions` is actually `kernel_frames` because it is (was?) intended to me more generic actually. I've seen that some about 3 years ago the auto-generated array `dev_p_reg` was removed from the code. But somehow I still don't fully understand why this wasn't added to `kernel_device_regions` then. Instead there seem some magic going on that tries to determine this automatically now? And `kernel_device_regions` still contains peripheral mappings only.

src/arch/arm/machine/hardware.c
```diff
@@ -31,8 +31,12 @@ BOOT_CODE const p_region_t *get_avail_p_regs(void)
...
 BOOT_CODE void map_kernel_devices(void)
 {
-    for (unsigned int i = 0; i < ARRAY_SIZE(kernel_devices); i++) {
-        const kernel_frame_t *frame = &kernel_device_regions[i];
+    /* If there are no kernel frames at all, then kernel_frames is NULL. Thus we
+     * can't use ARRAY_SIZE(kernel_frames) here directly, but have to use
+     * NUM_KERNEL_FRAMES that is defined accordingly.
+     */
+    for (unsigned int i = 0; i < NUM_KERNEL_FRAMES; i++) {
+        const kernel_frame_t *frame = &kernel_frames[i];
```
pingerino, line 39: I don't understand why this is being renamed. We aren't mapping generic frames, we're mapping devices, in a function called `map_kernel_devices`. Seems like more needless churn. 

Also, this commit does a lot more than is said in the commit message. The part it doesn't admit to doing is the useful part of the commit that should be kept.

src/arch/arm/machine/hardware.c
```diff
@@ -32,17 +32,15 @@ BOOT_CODE const p_region_t *get_avail_p_regs(void)
...
-                         VMKernelOnly,
-                         vm_attributes_new(kernel_devices[i].armExecuteNever,
-                                           false, false));
-        if (!kernel_devices[i].userAvailable) {
-            p_region_t reg = {
-                .start = kernel_devices[i].paddr,
-                .end = kernel_devices[i].paddr + BIT(PAGE_BITS),
-            };
-            reserve_region(reg);
+        const kernel_frame_t *frame = &kernel_device_regions[i];
```
pingerino: this commit also seems like needless churn. A code review and a commit to save 2 lines of code by using an alias?

pingerino:
Looks good, thanks for contributing the improvements. 

Please try to minimise stylistic code changes that aren't necessary though, to save everyone's time, and make sure the commits do exactly what they say they do, for history preserving purposes and to make life easier for reviewers. 

Every line of code needs to be scrutinised by reviewers. Every line of code also needs to be rebased over potentially by someone on a feature branch. It's not cheap to refactor.

pingerino (changes requested):

pingerino:
> After look a bit into the history of the kernel code I wonder if `kernel_device_regions` is actually `kernel_frames` because it is (was?) intended to me more generic actually. I've seen that some about 3 years ago the auto-generated array `dev_p_reg` was removed from the code. But somehow I still don't fully understand why this wasn't added to `kernel_device_regions` then. Instead there seem some magic going on that tries to determine this automatically now? And `kernel_device_regions` still contains peripheral mappings only.

It used to be a hard coded array of devices for each platform, until we refactored it to be generated from the device tree to make adding new platforms much easier. It has always been memory for devices though.

src/arch/arm/machine/hardware.c
axel-h: Thanks for the feedback, I've split up the commit further also the clarified renaming. This was a misunderstanding on my side in the first place. I have picked the name `kernel_device_frame` now because this seems to be what the structure contains. A device can use more than one page, so this structure can have more elements than kernel devices and this must not be mixed up. I've see the case on RISC-V, where a peripheral has 48 KiByte and there are 12 entries here. Not sure is there is a similar case on an ARM platform,

src/arch/arm/machine/hardware.c
axel-h: It's not about saving 2 lines of code, and I also have string faith in the compiler to be smart enough to detect that it can cache the pointer instead of calculating is everytime. This is about removing redundancy from the code by using a helper alias variable, because the array element is used multiple times. That make the code easier to understand as is separates what is done with the array element from the fact that this is an element from an array.
I also thing somebody mentioned that this pattern makes things a bit easier for verification, but I might be wrong here if have ripped this out of the original context.

axel-h:
I've split up the commit and extended the commit message to state what I'm up to there. Thingd should not build towards the final commit. I have two more commits on top of this that remove `MODE_RESERVED` in https://github.com/axel-h/seL4/commits/patch-axel-16a but will make a separate PR for this once this is merged.

src/arch/arm/machine/hardware.c
pingerino: I'm not convinced optimising boot code is worth the review time, nor the very minor readability improvement. My main point is there's a cost to every PR line of code, so if we can reduce the unnecessary changes that would reduce reviewer load and also get your changes in faster. However it's done so I'll approve it. Please be aware changes like this reduce the reviewer bandwidth so other more important changes are delayed.

pingerino (approved):

axel-h:
rebased

axel-h:
Can we merge this?

PR #445, opened 2021-07-06
risc-v: add COMMON section to linker script

axel-h:
Follow up from a comment from @kent-mcleod at https://github.com/seL4/seL4/pull/421#issuecomment-874375725

With gcc < v10, global variables defined but not initialized get put into the COMMON section. If this section is not declared explicitly, these variables end up at the end of the .bss, which wasts a page.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,15 @@ add_compile_options(
     -Wpointer-arith
     -Wno-nonnull
     -ffreestanding
+    # GCC < 10 and clang < 11 put uninitialized global variables into a 'COMMON'
+    # section unless '-fno-common' is specified. The linker will put anything
+    # from 'COMMON' as the end of the '.bss' it nothing else is specified in the
+    # linker script. Besides making the variable placement look odd, this also
+    # tends to waste a page because we puts large aligned block at the end.
+    # Eventually, GCC 10 and clang 11 made '-fno-common' the default, see
+    # - https://gcc.gnu.org/gcc-10/changes.html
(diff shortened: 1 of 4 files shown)


kent-mcleod (approved):

axel-h:
@mbrcknl, @lsf37: since this changes the linker file and thus affects the binary, what verification should run to merge this?

lsf37:
Ideally we'd want the hardware test to run for this (which unfortunately are down now). Can you confirm that this works for yo on your riscv boards?

On the formal side, the linker behaviour is in the assumptions, so nothing to do there.

lsf37 (approved):

mbrcknl:
I don't know much about linker configuration. Would these symbols in the `COMMON` section still be emitted in the symbol  table?

Binary verification needs `kernel.elf.symtab` to compare accesses via variable names in the C with accesses via pointers in the binary. As long as they're somewhere in the symbol table, BV will be fine. It won't matter which section, or where in the section.

axel-h:
Added `*(COMMON)` to the ad linker script also. Improved the comments based on https://gcc.gnu.org/gcc-10/changes.html.

mbrcknl:
> * Improved the comments based on https://gcc.gnu.org/gcc-10/changes.html.

That doc makes me wonder if it would be better to just add `-fno-common` to all GCC builds, to get more consistency between pre-v10 and post-v10.

axel-h:
Good point, maybe we should do both. And check what clang does.

axel-h:
So, what do you think now?

PR #447, opened 2021-07-08
Adds support for the arm Cortex A55 and ODroid C4

xurtis:
This adds support for the ODroidC4. The ODroidC4 is just a beefier ODroidC2 with some bits moved around. The devices we care about here are all essentially the same.

Related: https://github.com/seL4/util_libs/pull/97 and https://github.com/seL4/seL4_tools/pull/95

### DIFF
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -20,6 +20,10 @@ elseif(KernelArmCortexA53)
     set(KernelArmICacheVIPT ON)
     set(KernelArmPASizeBits40 ON)
     math(EXPR KernelPaddrUserTop "(1 << 40)")
+elseif(KernelArmCortexA55)
+    set(KernelArmICacheVIPT ON)
+    set(KernelArmPASizeBits40 ON)
+    math(EXPR KernelPaddrUserTop "(1 << 40)")
 elseif(KernelArmCortexA57)
     set(KernelArmPASizeBits44 ON)
     math(EXPR KernelPaddrUserTop "(1 << 44)")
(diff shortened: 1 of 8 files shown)


src/arch/arm/config.cmake
```diff
@@ -20,6 +20,10 @@ elseif(KernelArmCortexA53)
     set(KernelArmICacheVIPT ON)
     set(KernelArmPASizeBits40 ON)
     math(EXPR KernelPaddrUserTop "(1 << 40)")
+elseif(KernelArmCortexA55)
+    # set(KernelArmICacheVIPT ON)
```
axel-h, line 24: Why is this commented out?

src/plat/odroidc4/config.cmake
```diff
@@ -0,0 +1,35 @@
...
+declare_platform(odroidc4 KernelPlatformOdroidc4 PLAT_ODROIDC4 KernelSel4ArchAarch64)
+
+if(KernelPlatformOdroidc4)
+    declare_seL4_arch(aarch64)
+    set(KernelArmCortexA55 ON)
+    set(KernelArchArmV8a ON)
+    config_set(KernelARMPlatform ARM_PLAT odroidc4)
+    set(KernelArmMachFeatureModifiers "+crc" CACHE INTERNAL "")
+    list(APPEND KernelDTSList "tools/dts/odroidc4.dts")
+    list(APPEND KernelDTSList "src/plat/odroidc4/overlay-odroidc4.dts")
```
axel-h, line 18: could be merged `list(APPEND KernelDTSList "tools/dts/odroidc4.dts" "src/plat/odroidc4/overlay-odroidc4.dts")`

src/arch/arm/config.cmake
xurtis: Not clear on whether it is needed. The manual states the D-Cache is VIPT but doesn't mention anything about the I-Cache.

axel-h: file:///home/axehei01/Downloads/temp/cortex_a55_trm_100442_0200_00_en.pdf at "B1.27 CTR, Cache Type Register" says 
```
L1Ip, [15:14]
Instruction cache policy. Indicates the indexing and tagging policy for the L1 Instruction cache:
0b10 Virtually Indexed Physically Tagged (VIPT). 
```

xurtis: Ty, uncommented.

nomadeel (approved):
Aside from axel-h's issues, looks good to me.

PR #450, opened 2021-07-11
remove redundant defines

axel-h:
These defines seem redundant, they exist the architecture specific `bootinho.h` already and this seems the better place due to the subtle differences

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -8,16 +8,6 @@
 #include <bootinfo.h>
 #include <arch/bootinfo.h>
 
-#ifndef CONFIG_ARCH_ARM
-#define MAX_NUM_FREEMEM_REG 16
-#else
-/* Modifiers:
- *  + 1: allow the kernel to release its own boot data region
- *  + 1: possible gap between ELF images and rootserver objects;
- *       see arm/arch_init_freemem */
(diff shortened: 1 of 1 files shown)


axel-h:
Seems even github has hiccups `fatal: unable to access 'https://github.com/seL4/seL4.git/': Failed to connect to github.com port 443: Connection timed out`

lsf37:
> Seems even github has hiccups `fatal: unable to access 'https://github.com/seL4/seL4.git/': Failed to connect to github.com port 443: Connection timed out`

Huh, I've seen a few of these recently locally, but I though that was my home network. Maybe it wasn't. I'll rerun the tests.

kent-mcleod:
Context: https://sel4.discourse.group/t/why-is-max-num-freemem-reg-defined-in-two-different-places/330

kent-mcleod (approved):

axel-h:
@kent-mcleod: in my local patch set I'm considering deleting the architecture's `bootinho.h` because
- the content could be put into the existings architecture's `arch/kernel/boot.h` also
- there is a architecture agnostic `bootinfo.h` and having the `arch/bootingo.h`files is confusing.

axel-h:
Test SCHED0021 failed on PC99_debug_gcc_64, but I think this is related to my changes here.

lsf37:
The simulation test worked after re-run which confirms that it's the known SCHED0021 issue.

PR #451, opened 2021-07-11
clarify that parameter is initial thread's region

axel-h:
Align the parameter naming within each architecture's file and also across architectures.


mbrcknl (approved):

axel-h:
Rebased. Preprocess fails because of the renaming of the function parameter.

axel-h:
Based on the work in my other patches, I'm extending this patch to do the parameter renaming in a few more places.

axel-h:
Can we merge this?

PR #452, opened 2021-07-11
Another round of code cleanup and readabity improvements

axel-h:
This is another patch extracted from from a bigger change set, that aims at improving the code quality and readability. None of this is really necessary at a pure technical level, but it's supposed to make the code easier to understand and re-usable.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -21,6 +21,7 @@
  */
 #define UL_CONST(x) x
 #define ULL_CONST(x) x
+#define NULL 0
 
 #else /* not __ASSEMBLER__ */
 
@@ -31,6 +32,7 @@
  */
 #define UL_CONST(x) PASTE(x, ul)
 #define ULL_CONST(x) PASTE(x, llu)
+#define NULL ((void *)0)
 
 #endif /* [not] __ASSEMBLER__ */
 
@@ -66,8 +68,6 @@
 #define HZ_IN_MHZ   ULL_CONST(1000000)
 
 #ifndef __ASSEMBLER__
-
-#define NULL ((void *)0)
 
 #define PACKED       __attribute__((packed))
 #define NORETURN     __attribute__((__noreturn__))
--- a/include/util.h
+++ b/include/util.h
@@ -264,4 +264,3 @@ CONST popcountl(unsigned long mask)
 #define UL_CONST(x) x
 
 #endif /* !__ASSEMBLER__ */
-
--- a/include/util.h
+++ b/include/util.h
@@ -6,6 +6,33 @@
 
 #pragma once
 
+#define PASTE(a, b) a ## b
+#define _STRINGIFY(a) #a
+#define STRINGIFY(a) _STRINGIFY(a)
+
+#ifdef __ASSEMBLER__
+
+/* Provide a helper macro to define integer constants that are not of the
+ * default type 'ìnt', but 'unsigned long'. When such constants are shared
+ * between assembly and C code, some assemblers will fail because don't support
+ * C-style integer suffixes like 'ul'. Using a macro works around this, as the
+ * suffix is only applies when the C compiler is used and dropped when the
+ * assembler runs.
+ */
+#define UL_CONST(x) x
+
+#else /* not __ASSEMBLER__ */
+
+/* There is no difference between using 'ul' or 'lu' as suffix for numbers to
+ * enforce a specific type besides the default 'int'. Just when it comes to the
+ * printf() format specifiers, '%lu' is the only form that is supported. Thus
+ * 'ul' is the preferred suffix to avoid confusion.
+ */
+#define UL_CONST(x) PASTE(x, ul)
+
+#endif /* [not] __ASSEMBLER__ */
+
+#define BIT(n) (UL_CONST(1) << (n))
 #define MASK(n) (BIT(n) - UL_CONST(1))
 #define IS_ALIGNED(n, b) (!((n) & MASK(b)))
 #define ROUND_DOWN(n, b) (((n) >> (b)) << (b))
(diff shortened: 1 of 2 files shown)


lsf37 (approved):

lsf37:
I'd like to use this Pr to get the proof test fully working, if that's Ok (see comments at the bottom of #457)

include/util.h
```diff
@@ -13,27 +13,27 @@
...
 #endif /* [not] __ASSEMBLER__ */
 
-#define MASK(n) (BIT(n)-UL_CONST(1))
-#define IS_ALIGNED(n, b) (!((n) & MASK(b)))
-#define ROUND_DOWN(n, b) (((n) >> (b)) << (b))
-#define ROUND_UP(n, b) (((((n) - UL_CONST(1)) >> (b)) + UL_CONST(1)) << (b))
-#define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
-#define MIN(a,b) (((a)<(b))?(a):(b))
-#define MAX(a,b) (((a)>(b))?(a):(b))
+#define MASK(n)           ( BIT(n) - UL_CONST(1) )
```
pingerino, line 33: @axel-h  this isn't part of our style guide and shouldn't be merged. If you want to propose a change to the style guide, do so, but it's not an argument to have on a PR.

include/util.h
```diff
@@ -24,9 +25,11 @@
  */
 #define UL_CONST(x)   PASTE(x, lu)
 #define ULL_CONST(x)  PASTE(x, llu)
+#define NULL          ((void *)0)
```
pingerino, line 28: Does removing the void* not cause any warnings?

Not sure how you would tell unless you are compiling for all platforms.

include/util.h
```diff
@@ -14,10 +14,16 @@
...
 
 #else /* not __ASSEMBLER__ */
 
-#define UL_CONST(x) PASTE(x, ul)
+/* There is no difference between using 'ul' or 'lu' as number postfix. Just
+ * when it comes to the printf() format specifiers, '%lu' is the only form that
+ * is supported.
+ */
+#define UL_CONST(x) PASTE(x, lu)
+#define ULL_CONST(x) PASTE(x, llu)
```
pingerino: I honestly don't think adding a layer of indirection around `llu` is improving code readability. This might be a convention that you are used to, but for me it's harder to read.

include/util.h
```diff
@@ -6,16 +6,28 @@
...
 #pragma once
 
+#define PASTE(a, b) a ## b
+#define _STRINGIFY(a) #a
+#define STRINGIFY(a) _STRINGIFY(a)
+
+#ifdef __ASSEMBLER__
+
+/* Some assemblers don't recognise the ul/ull (unsigned long) suffix */
+#define UL_CONST(x) x
```
pingerino: ok, if this is the reason for using the macros (using the constants in assembler) it makes a lot more sense, can you add that to the initial commit?

pingerino (changes requested):

pingerino:
there's a merge commit in here that needs removing

lsf37:
> there's a merge commit in here that needs removing

The GitHub merge commits from master disappear in the rebase merge that GitHub does when you merge (if those are the only merge commits). It's still nicer to rebase before merging to make sure and I'd recommend it to see the final commits, but it's not strongly required for avoiding merge commits in master.

include/util.h
lsf37: Hm, yes, it might, you're right.

I'll see if I can get at least the build part of the HW tests going so we can see this, but it would be good to not remove the cast in any case in the assembler version. @axel-h or is this something a specific assembler tripped up on for you?

include/util.h
```diff
@@ -13,27 +13,27 @@
...
-#define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
-#define MIN(a,b) (((a)<(b))?(a):(b))
-#define MAX(a,b) (((a)>(b))?(a):(b))
+#define MASK(n)           ( BIT(n) - UL_CONST(1) )
+#define IS_ALIGNED(n, b)  ( !((n) & MASK(b)) )
+#define ROUND_DOWN(n, b)  ( ((n) >> (b)) << (b) )
+#define ROUND_UP(n, b)    ( ((((n) - UL_CONST(1)) >> (b)) + UL_CONST(1)) << (b) )
+#define ARRAY_SIZE(x)     ( sizeof(x) / sizeof(x[0]) )
+#define MIN(a,b)          ( ((a)<(b))?(a):(b) )
+#define MAX(a,b)          ( ((a)>(b))?(a):(b) )
```
yyshen, line 39: why are additional spaces introduced for the outer-most parenthesis?

include/util.h
axel-h: `NULL` is define as `0` if `__ASSEMBLER__` is defined, as the assembler does not know C types. Otherwise is remains as `(void*)0`. Thast allows having constants shares between C and assembly code to use NULL.

include/util.h
axel-h: ok, will drop the reformatting

include/util.h
axel-h: I'm just following what already existed in the code and adapt the changes I brought in a while ago to this. I do not have any preference for one was or the other. The argument for these macros is, that constants defined with `UL_CONST(x)` or `ULL_CONST(x)` can also be used in the assembly code easily, as the macro is defined to just drop any postfix there. If we use `PASTE(x, something)` directly, there is no way to use such constants in assembly code,`PASTE()` is a helper macro for the preprocessor that is used in other places also and we can't define this to drop the `something` there.

include/util.h
axel-h: I will remove this reformatting since it is not in the style guide. I picked up the habit to put a space between the outmost bracket in macros to clearly show the whole expression macro is expression is made an atom.

include/util.h
axel-h: done

axel-h:
Thanks to everybody for the feedback, I've reworked the PR to address the remarks and droppe the macro re-formatting.

pingerino (approved):

lsf37:
I'm happy with the proof test for now (some cosmetic improvements for later), and I can also confirm that a build for all platforms succeeds (a PR for that test is up as well)

PR #453, opened 2021-07-12
mcs: Don't move head past max time on unblock

xurtis:
`refill_unblock_check` didn't preserve the property that the head release of a scheduling context was always less than the `MAX_RELEASE_TIME`. This aborts delaying releases once the system time has passed `MAX_RELEASE_TIME`.


xurtis:
This is not needed as we assume that [`updateTimestamp` will never set `ksCurTime` to be greater than `MIN_RELEASE_TIME`](https://github.com/seL4/seL4/blob/caf6eb2a6b51f7e31e0e190acd7259438dfb0662/include/kernel/thread.h#L228).

PR #454, opened 2021-07-12
rpi3: Mark first memory page as reserved

sannum:
When the kernel is built the build system is responsible for finding
a suitable physical memory location where the kernel can be put on
the given target platform.
This is recorded into the kernel.elf file and read by the elfloader
program when it loads the kernel to that memory address.

In order to find memory blocks to avoid, the build system looks for
the /reserved-memory node in the target platform's device tree dts,
along with other device memory blocks.

The bcm2837 / Raspberry Pi 3 bootloader uses the first memory page
at address 0x0 to load a so called armstub which is used to set up
the ARM processor's initial state. It is also used to "park" the
secondary cores by putting them in a spin loop located within the
armstub from which the boot core can release them when ready.

The rpi3.dts already contains a /memreserve/ node reserving this
page, however as the build system only looks for the standardized
reserved-memory node it promptly disregards it and allow the kernel
to be loaded at physical address 0x0, overwriting the armstub.

A side effect of this is that the spinloop code also is overwritten,
potentially releasing the secondary cores to execute whatever kernel
code is written in the place of their spinloops, causing all kinds
of undefined behavior dependent on both race conditions and kernel
elf layout. It also implies that the kernel SMP boot code would not
be able to release the cores if implemented for the platform.

This patch adds the /reserved-memory node to the overlay-rpi3.dts
file and a child node reserving the memory region for the first
memory page. This in effect causes the kernel to instead be loaded
to 0x1000000 (aligned to a supersection).

Signed-off-by: Viktor Sannum <sannum.viktor@gmail.com>


src/plat/bcm2837/overlay-rpi3.dts
```diff
@@ -20,4 +20,14 @@
 		/* This is configurable in the Pi's config.txt, but we use 128MiB of RAM by default. */
 		reg = <0x00 0x08000000>;
 	};
+
+	reserved-memory {
```
axel-h, line 24: Please add a comment here explaining in detail why this memory is reserved.

axel-h:
Doesn't the ELF loader get the other cores out of this park loop there and makes them spin in the kenrel's init code then? IN Which scenarios have you seen that the cores are parked in and the start doing strange things?

src/plat/bcm2837/overlay-rpi3.dts
sannum: Fixed!

sannum:
> Doesn't the ELF loader get the other cores out of this park loop there and makes them spin in the kenrel's init code then? IN Which scenarios have you seen that the cores are parked in and the start doing strange things?

Not when SMP is not enabled as elfloader is agnostic to the other core's existence in that case. The rpi3 kernel platform does not yet support SMP as it is still unimplemented for it. 

There is not yet any smp driver for the rpi3 platform either which would be responsible for doing just that, bringing them out of the spin loop and into the kernel (this is partly what I'm currently working on).

The reason I found this bug was that I was attempting to move a kernel source file location withing the source tree. This somehow caused the kernel to do an undefined instruction abort just as it was entered from the elfloader at runtime. 

It turned out that moving the file must have changed the layout of my kernel binary such that the code placed in the secondary thread spin loops caused them to execute instructions writing 0x0 (udf in arm assembly) to physical addresses 0x8 and 0xc which were early instructions just after the kernel entry point. This race happened just after elfloader had loaded the kernel to 0x0 and before it jumped to that entry point.

I guess that without the layout changed that I tried to introduce. The secondary cores ended up doing something else which did not cause the kernel to abort. If that is the more common outcome it would explain how this could fly under the radar.

sannum:
Also note that, since the interrupt vectors are installed at address 0x0, the same early kernel boot code loaded to that area would once again be implicitly overwritten when the interrupt vector installation happens later on in the boot process.

Usually that is not a problem since the kernel entry point seems to end up at the start of the image in most cases and the code overwritten then is part of the at that point already run, run-once crt. But I think overwriting it should be done explicitly in that case.

src/plat/bcm2837/overlay-rpi3.dts
```diff
@@ -20,4 +20,15 @@
...
+		#address-cells = < 0x01 >;
+		#size-cells = < 0x01 >;
+		ranges = < >;
+/* Memory page holding the bootloader armstub and secondary core parking spinloops */
```
axel-h, line 28: I think this comment should be much more verbose explaining all the things you also mentioned in this PR. Please note that the PR and commit comment may not be available to somebody looking at this code in the future. And the importance of this when *not* building for SMP is not this obvious.

sannum: Added a more in depth explanation to the comment. Do you think this take covers the issue good enough or did I miss any important aspect?

sannum:
Here is a good readup on exactly what the armstub does and how the secondary cores are parked by the bootloader https://leiradel.github.io/2019/01/20/Raspberry-Pi-Stubs.html#armstub8s for anyone interested.

src/plat/bcm2837/overlay-rpi3.dts
```diff
@@ -20,4 +20,21 @@
 		/* This is configurable in the Pi's config.txt, but we use 128MiB of RAM by default. */
 		reg = <0x00 0x08000000>;
 	};
+
+	reserved-memory {
+		#address-cells = < 0x01 >;
+		#size-cells = < 0x01 >;
+		ranges = < >;
+/*
```
axel-h, line 28: nitpicking: please indent this comment

src/plat/bcm2837/overlay-rpi3.dts
```diff
@@ -20,4 +20,21 @@
...
+ * Memory page holding irq vector as well as the bootloader armstub and secondary core parking spinloops.
```
axel-h, line 29: I suggest to reword the comment:
```
Keep the first page of physical memory reserved for 
the initial bootloader (e.g. armstub). It has parked
the secondary cores there, they spin until they get
released. In SMP configuration, the seL4 kernel will
release them duing boot and this memory could be 
used - but we still have to ensure the kernel image
itself is not placed here. In non-SMP configurations,
the cores must keep sinning forever. Re-using this
memory will cause the secondary cores to execute 
whatever content is placed there, which likely makes
them run amok.
```

axel-h:
@sannum from my side this would be good to get merge now.

lsf37:
> @lsf37, do you know why the DCO isn't running?

No, first time I've seen this happening. Might be an issue with the app itself (server not reachable or similar). It did pass before, so it's fine to override the merge.

PR #455, opened 2021-07-14
Introduce banked IRQ addressing

sannum:
Currently, kernel interrupt controller drivers expose a linear space
of IRQ numbers which the kernel and userspace uses to address specific
IRQs.

Some platform uses multi-stage interrupt controllers and / or banked
IRQs where IRQ indices are relevant only to the relevant interrupt
controller and / or IRQ bank within a controller.

An example of such is the bcm2837 / rpi3 platform which uses two
interrupt controllers in tandem and the second stage interrupt
controller contains 3 differently sized banks of IRQs.

Currently, the IRQ spaces that those controller and banks represent
must be mapped into a single linear space. In the case of bcm2837
/ rpi3, they are mapped sequentially after one another.

This mapping is in fact arbritrary. In userspace, interrupts as
parsed from the device tree are addressed relative a specific
parent interrupt controller device and optional bank. Drivers
must somehow know how the kernel driver maps this to the linear
IRQ space as used by the kernel. This creates an implicit API
between userspace and the kernel interrupt controller driver
for a particular platform interrupt controller.

This commit implements a generic explicit API for handling this
by introducing the shared type seL4_BankedIRQ and the macro
BANKED_IRQ_TO_IRQ.

seL4_BankedIRQ encodes an interrupt controller device, an IRQ bank
and IRQ within a word. It can be used in place of IRQ numbers wherever
currently used in the seL4 API. This is similar to how seL4_CapRights
or seL4_MessageInfo are used. The controller device value is large
enough to address a physical memory page used by a interrupt controller.
The bank value supports 8 banks per interrupt controller and the IRQ
value can represent up to 512 IRQs per bank.

The BANKED_IRQ_TO_IRQ should be implemented by platform IRQ
drivers to map seL4_BankedIRQ:s to the linear IRQ space used
internally by the kernel. The IRQ driver may do that mapping
at its discretion in a way which makes sense according to what
addressing is used in device trees for that platform. This is
similar to how PPIs (in practice per core banking) is supported
using the CORE_IRQ_TO_IRQT macro.

This change should be completely backwards compatible, platforms
which do not care about staged interrupt controllers or banked
irqs do not need to implement the BANKED_IRQ_TO_IRQ macro and
can continue to rely on the existing implicit mapping.
Non-seL4_BankedIRQ values can still be used with the existing
APIs.

Signed-off-by: Viktor Sannum <sannum.viktor@gmail.com>

### DIFF
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -90,6 +90,19 @@ static inline word_t CONST wordFromMessageInfo(seL4_MessageInfo_t mi)
     return mi.words[0];
 }
 
+static inline seL4_BankedIRQ_t CONST bankedIrqFromWord(word_t w)
+{
+    seL4_BankedIRQ_t irq;
+
+    irq.words[0] = w;
+    return irq;
+}
+
+static inline word_t CONST wordFromBankedIrq(seL4_BankedIRQ_t irq)
+{
+    return irq.words[0];
+}
+
 #ifdef CONFIG_PRINTING
 #ifdef CONFIG_COLOUR_PRINTING
 #define ANSI_RESET "\033[0m"
--- a/libsel4/mode_include/32/sel4/shared_types.bf
+++ b/libsel4/mode_include/32/sel4/shared_types.bf
@@ -8,6 +8,12 @@
 
 base 32
 
+block seL4_BankedIRQ {
+    field interrupt_controller 20
+    field bank 3
+    field irq 9
+}
+
 block seL4_MessageInfo {
     field label 20
     field capsUnwrapped 3
--- a/libsel4/mode_include/64/sel4/shared_types.bf
+++ b/libsel4/mode_include/64/sel4/shared_types.bf
@@ -8,6 +8,12 @@
 
 base 64
 
+block seL4_BankedIRQ {
+    field interrupt_controller 52
+    field bank 3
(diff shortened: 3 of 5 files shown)


sannum:
Since this PR affects the API I'm very open for suggestions and discussion on how this can be implemented in the best way possible. There are other options such as introducing new API calls (similar to seL4_IRQControl_GetTrigger) or allowing IRQ drivers to generate additional IRQControl caps for each IRQ Controller / bank which is then given to the roottask to use as it sees fit. The proposed implementation is however the most unintrusive I could come up with.

axel-h:
If we assume the device tree is the only source of truth and it is available to every piece of (driver) software, what would this improve? An "seL4 SDK" (e.g. for bare metal or CAmkES) would comes with a device tree parsing script that would know how to serialize the interrupts for each platform. Is a platform specific API contract about the interrupt then really an issue? How is this handled on e.g. Linux for drivers and the actual interrupt number?

kent-mcleod:
Hi @sannum, I'm not sure if this is something you are still interested in pursuing. If so, it would be better to bring this up as a new discussion thread on https://sel4.discourse.group/c/rfc-discussion/9 and linking to your branch as an example starting point would be helpful.

PR #457, opened 2021-07-15
C proof check for labelled PR

lsf37:
This action will run the C proofs when the label `proof-test` is added to a pull request. Intended to be used when the preprocess test has failed and the PR is ready to check proofs.

The idea is that this replaces the ssrg-bamboo bot for proofs.

This test should run much faster than a full proof test board, with the same predictive power. It will likely need a bit longer on the first run, but after that use cached proof images for things that have not changed.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -1,0 +1,37 @@
+# Copyright 2021 Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+name: Proofs
+
+on:
+  pull_request_target:
+    types: [labeled, synchronize]
+
+jobs:
+  cproof:
+    name: C Proofs
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'proof-test')
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ARM, ARM_HYP, RISCV64, X64]
+    steps:
+    - name: Proofs
+      uses: seL4/ci-actions/aws-proofs@master
+      with:
+        L4V_ARCH: ${{ matrix.arch }}
+        isa_branch: ts-2020
+        session: CRefine SimplExportAndRefine
+        manifest: default.xml
+      env:
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        AWS_SSH: ${{ secrets.AWS_SSH }}
+        GH_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+    - name: Upload logs
+      uses: actions/upload-artifact@v2
+      with:
+        name: logs-${{ matrix.arch }}
+        path: logs.tar.xz


axel-h:
How "expensive" is this? Can one use this freely?

Xaphiosis:
Does this utilise `SKIP_DUPLICATED_PROOFS`?
Is `proof-test` clear enough that it only runs the C-related proofs? If at some distant point in the future we pull constants from the CMake config into the abstract spec, `seL4` PRs could cause issues there. If we're being paranoid, could call this `c-proof-test`.
Anyway, good idea.

Xaphiosis (approved):

axel-h:
Ture, `c-proof-test` seem more precise then

lsf37:
> Does this utilise `SKIP_DUPLICATED_PROOFS`?

Yes, that is on by default.

> Is `proof-test` clear enough that it only runs the C-related proofs? If at some distant point in the future we pull constants from the CMake config into the abstract spec, `seL4` PRs could cause issues there. If we're being paranoid, could call this `c-proof-test`

`proof-test` was deliberate: The idea is that if this test passes, it should be equivalent to running a full test board, and you can be certain that the proofs will work. 

When we start pulling things from the C config, this test is supposed to run more sessions -- everything that can be affected by a seL4 PR. Currently that is only CRefine (from CKernel up) and SimplExportAndRefine, but in the future these can be more (or even all of them). We only need to tweak the `session` parameter in this repo to make that happen.

lsf37:
> How "expensive" is this? Can one use this freely?

Relatively liberally, yes. We have enough credits calculated to run about 5 of these every single calendar day (on top of all the other proof tests), and we don't usually have that much PR traffic that needs this test.

It is possible that we'll be hitting VM number limits, because this uses the same VMs as the standard proof regression tests, and it needs 4 of them. If you see it fail with an error message about limits, it's Ok to just try again (re-run jobs in the Actions panel). If it happens too often I can also increase the limit (takes about a day, you have to file a request with AWS).

lsf37:
I'll try this out for a few days and then announce the label thing more widely if everything works as expected (it did in the tests, but there's always something..)

lsf37:
And turns out, there is something: the SimplExportAndRefine session is switched off in ARM_HYP and X64, because we don't have binary verification for them (my test only ran ARM, because why would a different platform make any difference..).

See #452 for an example.

Two ways we can deal with this: either we don't run SimplExportAndRefine for these, or we switch them on globally in the proofs. Since these were probably never even tried, they probably don't currently work (likely fixably). But also we don't really need them, so should we really burn cycles for running them all the time?

I think I'll put in a conditional here for now. Any thoughts @mbrcknl @Xaphiosis?

Xaphiosis:
I haven't been able to think of any way to bypass this without modifying run_tests.py to have some kind of option to "softly" request a test, i.e. add it if not already excluded (which raises a question what to do if any of its dependencies are excluded).

Currently as you noticed:
`L4V_ARCH=ARM_HYP ./run_tests -L SimplExportAndRefine CRefine`
results in:
`run_tests.py: error: Cannot run these tests because they depend on removed tests: SimplExportAndRefine`

I guess being able to run SimplExportAndRefine and keep it running on all arches is not too bad unless there's a lot of work to make that happen. How long does it run for?

lsf37:
> I guess being able to run SimplExportAndRefine and keep it running on all arches is not too bad unless there's a lot of work to make that happen. How long does it run for?

It's significant CPU hours (> 5 at some point), but parallelises well, so it reduces to < 1h. I think someone improved it a while ago, so it might just by 2-3 CPU hours now, but it's still not fast.

lsf37:
Ok, for posterity:

ARM: `Finished SimplExportAndRefine      passed      ( 0:30:00 real,  2:44:32 cpu,  8.26GB)`
RISCV64: `Finished SimplExportAndRefine      passed      ( 0:22:51 real,  2:15:32 cpu,  7.18GB)`

PR #458, opened 2021-07-15
README: add badge for simulation test

lsf37:
The badge name shows as `seL4Test`, which would imply also the hardware tests. We do intend to include them in that job group in the future, so I'd say it's fine to leave as is for a few weeks until that is done.


lsf37:
With #460 this bade now automatically includes the hardware builds (but not yet hardware test runs)

axel-h (approved):

PR #461, opened 2021-07-23
risc-v: remove unused L2 cache functions

axel-h:
The L2 cache handling functions were copied from the ARM code in the initial port,but they are not used on RISC-V. Remove them from the code base, they can be brought back if a platform has an L2 cache that needs to be maintained.

Is there any seL4 RISC-V code lurking in private repos that uses this, e.g. for HiFive/Polarfire?

### DIFF
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -223,8 +223,6 @@ void map_kernel_devices(void);
 
 /** MODIFIES: [*] */
 void initTimer(void);
-/* L2 cache control */
-void initL2Cache(void);
 void initLocalIRQController(void);
 void initIRQController(void);
 void setIRQTrigger(irq_t irq, bool_t trigger);
@@ -241,11 +239,6 @@ static inline void arch_pause(void)
 }
 
 #endif
-void plat_cleanL2Range(paddr_t start, paddr_t end);
-
-void plat_invalidateL2Range(paddr_t start, paddr_t end);
-
-void plat_cleanInvalidateL2Range(paddr_t start, paddr_t end);
 
 /* Update the value of the actual register to hold the expected value */
 static inline void Arch_setTLSRegister(word_t tls_base)
--- a/src/arch/riscv/machine/hardware.c
+++ b/src/arch/riscv/machine/hardware.c
@@ -253,21 +253,6 @@ BOOT_CODE void initTimer(void)
 }
 #endif /* !CONFIG_KERNEL_MCS */
 
-void plat_cleanL2Range(paddr_t start, paddr_t end)
-{
-}
-void plat_invalidateL2Range(paddr_t start, paddr_t end)
-{
-}
-
-void plat_cleanInvalidateL2Range(paddr_t start, paddr_t end)
-{
-}
-
-BOOT_CODE void initL2Cache(void)
-{
-}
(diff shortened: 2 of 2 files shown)


lsf37:
Running the proofs just in case there was some copy/past in those as well.

lsf37:
It looks like I might have screwed up the build trigger in my latest optimisation even though I tested them a bunch of times. Checking..

kent-mcleod (approved):

kent-mcleod:
> they can be brought back if a platform has an L2 cache that needs to be maintained.

Isn't all cache maintenance supposed to be handled via the FENCE instructions on RISC-V? Currently the kernel provides no invocations for performing cache maintenance on RISC-V.  So having a platform that needs specific L2 cache maintenance ops would become a large change requiring new invocations.

yyshen:
Some interesting read here
[https://forums.sifive.com/t/flush-invalidate-l1-l2-on-the-u54-mc/4483]

This reply is quite interesting.
"
Hi Antonin,

The only cache operations supported on the PolarFire SoC and FU540 SoC (on HiFive Unleashed) are the L2 Cache Flush operations (through the Flush32/Flush64 registers) and FENCE.I. Flushing a line in the L2 will also back probe into the L1 caches and flush them if required.

The CFLUSH.D.L1 and CDISCARD.D.L1 custom instructions were added after the designs for those SoCs were completed. CFLUSH.I.L1 isn’t supported in a released product yet, but it accidentally appeared in a few manuals as you found.
"

axel-h:
Thanks for the input. The goal of this is PR is just to remove things from the code base that are not used or get people to shout if this is used somewhere and then push these. Adding L2 cache maintenance support for RISC-V can happen at a later state if there is a platform that needs this and somebody is actively working on this, but then it can happen free of any ARM legacy.

yyshen:
> Thanks for the input. The goal of this is PR is just to remove things from the code base that are not used or get people to shout if this is used somewhere and then push these. Adding L2 cache maintenance support for RISC-V can happen at a later state if there is a platform that needs this and somebody is actively working on this, but then it can happen free of any ARM legacy.

It looks like the hifive unleashed board does have L2 cache flush supports, and that board is one of our supported platform.

axel-h:
@yyshen It is a supported platform, but from what you can see in the PR, no L2 handling is implemented so far and is seem none is needed. The functions are empty and they are not called from anywhere. So this is about deleting dead code avoid building legacy. RISC-V SMP support that may require L2 cache maintenance is also unclear, the comments in the code indicate it is still highly experimental. Soon I may have access to a HiFive Unmatched board and then hopefully find time to play with it...

yyshen:
> @yyshen It is a supported platform, but from what you can see in the PR, no L2 handling is implemented so far and is seem none is needed. The functions are empty and they are not called from anywhere. So this is about deleting dead code avoid building legacy. RISC-V SMP support that may require L2 cache maintenance is also unclear, the comments in the code indicate it is still highly experimental. Soon I may have access to a HiFive Unmatched board and then hopefully find time to play with it...
> 
> 

polarfire uses the same u54 core as hifive unleaded and has L2 cache. it is also a supported platform? Thus, I think it is time to fill in these functions.

lsf37 (approved):

PR #462, opened 2021-07-24
cleanup: Enforce ull for TIMER_FREQUENCY

axel-h:
- use ULL_CONST() for timer value. Enforce the maximum possible integer size in the generated C headers instead of doing this in CMake.
- merge TIMER_FREQUENCY definitions. There is no need for separate definitions, they define the same value.


axel-h:
Not sure how this failure in `PC99_release_gcc_32` could be related to my changes that only affect ARM andRISC-V platforms.
```
<testcase classname="sel4test" name="BREAKPOINT_002">
    <failure type="failure">
        result == SUCCESS 
        at line 291 of file /github/workspace/projects/sel4test/apps/sel4test-driver/src/testtypes.c
    </failure>
    <error>
        result == SUCCESS 
        at line 217 of file /github/workspace/projects/sel4test/apps/sel4test-driver/src/main.c
    </error>
</testcase>
```

lsf37:
I'm not sure either. Rerunning the test -- if it disappears in the rerun, this test should go on the list for close investigation for race conditions or other problems.

lsf37 (approved):

PR #466, opened 2021-07-25
prefer macro CURRENT_CPU_INDEX()

axel-h:
- Replace the macro `SMP_TERNARY(getCurrentCPUIndex(), 0)` by the much simpler macro `CURRENT_CPU_INDEX()` that does the same.
- define macro `SEL4_WORD_CONST()`. `CURRENT_CPU_INDEX()` is supported to return a `word_t`. The C parser from the verification toolchain requires declaring `word_t` constants without casting integer values to `word_t`.
- The field `ksKernelEntry.core` always exists, ensure it is properly set on non-SMP configurations also.

### DIFF
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -125,14 +125,14 @@ extern struct debug_syscall_error current_debug_error;
  * kernel is not performing their requested operation.
  */
 #define userError(M, ...) \
-    do {                                                                     \
-        out_error(ANSI_DARK "<<" ANSI_GREEN "seL4(CPU %lu)" ANSI_DARK        \
-                " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n",         \
-                SMP_TERNARY(getCurrentCPUIndex(), 0lu),                      \
-                __func__, __LINE__, NODE_STATE(ksCurThread),                 \
-                THREAD_NAME,                                                 \
-                (word_t)getRestartPC(NODE_STATE(ksCurThread)),               \
-                ## __VA_ARGS__);                                             \
+    do {                                                                       \
+        out_error(ANSI_DARK "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
+                ANSI_DARK " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n", \
+                CURRENT_CPU_INDEX(),                                           \
+                __func__, __LINE__, NODE_STATE(ksCurThread),                   \
+                THREAD_NAME,                                                   \
+                (word_t)getRestartPC(NODE_STATE(ksCurThread)),                 \
+                ## __VA_ARGS__);                                               \
     } while (0)
 #else /* !CONFIG_PRINTING */
 #define userError(...)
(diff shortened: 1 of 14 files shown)


lsf37:
This looks fine from the verification side, have started a proof test just to make sure.

lsf37 (approved):
This is fine from the verification side, and I think it makes sense for the code as well. It'd be good to get a second opinion from someone who has worked more with the SMP code, though.

@yyshen would you be happy to have a look?

axel-h:
Adding a comment about the C preprocessor macro evaluation rules, this should clarify the mysterious indirection used there.

axel-h:
The `SCHED0000`failure on nehalem/clang looks unrelated, In the test it says "This test non-deterministically fails. If you find only this test try re-running the test suite."
```
<testcase classname="sel4test" name="SCHED0000">
  		<failure type="failure">result == SUCCESS at line 291 of file /github/workspace/projects/sel4test/apps/sel4test-driver/src/testtypes.c</failure>
  	</testcase>
 ```

There is https://github.com/seL4/seL4_libs/pull/39 now to get rid of one of the many clang build warnings.

File: include/model/statedata.h [hunk 1]
```diff
@@ -48,7 +48,7 @@
 #define ARCH_NODE_STATE_ON_CORE(_state, _core) _state
 #define NODE_STATE_ON_CORE(_state, _core)      _state
 
-#define CURRENT_CPU_INDEX() 0
```
yyshen, line 51: any issues when a plain 0 was used  before?

lsf37:
> The `SCHED0000`failure on nehalem/clang looks unrelated, In the test it says "This test non-deterministically fails. If you find only this test try re-running the test suite."

Yes, that was unrelated, and simulation now succeeds.

lsf37:
The hw run failed for `MULTICORE0004` on sabre, for which I've put up seL4/sel4test#44 (is happening way too frequently). Will re-run them.

include/model/statedata.h
axel-h: Yes. Note that when the error message is printed there is a `SMP_TERNARY(getCurrentCPUIndex(), 0lu)` there actually, using `0lu` instead of `0`.

yyshen (approved):

include/api/types.h
```diff
@@ -125,14 +125,14 @@ extern struct debug_syscall_error current_debug_error;
...
-                " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n",         \
-                SMP_TERNARY(getCurrentCPUIndex(), 0lu),                      \
-                __func__, __LINE__, NODE_STATE(ksCurThread),                 \
-                THREAD_NAME,                                                 \
-                (word_t)getRestartPC(NODE_STATE(ksCurThread)),               \
-                ## __VA_ARGS__);                                             \
+    do {                                                                       \
+        out_error(ANSI_DARK "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
+                ANSI_DARK " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n", \
+                CURRENT_CPU_INDEX(),                                           \
```
yyshen, line 131: nitpick, are these light green squares spaces?

kent-mcleod: They're just showing the diff from the removal of the SMP_TERNARY line.  Because there's a trailing \ at the end of the line, there needs to be new spaces added to make up for the removed non-space characters.

PR #467, opened 2021-07-25
mcs: Don't check if old SC is still configured

xurtis:
When we are changing to a new SC, it doesn't matter whether the old SC is still configured. We should also know statically within `switchSchedContext` that whichever SC we will use to execute with next is both present and configured.

As such, this check can be removed.


mktnk3 (approved):

lsf37:
@mktnk3 do we have invariant proofs for this already?

mktnk3:
> @mktnk3 do we have invariant proofs for this already?

This is a change that came up from "add `refill_unblock_check`" change, which is in progress. But now you mention, I could and probably should do this as a separate change.

lsf37:
Ok, I'll wait with this one until we have the proof that it is safe to remove.

lsf37:
Proofs have passed

pingerino:
I vaguely remember this being quite important. 

If the old SC is configured, we want to make sure we maintain the bandwidth as we change SCs - we shouldn't violate bandwidth constraints during the change over. 

If the old SC is not configured, we don't care, as it's new. 

How is this resolved now?

kent-mcleod:
> I vaguely remember this being quite important. Something to do with revoke and avoiding switching to an SC that has been revoked. Perhaps that has been fixed in other ways.

I think that is handled by the same guard that is present inside `commitTime()`.

lsf37:
Yes, I think that is where it's from (but I haven't looked at it in detail). The proof does show that there is no change in behaviour from this, i.e. the condition is always true at this point.

kent-mcleod:
> Yes, I think that is where it's from (but I haven't looked at it in detail). The proof does show that there is no change in behaviour from this, i.e. the condition is always true at this point.

Is the condition always true? A variant on the example that @pingerino refers to is that the current thread is doing a revoke operation and revokes it's current SC.  I thought that the design allowed this by setting the memory such that scRefillMax is 0 and relies on the memory not being reallocated until after the kernel returns to userlevel (https://github.com/seL4/seL4/blob/master/src/object/objecttype.c#L232).  Hence the `NODE_STATE(ksCurSC)->scRefillMax` checks in `switchSchedContext()` and `commitTime()`.

Has this changed? Regardless, it still seems like removing the guard in this case is necessary as the `refill_unblock_check()` for the new SC needs to be called regardless of the value of ksCurSC (assuming that they are being switched).

lsf37:
Hm, you are totally right, I didn't look deeply enough. The proof needed no change at all, which is usually because the condition is already floating around the proof state. There is another possible explanation, though, which is that changing the `if` does not generate any difference in the rest of the conditions, even though the behaviour has changed. So everything is still as safe in terms of invariants as before, there was nothing that the condition contributed to that, but the behaviour still can have changed.

So I merged this a bit too eagerly, sorry.

You are right that `commitTime` has a `scRefillMax` guard as well, so the behaviour is still *almost* the same as before. The difference is that `commitTime` sets `NODE_STATE(ksConsumed) = 0llu` unconditionally. All the other action in `switchSchedContext` is on `NODE_STATE(ksCurThread)->tcbSchedContext`, not on `ksCurSC`.

PR #468, opened 2021-07-26
RISC-V interrupt handling and PLIC cleanup

axel-h:
Improving the RISC-V code dealing with interrupt and PLIC handling.

### DIFF
--- a/include/drivers/irq/riscv_plic0.h
+++ b/include/drivers/irq/riscv_plic0.h
@@ -3,9 +3,17 @@
  * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
+ *
+ * SiFive U54/U74 PLIC handling (HiFive Unleashed/Unmatched, Polarfire)
  */
 
 #pragma once
+
+/* This is a check that prevents using this driver blindly. Extend the list if
+ * this driver is confirmed to be working on other platforms. */
+#if !defined(CONFIG_PLAT_HIFIVE) && !defined(CONFIG_PLAT_POLARFIRE)
+#error "This code supports the SiFive U54/U74 PLIC only."
+#endif
 
 /* tell the kernel we have the set trigger feature */
 #define HAVE_SET_TRIGGER 1
--- a/include/drivers/irq/riscv_plic0.h
+++ b/include/drivers/irq/riscv_plic0.h
@@ -106,26 +106,28 @@ static inline bool_t plic_pending_interrupt(word_t interrupt)
     }
 }
 
-static inline word_t get_hart_id(void)
+/* The PLIC has separate register sets for each hart and the hart's context.
+ * This returns the hart ID used by the PLIC for the hart this code is currently
+ * executing on.
+ */
+static inline word_t plic_get_current_hart_id(void)
 {
-#ifdef ENABLE_SMP_SUPPORT
-    return cpuIndexToID(getCurrentCPUIndex());
-#else
-    return CONFIG_FIRST_HART_ID;
-#endif
+    return SMP_TERNARY(
+               cpuIndexToID(getCurrentCPUIndex()),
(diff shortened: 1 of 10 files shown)


axel-h:
The Preprocess failures are expected due to the changes.

lsf37:
This looks overall Ok to me, and might not even have (much) verification impact. I'll try a manual hardware test.

src/arch/riscv/machine/hardware.c
```diff
@@ -62,67 +62,70 @@ BOOT_CODE void map_kernel_devices(void)
...
- * @return     The active irq.
+ * @return     The active irq or irqInvalid.
  */
 static inline irq_t getActiveIRQ(void)
 {
+    irq_t *active_irq_slot = &active_irq[CURRENT_CPU_INDEX()];
 
-    uint32_t irq;
-    if (!IS_IRQ_VALID(active_irq[CURRENT_CPU_INDEX()])) {
-        active_irq[CURRENT_CPU_INDEX()] = getNewActiveIRQ();
```
kent-mcleod, line 126: It was originally done this way in order to be fairly identical to the Arm verified platform: https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/machine/gic_v2.h#L162 to make verification easier and for implementation consistency.

include/drivers/irq/riscv_plic0.h
```diff
@@ -3,10 +3,18 @@
...
+ *
+ * SiFive U54/U74 PLIC handling (HiFive Unleashed/Unmatched, Polarfire)
  */
 
 #pragma once
 
+/* This is a check that prevents using this driver blindly. Extend the list if
+ * this driver is confirmed to be working on other platforms. */
+#if !defined(CONFIG_PLAT_HIFIVE) && !defined(CONFIG_PLAT_POLARFIRE)
+#error "This code supports the SiFive U54/U74 PLIC only."
```
kent-mcleod, line 15: This change seems premature to me based on the following assumptions:
- RISC-V is trying to have a standardized PLIC that is used across many implementations
- This driver is written for a device with a compatibility string `riscv,plic0` in the device tree, which doesn't indicate that it's SiFive specific as compatibility strings are supposed to be in a global namespace.
- We don't have any platforms in the kernel that need a different PLIC driver yet (ignoring the platforms that don't use a PLIC) and as far as we know, the next platforms will also need to use this driver.

axel-h: I've put this in as a marker. It is expected to be lifted again, but this should be a deliberate action then to track this better. You have a point that it is premature, but since RISC-V is a quite open platform where any chip could be special, I'd prefer to be restrictive here for a while and see what happens.

src/arch/riscv/machine/hardware.c
axel-h: I can rework it back and postpone this change if that is a blocker.

kent-mcleod: Up to @lsf37 as I only have a weak opinion that the implementation should follow a similar pattern across architectures if possible, but it doesn't have to.

include/drivers/irq/riscv_plic0.h
kent-mcleod: But adding this header file to other platforms already needs to be a deliberate action in declare_default_headers() so this seems overly conservative to me.

axel-h: True, but we had some discussions internally about the PLIC spec, what our chip does, what other implementations seem to be doing and where the spec is evolving. There is a certain fear of fragmentation here and due to the lack of actual silicon form other vendors I tend to be very cautious here. I'm not too happy with `riscv,plic0` also, even if SiFive practically sets RISC-V standards.

kent-mcleod:
I manually checked that this passes sel4test on hifive due to the CI having it disabled currently.

PR #469, opened 2021-07-26
github: use correct label field

lsf37:
The trigger matched against the whole label object, not the name as it should have.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -16,7 +16,7 @@ jobs:
     # don't run again when other labels are added
     if: ${{ github.event.action != 'labeled' &&
               contains(github.event.pull_request.labels.*.name, 'proof-test') ||
-            github.event.action == 'labeled' && github.event.label == 'proof-test' }}
+            github.event.action == 'labeled' && github.event.label.name == 'proof-test' }}
     strategy:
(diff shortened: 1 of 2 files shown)


axel-h (approved):

lsf37:
Hm this is not great: the simulation builds show up as "skipped" in the checks list at the bottom of the PR, because they are skipped on the "labeled" event, but they were running on the commit itself from the original "open PR" event. 

In fact, you will see that the commit has a red cross for failed simulation[*], but the PR checks don't flag that. 

We might have to let the simulations run unconditionally, i.e. a new run for each label event. Not really great to heat the planet with that. Will try to think of something.

[*]: somehow it seems to have picked up the same broken docker build that Bamboo has. Possibly because Bamboo pushed something. Not sure where that is coming from, but will have a look at that as well..

lsf37:
Alright, this seems to be working as intended now, and the docker mystery is also cleared up: The failure is indeed the same as on Bamboo, and it is because Bamboo has *not* pushed out Damon's fix for the problem and the GitHub test is now testing the same manifest as Bamboo (master).

It'll take a few hours to build all the docker images, but then things should be working again.

lsf37:
And the containers are updated now as well. Will merge..

PR #473, opened 2021-07-31
cleanup explicit function Arch_finaliseInterrupt()

axel-h:
PC99 is the only platform that uses `Arch_finaliseInterrupt()`. There seems no need to have this explicit function, the necessary things can be done in `ackInterrupt()` also, which is architecture or platform specific anyway.

### DIFF
--- a/include/arch/x86/arch/machine.h
+++ b/include/arch/x86/arch/machine.h
@@ -361,7 +361,7 @@ static inline unsigned long getFaultAddr(void)
 
 static inline void Arch_finaliseInterrupt(void)
 {
-    ARCH_NODE_STATE(x86KScurInterrupt) = int_invalid;
+    /* Nothing architecture specific to be done. */
 }
 
 static inline void x86_set_tls_segment_base(word_t tls_base);
--- a/include/arch/x86/arch/machine.h
+++ b/include/arch/x86/arch/machine.h
@@ -359,11 +359,6 @@ static inline unsigned long getFaultAddr(void)
     return read_cr2();
 }
 
-static inline void Arch_finaliseInterrupt(void)
-{
-    /* Nothing architecture specific to be done. */
-}
-
 static inline void x86_set_tls_segment_base(word_t tls_base);
 
 /* Update the value of the actual register to hold the expected value */
--- a/include/arch/arm/arch/machine.h
+++ b/include/arch/arm/arch/machine.h
@@ -65,11 +65,6 @@ static inline void arch_pause(void)
 }
 #endif /* ENABLE_SMP_SUPPORT */
 
-static inline void Arch_finaliseInterrupt(void)
-{
-    /* Nothing architecture specific to be done. */
-}
-
 /* Update the value of the actual regsiter to hold the expected value */
 static inline exception_t Arch_setTLSRegister(word_t tls_base)
 {
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -254,11 +254,6 @@ static inline void setVSpaceRoot(paddr_t addr, asid_t asid)
 #endif
 }
 
-static inline void Arch_finaliseInterrupt(void)
-{
-    /* Nothing architecture specific to be done. */
-}
-
 void map_kernel_devices(void);
 
 /** MODIFIES: [*] */
--- a/include/plat/pc99/plat/machine/interrupt.h
+++ b/include/plat/pc99/plat/machine/interrupt.h
@@ -96,6 +96,8 @@ static inline void ackInterrupt(irq_t irq)
     } else {
         apic_ack_active_interrupt();
     }
+
+    ARCH_NODE_STATE(x86KScurInterrupt) = int_invalid;
 }
 
 static inline void handleSpuriousIRQ(void)
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -46,7 +46,6 @@ exception_t handleInterruptEntry(void)
 
     if (IRQT_TO_IRQ(irq) != IRQT_TO_IRQ(irqInvalid)) {
         handleInterrupt(irq);
-        Arch_finaliseInterrupt();
     } else {
 #ifdef CONFIG_IRQ_REPORTING
         userError("Spurious interrupt!");
@@ -628,7 +627,6 @@ exception_t handleSyscall(syscall_t syscall)
                 mcsPreemptionPoint(irq);
                 if (IRQT_TO_IRQ(irq) != IRQT_TO_IRQ(irqInvalid)) {
                     handleInterrupt(irq);
-                    Arch_finaliseInterrupt();
                 }
             }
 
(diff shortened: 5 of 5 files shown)


lsf37 (commented):
I don't think we should do this. We can't assume that the current architectures will remain the only architectures, so the argument that an arch interface is used in only one architecture isn't a real reason for removing such an interface.

Architecture interfaces also stretch into the proofs ([which do break for this one](https://github.com/seL4/seL4/runs/3206842850)), and specifically finalising things in delete has a lot of invariants and conditions that we need good interface points for where we can state these.

The functions are all inline, so there is no performance impact from having them.

(not sure why the Checks summary says the proofs are skipped -- they do show when you click on the commit)

axel-h:
I have been looking though the code to understand what purpose `Arch_finaliseInterrupt()` really serves currently, but I could not find any. We delegate the interrupt handling to an interrupt controller "driver" everywhere in the generic code. There seem is a clear rule that `ackInterrupt()` must be called when the kernel is done with the interrupt and this is supposed to be called in the current kernel entry path. `handleInterrupt()` from the generic code does exactly this everywhere. Thus, whatever is done in `Arch_finaliseInterrupt()` can also be done in `ackInterrupt()`.

lsf37:
> I have been looking though the code to understand what purpose `Arch_finaliseInterrupt()` really serves currently, but I could not find any. We delegate the interrupt handling to an interrupt controller "driver" everywhere in the generic code. There seem is a clear rule that `ackInterrupt()` must be called when the kernel is done with the interrupt and this is supposed to be called in the current kernel entry path. `handleInterrupt()` from the generic code does exactly this everywhere. Thus, whatever is done in `Arch_finaliseInterrupt()` can also be done in `ackInterrupt()`.

I'm not disputing that the refactor is correct or saying that the `Arch_finaliseInterrupt` is necessary in the code, I'm trying to express that the `Arch_finaliseInterrupt` interface is useful in the proofs for stating intermediate properties. This can in theory also be refactored, but it is a few day's worth of work which I don't think would improve the proof structure nor be worth the effort for the slightly simpler code structure.

Does that make sense?

axel-h:
Yes, I see you point. Actually, I wanted to convert this PR into a draft-PR because I'm started reworking. What I will do is keeping `Arch_finaliseInterrupt()` function and add a comment why it exists, why it is empty any why it should remain empty unless there there is a good reason to use this hook. For the PC99 I will keep the refactoring, because there seem no need why it need to use the `Arch_finaliseInterrupt()` hook - unless somebody disagrees, but then I'd like to add a comment in the code explaining this, because it seems not trivial then.

axel-h:
Is there anything more specific I can put in to code comment about what `Arch_finaliseInterrupt()` is really for. Calling it when running though an interrupt trap seems quite straight forward. But calling it in a syscall after `mcsPreemptionPoint()` when polling the interrupts manually (i.e. everytme `handleInterrupt()` is called) seem less intuitive. I wonder if this is really intended to happen in the end or this just happen to end up there due to the way things were done on the pc99 platform and what this PR cleans up now.

lsf37:
> Is there anything more specific I can put in to code comment about what `Arch_finaliseInterrupt()` is really for. Calling it when running though an interrupt trap seems quite straight forward. But calling it in a syscall after `mcsPreemptionPoint()` when polling the interrupts manually (i.e. everytme `handleInterrupt()` is called) seem less intuitive. I wonder if this is really intended to happen in the end or this just happen to end up there due to the way things were done on the pc99 platform and what this PR cleans up now.

`Arch_finaliseInterrupt()` used to be called `Arch_handleInterrupt` (replaced in 50995b4c5f5bfd88c0ae049966f58f5cdcd6d8df), and was renamed to indicate that it should happen after the generic interrupt part. The initial intention seems to have been (I don't remember that bit) a place holder for anything arch specific that might need to be done. It happens to be empty for most architectures, and it happened to be a good cut point for the proofs.

src/object/interrupt.c
```diff
@@ -182,74 +182,81 @@ void deletedIRQHandler(irq_t irq)
...
+         * driver reported this interrupt. Maybe the value maxIRQ is just wrong
+         * it set to a lower value because the interrupts are unused.
+         */
+        printf("Received IRQ %d, which is above the platforms maxIRQ of %d\n",
+               (int)IRQT_TO_IRQ(irq), (int)maxIRQ);
         maskInterrupt(true, irq);
-        ackInterrupt(irq);
-        return;
-    }
-    switch (intStateIRQTable[IRQT_TO_IDX(irq)]) {
```
lsf37, line 194: The GitHub diff is hard to read at this point. It looks like this only factors out `ackInterrupt` and moves it down, but I don't think we should change the structure here. This looks like a small change, but the early return makes the proof nicer in this case and we'd lose that with the big `else`. I'd rather have the small duplication in the code than touch those proofs (they are pretty hairy).

If we remove `Arch_finaliseInterrupt`, which now does seem possible, this function could actually remain unchanged.

lsf37:
Alright, after looking at the proofs more, I'm coming around to the view that `Arch_finaliseInterrupt` can actually be removed, because I can remove it in the proofs too, which is actually a significant plus. It'll cost 1-2 full days of work, but getting rid of the entire interface point would be worth it. I didn't realise initially that we had it on X64 to then just show it does nothing interesting.

If you're happy with leaving `handleInterrupt` unchanged (comment updates fine, of course), I'd be happy to look at the verification side.

We still should double-check that we have caught all contexts of `ackInterrupt` (you might have already done that, just making sure).

src/object/interrupt.c
axel-h: yes. I have move this to the end of the function to avoid code duplication because `ackInterrupt()` is done on any case and `Arch_finaliseInterrupt()` also.

include/arch/x86/arch/machine.h
```diff
@@ -361,7 +361,7 @@ static inline unsigned long getFaultAddr(void)
 
 static inline void Arch_finaliseInterrupt(void)
 {
-    ARCH_NODE_STATE(x86KScurInterrupt) = int_invalid;
+    /* Nothing architecture specific to be done. */
```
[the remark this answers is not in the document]
axel-h: I have look through the code paths several times while working on this PR, so I have a certain confidence doing this in `ackInterrupt()` is sufficient and aligns with the design idea behind this (that it must be called in the same kernel entry call flow when the interrupt subsystem has reported an interrupt). I will do another round of check when I update this PR.

axel-h:
@lsf37: what do you thing about these commits now. And thengeneralizing `mcsPreemptionPoint()` to `syscall_preemption_point()` that wraps the interrupt checking?

lsf37:
Sorry, it took me a while to get back to this. This is too much code movement for the MCS proofs, which are still under active development and need to be kept in sync manually. Ideally we should touch this only if really necessary or requested from the proof side.

I was actually happy with just removing `Arch_finaliseInterrupt` in the end (factoring out just the x86 bit) -- it simplifies both the code and the proofs, even if it costs a bit of work on the verification side. I don't think we should try to do more in this iteration.

axel-h:
Removed the commits that creates `create syscall_preemption_point()` from this PR and just put the commit there that removes `Arch_finaliseInterrupt()`. Is this fine now?

lsf37:
Yes, thank you for your patience :-). Now I only need to do the proof bit, but otherwise it is good to merge.

lsf37:
Proof update for this is now done.

PR #476, opened 2021-08-02
mcs: Explicitly check that sporadic scs are active

xurtis:
Easier to check this explicitly than prove the invaraint.

### DIFF
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -159,7 +159,10 @@ static inline bool_t sc_released(sched_context_t *sc)
  */
 static inline bool_t sc_sporadic(sched_context_t *sc)
 {
-    return sc != NULL && sc->scSporadic;
+    /* We do expect sc_active to be always true if sc is not NULL when
+       sc_sporadic is called, but we have not proved it yet. */
+    assert(sc == NULL || sc_active(sc));
+    return sc != NULL && sc_active(sc) && sc->scSporadic;
 }
 
 /*


include/kernel/sporadic.h
```diff
@@ -159,7 +159,7 @@ static inline bool_t sc_released(sched_context_t *sc)
  */
 static inline bool_t sc_sporadic(sched_context_t *sc)
 {
-    return sc != NULL && sc->scSporadic;
+    return sc != NULL && sc_active(sc) && sc->scSporadic;
```
mktnk3, line 162: This surely works. But this would duplicate the `sc_active` test when the guard actually checks both sporadic and active, just to mention. I guess it doesn’t hurt.

mktnk3 (approved):

lsf37:
Hardware test succeeded (modulo one boot failure for one pc99 run that succeed in the other pc99 compiler combinations).

I'd be happy to merge this. @mktnk3 is the spec already updated for this change?

lsf37 (approved):

mktnk3:
> I'd be happy to merge this. @mktnk3 is the spec already updated for this change?

Spec update is in progress as part of “add refill_unblock_check” change. It should be fine to merge this now. The “adding refill_unblock_check" part of the change is already merged.

axel-h:
I'd like to understand something about the proofs: do they exist to verify things like this, because this is in the design and thus there should be no need for an explicit check? Assuming there is a certain faith that `sc_active(sc)` always holds here or even a design rule, that this must holds, putting an `assert()` here might safe doing an explict check in (proved) release builds, while it still catches things in debug/hacker builds:
```
if (!sc) {
  return false;
} 
assert(sc_active(sc)); /* the sc must be active */
return sc->scSporadic;
```

Which brings me to the question, are the profs supposed to show that `assert()` hold?

lsf37:
> I'd like to understand something about the proofs: do they exist to verify things like this, because this is in the design and thus there should be no need for an explicit check? 

The proofs previously tried to show that, but it so far not successful and is too much work in relation to what it achieves. Since this project is super late, we're cutting the proof cost here in favour of a runtime check. Later, when there is more time, we could add the condition to the invariant again, prove it, and then remove the runtime check, because we have shown it is unnecessary. 

> Assuming there is a certain faith that `sc_active(sc)` always holds here or even a design rule, that this must holds, putting an `assert()` here might safe doing an explict check in (proved) release builds, while it still catches things in debug/hacker builds:
> 
> ```
> if (!sc) {
>   return false;
> } 
> assert(sc_active(sc)); /* the sc must be active */
> return sc->scSporadic;
> ```

Yes, it is a not a bad idea to put an assert in debug builds for things we think must always be true, esp during development when there are no proofs yet, as long as the behaviour is be the same with and without assert (i.e. if the assert is correct, there should be no difference between the builds). Before the proof is finished, this can also give the verification team hints of what the author thought should be true or relied upon at that point.

> Which brings me to the question, are the profs supposed to show that `assert()` hold?

No, the verification does not see C-level asserts (since they are off for release builds). For specific asserts that we are interested in and the proofs don't force us to show already for other reasons anyway, we can manually add them to the Haskell/design level. Then the proofs are force to show them as well. Most things we tend to assert in the C do come up naturally in the proofs anyway, so it's usually not necessary.

lsf37:
(recreated old comment, unfortunately now showing up out of order)

> I'd like to understand something about the proofs: do they exist to verify things like this, because this is in the design and thus there should be no need for an explicit check?

The proofs previously tried to show that, but it so far not successful and is too much work in relation to what it achieves. Since this project is super late, we're cutting the proof cost here in favour of a runtime check. Later, when there is more time, we could add the condition to the invariant again, prove it, and then remove the runtime check, because we have shown it is unnecessary.

> Assuming there is a certain faith that sc_active(sc) always holds here or even a design rule, that this must holds, putting an assert() here might safe doing an explict check in (proved) release builds, while it still catches things in debug/hacker builds:
>
> ```
> if  (!sc) {
>   return false;
> } 
> assert(sc_active(sc)); /* the sc must be active */
> return sc->scSporadic;
> ```

Yes, it is a not a bad idea to put an assert in debug builds for things we think must always be true, esp during development when there are no proofs yet, as long as the behaviour is be the same with and without assert (i.e. if the assert is correct, there should be no difference between the builds). Before the proof is finished, this can also give the verification team hints of what the author thought should be true or relied upon at that point.

> Which brings me to the question, are the profs supposed to show that assert() hold?

No, the verification does not see C-level asserts (since they are off for release builds). For specific asserts that we are interested in and the proofs don't force us to show already for other reasons anyway, we can manually add them to the Haskell/design level. Then the proofs are force to show them as well. Most things we tend to assert in the C do come up naturally in the proofs anyway, so it's usually not necessary.

lsf37:
I've rebased this and added the assertion with a comment.

lsf37:
And it promptly turns out that this is not an invariant and the assertion is not always true (see #496)

PR #479, opened 2021-08-08
Hardware tests triggered from GitHub

lsf37:
This PR adds the trigger that replaces Bamboo hardware tests. It runs the same general sel4test setup with the UNSW machine queue, just triggered from GitHub.

The current set of available boards is fairly small, more boards will come online over time.

The test runs on push to the `master` branch or on PR when explicitly labeled with `hw-test`.

See also seL4/ci-actions#135

### DIFF
--- a/.github/workflows/sel4test-hw.yml
+++ b/.github/workflows/sel4test-hw.yml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-# sel4test hardware builds
+# sel4test hardware builds and runs
 #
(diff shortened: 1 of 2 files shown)


lsf37:
Maybe we should change that link after all, until the website is finalised.

lsf37:
odriod_c2 seems to fail to boot every one in 20-ish times, and the boot script is not recognising it correctly. That's independent of this PR, though, and will be fixed on the machine queue side.

Xaphiosis (approved):

PR #480, opened 2021-08-10
mcs: Remove domain time check from preemptionPoint

xurtis:
This removes the operations that trigger a reschedule or reprogram the timer from `preemptionPoint` to ensure they relevant state updates in the proof occur where it they are easier to verify.

### DIFF
--- a/include/kernel/thread.h
+++ b/include/kernel/thread.h
@@ -229,18 +229,23 @@ static inline void updateTimestamp(void)
     time_t consumed = (NODE_STATE(ksCurTime) - prev);
     NODE_STATE(ksConsumed) += consumed;
     if (CONFIG_NUM_DOMAINS > 1) {
-
         if ((consumed + MIN_BUDGET) >= ksDomainTime) {
             ksDomainTime = 0;
         } else {
             ksDomainTime -= consumed;
         }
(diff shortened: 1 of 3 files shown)


michaelmcinerney (approved):
Looks great. Thanks Curtis!

include/kernel/thread.h
```diff
@@ -229,18 +229,23 @@ static inline void updateTimestamp(void)
...
-        }
     }
+}
 
+/*
+ * Check if domain time has expired
+ */
+static inline void checkDomainTime(void)
+{
+    if (unlikely(CONFIG_NUM_DOMAINS > 1 && isCurDomainExpired())) {
```
michaelmcinerney, line 245: `CONFIG_NUM_DOMAINS > 1` is already included in `isCurDomainExpired`, so this part can be removed

michaelmcinerney:
Would we also like to change the guard in `scheduleChooseNewThread` to `isCurDomainExpired`?

michaelmcinerney:
I've shown with this spec that we do indeed have nonzero domain time on kernel exit. It was very straightfoward with these changes. There are still most of the proofs for the scheduler invariants to go, but I might come back to those later

lsf37:
Closing in favour of #505

Issue #481, opened 2021-08-12
aarch64 clearMemory does not clear to RAM


lsf37:
This bug is relevant for correctness and security.

The likelihood to trigger this in deployed systems in a way that actually leaks confidential information should be very low since re-using memory over information security domains is already excluded in the security theorems, but it is possible to set up a system such that it can be observed.

This is a caching bug, so explicitly not covered by the seL4 proofs.

## Details

(originally reported by @bennoleslie)

I believe that the implementation of clearMemory on aarch64 may be incorrect

<https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/machine.h#L50>

    static inline void clearMemory(word_t *ptr, word_t bits)
    {
        memzero(ptr, BIT(bits));
        cleanCacheRange_PoU((word_t)ptr, (word_t)ptr + BIT(bits) - 1,
                            addrFromPPtr(ptr));
    }

As can be seen from the implementation this cleans the cache to the point of unification.

However, it does not clear the cache to the point of coherency.

## Why does this matter?

clearMemory is called during the retype operation to ensure that any memory handed out to the user does not have any sensitive data in it.

If clearMemory does not actually clear the memory then a user may access sensitive data.

## Why does PoU vs PoC matter?

From the Armv8 architecture reference manual:

PoC is described as:

> The point at which all agents that can access memory are guaranteed to see the same copy of a memory location for accesses of any memory type or **cacheability attribute**.

(Emphasis added)

PoU is described as:

> The PoU for a PE is the point by which the instruction and data caches and the translation table walks of that PE are guaranteed to see the same copy of a memory location.

My reading of this is that cleaning to PoU means that any access via the cache subsystem will return the cleaned data, however, access to the memory via another agent or not via the cache subsystem has the potential to access stale (non-zero) data.

## Can the user skip the cache?

Well, as it is possible to map a page uncached, this can be achieved trivially. There may be other options via user control of additional bus masters in the system as well.

## Does this happen in practise?

Yes, I've observed on a real system that this happens in practise. User code was able to access a page containing non-zeroed data left over from the boot process.

## What's the fix?

[...]

PR #485, opened 2021-08-13
arm: flush cache to RAM on retype reset

lsf37:
This fixes a correctness and security issue where uncached user mappings might see old data from before the clearMemory operation.

See also the discussion on GitHub issue #481 and corresponding seL4/sel4test#45

### DIFF
--- a/include/arch/arm/arch/machine.h
+++ b/include/arch/arm/arch/machine.h
@@ -50,14 +50,15 @@ void cleanInvalidateL1Caches(void);
 static inline void clearMemory(word_t *ptr, word_t bits)
 {
     memzero(ptr, BIT(bits));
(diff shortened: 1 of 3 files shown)


kent-mcleod (approved):

lsf37:
I'll rebase to remove the merge commit and get the new cache test included in the hw test.

lsf37:
This is ready to merge once the proofs are updated.

lsf37:
Verification reminded us that `clearMemory` is not only called in retype. So there are two `clearMemory` versions now for Arm: the default flushes to RAM, and basically the old version that flushes to PoU, which is used for page table operations.

lsf37:
Free proof update in seL4/l4v#326, courtesy of [Proocraft](https://proofcraft.systems).

When that is approved, we can merge.

PR #486, opened 2021-08-14
github: enable hw test for imx8mm

lsf37:
One more board looking stable again.


lsf37:
(removed hw-test label here again, because the new board will only be included after this is merged, so there is no point)

Xaphiosis (approved):

PR #490, opened 2021-08-17
aarch64: fix storing of thread context

nspin:
TPIDR_EL0 and TPIDRRO_EL0 had been omitted from the storing of thread context upon kernel entry. This commit corrects that omission.

### DIFF
--- a/src/arch/arm/64/traps.S
+++ b/src/arch/arm/64/traps.S
@@ -91,8 +91,11 @@ END_FUNC(arm_vector_table)
     mrs     x21, sp_el0
     mrs     x22, ELR
     mrs     x23, SPSR
+    mrs     x24, TPIDR_EL0
+    mrs     x25, TPIDRRO_EL0
     stp     x30, x21, [sp, #PT_LR]
     stp     x22, x23, [sp, #PT_ELR_EL1]
+    stp     x24, x25, [sp, #PT_TPIDR_EL0]
 .endm
 
 BEGIN_FUNC(invalid_vector_entry)


kent-mcleod:
Hey @nspin 

Is there a test you have that show these registers not being saved/restored during context switching?

They should be being correctly managed in `arm_save_thread_id` and `arm_load_thread_id`: https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/64/mode/machine.h#L116
Called from `arch_c_entry_hook();` and `arch_c_exit_hook();` on entry and exit from the kernel.

nspin:
I was debugging a setup with multiple VMs spanning multiple overlapping cores. I was observing segfaults in userland which I discovered were due to cases of corrupted `TPIDR_EL0`. After this patch, I haven't observed such a segfault.

I will figure out what was going on more precisely in order to account for `arm_*_thread_id`.

nspin:
In the meantime, here is some evidence towards the existence of a bug:

If I apply `./magic.diff` on top of this PR (https://github.com/seL4/seL4/pull/490/commits/7c882363cee2f05f85d90d29cca796e2207e449a at the time of writing), the magic value `0xbadbadbadbaaaaad` eventually leaks into userspace, as witnessed by the crash shown in `./crash.txt`. According to `./gdb-session.txt`, this crash is due to a user thread accessing an address at a small offset from `TPIDR_EL0`.

`./magic.diff`:
```
diff --git a/src/arch/arm/64/traps.S b/src/arch/arm/64/traps.S
index 692c394b..4604a472 100644
--- a/src/arch/arm/64/traps.S
+++ b/src/arch/arm/64/traps.S
@@ -91,7 +91,7 @@ END_FUNC(arm_vector_table)
     mrs     x21, sp_el0
     mrs     x22, ELR
     mrs     x23, SPSR
-    mrs     x24, TPIDR_EL0
+    ldr     x24, =0xbadbadbadbaaaaad
     mrs     x25, TPIDRRO_EL0
     stp     x30, x21, [sp, #PT_LR]
     stp     x22, x23, [sp, #PT_ELR_EL1]
```

`./crash.txt`:
```
Caught cap fault in send phase at address 0
while trying to handle:
vm fault on data at address 0xbadbadbadbaaaabd with status 0x93c08004
in thread 0xff811fffb400 "realm_vmm_primary_tcb" at address 0x41c404
With stack:
0x100003feca0: 0x100003fecd0
0x100003feca8: 0x41c7e4
0x100003fecb0: 0x69
0x100003fecb8: 0xfffffffb
0x100003fecc0: 0x0
0x100003fecc8: 0x0
0x100003fecd0: 0x100003fed20
0x100003fecd8: 0x41d874
0x100003fece0: 0x100003ff3b0
0x100003fece8: 0x4
0x100003fecf0: 0xffffe000103600c8
0x100003fecf8: 0xffffe0001035fd20
0x100003fed00: 0xffffe0001035fcc2
0x100003fed08: 0x69
0x100003fed10: 0x1
0x100003fed18: 0x200d
```

`./gdb-session.txt`
```
Reading symbols from realm-vmm.elf...
(ins)(gdb) disas 0x41c404
Dump of assembler code for function seL4_GetIPCBuffer:
   0x000000000041c3f4 <+0>:     mrs     x1, tpidr_el0
   0x000000000041c3f8 <+4>:     adrp    x0, 0x435000
   0x000000000041c3fc <+8>:     ldr     x0, [x0, #3560]
   0x000000000041c400 <+12>:    add     x0, x1, x0
   0x000000000041c404 <+16>:    ldr     x0, [x0]
   0x000000000041c408 <+20>:    ret
End of assembler dump.
```

nspin:
I found a bug in my userspace that was causing a lot of cross-core TCB and VCPU operations. After fixing that bug and using the kernel _without_ the patch in this PR, I no longer observed the segfaults mentioned above.

However, I still believe that those segfaults were ultimately being caused by a bug in the kernel, albeit one that was only being triggered due to the extreme conditions of the cross-core TCB and VCPU operations caused by my userspace bug.

Interestingly, after fixing my userspace bug and using the kernel _with_ this PR, my system hangs after a few minutes of operation. I have not observed such a hang without the patch in this PR.

I'm still working on getting to the bottom of this.

For completeness, here are instructions to reproduce each of the three cases above:

- Userspace bug fixed, without the patch in this PR; test runs fine: https://gitlab.com/arm-research/security/icecap/icecap/-/blob/nspin/repro/tpidr-corruption/ok/REPRO.md
- Userspace bug present, without the patch in this PR; segfaults in userspace, with coredumps showing TPIDR_EL0 corruption: https://gitlab.com/arm-research/security/icecap/icecap/-/blob/nspin/repro/tpidr-corruption/corruption/REPRO.md
- Userspace bug present, with the patch in this PR; system hangs after a few minutes: https://gitlab.com/arm-research/security/icecap/icecap/-/blob/nspin/repro/tpidr-corruption/hang/REPRO.md

axel-h:
According to the comments, there is no immediate bug that this PR fixes. Would it make sense to put this PR into draft state then and remove the bug label?

lsf37:
It looks to me like there is still a potential bug here, but it needs more investigation. Draft would be appropriate, yes.

kent-mcleod:
Hi @nspin, is this something that you are still interested in pursuing? It looks like there's a good chance the issue is caused by incorrect synchronization around `c_entry_hook` and `c_exit_hook` as described in https://github.com/seL4/seL4/issues/857.  At a guess, it could be due to there being a execution path that calls c_exit_hook without calling c_entry_hook which would corrupt the contents of just those two registers (https://github.com/seL4/seL4/issues/857#issuecomment-1170125306).

kent-mcleod:
With that in mind, I'm more in favor of taking your proposed approach of treating them the same as other registers. (Which would also require adding them to `restore_user_context()`)

axel-h:
> With that in mind, I'm more in favor of taking your proposed approach of treating them the same as other registers. (Which would also require adding them to `restore_user_context()`)

I have done this change in https://github.com/Hensoldt-Cyber/seL4/tree/patch-axel-83 (https://github.com/Hensoldt-Cyber/seL4/pull/18). Maybe @nspin could give this a try and see if the problem also happens there?

lsf37:
Is there any reason why we would not want these two registers to be treated as part of the user context like the others?

It seems cleaner to treat them that way in any case, no matter if there still is a potential issue here or not and it also looks like it would be potentially a tiny bit faster.

Is the special case for them in `arch_c_entry_hook`/`arch_c_exit_hook` a left-over from the globals frame we had on ARMv6? For x64 and ia32 we seem to be doing a similar special case, but for RISCV-V we do not.

src/arch/arm/64/traps.S
```diff
@@ -91,8 +91,11 @@ END_FUNC(arm_vector_table)
     mrs     x21, sp_el0
     mrs     x22, ELR
     mrs     x23, SPSR
+    mrs     x24, TPIDR_EL0
+    mrs     x25, TPIDRRO_EL0
     stp     x30, x21, [sp, #PT_LR]
     stp     x22, x23, [sp, #PT_ELR_EL1]
+    stp     x24, x25, [sp, #PT_TPIDR_EL0]
```
Indanz, line 98: This won't be a 16-byte aligned write because `FaultIP` disrupts the flow, it will be faster on most machines if we move `FaultIP` lower in `enum _register`.

Indanz (commented):
The same change should be done in `restore_user_context`, and we should get rid of `arm_save_thread_id` and `arm_load_thread_id` too.

Indanz:
> Is there any reason why we would not want these two registers to be treated as part of the user context like the others?

Not that I can find.

> It seems cleaner to treat them that way in any case, no matter if there still is a potential issue here or not and it also looks like it would be potentially a tiny bit faster.

Agreed.

The same is true for x86 and 32-bit ARM though.

kent-mcleod:
https://github.com/seL4/seL4/pull/1414/files#diff-21a9686dc2b16e105dc49806d70208289d37adc2f0147823a1f497970213694a changed when the kernel lock is released so that it's after c_exit_hook().

kent-mcleod:
> The same change should be done in `restore_user_context`, and we should get rid of `arm_save_thread_id` and `arm_load_thread_id` too.

I agree.

PR #493, opened 2021-08-19
github: separate sel4test for deployment on master

lsf37:
This commit pulls out a separate workflow for sel4test (simulation + hardware runs) on pushes to master, and deploys a new default.xml to sel4test-manifest when the test is successful.

Also puts the machine queue run into a concurrency group, which means only one of these will run at the same time. This should reduce contention on the the machine queue and finish individual tests more quickly. If a test is already running, new tests will remained in "queued" state until the running one is done. If there is already a queued test, the previously queued test is cancelled and the latest one is queued. This means only the most recent commit will be deployed to the manifest if previous tests haven't started yet (but previous tests will finish and deployed if they have already started).

This applies only to the master branch. PRs hardware tests are separate and still run concurrently.

### DIFF
--- a/.github/workflows/sel4test-deploy.yml
+++ b/.github/workflows/sel4test-deploy.yml
@@ -2,36 +2,61 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-# sel4test hardware builds and runs
-#
-# See sel4test-hw/builds.yml in the repo seL4/ci-actions for configs.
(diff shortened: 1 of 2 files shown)


lsf37:
I've tested this fairly extensively in a smaller version in the draft PR, but there is no good way to really test the full deployment chain without merging this, so there might have to be a follow-up PR if I stuffed something up.

Xaphiosis (approved):

PR #494, opened 2021-08-19
use proper types for bootloader parameters

axel-h:
Use word_t and sword_t to be architecture agnostic.

Notes:
- This PR is on top of https://github.com/seL4/seL4/pull/425
- This PR needs https://github.com/seL4/seL4_tools/pull/100

(edit: removed test reference, because these two are now merged)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -38,6 +38,10 @@ Upcoming release: BINARY COMPATIBLE
   constant seL4_GlobalsFrame from libsel4.
 * Implement KernelArmExportPTMRUser and KernelArmExportVTMRUser options for Arm generic timer use access on aarch32.
 * Removed obsolete define `HAVE_AUTOCONF`
+* Changed kernel boot parameters to become architecture agnostic with using word_t instead of uint32. Stop using
+  sword_t due to C rule violations, use word_t instead and explicitly allow overflows and underflows in the
+  unsigned integer calculations. These changes could break the boot process on 64-bit platforms, as the upper 32 bit
+  must be set correctly now also. The elfloader has been updated, custom bootloaders might need to be updated also.
 
 ## Upgrade Notes
 ---
--- a/include/arch/arm/arch/kernel/boot.h
+++ b/include/arch/arm/arch/kernel/boot.h
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014, General Dynamics C4 Systems
+ * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -12,12 +13,15 @@ cap_t create_unmapped_it_frame_cap(pptr_t pptr, bool_t use_large);
 cap_t create_mapped_it_frame_cap(cap_t pd_cap, pptr_t pptr, vptr_t vptr, asid_t asid, bool_t use_large,
                                  bool_t executable);
 
+/* This is called from assembly code and thus there are no specific types in
+ * the signature.
+ */
 void init_kernel(
-    paddr_t ui_p_reg_start,
-    paddr_t ui_p_reg_end,
-    sword_t pv_offset,
-    vptr_t  v_entry,
-    paddr_t dtb_addr_p,
-    uint32_t dtb_size
+    word_t ui_p_reg_start,
+    word_t ui_p_reg_end,
+    word_t pv_offset,
+    word_t v_entry,
+    word_t dtb_addr_p,
+    word_t dtb_size
 );
 
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -310,7 +310,6 @@ BOOT_CODE static void release_secondary_cpus(void)
 #endif /* ENABLE_SMP_SUPPORT */
 
 /* Main kernel initialisation function. */
-
 static BOOT_CODE bool_t try_init_kernel(
     paddr_t ui_p_reg_start,
     paddr_t ui_p_reg_end,
@@ -336,7 +335,27 @@ static BOOT_CODE bool_t try_init_kernel(
     create_frames_of_region_ret_t create_frames_ret;
     create_frames_of_region_ret_t extra_bi_ret;
 
-    /* convert from physical addresses to userland vptrs */
+    /* Conversion from physical addresses to userland vptrs uses pv_offset,
+     * which is defined as:
+     *     virt_address + pv_offset = phys_address
+     * The offset is usually a positive value, because the virtual address of
+     * the user image is a low value and the actually physical address is much
+     * greater. However, there is no restrictions on the physical and virtual
+     * image location in general. Defining pv_offset as a signed value might
+     * seem the intuitive choice how to handle this, but there are two catches
+     * here that break the C rules. We can't cover the full integer range then,
+     * and overflows/underflows are well defined for unsigned values only. They
+     * are undefined for signed values, even if such operations practically work
+     * in many cases due to how compiler and machine implement negative integers
+     * using the two's-complement.
+     * Assume a 32-bit system with virt_address=0xc0000000 and phys_address=0,
+     * then pv_offset would have to be -0xc0000000. This value is not in the
+     * 32-bit signed integer range. Calculating '0 - 0xc0000000' using unsigned
+     * integers, the result is 0x40000000 after an underflow, the reverse
+     * calculation '0xc0000000 + 0x40000000' results in 0 again after overflow.
+     * If 0x40000000 is a signed integer, the result is likely the same, but the
+     * whole operation is undefined by C rules.
+     */
     v_region_t ui_v_reg = {
         .start = ui_p_reg_start - pv_offset,
         .end   = ui_p_reg_end   - pv_offset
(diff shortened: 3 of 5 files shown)


lsf37:
Ok, proofs succeeded without change, test works together with seL4/seL4_tools#100 and I'm happy with this from the verification side. It'd be good if someone from the systems side could also take a look.

seL4/seL4_tools#100 should be merged first.

lsf37 (approved):

axel-h:
fixed a typo

axel-h:
Setting this to draft, as I still have to rework this to drop using `sword_t`

axel-h:
This should be the final version now. Do you thing the comments/docu is sufficient?

CHANGES
```diff
@@ -22,12 +22,16 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
 ---
-Upcoming release: BINARY COMPATIBLE
+Upcoming release: BREAKING
 
 ## Changes
 
  * Added support for the ARM Cortex A55
  * Added support for the ODroid C4
+ * Changed kernel boot parameters to become architecture agnostic with using word_t instead of uint32. Stop using
+   sword_t due to C rule violations, use word_t instead and explicitly allow overflows and underflows in the
```
lsf37, line 32: `sword_t` never made it into a permanent branch, so the reference to it in the comment is probably confusing.

kent-mcleod: sword_t is present in this repository.  It just never made it into the elfloader repository.

include/arch/arm/arch/kernel/boot.h
```diff
@@ -12,12 +13,15 @@ cap_t create_unmapped_it_frame_cap(pptr_t pptr, bool_t use_large);
...
+ * the signature.
+ */
 void init_kernel(
-    paddr_t ui_p_reg_start,
-    paddr_t ui_p_reg_end,
-    sword_t pv_offset,
-    vptr_t  v_entry,
-    paddr_t dtb_addr_p,
-    uint32_t dtb_size
+    word_t ui_p_reg_start,
```
kent-mcleod, line 20: I understand changing the types for dtb_size and pv_offset, but the other parameters don't need to change.  It's still valid to use types that are aliases of word_t.

CHANGES
```diff
@@ -22,12 +22,16 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 Further information about [seL4 releases](https://docs.sel4.systems/sel4_release/) is available.
 
 ---
-Upcoming release: BINARY COMPATIBLE
+Upcoming release: BREAKING
```
kent-mcleod, line 25: I don't think that this is a breaking change.

CHANGES
lsf37: Oh yes, you're right. I double checked and still managed to miss it, sorry. All good then.

src/arch/arm/kernel/boot.c
```diff
@@ -315,7 +315,27 @@ static BOOT_CODE bool_t try_init_kernel(
     create_frames_of_region_ret_t create_frames_ret;
     create_frames_of_region_ret_t extra_bi_ret;
 
-    /* convert from physical addresses to userland vptrs */
+    /* Convert from physical addresses to userland vptrs with the parameter
+     * pv_offset, which is defined as:
+     *     virt_address + pv_offset = phys_address
+     * The offset is usually a positive value, because the virtual address of
```
yyshen, line 321: Well, it is not unusual that physical memory starts from 0.

src/arch/arm/kernel/boot.c
```diff
@@ -315,7 +315,27 @@ static BOOT_CODE bool_t try_init_kernel(
...
+     * the user image is a low value and the actually physical address is much
+     * greater. However, there is no restrictions on the physical and virtual
+     * image location in general. Defining pv_offset as a signed value might
+     * seem the intuitive choice how to handle this, but there is are two
```
yyshen, line 325: "there is are"

src/arch/arm/kernel/boot.c
```diff
@@ -315,7 +315,27 @@ static BOOT_CODE bool_t try_init_kernel(
...
+     * catches here that break the C rules. We can't cover the full integer
+     * range then, and overflows/underflows are well defined for unsigned values
+     * only. They are undefined for signed values, even if such operations
+     * practically work in many cases due to how compiler and machine implement
+     * negative integers using the two's-complement.
+     * Assume a 32-bit system with virt_address=0xc0000000 and phys_address=0,
```
yyshen, line 331: I do not quite get the example here.

src/arch/riscv/kernel/boot.c
```diff
@@ -196,7 +197,27 @@ static BOOT_CODE bool_t try_init_kernel(
...
+     * the user image is a low value and the actually physical address is much
+     * greater. However, there is no restrictions on the physical and virtual
+     * image location in general. Defining pv_offset as a signed value might
+     * seem the intuitive choice how to handle this, but there is are two
+     * catches here that break the C rules. We can't cover the full integer
+     * range then, and overflows/underflows are well defined for unsigned values
+     * only. They are undefined for signed values, even if such operations
+     * practically work in many cases due to how compiler and machine implement
+     * negative integers using the two's-complement.
+     * Assume a 32-bit system with virt_address=0xc0000000 and phys_address=0,
```
yyshen, line 213: I think the comment can be put in the commit message instead of being repeated twice for different architectures.

yyshen (changes requested):

include/arch/arm/arch/kernel/boot.h
axel-h: Die idae is, that at this point we are called from a raw assembly interface, so all parameters are register-sized and thus word_t is used. This function will then to the first level of semantics/casting to put the values into a specific domain. This allows the compiler to emit a warning if there are issue with the cast and we do not have to trust on implicit assumption that one type is an alias of another type.

CHANGES
axel-h: I will test this again, but locally I had an issue that on RISC-V an old elf loader and an new kernel don' fit together, this started all the cleanup activity here.

include/arch/arm/arch/kernel/boot.h
```diff
@@ -12,12 +13,15 @@ cap_t create_unmapped_it_frame_cap(pptr_t pptr, bool_t use_large);
 cap_t create_mapped_it_frame_cap(cap_t pd_cap, pptr_t pptr, vptr_t vptr, asid_t asid, bool_t use_large,
                                  bool_t executable);
 
+/* This is called from assembly code and thus there are no specific types in
+ * the signature.
+ */
 void init_kernel(
-    paddr_t ui_p_reg_start,
```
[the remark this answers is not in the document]
axel-h: Some years back I asked if there is an assumption that virt and phys addresses are always the same i the context of ARM's LPAE with 64bit phys addresses and 32-bit virt addresses. The answer then was, that this is not supported and all are assumed to fit in a word_t. Furthermore, this is raw binary interface (even with some asm stub code in between), where we only have registers (ie `word_t`). Thus I think it is cleaner to keep the interface strict and to the casting in the function then.

src/arch/arm/kernel/boot.c
axel-h: I'm unsure do you want me to change anything here? We have platform where this is the case and platform there this is not the case.

src/arch/arm/kernel/boot.c
axel-h: fixed

src/arch/riscv/kernel/boot.c
axel-h: I can add it there also, but my vie is that people mostly look at the code and less at the commit messages in the long run, that is why I put the comment here. One day we might be able to unify the ARM and RISC-V code here, so this exists only once, but for now RISC-V cam into existence as a clone of ARM, thus we have some duplication.

include/arch/arm/arch/kernel/boot.h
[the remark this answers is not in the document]
lsf37: I do agree that there is a fundamental assumption in the kernel at the moment that virtual space is equal to or larger than physical space. If we do at some point want to support physical addresses that are wider than virtual addresses, we will need a bigger review and change. So I'd be fine with making that assumption explicit here.

src/arch/riscv/kernel/boot.c
```diff
@@ -1,6 +1,7 @@
 /*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright 2015, 2016 Hesham Almatary <heshamelmatary@gmail.com>
+ * Copyright 2021, HENSOLDT Cyber
```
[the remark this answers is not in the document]
lsf37: You should add one if you've made a substantial change. We've been quite relaxed about what constitutes "substantial". If we accumulate too many copyright lines for minor changes we might have to define it more strictly.

include/arch/arm/arch/kernel/boot.h
kent-mcleod: Ok. This does make more sense now.

src/arch/arm/kernel/boot.c
axel-h: What exactly is unclear here? We assume we are running vom phys address 0 and are linked on virtual address 0xc000000. The calculations will then leave the signed integer range, that is why we can't use signed integers here. Technically, things still works because of the 2s-complement number representation. In the end all this is about staying in the valid C domain, because over/underflow on unsigned values is undefined (even if it likely works)

CHANGES
axel-h: I've remove this, as it is not breaking on the upper layer. It may only affect custom bootloaders.

axel-h:
Dropping this one and split it up. See https://github.com/seL4/seL4/pull/704 for the actual issue to fix.

PR #495, opened 2021-08-20
github workflow improvements

lsf37:
- fix concurrency groups
- better name for kernel compile action
- show badge for full sel4test run


lsf37:
(When this is approved, I'll switch the "required" compile test to the new name before merging, just don't want to do that yet right now, because this will make an impossible requirement for other PRs)

mbrcknl (approved):

PR #496, opened 2021-08-20
mcs: sc_active not always true in sc_sporadic

lsf37:
Turns out the invariant 17109eb8c986cd0ac refers to (discussed in #476) is hard to prove because it is not true, and the runtime check is necessary. This assertion fails in sel4test SCHED_CONTEXT_0003 (Basic api_sc_bind/UnbindObject testing).


mktnk3 (approved):

lsf37:
(Cancelling the hardware test to reduce load. The MCS test that previously failed now does work again).

PR #502, opened 2021-08-22
boot: don't enforcing defining MODE_RESERVED

axel-h:
On AARCH32 there is a special need for an ASID region. Since no other architecture needs this mechanism, handle this within the AARCH32 specific code and keep the generic code clean.

### DIFF
--- a/include/arch/arm/arch/bootinfo.h
+++ b/include/arch/arm/arch/bootinfo.h
@@ -18,11 +18,19 @@
  */
 #define MAX_NUM_FREEMEM_REG (ARRAY_SIZE(avail_p_regs) + MODE_RESERVED + 1 + 1)
 
+/* The regions reserved by the boot code are:
+ * +1 for kernel
+ * +1 for device tree binary
+ * +1 for user image.
+ * +1 for each the MODE_RESERVED region, there might be none
+ */
+#define NUM_RESERVED_REGIONS (3 + MODE_RESERVED)
+
+
 /* The maximum number of reserved regions is:
  * +1 for each free memory region (MAX_NUM_FREEMEM_REG)
  * +1 for each kernel frame (NUM_KERNEL_DEVICE_FRAMES, there might be none)
- * +1 for each mode-reserved region (MODE_RESERVED)
- * +1 for each region reserved by the boot code (3: kernel, dtb, user image)
+ * +1 for each region reserved by the boot code (NUM_RESERVED_REGIONS)
  */
 #define MAX_NUM_RESV_REG (MAX_NUM_FREEMEM_REG + NUM_KERNEL_DEVICE_FRAMES + \
-                          MODE_RESERVED + 3)
+                          NUM_RESERVED_REGIONS)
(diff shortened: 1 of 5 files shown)


lsf37:
From the verification side, I'm relatively neutral on this, although I'd prefer the fewer `#ifdef`s of the old version.

From the code side, I'm not entirely convinced that this is really cleaner than before. Isn't it nicer to have the concept of a potential extra region for arch specific code? Currently only Arm32 needs it, but that might not stay that way forever, the current set of architectures is not all that could be supported in the future. It doesn't look like the new code is much shorter either.

Happy to be convinced otherwise, but so far I don't see it meeting the benefit threshold yet.

kent-mcleod:
> On AARCH32 there is a special need for an ASID region. Since no other architecture needs this mechanism, handle this within the AARCH32 specific code and keep the generic code clean.

AARCH32 isn't the only arch with this need.  Any platform that has less hardware ASIDs than seL4 ASIDs will require a way of linking the HW ASID with root pagetable object. Currently this can only be achieved by reserving a slot in the root page table and making it an invalid mapping that also holds the current hw_asid. aarch32 isn't the only architecture that has this constraint. RISC-V will need this once a CPU is added that has HWASIDs but less than 16bits of ASID space and greater than 1Bit of asid space.

axel-h:
@kent-mcleod: Thanks for the explanation. I think we can still add for RISC-V after my changes easily.
@lsf37: The main motivation for this is removing defines that are just dummies in some architectures and make the generic code simpler by forcing platforms/architectures to set the defines right. I have more patches on top of this that would allow removing the `#ifdef` eventually and use helper functions the for the memory setup.

yyshen:
> From the verification side, I'm relatively neutral on this, although I'd prefer the fewer `#ifdef`s of the old version.
> 
> From the code side, I'm not entirely convinced that this is really cleaner than before. Isn't it nicer to have the concept of a potential extra region for arch specific code? Currently only Arm32 needs it, but that might not stay that way forever, the current set of architectures is not all that could be supported in the future. It doesn't look like the new code is much shorter either.
> 
> Happy to be convinced otherwise, but so far I don't see it meeting the benefit threshold yet.

@lsf37 I think the changes also need reverification effort which seems not trivial?

axel-h:
I've reworked this now. It's split it in a set of changes that add comments and some refactoring to reduce global dependencies. That will also be helpful for further changes. The ARM part is left as it is as much as possible and `MODE_RESERVED` is not removed there. The changes left out from this PR are parked in https://github.com/axel-h/seL4/commits/patch-axel-14a. Depending how the other merges go, these change can be dropped anyway, as I'm aiming for a generic function `reserve_region()` that can be called in various places in the boot code. It handles managing/merging the reserved regions internally then, so no care needs to be taken about the order when calling it. That make handling dynamically reserved areas on all architectures much easier.

axel-h:
can we merge this?

kent-mcleod:
> can we merge this?

Just checking the proofs are happy with the extra +0 and if blocks

kent-mcleod (approved):

PR #505, opened 2021-08-23
mcs: Remove domain time check from preemptionPoint (v2)

lsf37:
As Curtis said: this removes the operations that trigger a reschedule or reprogram the timer from `preemptionPoint` to ensure the relevant state updates in the proof occur where they are easier to verify.

This simplifies and supersedes #480. The idea is that it is simpler and easier in both proof and code to do the check once after the main kernel event is finished, and on the path that has to deal with checking time and scheduling anyway.

The preemption point and all other places that call `updateTimestamp` now just flag that the domain will have to be changed (`ksDomainTime = 0`) and don't change the scheduler action/timer directly, so the proofs that depend on the scheduler action only have to deal with that state update once, but still before anything that depends on it.

### DIFF
--- a/include/kernel/thread.h
+++ b/include/kernel/thread.h
@@ -229,18 +229,24 @@ static inline void updateTimestamp(void)
     time_t consumed = (NODE_STATE(ksCurTime) - prev);
     NODE_STATE(ksConsumed) += consumed;
     if (CONFIG_NUM_DOMAINS > 1) {
-
         if ((consumed + MIN_BUDGET) >= ksDomainTime) {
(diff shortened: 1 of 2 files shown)


michaelmcinerney (approved):

kent-mcleod (approved):

michaelmcinerney:
Proofs are done https://github.com/seL4/l4v/pull/343

lsf37:
Hardware tests succeeded as well, so I'll merge this when the checks come back for the rebased version.

PR #510, opened 2021-08-24
x86: Always eliminate function prologue for idle_thread()

cl91:
The idle_thread() cannot perform any stack manipulations since it runs in the idle thread TCB context. Declare the function with the naked attribute, to ensure that the compiler always eliminates the function prologue. Previously we rely on the -O2 optimization flag for this, which on certain compilers may not guarantee that the function prologue gets eliminated.

This is in response to a kernel crash on ia32 when compiled with clang 12.0.1. See  https://sel4.discourse.group/t/kernel-crash-on-x86-but-not-on-x86-64/398

### DIFF
--- a/src/arch/x86/idle.c
+++ b/src/arch/x86/idle.c
@@ -7,11 +7,21 @@
 #include <config.h>
 #include <api/debug.h>
 
-void idle_thread(void)
+/*
+ * The idle thread does not have a dedicated stack and runs in
+ * the context of the idle thread TCB. Make sure that the compiler
+ * always eliminates the function prologue by declaring the
+ * idle_thread with the naked attribute.
+ */
+__attribute__((naked)) void idle_thread(void)
 {
-    while (1) {
-        asm volatile("hlt");
-    }
+    /* We cannot use for-loop or while-loop here because they may
+     * involve stack manipulations (the compiler will not allow
+     * them in a naked function anyway). */
+    asm volatile(
+        "1: hlt\n"
+        "jmp 1b"
+    );
 }
 
 /** DONT_TRANSLATE */
--- a/src/arch/x86/idle.c
+++ b/src/arch/x86/idle.c
@@ -13,7 +13,7 @@
  * always eliminates the function prologue by declaring the
  * idle_thread with the naked attribute.
  */
-__attribute__((naked)) void idle_thread(void)
+__attribute__((naked)) NORETURN void idle_thread(void)
 {
     /* We cannot use for-loop or while-loop here because they may
      * involve stack manipulations (the compiler will not allow


src/arch/x86/idle.c
```diff
@@ -7,11 +7,18 @@
...
+ * the context of the idle thread TCB. Make sure that the compiler
+ * always eliminates the function prologue by declaring the
+ * idle_thread with the naked attribute.
+ */
+__attribute__((naked)) void idle_thread(void)
 {
-    while (1) {
-        asm volatile("hlt");
-    }
+    asm volatile(
```
axel-h, line 18: Can you add a warning here that a `for(;;) { ... }` or ``while(1) { ... }` is also dangerous because the compiler creates additional code then that might need stack.

cl91: Ok yeah I can do that. The compiler would not allow any `for` or `while` statement in a naked function anyway --- the function body of a naked function must be one big `asm` statement.

cl91: Done!

yyshen (approved):

src/arch/x86/idle.c
axel-h: This seems a way to write plain asm code in a C file also without the pitfalls of `naked` (only simple asm, no parameters...), but I wonder if the C parser used in verification can hadle this.
 
```
asm("                     \n\
  _asm_idle_thread:       \n\
    hlt                   \n\
	jmp _asm_idle_thread  \n\
");
int idle_thread() asm ("_asm_idle_thread"); 

int kernel_has_nothing_to_do() {
        idle_thread();
        return 0;
}
```

Using a dedicated `idle.s` might be the better approach then?

src/arch/x86/idle.c
```diff
@@ -7,11 +7,21 @@
...
 #include <api/debug.h>
 
-void idle_thread(void)
+/*
+ * The idle thread does not have a dedicated stack and runs in
+ * the context of the idle thread TCB. Make sure that the compiler
+ * always eliminates the function prologue by declaring the
+ * idle_thread with the naked attribute.
+ */
+__attribute__((naked)) void idle_thread(void)
```
axel-h, line 16: should add the noreturn attribute also

src/arch/x86/idle.c
cl91: We would need two `idle.s` files (one for `x86` one for `x86_64`) that are identical so there might be a duplication of code problem there. The `naked` attribute seems to be the clearest way of expressing the intention here.

src/arch/x86/idle.c
cl91: Yeah true the function should be marked `noreturn`. I think it's better to do that in a separate commit though to keep changes small.

axel-h: just add the commit to this PR then.

cl91: Done!

axel-h:
@lsf37: how well do the proofs handle inline assembly and the naked attribute?

lsf37:
> @lsf37: how well do the proofs handle inline assembly and the naked attribute?

It depends on how the function is referenced in the proofs. The parser as such accepts it (CParser test in the checks succeeded), but it wouldn't create semantics for the function. It didn't do that before either, because there was inline asm, so it is most likely fine. The easiest way to find out for sure is to just run the proofs, which I've started now.

lsf37 (commented):
Two trivial things before we can merge:

* The style checker wants one more space indentation and the closing parenthesis to line up with the `asm`

```diff
     asm volatile(
-	"1: hlt\n"
-	"jmp 1b"
-	);
+        "1: hlt\n"
+        "jmp 1b"
+    );
```

* commit message checker wants the title to be shorter (50 chars max). Probably easiest to leave out the word "function"

(I can also do these for you if you want)

cl91:
@lsf37 Done! The indentation lint complaint was about replacing the tab with 8 spaces.

lsf37:
@kent-mcleod and @yyshen what do you think of @axel-h's concern about the `bare` attribute being buggy in compiler implementations?

It does seem to work correctly in our situation, but that is of course also what we thought about `-O2` before. At least this time it would be an actual compiler bug.

yyshen:
> @kent-mcleod and @yyshen what do you think of @axel-h's concern about the `bare` attribute being buggy in compiler implementations?
> 
> It does seem to work correctly in our situation, but that is of course also what we thought about `-O2` before. At least this time it would be an actual compiler bug.

I think you meant the "naked" attribute? Functions marked with "naked" should only allow basic asm statement, so I think we do not need to worry about "while (1)" in naked functions. Previously we rely on that the compiler is smart enough with -O2 to optimise the prologue and epilogue. We could always put the function in an assembly file later, if the current fix still fails on some compilers.

lsf37:
> If we think this is a decent strategy, then we should probably also go through all the other arch idle_thread implementations and do the same. aarch32 has a `FORCE_O2` attribute, and aarch64 is still written in C without any attribute. The RISC-V idle thread stands out because it goes the route of using the same stack pointer as the kernel stack with it probably shouldn't be doing.

I think that is a good plan. Will merge this PR for now and put the rest up as an issue.

HHPLow:
Hi @cl91 @lsf37 
When I following the seL4 doc to build Stand-alone seL4 and seL4test
Will meet compile error like below

```text
seL4_build# ninja kernel.elf
[33/34] Building C object CMakeFiles/kernel.elf.dir/kernel_all.c.obj
FAILED: CMakeFiles/kernel.elf.dir/kernel_all.c.obj
/usr/bin/ccache /usr/bin/gcc --sysroot=/ws/seL4_build  -I/ws/seL4/include -I/ws/seL4/include/64 -I/ws/seL4/include/arch/x86 -I/ws/seL4/include/arch/x86/arch/64 -I/ws/seL4/include/plat/pc99 -I/ws/seL4/include/plat/pc99/plat/64 -I/ws/seL4/libsel4/include -I/ws/seL4/libsel4/arch_include/x86 -I/ws/seL4/libsel4/sel4_arch_include/x86_64 -I/ws/seL4/libsel4/sel4_plat_include/pc99 -I/ws/seL4/libsel4/mode_include/64 -I/ws/seL4_build/gen_config -I/ws/seL4_build/autoconf -I/ws/seL4_build/gen_headers -I/ws/seL4_build/generated -m64  -march=nehalem -D__KERNEL_64__ -O3 -DNDEBUG -nostdinc -nostdlib -O2 -DHAVE_AUTOCONF -mcmodel=kernel -fno-pic -fno-pie -fno-stack-protector -fno-asynchronous-unwind-tables -std=c99 -Wall -Werror -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Wmissing-declarations -Wundef -Wpointer-arith -Wno-nonnull -ffreestanding -fno-common -mno-mmx -mno-sse -mno-sse2 -mno-3dnow -MD -MT CMakeFiles/kernel.elf.dir/kernel_all.c.obj -MF CMakeFiles/kernel.elf.dir/kernel_all.c.obj.d -o CMakeFiles/kernel.elf.dir/kernel_all.c.obj -c /ws/seL4_build/kernel_all.c
/ws/seL4/src/arch/x86/idle.c:17:1: error: ‘naked’ attribute directive ignored [-Werror=attributes]
 {
 ^
/ws/seL4/src/arch/x86/idle.c: In function ‘idle_thread’:
/ws/seL4/src/arch/x86/idle.c:25:1: error: ‘noreturn’ function does return [-Werror]
 }
 ^
cc1: all warnings being treated as errors
```

When I remove the `NORETURN`, still error
```text
seL4_build# ninja kernel.elf
[21/22] Building C object CMakeFiles/kernel.elf.dir/kernel_all.c.obj
FAILED: CMakeFiles/kernel.elf.dir/kernel_all.c.obj
/usr/bin/ccache /usr/bin/gcc --sysroot=/ws/seL4_build  -I/ws/seL4/include -I/ws/seL4/include/64 -I/ws/seL4/include/arch/x86 -I/ws/seL4/include/arch/x86/arch/64 -I/ws/seL4/include/plat/pc99 -I/ws/seL4/include/plat/pc99/plat/64 -I/ws/seL4/libsel4/include -I/ws/seL4/libsel4/arch_include/x86 -I/ws/seL4/libsel4/sel4_arch_include/x86_64 -I/ws/seL4/libsel4/sel4_plat_include/pc99 -I/ws/seL4/libsel4/mode_include/64 -I/ws/seL4_build/gen_config -I/ws/seL4_build/autoconf -I/ws/seL4_build/gen_headers -I/ws/seL4_build/generated -m64  -march=nehalem -D__KERNEL_64__ -O3 -DNDEBUG -nostdinc -nostdlib -O2 -DHAVE_AUTOCONF -mcmodel=kernel -fno-pic -fno-pie -fno-stack-protector -fno-asynchronous-unwind-tables -std=c99 -Wall -Werror -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Wmissing-declarations -Wundef -Wpointer-arith -Wno-nonnull -ffreestanding -fno-common -mno-mmx -mno-sse -mno-sse2 -mno-3dnow -MD -MT CMakeFiles/kernel.elf.dir/kernel_all.c.obj -MF CMakeFiles/kernel.elf.dir/kernel_all.c.obj.d -o CMakeFiles/kernel.elf.dir/kernel_all.c.obj -c /ws/seL4_build/kernel_all.c
/ws/seL4/src/arch/x86/idle.c:17:1: error: ‘naked’ attribute directive ignored [-Werror=attributes]
 {
 ^
cc1: all warnings being treated as errors
```

until I change this file to
```diff
diff --git a/src/arch/x86/idle.c b/src/arch/x86/idle.c
index 82bd30b6..c0a8cc88 100644
--- a/src/arch/x86/idle.c
+++ b/src/arch/x86/idle.c
@@ -13,7 +13,7 @@
  * always eliminates the function prologue by declaring the
  * idle_thread with the naked attribute.
  */
-__attribute__((naked)) NORETURN void idle_thread(void)
+void idle_thread(void)
 {
     /* We cannot use for-loop or while-loop here because they may
      * involve stack manipulations (the compiler will not allow
```

Is there something wrong with my toolchain?

Thanks

lsf37:
It appears that the compiler doesn't understand the attribute. What `gcc` version are you using?

(This is a PR that is merged already. Can you please open a new issue for this so we can continue there?)

HHPLow:
> It appears that the compiler doesn't understand the attribute. What `gcc` version are you using?
> 
> (This is a PR that is merged already. Can you please open a new issue for this so we can continue there?)

Please see this issue https://github.com/seL4/seL4/issues/574

PR #517, opened 2021-08-28
improve comments and style

axel-h:
collect some trivial things form the other PR that could be merged easily:
- trivial: improve style for code and comments
- add comment about empty Arch_finaliseInterrupt()
- add comment about handleUnknownSyscall()

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -538,9 +538,12 @@ static BOOT_CODE bool_t try_init_kernel(
     /* finalise the bootinfo frame */
     bi_finalise();
 
-    /* make everything written by the kernel visible to userland. Cleaning to PoC is not
-     * strictly neccessary, but performance is not critical here so clean and invalidate
-     * everything to PoC */
+    /* Flushing the L1 cache and invalidating the TLB is good enough here to
+     * make sure everything written by the kernel is visible to userland. There
+     * are no uncached userland frames at this stage that require enforcing
+     * flushing to RAM. Any retyping operation will clean the memory down to RAM
+     * anyway.
+     */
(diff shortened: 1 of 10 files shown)


src/arch/arm/kernel/boot.c
```diff
@@ -538,16 +538,16 @@ static BOOT_CODE bool_t try_init_kernel(
...
     bi_finalise();
 
-    /* make everything written by the kernel visible to userland. Cleaning to PoC is not
-     * strictly neccessary, but performance is not critical here so clean and invalidate
-     * everything to PoC */
+    /* make everything written by the kernel visible to userland. Cleaning to
+     * PoC is not strictly neccessary, but performance is not critical here so
+     * clean and invalidate everything to PoC
+     */
     cleanInvalidateL1Caches();
```
lsf37, line 545: Nothing to do with your PR, but it looks like that comment is actually wrong. `cleanInvalidateL1Caches()` doesn't clean to PoC. This was wrong already in the commit that introduced it in 2013. I think the code is right in this case: flushing L1 and invalidating the TLB is good enough here, unless we are handing any frames to user level uncached that we modified here, which I don't think we do.

The rest is handled in `retype`, which (now) does flush correctly.

lsf37 (approved):
LGTM. We can fix that comment in a separate PR if you like.

src/arch/arm/kernel/boot.c
axel-h: I'll extend the comment.

axel-h: change the comment, see https://github.com/seL4/seL4/pull/517/commits/c75e6c76d09309b0d3db795f74ef050430743df1

lsf37 (approved):

PR #518, opened 2021-08-29
fastpath: placate clang-11

lsf37:
clang-11 warns "converting the enum constant to a boolean". The comparison generates the same code, since the expression can be evaluated at compile time (I checked the objdump).

This is in preparation for upgrading the docker containers to clang-11 and gcc-10 (see initial discussion on seL4/seL4-CAmkES-L4v-dockerfiles#39 -- debian bullseye is released, which means our docker build no longer works).


lsf37:
This might need a proof update, but maybe we're lucky and automation happens to hide it.

lsf37:
Yup, proofs still succeed.

yyshen (approved):

PR #523, opened 2021-08-30
Allow Lazy SC Rebind

Indanz:
Before, binding a scheduling context to a TCB was not allowed if the
SC was bound to a notification object. Also, binding a SC to a NF was
not allowed if that scheduling context was already bound to a TCB.

Without these restriction it is much easier to move scheduling contexts
around: In effect having a SC bound on both the TCB and a NF acts as if
the thread is running on a donated SC which will be returned when the
tasks calls Recv/Wait, which is done by maybeReturnSchedContext().

This greatly simplifies startup and time-out exception handling for
passive threads having a NF with a SC bound to it.

[Edit for test:
Test with seL4/seL4test#50 ]

### DIFF
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -46,7 +46,7 @@ or \apifunc{seL4\_TCB\_Configure}{tcb_configure} methods and then calling
 \apifunc{seL4\_TCB\_WriteRegisters}{tcb_writeregisters} with an initial stack pointer and instruction
 pointer. The thread can then be activated either by setting the
 \texttt{resume\_target} parameter in the \apifunc{seL4\_TCB\_WriteRegisters}{tcb_writeregisters} invocation to true
-or by seperately calling the \apifunc{seL4\_TCB\_Resume}{tcb_resume} method. Both of these methods
+or by separately calling the \apifunc{seL4\_TCB\_Resume}{tcb_resume} method. Both of these methods
 place the thread in a runnable state.
 
 On the master kernel, this will result in the thread immediately being added to
@@ -98,7 +98,7 @@ controlled priority (MCP)] the highest priority a thread can set itself or anoth
 
 \subsection{MCS Scheduling}
 
-This section only applies to configrations with MCS enabled, where threads must have
+This section only applies to configurations with MCS enabled, where threads must have
 a scheduling context object available in order to be admitted to the scheduler.
 
 \subsection{Scheduling Contexts}
@@ -112,7 +112,7 @@ The tuple $(b, p)$ forms an upper bound on the thread's execution -- the kernel 
 thread to run for more than $b$ out of every $p$ microseconds. However, $\frac{b}{p}$ does not
 represent a lower bound on execution, as a thread must have the highest or equal highest priority
 of all runnable threads to be guaranteed to be scheduled at all, and the kernel does not conduct
-an admission test. As a result the set of all parameters is not neccesserily schedulable. If
+an admission test. As a result the set of all parameters is not necessarily schedulable. If
 multiple threads have available budget concurrently they are scheduled first-in first-out, and
 round-robin scheduling is applied once the budget is expired.
 
@@ -157,7 +157,7 @@ limits the number of refills for that specific scheduling context. This value is
 value of 2 and is limited by the size of the scheduling context.
 \end{itemize}
 
-Threads that have short exection times (e.g interrupt handlers) and are not frequently preempted
+Threads that have short execution times (e.g interrupt handlers) and are not frequently preempted
 should have less refills, while longer running threads with long values of $b$ should have a higher
 value. Threads bound to a scheduling context with 0 extra refills will behave periodically -- tasks
 that use their head replenishment, or call yield, will not be scheduled again until the start of
@@ -186,7 +186,7 @@ the scheduling context is active, it will be added to the scheduler.
 Threads can be unbound from a scheduling context with
 \apifunc{seL4\_SchedContext\_UnbindObject}{schedcontext_unbindobject}.  This is distinct from
 suspending a thread, in that threads that are blocked waiting in an endpoint or notification queue
-will remain in the queue and can still recieve messages and signals.  However, the unbound thread
+will remain in the queue and can still receive messages and signals.  However, the unbound thread
 will not be schedulable again until it receives a scheduling context.  Threads without scheduling
 contexts are referred to as \emph{passive} threads, as they cannot execute without the action of
 another thread.
@@ -271,11 +271,11 @@ Threads of sufficient maximum controlled priority and with possession of the
 appropriate scheduling context capability can manipulate the scheduler and
 implement user-level schedulers using IPC.
 
-Scheduling contexts provide access to and an upper bound on exection CPU time,
+Scheduling contexts provide access to and an upper bound on execution CPU time,
 however when a thread executes is determined by thread priority.  Consequently,
 access to CPU is a function of thread MCPs, scheduling contexts and the
 \obj{SchedControl} capability.  The kernel will enforce that threads do not
-exceeed the budget in their scheduling context for any given period, and that
+exceed the budget in their scheduling context for any given period, and that
 the highest priority thread will always run, however it is up to the system
 designer to make sure the entire system is schedulable.
 
@@ -297,7 +297,7 @@ capability address for the exception handler can be associated with a thread.
 This address is then used to lookup the handler endpoint, and the capability to
 the endpoint is installed into the threads' kernel CNode.  For threads without
 an exception handler, a null capability can be used, however the consequences
-are different per exeception handler type.  Before raising an exception the
+are different per exception handler type.  Before raising an exception the
 handler capability is validated. The kernel does not perform another lookup,
 but checks that the capability is an endpoint with the correct rights.
 
@@ -565,7 +565,7 @@ Timeout faults are raised when a thread consumes all of its budget and has a
 timeout fault handler that is not a null capability.  They allow a timeout
 exception handler to take some action to restore the thread, and deliver a
 message containing the scheduling context data word, as well as the amount of
-time consumed since the last timeout fault occured on this scheduling context,
+time consumed since the last timeout fault occurred on this scheduling context,
 or since \apifunc{seL4\_SchedContext\_YieldTo}{schedcontext_yieldto} or
 \apifunc{seL4\_SchedContext\_Consumed}{schedcontext_consumed} was last called.
 Timeout exception handlers can reply to a temporal fault with the registers set
@@ -574,7 +574,7 @@ in the same format as outlined in \autoref{sec:read_write_registers}.
 \begin{table}[htb] \noindent\begin{tabularx}{\textwidth}{XX} \toprule
     \textbf{Meaning} & \textbf{IPC buffer location} \\ \midrule Data word from
     the scheduling context object that the thread was running on when the fault
-    occured. & \ipcbloc{seL4\_TimeoutFault\_Data} \\ Upper 32-bits of
+    occurred. & \ipcbloc{seL4\_TimeoutFault\_Data} \\ Upper 32-bits of
     microseconds consumed since last reset &
     \ipcbloc{seL4\_TimeoutFault\_Consumed} \\ Lower 32-bits of microseconds
     consumed since last reset & \ipcbloc{seL4\_TimeoutFault\_Consumed\_LowBits}
@@ -640,7 +640,7 @@ a \obj{VCPU} to a \obj{TCB} is done by an invocation on the \obj{VCPU} only, and
 
 The provided objects and invocations are, generally speaking, the thinnest possible shim over
 the underlying hardware primitives and operations. As a result an in depth familiarity with
-the underlying architecture specific harwdare mechanisms is required to use these objects, and
+the underlying architecture specific hardware mechanisms is required to use these objects, and
 such familiarity is therefore assumed in description.
 
 \subsection{Arm}
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -253,7 +253,10 @@ Scheduling contexts can also be bound to notification objects using
 a notification object with a passive thread blocked waiting on it, the passive thread will receive
 the scheduling context that is bound to the notification object.  The scheduling context is returned
 when the thread blocks on the notification object.  This feature allows for passive servers to use
-notification binding (See \autoref{sec:notification-binding}).
+notification binding (See \autoref{sec:notification-binding}).  If a scheduling context is bound to
+both a notification object and a thread, the behaviour will be the same as for a passive server:
+The scheduling context will be unbound from the thread when it blocks on the bound notification object.
+This is useful when launching passive servers or handling timeout exceptions.
 
 Scheduling contexts can be unbound from all objects (notification objects and TCBs that are bound or
 have received a scheduling context through donation) using
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -94,14 +94,14 @@ static exception_t decodeSchedContext_Bind(sched_context_t *sc)
 
     cap_t cap = current_extra_caps.excaprefs[0]->cap;
 
-    if (sc->scTcb != NULL || sc->scNotification != NULL) {
-        userError("SchedContext_Bind: sched context already bound.");
-        current_syscall_error.type = seL4_IllegalOperation;
-        return EXCEPTION_SYSCALL_ERROR;
-    }
-
     switch (cap_get_capType(cap)) {
     case cap_thread_cap:
+        if (sc->scTcb != NULL) {
+            userError("SchedContext_Bind: sched context already bound.");
+            current_syscall_error.type = seL4_IllegalOperation;
+            return EXCEPTION_SYSCALL_ERROR;
+        }
+
         if (TCB_PTR(cap_thread_cap_get_capTCBPtr(cap))->tcbSchedContext != NULL) {
             userError("SchedContext_Bind: tcb already bound.");
             current_syscall_error.type = seL4_IllegalOperation;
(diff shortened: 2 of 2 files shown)


Indanz:
Conceptually what is missing is a way to donate a SC, which will be released when a task would block, outside of the normal call and notification paths. If it's unwanted that the behaviour of seL4_SchedContext_Bind() changes, a new syscall can be introduced.

With this patch the launch of a passive task can become:

`seL4_SchedContext_Bind(sc_cap, tcb_cap);
seL4_SchedContext_Bind(sc_cap, nf_cap);`

and then just start the task as usual, with no racy or failure prone synchronisation between passive task and the task that starts the thread. Without this patch the launcher tasks is forced to make a call to the to-be passive task which needs to call ReplyRecv(). (As far as I know there is no other robuust way of launching a passive task.)

Same for timeout-exception handling: The handler can just assign a temporary SC when the exception occurs, and that SC will be released automatically when it's not needed anymore. Without this the handler and the passive tasks need to know when to return the temporary SC, which is nontrivial.

(It seems we're not going to use passive servers because those don't work well with sporadic clients, but for others than want to retain sanity this seems like a nice feature to have.)

pingerino:
This is an interesting idea and does make configuration of passive servers much easier. 

I'm worried about the proof/correctness implications though, and also a lot of potential edge cases (I believe we rely on the fact that an sc can not be bound to a TCB and a notification in many places). To really work out the rammifications we'd need to work through the proofs. 

The races in initialisation can be dealt with using priorities and user level code (there are examples in sel4 test), so it may be that this PR would need someone to be willing to verify it before it can be considered. I'll leave @lsf37 to comment though. 

Curious: what about passive servers doesn't work for sporadic threads for your usecase? 
Note that the design of MCS is intended to use sporadic servers to contain thread execution, it doesn't work well as a scheduler in itself. I always wanted to add more system calls to allow the kernel to provide proper support for sporadic scheduling with timeouts, however I did have to graduate at some point :).

lsf37:
Interesting. @mktnk3 do the MCS invariants currently say anything about SCs not being bound simultaneously to a notification and to a TCB?

If not, this change could be easy in terms of verification. If the invariants forbid it, though, it looks like it might need significant proof effort.

mktnk3:
> Interesting. @mktnk3 do the MCS invariants currently say anything about SCs not being bound simultaneously to a notification and to a TCB?
> 
> If not, this change could be easy in terms of verification. If the invariants forbid it, though, it looks like it might need significant proof effort.

No, the current proof doesn’t depend on not allowing simultaneous binding, actually. We only need to know that the SC doesn’t already have anything bound for what you want to bind.

lsf37:
> No, the current proof doesn’t depend on not allowing simultaneous binding, actually. We only need to know that the SC doesn’t already have anything bound for what you want to bind.

Awesome, this sounds entirely feasible then. I can have a look at the small spec update his would imply and check if anything else breaks.

@pingerino and @kent-mcleod do you have any other concerns for this one? It looks to me like this would make sense to add.

pingerino:
@lsf37 to quote @kevinelp I don't have the time to "play the mind games" to work out the consequences of this change.

I'd really like to see tests for the different cases where this could have odd effects in scheduling, especially on the notification path, and bound notification path. E.g if the thread is bound to the sc and the notification, and someone else then runs on that notification, what happens? etc etc. What if I revoke one? What if I suspend the thread? Most of these can likely be ruled out but it needs careful consideration. 

Most of the MCS bugs I found were in testing odd scenarios where the scheduling context would go somewhere it shouldn't. The proofs can give us some confidence but we know that proof invariants don't tell us everything.

kent-mcleod:
> and then just start the task as usual, with no racy or failure prone synchronisation between passive task and the launcher task. Without this patch the launcher task is forced to make a call to the to-be passive task which needs to call ReplyRecv(). (As far as I know there is no other robuust way of launching a passive task.)

So if I understand you right, this allows you to configure a set of notification, SC and thread objects as if the SC had been bound to the notification and then donated to the thread without having to first get the thread to block on the notification before removing a setup SC from it?  Then when the thread eventually blocks on the notification object, the SC is returned back to to the notification as normal.  It makes sense that this has no proof impacts as the state is the same when the kernel completes the changed invocation as if the SC had been donated to the thread by the notification object.

> 
> Same for timeout-exception handling: The handler can just assign a temporary SC when the exception occurs, and that SC will be released automatically when it's not needed anymore. Without this the handler and the passive tasks need to know when to return the temporary SC, which is nontrivial.

This assumes that the thread being given the temporary SC is going to block back on the notification object that the SC is already bound on?

lsf37:
> I'd really like to see tests for the different cases where this could have odd effects in scheduling, especially on the notification path, and bound notification path. E.g if the thread is bound to the sc and the notification, and someone else then runs on that notification, what happens? etc etc. What if I revoke one? What if I suspend the thread? Most of these can likely be ruled out but it needs careful consideration.

Those are some of the scenarios I was looking for that we can think through. Looking further through Kent's comment, I think if we can show that we can end up in the same state with the old behaviour already, then it should at least be safe. 


> Most of the MCS bugs I found were in testing odd scenarios where the scheduling context would go somewhere it shouldn't. The proofs can give us some confidence but we know that proof invariants don't tell us everything.

Absolutely, the invariants so far only tell us that the kernel won't crash and that objects fit together etc, but not if we wouldn't get odd behaviour that we don't want.

Indanz:
> I'm worried about the proof/correctness implications though, and also a lot of potential edge cases (I believe we rely on the fact that an sc can not be bound to a TCB and a notification in many places).

When a TCB runs on a notification's SC the TCB state is the same, so I do not except widespread assumption problems.

However, the assumption that a SC is only used by one TCB at a time is probably made and I think the current patch breaks that assumption. I probably need to limit the simultaneous binding of TCB and NF to only when the NF is bound to the same TCB. This would solve all corner cases I believe, but I will check the unbind paths.

> To really work out the rammifications we'd need to work through the proofs.
> The races in initialisation can be dealt with using priorities and user level code (there are examples in sel4 test), so it may be that this PR would need someone to be willing to verify it before it can be considered. I'll leave @lsf37 to comment though.

This seems like a good start to learn verification, I would like to help with this if possible. Is there a practical guide to proof updates yet? I remember someone wanting to write something like that. If it doesn't exist yet I wouldn't mind writing one if I get help with the verification side of things. (This would be a hobby project, so would time some time though, assuming verification is as much work as it seems.)

> Curious: what about passive servers doesn't work for sporadic threads for your usecase?

Use case: Many threads on different cores with equal priorities and periods, but uneven budgets. If a task runs out of budget during a call, we don't want to block all other tasks (it's fine to block all tasks if the server would be running). The tasks have a bursty load, so they can't guarantee they won't overrun their budget. Because of the number of tasks and the small period running out of budget is expected to happen frequently at peak load (during a burst, when performance is critical).

We could assign an emergency SC when this happens, but that would slow down all the cross-core calls. I don't see any robust way to release the SC without making a call to the untrusted server (not trusted to behave correctly at all times). All in all it seems way too much complexity and hassle for a critical part of code, using signals seems simpler and overal more robuust. It also reduces the number of context switches.

Granted, the above is also true for non-periodic threads, but then at least we know for sure one CPU core is busy doing useful work while the passive server is blocked.

> Note that the design of MCS is intended to use sporadic servers to contain thread execution, it doesn't work well as a scheduler in itself. I always wanted to add more system calls to allow the kernel to provide proper support for sporadic scheduling with timeouts, however I did have to graduate at some point :).

Fair enough. :-) We want to use MCS for dividing CPU unequally between same-priority tasks and protecting low-priority task from high priority tasks, and it seems suitable for that.

Indanz:
> So if I understand you right, this allows you to configure a set of notification, SC and thread objects as if the SC had been bound to the notification and then donated to the thread without having to first get the thread to block on the notification before removing a setup SC from it?

Correct. Although now I wonder if I should limit this to notifications bound to the same TCB, or if setting sc->tcb is enough to avoid re-use of the SC by multiple tasks.

> Then when the thread eventually blocks on the notification object, the SC is returned back to to the notification as normal. It makes sense that this has no proof impacts as the state is the same when the kernel completes the changed invocation as if the SC had been donated to the thread by the notification object.

Yes.

> > Same for timeout-exception handling: The handler can just assign a temporary SC when the exception occurs, and that SC will be released automatically when it's not needed anymore. Without this the handler and the passive tasks need to know when to return the temporary SC, which is nontrivial.
> 
> This assumes that the thread being given the temporary SC is going to block back on the notification object that the SC is already bound on?

Yes if calling receiveSignal. For receiveIPC the SC must be bound to the NF which is bound to the TCB. Otherwise the thread continues with the given SC. The behaviour is exactly the same as if the SC was donated by the notification object bound to the SC.

This is the current behaviour, perhaps different behaviour would make more sense, like the SC also being returned when another SC could be donated. That is a change independent of this patch though.

kent-mcleod:
> Correct. Although now I wonder if I should limit this to notifications bound to the same TCB, or if setting sc->tcb is enough to avoid re-use of the SC by multiple tasks.

I think it can be less restrictive than that.  If the notification is bound to a TCB then other TCBs aren't allowed to receive on the notification.  So in this case, you would need to restrict an SC that is bound to a TCB to be also bound to a notification if the notification is bound to a TCB that is different.  But if the notification is not bound to a TCB, then the SC that is bound to the notification could be used by multiple TCBs (just not at the same time) if they all have access to receive on the notification.

kent-mcleod:
> Use case: Many threads on different cores with equal priorities and periods, but uneven budgets. If a task runs out of budget during a call, we don't want to block all other tasks (it's fine to block all tasks if the server would be running). The tasks have a bursty load, so they can't guarantee they won't overrun their budget. Because of the number of tasks and the small period running out of budget is expected to happen frequently at peak load (during a burst, when performance is critical).
> 
> We could assign an emergency SC when this happens, but that would slow down all the cross-core calls. I don't see any robust way to release the SC without making a call to the untrusted server (not trusted to behave correctly at all times). All in all it seems way too much complexity and hassle for a critical part of code, using signals seems simpler and overal more robuust. It also reduces the number of context switches.

Are your many client threads across different cores being served by the a single passive server thread on a single core with the expectation that the server thread migrates across cores to handle each client call?

lsf37:
The errors are only the test that expects to fail when binding a sched context that is already bound to a notification, see https://github.com/seL4/sel4test/blob/5d607ee8b8cc8b8f84cbe4e2db45afcb42355366/apps/sel4test-tests/src/tests/schedcontext.c#L101

For the proofs, @mktnk3 has volunteered to look at what breaks, and has confirmed that changes are minimal (draft at seL4/l4v#348)

kent-mcleod:
Well the kernel currently doesn't seem to prevent having an SC be bound to a TCB and notification at the same if the notification is bound to a different TCB.  The kernel allows the following to happen currently:
1. TCB T1 with SC S blocks on notification N that is currently unbound and inactive.
2. SC S is unbound from TCB T1 and bound to notification N.
3. notification N is signaled causing T1 to wake up with SC S now bound to N and T1.
4. Notification N is bound to TCB T2
5. T1 can no longer wait on N but is still bound to SC S that is still bound to N.
At this point T1 cannot return SC S to N by only sending or receiving on notification/endpoint objects.

So by not preventing you from being able to bind an SC (that's bound to a notification that's bound to thread A) to thread B we aren't adding a regression.

Indanz:
> Are your many client threads across different cores being served by the a single passive server thread on a single core with the expectation that the server thread migrates across cores to handle each client call?

Yes, if using a passive server. And because it's also the ethernet driver it's not easy to create a version per core.

> For the proofs, @mktnk3 has volunteered to look at what breaks, and has confirmed that changes are minimal (draft at seL4/l4v#348)

That looks minimal indeed. Thanks @mktnk3 for doing this!

> At this point T1 cannot return SC S to N by only sending or receiving on notification/endpoint objects.

I think the root cause of this problem is not allowing other threads to wait on bound notifications. That is enforcing random policy by the kernel. It allows a thread to claim exclusive ownership of a notification by just having read access and having access to its TCB cap.

lsf37:
> So by not preventing you from being able to bind an SC (that's bound to a notification that's bound to thread A) to thread B we aren't adding a regression.

Cool that should really mean we're not adding any fundamentally new behaviour. We still need to update sel4test to expect success instead of failure for the new configurations, but otherwise I'd be happy to merge this.

Indanz:
> We still need to update sel4test to expect success instead of failure for the new configurations, but otherwise I'd be happy to merge this.

I'll have a look at adding an extra passive test using this method, in addition to just fixing the bind test.

kent-mcleod:
> I think the root cause of this problem is not allowing other threads to wait on bound notifications. That is enforcing random policy by the kernel. It allows a thread to claim exclusive ownership of a notification by just having read access and having access to its TCB cap.

Do you have any examples of this being an issue in practice?  I made https://sel4.atlassian.net/browse/SELFOUR-2526 a while ago to start collecting any motivating examples to justify changing it.  Notification-TCB binding was pretty hard to get proven originally, and restricting other threads from being able to wait on the bound notification was a simplifying restriction as far as I understand.

Indanz:
> Do you have any examples of this being an issue in practice? I made https://sel4.atlassian.net/browse/SELFOUR-2526 a while ago to start collecting any motivating examples to justify changing it. Notification-TCB binding was pretty hard to get proven originally, and restricting other threads from being able to wait on the bound notification was a simplifying restriction as far as I understand.

No, but conceptually it seems much simpler if Notification-TCB binding is just a way for a TCB to wait for both notifications and end points. It seems like a very unintuitive and random restriction that binding a TCB to a notification object would affect other users of the notification object. The limit of only one TCB being allowed per notification object is also unintuitive (though also not a practical problem). It would all be fine it would simplify the proofs, but looking at the extra complexity of bound notifications in the code, it seems the current implementation makes everything hard.

Indanz:
I updated sel4test, see [sel4test/#50](https://github.com/seL4/sel4test/pull/50)

lsf37:
So, verification is good to go, we have an argument why the change is safe in terms of scheduling (because we can already get to that state with existing ops), and it looks like we want this in the API.

Any further concerns, or is this good to merge?

lsf37 (approved):

Indanz:
I'll update the manual and then it's ready I think.

kent-mcleod (approved):
I'm excited for this change.  I made an issue for eventually updating the MCS tutorial in sel4-tutorials: https://github.com/seL4/sel4-tutorials/issues/66.  We could probably update it once we're comfortable with the change.

Indanz:
Ok, the time & date in my Linux VM was messed up, so the spelling mistakes commit has the wrong date/time. Trying to fix it...

Indanz:
Ok, messed it up, will create a new branch and redo...

Indanz:
I think I fixed everything now, please double check before merging.

lsf37 (approved):

lsf37:
Looks good. It'll be a bit before all the tests pass, there is a lot of load on the CI at them moment (in a good way :-)). From my side this is ready to merge.

manual/parts/threads.tex
```diff
@@ -253,7 +253,10 @@ \subsection{Scheduling Context Donation}
 a notification object with a passive thread blocked waiting on it, the passive thread will receive
 the scheduling context that is bound to the notification object.  The scheduling context is returned
 when the thread blocks on the notification object.  This feature allows for passive servers to use
-notification binding (See \autoref{sec:notification-binding}).
+notification binding (See \autoref{sec:notification-binding}).  If a scheduling context is bound to
+both a notification object and a thread, the behaviour will be the same as for a passive server:
```
axel-h, line 257:
```suggestion
both a notification object and a thread, the behaviour will be the same as for a passive server.
```

manual/parts/threads.tex
```diff
@@ -253,7 +253,10 @@ \subsection{Scheduling Context Donation}
...
+The scheduling context will be unbound from the thread when it blocks on the bound notification object.
+This is useful when launching passive servers or handling timeout exceptions.
```
axel-h, line 259: Could you another sentence here saying why this is useful? Or is this just my lack of detailed understanding here that does not make this obvious?

manual/parts/threads.tex
lsf37: I think the `:` is better. What follows is an explanation for the preceding sentence.

Indanz: To me ':' seemed more appropriate because the following sentence is just a summary of previously described behaviour for passive tasks and not new information, but I'll change to whatever you prefer.

manual/parts/threads.tex
Indanz: How to launch passive servers is not documented anywhere in the manual. The practical difficulty is that the launching thread must know when the to-be passive thread is done initializing and is blocking on IPC so it can safely unbind the initial SC.  If it unbinds too soon, the thread will be stuck forever. So some kind of startup handshake protocol is needed between the launcher task and the just launched passive task.

When binding a SC to both the passive task and a notification object the task will wait for, no such handshake is necessary, as the SC will be unbound from the TCB when the task blocks on the bound notification object. Of course this is more useful for passive servers handling notifications to begin with, but it can also simplify launching other passive servers (at the cost of a notification object).

The same for timeout exceptions: With this feature you can just bind a SC to the TCB when the task runs out of time, and the SC will be automatically "returned" when done (blocking on the notification again). Otherwise you have the whole startup problem again, where the passive task needs to know when to tell the monitor tasks that it's safe to unbind the SC.  But this time you can't implement it simply and synchronously.

My intention was to point the user in the right direction, without going into all the details.

axel-h: But why not put the explanation you wrote above here into the manual to have something at least. We can still move this around at a later point in time.

Indanz: I think the manual could use some appendices with code examples with details like this further explained.

Personally I don't like the tutorials because they're made as exercises and not very useful if you just want information. The solutions do give some additional information, but they are hard to get (at least I haven't found a link on the website for them, you need a local checkout AFAIK).

Should I remove the "This is useful when launching passive servers or handling timeout exceptions." altogether then?

lsf37: Personally, I'm Ok with it in, but it's hard for me to judge what people get confused about when they read it in context. It's more information to have it in. It's potentially less confusion if we remove it.

@axel-h what do you think? I can understand that it sparks the desire for more information, but this spot is not a good place to provide that information. Better not to have it at all then?

Fully agree that code examples in the manual would be good. Then this could be an easy reference to one of those. More work for the future..

axel-h: let's leave it as it is.

lsf37:
Now merged. Thanks for the contribution, this should make things a little bit nicer to use!

@mktnk3 can you merge the corresponding proof PR seL4/l4v#348?

PR #525, opened 2021-08-31
trivial: use #ifdef instead of #if

lsf37:
`#if` fails if the symbol is not defined, at least for the older tool chain currently in docker.


axel-h (approved):

bennoleslie:
Sorry, that was my bad. It was intended to be #ifdef

bennoleslie:
Note: Failing check is only due to hyp not able to be compiled for tqm yet.

bennoleslie (approved):

lsf37:
No worries. I've meanwhile removed hyp as well, so the checks should now succeed. I'll re-run them.

PR #526, opened 2021-08-31
aarch64: make gcc-10 happy

lsf37:
The option -mno-outline-atomics used to be default before gcc-10 and now needs to be provided explicitly. Without it gcc will produce references to `__aarch64_ldadd8_acq_rel` which it expects to exist in libgcc which we are not linking against.


bennoleslie (approved):
LGTM.

PR #528, opened 2021-08-31
Document object api returned error codes

### LINKED ISSUE #499: (minor) Improve seL4_ARM_PageUpperDirectory_Map  documentation
Note: This is very minor.

The description of the operations describe some error conditions, however it is also an error if the vspace has yet to be assigned to an ASID pool. This detail is described in the parameters description, but it would help the user (at least it would have helped me!) if the failure code for this case was noted in the description; other error cases are described there.

canarysnort01:
**Purpose:** Ready to merge.
**Testing performed:** Visual inspection of manual.

- Add ```error``` element on methods to IDL schema
- Generate ```retval``` doxygen tags from ```error``` elements
- Generate error code table in manual from ```retval``` doxygen tags
- Add error code descriptions to IDL files

Fixes #499.

A rendered version of the manual: [manual.pdf](https://github.com/seL4/seL4/files/7085596/manual.pdf)


lsf37:
This is really cool, thanks!

I'll need some time to check through the details, but it's definitely a good idea to have these in the manual.

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -62,6 +62,29 @@
                 in <autoref label="ch:vspace"/>.</docref>
             </description>
             </param>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A page is mapped in <texttt text="vspace"/> at <texttt text="vaddr."/>
```
axel-h, line 67:
```suggestion
                    A page is mapped in <texttt text="vspace"/> at <texttt text="vaddr"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -62,6 +62,29 @@
...
+                    Or, all required page tables are already mapped in <texttt text="vspace"/> at <texttt text="vaddr."/>
```
axel-h, line 68:
```suggestion
                    Or, all required page tables are already mapped in <texttt text="vspace"/> at <texttt text="vaddr"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -101,6 +140,40 @@
                 in <autoref label="ch:vspace"/>.</docref>
             </description>
             </param>
+            <error name="seL4_AlignmentError">
+                <description>
+                    The <texttt text="vaddr"/> is not aligned to the page size of <texttt text="_service."/>
```
axel-h, line 145:
```suggestion
                    The <texttt text="vaddr"/> is not aligned to the page size of <texttt text="_service"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -101,6 +140,40 @@
...
+                </description>
+            </error>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A mapping already exists in <texttt text="vspace"/> at <texttt text="vaddr."/>
```
axel-h, line 150:
```suggestion
                    A mapping already exists in <texttt text="vspace"/> at <texttt text="vaddr"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -101,6 +140,40 @@
...
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    The <texttt text="vspace"/> does not have a paging structure at the required level mapped at <texttt text="vaddr."/>
```
axel-h, line 155:
```suggestion
                    The <texttt text="vspace"/> does not have a paging structure at the required level mapped at <texttt text="vaddr"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -101,6 +140,40 @@
...
+            </error>
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidArgument">
+                <description>
+                    The <texttt text="_service"/> is already mapped in <texttt text="vspace"/> at a different virtual address.
+                    Or, <texttt text="vaddr"/> is in the kernel virtual address range.
```
axel-h, line 166: Is this just the kernel address space or more general, any reserved address space?

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -145,6 +238,41 @@
...
             description="Number of bits of index to resolve to find the destination slot."/>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    The destination slot contains a capability.
+                    Or, there are no more ASID pools available.
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    The <texttt text="index"/> or <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>).</docref>
```
axel-h, line 249:
```suggestion
                    The <texttt text="index"/> or <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>)</docref>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -145,6 +238,41 @@
...
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> or <texttt text="untyped"/> is a CPtr to a capability of the wrong type.
+                    Or, <texttt text="untyped"/> is not the exact size of an ASID pool object.
+                    Or, <texttt text="untyped"/> is a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>).</docref>
```
axel-h, line 262:
```suggestion
                    Or, <texttt text="untyped"/> is a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>)</docref>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -145,6 +238,41 @@
...
+                </description>
+            </error>
+            <error name="seL4_RangeError">
+                <description>
+                    The <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>).</docref>
```
axel-h, line 267:
```suggestion
                    The <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>)</docref>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -162,6 +290,23 @@
                     to an ASID pool.
             </description>
             </param>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    There are no more ASIDs available in <texttt text="_service."/>
```
axel-h, line 295:
```suggestion
                    There are no more ASIDs available in <texttt text="_service"/>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -180,6 +325,39 @@
...
             <param dir="in" name="index" type="seL4_Word" description="CPTR to the destination slot. Resolved from the root of the destination CSpace."/>
             <param dir="in" name="depth" type="seL4_Uint8" description="Number of bits of dest_index to resolve to find the destination slot."/>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    The destination slot contains a capability.
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    The <texttt text="index"/> or <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>).</docref>
```
axel-h, line 335:
```suggestion
                    The <texttt text="index"/> or <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>)</docref>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -180,6 +325,39 @@
...
+            </error>
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_RangeError">
+                <description>
+                    The <texttt text="irq"/> is invalid.
+                    Or, <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>).</docref>
```
axel-h, line 353:
```suggestion
                    Or, <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>)</docref>.
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
canarysnort01: This would look much better in the xml; however, when rendered in the manual, this results in some extra whitespace between ```vaddr``` and the punctuation that looks weird. I'm not sure why that happens.

canarysnort01:
> This is really cool, thanks!
> 
> I'll need some time to check through the details, but it's definitely a good idea to have these in the manual.

Thanks @lsf37. In regards to our discussion in #499, I intended the column header ```Possible Cause``` to express the "may" part of the description. Also, where possible I tried to avoid discussing the exact conditions of the error and focus on the root cause that a programmer would need to fix.

bennoleslie:
Wow, this is amazing @canarysnort01 ! 

@lsf37 do you want to break up the review? I'm happy to review the ARM architecture specific parts if that helps.

lsf37:
> Wow, this is amazing @canarysnort01 !
> 
> @lsf37 do you want to break up the review? I'm happy to review the ARM architecture specific parts if that helps.

That'd be great! I'll take the rest.

bennoleslie:
I've started going through the ARM calls. It will take a little while (sorry).

Some overall questions that I think worth getting out of the way early.

For all the IllegalOperation / InvalidCap with the possible cause: `The <texttt text="_service"/> is a CPtr to a capability of the wrong type.`

I was in two minds about having this repeated for every call. At one level, having that repeated for every call is kind of just noise.

On the other hand, having it explicitly listed means the user doesn't need to jump around. On balance, I think it is best to have it repeated, but wanted to bring it up explicitly in case anyone else had opinions.

I want to check however, is it possible to get an `InvalidCap` error by passing a (valid) capability address but with the wrong capability type?

As far as I can tell from the code path if you pass a valid capability of the wrong type you should always end up with an `IllegalOperation` (assuming we don't have overlapping labels).

Is there an example path where this would result in an `InvalidCap` error instead?

kent-mcleod:
> As far as I can tell from the code path if you pass a valid capability of the wrong type you should always end up with an `IllegalOperation` (assuming we don't have overlapping labels).
> 
> Is there an example path where this would result in an `InvalidCap` error instead?

You can get an InvalidCap error if you try and perform a valid invocation in a 2-phase syscall (under MCS) and your object type is (at least) one of:
cap_thread_cap
cap_domain_cap
cap_cnode_cap
cap_sched_control_cap
cap_sched_context_cap

By 2-phase syscall I mean seL4_ReplyRecv, seL4_NBSendRecv, seL4_NBSendWait, where the first phase performs an invocation and second phase does a Recv or Wait operation.

bennoleslie:
> > As far as I can tell from the code path if you pass a valid capability of the wrong type you should always end up with an `IllegalOperation` (assuming we don't have overlapping labels).
> > Is there an example path where this would result in an `InvalidCap` error instead?
> 
> You can get an InvalidCap error if you try and perform a valid invocation in a 2-phase syscall (under MCS) and your object type is (at least) one of:
> cap_thread_cap
> cap_domain_cap
> cap_cnode_cap
> cap_sched_control_cap
> cap_sched_context_cap
> 
> By 2-phase syscall I mean seL4_ReplyRecv, seL4_NBSendRecv, seL4_NBSendWait, where the first phase performs an invocation and second phase does a Recv or Wait operation.

That's interesting. This gets back to a question I had on some other thread of discussion. Is it valid to invoke a kernel service using anything other than seL4_Call or seL4_Send ?

Or maybe I misunderstanding, are you saying that if you try to use something other the seL4_Call/seL4_Send for a kernel service you'll get `InvalidCap` ?

canarysnort01:
> I was in two minds about having this repeated for every call. At one level, having that repeated for every call is kind of just noise.

I think it is for the scenario where the user is reading the manual straight-through to learn the invocations.

> On the other hand, having it explicitly listed means the user doesn't need to jump around.

I think it is important that the list be as complete as possible for the scenario where the user has received an error code and is trying to figure out what could have caused it, since they're probably going to jump to the specific invocation and not read any introductory text at the beginning of the section.

kent-mcleod:
The seL4_ReplyRecv, seL4_NBSendRecv and seL4_NBSendWait syscalls on MCS perform a "NBSend" on a capability before doing a blocking recv or wait on a different capability.  This operation happens atomically from the perspective of userlevel and other cores.  The NBSend part of the operation is typically performed on a Reply or Notification object but can also be use to do things like retype on an untyped cap (In which case, only the final attempt of retype after any preemption points would be atomic with the final Recv/Wait operation I think).  Some operations are restricted because they could cause state to change that could make the second stage invalid from a scheduling perspective.  The restrictions are pretty conservative at the moment and exclude entire object types.

canarysnort01:
> As far as I can tell from the code path if you pass a valid capability of the wrong type you should always end up with an `IllegalOperation` (assuming we don't have overlapping labels).
> 
> Is there an example path where this would result in an `InvalidCap` error instead?

There are some edge cases where this can happen even on a call, including any operation on a null or zombie cap, or any operation on an endpoint or notification cap that does not have the send right.

https://github.com/seL4/seL4/blob/12.1.0/src/object/objecttype.c#L638

I think it is especially important to document these rare edge cases because they would be the most difficult error codes to figure out.

bennoleslie:
> The seL4_ReplyRecv, seL4_NBSendRecv and seL4_NBSendWait syscalls on MCS perform a "NBSend" on a capability before doing a blocking recv or wait on a different capability. This operation happens atomically from the perspective of userlevel and other cores. The NBSend part of the operation is typically performed on a Reply or Notification object but can also be use to do things like retype on an untyped cap (In which case, only the final attempt of retype after any preemption points would be atomic with the final Recv/Wait operation I think). Some operations are restricted because they could cause state to change that could make the second stage invalid from a scheduling perspective. The restrictions are pretty conservative at the moment and exclude entire object types.

OK, interesting.

I think this brings up an important thing though. It would only be possible for that to happen if you had custom system call bindings. As far as I'm aware the *seL4 API Reference* of the user manual is specifically documenting the `libsel4` interface. `libsel4` only ever use call or send (I think). So, I guess I'm not sure if it is relevant to bring up things that could happen when directly invoking caps in a non-standard manner.

I had wondered in the past if that part of the manual should be documenting libsel4, or directly documenting the invocation interface and specifying things in terms of message register location.

bennoleslie:
> > As far as I can tell from the code path if you pass a valid capability of the wrong type you should always end up with an `IllegalOperation` (assuming we don't have overlapping labels).
> > Is there an example path where this would result in an `InvalidCap` error instead?
> 
> There are some edge cases where this can happen even on a call, including any operation on a null or zombie cap, or any operation on an endpoint or notification cap that does not have the send right.
> 
> https://github.com/seL4/seL4/blob/12.1.0/src/object/objecttype.c#L638
> 
> I think it is especially important to document these rare edge cases because they would be the most difficult error codes to figure out.

You're right. I had missed those, thanks. So, yeah. I agree we should mention this each time. I think the description in 10.1.2 should go into depth on the different ways in which an `Invalid Capability` error can occur so that we do explicitly document these rare edges cases in more details that can fit in the table. (But that should be a different PR).

kent-mcleod:
> I had wondered in the past if that part of the manual should be documenting libsel4, or directly documenting the invocation interface and specifying things in terms of message register location.

I don't know if a deliberate decision was ever made here.  The API section of the manual is generated from the same XML that the C interface is generated from, but there are additional C functions that don't appear in the manual: https://github.com/seL4/seL4/blob/master/libsel4/include/sel4/virtual_client.h#L21

canarysnort01:
> I had wondered in the past if that part of the manual should be documenting libsel4, or directly documenting the invocation interface and specifying things in terms of message register location.

So, I'm a bit biased here because I'm not using the C libsel4, so I'm probably in the small minority.

I would love to see this detailed information in the manual; however, I do like seeing things explained at a higher level. I wonder if a middle ground could be struck, where the invocations are explained in terms of the C API, but there are additional sections explaining how that gets allocated into registers on each platform.

Long-term, I would like to see the scope of the IDL XML files be greatly increased -- for example, to include the register-level ABI for syscalls -- such that libsel4 could be almost completely procedurally generated from them. As a bonus, this would allow much richer information to be automatically generated into the manual.

lsf37:
> I had wondered in the past if that part of the manual should be documenting libsel4, or directly documenting the invocation interface and specifying things in terms of message register location.

It does currently document `libsel4`, at least, as Kent says, the xml is shared between `libsel4` C generation and the manual. We might want to change that in the future or additionally document the ABI, but for this PR I think we should stick with `libseL4`.

Which also answers my question about `seL4_TruncatedMessage` -- it can't really occur via `libsel4`, so we should probably leave it out for now.

lsf37:
> I would love to see this detailed information in the manual; however, I do like seeing things explained at a higher level. I wonder if a middle ground could be struck, where the invocations are explained in terms of the C API, but there are additional sections explaining how that gets allocated into registers on each platform.
>
> Long-term, I would like to see the scope of the IDL XML files be greatly increased -- for example, to include the register-level ABI for syscalls -- such that libsel4 could be almost completely procedurally generated from them. As a bonus, this would allow much richer information to be automatically generated into the manual.

I like that idea, because that would also enable proof generation for `libsel4` at some point.

lsf37 (commented):
RISC-V is reviewed and overall fine, will move on to the generic part now.

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -62,6 +62,29 @@
                 in <autoref label="ch:vspace"/>.</docref>
             </description>
             </param>
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A page is mapped in <texttt text="vspace"/> at <texttt text="vaddr."/>
+                    Or, all required page tables are already mapped in <texttt text="vspace"/> at <texttt text="vaddr."/>
+                </description>
+            </error>
+            <error name="seL4_IllegalOperation">
```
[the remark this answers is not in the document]
lsf37: Resolved (`seL4_TruncatedMessage ` not possible when you're using `libsel4`)

bennoleslie:
> > I had wondered in the past if that part of the manual should be documenting libsel4, or directly documenting the invocation interface and specifying things in terms of message register location.
> 
> So, I'm a bit biased here because I'm not using the C libsel4, so I'm probably in the small minority.

I think I share your bias. I currently am using libsel4, but at the point of considering custom bindings; I don't need most of libsel4 (for a specific thing I'm working on).

> I would love to see this detailed information in the manual; however, I do like seeing things explained at a higher level. I wonder if a middle ground could be struck, where the invocations are explained in terms of the C API, but there are additional sections explaining how that gets allocated into registers on each platform.

I think that would be useful. I think something at the start of the reference section indicating that this describes the libsel4 bindings for the raw invocations, that libsel4 always uses either Call or Send, and that other options are available if you want to build your own bindings would be helpful.

> Long-term, I would like to see the scope of the IDL XML files be greatly increased -- for example, to include the register-level ABI for syscalls -- such that libsel4 could be almost completely procedurally generated from them. As a bonus, this would allow much richer information to be automatically generated into the manual.

Agree 100%. I hadn't got deep into this side of the kernel yet, but was hoping that the XML IDL was sufficient for this purpose.

libsel4/include/interfaces/sel4.xml
```diff
@@ -48,6 +48,35 @@
...
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidArgument">
+                <description>
+                    The <texttt text="size_bits"/> is too big or too small for the requested object type.
+                    Or, <texttt text="type"/> cannot be created from a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>).</docref>
+                </description>
```
lsf37, line 66:
```suggestion
                    Or, <texttt text="type"/> cannot be created from a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>).</docref>
                    Or, the requested object <texttt text="type"/> does not exist.
                </description>
```

libsel4/include/interfaces/sel4.xml
```diff
@@ -48,6 +48,35 @@
...
+            </error>
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidArgument">
+                <description>
+                    The <texttt text="size_bits"/> is too big or too small for the requested object type.
```
lsf37, line 64: This one generates a `seL4_RangeError`. https://github.com/seL4/seL4/blob/5fd9619fb589e672fd968e57ceb8a10f548dc98a/src/object/untyped.c#L80-L86

lsf37: Ah, but it can also generate a `seL4_InvalidArgument`, right. So it should stay here. Sorry for the noise.

libsel4/include/interfaces/sel4.xml
```diff
@@ -48,6 +48,35 @@
...
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_NotEnoughMemory" description="The total size of the new objects exceeds the space available."/>
+            <error name="seL4_RangeError">
+                <description>
+                    The <texttt text="num_objects"/> do not fit in the destination CNode at <texttt text="node_offset."/>
+                    Or, <texttt text="num_objects"/> is greater than <texttt text="CONFIG_RETYPE_FAN_OUT_LIMIT."/>
```
lsf37, line 77: So this bit here, which mostly guards against overflow can also generate a `seL4_RangeError`
https://github.com/seL4/seL4/blob/5fd9619fb589e672fd968e57ceb8a10f548dc98a/src/object/untyped.c#L80-L86

Maybe:
```
Or, <texttt text="size_bits"/> is too large.
```

libsel4/include/interfaces/sel4.xml
```diff
@@ -132,6 +198,25 @@
...
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_RevokeFirst">
+                <description>
+                    The <texttt text="bufferFrame,"/> <texttt text="cspace_root,"/> or <texttt text="vspace_root"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
```
lsf37, line 219: Maybe I missed an indirect path, but I don't think `TCBConfigure` can generate `seL4_RevokeFirst`

libsel4/include/interfaces/sel4.xml
```diff
@@ -152,6 +237,25 @@
...
```
lsf37, line 258: (Same as for non-MCS version of `TCBConfigure`, I don't think it can generate `seL4_RevokeFirst`)

libsel4/include/interfaces/sel4.xml
```diff
@@ -235,6 +416,23 @@
                 description="Location of the thread's IPC buffer. Must be 512-byte aligned. The IPC buffer may not cross a page boundary."/>
             <param dir="in" name="bufferFrame" type="seL4_CPtr"
                 description="Capability to a page containing the thread's IPC buffer."/>
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> or <texttt text="bufferFrame"/> is a CPtr to a capability of the wrong type.
+                    Or, <texttt text="buffer"/> is not aligned.
```
lsf37, line 422: This one leads to a `seL4_AlignmentError` instead

libsel4/include/interfaces/sel4.xml
```diff
@@ -235,6 +416,23 @@
...
+                    Or, <texttt text="bufferFrame"/> is retyped from a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>).</docref>
```
lsf37, line 423:
```suggestion
                    Or, <texttt text="bufferFrame"/> is retyped from a device untyped <docref>(see <autoref label="sec:kernmemalloc"/>).</docref>
                    Or, <texttt text="bufferFrame"/> is not a frame capability.
```

libsel4/include/interfaces/sel4.xml
```diff
@@ -235,6 +416,23 @@
...
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_RevokeFirst">
+                <description>
+                    The <texttt text="bufferFrame"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
```
lsf37, line 435: This probably the same cause as for `TCBConfigure`. If I've missed it there, please ignore this one as well, but if not, then I don't think `SetIPCBuffer` can generate `seL4_RevokeFirst`

libsel4/include/interfaces/sel4.xml
lsf37: I just realised that this is included in "capability of the wrong type" above. Ok to ignore this one.

libsel4/include/interfaces/sel4.xml
lsf37: Urgh, I finally found it, `deriveCap` for untypeds has an `ensureNoChildren` of course. All good.

lsf37 (approved):
Let me save the chunk up to TCBSetSpace to continue later. I found the bit that generates `seL4_RevokeFirst`, so will mark those as resolved. No change necessary.

libsel4/include/interfaces/sel4.xml
lsf37: found it..

libsel4/include/interfaces/sel4.xml
lsf37: found it

lsf37 (approved):
I'm done with everything apart from the file `libsel4/arch_include/arm/interfaces/sel4arch.xml` which I'm leaving to @bennoleslie.

It's an impressive amount of work getting all of these together, and I have no substantial requests, apart from the small bits in the annotations.

It's systematically missing the `seL4_FailedLookup` for the case where the cap has stale mapping info, which may have been intentional, but I think we should add it.

The text I've added is `<cap> is not assigned to an ASID pool`. Another formulation that might give a hint of what is going one and why it is different to the other unassigned case might be to just say `<cap> is no longer assigned to an ASID pool`. That doesn't really explain it, but at least it's different.

lsf37:
So the whitespace issue seems to come from the detour via doxygen which at some point produces `@xmlonly` and`@endxmlonly` tags around stuff, which needs whitespace around it to be valid. That whitespace ends up in LaTeX and Markdown between the `,` or `.` and the preceding word.

I don't have enough doxygen knowledge to say if that could be avoided, but at least there doesn't seem to be any obvious. A relatively easy thing to do would be to post-process with something like `sed` and remove whitespace between words and `,`, and between words and `.` if the `.` is only one `.`.

canarysnort01:
> I'm done with everything apart from the file `libsel4/arch_include/arm/interfaces/sel4arch.xml` which I'm leaving to @bennoleslie.

Thanks @lsf37, I will work on incorporating the changes.

> So the whitespace issue seems to come from the detour via doxygen which at some point produces `@xmlonly` and`@endxmlonly` tags around stuff, which needs whitespace around it to be valid.

Thanks for tracking this down. I will work on a solution.

canarysnort01:
> > Long-term, I would like to see the scope of the IDL XML files be greatly increased -- for example, to include the register-level ABI for syscalls -- such that libsel4 could be almost completely procedurally generated from them. As a bonus, this would allow much richer information to be automatically generated into the manual.
> 
> Agree 100%. I hadn't got deep into this side of the kernel yet, but was hoping that the XML IDL was sufficient for this purpose.

I think you'll find that it's not. There's [an old thread talking about it on the discourse](https://sel4.discourse.group/t/24).

I've been thinking about this and prototyping some stuff for robigalia. You're welcome to join us in our chat room if you'd like to discuss it further, links to the chat room are on the contact page of our website at robigalia dot org.

libsel4/arch_include/x86/interfaces/sel4arch.xml
```diff
@@ -105,6 +205,23 @@
...
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidCapability">
+                <description>
+                    The <texttt text="_service"/> or <texttt text="iospace"/> is a CPtr to a capability of the wrong type.
+                    Or, <texttt text="iospace"/> is not assigned to a PCI device.
+                    Or, <texttt text="_service"/> is already mapped in an IOSpace.
+                </description>
+            </error>
```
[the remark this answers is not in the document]
canarysnort01: I'm confused about how this could happen. Since there is only one IO PageTable object type, wouldn't mapping an IO PageTable always fill in the first open slot that is missing in the hieararchy?

libsel4/arch_include/arm/interfaces/sel4arch.xml
```diff
@@ -31,6 +26,33 @@
...
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A mapping already exists for this level in <texttt text="vspace"/> at <texttt text="vaddr."/>
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    On aarch64, <texttt text="vspace"/> does not have a Page Directory mapped at <texttt text="vaddr."/>
+                </description>
+            </error>
```
canarysnort01, line 38: Missing ASID pool error here as well

libsel4/arch_include/arm/interfaces/sel4arch.xml
```diff
@@ -83,6 +150,39 @@
...
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A mapping already exists in <texttt text="vspace"/> at <texttt text="vaddr."/>
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    The <texttt text="vspace"/> does not have a paging structure at the required level mapped at <texttt text="vaddr."/>
+                </description>
+            </error>
```
canarysnort01, line 167: Missing ASID pool error here as well

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -101,6 +140,40 @@
...
+            <error name="seL4_DeleteFirst">
+                <description>
+                    A mapping already exists in <texttt text="vspace"/> at <texttt text="vaddr."/>
+                </description>
+            </error>
+            <error name="seL4_FailedLookup">
+                <description>
+                    The <texttt text="vspace"/> does not have a paging structure at the required level mapped at <texttt text="vaddr."/>
+                </description>
+            </error>
```
canarysnort01, line 157: Missing ASID pool error here as well

libsel4/arch_include/x86/interfaces/sel4arch.xml
```diff
@@ -495,6 +1034,27 @@
...
+                <description>
+                    The <texttt text="eptpml4"/> is not assigned to an ASID pool.
+                </description>
+            </error>
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidCapability">
```
canarysnort01, line 1052: Missing ASID pool error here as well

libsel4/arch_include/x86/interfaces/sel4arch.xml
```diff
@@ -523,6 +1098,28 @@
...
+                <description>
+                    The <texttt text="_service"/> or <texttt text="eptpml4"/> is a CPtr to a capability of the wrong type.
+                    Or, <texttt text="_service"/> is already mapped in a VSpace.
```
canarysnort01, line 1120: Missing ASID pool error here as well

libsel4/arch_include/x86/interfaces/sel4arch.xml
```diff
@@ -551,6 +1163,28 @@
...
```
canarysnort01, line 1185: Missing ASID pool error here as well

libsel4/include/interfaces/sel4.xml
```diff
@@ -152,6 +237,25 @@
                 description="Location of the thread's IPC buffer. Must be 512-byte aligned. The IPC buffer may not cross a page boundary."/>
             <param dir="in" name="bufferFrame" type="seL4_CPtr"
                 description="Capability to a page containing the thread's IPC buffer."/>
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service,"/> <texttt text="bufferFrame,"/> <texttt text="cspace_root,"/> or <texttt text="vspace_root"/> is a CPtr to a capability of the wrong type.
+                    Or, <texttt text="vspace_root"/> is not assigned to an ASID pool.
+                    Or, <texttt text="cspace_root_data"/> is invalid.
+                    Or, <texttt text="buffer"/> is not aligned.
```
canarysnort01, line 245: This should be seL4_AlignmentError as well

libsel4/arch_include/x86/interfaces/sel4arch.xml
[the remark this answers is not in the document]
lsf37: I need to look at the code again, but what I thought when I wrote this, I think, was that the lookup could hit a page instead of an empty slot. I might well be wrong about that.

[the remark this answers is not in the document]
canarysnort01: I didn't think that would be possible since it only allows 4K pages to be mapped.

Here's the code for reference:
https://github.com/seL4/seL4/blob/12.1.0/src/arch/x86/object/iospace.c#L258

It looks like it's checking for the case where the page tables are full separately.

canarysnort01:
> So the whitespace issue seems to come from the detour via doxygen which at some point produces `@xmlonly` and`@endxmlonly` tags around stuff, which needs whitespace around it to be valid. That whitespace ends up in LaTeX and Markdown between the `,` or `.` and the preceding word.
> 
> I don't have enough doxygen knowledge to say if that could be avoided, but at least there doesn't seem to be any obvious solution. A relatively easy thing to do would be to post-process with something like `sed` and remove whitespace between words and `,`, and between words and `.` if the `.` is only one `.`.
> 
> Edit: or, of course, try to recognise that situation inside `parse_doxygen_xml.py`

I went with trying to recognize the situation inside `parse_doxygen_xml.py`.

I don't think it's a good solution as there's not enough information available at that time to determine if the intent was to have the whitespace there or not.

But, I'm only stripping the whitespace if it exists before a `.` or `,` at the beginning of a text run, so I think the risk of a false positive is low.

lsf37:
> I went with trying to recognize the situation inside parse_doxygen_xml.py.
>
> I don't think it's a good solution as there's not enough information available at that time to determine if the intent was to have the whitespace there or not.
>
> But, I'm only stripping the whitespace if it exists before a . or , at the beginning of a text run, so I think the risk of a false positive is low.

That's fine, I think, certainly much better than my initial idea of using `sed`.

@bennoleslie how are you on time for looking at the Arm part? Do you want me to take that one over?

lsf37:
@canarysnort01 what do you think of the idea of moving swapping the position of error code table and the description section?

Currently, we have

```
text (short)
table (parameters)
text (return value)
table (errors)
text (description, optional)
```

I think it would flow slightly better as 

```
text (short)
table (parameters)
text (return value)
text (description, optional)
table (errors)
```

This would also catch the case of a description expecting to be next to the return value section in the original. I haven't spotted any particularly strong cases of this, but I wouldn't be surprised if there were some.

canarysnort01:
> @canarysnort01 what do you think of the idea of swapping the position of error code table and the description section?

Sounds fine to me. I will add the change.

canarysnort01:
Updated PDF with latest changes: [manual.pdf](https://github.com/seL4/seL4/files/7128351/manual.pdf)

bennoleslie:
@lsf37 I'm reviewing the ARM stuff now. Should have feedback in the next 60 minutes or so.

bennoleslie:
Wow, there sure are a lot of ARM methods! I found it useful to have a spreadsheet in front of me: https://docs.google.com/spreadsheets/d/1cbQo8vV3syFl3A6_sUEX5pVI3jJyhhVCML-c1J9NobI/edit?usp=sharing others might find it beneficial also.

bennoleslie:
@lsf37 @canarysnort01 One thing that appears missing in all cases its the `seL4_TruncatedMessage`. I think this can occur for most, if not all, methods.

The exact times it can occur is (I think) if you incorrectly specify the number of MRs in the message (which only happens if you are making your own system calls or maybe have a libsel4 that is somehow out of date with the kernel).

But, it can also happen if you somehow don't have an IPCBuffer (for cases that have more than 4 arguments).

This is where I do wonder about whether or not it makes sense to document these 'generic' failures for each and every method. Adding just that (and/or figuring out which ones can do it and which can't) is a lot of effort in itself. 

I do worry that the `IllegalOperation` and `InvalidCapability` that can happen on every invocation is actually more harm than good from a documentation point of view. I think in both cases there is always the `_service CPtr is wrong` type of failure, but for some cases `InvalidCapability` and `IllegalOperation` can also have method specific meanings. I feel that we end up obscuring those ones by repeating `Invalid`/`Illegal` every time.

bennoleslie:
Sorry, I'm not going to finish this today. I'm through 2 methods our of 45 after an hour or so. I'll keep working on this tomorrow.

One thing I have found is that `ASIDPoolAssign` can have a 'FailedLookup' error which is not currently documented.

lsf37:
> Sorry, I'm not going to finish this today. I'm through 2 methods our of 45 after an hour or so. I'll keep working on this tomorrow.

No worries

> @lsf37 @canarysnort01 One thing that appears missing in all cases its the seL4_TruncatedMessage. I think this can occur for most, if not all, methods.

I stumbled over this as well (it's in one of the resolved comments I think). Since we're documenting `libseL4` and not the ABI, this error doesn't actually occur, because `libsel4` constructs the message to the kernel. For the ABI, it would potentially occur for almost every call, yes.

> This is where I do wonder about whether or not it makes sense to document these 'generic' failures for each and every method. Adding just that (and/or figuring out which ones can do it and which can't) is a lot of effort in itself.
>
> I do worry that the IllegalOperation and InvalidCapability that can happen on every invocation is actually more harm than good from a documentation point of view. I think in both cases there is always the _service CPtr is wrong type of failure, but for some cases InvalidCapability and IllegalOperation can also have method specific meanings. I feel that we end up obscuring those ones by repeating Invalid/Illegal every time.

I had a similar feeling when I was reading through this, but then convinced myself that from a reference point-of-view you'd want to have the list of things that can go wrong.

Maybe we should have two sections in the error table. One for generic errors and one for errors specific to this call. Ideally, we'd then just have short XML references for the generic one instead of repeating the text. I'd say that is work for a separate PR, though. It is much better to have the current table than no table.

bennoleslie:
Thanks @lsf37 sorry I didn't keep up the all the previous discussion and missed the truncate.

I think one thing missing is that this *can* happen (for some methods) even if using `libsel4`. Specifically if you end up without an IPC buffer you can get this error.

If the idea is that we cover any error that *could* happen, I think we should include it.

But I'd be happy leaving that to another PR. And I like the idea of the 'generic error codes' table + method specific errors codes.
Another thing that I hadn't realised is that 'Invalid Capabilty' error actually returns additional data. Specifically the "invalid capability argument number'. For a 2nd round of this I think it would be good to include that in the documentation. As in "for this method capability argument number X refers to Y and can fail for reasons A, B and C". (But possibly this should also be done in the context of more exactly specifying the interface at a protocol, rather than libsel4 level).

lsf37:
> I think one thing missing is that this _can_ happen (for some methods) even if using `libsel4`. Specifically if you end up without an IPC buffer you can get this error.

Huh, you're right, I missed that one. So we do need to track down which ones need the IPC buffer.
 
> But I'd be happy leaving that to another PR. 

Yes, let's get this one merged, then we can improve on it later.

> And I like the idea of the 'generic error codes' table + method specific errors codes.

👍 

> Another thing that I hadn't realised is that 'Invalid Capabilty' error actually returns additional data. Specifically the "invalid capability argument number'. For a 2nd round of this I think it would be good to include that in the documentation. As in "for this method capability argument number X refers to Y and can fail for reasons A, B and C". (But possibly this should also be done in the context of more exactly specifying the interface at a protocol, rather than libsel4 level).

Agreed.

bennoleslie:
I finished going through all the ARM error codes. Want to just say again big thanks to @canarysnort01 for this work. It's really good!

This things that I think were missed and need updates are:

1: ASIDPool / Asid Pool Assign: FailedLookup can occur, not documented.
2: ARMPage / {Clean Data, Clean and Invalidate Data, Invalidate Data, Unify Instruction}: Failed lookup can occur, not documented.
3: ARMPage / Map:   on aarch64 I can't find a code path that leads to 'DeleteFirst', but I can on aarch32, so I guess it is there somewhere. Probably no change needed to documentation.

I had one other comment, but I don't think it is for documentation and may be a bug:

VCPU / Set TCB: This uses IllegalOperation error code in the case that the TCB cap is not actually a TCB captype. I think every cap type check uses InvalidCapability for this kind of check.

canarysnort01:
> 1: ASIDPool / Asid Pool Assign: FailedLookup can occur, not documented.

I don't understand what would cause this error. Any suggestions for error description?

> 2: ARMPage / {Clean Data, Clean and Invalidate Data, Invalidate Data, Unify Instruction}: Failed lookup can occur, not documented.

How about ```The VSpace of _service is not assigned to an ASID pool```?

> 3: ARMPage / Map: on aarch64 I can't find a code path that leads to 'DeleteFirst', but I can on aarch32, so I guess it is there somewhere. Probably no change needed to documentation.

Nice catch!

I don't see it either. Is the behavior of ARMPageMap different on aarch64 than aarch32?

It looks to me like on aarch64 you can map a page into an address even if there is a different page already mapped at that address, but you can't on aarch32.

I think this should be documented, and possibly changed to be consistent in the future.

bennoleslie:
@canarysnort01 I'm not sure the answer for #1. It seems weird that invalid base and invalid pool are treated differently. In fact, I'm not sure if the lookup fail is possible. I *think* ASIDBase is created valid in MakePool, and there is no way modify after that.

I'm also not 100% sure about cache one either I'm afraid. I think what you wrote is true, I'm just not sure how you can get the system in that state in practise.

And yes, I think we agree on the analysis for point 3. I think we probably need to raise a separate issue for that one. Not a doc problem.

lsf37:
> And yes, I think we agree on the analysis for point 3. I think we probably need to raise a separate issue for that one. Not a doc problem.

Yes this one is probably a bug.

lsf37:
> > 1: ASIDPool / Asid Pool Assign: FailedLookup can occur, not documented.
> I don't understand what would cause this error. Any suggestions for error description?

It's possible that this check exists for proof convenience, I'll put up an issue to see if it can be removed, I don't really see how it can be triggered either. In general, the mapping information in caps can be out of date and therefore needs to be checked -- that should not apply to ASIDPool caps, but we might not have created a special case for that in the proofs.

So in terms of description, we could say `if the ASID pool is no longer assigned` (even though that can likely not happen). Or if we're very confident about this, I'd also be fine with leaving it out.

> > 2: ARMPage / {Clean Data, Clean and Invalidate Data, Invalidate Data, Unify Instruction}: Failed lookup can occur, not documented.
> How about The VSpace of _service is not assigned to an ASID pool?

Yes, I think that would be fine.

lsf37:
Awesome, thanks! I think we should merge this now and add any additional error codes in a separate PR. I don't think there was much relevant code movement in that time, but it would be worth checking the diff to aec0f5a3f00147815e35290819cb2e008a916402 which is the parent of the first commit in this PR.

lsf37:
Had a look through that diff, and the only missing one I could find was a small update for the lazy SchedContext binding that was merged in the meantime. (PR for that now at #617)

lsf37:
And thank you for the impressive amount of work on this one!

PR #531, opened 2021-09-01
vtx: fix EPT cache attribute setting

lsf37:
Credits for this one should go to clang-11, which correctly flags that the big `||` always yielded true and was not doing what was intended.

This means, previously the only possible cache attribute for EPT was `EPTWriteBack`.


lsf37:
The enum `ept_cache_options` also has `EPTWriteProtected`, which is omitted in this list, but I think this was intentional.

kent-mcleod (approved):

PR #533, opened 2021-09-01
cleanup code using BI_PTR(), use BI_FRAME_SIZE_BITS instead of seL4_PageBits

axel-h:
- simplify the boot code by removing the trivial macros BI_PTR() and BI_REF().
- clarify assumptions that boot info is one page

### DIFF
--- a/include/bootinfo.h
+++ b/include/bootinfo.h
@@ -10,8 +10,10 @@
 #include <types.h>
 #include <sel4/bootinfo_types.h>
 
+/* declare object-specific macros to hide the casting */
 #define BI_PTR(r) ((seL4_BootInfo*)(r))
 #define BI_REF(p) ((word_t)(p))
+
 #define S_REG_EMPTY (seL4_SlotRegion){ .start = 0, .end = 0 }
 
 /* The boot info frame takes at least one page, it must be big enough to hold
--- a/include/bootinfo.h
+++ b/include/bootinfo.h
@@ -12,8 +12,12 @@
 
 #define BI_PTR(r) ((seL4_BootInfo*)(r))
 #define BI_REF(p) ((word_t)(p))
-#define BI_FRAME_SIZE_BITS PAGE_BITS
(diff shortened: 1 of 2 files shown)


kent-mcleod:
Macros such as `BI_PTR()` and `BI_REF()` are explicitly called for in the kernel style guide for typecasting: https://docs.sel4.systems/processes/style-guide.html#kernel-code

bennoleslie:
> Macros such as `BI_PTR()` and `BI_REF()` are explicitly called for in the kernel style guide for typecasting: https://docs.sel4.systems/processes/style-guide.html#kernel-code

The guide doesn't provide a rationale for that particular rule. Does anyone else know?

On the face of it this code looks cleaner than before.

Assuming the requirement to keep BI_PTR is needed, I still think this refactor by assigning to a shorter named temp variable is cleaner than what the code was before. (Assuming no proof impact).

axel-h:
Sticking to the coding guideline and leave the macro in. But reduce the usage to simplify the code. My understanding it, that all this boot code is not verified yet but should be one day . So sticking to guidelines coming form verification makes sense, but putting clear comment in the code also helps to identify areas that are not to be "simplified".

bennoleslie (approved):

lsf37 (approved):
I'd be happy with this.

axel-h:
`calculate_rootserver_size()` also uses `BI_FRAME_SIZE_BITS`now instead of `seL4_PageBits`

axel-h:
Seems the whole root server allocation needs some rework if `BI_FRAME_SIZE_BITS` ever gets larger than `seL4_PageBits`. Maybe all this could be moved to some Python code that works out the allocation order and then generate the C code. That avoids a define-orgy to handle this. The only thing that changes dynamically here is `extra_bi_size_bits` then, but there is code that can handle it.
I'd prefer to merge the current state now, otherwise this PR get's even more beyond of the cleanup it intended to do.

PR #535, opened 2021-09-02
Fix circular_includes.py script for GCC 11

alexf91:
The preprocessor for GCC 11 produces a different output than for older versions:

``kernel_all.i`` with GCC 10:
```C
# 1 "/home/alex/tmp/sel4test/build/kernel/kernel_all_copy.c"
# 1 "/home/alex/tmp/sel4test/build//"
# 1 "<built-in>"
# 1 "<command-line>"
...
```

``kernel_all.i`` with GCC 11:
```C
# 0 "/home/alex/tmp/sel4test/build/kernel/kernel_all_copy.c"
# 1 "/home/alex/tmp/sel4test/build//"
# 0 "<built-in>"
# 0 "<command-line>"
...
```

The ignore pattern regex doesn't recognize this, because it assumes the line starts with ``# 1``, resulting in a crash:
```Python
  File "/home/alex/tmp/sel4test/kernel/tools/circular_includes.py", line 74, in <module>
    sys.exit(main(args))
  File "/home/alex/tmp/sel4test/kernel/tools/circular_includes.py", line 61, in main
    while file_stack[-1] != header:
IndexError: list index out of range
```

This patch fixes the regex for ignore patterns to support GCC 11.


tools/circular_includes.py
```diff
@@ -28,7 +28,7 @@ def main(parse_args):
     if ignore_args and len(ignore_args):
         ignore_args = [re.escape(ignore) for ignore in ignore_args]
         ignore_re_string = '(' + '|'.join(ignore_args) + ')'
-        ignore_re = re.compile(r'^# 1 ".*' + ignore_re_string + '"')
+        ignore_re = re.compile(r'^# (0|1) ".*' + ignore_re_string + '"')
```
lsf37, line 31: Slightly more efficient and concise: 
```suggestion
        ignore_re = re.compile(r'^# [01] ".*' + ignore_re_string + '"')
```
(Sorry, I'm a regex nerd :-))

lsf37 (approved):
Thanks for that, @wom-bat mentioned something like this recently, but I didn't understand what he was referring to.

PR #536, opened 2021-09-02
Remove ability to run CMake config files as scripts

bennoleslie:
Currently it is possible to run configs/*verified.cmake files
as a script.

Having the configuration CMake files act as both a source
of configuration (that can be passed using the -C) and as
a script (using -P) is somewhat confusing, especially
to newcomers to the project that are trying to understand
how the build system works.

The PR removes the ability to use the config file as a
a script (leaving it purely as a pre-load script for
populating the cache).


bennoleslie:
Note: This is somewhat speculative. I've had some offline discussions with @kent-mcleod about this being a possibility, but obviously needs wider feedback.

I'm not sure what else this may break (and having it run via CI is one good way to see exactly what breaks).

If the specific use case "have a simple script to build a specific verification configuration" is a common use case that isn't already covered by other scripts it would be possible to make a very short shell script to accommodate that use case.

I suspect the "use cmake config as script to build the kernel" may be used by those doing verification, so they should probably comment.

Note: Even if this is rejected (which is fine) we should ensure all the *verified.cmake files have the same file permissions. Currently some have the 'x' bit set while others do not. (The fact that not all files have the 'x' bit set gives me the impression that perhaps no one is using this feature)

canarysnort01:
This confused me.

bennoleslie:
As far I can tell from searching the *verified.config files are reference in two places:

https://docs.sel4.systems/projects/buildsystem/standalone.html
https://docs.sel4.systems/projects/sel4/verified-configurations.html

The first of these describing standalone builds doesn't mention anything about using the scripts, and indeed the example of a 'typical' configuration doesn't have the script building support.

The verified build page does indeed reference running `cmake -P` explicitly. (So at the very least getting rid of the shebang and the 'x' bit makes sense I guess?). Or that documentation needs to be updated.

lsf37:
I'd be happy with this, I was initially also confused by it and eventually figured it out. We don't use the shebang feature on the verification side, so setting `-x` and removing the shebang won't break anything there.

lsf37:
Oh, but I think we do make use of `cmake -P` part. Let me start a proof run to see if that breaks or not.

kent-mcleod:
> If the specific use case "have a simple script to build a specific verification configuration" is a common use case that isn't already covered by other scripts it would be possible to make a very short shell script to accommodate that use case.

I made this to have a pretty simple and foolproof way of building and varying kernel configuration options to generate binaries and singular C files.  They perform the same build steps as for the "Compile / kernel" jobs for the same configurations and could be used in those plans (Because those plans potentially need to vary other configuration options they don't just invoke the scripts, but invoke CMake with all the options.).  They also should be almost equivalent to the build steps used in the verification build process and generate the same source C file (kernel_all_pp.c) and binary file that would get verified for the input verified configuration.  (Verification tooling doesn't directly invoke them either because it provides additional configuration paths to files located inside the L4V repository that are used as part of the kernel_theories target to generate input proofs to L4V.  

I wouldn't really be happy with removing this functionality without already having some alternative functionality to use.  One of my use cases is copying a config file, varying the configuration and then building both versions to produce diffs of the resulting artifacts.  Then committing both configs and sending the commit to someone else to reproduce the diff and talk about relative verification effort for the new configuration for proving a different platform, or when turning on a feature such as FPU.  Being able to exec the two scripts in different folders makes it easier to do a lot of comparisons without the cognitive load or communication effort of making sure all the trivial config and build target options are correct. 

I don't know if a shell script can replicate this as easily because it'd either have to be able to do config discovery or require the user to know how to modify the script to add new configs.  We have a shell script for easy configuration of projects like seL4test and seL4bench, but once you handle all the different error cases and config discovery it becomes a lot of effort to build and maintain: https://github.com/seL4/seL4_tools/blob/master/cmake-tool/griddle

It's also easy to use these executable scripts to embed a verified kernel build into other projects. 

You can also give configuration arguments to the command line:
`../kernel/configs/ARM_verified.cmake -DKernelPlatform=tk1` builds an ARM_verified config but using the tk1 platform instead of sabre.

bennoleslie:
> > If the specific use case "have a simple script to build a specific verification configuration" is a common use case that isn't already covered by other scripts it would be possible to make a very short shell script to accommodate that use case.
> 
> I made this to have a pretty simple and foolproof way of building and varying kernel configuration options to generate binaries and singular C files. They perform the same build steps as for the "Compile / kernel" jobs for the same configurations and could be used in those plans (Because those plans potentially need to vary other configuration options they don't just invoke the scripts, but invoke CMake with all the options.). They also should be almost equivalent to the build steps used in the verification build process and generate the same source C file (kernel_all_pp.c) and binary file that would get verified for the input verified configuration. (Verification tooling doesn't directly invoke them either because it provides additional configuration paths to files located inside the L4V repository that are used as part of the kernel_theories target to generate input proofs to L4V.
> 
> I wouldn't really be happy with removing this functionality without already having some alternative functionality to use. One of my use cases is copying a config file, varying the configuration and then building both versions to produce diffs of the resulting artifacts. Then committing both configs and sending the commit to someone else to reproduce the diff and talk about relative verification effort for the new configuration for proving a different platform, or when turning on a feature such as FPU. Being able to exec the two scripts in different folders makes it easier to do a lot of comparisons without the cognitive load or communication effort of making sure all the trivial config and build target options are correct.
> 
> I don't know if a shell script can replicate this as easily because it'd either have to be able to do config discovery or require the user to know how to modify the script to add new configs. We have a shell script for easy configuration of projects like seL4test and seL4bench, but once you handle all the different error cases and config discovery it becomes a lot of effort to build and maintain: https://github.com/seL4/seL4_tools/blob/master/cmake-tool/griddle
> 
> It's also easy to use these executable scripts to embed a verified kernel build into other projects.
> 
> You can also give configuration arguments to the command line:
> `../kernel/configs/ARM_verified.cmake -DKernelPlatform=tk1` builds an ARM_verified config but using the tk1 platform instead of sabre.

Seems like your preference would be to add the 'x' bit to all the other cmake files then?

kent-mcleod:
Yes my preference would be adding executable bits.  And also documenting it better to not confuse people somehow.  It doesn't help that the CMake format tool reports executable files to be an error and I think that is the pressure that makes people convert them to be not executable (that tool causes a lot of grief in it's formatting rules too).

PR #537, opened 2021-09-03
Reset SBC on kernel fault for AOS

xurtis:
This ensures the hardware reset is triggered when the kernel faults ensuring that the hardware is rebooted rather than causing the hardware to hang requiring a full power down (which can't currently be performed remotely).

This only needs to be added to the AOS branch of the kernel.

Signed-off-by: Curtis Millar <curtis@curtism.me>

### DIFF
--- a/src/arch/arm/64/idle.c
+++ b/src/arch/arm/64/idle.c
@@ -15,6 +15,8 @@ void idle_thread(void)
     }
 }
 
+void wdog_reset(void);
+
 /** DONT_TRANSLATE */
 void NORETURN NO_INLINE VISIBLE halt(void)
 {
@@ -27,6 +29,7 @@ void NORETURN NO_INLINE VISIBLE halt(void)
     debug_printKernelEntryReason();
 #endif
 #endif
+    wdog_reset();
     idle_thread();
     UNREACHABLE();
 }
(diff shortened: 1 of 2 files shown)


lsf37:
I'm not that familiar with the AOS branch -- it looks like the compile failures are unrelated to this PR, but it also looks like we're expecting it to compile (ARM/ARM_HYP are Arm32 configs, so maybe those are not in scope).

In any case, could you fix up the style bit? Preprocess and links are fine to ignore for the branch.

src/arch/arm/64/idle.c
```diff
@@ -27,6 +29,7 @@ void NORETURN NO_INLINE VISIBLE halt(void)
     debug_printKernelEntryReason();
 #endif
 #endif
+    wdog_reset();
```
axel-h, line 32: If we declare `wdog_reset()` as `weak`, then we could use
```
if (wdog_reset) {
    wdog_reset(); /* this is no expected to return */
}
```
And it would also compile on platforms where this is not implemented

kent-mcleod:
> I'm not that familiar with the AOS branch -- it looks like the compile failures are unrelated to this PR, but it also looks like we're expecting it to compile (ARM/ARM_HYP are Arm32 configs, so maybe those are not in scope).

This should only be required to compile on the AOS platform, which I believe is an odroidc4, and so compilation for other arm platforms shouldn't matter.

src/arch/arm/64/idle.c
lsf37: That would work, but it turns out the AOS branch already doesn't compile for other platforms, so it's probably not worth it.

PR #538, opened 2021-09-03
Add 'x' bit to *verified.cmake

bennoleslie:
The verified.cmake files are designed to be executable scripts.
Ensure all have the 'x' bit.


bennoleslie:
Based on feedback on #536 I'm going with a more conservative fixup which just ensures all scripts have the 'x' bit.

I still think the motivations in #536 hold (and am interested to see if it breaks proofs), but any change there certainly needs more work.

This at least makes things consistent and slightly less confusing.

kent-mcleod (approved):

lsf37 (approved):

PR #545, opened 2021-09-04
Remove unused PYTHON tool definition

bennoleslie:
python2 is not used in the kernel build anymore so the tool
definition is unnecessary.

python2 is also end-of-line so there is no reason to think
we would ever depend on it again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,6 @@ endif()
 include(tools/internal.cmake)
 
 # Define tools used by the kernel
-set(PYTHON "python2" CACHE INTERNAL "")
 set(PYTHON3 "python3" CACHE INTERNAL "")
 RequireTool(CPP_GEN_PATH cpp_gen.sh)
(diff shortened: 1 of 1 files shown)


axel-h (approved):

kent-mcleod:
I think we should do this, but there will probably be breakages in projects that have lifted the PYTHON variable out of the global cache (I think camkes-tool does this)

lsf37:
Yes, that will need a larger test set. When updating the docker containers, something broken when python2 was not available, and I still need to track down what that was.

Otherwise agreed, we should remove python2 and drop support for it.

bennoleslie:
Just rebase to bring this up to date with mainline.

kent-mcleod (approved):
camkes, capdl and sel4-tutorials are all using python3 so I think this is safe to merge.

kent-mcleod:
> camkes, capdl and sel4-tutorials are all using python3 so I think this is safe to merge.

It'll also break in obvious ways for anything coincidentally depending on the variable in a build.

PR #546, opened 2021-09-04
Add additional cmake checks

bennoleslie:
Verify that KernelArch and KernelWordSize have been correctly
set after including the platform cmake files.

### DIFF
--- a/configs/seL4Config.cmake
+++ b/configs/seL4Config.cmake
@@ -166,6 +166,18 @@ foreach(file ${result})
     include("${file}")
 endforeach()
 
+# Verify that, as a minimum any variables that are used
+# to find other build files are actually defined at this
+# point. This means at least: KernelArch KernelWordSize
+
+if("${KernelArch}" STREQUAL "")
+    message(FATAL_ERROR "Variable 'KernelArch' is not set.")
+endif()
+
+if("${KernelWordSize}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


bennoleslie:
This is one thing I added when performing some refactoring on an experimental branch.

I found it quite useful while refactoring things, but perhaps it is complete overkill and unnecessary.

If there is no value in it I will just close this out. I'm not at all attached to this patch!

axel-h (approved):
I have something similar in some test branches locally. If you bring this up now, it seems there is a need for this to aid porting. I'll prepare a PR with my checks then also.

configs/seL4Config.cmake
```diff
@@ -166,6 +166,18 @@ foreach(file ${result})
     include("${file}")
 endforeach()
 
+# Verify that, as a minimum any variables that are used
+# to find other build files are actually defined at this
+# point. This means at least: KernelArch KernelWordSize
+
+if ("${KernelArch}" STREQUAL "")
+    message(FATAL_ERROR "KernelArch must be set.")
```
axel-h, line 174: might be nitpicking, but I'd change this error message to just say what is wrong here.
```suggestion
    message(FATAL_ERROR "Variable 'KernelArch' is not set.")
```

lsf37:
@kent-mcleod, any opinion on this?

kent-mcleod (approved):

PR #550, opened 2021-09-04
Change freemem array as physical regions

bennoleslie:
During boot most architectures reference memory as physical
memory (not kernel virtual pointers). However, in the current
code freemem (and related helper functions) all expect
region_t, which is a kernel virtual memory region.

This patch changes this so that things stay as p_region_t as
much as possible. Overall this leads to much fewer conversions
between region_t/p_region_t and void */paddr.

With the change the only time the generic boot code requires
a kernel virtual address (and not a paddr) is right at the point
of creating the untyped cap.

The only region that starts as virtual that needs converting
to physical is the kernel reusable region of boot memory.
(And even then, that only appears to be the case on ARM. RISC-V
and X64 start with the kernel reusable region as a p_region_t).

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -96,15 +96,15 @@ BOOT_CODE bool_t reserve_region(p_region_t reg)
     return true;
 }
 
-BOOT_CODE static bool_t insert_region(region_t reg)
+BOOT_CODE static bool_t insert_region(p_region_t reg)
 {
     assert(reg.start <= reg.end);
-    if (is_reg_empty(reg)) {
+    if (is_p_reg_empty(reg)) {
         return true;
     }
     for (word_t i = 0; i < ARRAY_SIZE(ndks_boot.freemem); i++) {
-        if (is_reg_empty(ndks_boot.freemem[i])) {
-            reserve_region(pptr_to_paddr_reg(reg));
+        if (is_p_reg_empty(ndks_boot.freemem[i])) {
+            reserve_region(reg);
             ndks_boot.freemem[i] = reg;
             return true;
         }
@@ -562,7 +562,7 @@ BOOT_CODE void init_core_state(tcb_t *scheduler_action)
 BOOT_CODE static bool_t provide_untyped_cap(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
-    pptr_t     pptr,
+    paddr_t    paddr,
     word_t     size_bits,
     seL4_SlotPos first_untyped_slot
 )
@@ -572,10 +572,10 @@ BOOT_CODE static bool_t provide_untyped_cap(
     word_t i = ndks_boot.slot_pos_cur - first_untyped_slot;
     if (i < CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS) {
         ndks_boot.bi_frame->untypedList[i] = (seL4_UntypedDesc) {
-            pptr_to_paddr((void *)pptr), size_bits, device_memory, {0}
+            paddr, size_bits, device_memory, {0}
         };
         ut_cap = cap_untyped_cap_new(MAX_FREE_INDEX(size_bits),
-                                     device_memory, size_bits, pptr);
+                                     device_memory, size_bits, (word_t) paddr_to_pptr(paddr));
         ret = provide_cap(root_cnode_cap, ut_cap);
     } else {
         printf("Kernel init: Too many untyped regions for boot info\n");
@@ -587,14 +587,14 @@ BOOT_CODE static bool_t provide_untyped_cap(
 BOOT_CODE static bool_t create_untypeds_for_region(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
-    region_t   reg,
+    p_region_t   reg,
     seL4_SlotPos first_untyped_slot
 )
 {
     word_t align_bits;
     word_t size_bits;
 
-    while (!is_reg_empty(reg)) {
+    while (!is_p_reg_empty(reg)) {
         /* Determine the maximum size of the region */
         size_bits = seL4_WordBits - 1 - clzl(reg.end - reg.start);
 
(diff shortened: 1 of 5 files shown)


bennoleslie:
I'm not strongly attached to this PR. I don't want to unnecessarily create churn so happy to drop it if required.

In attempting to read / understand / modify this code I found the back and forth between region_t and p_region_t confusing, so this was a win on that front.

src/kernel/boot.c
```diff
@@ -623,16 +623,14 @@ BOOT_CODE static bool_t create_untypeds_for_region(
...
 {
     seL4_SlotPos first_untyped_slot = ndks_boot.slot_pos_cur;
 
     paddr_t start = 0;
     for (word_t i = 0; i < ndks_boot.resv_count; i++) {
         if (start < ndks_boot.reserved[i].start) {
-            region_t reg = paddr_to_pptr_reg((p_region_t) {
-                start, ndks_boot.reserved[i].start
-            });
+            p_region_t reg = (p_region_t) {start, ndks_boot.reserved[i].start};
```
axel-h, line 633: There should be no need for the cast `(p_region_t)` there. Using explicit field initializers is good practice. 
```suggestion
            p_region_t reg = {.start = start, .end = ndks_boot.reserved[i].start};
```

src/kernel/boot.c
```diff
@@ -642,9 +640,7 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap,
     }
 
     if (start < CONFIG_PADDR_USER_DEVICE_TOP) {
-        region_t reg = paddr_to_pptr_reg((p_region_t) {
-            start, CONFIG_PADDR_USER_DEVICE_TOP
-        });
+        p_region_t reg = (p_region_t) {start, CONFIG_PADDR_USER_DEVICE_TOP};
```
axel-h, line 643:
```suggestion
        p_region_t reg = {.start = start, .end = CONFIG_PADDR_USER_DEVICE_TOP};
```

axel-h:
I like this PR, because it does something I also have pending locally. Doing everything in the boot process based on the physical addresses is much more friendly for debugging and error handling/printing. Otherwise, some message are really hard to interpret because the addresses look quite strange and need much more explanation.

lsf37:
From a verification perspective, having everything in physical addresses is not convenient. Instead, everything as far as possibly should be kernel virtual, otherwise things are not comparable to what else goes on in the kernel.

An additional reason is that assertions about objects being there or not there, or memory being mapped/existing all talk about kernel virtual space and don't necessarily make sense on physical addresses. Converting all the time makes something that would be already pretty messy even worse. 

At this point this is hypothetical, because there is no verification yet on boot code, but this is the first thing I would rewrite if we did, beacuse something like `create_untypeds` not working on kernel virtual addresses would definitely be a problem, because that is where you have to make assertions that nothing else in kernel will be affected by this etc, all of which is in kernel virtual space.

I do take the point about physical addresses being easier to debug at this stage in the process. It would be perfectly fine to convert and print the physical addresses, though, that doesn't mean the code has to directly work with them.

axel-h:
The way I see the boot code is, that there is a initial phase where the whole memory region setup happens, and that seems very agnostic of what the kernel does with this memory. Then at some point all this has been sorted out and we are about to start the kernel world and the "kernel window" really comes into the picture. For verification is might also make sense to separate these phases clearly. I don't see much point in trying to reason about a virtual view of memory as long as we are still sorting out what we have physically to work with or even live in.

lsf37:
> The way I see the boot code is, that there is a initial phase where the whole memory region setup happens, and that seems very agnostic of what the kernel does with this memory. Then at some point all this has been sorted out and we are about to start the kernel world and the "kernel window" really comes into the picture. For verification is might also make sense to separate these phases clearly. I don't see much point in trying to reason about a virtual view of memory as long as we are still sorting out what we have physically to work with or even live in.

I'd be happy to have that separated into phases clearly, but when we use `create_untypeds` we are way beyond the point where we should be talking about physical addresses.

Even at the very beginning, though, the first thing you want to say about those regions is that they are distinct from any memory the kernel is currently using -- that is the major invariant you have to keep. What the kernel is currently using is all "measured" in kernel virtual space. So I'm not sure where a good cut point would be.

kent-mcleod:
I think if you are already familiar with dealing with addresses in kernel virtual space from the rest of the kernel code, it's no different dealing with them as kernel virtual addresses in the boot code too.

axel-h:
@kent-mcleod: on 64-bit platforms things are usually quite simple, because the VA space is big enough that virtual addresses are just physical addresses with a lot or "f" before them. On 32-bit platforms with things are more tricky and I found it a lot easier to sort out basic memory issues when it's all shown as physical address. Or when at least each message clearly says if it is referring to a phys or a virt address here. Otherwise this gets really painful sometimes to interpret error messages.

kent-mcleod:
Kernel virtual addresses are also the address domain where kernel object alignment requirements are applied.  Aligning up or down a large untyped object in the physical address range may give you a different result compared to doing it in the kernel virtual address range.

lsf37:
> On 32-bit platforms with things are more tricky and I found it a lot easier to sort out basic memory issues when it's all shown as physical address. Or when at least each message clearly says if it is referring to a phys or a virt address here. Otherwise this gets really painful sometimes to interpret error messages.

I think printing as physical addresses is perfectly fine.

bennoleslie:
> I'd be happy to have that separated into phases clearly, but when we use `create_untypeds` we are way beyond the point where we should be talking about physical addresses.

I guess this is where I disagree. In my understanding `create_untypeds` is actually **the point** at which we stop talking about physical addresses and start talking about kernel virtual addresses. Everything up to that point is physical regions of memory. There is are no pointers, it's all about regions of physical memory (a mix of normal and device memory). 

Just to be clear on what this PR changes: instead of the `freemem` data structures being in terms of kernel address it is in terms of physical addresses.

The change here is instead of having to convert physical regions to kernel address regions in *many* places as they are interested into the `freemem` data structure, instead leave them as physical regions right up until the point they should be kernel addresses and do that change in a single place. I would have thought the single place of change would have made any future potential verification simpler, not harder. But, I am very easily wrong here. I'm more than happy to leave the decision to those with deeper insights than I have.

bennoleslie:
Just to be clear, I'm not really advocating one way or the other for the superiority or not of kernel addresses vs physical addresses.

My main goal is to *reduce the number of conversions back and forward*. Clearly we have both physical and kernel addresses, this is just changing where and how often of conversion between the two happens, and ideally so we avoid having to go paddr -> kaddr -> paddr (or kaddr -> paddr -> kaddr).

To me it's not about the printing, and indeed changing the printing so that it displays paddr when the data structure is kaddr is net negative as it involves yet another conversion between the two compounding what I see as the initial problem.

bennoleslie:
One other point: the `reserved` array which is also part of the `ndks_boot` data structure is already `p_region_t`. Making `freemem` the same seems like it is more consistent.

bennoleslie:
And, I just realised that the `reserved` argument to `init_freemem` is a `region_t` array, while the `available` array is `region_t`. Which means `init_freemem` is also converting back and forth between the spaces (this after most of the things placed in the reserved array actually started life as a `p_region_t`).

bennoleslie:
I don't think there is enough consensus on this one. I'm closing it out for now.

Issue #540, opened 2021-09-03
On error ARM boot code attempts to print prior to serial driver setup


bennoleslie:
From the ARM boot code:

```
    if (it_v_reg.end >= USER_TOP) {
        printf("ERROR: userland image virtual end address too high\n");
        return false;
    }

    /* setup virtual memory for the kernel */
    map_kernel_window();

    /* initialise the CPU */
    if (!init_cpu()) {
        printf("ERROR: CPU init failed\n");
        return false;
    }
```

Serial debug output is not enabled until `init_cpu` returns (successfully).

Attempting to print prior to that results in the kernel itself triggering an exception (which depending on the `loader` you've used to boot the kernel may result in the loader catching the exception and printing out some details).

This may of little consequence as returning false will result in `init_kernel` to attempt to print anyway.

I'm not sure what the preferred approach here is.

I ran in to this as I was attempting to modify the boot code and the presence of these printfs led me to believe that this was a safe place to print.

I'd suggest there is a big comment that says `/* serial drive initialised at this point -- printf allowed */` at the appropriate point, and we should not have any printf calls prior (even in error cases).

I guess another option is that printf (or putc, I don't know the details) checks to see if is initialised and becomes a `nop` if not initialised.

Of course, we could do both.

I'm happy to put together a PR if we can decide on an approach.

kent-mcleod:
The point when printf starts working on arm is after `activate_global_pd();` at the top of `init_cpu()` which is when the device window mapping becomes active.  

Possible third alternative would be that the putchar driver finds the serial device via the direct-mapping left behind by the bootloader.

axel-h:
I have a pending PR that moves the userland image check to a later stage, that would be the cleanes fix there, because we need this message to work . For the CPU init I was wondering if this should not return an error but just halt due to the early boot stage. Adding a flag in putchar() that keeps it muted until enabled was also something I was considering.

kent-mcleod:
Another argument for having printing work from the start of the kernel boot process is that assert failures also try and print and there are several asserts before printing is setup that just cause a fault

kent-mcleod:
Updated issue: https://github.com/seL4/seL4/issues/1191

PR #551, opened 2021-09-04
boot: improve user image location check

axel-h:
Addresses parts of https://github.com/seL4/seL4/issues/540

- On ARM systems, the UART must be set up first before any error messages are visible. Move the user image location check after the platform init that brings up the UART.
- Add the user image location check on RISC-V systems also.
- Make the error message more verbose to clearly show the problem.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -1,5 +1,6 @@
 /*
  * Copyright 2014, General Dynamics C4 Systems
+ * Copyright 2021, HENSOLDT Cyber
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -344,11 +345,6 @@ static BOOT_CODE bool_t try_init_kernel(
     it_v_reg.start = ui_v_reg.start;
     it_v_reg.end = extra_bi_frame_vptr + BIT(extra_bi_size_bits);
 
-    if (it_v_reg.end >= USER_TOP) {
-        printf("ERROR: userland image virtual end address too high\n");
-        return false;
-    }
-
     /* setup virtual memory for the kernel */
     map_kernel_window();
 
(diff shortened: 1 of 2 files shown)


src/arch/arm/kernel/boot.c
```diff
@@ -364,6 +360,12 @@ static BOOT_CODE bool_t try_init_kernel(
     /* initialise the platform */
     init_plat();
 
+    if (it_v_reg.end >= USER_TOP) {
+        printf("ERROR: userland image virt [%p..%p] exceeds USER_TOP (%p)\n",
```
bennoleslie, line 364: What's the rationale for casting to `(void *)` and using `%p`? Would it be clearer to simply use `%lx` ?

axel-h: We can't use `%lx` but it would have to be `SEL4_PRIx_word` then to do it right. At some point I just got a bit lazy to do it everywhere if this is not absolutely necessary - like here in case of virtual addresses and a `prinf()`. If we are dealing with a virtual address that is in the kernel's address space, this is semantically something we can cast to `void*` safely. Also, that saves a bit of typing and make it  a bit nicer to read. Besides `%p` was invented to print pointers and most compilers/analyzer are aware of that that concept, so I'm just using this.

kent-mcleod: It isn't invalid for it_v_reg.start to be 0 in which case %p will format it to "(nil)" compared to 0x%lx which will format it as 0x0.

axel-h: @kent-mcleod: good point, even if our `printf()` just prionts `0` here.
I've changes this to use `SEL4_PRIx_word` now. Seem the definition of the constant `USER_TOP` could get some cleanup across all platforms to either make it `ul` or `void*` or `word_t` everywhere. A lot platforms use `ul`, is this the type to use for integers that are pointers? It's a pity we don't have the generic `unitptr_t` for this.

pingerino: History lesson: For a long time `%p` was not implemented in the kernel's printf, which is why you will see `0x%lx` everywhere.

axel-h:
Since this just moves code around, can we merge this? I can't find a strong reason why this check must happen very early in the boot process. The benefit of doing this later is, that we can print a message - which is quite helpful.

lsf37 (approved):

lsf37:
> Since this just moves code around, can we merge this?

"Moves code" and "just" in the same sentence :-). But this looks fine, none of the intermediate parts access `it_v_reg`.

> I can't find a strong reason why this check must happen very early in the boot process. The benefit of doing this later is, that we can print a message - which is quite helpful.

Happy to merge this from my side. Any objections otherwise?

axel-h:
One last change, I've added an empty line after the if-blokc to improve code readability.

kent-mcleod (approved):

PR #552, opened 2021-09-05
boot: fix selection of root server memory region

axel-h:
- return an error when no suitable region exists
- avoid potential overflow situations
- avoid broken freemem list
- clarify variable
- improve comments


axel-h:
adding another commit on top that 
* determines the allocation order for the root server object dynamically to get rig of the define there and the manual management. 
* adds a few more sanity check to catch inconsistencies and development error 
* prints what is allocated and what is left as free memory

lsf37:
This looks way too complex to me. Boot code is correctness critical for the verified part of the kernel and it needs to be as simple as possible and do as little as possible. All this dynamism is going in the wrong direction.

I think it would be better to to generate data structures offline and in the boot code only check for correctness, not trying to figure out things online. I could also imagine a two-phase boot where there is uncritical dynamic code (although it must at least not crash, so having it in C is a worry), and then a second phase that is critical and checks the output of the first phase. What is definitely is not acceptable is critical code trying to figure out things in big complex loops. (The current boot code is already too complex -- the first step in any verification would be to simplify it).

We probably need a more fundamental discussion about how seL4 boot code should evolve. I think @bennoleslie and @kent-mcleod had some strong interest in the boot code as well, and of course this has been a recurring topic in the past. 

Should we start a discussion on this on discourse or similar?

pingerino:
Agree this is the wrong direction. I thought we also had tools that worked out the boot allocation to do further offline memory allocation for the whole system? The plan there was to eventually have the boot code just parse what was generated offline and load that up.

axel-h:
I've remove the additional commit again, thus just leaves the fixes in this PR.

src/kernel/boot.c
```diff
@@ -838,34 +837,88 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+    *reg_post = (region_t) { /* fill the empty slot */
+        .start = rootserver_mem.end,
+        .end = reg_pre->end
+    };
+    reg_pre->end = rootserver_mem.start; /* shrink the region */
+
+    /* The free memory list has been updated, create the root server objects
+     * from the memory block carved out and now described in 'rootserver_mem'.
+     */
+    create_rootserver_objects(it_v_reg, extra_bi_size_bits);
```
pingerino, line 922: This doesn't look like a 'fix' to me but a rewrite. Will this change the ordering of memory objects? 

@lsf37 I vaguely recall the trusted loader / capdl / camkes relied on python code that emulated how the boot code assigns regions in order to do offline allocation. 

Does this change that?

IMO still wrong direction - we should be fixing minimal bugs, keeping the current layout, and revisiting the boot code as a whole with a verification plan and the ability to do offline allocaiton, as well as supporting other requirements (like swapping out boot code).

lsf37: @corlewis probably knows the current capDL loader assumptions best. @corlewis can you comment on that?

axel-h: This does not change anything in the algorithm itself, it's just more verbose now. It also moves things out of the loop that just tries to find a place.  After the loop it catches the error now if no suitable region was found - which was missing before, but usually worked because the regions are big enough.

corlewis: > I vaguely recall the trusted loader / capdl / camkes relied on python code that emulated how the boot code assigns regions in order to do offline allocation.

That's my understanding as well. See https://github.com/seL4/capdl/commit/ef2fe74242214da796179cab15821b66e7468a07 for what I think is the original commit dealing with this.

axel-h:
I've factored a few change out to separate commits and dropped some changes. For the last commit with the fixes, I can't really find a way to further split this up and reduce the changes. This will catch everything that can go wrong now. Even if some things are quite unlikely, they may still happen during porting or testing when running into odd memory configurations. Also, I think the comments and extra verbosity I added there are helpful to explain what this boot code is doing in detail.

src/kernel/boot.c
kent-mcleod: >  After the loop it catches the error now if no suitable region was found - which was missing before, but usually worked because the regions are big enough.

Adding this check only requires making the loop variable signed right?  It doesn't require moving the `create_rootserver_objects()` outside of the loop.

axel-h: Making the variable unsigned is essential for the loop to terminate. An unsigned values would just roll over and `i >= 0` is always true. 
The update of the free regions and the call to `create_rootserver_objects()` can be left inside the loop also. Then we could return 'true' there instead of doing a `break`. If we do this, we could unconditionally return `false`when the loop terminates and don't have to check explicitly `i < 0` to catch the error.
My intention of moving things out of the loop was keeping it tighter and just use it for searching a region. Then the function has a way to continue on success in case one wants to add more code here and exists early on any error.

src/kernel/boot.c
```diff
@@ -873,21 +881,60 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+         * if we arrive here, there is an empty region in the next higher slot.
+         * Swap the slots and try the next lower region:
+         *
+         *                region too small
+         *                v
+         * Free: [a] [b] [c] [empty] [...]  ->  [a] [b] [empty] [c] [...]
+         *            ^
+         *            region to check next
+         */
+        region_t *next = &ndks_boot.freemem[i + 1];
```
mbrcknl, line 908: This line is my guess as to why the SimplExportAndRefine test is failing. Unfortunately, SimplExportAndRefine is really bad at handling array indices that are sums of variables and constants. This is probably why we previously had the `next` variable.

src/kernel/boot.c
```diff
@@ -873,21 +881,60 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
      * from. If possible, retain any left over memory as an extra p region */
     word_t size = calculate_rootserver_size(it_v_reg, extra_bi_size_bits);
     word_t max = rootserver_max_size_bits(extra_bi_size_bits);
+    pptr_t start;
     for (; i >= 0; i--) {
-        word_t next = i + 1;
```
mbrcknl, line 877: This was probably here to avoid SimplExportAndRefine's flaky handling of array index expressions. I would try reinstating this `next` variable to see if that fixes the SimplExportAndRefine failure.

If that works, could you add a comment, e.g.:
```
// This avoids SimplExportAndRefine's poor handling of array indices that
// are sums of variables and constants.
```

My apologies if it was me that previously added `next` without such a comment!

mbrcknl (commented):
I've not yet understood the content of this PR, but have a possible explanation for the SimplExportAndRefine failure...

src/kernel/boot.c
mbrcknl: > This was probably here to avoid SimplExportAndRefine's flaky handling of array index expressions. I would try reinstating this `next` variable to see if that fixes the SimplExportAndRefine failure.

I just went ahead and pushed a commit that does this. We'll see how it goes...

mbrcknl: Ugh, sorry for all the noise. That's what I get for trying to make a small change while in a meeting, and using GitHub as my compiler.

mbrcknl: That didn't work, so I don't know why SimplExportAndRefine is failing. I'm not sure it's worth looking into it more deeply, so I suggest trying to minimise the code changes in this PR.

Some general comments, now that I've looked more closely at the PR...

Firstly, thanks for finding the issue, and proposing a fix!

Comments are worth adding to this code, but I think they could be more concise. May I recommend explicitly stating (and where reasonable, asserting) loop invariants, e.g.:
```
/* Invariant: both i and (i + 1) are valid indices in ndks_boot.freemem. */
assert(i < ARRAY_SIZE(ndks_boot.freemem) - 1);
/* Invariant; the region at index i is the current candidate.
 * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
 * Invariant: region (i + 1) is undefined, and should not be read.
 * Invariant: regions above (i + 1), if any, are empty or too small to use.
 * Invariant: defined regions (all except i + 1) are ordered,
 *            non-overlapping and unallocated. */
```
Having written down these invariants, it's immediately clear that two of the conditionals in the loop (prior to this PR) are redundant, and it's easier to explain some other things. I'll post what I mean separately...

However, I'm not sure the other changes in this PR actually make things clearer. When all else is equal, concision is preferable to verbosity, and I'm not convinced this PR has a payoff to compensate for the increased verbosity. On the contrary, the PR (slightly) increases the cognitive distance between things that readers need to know are connected. For example, when looking at the code that shuffles the current region up to the unused slot, we now need to remember which slots `reg` and `next` point to. IMO, the array expressions they replace are not so cumbersome that they warrant fresh names. Also, I think the code to perform the root server setup is not so big that it deserves to be placed after the loop, where the reader needs to figure out how to differentiate between success and failure cases.

As a point of comparison, here's an alternative version that I think clarifies how the code works with minimal changes to it: https://github.com/seL4/seL4/pull/587

I'm not sure if the `i + 1` will make it through SimplExportAndRefine, but worth a try...

src/kernel/boot.c
```diff
@@ -873,21 +881,64 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
-        pptr_t start = ROUND_DOWN(ndks_boot.freemem[i].end - size, max);
-        if (start >= ndks_boot.freemem[i].start) {
-            create_rootserver_objects(start, it_v_reg, extra_bi_size_bits);
-            if (i < ARRAY_SIZE(ndks_boot.freemem)) {
-                ndks_boot.freemem[next].end = ndks_boot.freemem[i].end;
-                ndks_boot.freemem[next].start = start + size;
+        // SimplExportAnRefine has poor handling of array indices that are sums
```
axel-h, line 886: I recommend extending this comment to say this is about the proofs. For you putting `SimplExportAnRefine` into that context might be obvious, but for people with just a C view on the code this comment might feel a bit mysterious. Could say `"The proof's rule SimplExportAnRefine has poor ..."` if "rule" is the right word there.

src/kernel/boot.c
```diff
@@ -873,21 +881,64 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+         * if we arrive here, there is an empty region in the next higher slot.
+         * Swap the slots and try the next lower region:
+         *
+         *                region too small
+         *                v
+         * Free: [a] [b] [c] [empty] [...]  ->  [a] [b] [empty] [c] [...]
+         *            ^
+         *            region to check next
+         */
+        region_t *next = &ndks_boot.freemem[next_i];
```
axel-h, line 912: Maybe we can avoid the need for having `next_i` completely by just using:
```
region_t *next = reg + 1;
```

src/kernel/boot.c
mbrcknl: Yeah, good point. Although, it turned out this wasn't the problem, so my commit should be dropped anyway.

src/kernel/boot.c
mbrcknl: Might be best to read my other comments before investing any more effort in my flawed commit. :-)

axel-h:
Closing this PR, we will pick https://github.com/seL4/seL4/pull/587

PR #553, opened 2021-09-05
boot: assert() on RISC-V also if insert_region() fails

axel-h:
Unify ARM and RISC-V behavior for `insert_region()`. Just print a message and return an error, but don't call `assert()`. The caller should decide what to do with the error.

I can't find a good reason why ARM and RISC-V should be handled differently here. This was added with https://github.com/seL4/seL4/commit/4a3685df33470396c3468f9b77360f2c97cae730 to catch error where the calculation of `MAX_NUM_FREEMEM_REG` does not result in a good value. For ARM there indeed a is a calculation, for RISC-V there is just a constant that seem fine also from a practical point of view. But if the caller does not check the result, we will never find any issues on release builds anyway. Unless there is a userland root task that check how much memory is there - and this one would be a pain to rest in debug builds if the `assert()` kicks in.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -102,6 +102,7 @@ BOOT_CODE static bool_t insert_region(region_t reg)
     if (is_reg_empty(reg)) {
         return true;
     }
+
     for (word_t i = 0; i < ARRAY_SIZE(ndks_boot.freemem); i++) {
         if (is_reg_empty(ndks_boot.freemem[i])) {
             reserve_region(pptr_to_paddr_reg(reg));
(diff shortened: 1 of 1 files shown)


bennoleslie (commented):
I don't think this should be changed. This very specific assert saved me 15 minutes ago. As far as I can tell the caller does not check this. I think the assert should stay and the function signature changed to 'void'.

axel-h:
My worry if we keep the assert is, what about the behavior on release builds where `assert()` is gone? Is it acceptable that is passes there or should this be a kernel halt during boot then? The issues is, that this is not as assert from static configuration only, but it depends on dynamic parameters and where the bootloader before the ELF loader was putting things. That may change anytime, so you can't trust you debug builds for testing. Since the size of the release builds also change, it might get into a chase-your-tail thing unless you have a deep insight was happens here. Thus I'd prefer to have the error bubble up and this breaks the boot.

lsf37:
@bennoleslie With the assert removed, you'd still get the error message and the boot halting, no? I.e. it should be as useful for debugging as before, unless I'm missing something.

axel-h:
- brought the assert back and added a comment that this is for debug
- @lsf37: not, when we print a message nothing happens. The problem here is that nobody cares for the return code. I have another PR locally that addresses this, but need to extract this from some larger change set first...

bennoleslie:
> @bennoleslie With the assert removed, you'd still get the error message and the boot halting, no? I.e. it should be as useful for debugging as before, unless I'm missing something.

I don't think there is any code that would halt the boot. If there is I'd be fine with things.

If there is another PR that introduces code that ensures it halts on this case, then let's get that merged prior to this and then the assert can go away.

lsf37:
> * @lsf37: not, when we print a message nothing happens. The problem here is that nobody cares for the return code. 

Got it.

lsf37 (approved):

lsf37:
@bennoleslie happy with this form?

bennoleslie (approved):

axel-h:
extended the comments in the code a bit more to say we are asserting in debug because the caller ignores errors silently.

PR #554, opened 2021-09-05
risc-v: fix signature for Arch_setTLSRegister()

axel-h:
This fixed that issue that if `KernelSetTLSBaseSelf` is enabled on RISC-V, the compilation fails with 
```
...src/api/syscall.c:320:20: error: void value not ignored as it ought to be
  320 |             return Arch_setTLSRegister(tls_base);
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


lsf37 (approved):

lsf37:
LGTM, just double-checking if there needs to be a proof tweak, but that shouldn't be the case.

PR #400, opened 2021-06-19
Set KernelPaddrUserTop correctly on RISC-V platforms

axel-h:
The top phys addr is exclusive, not inclusive. On 64-bit platforms there is a trivial way to fix this as long as the whole address space is never used. For 32-bit platforms the fix will be more complicated, so this is postponed.

See also https://github.com/seL4/seL4/pull/376 where this got fixed on ARM platforms.

PR #556, opened 2021-09-06
rework physical address space limits

axel-h:
This PR is on top of https://github.com/seL4/seL4/pull/572

Remove `KernelPaddrUserTop` and `CONFIG_PADDR_USER_DEVICE_TOP`, use `KernelPhysAddressSpaceBits` and `PHYS_ADDR_SPACE_BITS` instead. This avoids having to deal with large number that potentially touch the
integer limits and create pitfalls, especially on 32-bit systems.

There have been a discussion about this in the past, see
- https://sel4.discourse.group/t/device-region-untypeds/328
- https://github.com/seL4/seL4/issues/347
- https://github.com/seL4/seL4/pull/376
- https://github.com/seL4/seL4/pull/400

Using the bit size should avoid some issues completely and give move flexibility. What is the impact on verification of this change?


src/arch/arm/config.cmake
```diff
@@ -40,7 +40,7 @@ config_set(KernelArmICacheVIPT ARM_ICACHE_VIPT "${KernelArmICacheVIPT}")
 if(KernelSel4ArchAarch32)
     # 64-bit targets may be building in 32-bit mode,
     # so make sure maximum paddr is 32-bit.
-    math(EXPR KernelPaddrUserTop "(1 << 32) - 1")
+    set(KernelPhysAddressSpaceBits 32)
```
lsf37, line 43: The `1 << 32` is undefined behaviour in C when you're in u32. This is also the reason, the kernel uses the `2^n-1` form in general, i.e. talks about the last valid address and compares with `<=`, not the one after and comparing with `<` (not just here, for other comparisons as well). The former is always safe wrt overflow and you don't have to think (or prove anything) about it, whereas `+1` is an operation that can overflow, and then `<` can have unexpected behaviour.

So if we standardise, we should standardise on `(1 << n) - 1`

lsf37 (commented):
From the verification perspective that is the wrong direction (see inline comment)

src/arch/arm/config.cmake
axel-h: We can take care of this in the C code then also and there is a compiler warning anyway if the shift exceeds the limits. But having the address space bit size information available give much more flexibility than having some address and then try deriving the phys address space size in bits from that somehow.

lsf37: Doing that in C is awkward. But It would be perfectly fine to configure the bit size as you have it, and then compute `KernelPaddrUserTop` from it. Then you have both available.

bennoleslie:
I'm for rework in this area, but is changing things to use 'bits' rather than a specific address the right way?

Is it always true that physical address spaces can be described in 2^n ? It seems that isn't true now for at least x64 32-bit. Do we expect it to be true in the future for all architectures? This seems less flexible than what is currently there to me.

axel-h:
For the x86 32-bit I'm looking for help to understand what is the actual limit here and how this is explained. In my current work it seem we can create a reserved region for the end for the last 64k and that should solve the practical issues (and some code can be removed that is currently still on). I'm tempted to say that the basic phys address space size is always 2^n at a certain abstraction and reserved regions (as no-go areas) handle the rest good enough to catch cases where we can't start a system.

bennoleslie:
> For the x86 32-bit I'm looking for help to understand what is the actual limit here and how this is explained. In my current work it seem we can create a reserved region for the end for the last 64k and that should solve the practical issues (and some code can be removed that is currently still on). I'm tempted to say that the basic phys address space size is always 2^n at a certain abstraction and reserved regions (as no-go areas) handle the rest good enough to catch cases where we can't start a system.

OK. Fair enough. I need input from someone with more experience to review. I can't make a call one way or the other on that.

src/arch/arm/config.cmake
axel-h: I have brought `KernelPaddrUserTop` back, but this time in the generic CMake code, so there is only one place where to maintain this

axel-h:
@kent-mcleod: do you know why seL4 declares the last 64 KiB of the phys address space as reserved on x86 32-bit? What is located there?

src/plat/pc99/machine/hardware.c
```diff
@@ -17,16 +17,47 @@
...
-        if (!reserve_region((p_region_t) {
-        (word_t)0xFFFFFFF8, (word_t)0xFFFFFFF8 + 8
-        })) {
+
+#ifdef CONFIG_ARCH_IA32
+
+    /* The last 64 KiB [0xffff0000 - ffffffff] are reserved and can't be used.
+     * ToDo: add some details, why actually?
+     */
+    p_region_t p_reg_reserved = {
```
axel-h, line 26: I wonder, can't reserve this memory in the Python scrips already via a device tree overlay. Or maybe add this in the config and then have the script pick up some reserved regions from there.

src/plat/pc99/machine/hardware.c
```diff
@@ -17,16 +17,47 @@
...
+    }
+
+#else /* not CONFIG_ARCH_IA32 */
+
+    /* remove the 8 byte MSI region [0xfffffff8 - ffffffff] at the end of the
+     * 32-bit address range, as poking at this is undefined and may allow for
+     * the user to generate arbitrary MSI interrupts. Only need to consider this
+     * if it would actually be in the user device region.
+     */
+    p_region_t p_reg_msi = {
```
axel-h, line 45: Same here, would be nice do so this in the Python config generation already

kent-mcleod:
> @kent-mcleod: do you know why seL4 declares the last 64 KiB of the phys address space as reserved on x86 32-bit? What is located there?

https://sel4.atlassian.net/browse/SELFOUR-1179

kent-mcleod:
> Remove `KernelPaddrUserTop` and `CONFIG_PADDR_USER_DEVICE_TOP`, use `KernelPhysAddressSpaceBits` and `PHYS_ADDR_SPACE_BITS` instead. This avoids having to deal with large number that potentially touch the
> integer limits and create pitfalls, especially on 32-bit systems.

After looking into this a little bit in the last few days, I also don't think changing these constants to bits is the right way.  As already mentioned, the highest untyped physical address may not be 2^n as there are other constraints that can limit it.  Additionally, wanting to avoid having to deal with large number that potentially touch the integer limits and create pitfalls isn't possible because once untypeds have been created, we need to handle untypeds being retyped into any combination of smaller objects which requires handling object pointers that could be any address within the range of the untyped.

I think it'd still be worthwhile confirming whether it's necessary to use 2^n-1 instead of 2^n for existing definitions however. But that could be a different issue/pr

axel-h:
Added comments based on explanations from @kent-mcleod why the last page of the physical address space is no accessible.

kent-mcleod:
@axel-h can this be closed? It's very unlikely that the user address space limits will be defined as power's of two.

axel-h:
> @axel-h can this be closed? It's very unlikely that the user address space limits will be defined as power's of two.

If there is no support for bringing this in, then this can be closed, moving it to https://github.com/axel-h/seL4/pull/52. I'd make another PR with the more detailed comments on x86 about the address space exclusions if there is some interest to keep this.

PR #562, opened 2021-09-07
tools: merge adjacent memory regions

lsf37:
This is just the rebased version of #350 with the commit message shortened to appease gitlint.

### DIFF
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -26,6 +26,36 @@ def get_memory_regions(tree: FdtParser):
     return regions
 
 
+def merge_memory_regions(regions: Set[Region]) -> Set[Region]:
+    ''' Check all region and merge adjacent ones '''
+    all_regions = [dict(idx=idx, region=region, right_adj=None, left_adj=None)
+                   for (idx, region) in enumerate(regions)]
+
+    # Find all right contiguous regions
+    for dreg in all_regions:
+        for dnreg in all_regions[dreg['idx']+1:]:
+            if dreg['region'].owner == dnreg['region'].owner:
+                if dnreg['region'].base == dreg['region'].base + dreg['region'].size:
+                    dreg['right_adj'] = dnreg
+                    dnreg['left_adj'] = dreg
+                elif dreg['region'].base == dnreg['region'].base + dnreg['region'].size:
+                    dnreg['right_adj'] = dreg
+                    dreg['left_adj'] = dnreg
+
+    # Find all the left-most contiguous regions
+    contiguous_regions = set()
+    for reg in all_regions:
+        if reg['left_adj'] is None:
+            size = reg['region'].size
+            r_adj = reg['right_adj']
+            while r_adj is not None:
+                size += r_adj['region'].size
+                r_adj = r_adj['right_adj']
+            contiguous_regions.add(Region(reg['region'].base, size, reg['region'].owner))
+
+    return contiguous_regions
+
+
 def parse_reserved_regions(node: WrappedNode) -> Set[Region]:
     ''' Parse a reserved-memory node, looking for regions that are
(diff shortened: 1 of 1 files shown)


tools/hardware/utils/memory.py
```diff
@@ -26,6 +26,38 @@ def visitor(node: WrappedNode):
...
+    contiguous_regions = set()
+    for reg in all_regions:
+        if reg['left_adj'] is None:
+            size = reg['region'].size
+            r_adj = reg['right_adj']
+            while r_adj is not None:
+                size += r_adj['region'].size
+                r_adj = r_adj['right_adj']
+            contiguous_regions.add(Region(reg['region'].base, size, reg['region'].owner))
+    return contiguous_regions
```
axel-h, line 58: please add an empty line here, that improves code readability a bit

tools/hardware/utils/memory.py
```diff
@@ -57,6 +89,7 @@ def reserve_regions(regions: Set[Region], reserved: Set[Region]) -> Set[Region]:
 def get_physical_memory(tree: FdtParser, config: Config) -> List[Region]:
     ''' returns a list of regions representing physical memory as used by the kernel '''
     regions = get_memory_regions(tree)
+    regions = merge_memory_regions(regions)
```
axel-h, line 92: let's merge the lines
```suggestion
    regions = merge_memory_regions(get_memory_regions(tree))
```

tools/hardware/utils/memory.py
lsf37: Done.

tools/hardware/utils/memory.py
lsf37: Done.

lsf37:
I've also turned the loop for `all_regions` into a comprehension to make a bit shorter and more pythonic.

tools/hardware/utils/memory.py
```diff
@@ -26,6 +26,36 @@ def visitor(node: WrappedNode):
     return regions
 
 
+def merge_memory_regions(regions: Set[Region]) -> Set[Region]:
+    ''' Check all region and merge adjacent ones '''
```
axel-h, line 30: This will not handle region fully included in other regions. It's not really supposed to happen but we could add a comment here?

tools/hardware/utils/memory.py
```diff
@@ -26,6 +26,36 @@ def visitor(node: WrappedNode):
...
+    all_regions = [dict(idx=idx, region=region, right_adj=None, left_adj=None)
+                   for (idx, region) in enumerate(regions)]
+
+    # Find all right contiguous regions
+    for dreg in all_regions:
```
axel-h, line 35: I'm trying to understand the loging here and this seem over-complicates? Why not got over the array and build a new array with the merged region. We can modify the elements in the new array without worries.

tools/hardware/utils/memory.py
lsf37: This is not my code, but I find the comment pretty spot on, it explains exactly what the code does. This is strictly for adjacent regions, nothing with overlap or inclusion.

(AFAIK we do expect them to be non-overlapping, this is just trying to avoid trivial fragmentation when two regions are directly next to each other and could have been one region from the start.)

tools/hardware/utils/memory.py
lsf37: That's what it does, no? These are two sets, unordered. The first loop finds what is adjacent, the second builds a new set.

Doing it in one go would be fairly complex, you may have to merge one entry multiple times, and you don't want to iterate over a mutating structure.

lsf37: (Sorry, one set, not two, it just has to pair-wise compare all regions)

axel-h (approved):

PR #570, opened 2021-09-11
trivial: formatting, improve comments about interrupt handling

axel-h:


lsf37:
The contraction with the `cap_t` declaration should not break anything in the proofs, but let me double-check. Otherwise LGTM.

src/object/interrupt.c
```diff
@@ -182,21 +182,22 @@ void deletedIRQHandler(irq_t irq)
...
-         * above the claimed maxIRQ. i.e. we're assuming maxIRQ is wrong */
-        printf("Received IRQ %d, which is above the platforms maxIRQ of %d\n", (int)IRQT_TO_IRQ(irq), (int)maxIRQ);
+        /* The interrupt number is out of range. Pretend it did not happen by
+         * handling it like an inactive interrupt (mask and ack). We assume this
+         * is acceptable, because the platform specific interrupt controller
+         * driver reported this interrupt. Maybe the value maxIRQ is just wrong
+         * or set to a lower value because the interrupts are unused.
+         */
+        printf("Received IRQ %d, which is above the platforms maxIRQ of %d\n",
+               (int)IRQT_TO_IRQ(irq), (int)maxIRQ);
```
lsf37, line 192: It's not really a big problem here (Ok to leave it from my side), but in general there is not reason to wrap this line. The code wrap column for C is [120] and the previous one was correct. If it wasn't the style checker would have complained. In general rewrapping code should be avoided to reduce conflicts with other PRs.

[120]: https://github.com/seL4/seL4_tools/blob/6211394b3f83eed47a2abe80b81af608846cdb66/misc/astylerc#L20

src/object/interrupt.c
```diff
@@ -182,21 +182,22 @@ void deletedIRQHandler(irq_t irq)
...
         return;
     }
+
     switch (intStateIRQTable[IRQT_TO_IDX(irq)]) {
     case IRQSignal: {
-        cap_t cap;
-
-        cap = intStateIRQNode[IRQT_TO_IDX(irq)].cap;
-
+        cap_t cap = intStateIRQNode[IRQT_TO_IDX(irq)].cap;
```
lsf37, line 200: The proofs did unfortunately fail on this one. Probably the reason it was written this way in the first place.

lsf37 (changes requested):
We should revert the `cap` declaration fusion. The change is not really worth spending the hours to figure out what the proof wants there.

src/object/interrupt.c
pingerino: I don't think it was the reason - the kernel used to be compiled with C89 not C99 which lead to this style. 

I recall asking about it with some patches and while inlining the declaration and definition in new code is fine, changing all of the existing sites where this is done was going to be a huge amount of effort.

axel-h: Added a comment about this and created https://github.com/seL4/seL4/issues/576 as a reminder to look into this one day

src/object/interrupt.c
axel-h: reverted

lsf37 (approved):

PR #572, opened 2021-09-11
Cleanup python scripts

axel-h:
Changes factored out from 
* https://github.com/seL4/seL4/pull/556
* https://github.com/seL4/seL4/pull/365

Change are code cleanups and comment improvements. Using the type hints for class member function in the class definition requires Python 3.7 at least with the "import form future", it has been added officially in 3.10 then. Can we assume Python 3.7 here?

### DIFF
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -15,15 +15,6 @@ class Config:
         self.sel4arch = sel4arch
         self.addrspace_max = addrspace_max
 
-    def get_kernel_phys_align(self) -> int:
-        ''' Used to align the base of physical memory. Returns alignment size in bits. '''
-        return 0
-
-    def get_bootloader_reserve(self) -> int:
-        ''' Used to reserve a fixed amount of memory for the bootloader. Offsets
-            the kernel load address by the amount returned in bytes. '''
-        return 0
-
     def get_page_bits(self) -> int:
         ''' Get page size in bits for this arch '''
         return 12  # 4096-byte pages
@@ -35,13 +26,13 @@ class Config:
         ''' Get page size in bits for mapping devices for this arch '''
         return self.get_page_bits()
 
-    def align_memory(self, regions: Set[Region]) -> List[Region]:
-        ''' Given a set of regions, sort them and align the first so that the
-        ELF loader will be able to load the kernel into it. Will return the
-        aligned memory region list, a set of any regions of memory that were
-        aligned out and the physBase value that the kernel will use. memory
-        region list, a set of any regions of memory that were aligned out and
-        the physBase value that the kernel will use. '''
+    def align_memory(self, regions: Set[Region]) -> (List[Region], int):
+        '''
+        Given a set of regions, reserve memory to ensure proper alignment of the
+        first region so that the ELF loader will be able to load the kernel into
+        it. Will return a set of memory regions that are reserved (ie aligned
+        out) and the kernel_phys_base value that the kernel will use (it's
+        called 'phys_base' in the kernel headers). '''
         pass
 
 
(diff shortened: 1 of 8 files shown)


lsf37 (commented):
I like the type annotations, but I'm not 100% sure whether we should assume python 3.7 as minimal build version.

Docker (new Debian stable) has 3.9, GitHub runners (Ubuntu 20.04) have 3.8, so those would be fine. Ubuntu 18 is python 3.6, though, and Debian stretch (previous stable) is on 3.5, both are currently the officially described build platforms. We should probably update that together with the docker images when those are done, but they are not quite done yet.

@wom-bat and @seL4/tsc  do you have an opinion on how far back build host support should reach?

axel-h:
Thanks for the feedback. I've make this a draft again until I've addresses the remarks.

axel-h:
@lsf37 : I've remove all the changes that need more work. Do you thing this change set is fine to merge?

tools/hardware/memory.py
```diff
@@ -82,9 +88,13 @@ def align_base(self, align_bits):
         ''' align this region up to a given number of bits '''
         new_base = utils.align_up(self.base, align_bits)
         diff = new_base - self.base
-        new_size = self.size - diff
-        new = Region(new_base, new_size, self.owner)
-        return new
+        if self.size < diff:
+            raise ValueError(
+                'can''t align region base to {} bits, region {} too small'.format(
+                    align_bits, self))
```
lsf37, line 94: Would suggest using `f".."` strings for these:
```suggestion
                f"can't align region base to {align_bits} bits, {self} too small")
```
(Have left out `region`, because the `__str__` for `Region` will contain the word Region already)

lsf37 (approved):
This looks good from my side. Would slightly prefer the `f".."` string over `"..".format(..)` because it's more idiomatic and shorter.

tools/hardware/memory.py
axel-h: f-String are available in Python 3.6, so we have this version? We could merge this and then convert to f-String in a separate PR.

axel-h:
After syncing this with my other branches, this has endet up in a bigger rewrite of the scripts now. Basically there is just one function `get_physical_memory_regions()`left now, that gives the memory and device regions to the YAML and C-Header script. Seems that is all we need in the end.

axel-h:
It  fail on RISC-V because I'm running into a chicken-egg issues here with something that https://github.com/seL4/seL4/pull/365 fixes. I have to dig a bit deeper why this actually worked on RISC-V before and see if I can keep this working in this PR with a comment explaining why things work - and why https://github.com/seL4/seL4/pull/365 is then a nicer way to handle things, no only on RISC-V

axel-h:
Now it seems to be working everywhere again. Moving the reserved areas to device trees will be part of https://github.com/seL4/seL4/pull/365 then.

tools/hardware/memory.py
lsf37: Hm, I didn't actually check the python version, I just saw that we're using them in other scripts already. So I guess we must have a minimum of python 3.6. We really should set and check minimum versions for the tool chain. Happy to do this separately.

kent-mcleod:
> After syncing this with my other branches, this has endet up in a bigger rewrite of the scripts now. Basically there is just one function `get_physical_memory_regions()`left now, that gives the memory and device regions to the YAML and C-Header script. Seems that is all we need in the end.

Now that this PR has turned into a rewrite of these parts of the python scripts, I am unsure what the motivation is?  The original intention of these scripts were to make it easier to extract arbitrary information from device trees at build time in order to generate static code or configurations that makes working within proof constraints a lot easier (needing/prefering to avoid dynamic runtime querying in certain cases).  Removing a lot of utility functions and replacing them with a single function seems like it'll make it more complicated to use in the future when trying to translate device tree config into other configuration inputs.

axel-h:
I will break this PR up now to allow merging some of the smaller parts. This is all working towards some RISC-V improvements eventually that allow reserving some memory use by OpenSBI in the device tree overlays (https://github.com/seL4/seL4/pull/365)

axel-h:
- I've reworked this now. 
- The commit with the type annotation that requires python 3.7 if left out, I will bring in this separately
- config is part of the hw_yaml now because this is the wrapper object that describes the target we are building.

axel-h:
Is there a chance to get this merged? It streamlines things a bit and is easier to maintain, see the PR on top of this. 
I'm open for a discussion there, especially if python code has some compatibility strings attached that I have missed by removing some explicit functions.

lsf37:
This has turned into a pretty different PR to what it was initially and is also fairly large. Can you elaborate on why you think this new change is necessary and beneficial?

There was not indication that the previous code was problematic. I'm fine with adding type annotations and comments, but I see no reason to rewrite all this.

kent-mcleod:
> There was no indication that the previous code was problematic. I'm fine with adding type annotations and comments, but I see no reason to rewrite all this.

This PR was raised in preparation for https://github.com/seL4/seL4/pull/556 possible, which is now closed. So this PR could potentially be closed or be reduced to just the type annotations and comments for now.

axel-h:
This was also factoring out changes from https://github.com/seL4/seL4/pull/365

PR #573, opened 2021-09-13
hyp: remove check for kernel window conflict

szhuang:
In hyper mode, the VM access resources via stage 2 translation. The VM
should have access to the entire 4GB space. The check for the kernel
window should only apply to stage 1. Removing the check in hyp mode to
enable the VM's access to the extra space.

Signed-off-by: Siwei Zhuang <siwei.zhuang@data61.csiro.au>

### DIFF
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -2214,6 +2214,7 @@ static exception_t decodeARMPageTableInvocation(word_t invLabel, word_t length,
     pd = PDE_PTR(cap_page_directory_cap_get_capPDBasePtr(pdCap));
     asid = cap_page_directory_cap_get_capPDMappedASID(pdCap);
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
     if (unlikely(vaddr >= USER_TOP)) {
         userError("ARMPageTableMap: Virtual address cannot be in kernel window. vaddr: 0x%08lx, USER_TOP: 0x%08x", vaddr,
                   USER_TOP);
@@ -2222,6 +2223,7 @@ static exception_t decodeARMPageTableInvocation(word_t invLabel, word_t length,
 
         return EXCEPTION_SYSCALL_ERROR;
     }
+#endif
 
     {
         findPDForASID_ret_t find_ret;
@@ -2279,7 +2281,7 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
 {
     switch (invLabel) {
     case ARMPageMap: {
-        word_t vaddr, vtop, w_rightsMask;
+        word_t vaddr, w_rightsMask;
         paddr_t capFBasePtr;
         cap_t pdCap;
         pde_t *pd;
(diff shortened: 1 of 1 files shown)


src/arch/arm/32/kernel/vspace.c
```diff
@@ -2214,6 +2214,7 @@ static exception_t decodeARMPageTableInvocation(word_t invLabel, word_t length,
     pd = PDE_PTR(cap_page_directory_cap_get_capPDBasePtr(pdCap));
     asid = cap_page_directory_cap_get_capPDMappedASID(pdCap);
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
     if (unlikely(vaddr >= USER_TOP)) {
```
kent-mcleod, line 2218: USER_TOP prevents mapping entries being created that would overlap with:
1. the kernel window mapping entries, 
2. the kernel device memory mappings
3. the log buffer during bench marking
4. the PD_ASID_SLOT for storing the hardware ASID.

I think item 4. is still needed even in hyp mode and so we still need to prevent mappings from overlapping? 

Another thing to consider is if we need to restrict mapping the very last page of the address space?  I'm not sure what would happen if a syscall was performed on the last address of memory on Arm when the kernel returns after `setNextPC()`

yyshen: PD_ASID_SLOT is still needed. @szhuang Could that error, happened when addressing PT_ASID_SLOT on rk3388, is a result of bypassing this check?

lsf37: That would indeed explain things, well spotted.

It sounds to me like we should make the definition of `USER_TOP` based on `CONFIG_ARM_HYPERVISOR_SUPPORT`, then the check is unchanged, and what exactly it checks can be adjusted. And `USER_TOP` remains a useful concept.

yyshen: Maybe we should give the full 4GiB back to the 32-bit VM? But that means we need to find another place for the ASID_SLOT and verification changes. We just need to make sure that the kernel reserved slots in the stage-2 page directory do not conflict with MMIO addresses that must be used by the VM, so a configurable ASID slot might be enough? Still, I am think about where we could store the ASID.

yyshen: HYP uses long format, and several bits (58-52 and 11-2) in the PDE entry are IGNORED by hardware if the next level is page table. These bits are ignored by hardware, and hardware does not change these bits. They should be enough to contain HW ASID.

lsf37: That sounds possible, actually.

lsf37: Does it also work if there is a page mapped at that level?

yyshen: If the next level is a block, these bits do not exist.  But I do not think 1GiB page size is very useful at the moment.

kent-mcleod: The asid_map_t is an approach that seems to be suitable on aarch64 for solving the same problem. It should also apply here. I made an issue (https://github.com/seL4/seL4/issues/1027) and am going to close this PR for now.

kent-mcleod:
Closing due to discussion in https://github.com/seL4/seL4/pull/573#discussion_r1174563287

PR #577, opened 2021-09-13
boot: change order to improve debug output

axel-h:
Print the available physical memory regions before printing the reserved virtual memory regions, because as this order is more intuitive. For the checks the order does not make any difference. Factor out the checks into separate function to improve code readability.


lsf37 (approved):

lsf37:
Separate small functions instead of one long big function is something verification approves of :-)

There is one style thing to fix up (a space missing somewhere). I don't have access to the branch

axel-h:
Fixed the style issue.

PR #579, opened 2021-09-15
Use asid_map bitfield object on all architectures for ASID table entry

kent-mcleod:
When the ASID table entry only contains a reference to the vspace root object, it's not possible to add any attributes to the relationship.  Some of these attributes, such as a hardware_asid, end up getting stored in slots in the vspace root object itself which can have flow on limitations.  In the future, it is anticipated that more attributes will be needed to implement KernelImage mechanisms, and this is also in preparation for that.

EDIT: This PR is currently in a draft status because verification and performance impacts still need to be assessed.

### DIFF
--- a/include/arch/arm/arch/32/mode/object/structures.bf
+++ b/include/arch/arm/arch/32/mode/object/structures.bf
@@ -21,17 +21,28 @@ base 32
 
 -- 4k frame (these have a separate cap type as there is no room to
 -- store their size)
-block small_frame_cap {
-    field capFMappedASIDLow  10
+block small_frame_cap (
+    capFMappedASIDLow,
+    capFVMRights,
+    capFMappedAddress,
+    capFIsDevice,
+#ifdef CONFIG_ARM_SMMU
+    capFIsIOSpace,
+#endif
+    capFMappedASIDHigh,
+    capFBasePtr,
+    capType
+) {
+    field capFMappedASIDLow  9
     field capFVMRights       2
+    field capFIsDevice       1
     field_high capFMappedAddress 20
 
-    field capFIsDevice       1
 #ifdef CONFIG_ARM_SMMU
     field capFIsIOSpace      1
-    field capFMappedASIDHigh 6
-#else
     field capFMappedASIDHigh 7
+#else
+    field capFMappedASIDHigh 8
 #endif
     field_high capFBasePtr  20
     field capType            4
@@ -40,13 +51,14 @@ block small_frame_cap {
 -- 64k, 1M, 16M frames
 block frame_cap {
     field capFSize           2
-    field capFMappedASIDLow  10
+    field capFMappedASIDLow  9
     field capFVMRights       2
+    padding                  1
     field_high capFMappedAddress 18
 
-    padding                  2
+    padding                  1
     field capFIsDevice       1
-    field capFMappedASIDHigh 7
+    field capFMappedASIDHigh 8
     field_high capFBasePtr  18
     field capType            4
 }
@@ -583,5 +595,24 @@ block dbg_wcr {
     field enabled 1
 }
 #endif /* CONFIG_HARDWARE_DEBUG_API */
+
+block asid_map_none {
+    padding                         32
+
+    padding                         30
+    field type                      2
+}
+
+block asid_map_vspace {
+    field_high vspace_root          32
+
+    padding                         30
+    field type                      2
+}
+
+tagged_union asid_map type {
+    tag asid_map_none 0
+    tag asid_map_vspace 1
+}
 
 #include <sel4/arch/shared_types.bf>
--- a/include/arch/arm/arch/32/mode/object/structures.bf
+++ b/include/arch/arm/arch/32/mode/object/structures.bf
@@ -243,9 +243,7 @@ block stored_hw_asid {
 
 -- Page directory entries
 block pde_invalid {
-    field stored_hw_asid 8
-    field stored_asid_valid 1
-    padding 21
+    padding 30
     field pdeType 2
 }
 
@@ -341,9 +339,7 @@ tagged_union pte pteSize {
 
 block pdeS2_invalid {
     padding 32
-    field stored_hw_asid 8
-    field stored_asid_valid 1
-    padding 21
+    padding 30
     field pdeS2Type 2
 }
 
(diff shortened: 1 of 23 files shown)


src/arch/arm/64/kernel/vspace.c
```diff
@@ -1080,51 +1080,36 @@ pude_t *pageDirectoryMapped(asid_t asid, vptr_t vaddr, pde_t *pd)
...
-    asid_map_t asid_map = asidPool->array[asid & MASK(asidLowBits)];
-    assert(asid_map_get_type(asid_map) == asid_map_asid_map_vspace);
+    asid_map = findMapRefForASID(asid);
+    assert(asid_map_get_type(*asid_map) == asid_map_asid_map_vspace);
 
-    vspace_root_t *vtable = (vspace_root_t *)asid_map_asid_map_vspace_get_vspace_root(asid_map);
-
-    /* Store HW VMID in the last entry
-       Masquerade as an invalid PDGE */
-    vtable[VTABLE_VMID_SLOT] = vtable_invalid_new(hw_asid, true);
```
kent-mcleod, line 1118: seL4_UserTop can be updated as VTABLE_VMID_SLOT is no longer a removed slot.  Additionally, 1 slot at the end of the vspace is used to hold the number of SMMU CBs assigned to the vspace and this should also be added to the asid_map.

src/arch/arm/32/kernel/vspace.c
```diff
@@ -1122,52 +1132,24 @@ pde_t *pageTableMapped(asid_t asid, vptr_t vaddr, pte_t *pt)
...
-    assert(asid_map_get_type(asid_map) == asid_map_asid_map_vspace);
+    asid_map_t *asid_map;
 
-    pd = (pde_t *)asid_map_asid_map_vspace_get_vspace_root(asid_map);
+    asid_map = findMapRefForASID(asid);
+    assert(asid_map_get_type(*asid_map) == asid_map_asid_map_vspace);
 
-    /* Store HW ASID in the last entry
-       Masquerade as an invalid PDE */
-    pd[PD_ASID_SLOT] = pde_pde_invalid_new(hw_asid, true);
```
kent-mcleod, line 1161: This should mean that the MODE_RESERVED region for this region of untyped can be removed.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -190,7 +190,9 @@ block asid_map_none {
 
 block asid_map_vspace {
     field_high vspace_root          48
-    padding                         14
+    padding                         5
+    field hw_asid                   8
```
Indanz, line 194: But ASIDs are 16 bits for aarch64, not 8 bits.

kent-mcleod: I believe in this specific part, these are 8bit VMIDs for stage 2 translation TLB.  Because there are 16 bit ASIDs in aarch64, there isn't a need for a mapping as the seL4 ASID can be directly used for the hardware ASID when the kernel is operating stage 1 page tables.  When the kernel is configured with CONFIG_ARM_HYPERVISOR_SUPPORT it only does stage 2 translation management, but now needs to map 16bit seL4 ASIDs to 8bit VMIDs and needs to track hw_asid assignment.

Indanz: VMIDs are independent of ASIDs, so I have no idea what you mean. If hw_asid is actually an VMID it should be named as such, or if some hackery is going on, well documented that it's actually an VMID for 64-bit ARM with hypervisor support. Are you saying that generic ASID mapping code is only used for VMIDs mapping by aarch64?

kent-mcleod: These fields should be both renamed to VMIDs, and also #ifdef'd by CONFIG_ARM_HYPERVISOR_SUPPORT too.  The naming convention to use hw_asid is inherited from the 32bit hyp extensions.  We should probably look at fixing it for both if we can

yyshen: My reading is the hw_asid can be mapped to different things on different architectures, if I understand the author's intention correctly, without looking at other parts yet. 
On x86-64, it is mapped to PCID, VPID for EPT (not so sure about that part)? for armv7,v8 without HYP, it is mapped to asid, with HYP, it is mapped to vmid. RISCV also uses similar ASID and VMID naming. So the hw_asid might be a reasonable term to represent the underlying hardware mechanism without too many #ifdef in the code. Some early armv8 chips has 8-bit VMID instead of 16-bit.

include/arch/arm/arch/32/mode/object/structures.bf
```diff
@@ -21,17 +21,28 @@ base 32
 
 -- 4k frame (these have a separate cap type as there is no room to
 -- store their size)
-block small_frame_cap {
-    field capFMappedASIDLow  10
+block small_frame_cap (
```
lsf37, line 24: Just putting this here as a reminder: we should separate out introduction of `asid_map` for other platforms from changes in the frame caps and other caps. They should be fully independent.

lsf37:
As todo for later:
 - [ ] remove cap changes from this PR (factor out into separate PR if we want to keep them)
 - [ ] check impact on fastpath

yyshen:
@kent-mcleod This meant to be a more organised way to remove ASID_SLOT from page directories?

lsf37:
> @kent-mcleod This meant to be a more organised way to remove ASID_SLOT from page directories?

It's an old branch from I think @xurtis, who had a similar idea. It needs some cleanup and evaluation, but otherwise could be a solution for removing the ASID_SLOT and making platforms a bit more uniform. This might actually need an RFC since it changes the API, but we can explore first.

kent-mcleod:
> > @kent-mcleod This meant to be a more organised way to remove ASID_SLOT from page directories?
> 
> It's an old branch from I think @xurtis, who had a similar idea. It needs some cleanup and evaluation, but otherwise could be a solution for removing the ASID_SLOT and making platforms a bit more uniform. This might actually need an RFC since it changes the API, but we can explore first.

It also would make it possible to remove the VTABLE_SMMU_SLOT and handle any other cases in the future where other kernel objects need to be linked to a VSpace root object.

kent-mcleod:
Closing this as we'd likely incrementally apply it to any additional architectures on an as-needed basis.

PR #581, opened 2021-09-20
boot: simplify create_untypeds_for_region()

axel-h:
Simplify code and add comments.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -601,27 +601,27 @@ BOOT_CODE static bool_t create_untypeds_for_region(
     seL4_SlotPos first_untyped_slot
 )
 {
-    word_t align_bits;
-    word_t size_bits;
-
     while (!is_reg_empty(reg)) {
-        /* Determine the maximum size of the region */
-        size_bits = seL4_WordBits - 1 - clzl(reg.end - reg.start);
 
-        /* Determine the alignment of the region */
-        if (reg.start != 0) {
-            align_bits = ctzl(reg.start);
-        } else {
-            align_bits = size_bits;
-        }
-        /* Reduce size bits to align if needed */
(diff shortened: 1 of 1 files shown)


src/kernel/boot.c
```diff
@@ -601,27 +601,24 @@ BOOT_CODE static bool_t create_untypeds_for_region(
...
-        } else {
-            align_bits = size_bits;
-        }
-        /* Reduce size bits to align if needed */
-        if (align_bits < size_bits) {
-            size_bits = align_bits;
-        }
+        /* Calculate the bit size of the region. */
+        unsigned int size_bits = seL4_WordBits - 1 - clzl(reg.end - reg.start);
+        /* The bit size can't exceed a certain maximum. */
```
lsf37, line 608:
```suggestion
        /* The size can't exceed the largest possible Untyped size. */
```

src/kernel/boot.c
```diff
@@ -601,27 +601,24 @@ BOOT_CODE static bool_t create_untypeds_for_region(
...
+        /* The start address 0 satisfies any alignment needs, otherwise ensure
+         * the region's bit size does not exceed the alignment of the region.
+         */
+        if (0 != reg.start) {
+            unsigned int align_bits = ctzl(reg.start);
+            if (size_bits > align_bits) {
+                size_bits = align_bits;
+            }
+        }
+        /* Create a region only if it has a at least a certain size. */
```
lsf37, line 621:
```suggestion
        /* Create a region only if it is large enough to be later retyped into objects. */
```

lsf37 (approved):
Would make the comments slightly more specific, otherwise I'd agree that this is slightly clearer.

axel-h:
Thanks for the input, I've change the comments to be more clear.

PR #584, opened 2021-09-20
cleanup includes: unify definition of wordBits, move check from arch to mode headers

axel-h:
- trivial: remove superfluous empty lines
- cleanup includes (remove unused includes, add missing includes)
- move type checks from arch/types.h to mode/types.h
- unify definition of wordBits


include/basic_types.h
```diff
@@ -8,8 +8,15 @@
 
 #include <stdint.h>
 #include <arch/types.h>
+#include <assert.h>
 
-/* arch/types.h is supposed to define word_t and _seL4_word_fmt */
+/* arch/types.h defines wordRadix, wordBits is 2^wordRadix */
+#define wordBits     BIT(wordRadix)
```
axel-h, line 14: This will cause the some CI Preprocess complaints. But instead of using the `BIT(wordRadix)` macro we could also use the old `(1 << wordRadix)` definitions here. Seems to me `BIT()` is more intended for "real" bit masks, but here we just do some plain integer stuff.

kent-mcleod: If the proofs can handle it without an update then it's likely fine.

kent-mcleod: It looks like the proofs can't automatically handle the change between 1 and 1llu here.  Are you able to remove this commit as RISCV proofs use 1llu << wordRadix and the other proofs use 1<< wordRadix, and unifying them likely isn't high enough priority.

include/arch/arm/arch/32/mode/types.h
```diff
@@ -6,6 +6,8 @@
 
 #pragma once
 
+#include <stdint.h>
```
kent-mcleod, line 10: Are you aware that all of the C code in the kernel is compiled as a single C file? All files are concatenated together into kernel_all.c and then compiled in one go.  This means that moving these header files around can change the kernel in unexpected ways.  Something we'd be interested in doing though would be having a kernel build option that compiles each file individually.  Once we have this then it'd be a lot easier to reorganise includes without breaking things on certain configurations

lsf37: > Something we'd be interested in doing though would be having a kernel build option that compiles each file individually. 

If we do this, we need to be careful with the verified configurations. We want the compiler to see the same input that the verification toolchain sees.

kent-mcleod: I should have clarified that this wouldn't be for verified configurations.  (Although, would translation validation be able to show that a kernel.elf assembled from many .o is the same as the original source that had been assembled as a single C source file?

lsf37: Translation validation should be able to deal with that, yes (but we only have this for 2 of the verified platforms)

lsf37:
If the proofs pass, I have no strong feelings either way for this PR. Removing a little bit of duplication makes sense.

I'm unclear on wether the new includes are really missing. If they are missing, did this only compile before because we compile from one single file?

include/basic_types.h
```diff
@@ -8,8 +8,22 @@
...
 
-/* arch/types.h is supposed to define word_t and _seL4_word_fmt */
+/* arch/types.h defines wordRadix, wordBits is 2^wordRadix. */
+#ifdef CONFIG_ARCH_RISCV
+/* Unfortunately, the proofs for RISC-V are build around using BIT() and need to
+ * be updated to unify the definitions.
+ */
+#define wordBits    BIT(wordRadix)
+#else
+#define wordBits    (1 << wordRadix)
```
kent-mcleod, line 20: I don't think this is clearer.  wordBits is an arch or mode specific definition, so it's not unclear to find it in the arch/types.h or arch/mode/types.h header next to wordRadix which is a related closely related definition.

axel-h: My intention here is unifying definition as much as possible - or prepare this at least and ad comment what is still blocking this. Also, I try to provide some guidance, which definition is preferred (seems `(1 << wordRadix)`) in case there is another architecute/mode arising.
I agree that `wordBits` is an arch/mode type, but is seems we build this upon `wordRadix`, which is the only thing an architecture has to provide. Assuming that `wordBits` is of the from `2^n` might be a bit of gambling since there have been other architectures in the past with 7 or 9 or 26 bit words. But I really wonder if they ever come back or sel4 is ever supposed to run there.

kent-mcleod:
@axel-h, what is your opinion about moving some of these long-running draft PRs to be on your own fork of seL4/sel4?  Then if we decide to reconsider them at some point in the future, a new PR can be re-opened on the main repo.  This way it's easier to tell the difference between draft PRs that are getting actively reviewed and draft PRs that are tracking a development branch that isn't currently in the queue for review?

axel-h:
Yes, that would make sense. Now that the github CI action works, all the builds and simulates test work there also nicely. Just the proofs can't run there to see the impact there.

libsel4/include/sel4/macros.h
```diff
@@ -43,6 +43,14 @@
 
 #endif
 
+/* An expression that does nothing, but it is still a token that requires a
+ * semicolon afterwards. Use this for macros that are reduced to nothing in
+ * certain configurations, so the semicolon is properly consumed and does not
+ * create side effects in corner cases. Using "do {} while(0)" here is another
+ * common method, but "((void)(0))" is preferred for verification.
```
lsf37, line 50: Is it? I have the feeling it's the first time I see that expression.

I suspect the proof failures are because of it. I'll have to look at what this translates to in the semantics, but I suspect it'll be easy enough to make a rewrite rule that turns this into SKIP, so it's probably fine to use. I don't think it's substantially different to `do {} while (0)` for verification (which probably also will need its own rule).

lsf37: The other reason the proofs might be failing (the error message in the log doesn't show enough info) is that `wordRadix` isn't used in the expression any more, which means there is less simplification necessary, which means means the proofs steps might change slightly.

I still don't really mind either way. But as it stands it'll need a small proof update. I can look at that if needed.

lsf37:
> Yes, that would make sense. Now that the github CI action works, all the builds and simulates test work there also nicely. Just the proofs can't run there to see the impact there.

If you want to set up your own AWS proof checks I'm happy to help. I can probably make the VM image public so you can use it in a different account, and then it should mostly be a question of making login and password of the test runner account available as secrets in the repo that is being tested.

It'll need a tweak to the action code to detect that it can run outside the seL4 org, but that is entirely doable.

lsf37:
Checking proofs for the current version -- if they fail, I'll close this PR, it's not worth verification effort.

PR #585, opened 2021-09-20
remove goto in printf_core() and return directly

axel-h:
There is no need to keep an option for potentially debugging the error.

### DIFF
--- a/src/machine/io.c
+++ b/src/machine/io.c
@@ -311,12 +311,13 @@ static int printf_core(out_wrap_t *f, const char *fmt, va_list *ap, union arg *n
     int t, pl;
 
     for (;;) {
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

kent-mcleod:
This isn't a valid cleanup in my opinion.  We tried to use a BSD licenced implementation of printf (from libmusl c) so that it would be easy to apply updates from upstream.  Changing the code ourselves makes this harder to manage.

lsf37:
I should probably have waited longer on this one for more discussion, the upstream sync aspect was not on my radar. I do agree with Axel that the `goto`s here are a terrible idiom, though.

PR #586, opened 2021-09-21
Tidy up some untyped creation edge cases in boot code.

kent-mcleod:
- On x86 there was some device UT being created for a region that was intended to be reserved because of an incorrect extra call to `reserve_region()`
- On aarch32 if the user image was loaded into a region outside of the kernel memory window, then the kernel would create device untypeds as well as frame objects for the same memory.
- The last page of memory in the kernel window was being rounded off and not given to user level. It's possible to give smaller UT instead though. (It may even be possible to just allow the last page to be given anyway, but that could be a future change).
- A PPTR_TOP restriction in create_untypeds doesn't make sense as the bitfield code would only error if the address was non-canonical which wouldn't be a valid region and have been excluded before hand already.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -93,13 +93,18 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
             reserved[index] = ui_reg;
             index++;
         }
-    } else if (MODE_RESERVED == 1) {
-        if (index >= ARRAY_SIZE(reserved)) {
-            printf("ERROR: no slot to add the mode-reserved region\n");
-            return false;
+    } else {
+        if (MODE_RESERVED == 1) {
+            if (index >= ARRAY_SIZE(reserved)) {
+                printf("ERROR: no slot to add the mode-reserved region\n");
+                return false;
+            }
+            reserved[index] = mode_reserved_region[0];
+            index++;
         }
-        reserved[index] = mode_reserved_region[0];
-        index++;
+
+        /* Reserve the ui_p_reg region still so it doesn't get turned into device UT. */
+        reserve_region(ui_p_reg);
     }
 
     /* avail_p_regs comes from the auto-generated code */
--- a/src/arch/x86/kernel/boot.c
+++ b/src/arch/x86/kernel/boot.c
@@ -72,7 +72,10 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
                                           mem_p_regs_t *mem_p_regs,
                                           word_t extra_bi_size_bits)
 {
-    ui_p_reg.start = 0;
+    // Extend the reserved region down to include the base of the kernel image.
+    // KERNEL_ELF_PADDR_BASE is the lowest physical load address used
+    // in the x86 linker script.
+    ui_p_reg.start = KERNEL_ELF_PADDR_BASE;
     reserved[0] = paddr_to_pptr_reg(ui_p_reg);
     return init_freemem(mem_p_regs->count, mem_p_regs->list, MAX_RESERVED,
                         reserved, it_v_reg, extra_bi_size_bits);
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -655,13 +655,7 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap,
         region_t reg = paddr_to_pptr_reg((p_region_t) {
             start, CONFIG_PADDR_USER_DEVICE_TOP
         });
-        /*
-         * The auto-generated bitfield code will get upset if the
-         * end pptr is larger than the maximum pointer size for this architecture.
-         */
-        if (reg.end > PPTR_TOP) {
-            reg.end = PPTR_TOP;
-        }
+
         if (!create_untypeds_for_region(root_cnode_cap, true, reg, first_untyped_slot)) {
             return false;
         }
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -97,7 +97,9 @@ def get_physical_memory(tree: FdtParser, config: Config) -> List[Region]:
 def get_addrspace_exclude(regions: List[Region], config: Config):
     ''' Returns a list of regions that represents the inverse of the given region list. '''
     ret = set()
-    as_max = utils.align_down(config.addrspace_max, config.get_page_bits())
+    # We can't create untypeds that exceed the addrspace_max, so we round down to the smallest
+    # untyped size alignment so that the kernel will be able to turn the entire range into untypeds.
+    as_max = utils.align_down(config.addrspace_max, config.get_smallest_kernel_object_alignment())
     ret.add(Region(0, as_max, None))
 
     for reg in regions:
--- a/src/arch/x86/kernel/boot_sys.c
+++ b/src/arch/x86/kernel/boot_sys.c
@@ -213,7 +213,7 @@ static BOOT_CODE bool_t add_mem_p_regs(p_region_t reg)
     printf("Adding physical memory region 0x%lx-0x%lx\n", reg.start, reg.end);
     boot_state.mem_p_regs.list[boot_state.mem_p_regs.count] = reg;
     boot_state.mem_p_regs.count++;
-    return reserve_region(reg);
+    return true;
 }
 
 /*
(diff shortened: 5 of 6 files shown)


src/kernel/boot.c
```diff
@@ -655,13 +655,7 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap,
         region_t reg = paddr_to_pptr_reg((p_region_t) {
             start, CONFIG_PADDR_USER_DEVICE_TOP
         });
-        /*
-         * The auto-generated bitfield code will get upset if the
-         * end pptr is larger than the maximum pointer size for this architecture.
-         */
-        if (reg.end > PPTR_TOP) {
-            reg.end = PPTR_TOP;
-        }
```
axel-h, line 664: Good to see this is gone here. I have been wondering for some time now what this was for.

tools/hardware/utils/memory.py
```diff
@@ -97,7 +97,7 @@ def get_physical_memory(tree: FdtParser, config: Config) -> List[Region]:
 def get_addrspace_exclude(regions: List[Region], config: Config):
     ''' Returns a list of regions that represents the inverse of the given region list. '''
     ret = set()
-    as_max = utils.align_down(config.addrspace_max, config.get_page_bits())
+    as_max = utils.align_down(config.addrspace_max, config.get_smallest_kernel_object_alignment())
```
axel-h, line 100: This should get a comment that we deliberately want to exclude the very last part of the address space here.

src/arch/arm/kernel/boot.c
```diff
@@ -79,9 +79,14 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
...
-        reserved[index] = mode_reserved_region[0];
-        index++;
+    } else {
+        if (MODE_RESERVED == 1) {
+            reserved[index] = mode_reserved_region[0];
+            index++;
+        }
+
+        /* Reserve the ui_p_reg region still so it doesn't get turned into device UT. */
+        reserve_region(ui_p_reg);
```
axel-h, line 89: We are not doing the check for `ui_p_reg.start < PADDR_TOP` in RISC-V, so maybe we should remove it here also? Wouldn't `reserve_region()` then be calles also?

kent-mcleod: This is necessary on aarch32 but not riscv32 as far as I could tell.  (It's not an issue on 64bit because the physical space can easily fit in 64-bit pointer width).

Because aarch32 has platforms with seL4_UserTop high enough that the kernel window is too small to fit all of the platform's available RAM, when the available RAM region is converted to a pptr it's `.end` value can be wrapped around and is less than it's `.start` value.  Because the reserved region_t array given to `init_freemem` must have the regions ordered from lowest to highest in the pptr range, this overflow means a region that is higher in RAM becomes lower in the reserved list.  This made the arch_freemem implementation on Arm more complicated [1] and it ends up being much simpler to reserve directly like how kernel devices are reserved because we know that it can't become regular untypeds already.  This isn't an issue for riscv32 at the moment because there aren't any platforms with Ram outside of the kernel window and so there can't be any wraparound for the user image region. 

[1]
```
diff --git a/src/arch/arm/kernel/boot.c b/src/arch/arm/kernel/boot.c
index 65d30a12..0565193f 100644
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -42,10 +42,17 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
                                           v_region_t it_v_reg,
                                           word_t extra_bi_size_bits)
 {
-    reserved[0].start = KERNEL_ELF_BASE;
-    reserved[0].end = (pptr_t)ki_end;
-
-    int index = 1;
+    int index = 0;
+    region_t ui_reg = paddr_to_pptr_reg(ui_p_reg);
+    int ui_reg_placed = 0;
+    if (KERNEL_ELF_BASE > ui_reg.start) {
+        reserved[index] = ui_reg;
+        index++;
+        ui_reg_placed = 1;
+    }
+    reserved[index].start = KERNEL_ELF_BASE;
+    reserved[index].end = (pptr_t)ki_end;
+    index++;
     if (dtb_p_reg.start) {
         /* the dtb region could be empty */
         reserved[index].start = (pptr_t) paddr_to_pptr(dtb_p_reg.start);
@@ -58,29 +65,32 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
         return false;
     }
 
-    if (ui_p_reg.start < PADDR_TOP) {
-        region_t ui_reg = paddr_to_pptr_reg(ui_p_reg);
-        if (MODE_RESERVED == 1) {
-            if (ui_reg.end > mode_reserved_region[0].start) {
+    if (MODE_RESERVED == 1) {
+        if (!ui_reg_placed) {
+            if (ui_reg.start > mode_reserved_region[0].start) {
                 reserved[index] = mode_reserved_region[0];
                 index++;
                 reserved[index].start = ui_reg.start;
                 reserved[index].end = ui_reg.end;
+                index++;
+
             } else {
                 reserved[index].start = ui_reg.start;
                 reserved[index].end = ui_reg.end;
                 index++;
                 reserved[index] = mode_reserved_region[0];
+                index++;
             }
-            index++;
         } else {
-            reserved[index].start = ui_reg.start;
-            reserved[index].end = ui_reg.end;
+            reserved[index] = mode_reserved_region[0];
             index++;
         }
-    } else if (MODE_RESERVED == 1) {
-        reserved[index] = mode_reserved_region[0];
+    }
+
+    if (!ui_reg_placed) {
+        reserved[index] = ui_reg;
         index++;
+        ui_reg_placed = 1;
     }
 
     return init_freemem(get_num_avail_p_regs(), get_avail_p_regs(), index,
```

kent-mcleod:
> On x86 there was some device UT being created for a region that was intended to be reserved because of an incorrect extra call to `reserve_region()`

Looking at some of the test failures, it looks like [0x0, 0x100000] is expected to be given to userlevel.

kent-mcleod:
Updated to take [0x0, 0x100000] out of the declared `arch_init_freemem()` reserved regions.  Now there should be no ultimate change to what untypeds are given to userlevel on x86.

src/kernel/boot.c
kent-mcleod: Aha, finally found something that this was protecting: `TX2_debug_hyp_gcc_64 FAILED`.  I'll take a look at it tomorrow.

src/arch/x86/kernel/boot.c
```diff
@@ -72,7 +72,10 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
                                           mem_p_regs_t *mem_p_regs,
                                           word_t extra_bi_size_bits)
 {
-    ui_p_reg.start = 0;
+    // Round the reserved region down to the base of the kernel image too
```
axel-h, line 75:
```suggestion
    // Round the reserved region down to the base of the kernel image too.
```

kent-mcleod:
This PR is currently waiting on: https://github.com/seL4/seL4/pull/590

src/kernel/boot.c
kent-mcleod: Now fixed with https://github.com/seL4/seL4/pull/590

src/arch/arm/kernel/boot.c
axel-h: But for the sake of unifying the general behavior, should this be done in the generic RISC-V also? Just because there is no platform now it a poor argument, it could happen when testing/exploring something new and then we might see random crashes that are a pain to investigate.

kent-mcleod: > But for the sake of unifying the general behavior, should this be done in the generic RISC-V also? Just because there is no platform now it a poor argument, it could happen when testing/exploring something new and then we might see random crashes that are a pain to investigate.

This change shouldn't also be done for the RISC-V arch implementation of  arch_init_freemem.  The implementations are already arch specific.  RISC-V unconditionally adds the `ui_reg` region to the reserved region list unconditionally and so doesn't have the same issue on Arm where additional UT would be created.  

Whether RISC-V should be unconditionally adding the `ui_reg` to the reserved region isn't a concern of this PR. If some platform was added on RISC-V 32bit in the future that loaded the `ui_reg` above the kernel window, it would cause the boot code to print an error and abort during boot in the consistency checks that init_freemem performs.

tools/hardware/utils/memory.py
kent-mcleod: We aren't excluding the last part of the address space, but making sure that we don't create objects that exceed the max value of the address space.  On aarch64 and riscv64 the align_down doesn't actually exclude anything since the config.addrspace_max became power-of-2 aligned.

axel-h (approved):

PR #587, opened 2021-09-21
boot: handle failure to find root server memory

mbrcknl:
Previously, the case where there was no memory region large enough for
the root server objects was unhandled. The loop termination condition
(i>=0) was trivially true for all values of the loop variable. This
commit fixes this by making the loop variable signed, and adding a
failure case.

It also simplifies away some conditionals that were guaranteed to be
true, and adds some clarifying comments.

Co-authored-by: Axel Heider <axelheider@gmx.de>
Signed-off-by: Matthew Brecknell <matthew@brecknell.net>

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -861,7 +861,7 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
     }
 
     /* now try to fit the root server objects into a region */
-    word_t i = ARRAY_SIZE(ndks_boot.freemem) - 1;
+    int i = ARRAY_SIZE(ndks_boot.freemem) - 1;
     if (!is_reg_empty(ndks_boot.freemem[i])) {
         printf("Insufficient MAX_NUM_FREEMEM_REG\n");
         return false;
@@ -874,20 +874,50 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
     word_t size = calculate_rootserver_size(it_v_reg, extra_bi_size_bits);
     word_t max = rootserver_max_size_bits(extra_bi_size_bits);
     for (; i >= 0; i--) {
-        word_t next = i + 1;
-        pptr_t start = ROUND_DOWN(ndks_boot.freemem[i].end - size, max);
-        if (start >= ndks_boot.freemem[i].start) {
+        /* Invariant: both i and (i + 1) are valid indices in ndks_boot.freemem. */
+        assert(i < ARRAY_SIZE(ndks_boot.freemem) - 1);
+        /* Invariant; the region at index i is the current candidate.
+         * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
+         * Invariant: region (i + 1) is empty. */
+        assert(is_reg_empty(ndks_boot.freemem[i + 1]));
+        /* Invariant: regions above (i + 1), if any, are empty or too small to use.
+         * Invariant: all non-empty regions are ordered, disjoint and unallocated. */
+
+        /* We make a fresh variable to index the known-empty region, because the
+         * SimplExportAndRefine verification test has poor support for array
+         * indices that are sums of variables and small constants. */
+        int empty_index = i + 1;
+
+        /* Try to take the top-most suitably sized and aligned chunk. */
+        pptr_t unaligned_start = ndks_boot.freemem[i].end - size;
+        pptr_t start = ROUND_DOWN(unaligned_start, max);
+        /* if unaligned_start didn't underflow, and start fits in the region,
+         * then we've found a region that fits the root server objects. */
+        if (unaligned_start <= ndks_boot.freemem[i].end
+            && start >= ndks_boot.freemem[i].start) {
             create_rootserver_objects(start, it_v_reg, extra_bi_size_bits);
-            if (i < ARRAY_SIZE(ndks_boot.freemem)) {
-                ndks_boot.freemem[next].end = ndks_boot.freemem[i].end;
-                ndks_boot.freemem[next].start = start + size;
-            }
+            /* There may be leftovers before and after the memory we used. */
+            /* Shuffle the after leftover up to the empty slot (i + 1). */
+            ndks_boot.freemem[empty_index] = (region_t) {
+                .start = start + size,
+                .end = ndks_boot.freemem[i].end
+            };
+            /* Leave the before leftover in current slot i. */
             ndks_boot.freemem[i].end = start;
-            break;
-        } else if (i < ARRAY_SIZE(ndks_boot.freemem)) {
-            ndks_boot.freemem[next] = ndks_boot.freemem[i];
+            /* Regions i and (i + 1) are now well defined, ordered, disjoint,
+             * and unallocated, so we can return successfully. */
+            return true;
         }
+        /* Region i isn't big enough, so shuffle it up to slot (i + 1),
+         * which we know is unused. */
+        ndks_boot.freemem[empty_index] = ndks_boot.freemem[i];
+        /* Now region i is unused, so make it empty to reestablish the invariant
+         * for the next iteration (when it will be slot i + 1). */
+        ndks_boot.freemem[i] = REG_EMPTY;
     }
 
-    return true;
+    /* We didn't find a big enough region. */
+    printf("ERROR: no free memory region is big enough for root server "
+           "objects, need size/alignment of 2^%"SEL4_PRIu_word"\n", max);
+    return false;
 }


src/kernel/boot.c
```diff
@@ -874,20 +874,39 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+        /* Invariant: both i and (i + 1) are valid indices in ndks_boot.freemem. */
+        assert(i < ARRAY_SIZE(ndks_boot.freemem) - 1);
+        /* Invariant; the region at index i is the current candidate.
+         * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
+         * Invariant: region (i + 1) is undefined, and should not be read.
+         * Invariant: regions above (i + 1), if any, are empty or too small to use.
+         * Invariant: defined regions (all except i + 1) are ordered,
+         *            non-overlapping and unallocated. */
+
+        /* Can we fit a suitably aligned chunk in the top end of region i? */
```
axel-h, line 886: In my change I also added the check `if (reg->end - reg->start >= size) { ... ` here first, because blindly calculating `ndks_boot.freemem[i].end - size` could lead to an underflow and then we have a very large value that always looks right.

src/kernel/boot.c
```diff
@@ -874,20 +874,39 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
     word_t size = calculate_rootserver_size(it_v_reg, extra_bi_size_bits);
     word_t max = rootserver_max_size_bits(extra_bi_size_bits);
     for (; i >= 0; i--) {
-        word_t next = i + 1;
+        /* Invariant: both i and (i + 1) are valid indices in ndks_boot.freemem. */
+        assert(i < ARRAY_SIZE(ndks_boot.freemem) - 1);
```
axel-h, line 878: Can't we even state `assert(i < ARRAY_SIZE(ndks_boot.freemem) - 2);` here, because we enforce an empty element at the end and the "skip any empty regions" loop in line 870 will walk over that one.

src/kernel/boot.c
```diff
@@ -874,20 +874,39 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
-                ndks_boot.freemem[next].start = start + size;
-            }
+            /* There may be leftovers before and after the memory we used. */
+            /* Leave the before leftover in slot i. */
             ndks_boot.freemem[i].end = start;
-            break;
-        } else if (i < ARRAY_SIZE(ndks_boot.freemem)) {
-            ndks_boot.freemem[next] = ndks_boot.freemem[i];
+            /* Shuffle the after leftover up to slot (i + 1), which is unused. */
+            ndks_boot.freemem[i + 1].start = start + size;
```
axel-h, line 895: Make this an "atomic" assignment that clearly states we updates the whole slot here and ot parts of it.
```suggestion
            ndks_boot.freemem[i + 1] = (region_t) { 
                .start = start + size;
                .end = ndks_boot.freemem[i].end;
            };
```

src/kernel/boot.c
```diff
@@ -874,20 +874,39 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+            ndks_boot.freemem[i + 1].end = ndks_boot.freemem[i].end;
+            /* Regions i and (i + 1) is now well defined, ordered, non-overlapping,
+             * and unallocated, so we can return successfully. */
+            return true;
         }
+        /* Region i isn't big enough, so shuffle it up to slot (i + 1),
+         * which we know is unused. */
+        ndks_boot.freemem[i + 1] = ndks_boot.freemem[i];
+        /* Now region i is unused, reestablishing the invariant for the next
```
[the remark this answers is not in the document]
mbrcknl: I'd prefer minimal code changes for reasons stated elsewhere.

src/kernel/boot.c
```diff
@@ -874,20 +874,48 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+         * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
+         * Invariant: region (i + 1) is undefined, and should not be read.
+         * Invariant: regions above (i + 1), if any, are empty or too small to use.
+         * Invariant: defined regions (all except i + 1) are ordered,
+         *            non-overlapping and unallocated. */
+
+        /* We make a fresh variable for the unused index, because the
+         * SimplExportAndRefine verification test has poor support for array
+         * indices that are sums of variables and small constants. */
+        int unused = i + 1;
```
axel-h, line 889: Nitpicking: The name "unused" for a variable is a bit misleading. You could just leave `word_t next = i + 1;` form the original code in.

src/kernel/boot.c
mbrcknl: > for the sake of some code golf.

Sorry, this comment of mine was neither fair nor reasonable.

src/kernel/boot.c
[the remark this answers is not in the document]
axel-h: I agree with you here, but here I consider this a "kind-of" bug that we have not zero'd this before. Technically it's not needed, but then at least a clarifying comment could be placed here that says we don't explicitly have to zero is because we will overwrite it anyway in the next iteration. And we accept that the freemem list contains a duplicate entry in slot 0 and 1 if no usable slot was found and this returns false, because then booting will stop anyway and nobody will need the freemem list any longer.

src/kernel/boot.c
axel-h: No problem.

src/kernel/boot.c
```diff
@@ -874,20 +874,48 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
+
+        /* Can we fit a suitably aligned chunk in the top end of region i? */
+        pptr_t unaligned_start = ndks_boot.freemem[i].end - size;
+        pptr_t start = ROUND_DOWN(unaligned_start, max);
+        /* start might have underflowed, so check both ends of the region. */
+        if (start >= ndks_boot.freemem[i].start
+            && unaligned_start <= ndks_boot.freemem[i].end)
```
axel-h, line 896: I might be more intuitive is we check `unaligned_start` in both cases. And for checking the end, we should use `<` as we need some space in the slot anyway.
```suggestion
      if (unaligned_start >= ndks_boot.freemem[i].start
            && unaligned_start < ndks_boot.freemem[i].end)
```

src/kernel/boot.c
mbrcknl: Interesting. I always thought `next` was not quite right. Arguably, it could just as well be `prev`, since we're going backwards. :-)

At least at this point in  the code, and at the places it's used, its significance is that it's the index of the unused entry in the array. Is "unused" the wrong word? Or does it need more qualification, as in `unused_region_index`?

axel-h: Hmm.. `empty` might be better than `unused`. My urge against the name `unused` is just because it's often used for variables that are unused, but still have to be there because you need something.

src/kernel/boot.c
mbrcknl: I'm not sure what alignment guarantees the regions in `ndks_boot.freemem` provide, but we'd need some guarantees to know that the `unaligned_start >= ndks_boot.freemem[i].start` check implies that `start >= ndks_boot.freemem[i].start`. The latter is what we really need, so I think it's clearest if we check for that. (Remember, `ROUND_DOWN` means that `start <= unaligned_start` regardless of underflow.)

For the second part, we just need to know we didn't underflow, to be sure that the first check is meaningful. And `unaligned_start <= ndks_boot.freemem[i].end` is the precise check for non-underflow. The responsibility for "we need some space" should be on `calculate_rootserver_size`. :-)

I think the problem here is the comment, so I'll clarify that. If that doesn't fix it, then we should go back to the nested conditional that you had in your PR.

src/kernel/boot.c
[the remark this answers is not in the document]
mbrcknl: All right, between this and the `unused` variable name, you've convinced me to make the region empty. Then I can call the variable `empty`. But I'll wait for the current proof run to finish, so we'll be able to tell whether this change is the one that trips up SimplExportAndRefine.

axel-h (approved):

src/kernel/boot.c
```diff
@@ -874,20 +874,50 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
...
     for (; i >= 0; i--) {
-        word_t next = i + 1;
-        pptr_t start = ROUND_DOWN(ndks_boot.freemem[i].end - size, max);
-        if (start >= ndks_boot.freemem[i].start) {
+        /* Invariant: both i and (i + 1) are valid indices in ndks_boot.freemem. */
+        assert(i < ARRAY_SIZE(ndks_boot.freemem) - 1);
+        /* Invariant; the region at index i is the current candidate.
+         * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
+         * Invariant: region (i + 1) is empty. */
+        assert(is_reg_empty(ndks_boot.freemem[i + 1]));
```
axel-h, line 882: we could move the variable declaration up:
```suggestion
        /* Invariant; the region at index i is the current candidate.
         * Invariant: regions 0 up to (i - 1), if any, are additional candidates.
         * Invariant: region (i + 1) is empty. */
        int empty_index = i + 1;
        assert(is_reg_empty(ndks_boot.freemem[empty_index]));
```

mbrcknl: I guess what I was trying to do was explain the invariants directly in terms of the loop variable, firstly so that the progression through the array is as clear as possible, but also so that the significance of the variable is known by the time we declare it.

kent-mcleod (approved):
@mbrcknl, feel free to merge this when ready.

PR #590, opened 2021-09-24
aarch64,hyp: Move PPTR_BASE down to 0x8000000000

kent-mcleod:
On aarch64 in EL2, there aren't any addresses after 2^48 and so the
any kernel device untypeds that have very large physical addresses could
potentially move into an invalid address range when translated to a
Kernel window PPTR address when being stored in a cap slot. The kernel
in EL2 doesn't need to share its address space with user level and so we
can make the kernel window start low enough that we can't get overflows.
We start from the second entry in the top level page table so that we
don't conflict with any setup code running in the lowest 512GiB of
virtual addresses.


include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -194,7 +194,7 @@
 /* This is a page table mapping at the end of the virtual address space
  * to map objects with 4KiB pages rather than 4MiB large pages. */
```
kent-mcleod, line 195: I don't think that this PT mapping exists any more.

lsf37 (approved):
Looking good!

lsf37:
`skylake` seems to be having problems. I've temporarily removed it from the queue until fixed. I.e. rerunning the hardware test should now work.

PR #595, opened 2021-09-27
unify libsel4 type definitions

axel-h:
Seems these definition are all based on generic types, so they could be move one level up and get unified across architectures then.


axel-h:
I've picked this from my working branch, but it seems this needs some include cleanup then also.

axel-h:
Can we merge this?

lsf37 (approved):
Yes, I think this makes sense. I can't see a new architecture where we would define these differently.

kent-mcleod (approved):

PR #596, opened 2021-09-27
move type checks to mode/types.h

axel-h:
Moving the type checks avoid the need for conditional compilation. This also allows simplifying some include file dependencies.


kent-mcleod:
@axel-h the `proof-test` label isn't needed if the preprocess tests pass.  (using the label on too many PRs at the same time causes the runners to get overloaded and start failing).

kent-mcleod (approved):

PR #597, opened 2021-09-27
boot: remove unnecessary cast

axel-h:
The cast does not exist in the ARM version of this code either. Both `vptr_t` and `paddr_t` ars aliases for `word_t`, so this is safe at the moment. If this ever changes and `vptr_t` gets a smaller size than `paddr_t`then I'd expect the compiler to issue a warning here without the cast, so we will catch this. With the cast in the code we silence such a compiler warning, which seems undesired.

### DIFF
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -207,8 +207,8 @@ static BOOT_CODE bool_t try_init_kernel(
 
     /* convert from physical addresses to userland vptrs */
     v_region_t ui_v_reg = {
-        .start = (word_t)(ui_p_reg_start - pv_offset),
-        .end   = (word_t)(ui_p_reg_end   - pv_offset)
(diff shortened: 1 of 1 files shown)


mbrcknl (approved):

lsf37:
Looks like we're running too many proof tests at the moment. Should try to restart in half an hour or so.

lsf37:
It's boot code, so the only effect would be in SimplExportAndRefine, which I think should be fine for this one. Happy to merge without waiting for the proof test. If it fails, we can still fix afterwards.

lsf37 (approved):

mbrcknl:
> It's boot code, so the only effect would be in SimplExportAndRefine, which I think should be fine for this one. Happy to merge without waiting for the proof test. If it fails, we can still fix afterwards.

Of course, you're right.

PR #599, opened 2021-09-28
avoid dependency from util.h on stdint.h

axel-h:
This also aligns the definition with GCC's definitions based on the native types. Add compile time check to ensure the type sizes match. Update the comments.

Avoiding a dependency from `util.h` on `stdint.h` saves us from running into circular dependency issues. Since `assert.h` includes `util.h`, this allows using `assert.h` in more places without causing side effects.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -418,30 +418,33 @@ static UNUSED CONST inline unsigned ctz64(uint64_t x)
 // These are only provided when the relevant config items are set.
 // We define these separately from `ctz32` etc. so that we can verify all of
 // `ctz32` etc. without necessarily linking them into the kernel binary.
+compile_assert(int_is_max_64_bit, sizeof(unsigned int) <= 8);
+compile_assert(long_is_max_64_bit, sizeof(unsigned long) <= 8);
+
 #ifdef CONFIG_CLZ_32
-CONST int __clzsi2(uint32_t x)
+CONST int __clzsi2(unsigned int x)
 {
-    return clz32(x);
+    return (4 == sizeof(int)) ? clz32((uint32_t)x) : clz64((uint64_t)x);
 }
 #endif
 
 #ifdef CONFIG_CLZ_64
-CONST int __clzdi2(uint64_t x)
+CONST int __clzdi2(unsigned long x)
 {
-    return clz64(x);
+    return (4 == sizeof(long)) ? clz32((uint32_t)x) : clz64((uint64_t)x);
 }
 #endif
 
 #ifdef CONFIG_CTZ_32
-CONST int __ctzsi2(uint32_t x)
+CONST int __ctzsi2(unsigned int x)
 {
-    return ctz32(x);
+    return (4 == sizeof(int)) ? ctz32((uint32_t)x) : ctz64((uint64_t)x);
 }
 #endif
 
 #ifdef CONFIG_CTZ_64
-CONST int __ctzdi2(uint64_t x)
+CONST int __ctzdi2(unsigned long x)
 {
-    return ctz64(x);
+    return (4 == sizeof(long)) ? ctz32((uint32_t)x) : ctz64((uint64_t)x);
 }
 #endif
--- a/include/util.h
+++ b/include/util.h
@@ -138,11 +138,11 @@ long PURE str_to_long(const char *str);
 // Library functions for counting leading/trailing zeros.
 // GCC's builtins will emit calls to these functions when the platform
 // does not provide suitable inline assembly.
-// We only emit function definitions if CONFIG_CLZL_IMPL etc are set.
-CONST int __clzsi2(uint32_t x);
-CONST int __clzdi2(uint64_t x);
-CONST int __ctzsi2(uint32_t x);
-CONST int __ctzdi2(uint64_t x);
+// We only emit function definitions if CONFIG_CLZ_32 etc are set.
+CONST int __clzsi2(unsigned int x);
+CONST int __clzdi2(unsigned long x);
+CONST int __ctzsi2(unsigned int x);
+CONST int __ctzdi2(unsigned long x);
 
 // Used for compile-time constants, so should always use the builtin.
 #define CTZL(x) __builtin_ctzl(x)


mbrcknl:
When I was implementing the current clz/ctz, I found GCC's docs and definitions confusing, which was why I got it wrong the first time around. Some other implementations use their own internal type definitions. IIUC, the point of these functions is that  `si` is 32 bits and `di` is 64 bits, regardless of what sizes the implementation chooses for `int` and `long`.

Unless I'm (still!) wrong about that, I personally don't think this is an improvement. Can you explain why a dependency on `stdint.h` is problematic, or why we should follow what GCC does?

src/util.c
```diff
@@ -419,28 +419,32 @@ static UNUSED CONST inline unsigned ctz64(uint64_t x)
...
+compile_assert(clzsi2_int_is_32_bits, 4 == sizeof(unsigned int))
+CONST int __clzsi2(unsigned int x)
 {
     return clz32(x);
 }
 #endif
 
 #ifdef CONFIG_CLZ_64
-CONST int __clzdi2(uint64_t x)
+compile_assert(clzdi2_long_is_64_bits, 8 == sizeof(unsigned long))
```
mbrcknl, line 430: In fact, I think this is just plain wrong for all of our 32-bit platforms, and the only reason this compiles is that we don't yet have any platforms that define `CONFIG_CLZ_64`.

See the `clz_word_size` `compile_assert` further up, which is always on, and shows that `unsigned long` is 32 bits on our current 32-bit platforms.

kent-mcleod:
> Avoiding a dependency from `util.h` on `stdint.h` saves us from running into circular dependency issues. Since `assert.h` includes `util.h`, this allows using `assert.h` in more places without causing side 

Sorry but stdint.h is a pretty foundational level header and restricting its use isn't something we would like to do unnecessarily.  If you want to static assert sizes of types from stdint.h couldn't you do that from a different file.  Alternatively, what's preventing removing util.h from assert.h? The UNUSED macro can be explicitly expanded in this case.

src/util.c
axel-h: You are right, this is broken. Fixed the check.

axel-h:
> When I was implementing the current clz/ctz, I found GCC's docs and definitions confusing, which was why I got it wrong the first time around. Some other implementations use their own internal type definitions. IIUC, the point of these functions is that `si` is 32 bits and `di` is 64 bits, regardless of what sizes the implementation chooses for `int` and `long`.

I found this https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html and the llvm rt-library supports you print as 'ti' is for up to 128-bit integers then. What I understand from the definition is that - since C has not 32-bit or 64-bit type internals, but only 'int', 'long' abnd 'long long' - the compiler will basically fall back to this native types eventually. So the implementation must support then and any mapping to 32-bit or 64-bit types that is made in the runtime library. Even `di` can still be just a 32-bit type then if 'long' is 32-bit.

mbrcknl:
> > When I was implementing the current clz/ctz, I found GCC's docs and definitions confusing, which was why I got it wrong the first time around. Some other implementations use their own internal type definitions. IIUC, the point of these functions is that `si` is 32 bits and `di` is 64 bits, regardless of what sizes the implementation chooses for `int` and `long`.
> 
> I found this https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html and the llvm rt-library supports you print as 'ti' is for up to 128-bit integers then. What I understand from the definition is that - since C has not 32-bit or 64-bit type internals, but only 'int', 'long' abnd 'long long' - the compiler will basically fall back to this native types eventually. So the implementation must support then and any mapping to 32-bit or 64-bit types that is made in the runtime library. Even `di` can still be just a 32-bit type then if 'long' is 32-bit.

That's the GCC page that led me to break the build for some of our platforms. In the end, I think I concluded that it is wrong, or at least quite misleading. I then spent a while studying the library sources of both GCC and  LLVM to reach the conclusion that regardless of the sizes of `int` and `long`, `si` is always 32 bits, `di` is always 64 bits, and `ti` (which we don't use) is always 128 bits.

These are library functions intended to be called from code emitted by compiler intrinsics. Normal people should never see them. We only implement them because we want to use intrinsics where possible (in case the platform has an instruction), but we don't link to the compiler's library. The intended interfaces for a C program are `__builtin_clz(unsigned)`, `__builtin_clzl(unsigned long)` and `__builtin_clzll(unsigned long long)`. What this means is that if you have a platform where `long long` is only 32 bits, you won't be able to get its compiler to emit a call to `__clzdi2`. It will always use `__clzsi2`.

You should be able to test my theory by compiling the following for each of riscv32 and riscv64, and seeing what calls the compiler emits:
```
int my_clz(unsigned x) { return __builtin_clz(x); }
int my_clzl(unsigned long x) { return __builtin_clzl(x); }
int my_clzll(unsigned long long x) { return __builtin_clzll(x); }
```

src/util.c
mbrcknl: If my conviction that `si` is always 32 bits and `di` is always 64 bits turns out to be correct, then I think your update is in the wrong direction.

mbrcknl:
Anyway, this is all beside the point until you explain what's problematic about a dependency on `stdint.h`.

mbrcknl:
> ```
> int my_clz(unsigned x) { return __builtin_clz(x); }
> int my_clzl(unsigned long x) { return __builtin_clzl(x); }
> int my_clzll(unsigned long long x) { return __builtin_clzll(x); }
> ```

I ran this experiment, and as I expected:
- On rv32, where `unsigned long` is 32 bits, `__builtin_clzl(unsigned long)` emits a call to `__clzsi2`.
- On rv64, where `unsigned long` is 64 bits, `__builtin_clzl(unsigned long)` emits a call to `__clzdi2`.

I'm closing this PR, because I think it is both poorly motivated and incorrect.

axel-h:
Yes, this one is obsolete. Thanks for the investigation. I've created https://github.com/seL4/seL4/pull/601 to follow up on the finding from here.

PR #600, opened 2021-09-28
allwinnerA20: add option to use the ARM generic timer

cvluca:
The memory mapping for timer only uses 1K on AllwinnerA20, but the minimum device mapping is 4K in seL4. Other devices within this 4K page (CCU and PIO) cannot be accessed in userland. This option allows kernel to use the arm generic timer, and then we will have the access rights for those devices in userland.

### DIFF
--- a/src/plat/allwinnerA20/config.cmake
+++ b/src/plat/allwinnerA20/config.cmake
@@ -8,6 +8,13 @@ cmake_minimum_required(VERSION 3.7.2)
 
 declare_platform(allwinnerA20 KernelPlatformAllwinnerA20 PLAT_ALLWINNERA20 KernelSel4ArchAarch32)
 
+config_option(
+    KernelPlatformAllwinnerA20ArchTimer ARM_ALLWINNER_A20_ARCH_TIMER
+    "Use the ARM generic timer for kernel"
+    DEFAULT OFF
+    DEPENDS "KernelPlatformAllwinnerA20"
+)
+
 if(KernelPlatformAllwinnerA20)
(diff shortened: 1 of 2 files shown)


src/plat/allwinnerA20/config.cmake
```diff
@@ -8,6 +8,13 @@ cmake_minimum_required(VERSION 3.7.2)
 
 declare_platform(allwinnerA20 KernelPlatformAllwinnerA20 PLAT_ALLWINNERA20 KernelSel4ArchAarch32)
 
+config_option(
+    KernelPlatformAllwinnerA20ArchTimer ARM_ALLWINNER_A20_ARCH_TIMER
```
axel-h, line 12: I wonder, is this an Alwinner A20 specific thing or is this something that could be generic for all ARM platforms since this is about the  "generic timer"?

cvluca: I think I probably gave a bad naming. This configuration just gives a choice for the kernel to use either the AllwinnerA20 specific timer (currently using) or the ARM generic timer. And we already have a generic timer implementation for all ARM platforms.

kent-mcleod:
If this platform has an arm generic timer, then that's what the kernel should always use shouldn't it?

cvluca:
> If this platform has an arm generic timer, then that's what the kernel should always use shouldn't it?

Thanks for mentioning that. I just created a new pr to remove the allwinner timer. (https://github.com/seL4/seL4/pull/610)
You might close this pr now.

lsf37:
Closing as requested for #610

PR #601, opened 2021-09-29
Add missing include and improve comments

axel-h:
- Follow-up form https://github.com/seL4/seL4/pull/599
- adding the missing include
- adding comment to capture the result of the tests.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -135,10 +135,28 @@ int PURE strncmp(const char *s1, const char *s2, int n);
 long CONST char_to_long(char c);
 long PURE str_to_long(const char *str);
 
-// Library functions for counting leading/trailing zeros.
-// GCC's builtins will emit calls to these functions when the platform
-// does not provide suitable inline assembly.
-// We only emit function definitions if CONFIG_CLZL_IMPL etc are set.
+/* Library functions for counting leading/trailing zeros.
+ *
+ * GCC/LLVM provides builtin function like __builtin_clzl() for this, which
+ * either get translated to machine specific instructions or calls helper
+ * functions like __clzsi2() that a compiler library is expected to implement.
+ * At the time of writing this comment, the GCC documentation about the compiler
+ * library (https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html)
+ * is not very detailed and the signatures given for these helper functions
+ * appear incorrect. For example, is says "int __clzsi2(unsigned int a)", but
+ * both the GCC and LLVM libraries implement it in a way that is independent of
+ * the implementation choices for the sizes of `unsigned int`. Instead, it
+ * appears that `si` always signifies a 32-bit argument and `di` always
+ * signifies a 64-bit argument. Tests with __builtin_clzl() on RISC-V have shown
+ * that if 'unsigned long' is 32 bits __builtin_clzl() uses __clzsi2() and if
+ * the type is 64 bits __builtin_clzl() uses __clzdi2(). Thus using the types
+ * uint32_t and uint64_t from stdint.h in the signatures below is considered the
+ * semantically correct way.
+ * Note that we only emit actual function implementations for these functions if
+ * CONFIG_CLZ_32 etc. are set. Otherwise, the compiler's internal implementation
+ * may get used or compilation fails if there is no machine instruction.
+ */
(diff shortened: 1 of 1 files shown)


include/util.h
```diff
@@ -135,10 +137,27 @@ int PURE strncmp(const char *s1, const char *s2, int n);
...
 long PURE str_to_long(const char *str);
 
-// Library functions for counting leading/trailing zeros.
-// GCC's builtins will emit calls to these functions when the platform
-// does not provide suitable inline assembly.
-// We only emit function definitions if CONFIG_CLZL_IMPL etc are set.
+/* Library functions for counting leading/trailing zeros.
+ * GCC provides the following builtin function for this:
+ *   __builtin_clz(unsigned int x)
+ *   __builtin_clzl(unsigned long x)
```
mbrcknl, line 143: We don't currently use `clz` or `ctz` anywhere. We do use `clzll` for MCS on 32-bit platforms, and `ctzl` in some places. Perhaps just pick one example that we do use (the most important is `clzl`).
You've also buried the important bit about why these exist in the note about the GCC docs page. I'd prefer to keep that up front, especially since we know the GCC docs page is misleading.

include/util.h
```diff
@@ -135,10 +137,27 @@ int PURE strncmp(const char *s1, const char *s2, int n);
...
+ * At https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html it
+ * says that GCC will call helper functions in case the hardware does not
+ * provide instruction for counting leading/trailing zeros:
+ *   int __clzsi2 (unsigned int a)
+ *   int __clzdi2 (unsigned long a)
+ *   int __ctzsi2 (unsigned int a)
+ *   int __ctzdi2 (unsigned long a)
```
mbrcknl, line 150: The only reason we should be mentioning the GCC docs page is to point out that it's misleading, i.e. that the types it shows for these operations are not consistent with the way the GCC (and LLVM) libraries implements the functions or the way the compilers emit calls to them.

If we're going to paste in the misleading part, we should make it very clear that it is incorrect.

```suggestion
 * At the time of writing this comment, the GCC docs
 * (https://gcc.gnu.org/onlinedocs/gccint/Integer-library-routines.html)
 * show INCORRECT signatures for these library functions:
 *   int __clzsi2 (unsigned int a)
 *   int __clzdi2 (unsigned long a)
 *   int __ctzsi2 (unsigned int a)
 *   int __ctzdi2 (unsigned long a)
 * These signatures are NOT consistent with the way GCC and LLVM
 * libraries implement these functions, nor the way the compilers emit
 * calls to them.
 * In the actual libraries and compilers, the sizes of the arguments to these
 * library functions are independent of the implementation choices for
 * the sizes of `unsigned int` and `unsigned long`.
 * - `si` always signifies a 32-bit argument.
 * - `di` always signifies a 64-bit argument.
```

include/util.h
```diff
@@ -135,10 +137,27 @@ int PURE strncmp(const char *s1, const char *s2, int n);
...
+ * Tests with __builtin_clzl() on RISC-V have shown that actual GCC behavior
+ * depends on the size of 'unsigned long'.
+ *   'unsigned long' is 32 bits: __builtin_clzl() uses __clzsi2().
+ *   'unsigned long' is 64 bits: __builtin_clzl() uses __clzdi2().
+ * Based on this observation, the suffix 'si' seems to indicate 32-bit integers
+ * and 'di' 64-bit integer. Thus using the types uint32_t and uint64_t in the
+ * signatures below appears to be semantically correct. We only emit function
+ * implementations if CONFIG_CLZ_32 etc are set. Otherwise, the compiler's own
+ * internal implementation may get used or compilation fails.
```
mbrcknl, line 159: I don't think we need to tell the story of how we came to this conclusion. In any case, it's missing the much longer story of the day or two I spent trawling through GCC and LLVM library code to figure this out.

mbrcknl (changes requested):

lsf37:
@mbrcknl are you happy with the updated version?

mbrcknl:
> @mbrcknl are you happy with the updated version?

Yep, looks good. Sorry for letting this get buried in my inbox.

mbrcknl (approved):

PR #602, opened 2021-09-29
add macros UINT64_C() and INT64_C()

axel-h:
Add macros to allow defining properly typed 64-bit constants.

### DIFF
--- a/include/stdint.h
+++ b/include/stdint.h
@@ -18,9 +18,12 @@ typedef signed short int16_t;
 typedef signed int int32_t;
 typedef signed long long int64_t;
 
-#define UINT64_MAX (0xFFFFFFFFFFFFFFFF)
+#define UINT64_C(x)  (x ## llu)
+#define INT64_C(x)   (x ## ll)
+
+#define UINT64_MAX UINT64_C(0xFFFFFFFFFFFFFFFF)
 #define UINT32_MAX (0xFFFFFFFF)
-#define INT64_MAX  (0x7FFFFFFFFFFFFFFF)
+#define INT64_MAX  INT64_C(0x7FFFFFFFFFFFFFFF)
 #define INT32_MAX  (0x7FFFFFFF)
 
 #define PRId64     "lld"


kent-mcleod (approved):
If the proofs come back green then this should be fine.  I wonder if there could ever be an issue with using `long long` literals on 64bit targets where the `word_t` type is defined as a `long` though.  I know it'd at least cause an issue with printf `-Wformat` warnings and `SEL4_PRI_word` .  

```
test.c:5:12: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 2 has type ‘long long int’ [-Wformat=]
    5 |  printf("%ld\n", 1ll);
      |          ~~^     ~~~
      |            |     |
      |            |     long long int
      |            long int
      |          %lld
```

axel-h:
> I wonder if there could ever be an issue with using `long long` literals on 64bit targets where the `word_t` type is defined as a `long` though. I know it'd at least cause an issue with printf `-Wformat` warnings and `SEL4_PRI_word` .

I'm a bit lost -  is this a hypothetical question so did you see this happen anywhere? You can use `SEL4_PRI_word` (and friends) only to print variables of the type `word_t` (or `seL4_word`). There should be be `PRIu64` (and friends) to print a `uint64_t` type. But `uint64_t` is completely decoupled form `word_t` because `uint64_t` is a machine-agnostic 64-bit format  and `word_t` is the machine register size. Random mixing or `uint64_t` and `word_t` will not work and breaks the concept of portable code. Same for other types that can have variable sizes, writing really portable code is sometime not trivial.

axel-h:
> If the proofs come back green then this should be fine.

Unfortunately they fail with: `Cannot load theory "HOL-Library.Word"`

lsf37:
> > If the proofs come back green then this should be fine.
> 
> Unfortunately they fail with: `Cannot load theory "HOL-Library.Word"`

This was an effect of the proof update to Isabelle2021, which should be fixed in #609. Rerunning now.

lsf37:
> But uint64_t is completely decoupled form word_t because uint64_t is a machine-agnostic 64-bit format and word_t is the machine register size. Random mixing or uint64_t and word_t will not work and breaks the concept of portable code. Same for other types that can have variable sizes, writing really portable code is sometime not trivial.

This probably doesn't matter for this PR, just wanted to point out that this depends in which code parts this is in. If you're in mode 64 code, you do know that `word_t` is 64bit (that's the whole point of mode 64 files after all), so it would be legitimate to use `word_t` for 64 bit situations there.

axel-h:
> This probably doesn't matter for this PR, just wanted to point out that this depends in which code parts this is in. If you're in mode 64 code, you do know that `word_t` is 64bit (that's the whole point of mode 64 files after all), so it would be legitimate to use `word_t` for 64 bit situations there.

It is still not ideal that the kernel does not do `typedef uint64_t word_t` and `typedef uint32_t word_t`. In the libsel4 it is done this was (https://github.com/seL4/seL4/tree/master/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/simple_types.h) and I was wondering if we could do this in the kernel also.

mbrcknl:
> > I wonder if there could ever be an issue with using `long long` literals on 64bit targets where the `word_t` type is defined as a `long` though. I know it'd at least cause an issue with printf `-Wformat` warnings and `SEL4_PRI_word` .
> 
> I'm a bit lost - is this a hypothetical question so did you see this happen anywhere? You can use `SEL4_PRI_word` (and friends) only to print variables of the type `word_t` (or `seL4_word`). There should be be `PRIu64` (and friends) to print a `uint64_t` type. But `uint64_t` is completely decoupled form `word_t` because `uint64_t` is a machine-agnostic 64-bit format and `word_t` is the machine register size. Random mixing or `uint64_t` and `word_t` will not work and breaks the concept of portable code. Same for other types that can have variable sizes, writing really portable code is sometime not trivial.

I think @axel-h is mostly right in this comment. Writing portable code with different sizes of integral types is hard. Part of the reason for that is that C is very liberal with implicit integer conversions, so the compiler will rarely tell you when you do something wrong.

*However*, I don't think this is PR is useful, for essentially the same reason.

For a hexadecimal constant without a suffix, the compiler is required to pick the first type in the following list that can represent the value ([n1256](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) 6.4.4.1):
- `int`
- `unsigned int`
- `long int`
- `unsigned long int`
- `long long int`
- `unsigned long long int`

For `0xFFFFFFFFFFFFFFFF`, that will be a 64-bit unsigned type. For our 64-bit targets, `unsigned long int`. For our 32-bit targets, `unsigned long long int`.

For `0x7FFFFFFFFFFFFFFF`, that will be a 64-bit signed type. For our  64-bit targets, `long int`. For our 32-bit targets, `long long int`.

This PR forces the constants to be typed as `unsigned long long int` and `long long int` respectively on our 64-bit targets. It has no effect on our 32-bit targets.

There are two main places where you might think this could make a difference. The unsigned and signed cases are analogous, so I'll just consider the unsigned case.

The first is using the old (un-suffixed) constant in a `uint64_t` context. In other words, using a `unsigned long` constant in  a `unsigned long long` context. This is going upwards in the integer conversion rank ([n1256](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf) 6.3.1), which won't lose precision. Compilers won't issue warnings. In this case, the PR makes no difference.

The second is going the other way: using the new (suffixed) constant in a `word_t` context. In other words, using a `unsigned long long` constant in a `unsigned long` context. This is going down the integer conversion rank. But this is still perfectly well-defined, and not an error even if the converted type is smaller. Compilers, however, can issue warnings, but typically only do so if they see an actual loss of precision of a constant value. On the 64-bit targets where this PR changes the type of the constant, there is no actual loss of precision, meaning no warning. So in this case, the PR results in no difference.

To wrap this up, consider one more case. Compare these:
```
uint32_t i = 0xffffffffffffffff;
uint32_t j = 0xffffffffffffffffull;
```

*Both* cases have an immediate loss of precision, so both cases will issue the same warnings. Still no difference from this PR.

So AFAICT, this PR makes no difference to portability or safety. But it does create a small amount of churn, so I am mildly against merging it.

axel-h:
Thanks @mbrcknl to the detailed feedback and pointing to the C specs section with the rules applies here. You have a point there, that this is indeed not necessary. I have extracted this PR from a change set where having `UINT64_C()` came handy, but I drop this PR then until there is a stronger need to add it.

PR #605, opened 2021-09-30
arm,generic_timer: Add config user access on 32bit

kent-mcleod:
The KernelArmExportPTMRUser and KernelArmExportVTMRUser options for
enabling user access to the generic timer were previously only supported
on aarch64 but are now supported on aarch32 also.


lsf37 (approved):

lsf37:
Should there be a CHANGES entry for this?

PR #606, opened 2021-09-30
qemu-arm-virt: Add KernelUserTop option on aarch32

kent-mcleod:
32-bit architectures have to share a small 4GiB address space between
privileged and non-privileged modes. Picking the address to split the
address space at affects how much virtual memory userlevel can use, and
how much physical memory the kernel can turn into untypeds. This is
something that should be configurable policy as different divisions make
sense for different applications.  For now we only add the config
option to qemu-arm-virt to try it out before potentially adding it to
all 32-bit platforms.


lsf37:
I think we have a bunch of alignment and other requirements on `seL4_UserTop`, which would be good to check if we make this configurable.

Are there compile time asserts elsewhere for that already? I can try to extract what we need in the proofs (I might also be wrong and there aren't any real requirements on this specific one). @Xaphiosis do you remember anything from the top off your head?

kent-mcleod:
> I think we have a bunch of alignment and other requirements on `seL4_UserTop`, which would be good to check if we make this configurable.

Well seL4_UserTop is what becomes PPTR_BASE and across all of the aarch32 platforms it's given different values. It's a pretty annoying constant because it affects the highest address that userlevel is allowed to create vspace mappings for while also affecting how much physical memory can be used for kernel untypeds.

Xaphiosis:
I think we need these to be page-aligned. On our 64-bit platforms I'm seeing:

```
X64:
pptrUserTop = 0x00007fffffffffff

RISCV64:
"pptrUserTop \<equiv> mask canonical_bit && ~~mask 12" (* for page boundary alignment *)
```

Note that `pptrBase` is treated differently, i.e. on RISCV64:
`pptrBase = - (1 << canonical_bit)`

We also need to be sure that everything below pptrUserTop is in the user region, and demand that `user_vtop < pptr_base` which is really `pptrUserTop < pptrBase`. I think the most up-to-date guide to what we require on 64-bit platforms can be seen in: `l4v/proof/invariant-abstract/RISCV64/ArchInvariants_AI.thy`

For 32-bit platforms, which for us means ARM/ARM_HYP, there indeed does not appear to be a `pptrUserTop`, as Kent indicated via the code. We use `pptrBase`, and demand that the end of objects don't go past that.

So a preliminary investigation would indicate no issues, but of course this will come out the moment someone wants a verified platform with that setting changed.

PR #367, opened 2021-05-24
Fix d-cache flush on aarch32

cvluca:
I had trouble boot with allwinner A20 (Cortex-A7) and got it fixed by cleaning to PoC.
It seems all Cortex-A7 (I.MX7, AllwinnerA20), Cortex-A15 (Exynos5, Tk1), and Cortex-A53 boards require to flush to coherency.

PR #610, opened 2021-10-02
allwinnerA20: switch to the ARM generic timer

cvluca:
The memory mapping for the timer only uses 1K on AllwinnerA20, but the minimum device mapping is 4K in seL4. Other devices within this 4K page (CCU and PIO) cannot be accessed in the userland.

This change will replace the kernel timer with the ARM generic timer on AllwinnerA20, and remove the implementation for AllwinnerA20 specific timer in the kernel. So we should have user access to those devices now.


kent-mcleod:
Cool.  Can I ask what the details of the allwinnerA20 platform you are using are? As far as I was aware, there were no AllwinnerA20 boards available for use or testing by anyone and the kernel port was unmaintained.  There's not even allwinnerA20 platform support in the user libraries to allow sel4test to run.

cvluca:
> Cool. Can I ask what the details of the allwinnerA20 platform you are using are? As far as I was aware, there were no AllwinnerA20 boards available for use or testing by anyone and the kernel port was unmaintained. There's not even allwinnerA20 platform support in the user libraries to allow sel4test to run.

I am using CubieTruck. Yes, you are right. The support for allwinnerA20 was unmaintained and I got a lot of trouble to make sel4 running on it. There is still an unfixed issue in the kernel to make it run (https://github.com/seL4/seL4/pull/367). I am also working on the platform support in the user libraries, and I will contribute later.

PR #611, opened 2021-10-04
Remove C semantics from CMake files

axel-h:
There seems no need to pull in C semantics into the CMake files. The constant `CONFIG_KERNEL_WCET` in not supposed to be used directly, but the value must be retrieved by calling `getKernelWcetUs()`, which add the proper type semantics. Within the function itself, the compiler can do the integer conversion from the plain number automatically, which still allows constant folding to happen on higher optimization levels.

This basically follows the road of 7d379904650b76f4d33d5c78b232879c5bcbccc6 that did remove the suffix for `TIMER_FREQUENCY` in the CMake files.

### DIFF
--- a/include/arch/riscv/arch/machine/timer.h
+++ b/include/arch/riscv/arch/machine/timer.h
@@ -18,8 +18,10 @@
 
 static inline CONST time_t getKernelWcetUs(void)
 {
-    /* Copied from x86_64. Hopefully it's an overestimate here. */
-    return  10u;
+    /* The value is copied from x86_64, hopefully it's an overestimate here.
+     * Making this an explicit unsigned type simplifies verification.
+     */
+    return 10u;
 }
 
 static inline PURE ticks_t usToTicks(time_t us)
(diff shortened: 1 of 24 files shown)


axel-h:
I've keep the unsigned types and added a comment explaining why thy are there. Also added a 'U_CONST()' macro handle the constants. 
The main motivation for at the moment for removing the C semantics from CMake is avoiding setting a bad example, as it feels this copy/pasted for everything that is added. Furthermore, putting this explicitly in C code gives a pattern that can be used in other places also where the semantics matter and where the handling in the proof could be be aligned. Limiting the CMake constants to C seems artificial limitation to me that just exists due to historical reasons and convenience.

kent-mcleod:
> Limiting the CMake constants to C seems artificial limitation to me that just exists due to historical reasons and convenience.

These constants are being given to a CMake function that is named `declare_default_headers` because it is explicitly declaring C headers.  Many of the other arguments are also C literals.

PR #616, opened 2021-10-16
Fix indexing into an empty array

elelel:
When compiling hello world tutorial the following error happens:
```
  File "/home/sel4repo/sel4-tutorials-manifest/kernel/tools/circular_includes.py", line 61, in main
    while file_stack[-1] != header:
IndexError: list index out of range
```

This PR fixes the indexed access into potentially empty array

### DIFF
--- a/tools/circular_includes.py
+++ b/tools/circular_includes.py
@@ -58,7 +58,7 @@ def main(parse_args):
                 file_stack.append(header)
         else:
             # popped back up to an earlier header
-            while file_stack[-1] != header:
+            while (len(file_stack) > 0) and (file_stack[-1] != header):
                 file_stack.pop()
 
     return 0


lsf37:
Hi @elelel, thanks for that contribution! Would you mind adding a sign-off line to your commit message to indicate that you are agreeing with the [DCO]?

It's curious that this didn't trigger in any of the regression tests. Is this on the tutorial solution or on your own changes to the template?

(It's good to guard against this case in anyway, just trying to track down if there is something missing in the test setup).

[DCO]: https://docs.sel4.systems/processes/contributing.html

elelel:
@lsf37 This is neither tutorial solution, nor my changes to the tutorial. This is the base Hello World example, "building your first program" part from here: https://docs.sel4.systems/Tutorials/hello-world.html which does not require the learner anything other than compiling the source.

Here are the details if you're interested in tracking it down and cover with tests. The main difference from baseline setups, I think, is that I use native environment (not Docker image) with Archlinux. I installed the dependencies as described here: https://docs.sel4.systems/projects/buildsystem/host-dependencies.html as well as tutorial dependencies. Any non-matching package names were updated from Ubuntu's to Archlinux (by trivially searching the web). I set up a new system user sel4repo, because there is an identity leak with repo tool, that requires you to have git --global name/email to be present in the env, without it invoking repo on manifests will crash. I ran init as described here: https://docs.sel4.systems/Tutorials/hello-world.html and entered the CMake build directory and ran ninja. No changes or branch selections were made to any of the original sources.

My guess is that Archlinux which usually has more recent packages than Ubuntu has some package as a dependency with a behavior inconsistent with previous versions. Without any specific knowledge of seL4 build system I would look into Python part as this is usually the common place to produce non-deteministic results across versions. 

Out of curiousity, why seL4 uses Google repo tool? Is it for historical reasons (i.e. from olf times when git didn't have similar functionality) or it still supports some key functionality not supported by git?

axel-h:
I think repo is used for historical reasons and it does the job quite well so far. What would you suggest as native git replacement - submodules?

elelel:
I think git submodules is more natural hyper-meta management git repository management these days. But I'm not suggesting to use that in seL4, I'm more curious about the reasons of the choice, in case I consider use it myself for my seL4-related project. If there is some functionality repo tool has which is used in seL4 that is absent from git submodules mechanism, I thought it would be a good opprtunity to find out.

canarysnort01:
I can't speak to the seL4 team's decision to use repo, but I have worked on a project that tried to use git submodules as a means of organizing a larger project made up of multiple git repositories and the experience was awful.

I do not recommend trying to use git submodules for that.

axel-h:
At Hensoldt Cyber we are using submodules for our seL4 SDK, but the tight coupling there turns out to have drawbacks also, the more loose coupling of repos seems to scale nicer due to the xml files. I have not tried git subtrees yet.
In the end it's up to the tooling and scripts to handle this. The repo tool has a few advanced features implemented already (documentation could be better) that gives it a bit of a head start.

elelel:
@canarysnort01, @axel-h thank you for your opinion on submodules and repo tool, I shall look at the latter in more depth.

lsf37:
I can confirm that we picked `repo` because it allows you to choose the degree of coupling between the repos. For development, we prefer loose coupling by branch name, and for releases we prefer precise revisions/tags. `repo` allows you to do both relatively easily. The downside is that you need an external tool (repo) and more management.

Loose coupling is very hard to achieve with git submodules and previous experience has not been great. We used mercurial subrepos on the verification side for a long while, and that was alright, but less flexible than `repo`. We also discussed big mono-repos, but didn't go for that either. None of the solutions are really great.

Issue #509, opened 2021-08-24
Allow Github CI actions to run on custom local hardware


axel-h:
- Can we make Github CI actions run on custom local hardware also? 
- In a repo fork, what needs to be adapted that this is tested locally (both hardware and AWS proofs)
- Assume there are multiple test farms, how could this be handled?

lsf37:
Yes, that would be very nice to do. I haven't finished yet setting up the UNSW one for the rest of the repos and wanted to do that before thinking more about how to distribute it, but we can start discussing this now.

The current architecture for hardware tests is that there is
 * the GitHub CI action that compiles images
 * followed by an action that downloads those images and selects the correct board for it, then uses the UNSW machine queue commands to
     * request a lock for that board
     * issue a `run` command that uploads the image(s) and runs the test, getting output on the console and a log file in the end
     * release the lock
     * parse the log file for success/error

The machine queue is currently a set of shell scripts in [seL4/machine_queue](https://github.com/seL4/machine_queue) (currently still private, to be published, committers have read access at least atm) plus another set of python scripts to handle the actual power reset/boot/console for each board (also still private).

The latter is specific to the setup at UNSW. The former could be expanded to a reasonable larger interface. It's not published yet, because it currently requires `ssh` access to UNSW machines to use it, and @wom-bat  wanted to implement a web-hook solution for submitting tests first so that we can move away from that. There is also no reason that this has to be implemented as shell scripts, that's just historically grown (although it works fine for what it needs to so far).

In terms of how that interface could work, I would basically not change anything for the command line of `mq.sh`, but for each board, the scripts should know which test server to contact, instead of always going to UNSW. So if an organisation wants to add a new board to the test set, it would add the parameters that we have now, plus which server (or which submission URL if that web-hook interface is better in the end).

`mq.sh` currently calls an interface `reboot` with a bunch of options that actually runs the test. That is the bit each org would implement for themselves (or copy the UNSW setup if that is convenient). As mentioned, this is currently over `ssh`, so maybe not ideal, although if set up correctly that doesn't have to be a problem.

What I'm not clear on yet for the web-hook idea is how we would get live logs from the test back, which is much nicer than just a logfile in the end, because you can see if things are hanging or otherwise failing and then cancel tests.

Does this make some kind of sense? I haven't thought through this very deeply yet.

lsf37:
Logging here that we will likely need to adjust some parts of #618 to enable forks to do their own hardware tests.

[...]

PR #618, opened 2021-10-17
github: run actions with secrets only in seL4 org

lsf37:
Some of the Github actions in this repository require secrets for
access to hardware test or other compute infrastructure, or for
deployment. When these are run in a fork on GitHub they are safe,
but generate annoying error messages.

This commit adds tests in the relevant actions whether they are run
from the seL4 org or not, and skips the action when they are run
outside. This should make the remaining actions more useful on forks.

### DIFF
--- a/.github/workflows/preprocess-deploy.yml
+++ b/.github/workflows/preprocess-deploy.yml
@@ -40,6 +40,7 @@ jobs:
   deploy:
     name: Deploy manifest
     needs: [code, preprocess]
+    if: ${{ github.repository_owner == 'seL4' }}
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


axel-h:
I have not given up in the plan to set up a HW CI on our forks. Is there a way to check if the secrets exists instead of checking the repo owner?

lsf37:
Good question. I'm not sure, actually, it could be possible. Deployment actions should still not run, but a hardware test or even a custom proof test could. 

I'll look into it.

lsf37:
Ok, I've noted in #509 that we'll need to adjust this again, but I'll merge this now so that we get the intermediate improvement.

PR #626, opened 2021-10-22
Fix linking for armv8-a with gcc 10.1

Indanz:
Starting from gcc 10.1 -moutline-atomics is enabled by default.
This option causes gcc to generate code that will determine at
runtime whether the hardware supports the newer armv8.1-a atomic
instructions. This is useful if the same binary will be used on
multiple platforms. As this is not the case for seL4, disable this
option, it will only cause unnecessary runtime overhead.

If later it is decided to enable outline-atomics, seL4 will need
to link against libgcc.

### DIFF
--- a/src/arch/arm/64/config.cmake
+++ b/src/arch/arm/64/config.cmake
@@ -21,3 +21,7 @@ add_sources(
         kernel/vspace.c
     ASMFILES head.S traps.S
 )
+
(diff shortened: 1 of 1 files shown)


lsf37:
> I probably tested with an old kernel and not the most recent one, silly me. I'll check Monday and open another issue instead if necessary.

No worries, do let me know if it is still not working.

PR #628, opened 2021-10-25
Custom DTS overlay file support

Indanz:
Add KernelCustomDTSOverlay option to append user provided DTS overlay
files. This way users do not need to modify the seL4 kernel sources
just because they use a hardware module differently. Also useful for
defining VM memory regions.


kent-mcleod:
@Indanz, I'm happy for this to be merged if you are happy that it meets your use case.

Indanz:
Yes, please do.

Thanks!

Issue #633, opened 2021-10-26
seL4 failed assertion 'sc->scRefillHead < sc->scRefillMax'


Indanz:
Not sure what's going on, but since today occasionally I'm getting:
```
seL4 failed assertion 'sc->scRefillHead < sc->scRefillMax' at /home/iz/vmm-mep/kernel/src/kernel/sporadic.c:128 in function refill_pop_head
halting...
Kernel entry via Interrupt, irq 0
```
It may be triggered by application code calling seL4_DebugPutString() though.

seL4 commit 561c6e65bef2d9, SMP, MCS and HYP configured.

I never got this with an older kernel, it seems something fell over since June / commit db7bb413.

Indanz:
The only other place that sets ksCurThread is switchToThread(), which is only called by schedule(), either directly or indirectly, and schedule() always calls switchSchedContext().

switchToIdleThread() is only called via schedule() or by ipiStallCoreCallback(), so all in all ipiStallCoreCallback() is the only call path which changes the current thread without updating the ksCurSC.

Is the above fix sufficient, or should ipiStallCoreCallback() do more, like call switchSchedContext()?

lsf37:
Thanks for debugging this, `sc->scRefillHead < sc->scRefillMax` is definitely supposed to not fail. @pingerino or @kent-mcleod might be able to say if we need `switchSchedContext()` here or if the assignment is enough.

@mktnk3 and @michaelmcinerney anything from the verification side on this one? We don't see the `ipiStallCoreCallback()` code path in the verification, because that is SMP only, but I think we have an invariant about `ksCurSC` and `ksCurThread` that might be relevant here.

pingerino:
IIRC you don't need swtichSchedContext when switching to the idle thread (the idle thread always has budget and isn't doing anything fancy with thread states, nor can it donate its scheduling context). 

Are we still running the SMP + MCS version of sel4test? what about sel4bench? one of these should have caught this, and if not, we should add a test case. 

Thanks for reporting and investigating!

michaelmcinerney:
> @mktnk3 and @michaelmcinerney anything from the verification side on this one? We don't see the `ipiStallCoreCallback()` code path in the verification, because that is SMP only, but I think we have an invariant about `ksCurSC` and `ksCurThread` that might be relevant here.

[...]

PR #636, opened 2021-11-02
Avoid scheduling context use-after-free

Indanz:
When changing to the idle task in ipiStallCoreCallback(),
ksCurSC should change too, as a remote core may be deleting
the scheduling context. See also issue #633.


Indanz:
The proofs need updating because of the chargeBudget() cleanups. We can also merge the other commits first and do the chargeBudget() cleanup later.

Indanz:
An alternative approach would be to do all SC and TCB operations on the core they're currently running on. This would get rid of ipiStallCoreCallback() altogether and replace it with multiple separate IPI calls, but that would be a more invasive change overall.

pingerino:
> ghost/runtime/src/main/resources/ghostinit.conf:services."ghost.runtime.services.RecordingService" 

Honestly I think this would be a better design and it's one I have wished for. Cores modifying each others schedulers is generally not great for performance. It would also enable us to remove  more operations from the global lock on SMP. But that's a much, bigger change and needs deep thought and eval. 

I do not have the time available to look at this PR in the detail it needs - I'm hoping someone else can give it the attention needed.

lsf37:
From the proof side this looks fine, but I've added @mktnk3 and @michaelmcinerney to the review requests so one of them can confirm.

pingerino:
An alternative approach would be to do all SC and TCB operations on the core they're currently running on.

Honestly I think this would be a better design and it's one I have wished for. Cores modifying each others schedulers is generally not great for performance. It would also enable us to remove more operations from the global lock on SMP. But that's a much, bigger change and needs deep thought and eval.

I do not have the time available to look at this PR in the detail it needs - I'm hoping someone else can give it the attention needed.

mktnk3:
> From the proof side this looks fine, but I've added @mktnk3 and @michaelmcinerney to the review requests so one of them can confirm.

Yes, as far as I can see, I agree that this won’t affect the current MCS verification.

Indanz:
> ```
> An alternative approach would be to do all SC and TCB operations on the core they're currently running on.
> ```
> 
> Honestly I think this would be a better design and it's one I have wished for. Cores modifying each others schedulers is generally not great for performance.

The main benefit would be that the code paths are (almost) the same for SMP and non-SMP, so the proofs apply more to SMP configurations. My idea was to add macros that either call a function directly, or via IPI on another core. That way the argument passing and correctness of the functions is checked by the proven non-SMP build.

As IPI calls are synchronous, I don't think performance would be much different though, as the total kernel time won't be much reduced I think.

> It would also enable us to remove more operations from the global lock on SMP. But that's a much, bigger change and needs deep thought and eval.

If the data could be partitioned between cores and access is proven to be correct by the proofs, this would be much easier to do. Then only global data would need the global lock. However, that seems like a lot of work on the proof side of things.

Indanz:
(I cleaned up the history, not real changes with the last push.)

lsf37:
seL4/sel4test#56 adds a test for this

lsf37 (approved):
Ok, I'm happy with the change as is. As far as I can trace the time accounting, the change in `schedcontrol.c` is correct, setting `ksCurSc` for the IPI is also correct, and proof impact is minimal.

lsf37:
The test failure on TK1 is unrelated, and the absence of a PC99 and TX2 failure means the race condition was not triggered any more (either because this fixes it or it failed to trigger this time, but it has triggered fairly reliably before).

lsf37:
So, side discussion with @kent-mcleod does reveal one concern: with the fix as it is, we are switching time accounting to the idle thread when we switch to the idle thread. That is usually the right thing, but now it is possible to inflate the budget of a thread on core A if you have a cap to it from core B.

For instance, if you do a loop of many writeRegister calls from core B to A, you will stall core A every time. The unstall is expected to come from the core B user and could be delayed. All that stalled time is accounted to the idle thread now, but that is not really the right place for it to go.

It does sounds like the budget of the thread on core A should still be charged. Basically the idle thread running is a bit of a hack here.

michaelmcinerney:
> So, side discussion with @kent-mcleod does reveal one concern: with the fix as it is, we are switching time accounting to the idle thread when we switch to the idle thread. That is usually the right thing, but now it is possible to inflate the budget of a thread on core A if you have a cap to it from core B.
> 
> For instance, if you do a loop of many writeRegister calls from core B to A, you will stall core A every time. The unstall is expected to come from the core B user and could be delayed. All that stalled time is accounted to the idle thread now, but that is not really the right place for it to go.
> 
> It does sounds like the budget of the thread on core A should still be charged. Basically the idle thread running is a bit of a hack here.

I know next to nothing about multicore, and I haven't been closely following this issue, but the question of the idle thread has come up in trying to show that the current sc is ready and sufficient (with "buffer" the consumed time) after the kernel runs. @pingerino mentioned that it might be best to not charge the idle sc, since if the idle thread is running, then the system isn't really using time. I've started tinkering with this idea in the proofs, but it is quite annoying in places, and it might be best to make a decision on what to do with the idle thread/sc, especially if it might affect several issues

lsf37:
> @pingerino mentioned that it might be best to not charge the idle sc, since if the idle thread is running, then the system isn't really using time. I've started tinkering with this idea in the proofs, but it is quite annoying in places, and it might be best to make a decision on what to do with the idle thread/sc, especially if it might affect several issues

If I understand correctly you mean by "not charging the idle thread" not changing the SC when the idle thread is running. I think that would make sense in general, but is orthogonal to this specific problem where (at least I think) we do want to charge the time somewhere (to thread A if I'm correct).

In fact, that suggests to not make the idleSC assignment and leave the inconsistent state here (idleThread as current thread, old thread A SC as current SC), but deal with deletion explicitly instead.

That could be done by finaliseCap explicitly checking curSC and setting it to NULL if it is the one that is being deleted. Probably complicated by needing to check the curSC of all cores, which is annoying, but deletion is already slow anyway.

We'd then have to allow the combination of idleThread being the current thread, and another SC (not idle) being the current SC in the invariants, because we still want to charge time to that SC even though the thread is not strictly speaking running. Main question there is: are there other corner cases apart from delete where things could go horribly wrong?

(Maybe we should instead have a thread state "stalled" that signifies that the thread is still the active thread occupying CPU, but is currently idle. Sounds like a much bigger change, though)

lsf37:
One incremental option forward would be to merge the PR as is and not an issue that time accounting is off for IPIs. That we at least fix the crashing problem, turning it into a potential timing problem.

I'm not clear enough on how well time accounting is validated for the SMP case to say if that would make the situation significantly worse there.

michaelmcinerney:
I agree that these changes don't affect the MCS proofs, so I'd be happy to merge this as is and then consider the other issues separately.

I wish I did know a bit more about multicore so that I could add something to the discussion. Perhaps having the idle thread be the current thread, but not linked to the idle sc, is not significantly more difficult an issue than any other we face in multicore.

lsf37:
> Perhaps having the idle thread be the current thread, but not linked to the idle sc, is not significantly more difficult an issue than any other we face in multicore.

Do you have an impression of what the situation would be on the single core proof if we made that exception to the invariant there? (e.g. we wouldn't create the situation, but we wouldn't rely on it being true on kernel entry).

michaelmcinerney:
> Do you have an impression of what the situation would be on the single core proof if we made that exception to the invariant there? (e.g. we wouldn't create the situation, but we wouldn't rely on the old invariant being true on kernel entry).

For deletion? I think all those proofs were before my time. I would guess (maybe poorly) that if we still do not have non-zombie caps to the idle thread and idle sc, then deletion shouldn't be much of an issue.

As for charging scs, we have absolutely nothing about charging the idle sc; in fact, `valid_idle` doesn't even specify a refill list for the idle sc (or maybe it says it's empty or something else we don't want). If we do want to charge the idle sc, then we have to worry about underflow in `refill_budget_check_round_robin`, which is its own problem (which might require introducing a cap on the amount the consumed time can get to). In my little experiment with not charging the idle, it seems like we don't need it linked to the idle thread, but there are still some invariants to go (for example, "if the current sc is tied to a thread in the release queue, then the consumed time is 0")

lsf37:
> For deletion? I think all those proofs were before my time. I would guess (maybe poorly) that if we still do not have non-zombie caps to the idle thread and idle sc, then deletion shouldn't be much of an issue.
>
> As for charging scs, we have absolutely nothing about charging the idle sc; in fact, valid_idle doesn't even specify a refill list for the idle sc (or maybe it says it's empty or something else we don't want). If we do want to charge the idle sc, then we have to worry about underflow in refill_budget_check_round_robin, which is its own problem (which might require introducing a cap on the amount the consumed time can get to). In my little experiment with not charging the idle, it seems like we don't need it linked to the idle thread, but there are still some invariants to go (for example, "if the current sc is tied to a thread in the release queue, then the consumed time is 0")

Ok, this sounds mildly encouraging.

Indanz:
> I'm not clear enough on how well time accounting is validated for the SMP case to say if that would make the situation significantly worse there.

I don't think time is accounted properly for SMP, as all accounting is done per core, but on SMP time spend with the kernel lock held and IPIs may also affect available time on other cores. With SMP it has always been possible to "steal" time from other cores, which isn't accounted properly as far as I can tell.

- Before this patch, the current SC would be charged for the IPI time spend when the next reschedule happens, even if that SC doesn't exist any more.
- With this patch, the current SC gets charged immediately and no one gets charged for the IPI time.

> but now it is possible to inflate the budget of a thread on core A if you have a cap to it from core B.

I don't think so. You can mess up total time accounting for another core, but you can't give a thread more budget this way.

> For instance, if you do a loop of many writeRegister calls from core B to A, you will stall core A every time. The unstall is expected to come from the core B user and could be delayed. All that stalled time is accounted to the idle thread now, but that is not really the right place for it to go.

Neither is charging a random SC that didn't cause this the right place. Because that makes it possible to cause threads on core A to miss their deadline. If you charge no one for the IPI time, at least the remaining time is divided according to the scheduling policies of the threads.

> It does sounds like the budget of the thread on core A should still be charged.

Why? Either you account for globally locked operations and then it's the thread on core B that should be charged extra, or time accounting isn't strictly "correct" because other cores can steal time. If that's the case, charging such stolen time to a random thread won't improve anything.

> Basically the idle thread running is a bit of a hack here.

That we agree on.

Best would be if there are no synchronous IPI calls at all, and that everything happens either async or from one place with the global lock held. Any time with the global lock held should be multiplied by either the number of cores, or the number of cores waiting for the lock. That way at least the one responsible would be charged extra.

Indanz:
A different hack would be to switch the SC to the thread causing the IPI instead of the idle thread's SC, but that seems a bit scary.

lsf37:
> I don't think time is accounted properly for SMP, as all accounting is done per core, but on SMP time spend with the kernel lock held and IPIs may also affect available time on other cores. With SMP it has always been possible to "steal" time from other cores, which isn't accounted properly as far as I can tell.

That is fair enough, and means this change doesn't make that situation significantly worse (but also doesn't make it better).


> * Before this patch, the current SC would be charged for the IPI time spend when the next reschedule happens, even if that SC doesn't exist any more.
> * With this patch, the current SC gets charged immediately and no one gets charged for the IPI time.

Agreed.


> > but now it is possible to inflate the budget of a thread on core A if you have a cap to it from core B.
> 
> I don't think so. You can mess up total time accounting for another core, but you can't give a thread more budget this way.

True, it's not really added to the budget of that thread, it's removed from the overall time available on that core. In the scenario we were looking at (operation from core B that leads to a stall on core A) you would always have a cap to the A thread, so in that sense you're causing time delay on behalf of A. 
 
> > For instance, if you do a loop of many writeRegister calls from core B to A, you will stall core A every time. The unstall is expected to come from the core B user and could be delayed. All that stalled time is accounted to the idle thread now, but that is not really the right place for it to go.
> 
> Neither is charging a random SC that didn't cause this the right place.  Because that makes it possible to cause threads on core A to miss their deadline. If you charge no one for the IPI time, at least the remaining time is divided according to the scheduling policies of the threads.

It's not random -- you need authority to that A thread or its SC to cause this (otherwise you don't stall the core, at least if we've traced this correctly, it's entirely possible that I've missed a path), and you'd only be causing A to miss its deadline. I don't think you can cause other threads to miss their deadline if it's accounted to A -- even if you don't re-enable thread A at all, core A gets the timer interrupt that causes a re-schedule and will switch away from the idle thread. On the other hand if you don't charge anyone (i.e. idle) then if A is high priority, it will reclaim the time it has lost and starve out the rest. 

One could also make the argument that the B thread should be charged for that time, but it seems strange to me to account time on one core to another, because that core doesn't really get more or less time to run.

This argument hinges on the fact that you must have presented authority to thread A for a stall of A to happen. If that is wrong, then it really is a random thread, and things are worse.

All of this is a bit theoretical, because if you're doing real-time you really shouldn't be messing with threads while they are running on a different core if those have important deadlines. Other cores shouldn't have caps to anything on the real-time core(s). It'd be extremely hard to figure out schedulability for that otherwise, even with perfect accounting.

> > It does sounds like the budget of the thread on core A should still be charged.
> 
> Why? Either you account for globally locked operations and then it's the thread on core B that should be charged extra, or time accounting isn't strictly "correct" because other cores can steal time. If that's the case, charging such stolen time to a random thread won't improve anything.

I do take your point that the rest of MCS on SMP doesn't really account for other global locking scenarios either, so trying to fix that detail here doesn't help much in the bigger picture. 

> > Basically the idle thread running is a bit of a hack here.
> 
> That we agree on.
> 
> Best would be if there are no synchronous IPI calls at all, and that everything happens either async or from one place with the global lock held. Any time with the global lock held should be multiplied by either the number of cores, or the number of cores waiting for the lock. That way at least the one responsible would be charged extra.

I don't think fully async is possible when threads on one core can remove objects for another core that are necessary for user mode to run. Most directly the top-level page table -- if that is zeroed then kernel entry would crash because the kernel window disappears, but also the SC, because if that disappears the thread should stop running because it has no authority to time any more. Similar for the TCB of the running thread itself.  If we could set it up such that these can't happen (be modified from another core), then fully async could work, but I don't really see how.

(It's easy enough to set up a system with partitioned resources that satisfies this, but it's not easy to check from the kernel perspective that you are in such a case. Come to think of it, if you do set up a system that is partitioned that way there is a good chance that you already won't get any of these stalls and you only have the global lock to account for. I haven't checked everything, but all cases I've seen so far would not happen).

Indanz:
> In the scenario we were looking at (operation from core B that leads to a stall on core A) you would always have a cap to the A thread, so in that sense you're causing time delay on behalf of A.

Ah, I understand what you mean now. Yes, in that sense the old behaviour was indeed fairer.

> One could also make the argument that the B thread should be charged for that time, but it seems strange to me to account time on one core to another, because that core doesn't really get more or less time to run.

It's a way to limit the global lock hold time for a task. It doesn't help the other core directly to get its time back, but it does limit future stealing. I can imagine that this concept would make the theoretical side unnecessarily more complicated though. But MCS on SMP with passive threads has this local/global time duality already though.

> This argument hinges on the fact that you must have presented authority to thread A for a stall of A to happen. If that is wrong, then it really is a random thread, and things are worse.

I think that assumption is correct.

> All of this is a bit theoretical, because if you're doing real-time you really shouldn't be messing with threads while they are running on a different core if those have important deadlines. 

Agreed.

> If we could set it up such that these can't happen (be modified from another core), then fully async could work, but I don't really see how.

To handle cross-core operations, it could send an async-IPI request for the operation to another core. Instead of waiting, the initiating task can be blocked and some other task can run instead. This would behave a lot like a normal blocking IPC call. This way all operations are always core-local and all IPIs are async.

As far as I can tell, all IPI calls are triggered by a syscall. IIRC, there may be a corner case with scheduling donation though. Deletion or revocation of untyped regions or cnodes may cause mass deletions on possibly multiple cores. Those could all be run once on each core, where each core only deals with core-local data. These calls are already interruptible, so it shouldn't be too hard.

(It would be nice if a core could run another schedulable thread instead of waiting for the global kernel lock, but I don't see how without introducing per-core locks and partitioning all data.)

src/smp/ipi.c
```diff
@@ -65,7 +69,10 @@ void ipiStallCoreCallback(bool_t irqPath)
          * handle the pending interrupt. Its valid as interrups are async events! */
         SCHED_ENQUEUE_CURRENT_TCB;
         switchToIdleThread();
-
+#ifdef CONFIG_KERNEL_MCS
+        commitTime();
+        NODE_STATE(ksCurSC) = NODE_STATE(ksIdleThread)->tcbSchedContext;
```
kent-mcleod, line 74: Is there a reason these can't be replaced with switchSchedContext() which should do the same thing after a call to switchToIdleThread() but is more likely stay up to date with any edge cases?

kent-mcleod: One reason is that we need to ensure that ksReprogram is left as false because we aren't programming a timeout (assuming that the core stall will complete in << KERNEL_WCET). switchSchedContext sets ksReprogram to true and so we would need to set it to false right after.

kent-mcleod (approved):
I agree that while it's unfortunate that the time the idle spent running on the remote core isn't billed to any SC, there is currently no better way to safely reschedule if the SC is deleted.

lsf37:
> I agree that while it's unfortunate that the time the idle spent running on the remote core isn't billed to any SC, there is currently no better way to safely reschedule if the SC is deleted.

Ok, I agree as well. Let's merge this as is and add an issue so we don't forget about the time accounting wrinkle. I think it would be good to document some assumptions under which scenarios MCS for SMP makes sense and is expected to be reliable.

lsf37:
> > One could also make the argument that the B thread should be charged for that time, but it seems strange to me to account time on one core to another, because that core doesn't really get more or less time to run.

> It's a way to limit the global lock hold time for a task. It doesn't help the other core directly to get its time back, but it does limit future stealing.

I get it now, yes that does make sense.

> I can imagine that this concept would make the theoretical side unnecessarily more complicated though. But MCS on SMP with passive threads has this local/global time duality already though.

Yes.

> To handle cross-core operations, it could send an async-IPI request for the operation to another core. Instead of waiting, the initiating task can be blocked and some other task can run instead. This would behave a lot like a normal blocking IPC call. This way all operations are always core-local and all IPIs are async.

Right, that kind of async. Yes that would be a lot more possible.

> As far as I can tell, all IPI calls are triggered by a syscall. IIRC, there may be a corner case with scheduling donation though. Deletion or revocation of untyped regions or cnodes may cause mass deletions on possibly multiple cores. Those could all be run once on each core, where each core only deals with core-local data. These calls are already interruptible, so it shouldn't be too hard.

The accounting of what is on which core might get a bit tricky. Revocation changes the pointer structure it is traversing, so you'd either get race conditions or synchronisation if you do it independently on multiple cores. Not that it's not worth looking into, but I imagine it would be a reasonably invasive change.

PR #637, opened 2021-11-09
qemu-arm-virt: fix in cmake for STRIP

ruizjuanpablo:
Check if QEMU_OUTPUT_MESSAGE is empty before the STRIP
If no error in the previous qemu test command, the output
message is empty (QEMU_OUTPUT_MESSAGE), then the STRIP
will fail on empty input. Then, if it is empty do not
run STRIP.

Signed-off-by: Juan Pablo Ruiz <juanpablo@ssrc.tii.ae>


kent-mcleod (approved):
Thanks!

PR #640, opened 2021-11-11
enable more test platforms

lsf37:
Tests should now re-run automatically on boot failure, so we can hopefully add some of the more skittish boards.


lsf37:
Unfortunately the boards are still not reliable enough yet.

PR #642, opened 2021-11-14
python: use KernelSel4Arch instead of KernelArch

axel-h:
Also make 'arch' a required parameter to avoid implicit assumptions.

This PR make a change in the ELF-Loader script necessary, too. See https://github.com/seL4/seL4_tools/pull/130

Test with seL4/seL4_tools#130


kent-mcleod (approved):

lsf37:
This looks good, let me run all build variations to check if we haven't missed a corner case, but I don't think we did.

lsf37 (approved):
Ok, looks like everything has passed, so will now merge both PRs

PR #645, opened 2021-11-16
consolidate arm core specific constants

axel-h:


libsel4/sel4_plat_include/exynos5/sel4/plat/api/constants.h
```diff
@@ -7,14 +7,20 @@
...
+/* This is used for multiple SOCs:
+ * - Exynos5250 (A15)
+ * - Exynos5410 (A7/A15)
+ * - Exynos5422 (A7/A15)
+ * Since the A7 was designed as big.LITTLE companion for the A15, their settings
+ * are identical.
+ */
+#if defined(CONFIG_ARM_CORTEX_A7)
+#include <sel4/arch/constants-cortex-a7.h>
+#el#if defined(CONFIG_ARM_CORTEX_A15)
```
axel-h, line 19: need to fix this
```suggestion
#elif defined(CONFIG_ARM_CORTEX_A15)
```

axel-h:
anybody has an idea why `TX2_debug_gcc_64 FAILED` failes?

kent-mcleod (changes requested):
Thanks for this!  Are you able to use `_` instead of `-` in the new filenames to be consistent with other file in this interface.

kent-mcleod:
> anybody has an idea why `TX2_debug_gcc_64 FAILED` failes?

@lsf37 looks like the machine runner has lost permissions to the destination of the binary copy: 
`  Couldn't change mode of /tftpboot/tx2a/seL4-image: [Errno 1] Operation not permitted: '/tftpboot/tx2a/seL4-image'`

lsf37:
> > anybody has an idea why `TX2_debug_gcc_64 FAILED` failes?
> 
> @lsf37 looks like the machine runner has lost permissions to the destination of the binary copy: ` Couldn't change mode of /tftpboot/tx2a/seL4-image: [Errno 1] Operation not permitted: '/tftpboot/tx2a/seL4-image'`

This bit is normal, and should really be fixed, it's happening for all runs (the user running this doesn't have enough permissions for changing permissions of the image, but doesn't really need to change them). The thing that went wrong in this one is

```
Retrieving file: /tx2a/pxelinux.cfg/default-arm
ethernet@2490000 Waiting for PHY auto negotiation to complete........ done
*** ERROR: `ipaddr' not set
 ```

which looks like a network fluke. I'll restart the test. @wom-bat this looks like a DHCP failure (but did work fine for all the other boards and also the other test runs of `tx2`).

axel-h:
> Are you able to use `_` instead of `-` in the new filenames to be consistent with other file in this interface.

Changed to underscores.

lsf37 (approved):

axel-h:
The changes affect ARM only, so I don't think the failing of HIFIVE_verification_gcc_64 is related.

kent-mcleod (approved):

axel-h:
HW Run `IMX8MM_EVK, armv8a, gcc` and `IMX8MM_EVK, armv8a, clang` fail now, see the board is stuck?

lsf37:
Yes, someone was holding the lock for that board for a few days and blocked all the tests (you can see who in the logs ;-)). Seems like nobody but me is policing this, so it breaks when I'm on leave..

PR #646, opened 2021-11-16
RISC-V: align with ARM, move rv32 CSR access wrappers

axel-h:
- move C code out of hardware.h

postponed for another PR
- always use the second value
- handle roll over from MAX to 0 properly (and halt on roll over)
- simplify expression to build a uint64_t
- fix comment to mention cycles and not timer


axel-h:
Seems the URL check failure is a random internet hickup.

include/arch/riscv/arch/32/mode/hardware.h
```diff
@@ -69,40 +69,38 @@
...
         "rdcycle  %1\n"
         "rdcycleh %2\n"
-        : "=r"(nH1), "=r"(nL), "=r"(nH2));
-    if (nH1 != nH2) {
+        : "=r"(nH_prev), "=r"(nL), "=r"(nH));
+    if (nH != nH_prev) {
         /* Ensure that the time is correct if there is a rollover in the
          * high bits between reading the low and high bits. */
         asm volatile(
             "rdcycle  %0\n"
```
axel-h, line 100: Is there actually a strong reason to read the low value again and not just use `0` here? Rationale would be that there is some jitter on the low value anyway, but 0 would be closer to the actual point in time where we wanted to read things in the first place.

include/arch/riscv/arch/32/mode/hardware.h
```diff
@@ -75,7 +75,7 @@ static inline uint64_t riscv_read_time(void)
         "rdtime  %1\n"
         "rdtimeh %2\n"
         : "=r"(nH1), "=r"(nL), "=r"(nH2));
-    if (nH1 < nH2) {
+    if (nH1 != nH2) {
```
kent-mcleod: A 64-bit rollover of the kernel timer is not allowed to happen. It would be better to add a condition that halts if the nH1 > nH2 condition is seen.

include/arch/riscv/arch/32/mode/hardware.h
kent-mcleod: It seems like a reasonable optimization when you know that the value of `rdcycle`must have contained 0 at some time between the two reads of `rdcycleh`.

include/arch/riscv/arch/32/mode/hardware.h
axel-h: I wonder if this is something that should not be checked here, but at the higher layers in a generic way. The timer design for each platforms should be, that this does never roll over practically. This function just returns whatever is in the registers.

axel-h:
I've added a check in `getCurrentTime()` for RISC-V now. But I wonder if we should have this for all platforms/architectures, as the same issue can also happen there. We could move `getCurrentTime()` out of the architecture code and make it do the check. Platforms/Architectures then just have to provide a function that retrieves the values, but don't have to reason much about it,

axel-h:
I've factored out the timer sanity checks to https://github.com/seL4/seL4/pull/657

axel-h:
Adding one commit that moves the CSR access code from `hardware.h` to `machine.h`, because having C code in `hardware.h` generated some include issues. It seem better to just have defines there and have the C code in `machine.h`

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -8,4 +8,42 @@
...
+        "rdcycleh %0\n"
+        "rdcycle  %1\n"
+        "rdcycleh %2\n"
+        : "=r"(nH_prev), "=r"(nL), "=r"(nH)
+        : /* no input */
+        : "memory");
+    if (nH_prev != nH) {
+        /* Ensure that the time is correct if there is a rollover in the
+         * high bits between reading the low and high bits. */
+        asm volatile("rdtime  %0\n" : "=r"(nL) :: "memory");
```
kent-mcleod, line 46: This should be `rdcycle` instead of `rdtime`.

axel-h: thanks, fixed

include/arch/riscv/arch/32/mode/hardware.h
```diff
@@ -62,46 +62,3 @@
 
 #define LOAD  lw
 #define STORE sw
-
```
pingerino, line 65: this commit seems to shuffle code around with a new convention that we don't currently follow in the kernel?

include/arch/riscv/arch/32/mode/hardware.h
```diff
@@ -69,40 +69,38 @@
 
 static inline uint64_t riscv_read_time(void)
 {
-    uint32_t nH1, nL, nH2;
+    uint32_t nH_prev, nH, nL;
```
pingerino: what is the reasoning behind the changes in this commit?

pingerino (changes requested):

include/arch/riscv/arch/32/mode/hardware.h
axel-h: Putting C code in hardware.h does not scale too well in my dev branches, especially when this starts pulling in more include dependencies. Thus I'm moving this to machine.h, there almost everything can be included without problems. Also, the assembler never sees this file, so it does no need defines.

include/arch/riscv/arch/32/mode/hardware.h
axel-h: The variable names are more intuitive. Can also keep the old names.

pingerino: Changing the names makes the code new for anyone who understood it before. 1, 2 vs prev is not really intuitive, just a personal preference. To avoid code churn we should avoid renames like this.

include/arch/riscv/arch/32/mode/hardware.h
pingerino: Did you consider putting new C code in a different file,  and leaving the existing code as is?

axel-h: Yes, but since the ARM port also has the special register access wrapper code in `machine.h` also, why not follow this on RISC-V also?

pingerino: if it's following an existing convention, the change makes sense.

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -6,6 +6,42 @@
...
+static inline uint64_t riscv_read_time(void)
+{
+    register word_t nH_prev, nH, nL;
+    asm volatile(
+        "rdtimeh %0\n"
+        "rdtime  %1\n"
+        "rdtimeh %2\n"
+        : "=r"(nH_prev), "=r"(nL), "=r"(nH)
+        : /* no input */
+        : "memory");
```
kent-mcleod, line 22: These "memory" clobbers shouldn't be inserted as part of the move

kent-mcleod: (They shouldn't be as the inline asm doesn't touch memory contents)

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -6,6 +6,42 @@
...
+        "rdcycleh %0\n"
+        "rdcycle  %1\n"
+        "rdcycleh %2\n"
+        : "=r"(nH_prev), "=r"(nL), "=r"(nH)
+        : /* no input */
+        : "memory");
+    if (nH_prev != nH) {
+        /* Ensure that the cycles are correct if there is a rollover in the
+         * high bits between reading the low and high bits. */
+        asm volatile("rdcycle  %0\n" : "=r"(nL) :: "memory");
```
pingerino, line 44: this isn't a clean move - you've changed the code - can you please seperate the move into a different commit? it's nearly impossible to review if code is moved and changed in one commit

axel-h:
I've removed all the commits that change a thing and just keep one commit here that moved the function from `hardware.h` to `machine.h`. This is the most fundamental thing to change. For the other changes I will create a new PR then.

include/arch/riscv/arch/32/mode/hardware.h
freddy123098: But if it does not like I am trying to see the changes made by this person but I can not see any changes made by him/her yet so now what should take this as?

freddy123098 (changes requested):
requesting changes.

freddy123098:
Proofs / C Proofs (ARM_HYP, CRefine) (pull_request_target) Failing after 14s — C Proofs (ARM_HYP, CRefine)

The AWS has failed and needs to be restarted. 

And the others that have failed also there AWS has to restarted. 


Thank you.

axel-h:
Can we merge this one commit now to better align the RISCV and ARM port?

kent-mcleod (approved):

PR #648, opened 2021-11-16
boot: pass a slot_region when setting up untypeds

axel-h:
Inlining `provide_untyped_cap()` into `create_untypeds_for_region()` makes the code easier to understand. This also allows printing better error and debug messages to troubleshoot memory issues during porting.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -645,7 +645,7 @@ BOOT_CODE static bool_t pptr_in_kernel_window(pptr_t pptr)
  */
 BOOT_CODE static bool_t provide_untyped_cap(
     cap_t      root_cnode_cap,
-    bool_t     device_memory,
+    bool_t     is_device_memory,
     pptr_t     pptr,
     word_t     size_bits,
     seL4_SlotPos first_untyped_slot
@@ -670,7 +670,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
     }
 
     /* All cap ptrs apart from device untypeds must be in the kernel window. */
-    if (!device_memory && !pptr_in_kernel_window(pptr)) {
+    if (!is_device_memory && !pptr_in_kernel_window(pptr)) {
         printf("Kernel init: Non-device untyped pptr %p outside kernel window\n",
                (void *)pptr);
         return false;
@@ -679,7 +679,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
     /* Check that the end of the region is also in the kernel window, so we don't
        need to assume that the kernel window is aligned up to potentially
        seL4_MaxUntypedBits. */
-    if (!device_memory && !pptr_in_kernel_window(pptr + MASK(size_bits))) {
+    if (!is_device_memory && !pptr_in_kernel_window(pptr + MASK(size_bits))) {
         printf("Kernel init: End of non-device untyped at %p outside kernel window (size %"SEL4_PRIu_word")\n",
                (void *)pptr, size_bits);
         return false;
@@ -690,11 +690,11 @@ BOOT_CODE static bool_t provide_untyped_cap(
         ndks_boot.bi_frame->untypedList[i] = (seL4_UntypedDesc) {
             .paddr    = pptr_to_paddr((void *)pptr),
             .sizeBits = size_bits,
-            .isDevice = device_memory,
+            .isDevice = is_device_memory,
             .padding  = {0}
         };
         ut_cap = cap_untyped_cap_new(MAX_FREE_INDEX(size_bits),
-                                     device_memory, size_bits, pptr);
+                                     is_device_memory, size_bits, pptr);
         ret = provide_cap(root_cnode_cap, ut_cap);
     } else {
         printf("Kernel init: Too many untyped regions for boot info\n");
@@ -720,7 +720,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
  */
 BOOT_CODE static bool_t create_untypeds_for_region(
     cap_t      root_cnode_cap,
-    bool_t     device_memory,
+    bool_t     is_device_memory,
     region_t   reg,
     seL4_SlotPos first_untyped_slot
 )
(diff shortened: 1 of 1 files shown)


lsf37:
I'm not feeling too good about this one. The comment is good and should stay in, but I don't think the inlining improves readability (it does make it worse for me). In terms for verification I would definitely prefer a separate function and small loop bodies.

src/kernel/boot.c
```diff
@@ -623,12 +597,49 @@ BOOT_CODE static bool_t create_untypeds_for_region(
...
+             * does not use all system resources.
+             */
+            word_t idx = ndks_boot.slot_pos_cur - first_untyped_slot;
+            if (idx >= CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS) {
+                printf("WARNING: can't provide cap for %"SEL4_PRIx_word"/2^%u"
+                       " and beyond, CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS (%d)"
+                       " exceeded\n",
+                       reg.start, size_bits,
+                       (int)CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS);
+                assert(0); /* Make this fatal for debug builds only. */
```
lsf37, line 615: Not sure it should be fatal for debug either. In fact, it might be especially in debug mode where you want the warning but otherwise want things to proceed while you are bringing up something new.

lsf37: It might make sense to add to the warning that we're proceeding anyway and that this means the memory will not be available to user space.

axel-h: I'm ok with removing this assert if nobody thing that it should be kept. It's just there from my local patch set where this was useful to catch some issues early. If we remove this assert, the warning can show up multiple times when the configuration is tool small - which can get a bit annoying if this configuration is intended.

axel-h:
> I'm not feeling too good about this one. The comment is good and should stay in, but I don't think the inlining improves readability (it does make it worse for me). In terms for verification I would definitely prefer a separate function and small loop bodies.

I see your point. Inlining makes this a bit more porting/experimenting friendly, as you have more context available there for debug messages. And you can abort the whole loop on the first warning, as all further iteration will produce the same warning. The other option would ne making `provide_untyped_cap()` return more then just true/false, os we know that we can stop.

src/kernel/boot.c
axel-h: remove the assert

lsf37:
> I see your point. Inlining makes this a bit more porting/experimenting friendly, as you have more context available there for debug messages. And you can abort the whole loop on the first warning, as all further iteration will produce the same warning. The other option would be making `provide_untyped_cap()` return more then just true/false, as we know that we can stop.

I'd be Ok with making `provide_untyped_cap()` return an `int` or similar instead of `bool` to indicate wether we should break out of the loop to reduce the warning messages.

axel-h:
I was wondering, it the loop not supposed to terminate actually, because not being able to add something in the boot info frame list should not prevent caps to be added to the root C-Node as long as is has space? But I'm not sure if there is a practical use case for this, so keeping things in sync here might make more sense. Are there tools (CapDL?) that don't care about the boot info frame content and use the root C-Node slots directly because they know what to find in there?

kent-mcleod:
I'd expect that the kernel would stop copying caps once it reached the limit defined by KernelMaxNumBootinfoUntypedCaps.  And that there should probably be a static compilation error if KernelMaxNumBootinfoUntypedCaps is larger than what could fit inside the seL4_BootInfo_t struct.  A root task shouldn't expect that the kernel starts using more slots in it's initial CNode than what is the configured max.

kent-mcleod:
> I'd be Ok with making `provide_untyped_cap()` return an `int` or similar instead of `bool` to indicate wether we should break out of the loop to reduce the warning messages.

It looks like you can just change the return value to false when `(i >= CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS)` and this will terminate the loop after printing the warning message only one time.

axel-h:
restored the old behavior

axel-h:
- Leaving the function `provide_untyped_cap()` in and make it return an error code instead of just a boolean, so the caller can better differentiate there error cases.
- Created https://github.com/seL4/seL4/issues/716 do discuss `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` and behavior when this is exceeded.
- This PR is now on top of 
  - https://github.com/seL4/seL4/pull/715
  - https://github.com/seL4/seL4/pull/719

src/kernel/boot.c
```diff
@@ -580,19 +580,22 @@ BOOT_CODE static bool_t provide_untyped_cap(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
     pptr_t     pptr,
-    word_t     size_bits)
+    word_t     size_bits,
+    seL4_SlotRegion *slot_region
```
kent-mcleod, line 584: It's generally preferred to use global variables where possible in these functions. Is there a reason that this needs to be changed?

kent-mcleod: In this case the proofs are fine with the change.  Also I forgot that slot_region is just a fixed size struct and not an array.

src/kernel/boot.c
```diff
@@ -577,39 +577,50 @@ BOOT_CODE void init_core_state(tcb_t *scheduler_action)
...
+
+    /* This increments ndks_boot.slot_pos_cur if the cap can be provided. */
+    if (!provide_cap(root_cnode_cap,
+                     cap_untyped_cap_new(MAX_FREE_INDEX(size_bits),
+                                         is_device_memory, size_bits, pptr))) {
+        return -1;
+    }
+
+    /* Update boot info. */
+    ndks_boot.bi_frame->untyped.end = ndks_boot.slot_pos_cur;
```
kent-mcleod, line 615: I don't think adding a slot_region parameter to this function makes sense if internally the function requires that it can only be the ndks_boot.bi_frame->untyped region.  Also because this function is writing into the `ndks_boot.bi_frame->untypedList` array which is described by one specific slot_region.

axel-h: Yes, this should be `slot_region->end = ndks_boot.slot_pos_cur`.
The update of `ndks_boot.bi_frame->untypedList`remains a hard-coded thing - unless this also becomes a parameter. However, since I made `create_untypeds_for_region()` take a `slot_region` parameter, it should pass on this to `provide_untyped_cap()` and not fall back to something hard-coed again. Furthermore, the long term thing I'm exploring is passing the `slot_region` to `provide_cap()` also (see https://github.com/axel-h/seL4/pull/47)

lsf37:
This is too much code movement in unverified code to be certain that it doesn't introduce new bugs and the PR has been on draft for 2 years. Happy for you to re-open the PR when the work is ready and you think there is a strong need for it.

PR #651, opened 2021-11-17
add helper functions to get kernel image location

axel-h:
The helper functions hides the details about the ELF symbols and return the physical address of the kernel image. This removes some hard dependencies of the code on ELF images and works towards allowing raw kernel images also.

This is on top of https://github.com/seL4/seL4/pull/728


kent-mcleod:
> The helper functions hide the details about the ELF symbols and return the physical address of the kernel image. This removes some hard dependencies of the code on ELF images and works towards allowing raw kernel images also.

What do you mean by removing dependencies on ELF images? The kernel doesn't use any ELF headers to load itself (On x86 it does for user images, but this PR doesn't remove this), the `ki_end` and `ki_boot_end` symbols would be implemented by the linker regardless of the executable output format wouldn't they?

Do you need to remove the ki_end and ki_boot_end symbols for allowing raw kernel images or would they still exist?  What future changes are these changes preparing for?

kent-mcleod:
> Do you need to remove the ki_end and ki_boot_end symbols for allowing raw kernel images or would they still exist? What future changes are these changes preparing for?

I've just seen that https://github.com/seL4/seL4/pull/652/files builds on top of this PR currently

kent-mcleod:
> I've just seen that https://github.com/seL4/seL4/pull/652/files builds on top of this PR currently

Does this mean that the motivation for these changes are to be able to remove the `boot_mem_reuse_reg` parameter from create_untypeds?

axel-h:
> I've just seen that https://github.com/seL4/seL4/pull/652/files builds on top of this PR currently.
> Does this mean that the motivation for these changes are to be able to remove the boot_mem_reuse_reg parameter from create_untypeds?

yes, this is intention eventually. My feeling is that passing this parameter down the call stack does not really gain much, it just create unnecessary dependencies.

include/machine.h
```diff
@@ -55,5 +64,23 @@ static inline p_region_t CONST pptr_to_paddr_reg(const region_t pptr_reg)
...
+    };
+}
+
+/* Returns the physical region of the kernel image. */
+static inline p_region_t get_p_reg_kernel_img(void)
+{
+    return (p_region_t) {
+        .start = kpptr_to_paddr((const void *)KERNEL_ELF_BASE),
+        .end   = kpptr_to_paddr(ki_end)
+    };
```
kent-mcleod: It seems odd to try and translate pptr's to paddrs, and then have 4 out of the 5 callers translate back to pptr again. pptr is the natural region type as used by everything else in the kernel.

kent-mcleod:
I think this would be better to review as part of https://github.com/seL4/seL4/pull/652 as a single PR as the changes in this PR are to support the changes in the otehr PR.

PR #652, opened 2021-11-18
boot: simplify boot code recycling

axel-h:
This build on top of PR https://github.com/seL4/seL4/pull/651
With the changes the we can simplify boot code recycling and make this quite generic, there is no need to pass a parameter though the layers on all architectures.


axel-h:
@lsf37 how bad is this for the proofs?

kent-mcleod:
> With the changes the we can simplify boot code recycling and make this quite generic, there is no need to pass a parameter though the layers on all architectures.

Is there another motivation for this other than exchanging the parameter with `get_p_reg_kernel_img_boot()`?

include/machine.h
```diff
@@ -47,13 +56,31 @@ static inline region_t CONST paddr_to_pptr_reg(const p_region_t p_reg)
...
 }
 
+/* Returns the physical region of the kernel image. */
+static inline p_region_t get_p_reg_kernel_img(void)
+{
+    return (p_region_t) {
+        .start = kpptr_to_paddr((const void *)KERNEL_ELF_BASE),
+        .end   = kpptr_to_paddr(ki_end)
+    };
+}
```
kent-mcleod, line 84: Can these just return `region_t` types? Then if needed they can be wrapped in `pptr_to_paddr_reg` at any call sights? The kernel mostly deals with region_t by default.

axel-h:
rebased and fixed auto-merge corruption

include/machine.h
axel-h: It look a bit odd here, but the reason for using physical addresses here is, that the whole free memory setup during boot is simpler if this all works on the physical addresses eventually. And that is also much nicer to debug then for ports or memory changes. These changes are in a draft in my fork at https://github.com/axel-h/seL4/pull/100

src/kernel/boot.c
```diff
@@ -774,7 +793,18 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap,
...
+     * boot process. We can create UT objects for these frames, so the memory
+     * can be reused.
+     * ToDo: Actually, the region returned by get_p_reg_kernel_img_boot() could
+     *       be made part of ndks_boot.freemem[] also, then no special handling
+     *       is needed at all. Currently, this is non-trivial, because the
+     *       regions in this array are expected to be be properly ordered and
+     *       mutually non-overlapping. Furthermore, this can only be added to
+     *       ndks_boot.freemem[] after the rootserver object have been created
+     *       in there, otherwise this could overwrite boot code.
+     */
```
kent-mcleod, line 806: Please can you move this into an issue and out of the source code.

include/machine.h
kent-mcleod: Ok

kent-mcleod (approved):

src/kernel/boot.c
axel-h: Removed from the code. Note sure if this is worth creating an issue, though. It was just a note for something to explore one day, but there seem not much gain in this.

PR #653, opened 2021-11-18
boot: created untypeds up to PADDR_TOP only

axel-h:
Provide untyped device capabilities for the remaining physical address space that can be accessed via the kernel window. This could be less than the physical address space implemented by the platform or defined by the architecture (`CONFIG_PADDR_USER_DEVICE_TOP`). Due to the kernel design, only the part of the physical address space is accessible that fits into the kernel window.

This limit used to exists in the past, but the commit [fc1674](https://github.com/seL4/seL4/commit/fc167441f02ac35e0654047a323936e563bcff79) removed this, because the comment there was invalid thus not justification.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -102,9 +102,6 @@ BOOT_CODE static bool_t arch_init_freemem(p_region_t ui_p_reg,
             reserved[index] = mode_reserved_region[0];
             index++;
         }
-
-        /* Reserve the ui_p_reg region still so it doesn't get turned into device UT. */
-        reserve_region(ui_p_reg);
     }
 
     /* avail_p_regs comes from the auto-generated code */
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -654,9 +654,17 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap,
         start = ndks_boot.reserved[i].end;
     }
 
-    if (start < CONFIG_PADDR_USER_DEVICE_TOP) {
+    if (start < PADDR_TOP) {
+        /* Provide untyped device capabilities for the remaining physical
+         * address space that can be accessed via the kernel window. This could
+         * be less than the physical address space implemented by the platform
+         * or defined by the architecture (CONFIG_PADDR_USER_DEVICE_TOP). Due to
+         * the kernel design, only the part of the physical address space is
+         * accessible that fits into the kernel window.
+         */
         region_t reg = paddr_to_pptr_reg((p_region_t) {
-            start, CONFIG_PADDR_USER_DEVICE_TOP
+            .start = start,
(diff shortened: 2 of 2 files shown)


axel-h:
Maybe this change is not necessary and we leave it this was. It seems much easier to rework the boot code to use the physical address view for all memory reservations and then make  `create_untypeds_for_region()` a `create_untypeds_for_phys_region()` as the one and only place that does the conversion to the kernel window view and caps. That avoid a lot of back and forth conversion and range checks then also. A lot of boot code does not have to care about the kernel runtime mapping view then

kent-mcleod:
> This limit used to exists in the past, but the commit fc1674 removed this, because the comment there was invalid thus not justification.

From my understanding of the changes in this PR, this isn't true.  The kernel has always created device untypeds outside of the kernel window.  Not supporting this would mean only being able to access 256MiB of the physical address space on 32bit platforms.  fc1674 was removing a work around that was trying to prevent an underlying issue that was resolved in https://github.com/seL4/seL4/pull/590 - that device untypeds for physical addresses outside of the kernel physical window mapping had KPPtr addresses that were non-canonical (Invalid virtual addresses and wouldn't fit into bitfields that were only 48bit long). 

On almost every configuration, CONFIG_PADDR_USER_DEVICE_TOP is larger than PADDR_TOP and so this change would prevent their creations.  On 64bit architectures this probably wouldn't cause any issue because there don't appear to be physical addresses that get used that high, but on 32-bit architectures this would limit access to almost all physical devices.

lsf37:
> It seems much easier to rework the boot code to use the physical address view for all memory reservations and then make `create_untypeds_for_region()` a `create_untypeds_for_phys_region()` as the one and only place that does the conversion to the kernel window view and caps. That avoid a lot of back and forth conversion and range checks then also. A lot of boot code does not have to care about the kernel runtime mapping view then

I have the feeling this came up before, but can't quite pinpoint where. In any case, `create_untypeds_for_region` is too late to switch to a kernel-virtual view. As soon as the first kernel object (incl first cap) exists, all addresses the kernel deals with should be kernel virtual. IIRC, the creation of untypeds is very late in the kernel init process and a good deal past that point.

The reason for that is that all predicates we have only talk about the kernel virtual view. We could do back-and-forth calculations there, but those would be a factor of 5-10 more conversions than what we are seeing in the code. 

In fact, I don't understand why there needs to be any calculation at all on physical addresses. I do understand that printing physical addresses makes sense, and I don't oppose that (conversion there could easily be factored out into a printing function), but otherwise it seems to me that the kernel should compute on kernel addresses. I do agree that using a mix is annoying and error-prone, but if we choose a view we should choose the one that the rest of the kernel uses.

axel-h:
Closing this based on the comment from @kent-mcleod, I will file another PR that adds a comment in the code to explicitly state this is the desired behavior.

PR #654, opened 2021-11-19
generate test hardware matrix

lsf37:
This uses the corresponding action for seL4/ci-actions#170 to generate the matrix for sel4test hardware runs centrally instead of replicating it over multiple workflows.


lsf37:
Testing successful, this should make switching on/off additional boards much easier.

Xaphiosis (approved):

PR #655, opened 2021-11-20
riscv: provide functions for CSR access

axel-h:
Provide generic CSR access function macros that can be extended easily for other CSRs.
Note that this PR is on top of https://github.com/seL4/seL4/pull/646

### DIFF
--- a/include/arch/riscv/arch/32/mode/machine.h
+++ b/include/arch/riscv/arch/32/mode/machine.h
@@ -18,7 +18,7 @@ static inline uint64_t riscv_read_time(void)
         "rdtime  %1\n"
         "rdtimeh %2\n"
         : "=r"(nH1), "=r"(nL), "=r"(nH2));
-    if (nH1 < nH2) {
+    if (nH1 != nH2) {
         /* Ensure that the time is correct if there is a rollover in the
          * high bits between reading the low and high bits. */
         asm volatile("rdtime  %0\n" : "=r"(nL));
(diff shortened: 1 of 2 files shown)


kent-mcleod:
This doesn't seem like a good change to me.  It's more convenient that the assembler understands `rdcycle` and we don't need to copy down all of the CSR codes for the right instruction.  Writing inline asm is already encapsulated via the helper functions.

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -12,7 +12,7 @@
 
 static inline uint64_t riscv_read_time(void)
 {
-    uint32_t nH1, nL, nH2;
+    register word_t nH1, nL, nH2;
```
kent-mcleod, line 15: The kernel doesn't use `register` keyword anywhere.  Also why do you wish to switch uint32_t to word_t as they're the same type in this context.

axel-h: - I've been digging a bit deeper into the `register` keyword and the history.Seems nowadays it's indeed better to avoid using it unless there is a really good reason. Remove this. 
- `word_t` seems semantically the better choice for case like this, that deal with register data. Furthermore, it's a pattern that is copy/paste safe. Technically it does not make any difference.

kent-mcleod:
@axel-h can you explain why you consider fc60cd79a6a602a6e3637880493f9f90868e7386 an improvement over the existing implementation? These CSR functions are already only 4 or 5 lines each and so there isn't really that much need to minimize any possible duplication. The added indirection layer seems to make the final result more complex than the original.

kent-mcleod (approved):

lsf37 (approved):
Happy with this from the verification side.

Issue #656, opened 2021-11-20
Generic check that the timer does not run backwards or rolls over


axel-h:
This is a follow-up from the discussion at https://github.com/seL4/seL4/pull/646 that the system timer must never roll over or run backwards. It seems there is no sanity check in the kernel about this. Thus I wonder if such a check should be implemented in a generic function `getCurrentTime()`, which would internally call a architecture/platform specific function then that is just providing the raw value. If the timer is seen to roll over or go backwards, the system is halted immediately. This would give all functions that use `getCurrentTime()` a strong and enforced guarantee that the timer is always incrementing monotonically and it never rolls over.

lsf37:
To be honest, that condition would be very handy to have in verification.

lsf37:
@kent-mcleod and @pingerino do you see any potential problems with that?

Indanz:
It seems wrong to keep track of the same state in two places, and updateTimestamp() is pretty much the only caller of getCurrentTime(). If you want a check that has no performance impact, do the check in updateTimestamp() against ksCurTime instead of some static variable hidden away in a function.

kent-mcleod:
`updateTimestamp()` also already has a similar check in place:
`    assert(NODE_STATE(ksCurTime) < MAX_RELEASE_TIME);`
So adding another assert that checks that the new time is greater than the previous time would fit here.

axel-h:
I've moved the PR to draft state. Any feed back is appreciated on the current proposal.

Indanz:
I think that if a check is done it should be in updateTimestamp(), and I agree that x86's setDeadline() should use ksCurTime. setDeadline() is only called by setNextInterrupt() which does:

``time_t next_interrupt = NODE_STATE(ksCurTime) +  refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount;``

Also, I remember an issue where the conclusion was that it's Elfloader's responsibility to make sure that the clock is initialised such that it won't overflow. If seL4 can't handle an overflowing clock, it really should init the clock itself such that it can't happen and not count on Elfloader doing it IMHO.

kent-mcleod:
> Also, I remember an issue where the conclusion was that it's Elfloader's responsibility to make sure that the clock is initialised such that it won't overflow. If seL4 can't handle an overflowing clock, it really should init the clock itself such that it can't happen and not count on Elfloader doing it IMHO.

[...]

PR #657, opened 2021-11-21
check monotonic increment of timer

axel-h:
Halt the system if the timer rolls over or moved backward.

This adresses issue https://github.com/seL4/seL4/issues/656
This is on top of PR https://github.com/seL4/seL4/pull/678

### DIFF
--- a/include/arch/x86/arch/machine/timer.h
+++ b/include/arch/x86/arch/machine/timer.h
@@ -13,6 +13,7 @@
 #ifdef CONFIG_KERNEL_MCS
 #include <mode/util.h>
 #include <arch/kernel/xapic.h>
+#include <machine/timer.h>
 
 static inline CONST time_t getKernelWcetUs(void)
 {
@@ -39,7 +40,7 @@ static inline void ackDeadlineIRQ(void)
 {
 }
 
-static inline ticks_t getCurrentTime(void)
+static inline ticks_t driver_getSystemTime(void)
 {
     return x86_rdtsc();
 }
@@ -60,8 +61,12 @@ static inline void setDeadline(ticks_t deadline)
     if (likely(x86KSapicRatio == 0)) {
         x86_wrmsr(IA32_TSC_DEADLINE_MSR, deadline);
     } else {
-        /* convert deadline from tscKhz to apic khz */
-        deadline -= getCurrentTime();
+        /* Convert deadline from tscKhz to apic khz and program it. Obtaining a
+         * fresh timestamp is needed, because ksCurTime could be too old here.
+         */
+        ticks_t now = getCurrentTime();
+        assert(now >= deadline);
+        deadline -= now;
         apic_write_reg(APIC_TIMER_COUNT, div64(deadline, x86KSapicRatio));
     }
 }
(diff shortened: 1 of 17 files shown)


src/machine/timer.c
```diff
@@ -0,0 +1,27 @@
...
+#include <config.h>
+
+#ifdef CONFIG_KERNEL_MCS
+
+#include <machine/timer.h>
+
+/* Read the current time from the timer. */
+ticks_t getCurrentTime(void)
+{
+    static uint64_t time_last = 0;
```
Indanz, line 16: Shouldn't this be a core local variable instead, in case a platform does not has cross-core synchronized timers?

And maybe make the if-check unlikely.

axel-h: Made the if-check unlikely and tracking things in core-local variables now.

include/arch/x86/arch/machine/timer.h
```diff
@@ -61,7 +62,9 @@ static inline void setDeadline(ticks_t deadline)
         x86_wrmsr(IA32_TSC_DEADLINE_MSR, deadline);
     } else {
         /* convert deadline from tscKhz to apic khz */
-        deadline -= getCurrentTime();
+        ticks_t now = getCurrentTime(); /* ToDO: just use ksCurTime? */
```
axel-h, line 65: should this use ksCurTime here instead of getting an update time stamp from the timer driver?

kent-mcleod: No, because this is programming a timeout in the future by a countdown timer.  So if it uses ksCurTime (which is less than getCurrentTime()), then the timeout will actually be  late because of the amount of time passed while the kernel was executing after calling getCurrentTime previously.

Indanz: setDeadline() is only called by setNextInterrupt() which does:

``time_t next_interrupt = NODE_STATE(ksCurTime) + refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount;``

and then immediately calls setDeadline(), so in practise you can as well use ksCurTime. Perhaps the API needs to be changed to make this explicit though.

kent-mcleod: setNextInterrupt() is called very near the end of the kernel's execution whereas ksCurTime is updated upon entry to the kernel (or at preemption points).  We need to ensure that an interrupt is fired so that the latest next time the kernel is entered ksCurTime is updated to a value that isn't greater than `NODE_STATE(ksCurTime) + refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount`.  If there has been `t` time since ksCurTime was updated at the point when we are programming the timer, then an interrupt programmed to be `refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount` in the future will be late by `t` amount. `t` could be as large as the kernel's WCET, which I believe would cause the interrupt to come in 1 WCET late.

Indanz: Thanks for the clear explanation.

kent-mcleod:
For similar reasons outlined in this comment: https://github.com/seL4/seL4/pull/752#issuecomment-1025056495, it's unlikely that these changes would be accepted until after MCS verification has completed.

axel-h:
Closing this

PR #658, opened 2021-11-22
Remove userError from seL4_ReplyRecv path

kent-mcleod:
Remove a userError that is present on mainline kernel but not on MCS.

seL4_ReplyRecv is often used in a loop to create an event handler. When
first entering this loop, or when handling signals from notifications,
there won't already be a pending caller blocked on the reply object. In
this case the kernel doesn't perform a reply transfer and continues with
the receive operation. This is a common operation and shouldn't result
in a userError being printed by the kernel each time the reply phase
ends up as a no-op.

### DIFF
--- a/libsel4/include/sel4/syscalls_master.h
+++ b/libsel4/include/sel4/syscalls_master.h
@@ -72,7 +72,9 @@ seL4_Call(seL4_CPtr dest, seL4_MessageInfo_t msgInfo);
 /**
  * @xmlonly <manual name="Reply" label="sel4_reply"/> @endxmlonly
  * @brief Perform a send to a one-off reply capability stored when
- *        the thread was last called
+ *        the thread was last called. Does nothing if there is no
+ *        reply capability which can happen if the blocked thread
+ *        was unblocked via an operation such as destroying it.
  *
  * @xmlonly
  * <docref>See <autoref label="sec:sys_reply"/></docref>
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -490,7 +490,7 @@ static void handleReply(void)
     }
 
     case cap_null_cap:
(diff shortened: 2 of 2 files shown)


gernotheiser:
I would have thought that the event look has an initialisation phase that consists of a receive-only operation. Considering a user error to do a ReplyRecv with an empty send intuitively makes more sense to me than just ignoring the send phase.

kent-mcleod:
> I would have thought that the event loop has an initialisation phase that consists of a receive-only operation. Considering a user error to do a ReplyRecv with an empty send intuitively makes more sense to me than just ignoring the send phase.

userError is a macro that generates a debug print.  This doesn't change the actual (non-debug) behavior of the kernel.  The kernel has always treated it as a non-error condition and moved onto the receive phase.  This code path is also only executed on non-mcs configs. MCS uses a different implementation that doesn't print an error and also continues to the receive phase.  (This came up because of needing to run user level code that was written for MCS on a master kernel which prints large amount of error debug messages even though the kernel doesn't actually return any error results).

lsf37 (approved):

axel-h:
I think this PR is a nice practical thing. The question seems to be, do we want to print this warning message to educate the developer that the code is a bit sloppy or do we say that this API usage is perfectly fine and well within the defined behavior. The comment could be a bit more verbose then, stating that we explicitly allow this. And maybe even the manual should explicitly state that this usage is allowed.

lsf37:
The comment already implies that it is explicitly allowed, so I think that part is fine. It would be good to add to the manual, though, that this behaviour is defined and intended. @kent-mcleod would you be happy to add that?

lsf37:
Thanks for adding that, happy to merge when the tests are through.

PR #659, opened 2021-11-22
print 64-bit integers on 32-bits systems fully

axel-h:
`PRIu64`/`PRIx64` were only printing the lower 32 bits of the value on 32-bit systems.

Found while experimenting with the system time for https://github.com/seL4/seL4/pull/657

### DIFF
--- a/src/machine/io.c
+++ b/src/machine/io.c
@@ -109,11 +109,6 @@ static inline bool_t isdigit(char c)
 
 #define FLAGMASK (ALT_FORM|ZERO_PAD|LEFT_ADJ|PAD_POS|MARK_POS|GROUPED)
 
-#define INTMAX_MAX INT32_MAX
-#define INT_MAX  0x7fffffff
-
-#define ULONG_MAX ((unsigned long)(-1))
-
 /* State machine to accept length modifiers + conversion specifiers.
  * Result is 0 on failure, or an argument type to pop on success.
  */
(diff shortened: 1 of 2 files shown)


kent-mcleod (approved):

src/machine/io.c
```diff
@@ -248,30 +243,49 @@ static const char xdigits[16] = {
...
-        *--s = '0' + x % 10;
-    }
-    for (y = x;           y; y /= 10) {
-        *--s = '0' + y % 10;
+    while (0 != x) {
+#if defined(CONFIG_ARCH_AARCH32) || defined(CONFIG_ARCH_RISCV32)
+        /* On 32-bit systems dividing a 64-bit number by 10 make the compiler
+         * call a helper functions. On ARM it's __aeabi_uldivmod(), on RISC-V
+         * it's __umoddi3(). These are not available in the kernel, thus the
+         * algorithm below is needed. Credits go to Hacker's Delight.
```
kent-mcleod, line 269: How was the old version of this code working in the past then? It seems like the compiler was previously able to optimize the code into something that didn't require calling helper functions?

axel-h: Looks to me as if it has never run into code that has to divided a 64-bit value by a arbitrary non-power-of-2 value. Also, integers where limited to 32-bit on 32-bit platforms. I've found all this because I was trying to print a 64-bit time value and it turned out only the lower 32-bit were printed. Furthermore, the division issue just "really" hows up when printing a number in decimal.

axel-h:
Added `CONFIG_ARCH_IA32` also.

PR #662, opened 2021-11-25
cmake: cleanup compiler and linker flag setup

axel-h:
- consolidate macro invocations
- improve comments
- improve readability by adding a few empty lines in between
- due to convenience this PR is on top of https://github.com/seL4/seL4/pull/668 (remove `HAVE_AUTOCONF`)

Motivated by rebasing https://github.com/seL4/seL4/pull/224, the rebased branch on top of this is https://github.com/axel-h/seL4/commits/riscv_gp_fixes then.


CMakeLists.txt
```diff
@@ -175,7 +138,19 @@ add_compile_options(
...
+string(APPEND CMAKE_EXE_LINKER_FLAGS " -Wl,--build-id=none -static -Wl,-n ")
+
+KernelCommonFlags(
+    ${KernelOptimisation}
+    -nostdlib
     -ffreestanding
+    -fno-pic
+    -fno-pie
+    -fno-stack-protector
+    -fno-asynchronous-unwind-tables
```
kent-mcleod, line 153: Is there a reason why these are now added to the linker flags too?

axel-h: Good point, that might not be needed for the linker. I'll remove this.

axel-h: Actually, when looking at the ld documentation at https://sourceware.org/binutils/docs-2.37/ld.html the following things also seem odd in the original file:
- `-fno-pie`is a gcc flag, the ld flag should be `-no-pie`
- `-fno-pic` is undocumented for ld, it's documented for gcc only 
- `-ffreestanding` is undocumented for ld, it's documented for gcc only

kent-mcleod: GCC (or clang) is still invoked for both compiling and linking phases.
CMake has 2 stages for executables with C:
- Compile stage when gcc is invoked to compile source files into object files, add_compile_options is applied here.
- Link stage when gcc is invoked to link object files and libraries into final programs. CMAKE_EXE_LINKER_FLAGS is applied here.
So I think -fno-pie should be translated by GCC into -no-pie when it calls the linker.  Likewise for the other options.

axel-h: When running it with `-v` it shows me this for the linker invocation:
```
/usr/lib/gcc-cross/aarch64-linux-gnu/8/collect2
    -plugin /usr/lib/gcc-cross/aarch64-linux-gnu/8/liblto_plugin.so
    -plugin-opt=/usr/lib/gcc-cross/aarch64-linux-gnu/8/lto-wrapper
    -plugin-opt=-fresolution=/tmp/ccMqglI5.res
    --sysroot=/host/build-qemu-arm-virt-a53-Debug-hello_world
    --build-id
    --hash-style=gnu
    -Bstatic
    -X
    -EL
    -maarch64linux
    --fix-cortex-a53-843419
    -o kernel/kernel.elf
    -L/usr/lib/gcc-cross/aarch64-linux-gnu/8
    -L/usr/lib/gcc-cross/aarch64-linux-gnu/8/../../../../aarch64-linux-gnu/lib/../lib
    -L/host/build-qemu-arm-virt-a53-Debug-hello_world/lib/../lib
    -L/usr/lib/gcc-cross/aarch64-linux-gnu/8/../../../../aarch64-linux-gnu/lib
    -L/host/build-qemu-arm-virt-a53-Debug-hello_world/lib
    --build-id=none
    -n
    -T /host/build-qemu-arm-virt-a53-Debug-hello_world/kernel/linker.lds_pp
    kernel/CMakeFiles/kernel.elf.dir/src/arch/arm/armv/armv8-a/64/machine_asm.S.obj
    kernel/CMakeFiles/kernel.elf.dir/src/arch/arm/64/head.S.obj
    kernel/CMakeFiles/kernel.elf.dir/src/arch/arm/64/traps.S.obj
    kernel/CMakeFiles/kernel.elf.dir/kernel_all.c.obj
COLLECT_GCC_OPTIONS=
    '-march=armv8-a'
    '-D __KERNEL_64__'
    '-O2'
    '-g'
    '-D NDEBUG'
    '-D __KERNEL_64__'
    '-static'
    '-v'
    '-O2'
    '-nostdlib'
    '-fno-pic'
    '-ffreestanding'
    '-D DEBUG'
    '-g'
    '-ggdb'
    '-mgeneral-regs-only'
    '-o kernel/kernel.elf'
    '-mlittle-endian'
    '-mabi=lp64'
```
So it seem not much is passed that is not passed as "Wl,...". Just `-Bstatic` come from "-static". The '-fno-pic' seems to be remvoed completely, though. For the rest to show up in the parameters it requires this a ` -Wl,xxx`.

kent-mcleod: Yea you're right.  From the manual:
```
       -fpie
       -fPIE
           These options are similar to -fpic and -fPIC, but the
           generated position-independent code can be only linked into
           executables.  Usually these options are used to compile code
           that will be linked using the -pie GCC option.

           -fpie and -fPIE both define the macros "__pie__" and
           "__PIE__".  The macros have the value 1 for -fpie and 2 for
           -fPIE.

```
-static overrides -pie to become -no-pie.  (there is an option -static-pie which is static linked position independent exe). I assumed that -ffreestanding also had an impact at link time (to not link the standard libraries) but it seems you still need -nostdlib and it only asserts not stdlibs for the compiler stage.

CMakeLists.txt
```diff
@@ -102,33 +106,88 @@ set(
...
+    -fno-common
+)
+
+# Linker parameters
+string(
+    APPEND
+    CMAKE_EXE_LINKER_FLAGS
+    " -static" # Implies "-no-pie" (and overrides "-pie"). The ld 2.37 docs say
+               # "-no-pie" is a linker option, but passing "-Wl,-no-pie" fails.
+    " -Wl,-nostdlib" # linker counterpart for "-ffreestanding"
```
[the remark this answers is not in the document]
axel-h: Thanks, I've improved the comments to state this and have removed `-Wl,-nostdlib`

kent-mcleod (approved):

PR #665, opened 2021-11-25
Make debug output for fault and dump a bit nicer

axel-h:
Improve the messages a bit to no talk about a "cap fault" just because there is no fault handler. Print a register snapshot besides the stack dump if  `CONFIG_DEBUG_BUILD` is enabled.

Fault printout for a null pointer dereferecne on RISC-V
```
## ==============================================================
## FAULT at PC=0x0 in thread 0xffffffc17fec7200 (rootserver)
## Cause: access fault on code at address 0x0, status 0x1
## Thread suspended, no userland fault handler (code 0, cap type 0)
## State snapshot:
##     x1/ra: 0x00000000'000150c6     x2/sp: 0x00000000'00143ea0
##     x3/gp: 0x00000000'00024800     x4/tp: 0x00000000'00144000
##     x5/t0: 0x00000000'00000000     x6/t1: 0x00000000'0001f5d8
##     x7/t2: 0x00000000'00000000     x8/s0: 0x00000000'00143ec0
##     x9/s1: 0x00000000'00000000    x10/a0: 0x00000000'00000000
##    x11/a1: 0x00000000'00000000    x12/a2: 0x00000000'00000000
##    x13/a3: 0x00000000'00000000    x14/a4: 0x00000000'00000000
##    x15/a5: 0x00000000'00000000    x16/a6: 0x00000000'00000000
##    x17/a7: 0xffffffff'fffffff7    x18/s2: 0x00000000'00000000
##    x19/s3: 0x00000000'00000000    x20/s4: 0x00000000'00000000
##    x21/s5: 0x00000000'00000000    x22/s6: 0x00000000'00000000
##    x23/s7: 0x00000000'00000000    x24/s8: 0x00000000'00000000
##    x25/s9: 0x00000000'00000000   x26/s10: 0x00000000'00000000
##   x27/s11: 0x00000000'00000000    x28/t3: 0x00000000'00000012
##    x29/t4: 0x00000000'20000000    x30/t5: 0x00000000'00000000
##    x31/t6: 0x00000000'00000000    scause: 0x00000000'0000000c
##   sstatus: 0x00000000'00000020   FaultIP: 0x00000000'00000000
##    NextIP: 0x00000000'00000000
##

Stack trace:
0x143ea0: 0x143ec0
0x143ea8: 0x101ea
0x143eb0: 0x143f00
0x143eb8: 0x14bcc
0x143ec0: 0x0
0x143ec8: 0x143f30
0x143ed0: 0x143f20
0x143ed8: 0x143f10
0x143ee0: 0x1
0x143ee8: 0x101d6
0x143ef0: 0x144000
0x143ef8: 0x14cf8
0x143f00: 0x0
0x143f08: 0x14a000
0x143f10: 0x20280
```

Scheduler dump ("RelQ" is only visible with MCS, "Core" only with SMP)
```
Dump of all TCBs (4):
  TCB              | State         | PC               | Prio | Core | RelQ | Name
  -----------------+---------------+------------------+------+------+------+----------------
  ffffffc100137600 | blocked/recv  | 1053a            |  255 |    0 |   no | main:fault_handler
  ffffffc100137200 | running       | 101f0            |  254 |    0 |   no | main:control
  ffffffff84028000 | idle          | 0                |    0 |    0 |   no | idle_thread
  ffffffc27ff0a200 | inactive      | 105dc            |  255 |    0 |   no | rootserver
```

### DIFF
--- a/src/kernel/debug.c
+++ b/src/kernel/debug.c
@@ -1,0 +1,210 @@
+/*
+ * Copyright 2014, General Dynamics C4 Systems
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <config.h>
+
+#ifdef CONFIG_PRINTING
+
+#include <types.h>
+#include <api/debug.h>
+#include <machine/io.h>
+#include <kernel/thread.h>
+
+static void print_fault(seL4_Fault_t f)
+{
+    seL4_Word fault_type = seL4_Fault_get_seL4_FaultType(f);
+    switch (fault_type) {
+    case seL4_Fault_NullFault:
+        printf("null fault");
+        break;
+    case seL4_Fault_CapFault:
+        printf("cap fault in %s phase at address %p",
+               seL4_Fault_CapFault_get_inReceivePhase(f) ? "receive" : "send",
+               (void *)seL4_Fault_CapFault_get_address(f));
+        break;
+    case seL4_Fault_VMFault:
+        printf("vm fault on %s at address %p with status %p",
+               seL4_Fault_VMFault_get_instructionFault(f) ? "code" : "data",
+               (void *)seL4_Fault_VMFault_get_address(f),
+               (void *)seL4_Fault_VMFault_get_FSR(f));
+        break;
+    case seL4_Fault_UnknownSyscall:
+        printf("unknown syscall %p",
+               (void *)seL4_Fault_UnknownSyscall_get_syscallNumber(f));
+        break;
+    case seL4_Fault_UserException:
+        printf("user exception %p code %p",
+               (void *)seL4_Fault_UserException_get_number(f),
+               (void *)seL4_Fault_UserException_get_code(f));
+        break;
+#ifdef CONFIG_KERNEL_MCS
+    case seL4_Fault_Timeout:
+        printf("Timeout fault for badge %p",
+               (void *)seL4_Fault_Timeout_get_badge(f));
+        break;
+#endif
+    default:
+        printf("unknown fault type %"SEL4_PRIu_word, fault_type);
+        break;
+    }
+}
+
+void debug_print_fault_handler(
+    tcb_t *tptr,
+    seL4_Fault_t fault
+#ifndef CONFIG_KERNEL_MCS
+    ,
+    exception_t status,
+    seL4_Fault_t ipcFault
+#endif /* not CONFIG_KERNEL_MCS */
+)
+{
+
+#ifdef CONFIG_KERNEL_MCS
+    printf("Found thread has no fault handler while trying to handle:\n");
+#else /* not CONFIG_KERNEL_MCS */
+    printf("Caught ");
+    print_fault(ipcFault);
+    printf("with IPC send status 0x%"SEL4_PRIx_word"\n"
+           "while trying to handle:\n",
+           status);
+#endif /* [not] CONFIG_KERNEL_MCS */
+    print_fault(fault);
+
+    const char *name = config_ternary(CONFIG_DEBUG_BUILD,
+                                      TCB_PTR_DEBUG_PTR(tptr)->tcbName,
+                                      null);
+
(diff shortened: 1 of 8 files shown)


src/kernel/faulthandler.c
```diff
@@ -43,129 +30,78 @@ bool_t sendFaultIPC(tcb_t *tptr, cap_t handlerCap, bool_t can_donate)
...
         }
-        sendIPC(true, true,
-                cap_endpoint_cap_get_capEPBadge(handlerCap),
-                cap_endpoint_cap_get_capCanGrant(handlerCap), true, tptr,
-                EP_PTR(cap_endpoint_cap_get_capEPPtr(handlerCap)));
 
-        return EXCEPTION_NONE;
-    } else {
-        current_fault = seL4_Fault_CapFault_new(handlerCPtr, false);
         current_lookup_fault = lookup_fault_missing_capability_new(0);
```
axel-h, line 85: I wonder, do we really have to update `current_lookup_fault` here? Seems no body uses this.

src/kernel/debug.c
```diff
@@ -0,0 +1,281 @@
...
+        config_ternary(CONFIG_KERNEL_MCS, "+------", "")
+        "+----------------"
+        "\n",
+        CONFIG_WORD_SIZE / 4, "TCB",
+        "State",
+        CONFIG_WORD_SIZE / 4, "PC",
+        CONFIG_WORD_SIZE / 4, DASHES,
+        CONFIG_WORD_SIZE / 4, DASHES);
+
+    for (tcb_t *tcb = specific_tcb ? specific_tcb : NODE_STATE(ksDebugTCBs);
```
axel-h, line 242: I wonder if this should try to dump the scheduler on all cores on day and not just for the current core.

lsf37 (changes requested):
It looks like there are lots of changes in here that break verification and are not needed to achieve the increased debug info. The effort for that would be in no relation to the improvement.

I'm also not convinced we do want that much info for every fault, e.g. there are lots of intended cap faults in sel4test and this looks like it would swamp the entire output. On the other hand I can see that this info could be very useful when no cap faults are intended.

What do others think of this?

axel-h:
I've reworked this now to have the changes only in the debug code that is invisible to the proofs. Let's see if this passes the proof test. Concerning the verbose output we could have another config flag here to reduce the debug output for seL4 test. Aside from this I'm open for ans specific help how we can tailor what it printed better to the cause.

kent-mcleod:
> I've reworked this now to have the changes only in the debug code that is invisible to the proofs. Let's see if this passes the proof test. Concerning the verbose output we could have another config flag here to reduce the debug output for seL4 test. Aside from this I'm open for ans specific help how we can tailor what it printed better to the cause.

Is there any existing issue about the kernel fault debug messages needing to be improved?  The kernel currently provides very simple debug fault console output and more detailed fault debug output can be implemented at user level. Shorter and simpler fault messages have an advantage of requiring the kernel spend less time generating and printing them and have less chance that the code to generate an error has an error itself.  It's also possible to modify the kernel code when debugging if additional information is needed without requiring this functionality to be merged upstream as it's not part of the verified configuration.

> Improve the messages a bit to no talk about a "cap fault" just because there is no fault handler. Print a register snapshot besides the stack dump if CONFIG_DEBUG_BUILD is enabled.

If there is no fault handler then it is correct to call it a cap fault because the kernel found no fault handler cap.

axel-h:
> It's also possible to modify the kernel code when debugging if additional information is needed without requiring this functionality to be merged upstream as it's not part of the verified configuration.

OK, making this PR a draft for now. Maybe this is also a candidate for having a configurable verbosity if debug kernel messages.

> If there is no fault handler then it is correct to call it a cap fault because the kernel found no fault handler cap.

Technically correct, my aim is just making the message be a bit more user friendly. I had to explain this message a couple of times now.

PR #668, opened 2021-11-27
remove obsolete define HAVE_AUTOCONF

axel-h:
All usages of `HAVE_AUTOCONF` have been removed already, so we can finally drop defining this for the kernel


lsf37:
Quick grep shows that `camkes-vm-apps` and `camkes-vm-examples` still mention `#ifdef HAVE_AUTOCONF` although I'm not sure that is actually used. Have you tested those?

lsf37:
(as in: I'm not sure wether setting this or not in the kernel cmake files is visible in those examples -- I can't see it being set there, only consumed)

axel-h:
- `camkes-vm-apps` is deprecated.
- the build instruction for `camkes-vm-examples` are outdated. What I can get building does not seem to use this because nobody sets `HAVE_AUTOCONF`. Seem we can remove this as dead code then?

lsf37:
> * `camkes-vm-apps` is deprecated.
> * the build instruction for `camkes-vm-examples` are outdated. Created [update readme camkes-vm-examples#15](https://github.com/seL4/camkes-vm-examples/pull/15) What I can get building does not seem to use this because nobody sets `HAVE_AUTOCONF`. Seem we can remove this as dead code then?

Yes, agreed, just wanted to give people an opportunity to shout out in case there was some other case we've overlooked. Will merge both.

lsf37 (approved):

PR #674, opened 2021-11-29
align include guards for seL4_DebugSendIPI()

axel-h:
Align the API wrapper guards with the kernel syscall implementation, the function is available if `CONFIG_ENABLE_SMP_SUPPORT` is set instead of depending on `CONFIG_MAX_NUM_NODES`.


lsf37 (approved):
Those are equivalent, but I agree that it's nicer to be consistent.

PR #676, opened 2021-11-30
Move debug implementation out of syscall.c

axel-h:
Make `syscall.c` call syscall handler function for debug syscalls, so the implementation is no longer visible there. That makes the dispatcher code easier to read and maintain. Also changes in the debug syscalls no longer cause cause preprocess failures, this is all hidden from verification anyway. Furthermore the this enforces a cleaner implementation of the debug syscalls, then need to fully handle the syscall.
The PR is on top of https://github.com/seL4/seL4/pull/673 and a preparation for reworking https://github.com/seL4/seL4/pull/298 eventually.

### DIFF
--- a/src/benchmark/benchmark.c
+++ b/src/benchmark/benchmark.c
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2016, General Dynamics C4 Systems
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <config.h>
+
+#ifdef CONFIG_ENABLE_BENCHMARKS
+
+#include <types.h>
+#include <mode/machine.h>
+#include <benchmark/benchmark.h>
+#include <benchmark/benchmark_utilisation.h>
+
+
+exception_t handle_SysBenchmarkFlushCaches(void)
+{
+#ifdef CONFIG_ARCH_ARM
+    tcb_t *thread = NODE_STATE(ksCurThread);
+    if (getRegister(thread, capRegister)) {
+        arch_clean_invalidate_L1_caches(getRegister(thread, msgInfoRegister));
+    } else {
+        arch_clean_invalidate_caches();
+    }
+#else
+    arch_clean_invalidate_caches();
+#endif
+    return EXCEPTION_NONE;
+}
+
+exception_t handle_SysBenchmarkResetLog(void)
+{
+#ifdef CONFIG_ENABLE_KERNEL_LOG_BUFFER
+    if (ksUserLogBuffer == 0) {
+        userError("A user-level buffer has to be set before resetting benchmark.\
+                Use seL4_BenchmarkSetLogBuffer\n");
+        setRegister(NODE_STATE(ksCurThread), capRegister, seL4_IllegalOperation);
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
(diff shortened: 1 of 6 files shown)


kent-mcleod:
The preprocess guards should already be sufficient to hide debug syscalls from preprocess failures.  Moving the implementation into functions shouldn't change this.

I agree with you that there's a bit of a scalability issue with implementing many debug system calls within the same function. But I think a different solution is needed, like making it possible to compile in a different handleUnknownSyscall implementation (similar to how the domain schedule is added: `list(APPEND c_sources "${KernelDomainSchedule}")`) which can make it easier to extend the number of syscalls in unverified kernel configurations with additional profile and test/debug points with different specialized implementations and save us from the complexity of trying to design general purpose ones that will work for everyone's use cases.

axel-h:
At the moment this is mainly intended to simplify maintainability, especially as I'm working on cleaning up the benchmak/logging and integrating https://github.com/seL4/seL4/pull/298 then. Moving the function out allows tracking the changes better and having a `benchmark.c` comes handy to consolidate more things there.
Similar for the IPI syscall, making this work in RISC-V is a bit nicer to track if this is becomes a handler function in `ipi.c`.
 I can look into way for completely customizing `handleUnknownSyscall()`after this. But moving things out into helper function seems a way towards this anyway.

axel-h:
Can this get merged? It would really help having the handlers in separate files, changes causes less conflicts then in `syscall.c` then. It might be an interim step to make life/work a bit easier until there is a proposal for the way @kent-mcleod  suggested.

src/benchmark/benchmark.c
```diff
@@ -0,0 +1,161 @@
...
+{
+#ifdef CONFIG_ENABLE_KERNEL_LOG_BUFFER
+    word_t cptr_userFrame = getRegister(NODE_STATE(ksCurThread), capRegister);
+    if (benchmark_arch_map_logBuffer(cptr_userFrame) != EXCEPTION_NONE) {
+        setRegister(NODE_STATE(ksCurThread), capRegister, seL4_IllegalOperation);
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+#endif /* CONFIG_ENABLE_KERNEL_LOG_BUFFER */
+
+    /* ToDo: If CONFIG_ENABLE_KERNEL_LOG_BUFFER is not set it might be better to
```
kent-mcleod, line 85: Can you remove this new comment please.  It would be better suited as an issue instead of as a new ToDo in the code.
Also can you try and not add modifications at the same time as moving large chunks of code as it makes it a lot harder to review.

kent-mcleod (approved):

lsf37 (approved):
I'm fine with this from the verification side.

src/benchmark/benchmark.c
kent-mcleod: Sorry, I accidentally merged this without giving @axel-h a chance to resolve my comment.

kent-mcleod: I think this should turn into a seL4_Fault_UnknownSyscall if it isn't enabled.

axel-h: I will make a new PR an address this.

axel-h: See https://github.com/seL4/seL4/pull/748

PR #678, opened 2021-11-30
resolve include dependency issues

axel-h:
Include the timer header after `getKernelWcetUs()` is defined.

The include quirk this resolves is that If `CONFIGURE_TIMER` is `drivers/timer/arm_global.h`, this includes `machine/timer. h` which implements the function `getKernelWcetTicks()` that calls `getKernelWcetUs()` which `src/arch/arm/platform_gen.h.in` defines. So it is better if we move the include after `getKernelWcetUs()`.


kent-mcleod:
Is this fixing a compiler error somewhere?

axel-h:
I was running into this in one of my dev branches and that was the fix to ensure the right include order. Will investigate in more detail why this works in master.

kent-mcleod:
Closing this as it still doesn't appear to be necessary on the main kernel branch (which doesn't have any outstanding include errors).

PR #679, opened 2021-12-01
Fix getMaxUsToTicks function in RISC-V machine/timer.h

kgugala:
The `getMaxUsToTicks` function is supposed to return maximum amount of time you can pass to `usToTicks` without overflow,but in RISC-V implementation it returns ticks instead. This patch corrects it so it's implementation is akin to those in other platforms.

### DIFF
--- a/include/arch/riscv/arch/machine/timer.h
+++ b/include/arch/riscv/arch/machine/timer.h
@@ -39,12 +39,12 @@ static inline PURE ticks_t getTimerPrecision(void)
 
 static inline CONST ticks_t getMaxTicksToUs(void)
 {
-    return UINT64_MAX / TICKS_IN_US;
+    return UINT64_MAX;
 }
 
-static inline PURE time_t getMaxUsToTicks(void)
+static inline CONST time_t getMaxUsToTicks(void)
 {
-    return usToTicks(getMaxTicksToUs());
+    return UINT64_MAX / TICKS_IN_US;
 }
 
 /* Read the current time from the timer. */


lsf37 (approved):
It does look like you are embarrassingly correct, yes.

axel-h (approved):

axel-h:
@lsf37 Should proof-test run or can we merge this directly?

lsf37:
Safer to do the test, even if this one looks obvious. I've become lax about it and have been bitten by it once or twice (trivial changes only, but it's better to know about them before they happen).

kent-mcleod:
I think this fix is correct.  This came up before: https://github.com/seL4/seL4/pull/312#issuecomment-807857887 and supposedly back then we reasoned it out wrong by treating the RISC-V definitions as source of truth because of verification:  getMaxUsToTicks was converting the the max US value to ticks and getMaxTicksToUs was converting the max Ticks value to us even though this is not the same as other architectures, and not what the type signatures declare.

The reason this inconsistency has side-stepped the proofs is that `getMaxUsToTicks` is #ifdef'd out in a proof configuration and isn't actually used: https://github.com/seL4/seL4/blob/master/include/kernel/sporadic.h#L47

axel-h:
@kgugala could you allow edits to your branch (check "Allow edits and access to secrets by maintainers") on bottom right column), so this can be rebased and merged via the git workflow?

kgugala:
I'm afraid that, for some reason, I don't see this option:
image

I can rebase the PR manually

kgugala:
I have rebase the PR

PR #680, opened 2021-12-01
riscv: pass hart mask by value to SBI wrappers

axel-h:
Hide the actual SBI call parameter details of the legacy API and remove the pointer parameters in the higher code layers. This is a preparation step for switching to the newer SBI API.

### DIFF
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -82,13 +82,13 @@ static inline void ifence(void)
 {
     ifence_local();
 
-    unsigned long mask = 0;
+    word_t mask = 0;
     for (int i = 0; i < CONFIG_MAX_NUM_NODES; i++) {
         if (i != getCurrentCPUIndex()) {
             mask |= BIT(cpuIndexToID(i));
         }
     }
-    sbi_remote_fence_i(&mask);
+    sbi_remote_fence_i(mask);
 }
 
 static inline void sfence(void)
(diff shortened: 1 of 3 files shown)


kent-mcleod (approved):
Someone will need to take a look at the CKernel proof test failure.

lsf37:
Yeah, it's complaining about code taking the address of a local variable. That is not in the C subset. I wonder why the C parser test didn't show this already, but it might need too much semantic info for that test to do it at the purely syntactic level.

I also wonder why this was possible before, because this is really just shifting the location where that happens. Let me have a closer look.

lsf37:
So it looks like all places where the address of a local variable is taken currently are SMP and should probably be under an SMP guard anyway. @axel-h does that makes sense?

This would make the problem temporarily invisible to verification, and if more recent SBI versions don't want a pointer anyway, this will go away before we need to deal with it on the verification side. Otherwise we need to move that parameter into a global, make sure it is core-local and make sure there are no conflicting accesses to it.

I haven't quite figured out yet why it wasn't visible before. Most likely because those functions were underneath the machine interface and weren't translated into Isabelle.

axel-h:
Yes, this is all compiled in for SMP only. I can hide the SBI call in the SMP configuration also, then we can keep this hidden for a while until we either switch to the new API or find a way to handle this.

axel-h:
Seem this passes the RISCV64 proofs now. Can we merge this now to have this pointer parameter removed from the higher code layers?

lsf37 (approved):
Yup, this is good to merge.

PR #690, opened 2021-12-06
Zynqmp ultra96 fixes improvements

Furao:
Just some improvements for configuration and flexibility of zynqmp platforms. Tested with zcu102 and ultra96v2.


src/plat/zynqmp/overlay-ultra96v2.dts
```diff
@@ -0,0 +1,307 @@
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * This software may be distributed and modified according to the terms of
+ * the GNU General Public License version 2. Note that NO WARRANTY is provided.
+ * See "LICENSE_GPLv2.txt" for details.
+ *
+ * @TAG(DORNERWORKS_GPL)
```
axel-h, line 8:
```suggestion
 * SPDX-License-Identifier: GPL-2.0-only
```

tools/dts/ultra96v2.dts
```diff
@@ -0,0 +1,2185 @@
+/dts-v1/;
```
axel-h, line 1: Seem the check also wants this:
```suggestion
/*
 * Copyright Linux Kernel Team
 *
 * SPDX-License-Identifier: GPL-2.0-only
 *
 * This file is derived from an intermediate build stage of the
 * Linux kernel. The licenses of all input files to this process
 * are compatible with GPL-2.0-only.
 */

/dts-v1/;
```

Furao: updated

axel-h (approved):

src/plat/zynqmp/overlay-ultra96v2.dts
```diff
@@ -0,0 +1,307 @@
+/*
+ * Copyright 2019, DornerWorks
+ *
+ * This software may be distributed and modified according to the terms of
+ * the GNU General Public License version 2. Note that NO WARRANTY is provided.
+ * See "LICENSE_GPLv2.txt" for details.
+ *
```
lsf37, line 7:
```suggestion
 * This software may be distributed and modified according to the terms of
 * the GNU General Public License version 2. Note that NO WARRANTY is provided.
 * See "LICENSE_GPLv2.txt" for details.
 *
```
With the new SPDX identifiers, this bit can also go. The file the comment refers to does not exist any more.

axel-h:
Could you add a note at the beginning of the DTS where you have taken this from? Would be good to mention a public repo and a commit id, ideally `tools/dts/update-dts.sh` would be able to pull it from somewhere.
I've just compared the v1 and v2 trees and what various github repos. Seem even the source v1 board's DTS we have is not really clear and it seem also not the most recent one. So checking for difference between the two files brings up a lot... seems there is not much chance having the differences just in an overlay

lsf37:
The zcu102 test is currently failing, because the board is not booting. As soon as @wom-bat has had a chance of fixing that and the test passes, I think this can be merged.

lsf37:
The test have sufficiently passed (the two failures on other boards were unrelated). If you could fix up that one copyright header, this one can be merged.

src/plat/zynqmp/overlay-ultra96v2.dts
Furao: Will update.

Furao:
> Could you add a note at the beginning of the DTS where you have taken this from? Would be good to mention a public repo and a commit id, ideally tools/dts/update-dts.sh would be able to pull it from somewhere.
I've just compared the v1 and v2 trees and what various github repos. Seem even the source v1 board's DTS we have is not really clear and it seem also not the most recent one. So checking for difference between the two files brings up a lot... seems there is not much chance having the differences just in an overlay

@axel-h  Unfortunately, it is not stored in a repo anywhere. I believe we originally got it from here: https://www.96boards.org/documentation/consumer/ultra96/ultra96-v2/downloads/linux.md.html
They don't really capture it because it is generated from the hardware definition using Xilinx tools.

axel-h:
The commit "ultra96: differentiate between v1 and v2" adds a overlay now with lots of definitions. Why do we have to keep this in an overlay, can't we put this into `tools/dts/ultra96v2.dts` also? Seem this DTS file it not used any longer, so why should we keep it then - or introduce in the previous commit?

Furao:
> The commit "ultra96: differentiate between v1 and v2" adds a overlay now with lots of definitions. Why do we have to keep this in an overlay, can't we put this into tools/dts/ultra96v2.dts also? Seem this DTS file it not used any longer, so why should we keep it then - or introduce in the previous commit?

You are right, that was an oversight. We should squash the first 2 commits and remove the overlay. Will update.

lsf37:
Looks good, will merge!

PR #691, opened 2021-12-06
Domain scheduler multicore support

Furao:
Gang scheduling domains and leaving thread scheduling within
the domain up to the round-robin scheduler.

Also updated scheduler dump info to make it a little prettier.

### DIFF
--- a/include/api/debug.h
+++ b/include/api/debug.h
@@ -77,7 +77,7 @@ static inline void debug_printUserState(void)
 
 static inline void debug_printTCB(tcb_t *tcb)
 {
-    printf("%40s\t", TCB_PTR_DEBUG_PTR(tcb)->tcbName);
+    printf("%50s\t", TCB_PTR_DEBUG_PTR(tcb)->tcbName);
     char *state;
     switch (thread_state_get_tsType(tcb->tcbState)) {
     case ThreadState_Inactive:
@@ -114,7 +114,7 @@ static inline void debug_printTCB(tcb_t *tcb)
     }
 
     word_t core = SMP_TERNARY(tcb->tcbAffinity, 0);
-    printf("%15s\t%p\t%20lu\t%lu", state, (void *) getRestartPC(tcb), tcb->tcbPriority, core);
+    printf("%20s\t%p\t0x%lx\t%3lu", state, (void *) getRestartPC(tcb), tcb->tcbPriority, core);
 #ifdef CONFIG_KERNEL_MCS
     printf("\t%lu", (word_t) thread_state_get_tcbInReleaseQueue(tcb->tcbState));
 #endif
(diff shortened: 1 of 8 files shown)


include/api/debug.h
```diff
@@ -123,14 +123,18 @@ static inline void debug_printTCB(tcb_t *tcb)
 
 static inline void debug_dumpScheduler(void)
 {
-    printf("Dumping all tcbs!\n");
-    printf("Name                                    \tState          \tIP                  \t Prio \t Core%s\n",
-           config_set(CONFIG_KERNEL_MCS) ?  "\t InReleaseQueue" : "");
-    printf("--------------------------------------------------------------------------------------\n");
-    for (tcb_t *curr = NODE_STATE(ksDebugTCBs); curr != NULL; curr = TCB_PTR_DEBUG_PTR(curr)->tcbDebugNext) {
-        debug_printTCB(curr);
+    /* Loop through each core and print out scheduler! */
```
axel-h, line 126: Should this take the big kernel log to ensure what is printed is really consistent?

Furao: That's a good question. I suppose it depends on what you are using it for. If it is just for some simple debugging I don't think you care if the numbers are off by a little bit. If you want to use it to help calculate some sort of performance metrics, then you would probably care. We have really only ever used it for the former.

If anyone cares about the latter, we could look into it.

axel-h: I think for now adding a comment would be sufficient that this is not fully synchronized across multiple cores, but works good enough for debug purposes. The rest could be a follow-up action if it is not trivial.

include/api/debug.h
```diff
@@ -123,14 +123,18 @@ static inline void debug_printTCB(tcb_t *tcb)
 
 static inline void debug_dumpScheduler(void)
 {
-    printf("Dumping all tcbs!\n");
-    printf("Name                                    \tState          \tIP                  \t Prio \t Core%s\n",
-           config_set(CONFIG_KERNEL_MCS) ?  "\t InReleaseQueue" : "");
```
kent-mcleod, line 128: Are you able to move this commit to a separate PR. It's more of a subjective discussion whether reformatting this debug syscall is wanted.  In particular removing the `InReleaseQueue` state and adding ANSI color codes is not an unambiguous improvement and discussing that is a different scope than the domain scheduler changes..

axel-h: Also, we could merge this change then with what I have pending in https://github.com/seL4/seL4/pull/665/commits/1f3c02f540293fad9e0dbed37a32df8dcb6e499f that dynamically adapts the numer of columns to the the configuration (MCS, SMP)

lsf37:
So I'm a bit confused what this is actually doing. We're introducing a per-core index into the one global domain schedule, but there is still only one global `ksDomainTime` which it looks like is still updated by each core individually. Am I missing something?

For gang-scheduling, I'd have expected still the one global schedule with one global index, but maybe one dedicated core being responsible for that schedule and forcing all cores to switch at the same time. There may be other ways to achieve the same, but I'm not really clear on the mechanism in the PR yet.

lsf37:
I think the mechanism behind SMP + domains needs some design rationale documentation. Ideally we'd do this in an RFC, since this is basically a new feature and how this works does affect future design decisions.

@kent-mcleod and I would be interested in working together with you on writing up something that lays out how domains+SMP should work. Would you have some time for that? We can also discuss a bit more on this PR and then put the result/summary up as an RFC.

include/api/debug.h
Furao: Yes, we could probably split this up into two PRs. I didn't think a debug print would have that much contention, but since it seems to, yes, we can move it.

Furao:
> I think the mechanism behind SMP + domains needs some design rationale documentation. Ideally we'd do this in an RFC, since this is basically a new feature and how this works does affect future design decisions.

Agreed, I do think we need a little more community thought behind this. If it helps, our use case is utilizing this to support ARINC653 scheduling. 
 
> @kent-mcleod and I would be interested in working together with you on writing up something that lays out how domains+SMP should work. Would you have some time for that? We can also discuss a bit more on this PR and then put the result/summary up as an RFC.

That is a good idea. @Apave24 is the one that actually implemented the change. He might have some more insights, along with @natestuder. We probably don't have much availability for this task until the new year.

kent-mcleod:
Marking this a draft as it will likely require more design discussion via an RFC.

PR #692, opened 2021-12-06
Mcs bug fixes

Furao:
We encountered and fixed a couple of issues with the MCS scheduler when enabling it with multicore for virtualized ARMv8 platforms.

### DIFF
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -533,7 +533,7 @@ void setPriority(tcb_t *tptr, prio_t prio)
 void possibleSwitchTo(tcb_t *target)
 {
 #ifdef CONFIG_KERNEL_MCS
-    if (target->tcbSchedContext != NULL && !thread_state_get_tcbInReleaseQueue(target->tcbState)) {
+    if (target->tcbSchedContext != NULL && !thread_state_get_tcbInReleaseQueue(target->tcbState) && isRunnable(target)) {
 #endif
         if (ksCurDomain != target->tcbDomain
             SMP_COND_STATEMENT( || target->tcbAffinity != getCurrentCPUIndex())) {
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -188,7 +188,9 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
                 current_fault = seL4_Fault_Timeout_new(receiver->tcbSchedContext->scBadge);
                 handleTimeout(receiver);
             } else {
-                postpone(receiver->tcbSchedContext);
+                if (!thread_state_get_tcbInReleaseQueue(receiver->tcbState)) {
+                    postpone(receiver->tcbSchedContext);
+                }
             }
         }
     }


src/kernel/thread.c
```diff
@@ -188,7 +188,9 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
                 current_fault = seL4_Fault_Timeout_new(receiver->tcbSchedContext->scBadge);
                 handleTimeout(receiver);
             } else {
-                postpone(receiver->tcbSchedContext);
+                if(thread_state_get_tcbInReleaseQueue(receiver->tcbState) == false){
```
axel-h, line 191: Maybe avoid comparing to "false" explicitly:
```suggestion
                if(!thread_state_get_tcbInReleaseQueue(receiver->tcbState)){
```

Furao: sure

Furao:
Fixed some issues reported by the style checker.

lsf37:
(The proofs don't see the MCS C code yet, they are still in the stage before that, so no point in running them, really)

kent-mcleod:
Hi @Furao, are you able to provide any more information about the motivation for these changes. Instructions for how we could potentially reproduce them? The original code is currently being verified and these conditions are being tracked very closely by the team working on the verification.  So we would need to know exactly what is going on to require these changes.

lsf37:
These changes do not look Aarch64 or SMP or HYP specific, so @mktnk3 or @michaelmcinerney might be able to say something about them from the proof side.

lsf37:
In particular: do we have any invariants about the release queue that say something about these two situations?

kent-mcleod:
In particular:
- https://github.com/seL4/seL4/pull/692/commits/d64d1f1dac6de0f4ca2683d55e809f9bb363cb6e: It shouldn't be possible for doReplyTransfer to be called with a receiver that is already in the release queue. If the thread is in the releaseQueue it cannot be blocked on a reply object at the same time. So if it has ended up in this state then it indicates there could be an issue somewhere else in the kernel.
- https://github.com/seL4/seL4/pull/692/commits/6a35bf9ee8367e4aca53204fb317f612f10bfbe5: possibleSwitchTo is called when a thread becomes runnable and is not in the release queue.  I believe at the moment it is the responsibility of the caller to only call this on threads that are already `isRunnable()` and it may make more sense for isRunnable() to be asserted here instead. (And then tracking down where possibleSwitchTo is being called where this isn't true.

mktnk3:
> * [6a35bf9](https://github.com/seL4/seL4/commit/6a35bf9ee8367e4aca53204fb317f612f10bfbe5): possibleSwitchTo is called when a thread becomes runnable and is not in the release queue.  I believe at the moment it is the responsibility of the caller to only call this on threads that are already `isRunnable()` and it may make more sense for isRunnable() to be asserted here instead. (And then tracking down where possibleSwitchTo is being called where this isn't true.

I agree that we expect the thread is expected to be runnable when possibleSwitchTo is called, and the proofs establish that information as needed, but it is often provided simply by a guard preceding to the call. I believe this is not MCS specific.

lsf37:
> I agree that we expect the thread is expected to be runnable when possibleSwitchTo is called, and the proofs establish that information as needed, but it is often provided simply by a guard preceding to the call. I believe this is not MCS specific.

So if this change fixes something, it looks like we might be running into a situation on SMP or HYP where that check is missing before `possibleSwitchTo` is called. That is a good lead to chase down.

lsf37:
> Hi @Furao, are you able to provide any more information about the motivation for these changes. Instructions for how we could potentially reproduce them? 

Ideally we'd want to add a sel4test case that reproduces the problem and reliably shows its absence when fixed.

mktnk3:
> In particular:
> 
> * [d64d1f1](https://github.com/seL4/seL4/commit/d64d1f1dac6de0f4ca2683d55e809f9bb363cb6e): It shouldn't be possible for doReplyTransfer to be called with a receiver that is already in the release queue. If the thread is in the releaseQueue it cannot be blocked on a reply object at the same time. So if it has ended up in this state then it indicates there could be an issue somewhere else in the kernel.

For the doReplyTransfer part, this also agrees with what the proofs expect. `valid_sched` and `invs` (via `valid_replies`) expect that all the threads in the release queue are runnable and the thread linked to a reply is in BlockedOnReply.

Furao:
We were using MCS on multicore AARCH64 VMM. We ran into issues without these changes. We implemented these a bit ago, but I am working to see if I can replicate the issues without the commits in place.

Furao:
https://github.com/seL4/seL4/commit/6a35bf9ee8367e4aca53204fb317f612f10bfbe5 The issue seems to occur when I have multiple VMs running on the same core. Also, the assert doesn't always seem to print, but I always halt when I have VMs running on the same core. When they were each running on their own cores, a halt never occurred. 
See the log here: [is_schedulable_error.log](https://github.com/seL4/seL4/files/7676276/is_schedulable_error.log)
is_schedulable_error_config

Furao:
https://github.com/seL4/seL4/commit/d64d1f1dac6de0f4ca2683d55e809f9bb363cb6e This issue seems to occur when we have a a VM that is running on multiple cores. 
See the log here: [tcb_in_release_queue_error.log](https://github.com/seL4/seL4/files/7676514/tcb_in_release_queue_error.log)

tcb_in_release_queue_error_config

Furao:
> Ideally we'd want to add a sel4test case that reproduces the problem and reliably shows its absence when fixed.
@lsf37 We tried to work on some test cases for this a bit ago. However, they seem to be caused by race conditions that are difficult to replicate with seL4 test. It is a consistent failure using camkes-vm though. 

Thoughts on ways to move forward with these issues?

kent-mcleod:
> Thoughts on ways to move forward with these issues?

I'd started looking into reproducing this, but found that the camkes-vm doesn't support MCS yet. This means there's not a way to reproduce MCS + Hyp + SMP with a multicore vm.

kent-mcleod:
> Oh right, we have more commits that enable that. We will work on getting pull requests for those as well. I will ping you when those are available.

Ok, I'll mark this as a draft for now.

lsf37:
Some progress on this (maybe) is that it turned out we had not enabled `hyp` at all for the `zynqmp` platforms in CI. I have done that now and combinations `hyp+smp+clang` seem to fail a lot with and without MCS.

I don't know if these are the same issues or not, but the fixes in the PR are basically all checking for things that should not be possible, so maybe there are some leads in these failures. Could also just be clang big-lock issues, though.

Indanz:
I'm going to close this assuming it has been resolved by the SMP locking fixes and/or MCS bugfixes we made since this was opened.

If it still occurs, feel free to reopen this.

PR #698, opened 2021-12-08
libsel4: adjust asm immediate syntax for clang

lsf37:
Commit 8035f1a7c4 removed "-fno-integrated-as" for llvm, which on more
recent versions (clang-11) means that we now need `#` instead of `$`
for immediate values.

### DIFF
--- a/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/syscalls.h
+++ b/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/syscalls.h
@@ -68,7 +68,7 @@ static inline void arm_sys_send(seL4_Word sys, seL4_Word dest, seL4_Word info_ar
     /* Perform the system call. */
     register seL4_Word scno asm("r7") = sys;
     asm volatile(
-        "swi $0"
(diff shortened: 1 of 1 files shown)


axel-h:
Does this mean the older version will no longer work now? Or should this always have been `#`?

lsf37:
I believe the older version should also still work, but @kent-mcleod can probably confirm. AFAIU it's the new version being stricter.

lsf37:
We finally merged the new docker containers yesterday, and I thought had tested everything, but apparently not. I expect a few more glitches, but hopefully nothing too dramatic. (Preprocess test seems to need an adjustment)

kent-mcleod:
> I believe the older version should also still work, but @kent-mcleod can probably confirm. AFAIU it's the new version being stricter.

Using `#` is the official[1] syntax and is already used everywhere else for aarch32. 

[1]: https://developer.arm.com/documentation/dui0473/k/writing-arm-assembly-language/register-usage-in-subroutine-calls?lang=en

kent-mcleod (approved):

Issue #660, opened 2021-11-23
Unify boot flow of ARM and RISC-V


axel-h:
The boot flow on ARM and RISC-V are largely similar, thus is seems worth to unify them. Either the common parts are factored out in a shared file, so the architecture specific part still has the full control or the common flow gets some architecture specific hooks that get called for the differences. Having a unified flow makes maintenance easier and avoid artificial differences that just seem to exists for historic reasons.

What would be the bests place for shared files in the directory structure?

kent-mcleod:
Functions that are generic across Arm and RISCV could go in the common src/kernel/boot.c file and guarded from x86 configurations using the preprocessor.

PR #699, opened 2021-12-10
boot: don't round DTB area down or up

axel-h:
Follow-up from comment https://github.com/seL4/seL4/pull/422#discussion_r715981928

There is no reason to round the DTB area up or down to page boundaries, because the content is copied anyway and reserving memory areas works on arbitrary granularity.

This is also a step towards https://github.com/seL4/seL4/issues/660 to align the ARM/RISC-V boot code, so it can be unified eventually.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -364,7 +364,7 @@ static BOOT_CODE bool_t try_init_kernel(
     /* If a DTB was provided, pass the data on as extra bootinfo */
     p_region_t dtb_p_reg = P_REG_EMPTY;
     if (dtb_size > 0) {
-        paddr_t dtb_phys_end = ROUND_UP(dtb_phys_addr + dtb_size, PAGE_BITS);
+        paddr_t dtb_phys_end = dtb_phys_addr + dtb_size;
         if (dtb_phys_end < dtb_phys_addr) {
             /* An integer overflow happened in DTB end address calculation, the
              * location or size passed seems invalid.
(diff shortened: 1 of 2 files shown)


axel-h:
Seems `ODROID_XU4_debug_hyp_clang_32` does not like this
```
Running test FRAMEEXPORTS0001 (Test that we can access all exported frames)
  [[Timeout]]
  None
```
Is there anything known here?

axel-h:
Any feedback on this? The remaining failing test seem be be unrelated to this change and these know issues exist:
- https://github.com/seL4/sel4test/issues/65
- https://github.com/seL4/sel4test/issues/64

lsf37:
This *looks* fine to me, but I'm not deep enough in all the assumptions the boot code makes to say for sure that we should not round here.

axel-h:
Can we merge this?

axel-h:
Is there an issue with the ZYNQ board?
```
Lock acquired, we are allowed to run
  '/tmp/tmp.9Utk8Deupt' -> '/tftpboot/zcu102/sel4-image'
  Couldn't change mode of /tftpboot/zcu102/sel4-image: [Errno 1] Operation not permitted: '/tftpboot/zcu102/sel4-image'
```

lsf37:
> `Couldn't change mode of /tftpboot/zcu102/sel4-image: [Errno 1] Operation not permitted: '/tftpboot/zcu102/sel4-image'`

This bit is normal and occurs for all boards (should be fixed at some point, but it's harmless the permissions of the image are always sufficient).

The log looks like the board isn't coming up for receiving the image. It'll need someone to look at it in person. Will check if there is someone around or if we have to wait for Peter to be back from holidays (if that is the case I can disable the board for the test temporarily).

lsf37:
> The log looks like the board isn't coming up for receiving the image. It'll need someone to look at it in person. Will check if there is someone around or if we have to wait for Peter to be back from holidays (if that is the case I can disable the board for the test temporarily).

`zcu102` should be operational again now.

kent-mcleod (approved):

PR #700, opened 2021-12-10
Merge ARM and RISC-V kernel init boot code 

axel-h:
This addresses https://github.com/seL4/seL4/issues/660 and unifies the kernel init boot flow of ARM and RISC-V to share the same code base. Both architectures have been doing almost the same anyway, making this use common code to reduced the maintenance efforts in the future. The kernel initialization is mostly high level C code already. There are just a few placed left with conditional code, they can be unified in a follow-up task after clarifying the details, 

This builds on top of the PRs
- https://github.com/seL4/seL4/pull/699
- https://github.com/seL4/seL4/pull/704

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -40,6 +40,7 @@ Upcoming release: BINARY COMPATIBLE
 * Removed obsolete define `HAVE_AUTOCONF`
 * Removed user address space reserved slots restriction on 40bit PA platforms when KernelArmHypervisorSupport is set.
   This change is reflected in the definition of the seL4_UserTop constant that holds the largest user virtual address.
+* Merge ARM and RISC-V generic kernel initialization code
 
 ## Upgrade Notes
 ---
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -135,3 +135,31 @@ static inline BOOT_CODE pptr_t it_alloc_paging(void)
 
 /* return the amount of paging structures required to cover v_reg */
 word_t arch_get_n_paging(v_region_t it_veg);
+
+#if defined(CONFIG_ARCH_ARM) || defined(CONFIG_ARCH_RISCV)
+
+bool_t arch_init_freemem(
+    p_region_t ui_p_reg,
+    p_region_t dtb_p_reg,
+    v_region_t it_v_reg,
+    word_t extra_bi_size_bits);
+
+void arch_init_irqs(cap_t root_cnode_cap);
+#ifdef CONFIG_ARM_SMMU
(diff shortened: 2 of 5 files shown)


kent-mcleod:
> Both architectures have been doing almost the same anyway, making this use common code to reduced the maintenance efforts in the future.

I don't think unification is the right approach here.  For similar reasons as why there was a request to support different versions of booting (https://sel4.discourse.group/t/pre-rfc-arm-boot-interface-versioning/410/5). As RISC-V and Arm architectures are independent, trying to fit the code to a single abstracted architecture is unlikely to reduce overall complexity as it adds a requirement for defining an abstract model and then fitting each actual architecture to the abstract model.  When working on or reviewing boot code you would now have to know what effect your changes would have across all architectures instead of just the one you are using.

axel-h:
> I don't think unification is the right approach here. For similar reasons as why there was a request to support different versions of booting (https://sel4.discourse.group/t/pre-rfc-arm-boot-interface-versioning/410/5). As RISC-V and Arm architectures are independent, trying to fit the code to a single abstracted architecture is unlikely to reduce overall complexity as it adds a requirement for defining an abstract model and then fitting each actual architecture to the abstract model. When working on or reviewing boot code you would now have to know what effect your changes would have across all architectures instead of just the one you are using.

On the practical side, ARM and RISC-V are quite similar, and most of the code I'm unifying here is basically the same. The RISC-V port largely seem a fork of the ARM port. Thus I'd argue that a certain model for the boot exists already and it works quite well. Also, currently some things need to be done twice, and more than once that uncovered a subtle difference that doesn't really have to be there at all. The overall gain would be better comments in the code where there are real differences that don't just come from things slowly diverging. There might be a price, but I'd like to find these places and then see if the compiler or config wont be able to handle this.
We have the GitHub that can be used to check how changes affect other platforms/architectures, this seems a mature way by now to ensure nothing is broken.

lsf37:
> We have the GitHub that can be used to check how changes affect other platforms/architectures, this seems a mature way by now to ensure nothing is broken.

Tests are really not enough to spot these subtle things, esp not in boot code, and it's not like the CI tests have super great coverage anyway.

I get that a lot of code is very similar between the RISC-V and ARM boot sequence, but going for a big abstraction/unification effort now without verification backing it up seems to be just a big opportunity to introduce subtle problems. Of course it might also fix subtle problems, but we don't see them. If there are issues we should fix them and check if they apply to other architectures, but otherwise, if it ain't broken, don't fix it.

What really should be happening is a verification effort there and maybe a principled design of what makes sense to share and what not.

PR #701, opened 2021-12-10
SMC Capability

Furao:
Add a new capability which certain threads can invoke so seL4 can make SMC calls on ARM platforms in EL2 (virtualized mode) on the thread’s behalf.

See https://sel4.atlassian.net/browse/RFC-9

### DIFF
--- a/include/arch/arm/arch/object/structures.h
+++ b/include/arch/arm/arch/object/structures.h
@@ -12,6 +12,16 @@
 
 static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 {
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    switch (cap_get_capType(derivedCap)) {
+    case cap_smc_cap:
+        return (cap_smc_cap_get_capSMCBadge(derivedCap) !=
+                cap_smc_cap_get_capSMCBadge(srcCap));
+
+    default:
+        return false;
+    }
+#endif
     return false;
 }
 
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -27,7 +27,8 @@ enum {
     seL4_CapSMMUSIDControl      = 12, /* global SMMU SID controller cap, null cap if not supported */
     seL4_CapSMMUCBControl       = 13, /* global SMMU CB controller cap, null cap if not supported */
     seL4_CapInitThreadSC        = 14, /* initial thread's scheduling context cap, null cap if not supported */
-    seL4_NumInitialCaps         = 15
+    seL4_CapSMC                 = 15, /* global SMC cap, null cap if not supported */
+    seL4_NumInitialCaps         = 16
 };
 
 /* Legacy code will have assumptions on the vspace root being a Page Directory
--- a/libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
+++ b/libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
@@ -44,6 +44,16 @@
         <member name="tpidr_el0"/>
         <member name="tpidrro_el0"/>
     </struct>
+    <struct name="seL4_ARM_SMCContext">
+        <member name="x0"/>
+        <member name="x1"/>
+        <member name="x2"/>
+        <member name="x3"/>
+        <member name="x4"/>
+        <member name="x5"/>
+        <member name="x6"/>
+        <member name="x7"/>
+    </struct>
     <interface name="seL4_ARM_VSpace" manual_name="Page Global Directory"
         cap_description="Capability to the top level translation table being operated on.">
         <method id="ARMVSpaceClean_Data" name="Clean_Data" manual_label="vspace_clean"
@@ -205,6 +215,22 @@
                     The specified range crosses a page boundary.
                 </description>
             </error>
+        </method>
+    </interface>
+    <interface name="seL4_ARM_SMC" manual_name="SMC" cap_description="Capability to allow threads to make Secure Monitor Calls.">
+        <method id="ARMSMCCall" name="Call" manual_name="SMC Call" manual_label="smc_call">
+            <brief>
+                Tell the microkernel to make the real SMC call.
+            </brief>
+            <description>
+                Takes x0-x7 as arguments to an SMC call which are defined as a seL4_ARM_SMCContext
+                struct. The microkernel makes the SMC call and then returns the results as a
+                new seL4_ARM_SMCContext.
+            </description>
+            <param dir="in" name="smc_args" type="seL4_ARM_SMCContext"
+                description="The structure that has the provided arguments."/>
+            <param dir="out" name="smc_response" type="seL4_ARM_SMCContext"
+                description="The structure to capture the responses."/>
         </method>
     </interface>
 </api>
--- a/src/arch/arm/64/object/objecttype.c
+++ b/src/arch/arm/64/object/objecttype.c
@@ -85,6 +85,12 @@ deriveCap_ret_t Arch_deriveCap(cte_t *slot, cap_t cap)
         ret.status = EXCEPTION_NONE;
         return ret;
 #endif
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    case cap_smc_cap:
+        ret.cap = cap;
+        ret.status = EXCEPTION_NONE;
+        return ret;
+#endif
     default:
         /* This assert has no equivalent in haskell,
          * as the options are restricted by type */
@@ -403,6 +409,10 @@ exception_t Arch_decodeInvocation(word_t label, word_t length, cptr_t cptr,
     case cap_cb_cap:
         return decodeARMCBInvocation(label, length, cptr, slot, cap, call, buffer);
 #endif /*CONFIG_ARM_SMMU*/
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    case cap_smc_cap:
+        return decodeARMSMCInvocation(label, length, cptr, slot, cap, call, buffer);
+#endif
     default:
 #else
 {
--- a/src/arch/arm/64/object/objecttype.c
+++ b/src/arch/arm/64/object/objecttype.c
@@ -405,7 +405,7 @@ exception_t Arch_decodeInvocation(word_t label, word_t length, cptr_t cptr,
     /* The C parser cannot handle a switch statement with only a default
      * case. So we need to do some gymnastics to remove the switch if
      * there are no other cases */
-#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT) || defined(CONFIG_ARM_SMMU)
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT) || defined(CONFIG_ARM_SMMU) || defined(CONFIG_ALLOW_SMC_CALLS)
     switch (cap_get_capType(cap)) {
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
     case cap_vcpu_cap:
--- a/src/arch/arm/64/object/objecttype.c
+++ b/src/arch/arm/64/object/objecttype.c
@@ -275,6 +275,13 @@ bool_t CONST Arch_sameRegionAs(cap_t cap_a, cap_t cap_b)
         }
         break;
 #endif
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    case cap_smc_cap:
+        if (cap_get_capType(cap_b) == cap_smc_cap) {
+            return true;
+        }
+        break;
+#endif
     }
     return false;
 }
--- a/src/arch/arm/64/object/objecttype.c
+++ b/src/arch/arm/64/object/objecttype.c
@@ -100,7 +100,19 @@ deriveCap_ret_t Arch_deriveCap(cte_t *slot, cap_t cap)
 
 cap_t CONST Arch_updateCapData(bool_t preserve, word_t data, cap_t cap)
 {
-    return cap;
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    if (cap_get_capType(cap) == cap_smc_cap) {
+        if (!preserve && cap_smc_cap_get_capSMCBadge(cap) == 0) {
+            return cap_smc_cap_set_capSMCBadge(cap, data);
+        } else {
+            return cap_null_cap_new();
+        }
+    } else {
+#endif
+        return cap;
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    }
+#endif
 }
 
 cap_t CONST Arch_maskCapRights(seL4_CapRights_t cap_rights_mask, cap_t cap)
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -208,6 +208,16 @@ config_option(
 )
 mark_as_advanced(KernelAArch64SErrorIgnore)
 
+config_option(
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow virtualized guests to make SMC calls. \
+    WARNING: Allowing SMC calls causes a couple of issues. Since seL4 cannot \
+    pre-empt the secure monitor, the WCET is no longer guaranteed. Also, since the \
+    secure monitor is a higher privilege level and can make any change in the \
+    system, the proofs can no longer be guaranteed."
+    DEFAULT OFF
+    DEPENDS "NOT KernelVerificationBuild; KernelArmHypervisorSupport"
+)
+
 if(KernelAArch32FPUEnableContextSwitch OR KernelSel4ArchAarch64)
     set(KernelHaveFPU ON)
 endif()
@@ -253,6 +263,7 @@ add_sources(
         object/iospace.c
         object/vcpu.c
         object/smmu.c
+        object/smc.c
         smp/ipi.c
 )
 
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -209,13 +209,13 @@ config_option(
 mark_as_advanced(KernelAArch64SErrorIgnore)
 
 config_option(
-    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow virtualized guests to make SMC calls. \
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow components to make SMC calls. \
     WARNING: Allowing SMC calls causes a couple of issues. Since seL4 cannot \
     pre-empt the secure monitor, the WCET is no longer guaranteed. Also, since the \
     secure monitor is a higher privilege level and can make any change in the \
     system, the proofs can no longer be guaranteed."
     DEFAULT OFF
-    DEPENDS "NOT KernelVerificationBuild; KernelArmHypervisorSupport"
+    DEPENDS "NOT KernelVerificationBuild; KernelSel4ArchAarch64"
 )
 
 if(KernelAArch32FPUEnableContextSwitch OR KernelSel4ArchAarch64)
--- a/src/arch/arm/object/smc.c
+++ b/src/arch/arm/object/smc.c
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2021, DornerWorks Ltd.
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+#include <config.h>
+
+#ifdef CONFIG_ALLOW_SMC_CALLS
+#include <arch/object/smc.h>
+
+exception_t decodeARMSMCInvocation(word_t label, unsigned int length, cptr_t cptr,
+                                   cte_t *srcSlot, cap_t cap, bool_t call, word_t *buffer)
+{
+    word_t i;
+    seL4_Word arg[NUM_SMC_REGS];
+    word_t *ipcBuffer;
+
+    switch (label) {
+    case ARMSMCCall:
+        for (i = 0; i < NUM_SMC_REGS; i++) {
+            arg[i] = getSyscallArg(i, buffer);
+        }
+
+        ipcBuffer = lookupIPCBuffer(true, NODE_STATE(ksCurThread));
+
+        register seL4_Word r0 asm("x0") = arg[0];
+        register seL4_Word r1 asm("x1") = arg[1];
+        register seL4_Word r2 asm("x2") = arg[2];
+        register seL4_Word r3 asm("x3") = arg[3];
+        register seL4_Word r4 asm("x4") = arg[4];
+        register seL4_Word r5 asm("x5") = arg[5];
+        register seL4_Word r6 asm("x6") = arg[6];
+        register seL4_Word r7 asm("x7") = arg[7];
+        asm volatile("smc #0\n"
+                     : "+r"(r0), "+r"(r1), "+r"(r2), "+r"(r3),
+                     "+r"(r4), "+r"(r5), "+r"(r6), "+r"(r7)
+                     :: "x8", "x9", "x10", "x11", "x12", "x13", "x14", "x15", "x16", "x17", "memory");
+
+        arg[0] = r0;
+        arg[1] = r1;
+        arg[2] = r2;
+        arg[3] = r3;
+        arg[4] = r4;
+        arg[5] = r5;
+        arg[6] = r6;
+        arg[7] = r7;
+
+        for (i = 0; i < n_msgRegisters; i++) {
+            setRegister(NODE_STATE(ksCurThread), msgRegisters[i], arg[i]);
+        }
+
+        if (ipcBuffer != NULL && i < NUM_SMC_REGS) {
+            for (; i < NUM_SMC_REGS; i++) {
+                ipcBuffer[i + 1] = arg[i];
+            }
+        }
+
+        setRegister(NODE_STATE(ksCurThread), msgInfoRegister, wordFromMessageInfo(
+                        seL4_MessageInfo_new(0, 0, 0, i)));
+
+        setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
+        return EXCEPTION_NONE;
+
+    default:
+        userError("ARMSMCInvocation: Illegal operation.");
+        current_syscall_error.type = seL4_IllegalOperation;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+}
+
+#endif
--- a/src/arch/arm/object/smc.c
+++ b/src/arch/arm/object/smc.c
@@ -8,7 +8,7 @@
 #ifdef CONFIG_ALLOW_SMC_CALLS
 #include <arch/object/smc.h>
 
-static exception_t performSMCCall(word_t *buffer)
+static exception_t invokeSMCCall(word_t *buffer)
 {
     word_t i;
     seL4_Word arg[NUM_SMC_REGS];
@@ -60,32 +60,23 @@ static exception_t performSMCCall(word_t *buffer)
 exception_t decodeARMSMCInvocation(word_t label, unsigned int length, cptr_t cptr,
                                    cte_t *srcSlot, cap_t cap, bool_t call, word_t *buffer)
 {
-    word_t badge;
-    word_t smc_func_id;
-
-    switch (label) {
-    case ARMSMCCall:
-        badge = cap_smc_cap_get_capSMCBadge(cap);
-        smc_func_id = getSyscallArg(0, buffer);
-
-        if (badge == smc_func_id) {
-            performSMCCall(buffer);
-        } else if (badge == 0) {
-            performSMCCall(buffer);
-        } else {
-            userError("ARMSMCInvocation: Illegal operation.");
-            current_syscall_error.type = seL4_IllegalOperation;
-            return EXCEPTION_SYSCALL_ERROR;
-        }
-
-        setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
-        return EXCEPTION_NONE;
-
-    default:
+    if (label != ARMSMCCall) {
         userError("ARMSMCInvocation: Illegal operation.");
         current_syscall_error.type = seL4_IllegalOperation;
         return EXCEPTION_SYSCALL_ERROR;
     }
+
+    word_t badge = cap_smc_cap_get_capSMCBadge(cap);
+    word_t smc_func_id = getSyscallArg(0, buffer);
+
+    if (badge != 0 && badge != smc_func_id) {
+        userError("ARMSMCCall: Illegal operation.");
+        current_syscall_error.type = seL4_IllegalOperation;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
+    return invokeSMCCall(buffer);
 }
 
 #endif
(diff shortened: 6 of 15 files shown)


libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -44,6 +44,16 @@
...
     </struct>
+    <struct name="seL4_ARM_SMCContext">
+        <member name="x0"/>
+        <member name="x1"/>
+        <member name="x2"/>
+        <member name="x3"/>
+        <member name="x4"/>
+        <member name="x5"/>
+        <member name="x6"/>
+        <member name="x7"/>
```
yyshen, line 55: might need to state that this only works for SMC calls that follow the SMC calling convention.

src/arch/arm/config.cmake
```diff
@@ -194,6 +194,12 @@ config_option(
 )
 mark_as_advanced(KernelAArch64SErrorIgnore)
 
+config_option(
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow virtualized guests to make SMC calls"
+    DEFAULT OFF
```
yyshen, line 199: probably need to enable the option that traps SMC calls?

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,93 @@
...
+exception_t decodeARMSMCInvocation(word_t label, unsigned int length, cptr_t cptr,
+                                  cte_t *srcSlot, cap_t cap, bool_t call, word_t *buffer)
+{
+    word_t badge;
+    word_t smc_func_id;
+
+    switch (label) {
+    case ARMSMCCall:
+        badge = cap_smc_cap_get_capSMCBadge(cap);
+        smc_func_id = getSyscallArg(0, buffer);
```
yyshen, line 69: so the badge is used to represent the authorised smc_func_id for a minted SMC Cap. by default, the original  SMC Cap can authorise all smc func calls, right?

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,93 @@
...
+
+    register seL4_Word r0 asm("x0") = arg[0];
+    register seL4_Word r1 asm("x1") = arg[1];
+    register seL4_Word r2 asm("x2") = arg[2];
+    register seL4_Word r3 asm("x3") = arg[3];
+    register seL4_Word r4 asm("x4") = arg[4];
+    register seL4_Word r5 asm("x5") = arg[5];
+    register seL4_Word r6 asm("x6") = arg[6];
+    register seL4_Word r7 asm("x7") = arg[7];
+    asm volatile("smc #0\n"
```
yyshen, line 31: have you considered to release the lock for the SMP kernel before the call and acquire the lock again after the call. The smc call can run arbitrary long.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -24,11 +24,12 @@ enum {
     seL4_CapDomain              = 11, /* global domain controller cap */
     seL4_CapSMMUSIDControl      = 12,  /*global SMMU SID controller cap, null cap if not supported*/
     seL4_CapSMMUCBControl       = 13,  /*global SMMU CB controller cap, null cap if not supported*/
+    seL4_CapSMC                 = 14,  /*global SMC cap, null cap if not supported*/
```
yyshen, line 27: It looks like one SMC cap is used for all CPU cores, not sure if this is coarse-grained.

src/arch/arm/object/smc.c
lsf37: According to the code, badge 0 authorises all, which would be the same convention as for endpoints. I think that makes sense. We should check though, if smc call 0 is reserved or may be needed, because otherwise anyone who needs to do that call can do all calls.

src/arch/arm/object/smc.c
lsf37: Hm, that is an important observation -- I hadn't thought about WCET on these yet at all, and without knowing more about individual calls, the kernel WCET for a system with SMC enabled is now arbitrarily high. That is even true on uni-core, but it is worse on SMP, because the kernel is blocked for all cores during that time.

This is different to other user-level behaviour that can take arbitrarily long, because this is now kernel time.

I don't think that releasing the lock while inside the kernel works in the current design: it wouldn't be a big-lock any more, and when you re-acquire the lock you would have not guarantee that the current TSC, SC, or IPC buffer still exists (some other core could have retyped them, for instance). I'm sure that could be worked-around, but it would definitely break the big-lock idea. And even if that should somehow work, it would still not solve the uni-core case.

That might throw a real spanner in the works. Do we need to turn this into some kind of more asynchronous mechanism?

libsel4/include/sel4/bootinfo_types.h
lsf37: That looks fine to me, as long as the intended pattern is that the system initialiser makes badged copies of this cap, handing these out to whoever needs them, restricted to core, sub-system or whichever is needed. Usually the 0-badge version of the cap should stay in the initialiser, or in a high-privilege management component.

That scenario does suggest that we might want more derivation levels for this cap, though, like for endpoint caps.

libsel4/arch_include/arm/interfaces/sel4arch.xml
```diff
@@ -23,7 +23,7 @@
             <param dir="in" name="vaddr" type="seL4_Word"
             description="Virtual address to map the page into."/>
             <param dir="in" name="attr" type="seL4_ARM_VMAttributes">
-                <description> 
+                <description>
                     VM Attributes for the mapping. <docref>Possible values for this type are given in <autoref label="ch:vspace"/>  .</docref>
```
axel-h, line 27: Could you move these space changes to a dedicated commit?

libsel4/include/sel4/bootinfo_types.h
axel-h, line 27: Is this a potentially breaking change that should be noted in the readme?

libsel4/arch_include/arm/interfaces/sel4arch.xml
Furao: Yeah, I can do that.

libsel4/include/sel4/bootinfo_types.h
Furao: That is a good point I hadn't considered. Adding to the end avoids breaking old layouts, but that becomes much harder to do with the sched_context cap getting configured out. Maybe positions should be constant whether their features are enabled or not? 
ex:
```
enum {
    seL4_CapNull                =  0, /* null cap */
    seL4_CapInitThreadTCB       =  1, /* initial thread's TCB cap */
    seL4_CapInitThreadCNode     =  2, /* initial thread's root CNode cap */
    seL4_CapInitThreadVSpace    =  3, /* initial thread's VSpace cap */
    seL4_CapIRQControl          =  4, /* global IRQ controller cap */
    seL4_CapASIDControl         =  5, /* global ASID controller cap */
    seL4_CapInitThreadASIDPool  =  6, /* initial thread's ASID pool cap */
    seL4_CapIOPortControl       =  7, /* global IO port control cap (null cap if not supported) */
    seL4_CapIOSpace             =  8, /* global IO space cap (null cap if no IOMMU support) */
    seL4_CapBootInfoFrame       =  9, /* bootinfo frame cap */
    seL4_CapInitThreadIPCBuffer = 10, /* initial thread's IPC buffer frame cap */
    seL4_CapDomain              = 11, /* global domain controller cap */
    seL4_CapSMMUSIDControl      = 12,  /*global SMMU SID controller cap, null cap if not supported*/
    seL4_CapSMMUCBControl       = 13,  /*global SMMU CB controller cap, null cap if not supported*/
    seL4_CapInitThreadSC        = 14, /* initial thread's scheduling context cap */
    seL4_CapSMC                    = 15,  /*global SMC cap, null cap if not supported*/
    seL4_NumInitialCaps         = 16
};
```

Then if they feature isn't enabled, it just shows up as a nullcap?

src/arch/arm/object/smc.c
axel-h: if the short term goal is, that the feature is to be merged (behind a config option) , I'm fine with the comment https://sel4.atlassian.net/browse/RFC-9?focusedCommentId=19762 from @lsf37 that for now the system designer needs to be aware of the potential pitfalls. We can look into this as a follow-up then.

Furao: The warning message has been added here: https://github.com/seL4/seL4/pull/701/files#diff-bc9402a3a9f4e827bf5919980dd881d416c5a6225a82a380af72117a605ae299R198-R203

src/arch/arm/64/object/objecttype.c
```diff
@@ -495,6 +501,10 @@ exception_t Arch_decodeInvocation(word_t label, word_t length, cptr_t cptr,
     case cap_cb_cap:
         return decodeARMCBInvocation(label, length, cptr, slot, cap, call, buffer);
 #endif /*CONFIG_ARM_SMMU*/
+#ifdef CONFIG_ALLOW_SMC_CALLS
```
axel-h, line 516: I think we have to add this in line 488 also

Furao: CONFIG_ALLOW_SMC_CALLS has a dependency on CONFIG_ARM_HYPERVISOR_SUPPORT, so it works how it is. However, we can be more explicit and add it into 488 if you would like.

axel-h: I think having this explicitly improves maintainability. How strong us the dependency of the SMC feature on the Hypervisor feature - could they be separated one day?

Furao: We have actually already seen a use case to uncouple this from virtualization support, so we updated the specified ifdef line and changed the config dependency.

Furao:
@axel-h Ping

axel-h:
Could you rebase this in the current master?

libsel4/include/sel4/bootinfo_types.h
axel-h: @lsf37, @kent-mcleod; What do you think about this? 

A change here might also require updating `SIMPLE_SKIPPED_INIT_CAPS` here:
- https://github.com/seL4/seL4_libs/blob/master/libsel4simple/arch_include/arm/simple/arch/simple.h#L17-L19 (note that there is no RISC-V here, it uses the ARM files)
- https://github.com/seL4/seL4_libs/blob/master/libsel4simple/arch_include/x86/simple/arch/simple.h#L17-L23

Furao:
Ok, rebased.

libsel4/include/sel4/bootinfo_types.h
Furao: @lsf37, @kent-mcleod; ping.

Furao: @lsf37, @kent-mcleod; pinging again.

src/arch/arm/config.cmake
```diff
@@ -194,6 +194,17 @@ config_option(
...
 
+config_option(
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow components to make SMC calls.
+    WARNING: Some platforms require SMC calls to actually be made to boot VMs \
+    correctly, however allowing SMC calls causes a couple of issues. Since seL4 \
+    cannot pre-empt the supervisor the WCET is no longer guaranteed. Also, \
+    since the supervisor is a higher privilege level and can make any change \
+    in the system, the proofs can no longer be guaranteed."
+    DEFAULT OFF
+    DEPENDS "KernelSel4ArchAarch64"
```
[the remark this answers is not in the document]
Furao: Yeah, that sounds like a good idea. I will make that update.

libsel4/include/sel4/bootinfo_types.h
lsf37: Apologies for the long silence. As commented in #1088, from the verification side that would be Ok.

kent-mcleod: Once this is rebased on the changes from #1088, this should be resolved?

kent-mcleod:
@Furao, I added some suggested changes but can't PR off this PR due to a rebase.  https://github.com/seL4/seL4/compare/master...kent-mcleod:seL4:kent/smc.  Are you happy to pull these additional commits in?  In addition to some changes to the invocation implementation to more closely follow the structures required by the verification specs, I've made it so that seL4_CNode_Revoke should do the right thing:
- Revoking the root's original smc cap should delete all SMC caps in the system,
- Revoking the first badged smc cap should delete any SMC caps derived from it.

src/arch/arm/config.cmake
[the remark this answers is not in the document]
Furao: Updated this dependency in the latest.

src/object/cnode.c
```diff
@@ -806,6 +806,21 @@ bool_t PURE isMDBParentOf(cte_t *cte_a, cte_t *cte_b)
         break;
     }
 
+#ifdef CONFIG_ALLOW_SMC_CALLS
```
kent-mcleod, line 809: This would benefit from some arch-specific functions to call out to, but they don't currently exist... Thoughts?

src/object/cnode.c
```diff
@@ -806,6 +806,21 @@ bool_t PURE isMDBParentOf(cte_t *cte_a, cte_t *cte_b)
...
+    case cap_smc_cap: {
+        word_t badge;
+
+        badge = cap_smc_cap_get_capSMCBadge(cte_a->cap);
+        if (badge == 0) {
+            return true;
+        }
+        return
```
Indanz, line 817: The formatting of this return is very weird. I know that cap_notification_cap does the same, but I don't think there is a reason to spread this further.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -207,4 +217,18 @@
             </error>
         </method>
     </interface>
+    <interface name="seL4_ARM_SMC" manual_name="SMC" cap_description="Capability to allow threads to interact with SMC calls.">
+        <method id="ARMSMCCall" name="Call" manual_name="SMC Call" manual_label="smc_call">
+            <brief>
+                Tell the microkernel to make the real SMC call.
+            </brief>
+            <description>
+                <docref>See <autoref label="sec:read_write_registers"/></docref>
```
Indanz, line 226: Is this reference still up-to-date, considering that `seL4_ARM_SMCContext` has been added?

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,89 @@
...
+                 : "+r"(r0), "+r"(r1), "+r"(r2), "+r"(r3)
+                 : "r"(r4), "r"(r5), "r"(r6), "r"(r7)
```
Indanz, line 32: According to DEN0028E_SMC_Calling_Convention_1.4eac0:

- A Function Identifier is passed in register W0.
- Arguments are passed in registers W1-W7.
- Results are returned in W0-W7.
- Registers W4-W7 must be preserved unless they contain results, as specified in the function definition.

(Same for 64-bit calls, but with X0-X17 instead.)

So I think both the output and input lists should be extended to the r0-r7 registers, both with `+r`.

Also, if only the 32-bit calling convention is supported, the code should check that bit 30 is not set in `smc_func_id`.

Furao: So we are actually targeting a specific version of the SMCCC spec. v1.0. This document can be viewed here: https://developer.arm.com/documentation/den0028/a/?lang=en (The `a` in the URL should give you the correct version when you download.). You can query the SMCCC version number on your platform with an SMC call. We did our development on a Zynq UltraScale+ MPSoC, which implements spec v1.0. 

In this version, r4-r7 is just scratch space, so it won't ever be a result. 
Also, since we are using `x0-x7`, we are actually using the 64-bit calling convention.

This does mean that if future processors conform to newer calling convention versions, than we need to create version  specific implementations.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -207,4 +217,18 @@
             </error>
         </method>
     </interface>
+    <interface name="seL4_ARM_SMC" manual_name="SMC" cap_description="Capability to allow threads to interact with SMC calls.">
```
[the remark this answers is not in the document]
Furao: Agreed. Updated.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
Furao: Updated the description accordingly.

src/object/cnode.c
Furao: Do you just mean that its weird that the keyword is on its own line?

src/object/cnode.c
Furao: I suppose it could also be wrapped in an architecture ifdef so that if we ever make an architecture specific `cnode.c` file, then it is obvious that chunk of code should move.

src/object/cnode.c
Indanz: Yes.

src/arch/arm/object/smc.c
Indanz: We don't want version specific implementations. This is a new multi-platform feature, it should be as future proof as possible, which means following the latest spec. If in the future a platform does return data in registers x4-x7, with the current code it will silently clobber a register and that will be extremely hard to debug if it causes weird problems.

If you do follow the 64-bit calling convention, a lot more registers get clobbered, see table 3-1. The normal procedure call conventions are used, registers `x0-x17`  may be modified. On the upside, you don't have to check bit 30, as you can do 32-bit calls with the 64-bit convention.

With that in mind, maybe it's better to add a `do_smc` function to assembly and call it from here, I think that will get all clobbers and assumptions right automatically.

src/object/cnode.c
kent-mcleod: I thought that the original weirdness is due to an autostyle edge case. Turns out it's not, as the autostyle accepts it on the same line.. Agree we should change it.

kent-mcleod:
> Yeah, I can pull those in.

Thanks.  I'd be happy to apply any further fixes myself, but the PR branch permissions don't allow modifications by reviewers.

src/arch/arm/object/smc.c
Indanz: I'm fine with limiting it to `x0-x7` for now, as it will take ages before more is practically needed, and when the limit is reached, it's easy to extend. On the other hand, it's probably not much more work to extend it now.

>  I am not entirely sure what an assembly function gets us over the inline assembly we are already doing. The context still needs to be transferred in/out of the registers?

Doing the SMC call would be just a normal function call then. But I forgot that there is a need to transfer the result too, so the advantage would be much less, if any.

Furao: Playing around with with using all 18 registers revealed to me that inline assembly calls can only have 30 operands, so 18 inputs + 18 outputs doesn't work. Path of least resistance is probably to just modify `r4-r7` to be `r+`, so it is an input and output. 

Or if we really wanted to support all 18 registers, we would need to write a proper assembly function. Thoughts?

Furao: I implemented the simpler approach with my latest push.

Indanz: Almost there, you also need to add all other modified registers to the clobber list, see Table 3-1.

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,90 @@
...
+    register seL4_Word r3 asm("x3") = arg[3];
+    register seL4_Word r4 asm("x4") = arg[4];
+    register seL4_Word r5 asm("x5") = arg[5];
+    register seL4_Word r6 asm("x6") = arg[6];
+    register seL4_Word r7 asm("x7") = arg[7];
+    asm volatile("smc #0\n"
+                 : "+r"(r0), "+r"(r1), "+r"(r2), "+r"(r3),
+                 "+r"(r4), "+r"(r5), "+r"(r6), "+r"(r7)
+                 :
+                 : "memory");
```
Indanz, line 34:
```suggestion
                 :: "x8", "x9", "x10", "x11", "x12", "x13", "x14", "x15", "x16", "x17", "memory");
```

Furao: Made the suggested change

src/arch/arm/object/smc.c
Furao: I appreciate the input. This implementation is much more future-proof now.

axel-h:
I might be good to mention the new `seL4_CapSMC` and [RFC-9](https://sel4.atlassian.net/browse/RFC-9) in the [CHANGES](https://github.com/seL4/seL4/blob/master/CHANGES) file also.

src/arch/arm/config.cmake
```diff
@@ -208,6 +208,17 @@ config_option(
 )
 mark_as_advanced(KernelAArch64SErrorIgnore)
 
+config_option(
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow components to make SMC calls.
+    WARNING: Some platforms require SMC calls to actually be made to boot VMs \
+    correctly, however allowing SMC calls causes a couple of issues. Since seL4 \
+    cannot pre-empt the supervisor the WCET is no longer guaranteed. Also, \
```
axel-h, line 215:
```suggestion
    cannot pre-empt the supervisor, the WCET is no longer guaranteed. Also, \
```

axel-h: Also, I think we should use "monitor" (or "secure monitor") instead of "supervisor" here.

src/arch/arm/config.cmake
```diff
@@ -208,6 +208,17 @@ config_option(
 )
 mark_as_advanced(KernelAArch64SErrorIgnore)
 
+config_option(
+    KernelAllowSMCCalls ALLOW_SMC_CALLS "Allow components to make SMC calls.
+    WARNING: Some platforms require SMC calls to actually be made to boot VMs \
```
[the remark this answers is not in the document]
Furao: Agreed. I removed that line.

src/arch/arm/config.cmake
Furao: Updated both suggestions.

Furao:
> I might be good to mention the new `seL4_CapSMC` and [RFC-9](https://sel4.atlassian.net/browse/RFC-9) in the [CHANGES](https://github.com/seL4/seL4/blob/master/CHANGES) file also.

Added a new commit to update CHANGES file.

src/arch/arm/config.cmake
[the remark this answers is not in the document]
Indanz: Agreed. Other than doing standard PSCI calls, SCM is used for platform specific things. People needing SCM will know whether they need it or not.

Furao:
I just rebased onto the latest and I believe I have addressed all of the open comments. Would be nice to get this through.

axel-h (approved):
From my side this is fine now. I lack the resources to test this in more detail, tough. 
You could squash the last commit with the CHANGES update into the commit that adds the capability.

kent-mcleod (approved):

kent-mcleod:
> I just rebased onto the latest and I believe I have addressed all of the open comments. Would be nice to get this through.

I'll wait a couple more days for feedback from other reviewers and otherwise merge.

Ivan-Velickovic (commented):
The single comment I had has been addressed, but I’m not qualified to review/approve the actual contents of the patch so I’ll leave that for the others.

lsf37 (approved):
No further issues from my side. If we had a model of what SMC calls do this does even look fairly verifiable. 

(The preprocess check triggers on the change of `seL4_NumInitialCaps` which should not affect the proofs)

lsf37:
Have remove the proof-test label to preserve AWS instances, because the proofs have passed with this PR now. It looks like the build failed with sel4test because the PR needs to be rebased first to work with the latest libsel4.

src/arch/arm/object/smc.c
```diff
@@ -8,18 +8,17 @@
...
 {
     word_t i;
     seL4_Word arg[NUM_SMC_REGS];
-    word_t *ipcBuffer;
 
     for (i = 0; i < NUM_SMC_REGS; i++) {
         arg[i] = getSyscallArg(i, buffer);
     }
 
-    ipcBuffer = lookupIPCBuffer(true, NODE_STATE(ksCurThread));
```
Indanz: I don't think you can safely remove this, however tempting:

 > Require the input length to be >= NUM_SMC_REGS. This ensures that any  accesses to the IPC buffer memory is safe (if the IPC buffer isn't  present then the message length will be truncated to n_msgRegisters).

This misses the corner case where the IPC buffer is mapped read-only.

Indanz (commented):
I'd add a back reference to this pull request to the first commit, so people can find this easily when going through the code or git log.

src/arch/arm/object/smc.c
kent-mcleod: Yea you're right good catch. TIL that IPC buffer caps can be without write rights.  (seems like such an anti-feature)

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -207,4 +217,20 @@
...
+            </brief>
+            <description>
+                Takes x0-x7 as arguments to an SMC call which are defined as a seL4_ARM_SMCContext
+                struct. The microkernel makes the SMC call and then returns the results as a
+                new seL4_ARM_SMCContext.
+            </description>
+            <param dir="in" name="smc_args" type="seL4_ARM_SMCContext"
+                description="The structure that has the provided arguments."/>
+            <param dir="out" name="smc_response" type="seL4_ARM_SMCContext"
+                description="The structure to capture the responses."/>
```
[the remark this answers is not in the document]
Furao: Correct, `inout` is not supported. 

`syscall_stub_gen.py` does the parsing:
```py
assert (param_dir == "in") or (param_dir == "out")
```

src/arch/arm/object/smc.c
Furao: Added it back in

Furao:
> From my side this is fine now. I lack the resources to test this in more detail, tough. You could squash the last commit with the CHANGES update into the commit that adds the capability.

Squashed that change with the first one.

> I'd add a back reference to this pull request to the first commit, so people can find this easily when going through the code or git log.

Updated the first commit message with the link to this PR.

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,94 @@
...
+    word_t smc_func_id = getSyscallArg(0, buffer);
+
+    if (badge != 0 && badge != smc_func_id) {
+        userError("ARMSMCCall: Illegal operation.");
+        current_syscall_error.type = seL4_IllegalOperation;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
+    return invokeSMCCall(buffer, call);
```
Indanz, line 91: `buffer` is not used any more, perhaps move the `lookupIPCbuffer()` call to here and pass that instead of buffer? Or remove the buffer argument, like you had before.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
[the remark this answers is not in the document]
Indanz: Oh well, so be it, it's not the end of the world and can be changed later if we want.

Indanz (approved):
Looks good now.

Would be nice if:

https://github.com/seL4/sel4test/pull/85
https://github.com/seL4/sel4test/pull/101

would get merged together, but it's fine if they get merged upstream separately. Could you do a review of Kent's SMC tests?

src/arch/arm/object/smc.c
```diff
@@ -0,0 +1,94 @@
...
+    seL4_Word arg[NUM_SMC_REGS];
+    word_t *ipcBuffer;
+
+    for (i = 0; i < NUM_SMC_REGS; i++) {
+        arg[i] = getSyscallArg(i, buffer);
+    }
+
+    ipcBuffer = lookupIPCBuffer(true, NODE_STATE(ksCurThread));
+
+    register seL4_Word r0 asm("x0") = arg[0];
```
axel-h, line 25: Maybe this should have a compile time assert that `NUM_SMC_REGS` is at least 8. It's unlikey to ever change, but may catch cases where somebody is experimenting. I'd expect the compiler to issue a out-of-warning in this case automatically, though - but does this also break the build with the current defaults?

include/arch/arm/arch/object/structures.h
```diff
@@ -12,6 +12,16 @@
 
 static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 {
+#ifdef CONFIG_ALLOW_SMC_CALLS
+    switch (cap_get_capType(derivedCap)) {
+    case cap_smc_cap:
+        return (cap_smc_cap_get_capSMCBadge(derivedCap) !=
+                cap_smc_cap_get_capSMCBadge(srcCap));
+
```
Indanz, line 20: I would remove this empty line, it looks a bit weird.

src/arch/arm/object/smc.c
kent-mcleod: If you want to make that in a separate PR then I'd be happy to accept it.

axel-h: will do

Ivan-Velickovic:
My understanding is that this PR does not support the SMC64 calling convention, is that correct?

Ivan-Velickovic:
> > My understanding is that this PR does not support the SMC64 calling convention, is that correct?
> 
> No, it's not. You can do both 32-bit and 64-bit SMC calls with this interface. However, it's only implemented for Aarch64, not 32-bit ARM.

The reason I asked is being the SMC specification I was looking at (downloaded from here https://developer.arm.com/documentation/den0028/f?lang=en) says that for SMC64/HVC64 (section 2.7), registers x1-x17 are used for arguments and x0-17 are used for results. This patch takes in x0-x7 as arguments/results.

I *think* I've found a case where I need SMC64, so if I'm happy to add support for the SMC64 calling convention, assuming I've understood everything correctly.

Ivan-Velickovic:
> Technically, we are supporting 32-bit and 64-bit SMC calls but for an earlier version of the spec (v1.0) which didn't include x8-x17 for arguments or results. I looked at adding the rest of the registers, but ran into an issue with inline assembly.

Yep, I just read through the spec changes and saw that in August 2020 it was introduced to use x8-x17 as arguments/results. That makes sense, now. Although, we should document this. I assume then issue B/version 1.1 of the spec was used?

> To support this, you need to do a refactor and write a new function as just assembly.

Yes I ran into the same issue.

Ivan-Velickovic:
Ah GitHub hides all the existing conversations on the calling convention which is why I missed it, my bad.

Furao:
> Ah GitHub hides all the existing conversations on the calling convention which is why I missed it, my bad.

No problem. I knew to look for the comment and it was hard for me to find as well.

> > Technically, we are supporting 32-bit and 64-bit SMC calls but for an earlier version of the spec (v1.0) which didn't include x8-x17 for arguments or results. I looked at adding the rest of the registers, but ran into an issue with inline assembly.
> 
> Yep, I just read through the spec changes and saw that in August 2020 it was introduced to use x8-x17 as arguments/results. That makes sense, now. Although, we should document this. I assume then issue B/version 1.1 of the spec was used?

Yes documentation on the version would be good. Yes, I think 1.1 is technically the latest one we support.

PR #702, opened 2021-12-11
trivial: fix typos in comment

axel-h:


lsf37 (approved):

lsf37:
Thanks for the fix -- please don't add `hw-test` for comment PRs, there is only a limited number of boards and all hardware tests in the seL4 org compete for them.

axel-h:
I've added the label to have a reference run for the other pending PRs.

lsf37:
Fair enough, they have been failing a lot. All runs should succeed, apart from this one: https://github.com/seL4/sel4test/issues/64 (I *think* have seen it succeeding in the new build environment, but it fails a lot and really shouldn't)

axel-h:
But using the label here was not necessary indeed, there seems to be the deps-update that acts as reference. The status of it is just hidden a bit. Maybe having a more prominent status page of it similar to the benchmark history idea https://github.com/seL4/website/issues/158 would be handy.

lsf37:
Yes, although in theory these are the tests that should be always passing. In practice, theory and practice are not the same :-)

PR #704, opened 2021-12-14
boot/risc-v: pv_offset cannot be uint32_t

axel-h:
The phys/virt translation fails on 64-bit systems when uint32_t is used. With `virt_address = 0xc0000000` and `phys_address = 0x0` the calculation
```
   pv_offset = phys_address - virt_address
             = 0x0 - 0xc0000000 = 0xffffffff40000000
````
result truncated to `uint32_t` is `0x40000000`. Then the phys-to-virt reverse calculation
```
  virt_address = phys_address - pv_offset
               = 0 - 0x40000000 = fffffffc0000000
```
results in a wrong virtual address. Using the proper `(s)word_t` type
```
  virt_address = phys_address - pv_offset
               = 0 - 0xffffffff40000000 = 0xc0000000
```
gives the correct result.
Note that the C rules define that sword_t is converted to word_t in this operation automatically, thus using sword_t or word_t does not make any difference here. Using a signed type for offset is just more intuitive and aligned with what is done in the ARM port of this code.

### DIFF
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -181,7 +181,7 @@ BOOT_CODE static void release_secondary_cores(void)
 static BOOT_CODE bool_t try_init_kernel(
     paddr_t ui_p_reg_start,
     paddr_t ui_p_reg_end,
-    uint32_t pv_offset,
+    sword_t pv_offset,
     vptr_t  v_entry,
     paddr_t dtb_phys_addr,
     word_t  dtb_size


lsf37 (approved):
Yes, it does not really make sense to have only a 32 bit offset on a 64 bit kernel.

Do we need documentation that this changes the interface? IIRC, `try_init_kernel` is the top-level function.

axel-h:
`try_init_kernel()` is an internal function that is no accessible from the outside this file. It is called from `init_kernel()` in the same file, which is the C entry point. 
There was a similar issues in the ELF-Loader when it calculated the parameters, but that gt fixed already in https://github.com/seL4/seL4_tools/pull/100 a while ago, the `int32_t` there is now `word_t`. On the ARM port nothing has changed at all and things were always fine, it was mostly the RISC-V port that had some issues here due to the lack of practical test coverage.

kent-mcleod (approved):

axel-h:
CI test after merge: https://github.com/seL4/seL4/actions/runs/1589251898

PR #705, opened 2021-12-15
cmake: invoke python3 explicitly

mbrcknl:
When running a python script from CMake, explicitly invoke Python via
the `PYTHON3` variable defined in the seL4 CMake config, rather than
relying on the script to specify its interpreter with `#!`.

This makes it easier to override the Python interpreter used for all
Python scripts called from CMake builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -557,7 +557,7 @@ cppfile(
 
 add_custom_command(
     OUTPUT circular_includes_valid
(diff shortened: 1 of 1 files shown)


axel-h:
Make sense to enforce python3 everywhere give https://github.com/seL4/seL4_tools/pull/133 was just merged.

axel-h (approved):

PR #707, opened 2021-12-15
boot: sanity check for SMP lock variable state

axel-h:
- Describe the SMP lock variable mechanism detail.
- Add checks that the lock state is valid, this will catch potential kernel loader problems.

This might catch some porting issues, I was running into this with a broken loader.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -252,15 +252,13 @@ BOOT_CODE static void init_plat(void)
 #ifdef ENABLE_SMP_SUPPORT
 BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 {
-    unsigned i;
-
     /* need to first wait until some kernel init has been done */
     while (!node_boot_lock);
 
     /* Perform cpu init */
     init_cpu();
 
-    for (i = 0; i < NUM_PPI; i++) {
+    for (unsigned int i = 0; i < NUM_PPI; i++) {
         maskInterrupt(true, CORE_IRQ_TO_IRQT(getCurrentCPUIndex(), i));
     }
     setIRQState(IRQIPI, CORE_IRQ_TO_IRQT(getCurrentCPUIndex(), irq_remote_call_ipi));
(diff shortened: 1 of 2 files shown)


src/arch/arm/kernel/boot.c
```diff
@@ -29,8 +29,16 @@
...
- * until kernel data structures initialized */
+/* The SMP boot synchronization uses a simple mechanism based on a global
+ * variable with the well defined initial value 0. Secondary cores simply keep
+ * spinning until the primary node has initialized all kernel structures and
+ * then set this variable to 1. This mechanism works, because the C rules define
+ * that uninitialized global variables are zero. Technically, such variables are
+ * put either in the BSS segment filled with zeros or the data segment. Both
+ * segments are set up by a previous kernel loader. Any runtime initialization
+ * in the kernel boot code wont work for the SMP synchronization due to a race
+ * condition if all nodes start in an undefined order.
```
Indanz, line 40: First part is fine, but the second part is a bit verbose without being clear. What about:

"This mechanism only works if the kernel loader initialised the variable to zero correctly."?

But the above could be said of most kernel functionality. If the real problem is checking if BSS and data sections are initialised correctly, shouldn't we add explicit sanity checks for all known sections, instead of checking one random variable which happened to cause issues in the past? That would also do checking for non-SMP,

axel-h: You have a point there. I created this PR for the specific issue I ran into back then. In the end it's a definition of the ABI the kernel expects. I have a bias for not expecting anything, but this come from our "seL4 in ROM" where there is not really a loader. There is just power up ROM code that should better wipe memory contents for security reasons. But that costs boot time that we had concerns about. So it falls to non-ROM boot code that should do this.

kent-mcleod: @axel-h are you able to reduce this comment and the RISC-V one to just the first part:
```
The SMP boot synchronization uses a simple mechanism based on a global
 * variable with the well defined initial value 0. Secondary cores simply keep
 * spinning until the primary node has initialized all kernel structures and
 * then set this variable to 1.
 ```

axel-h: I've made this a bit shorter. But I think it's work mentioning here that BSS must be zero'd by the loader, otherwise this will not work. For most other BSS variables we have some kind if init option, so we don't depend on a specific init values and this is less critical.

kent-mcleod (approved):

PR #709, opened 2021-12-17
consolidate idle and arch thread prototypes

axel-h:
Consolidate arch thread function prototypes. They are the same on all architectures and thus there is no need to duplicate this in the architecture specific headers

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -31,7 +31,7 @@ jobs:
           - arch: X64
             session: CRefine
     # test only most recent push to PR:
-    concurrency: seL4-PR-C-proofs-${{ github.ref }}-${{ strategy.job-index }}
+    concurrency: seL4-PR-C-proofs-pr-${{ github.event.number }}-idx-${{ strategy.job-index }}
     steps:
     - name: Proofs
       uses: seL4/ci-actions/aws-proofs@master
--- a/include/arch/arm/arch/kernel/thread.h
+++ b/include/arch/arm/arch/kernel/thread.h
@@ -6,11 +6,4 @@
 
 #pragma once
 
-#include <object.h>
 #include <mode/kernel/thread.h>
-
-void Arch_switchToThread(tcb_t *tcb);
-void Arch_switchToIdleThread(void);
-void Arch_configureIdleThread(tcb_t *tcb);
-void Arch_activateIdleThread(tcb_t *tcb);
-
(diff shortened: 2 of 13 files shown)


lsf37:
The removal of the `config.h` include leads to a lot of re-ordering that is visible to the proofs, but they should still pass, let's see what the test says.

Otherwise I would agree that it makes sense to pull the declarations up, not because they are currently the same on all architectures, but because they have to be the same if this function is supposed to be an interface.

axel-h:
you mean `object.h`?

lsf37:
> you mean `object.h`?

I do indeed. Don't let me operate any heavy machinery today :-)

axel-h:
Adding two more commit to consolidate things
- consolidate idle thread prototypes 
- remove define idleThreadStart (unless this is something the proofs need?)

axel-h:
It's no longer running the proofs when I rebase this, the message is "Canceling since a higher priority waiting request for 'seL4-PR-C-proofs-refs/heads/master-1' exists". Is this conflicting with another PR now?

lsf37:
> It's no longer running the proofs when I rebase this, the message is "Canceling since a higher priority waiting request for 'seL4-PR-C-proofs-refs/heads/master-1' exists". Is this conflicting with another PR now?

Hm, this looks like a bug in the workflow. It shouldn't be `refs/heads/master`, but `refs/pulls/709`. Will investigate.

lsf37:
> - consolidate idle thread prototypes
> - remove define idleThreadStart (unless this is something the proofs need?)

It looks like something done for proof benefit, but it is not needed any more. Back in the dark ages at the very beginning of the verification, the C subset did not include taking the address of global variables either (which this would guard). It might be a remnant of that.

lsf37 (approved):
When the proofs pass, I'm fine with this change.

yyshen (approved):

axel-h:
The proofs still do not run, error remains "Canceling since a higher priority waiting request for 'seL4-PR-C-proofs-refs/heads/master-1' exists"

axel-h:
rebasing on top of https://github.com/seL4/seL4/pull/745 to see if this fixes the issue that the proofs do not run.

axel-h:
Will retry running the proofs later today, seem AWS is having issues
```
  Starting AWS instance...
   An error occurred (InsufficientInstanceCapacity) when calling the RunInstances operation (reached max retries: 2): We currently do not have sufficient c5.4xlarge capacity in zones with support for 'gp2' volumes. Our system will be working on provisioning additional capacity.
```

axel-h:
Seem the other proofs are fine with this change.

lsf37:
They have now all passed.

PR #710, opened 2021-12-19
risc-v: print kernel entry reason on halt

axel-h:
This also aligns RISC-V behavior with ARM port.


lsf37 (approved):

lsf37:
Hopefully we don't see that one a lot..

yyshen (approved):

PR #711, opened 2021-12-19
 improve SMP boot arch handling and documentation

axel-h:
- The cache maintenance is needed on `AARCH32`, so check explicitly for this architecture instead of doing this everywhere except `AARCH64`.
- Do nothing on `AARCH64` and add a build breaker for future unknown architectures this code might be compiled for.
- Improve the documentation to describe the potential pitfall here and why/when explicit action might be needed.
- RISC-V: Improve the documentation to describe about the potential pitfall in SMP boot and the non-recommended barrier that is used.

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -299,7 +299,7 @@ BOOT_CODE static void release_secondary_cpus(void)
      * On AARCH32 the elfloader uses strongly ordered uncached memory, but seL4
      * has caching enabled, thus explicit cache cleaning is required.
      */
-#ifndef CONFIG_ARCH_AARCH64
+#ifdef CONFIG_ARCH_AARCH32
     cleanInvalidateL1Caches();
     plat_cleanInvalidateL2Cache();
 #endif
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -284,20 +284,22 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 
 BOOT_CODE static void release_secondary_cpus(void)
 {
-
     /* release the cpus at the same time */
     node_boot_lock = 1;
 
-#ifndef CONFIG_ARCH_AARCH64
-    /* At this point in time the other CPUs do *not* have the seL4 global pd set.
-     * However, they still have a PD from the elfloader (which is mapping memory
-     * as strongly ordered uncached, as a result we need to explicitly clean
-     * the cache for it to see the update of node_boot_lock
+    /*
+     * At this point in time the primary core (executing this code) already uses
+     * the seL4 MMU/cache setup. However, the secondary cores are still using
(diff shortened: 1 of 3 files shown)


src/arch/riscv/kernel/boot.c
```diff
@@ -169,6 +169,18 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(word_t hart_id, word_t co
...
+     * the seL4 MMU/cache setup. However, the secondary cores are still using
+     * the elfloader's MMU/cache setup, and thus the update of node_boot_lock
+     * may not be visible there if the setups differ. Currently, the mappings
+     * match, so a write-before-read fence below is all that is needed. It acts
+     * as a barrier to ensure the write really happens and becomes globally
+     * visible to make the secondary harts boot before we start the polling loop
+     * that checks ksNumCPUs to determine when all nodes are up. However, the
+     * RISC-V Unprivileged ISA spec (V20191214-draft, section A.3.6 "Fences")
+     * states that "fence w,r" is one of the uncommon combination and thus not
+     * recommended to be used.
```
Indanz, line 182: As this happens once during bootup, I would use the strongest fence there is.

src/arch/arm/kernel/boot.c
```diff
@@ -284,20 +284,19 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
...
-     * (so is the attribute of the seL4 global PD) when SMP is enabled, and
-     * turns on the cache. Thus, we do not need to clean and invalidate the cache.
+    /* At this point in time the primary core (executing this code) already uses
+     * the seL4 MMU/cache setup. However, the secondary cores are still using
+     * the elfloader's MMU/cache setup, and thus the update of node_boot_lock
+     * may not be visible there if the setups differ. On AARCH32 the elfloader
+     * uses strongly ordered uncached memory, but seL4 has caching enabled, thus
+     * explicit cache cleaning is required. On AARCH64, both elfloader and seL4
+     * map memory inner shareable and have the caches enabled, so no explicit
+     * cache maintenance is necessary.
```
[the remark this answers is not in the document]
axel-h: I have applied your changes. 

When the primary core arrives here, is has done a cache flush already for all the things is did. All that is left to be done here is really the release of the other cores - so just the change to `node_boot_lock` has to become visible. But you have a point, better put a barrier here to ensure the write really happen. Some time ago I did some experimenting into that direction with https://github.com/Hensoldt-Cyber/seL4/pull/9/commits/9026e8ca0ea9e002fffb8cbcf65e960c7967672a

src/arch/riscv/kernel/boot.c
axel-h: I can add a commit on top that puts `fence_rw_rw()` there and cleans up the comment then. But I'd prever to get this PR merged first to have some more documentation in the code.

Indanz (approved):
Seems fine to me now.

kent-mcleod (approved):

PR #713, opened 2021-12-20
Add numDomains enumeration constant

Xaphiosis:
Defined to be equal to CONFIG_NUM_DOMAINS. seL4 makes control-flow
decisions based on whether the number of domains is greater than 1. To
perform refinement proofs independent of the number of domains, we need
to follow both branches of these if statements, pretending we don't know
which branch will be taken. This is made significantly harder when
preprocessed C code ends up with comparisons like `if (16 > 1)`.
By adding a numDomains that appears in the C code, we obtain a name we
can point to and link up to higher level specifications.

### DIFF
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -25,7 +25,11 @@ typedef uint64_t time_t;
 
 enum domainConstants {
     minDom = 0,
-    maxDom = CONFIG_NUM_DOMAINS - 1
+    maxDom = CONFIG_NUM_DOMAINS - 1,
+    /* To analyse branches of control flow decisions based on the number of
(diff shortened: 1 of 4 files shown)


kent-mcleod (approved):

Xaphiosis:
Thanks @kent-mcleod, that was really quick. I'm working on the much more complicated proof side of generalising number of domains, so this will sit until I finish that, hopefully not too long.

lsf37 (approved):

Xaphiosis:
rebased

Xaphiosis:
rebased again, no surprises, should be ready to merge once verification tests complete on seL4/l4v#410

PR #719, opened 2021-12-26
boot: drop explicit parameter first_untyped_slot

axel-h:
Creating untypeds must happen in in block and it's good to have some checks in place that ensure this is the case. The checks are just asserts that work in debug builds, but that is sufficient as this is mostly a porting or platform adaption issue.
The explicit parameter `first_untyped_slot` does not really gain much, the value can be provided in `ndks_boot.bi_frame->untyped.start` also to any function that case. The functon `provide_untyped_cap()` remains the only function that really cares about and contains the sanity check now also

This PR is on top of:
- https://github.com/seL4/seL4/pull/715

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -582,13 +582,19 @@ BOOT_CODE static bool_t provide_untyped_cap(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
     pptr_t     pptr,
-    word_t     size_bits,
-    seL4_SlotPos first_untyped_slot
-)
+    word_t     size_bits)
 {
     bool_t ret;
     cap_t ut_cap;
-    word_t i = ndks_boot.slot_pos_cur - first_untyped_slot;
+
+    /* Sanity check, the whole untyped creation cannot be split up during the
+     * boot process. In must happen in one block, otherwise the boot info meta
+     * data can't handle this with just a start and end field.
+     */
+    assert(ndks_boot.bi_frame->untyped.start <= ndks_boot.slot_pos_cur);
+    assert(ndks_boot.bi_frame->untyped.end = ndks_boot.slot_pos_cur);
+
+    word_t i = ndks_boot.slot_pos_cur - ndks_boot.bi_frame->untyped.start;
     if (i < CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS) {
         ndks_boot.bi_frame->untypedList[i] = (seL4_UntypedDesc) {
             .paddr    = pptr_to_paddr((void *)pptr),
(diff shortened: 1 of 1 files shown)


kent-mcleod (approved):
If I understand correctly, this is adding an assert that enforces, ndks_boot.slot_pos_cur hasn't been incremented in between multiple calls to `provide_untyped_cap()` in order to ensure that all created untyped caps are in a single range of slots in the root CNode?

axel-h:
> If I understand correctly, this is adding an assert that enforces, ndks_boot.slot_pos_cur hasn't been incremented in between multiple calls to `provide_untyped_cap()` in order to ensure that all created untyped caps are in a single range of slots in the root CNode?

Yes.

src/kernel/boot.c
```diff
@@ -580,13 +580,19 @@ BOOT_CODE static bool_t provide_untyped_cap(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
     pptr_t     pptr,
-    word_t     size_bits,
-    seL4_SlotPos first_untyped_slot
```
yyshen, line 578: This commit only removes the first_untyped_slot field? I think keeping it explicitly helps people to understand the code. Overloading with untyped.end with two purposes is not as obvious as having a first_untyped_slot.

kent-mcleod: The commit message could be worded more clearly.  The primary change is that `ndks_boot.bi_frame` is updated after each call to provide_untyped_cap with the updated region of untyped caps.  This change allows the code to check that no other caps are inserted between successive calls to provide_untyped_cap so that the final untyped region is correct.  It's not possible to check this without this change. Once this change has been made then first_untyped_slot becomes an unused parameter.

axel-h: The issue here is, that the whole concept only works out, if `ndks_boot.bi_frame->untyped.start` is properly setup. All untype caps must be in one "block". So my argument again that would be that passing a parameter `first_untyped_slot`explicitly creates an impression one can do something here - but that is not the case, if this gets out of sync this will cause strange effects eventually and we see an assert anyway.
Thus I think that not explicitly having this is better. However, what we could pass there is a pointer to the current boot info slot region that this is working on (`&ndks_boot.bi_frame->untyped`). Then `provide_untyped_cap()` and other would no really care which boot info slot region they are really working on, but could automatically update it. That is what I have tried out in https://github.com/axel-h/seL4/tree/patch-axel-67

kent-mcleod: It would also be possible to keep the first_untyped_slot parameter and update `ndks_boot.bi_frame->untyped.end` inside the function.  Then you could still use asserts to catch if the ndks_boot.slot_pos_cur doesn't match the last call.

axel-h:
I'm dropping this in favor of the change in https://github.com/seL4/seL4/pull/648

kent-mcleod:
> I'm dropping this in favor of the change in #648

I'm more in favor of what this PR was proposing (with the removal of the first_untyped_slot parameter) than what's being proposed in #648. This is because ultimately, `create_untypeds_for_region` is creating untyped caps into a single continuous range of slots and this isn't expected to be generalized to multiple ranges any time soon.

axel-h:
You have a point there. The bigger thing behind https://github.com/seL4/seL4/pull/648 is then what I keep in https://github.com/axel-h/seL4/pull/47 where the `slot_region` parameter comes handy and the boot info is no longer a global thing accessed in multiple places, but all access are based on a reference passed around that is obtained in exactly one place.

PR #720, opened 2021-12-26
boot: handle inclusive region end properly

axel-h:
The region end address can be inclusive or exclusive. Ensure the length is calculated properly for the inclusive form. Without this the device untypes on e.g. the zynq7000 platform are:
```
...
[0xf8f20000..0xf8f40000] size 2^17
[0xf8f40000..0xf8f80000] size 2^18
[0xf8f80000..0xf9000000] size 2^19
[0xf9000000..0xfa000000] size 2^24
[0xfa000000..0xfc000000] size 2^25
[0xfc000000..0xfe000000] size 2^25
[0xfe000000..0xff000000] size 2^24
[0xff000000..0xff800000] size 2^23
... continues for 2^22 down to 2^7 ...
[0xffffff80..0xffffffc0] size 2^6
[0xffffffc0..0xffffffe0] size 2^5
[0xffffffe0..0xfffffff0] size 2^4
```
And the last 16 bytes are left out. With this change they are:
```
[0xf8f20000..0xf8f40000] size 2^17
[0xf8f40000..0xf8f80000] size 2^18
[0xf8f80000..0xf9000000] size 2^19
[0xf9000000..0xfa000000] size 2^24
[0xfa000000..0xfc000000] size 2^25
[0xfc000000..0x100000000] size 2^26
```

Note that for the top n byte of the physical address space there might be a need to reserve them and not provide untypeds. But this should be done via a reserved region then explicitly and not be hidden in the untyped creation algorithm. 
See also:
- https://sel4.discourse.group/t/device-region-untypeds/328
- https://github.com/seL4/seL4/pull/556

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -724,18 +724,27 @@ BOOT_CODE static bool_t create_untypeds_for_region(
     seL4_SlotPos first_untyped_slot
 )
 {
+    /* The region end address can be inclusive or exclusive, this must be taken
+     * into account when calculating the region size. An end address is assumed
+     * to be inclusive if the LSB is 1.
+     */
+    word_t reg_size = reg.end - reg.start;
+    if (reg.end & 1) {
+        reg_size++; /* Increment length by 1 for inclusive end addresses. */
+        assert(reg_size > 0); /* Overflows are not expected. */
+    }
+
     /* This code works with regions that wrap (where end < start), because the loop cuts up the
        region into size-aligned chunks, one for each cap. Memory chunks that are size-aligned cannot
        themselves overflow, so they satisfy alignment, size, and overflow conditions. The region
        [0..end) is not necessarily part of the kernel window (depending on the value of PPTR_BASE).
        This is fine for device untypeds. For normal untypeds, the region is assumed to be fully in
        the kernel window. This is not checked here. */
-    while (!is_reg_empty(reg)) {
-
+    while (reg_size > 0) {
         /* Calculate the bit size of the region. This is also correct for end < start: it will
            return the correct size of the set [start..-1] union [0..end). This will then be too
            large for alignment, so the code further down will reduce the size. */
-        unsigned int size_bits = seL4_WordBits - 1 - clzl(reg.end - reg.start);
+        unsigned int size_bits = seL4_WordBits - 1 - clzl(reg_size);
         /* The size can't exceed the largest possible untyped size. */
         if (size_bits > seL4_MaxUntypedBits) {
             size_bits = seL4_MaxUntypedBits;
@@ -758,8 +767,14 @@ BOOT_CODE static bool_t create_untypeds_for_region(
                 return false;
             }
         }
-        reg.start += BIT(size_bits);
+
+        word_t chunk_size = BIT(size_bits);
+        reg.start += chunk_size;
+        assert(reg_size >= chunk_size);
+        reg_size -= chunk_size;
+
     }
+
     return true;
 }
 


src/kernel/boot.c
```diff
@@ -611,10 +611,22 @@ BOOT_CODE static bool_t create_untypeds_for_region(
     seL4_SlotPos first_untyped_slot
 )
 {
-    while (!is_reg_empty(reg)) {
+    /* The region end address can be inclusive or exclusive, this must be taken
+     * into account when calculating the region size. An end address is assumed
+     * to be inclusive if the LSB is 1.
+     */
```
kent-mcleod, line 617: @lsf37 what do you think about this?  The convention that a region start address is inclusive and a region end address is exclusive is not something that I believe can be easily changed. Treating an end address as inclusive if the LSB = 1 means that it's no longer possible to treat the region's size as being equal to its end minus its start.

lsf37: So in general the better convention (for verification and overflow reasoning) for intervals is either start + size or end inclusive. If the the end is exclusive there is a potential for overflow at the end of the value range, which doesn't occur in the other two representations.

That said, changing this here would be a pretty large operation for not much gain and doing both at the same time by conventions such as LSB strikes me as unnecessary complexity at this point -- if we add complexity it should be where the regions are created, not where they are consumed. If we do boot code verification at some point, this function would definitely be in scope, and I would not support doing both.

I would also not want this function to ever see empty regions -- the guard for an empty region was for the iteration, not because the function should be expect empty input.

axel-h: Rebased and brought back the `is_reg_empty()` call. I've happy to fixe the root cause also to remove this inconsistency and establish a clear rule about region. I was not sure if the end is inclusive or not, have both are use thoughout the code. My assumption was that the "inclusive end" comes from trying to handle the end of the address space (2^32 or 2^64) without overflow?

lsf37: That might be the underlying cause, yes (I'm not sure, I haven't followed the boot code that closely in the past; probably worth checking the git history on this one). 

If that is the only time, and we are sure about that, we could check for the end of the address space by explicit value instead of LSB.

lsf37:
Coming back to this after staring at the region code more and documenting it: the memory region types are definitely with exclusive end, we should not change that here, because a lot of stuff like check for empty intervals, overlapping intervals, etc will break in the rest of the code.

If some platforms provide the end address as inclusive, that should be changed (if it is device memory, it is fine to provide `0` as the end).

If the problem is from the definition of KernelPaddrUserTop I think Benno is correct in that making this value exclusive should fix the problem. This will have proof impact, but shouldn't be too catastrophic.

In short we should close this PR, and replace it with one or both of the above.

lsf37:
@axel-h you mentioned in the comment above that you've seen the region end used as inclusive in the code. Do you remember where? It'd be good to fix these and make sure we're treating them consistently everywhere as exclusive.

I've put up PR #1007 which documents that the region ends are exclusive, and I think we should also print them that way (#1008), because I've gotten confused by this as well.

lsf37:
Some analysis/summary: 

- memory regions are currently all exclusive, so the fix proposed here is wrong and the PR should be closed
- the underlying issue no longer occurs on any of the 64 bit platforms
- it still does occur on AArch32 and rv32, e.g. on Sabre hardware runs and rv32 qemu run, the device untypeds are a cascade that go up to `(2^32-1) & ~mask 4`. The issue there is not anything with the platform memory layout description or the regions provided to the kernel, but the definition of `CONFIG_PADDR_USER_DEVICE_TOP` which for these is `2^32-1` which is not according to spec (is not exclusive). Of course 2^32 would wrap. (Have I mentioned recently how inclusive regions don't have this problem?)

I'll open a separate issue and close this PR.

Issue #716, opened 2021-12-24
Why does CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS exist?


axel-h:
I wonder why `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` exists - what is/was the usecase? In the current form enforcing this seems to be of limited practical use, as there is no differentiation between device and free memory caps. We stop creating caps when this limit is hit, even is there is still space in the root C-Node. Device caps are created first, then boot code recycling happens and finally memory caps are created from the free memory. Should `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` be removed? Or should it be replaced by separate limits for CONFIG_MAX_NUM_BOOTINFO_UNTYPED_MEMORY_CAPS' and CONFIG_MAX_NUM_BOOTINFO_UNTYPED_DEVICE_CAPS'. However, I can see a usecase only for limiting the memory caps. Limiting the device caps by numer does not really make sense. Here it seem more useful to have a dedicated way to configure the overall device caps (see https://github.com/seL4/seL4/issues/714)
Assuming that the only reason `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` exists is, that the array size in the boot info must be define somehow and this value can't be computed during kernel compillation, the boot should abort if this number is too small. Currently the boot continues in this case and just provides less caps to the system - which does not really make much sense. See also https://github.com/seL4/seL4/pull/648#issuecomment-997480274.

yyshen:
It exists because the bootinfo structure  is 4KiB (if I read correctly that the extra bi_frame_vptr is only  used when DTB is provided), so there should be an upper bound for the array size.

axel-h:
We could also provide additional descriptors in the extra boot info then if this is ever needed. Currently there is space or 501 descriptors in the boot info page and that seems sufficient on all standard platforms.

axel-h:
As @lsf37 stated in https://github.com/seL4/seL4/pull/721#discussion_r777755358, the constant `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` exists because the verification requested this. Using open arrays as final element in structs is not supported, so we have to give something in `seL4_BootInfo` for the `seL4_UntypedDesc untypedList[]`array.
[...]

PR #721, opened 2021-12-26
calculate CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS

axel-h:
Calculate the value from the remaining space in boot info frame instead of explicitly defining it. Provide the define for userland in libsel4 to avoid breaking existing code.

See also https://github.com/seL4/seL4/issues/716

### DIFF
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -52,6 +52,10 @@ typedef struct seL4_UntypedDesc {
     seL4_Uint8 padding[sizeof(seL4_Word) - 2 * sizeof(seL4_Uint8)];
 } seL4_UntypedDesc;
 
+SEL4_COMPILE_ASSERT(
+    invalid_seL4_UntypedDesc,
+    sizeof(seL4_UntypedDesc) == 2 * sizeof(seL4_Word));
+
 typedef struct seL4_BootInfo {
     seL4_Word         extraLen;        /* length of any additional bootinfo information */
     seL4_NodeId       nodeID;          /* ID [0..numNodes-1] of the seL4 node (0 if uniprocessor) */
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -7,7 +7,7 @@
 #pragma once
 
 #include <autoconf.h>
-#include <sel4/macros.h>
+#include <sel4/sel4_arch/constants.h>
 
 /* caps with fixed slot positions in the root CNode */
 enum {
@@ -79,11 +79,26 @@ typedef struct seL4_BootInfo {
      * to make this struct easier to represent in other languages */
 } seL4_BootInfo;
 
-/* If extraLen > 0, then 4K after the start of bootinfo there is a region of the
- * size extraLen that contains additional boot info data chunks. They are
- * arch/platform specific and may or may not exist in any given execution. Each
- * chunk has a header that contains an ID to describe the chunk. All IDs share a
- * global namespace to ensure uniqueness.
+/* The boot info frame size is usually exactly one page. As she page size is
+ * 4 KiByte on x86, ARM and RISC-V, some userland code seems to have a hard
+ * coded assumption that the boot info frame size is 4 KiByte. New code should
+ * avoid this and use SEL4_BI_FRAME_SIZE instead. This allows the boot info page
+ * to become larger in case CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS needs to be
+ * increased. Keep in mind that a larger boot info frame might break older
+ * userland code and libs then.
+ */
+#define SEL4_BI_FRAME_PAGES  1
+#define SEL4_BI_FRAME_SIZE   (SEL4_BI_FRAME_PAGES * LIBSEL4_BIT(seL4_PageBits))
+
+SEL4_COMPILE_ASSERT(
+    invalid_SEL4_BI_FRAME_SIZE,
+    sizeof(seL4_BootInfo) <= SEL4_BI_FRAME_SIZE)
+
+/* If extraLen > 0, then at the offset SEL4_BI_FRAME_SIZE after the start of
+ * bootinfo there is a region of the size extraLen that contains additional boot
+ * info data chunks. They are arch/platform specific and may or may not exist in
+ * any given execution. Each chunk has a header that contains an ID to describe
+ * the chunk. All IDs share a global namespace to ensure uniqueness.
  */
 typedef enum {
     SEL4_BOOTINFO_HEADER_PADDING            = 0,
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -74,9 +74,8 @@ typedef struct seL4_BootInfo {
     seL4_SlotRegion   schedcontrol;    /* Caps to sched_control for each node */
 #endif
     seL4_SlotRegion   untyped;         /* untyped-object caps (untyped caps) */
-    seL4_UntypedDesc  untypedList[CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS]; /* information about each untyped */
-    /* the untypedList should be the last entry in this struct, in order
-     * to make this struct easier to represent in other languages */
+    seL4_UntypedDesc  untypedList[];   /* information about each untyped */
+    /* the untypedList[] must be the last entry */
 } seL4_BootInfo;
 
 /* The boot info frame size is usually exactly one page. As she page size is
@@ -93,6 +92,23 @@ typedef struct seL4_BootInfo {
 SEL4_COMPILE_ASSERT(
     invalid_SEL4_BI_FRAME_SIZE,
     sizeof(seL4_BootInfo) <= SEL4_BI_FRAME_SIZE)
+
+/* The number of elements in untypedList[] is limited by the by the remaining
+ * space in the boot info frame. Currently, all remaining space is used. The
+ * calculation below is safe, because the size of the open array untypedList[]
+ * counts as zero for sizeof(seL4_BootInfo), but this still takes into account
+ * padding after the field 'untyped' to ensure a proper alignment of the array.
+ */
+#define MAX_NUM_BOOTINFO_UNTYPED_CAPS \
+    ((SEL4_BI_FRAME_SIZE - sizeof(seL4_BootInfo)) / sizeof(seL4_UntypedDesc))
+
+/* The constant CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS was a CMake configuration
+ * parameter that allowed defining the size of the array untypedList[]. However,
+ * since the boot info frame always take full pages the array uses the remaining
+ * space. Thus, the configuration option has been dropped and the define is
+ * provided here for legacy compatibility.
+ */
+#define CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS  MAX_NUM_BOOTINFO_UNTYPED_CAPS
 
 /* If extraLen > 0, then at the offset SEL4_BI_FRAME_SIZE after the start of
  * bootinfo there is a region of the size extraLen that contains additional boot
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -23,8 +23,8 @@ enum {
     seL4_CapBootInfoFrame       =  9, /* bootinfo frame cap */
     seL4_CapInitThreadIPCBuffer = 10, /* initial thread's IPC buffer frame cap */
     seL4_CapDomain              = 11, /* global domain controller cap */
-    seL4_CapSMMUSIDControl      = 12,  /*global SMMU SID controller cap, null cap if not supported*/
-    seL4_CapSMMUCBControl       = 13,  /*global SMMU CB controller cap, null cap if not supported*/
+    seL4_CapSMMUSIDControl      = 12, /* global SMMU SID controller cap, null cap if not supported */
+    seL4_CapSMMUCBControl       = 13, /* global SMMU CB controller cap, null cap if not supported */
 #ifdef CONFIG_KERNEL_MCS
     seL4_CapInitThreadSC        = 14, /* initial thread's scheduling context cap */
     seL4_NumInitialCaps         = 15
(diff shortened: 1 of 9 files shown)


libsel4/include/sel4/bootinfo_types.h
```diff
@@ -69,9 +69,7 @@ typedef struct seL4_BootInfo {
     seL4_SlotRegion   schedcontrol; /* Caps to sched_control for each node */
 #endif
     seL4_SlotRegion   untyped;         /* untyped-object caps (untyped caps) */
-    seL4_UntypedDesc  untypedList[CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS]; /* information about each untyped */
-    /* the untypedList should be the last entry in this struct, in order
-     * to make this struct easier to represent in other languages */
+    seL4_UntypedDesc  untypedList[];   /* information about each untyped */
 } seL4_BootInfo;
```
Indanz, line 74: Wouldn't it make more sense to define MAX_NUM_BOOTINFO_UNTYPED_CAPS in the kernel underneath the seL4_BootInfo structure definition and define CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS  to that?

libsel4/include/sel4/macros.h
```diff
@@ -19,7 +19,14 @@
 #define SEL4_PACKED             __attribute__((packed))
 #define SEL4_DEPRECATED(x)      __attribute__((deprecated(x)))
 #define SEL4_DEPRECATE_MACRO(x) _Pragma("deprecated") x
+
+
+#ifdef CONFIG_VERIFICATION_BUILD
+/* __builtin_offsetof() is currently not supported by the C Parser */
+#define SEL4_OFFSETOF(type, member) ((word_t)&(((type *)0)->member))
+#else /* not CONFIG_VERIFICATION_BUILD */
 #define SEL4_OFFSETOF(type, member) __builtin_offsetof(type, member)
```
Indanz, line 28: If the new one is proven to be correct, why also have the built-in one?

libsel4/include/sel4/bootinfo.h
```diff
@@ -10,6 +10,23 @@
 #include <sel4/bootinfo_types.h>
 #include <sel4/macros.h>
 
+/* The constant CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS has been removed from
+ * CMake, it automatically adapts to the boot info frame's size now. To keep
+ * compatibility, the define is provide here now. Getting the boot info frame
+ * size within the kernel is easy, as BI_FRAME_SIZE_BITS exists. For userland,
+ * this is currently not exposed and there seems an assumption that this is
+ * always one page. We define BI_FRAME_SIZE_BITS for userland here.
```
Indanz, line 18: Isn't it better to just expose the kernel's BI_FRAME_SIZE_BITS?

Indanz (commented):
Having it as a configuration option is misleading, this change makes sense to me.

libsel4/include/sel4/bootinfo_types.h
axel-h: I wanted to do that, but have not found a way yet how to do it. It's a bit of a chicken-egg problem to calculate this. And it the end I wonder if all this is really worth the effort and who really needs `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS`. In the boot info's `seL4_SlotRegion untyped` you get the start and end slot, calculating the amount of descriptors from the delta is easy. The kernel will also guarantee that the start and end values in there is consistent with the items in the array. That just leave the corner case that there are more untyped to be created than there is space for descriptors in the boot info block, which is to be discussed in https://github.com/seL4/seL4/issues/716

libsel4/include/sel4/macros.h
axel-h: Actually, the macro (and various others) violates the C specs, see explanation at https://en.wikipedia.org/wiki/Offsetof. That is why using `__builtin_offsetof()` is the best option and ideally support for this in the verification toolchain should be added. I agree that the current solution is not really optimal and not using `__builtin_offsetof()` might be something to consider. Opinions from others?

libsel4/include/sel4/bootinfo.h
axel-h: Yes. I was wondering if this should be made a config option. Or if we just adapt the boot info frame size dynamically to the required size. But then we might have to communicate the size to the root task also, which requires a new fiel somewhere. However, it's also implicitly available due to the number of untyped slots and implies size of the descriptor array. 
Which reminds me that BI_FRAME_SIZE_BITS is fixed to 4KiB anyway because of the extra boot info that is defined to start 4 KiB after the boot info. I have to add this in the comment. And unless we break compatibility, there is a hard limit at 500 untyped descriptors in the 4 KiB page.

libsel4/include/sel4/bootinfo_types.h
Indanz: Inline would be best but also the ugliest, but you could just add a:
```
#define MAX_NUM_BOOTINFO_UNTYPED_CAPS ((SEL4_BIT(BI_FRAME_SIZE_BITS) - sizeof(struct seL4_BootInfo)) / sizeof(seL4_UntypedDesc))
```
underneath it, no need for offsetof() either then.

Indanz: Or perhaps have no CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS or any define at all and just use sizeof() in the one place that uses it instead of offsetof().

libsel4/include/sel4/macros.h
Indanz: I disagree that it violates the C specs as it is doing pointer arithmetic and not actually following the pointer, and the arithmetic itself has no side effects, so could never have undefined behaviour. But I have a practical view of standards.

libsel4/include/sel4/bootinfo_types.h
axel-h: > Inline would be best but also the ugliest, but you could just add a:
> ```
> #define MAX_NUM_BOOTINFO_UNTYPED_CAPS ((SEL4_BIT(BI_FRAME_SIZE_BITS) - sizeof(struct seL4_BootInfo)) / sizeof(seL4_UntypedDesc))
> ```
> underneath it, no need for offsetof() either then.

I'm not sure if this really works. I have though of splitting up `seL4_BootInfo` and declare the fixed part separately. Then it is known and a macro can be used to calculate the elements for an additional variable part.
```
typedef struct seL4_BootInfo_flex {
    seL4_Word         extraLen;
    ...
    seL4_SlotRegion untyped;
} seL4_BootInfo_fixed

#define CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS ((SEL4_BIT(BI_FRAME_SIZE_BITS) - sizeof(seL4_BootInfo_fixed)) / sizeof(seL4_UntypedDesc))

typedef struct seL4_BootInfo {
    // ... and here I'm lost how to do the magic to get seL4_BootInfo_fixed in here without 
    // giving it a name. We could make the struct boy a big macro that is used twice....

    seL4_UntypedDesc  untypedList[CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS];

} seL4_BootInfo_fixed

Comment (Indanz) [reply libsel4/include/sel4/bootinfo_types.h:74]: Just leave it as ``seL4_UntypedDesc  untypedList[];`` and it works fine?

**Comment by axel-h** [comment]
@Indanz: I've simplified things, what do you think if it now?

File: libsel4/include/sel4/bootinfo_types.h [hunk 4]
```diff
@@ -69,11 +70,24 @@ typedef struct seL4_BootInfo {
...
+ * 0 when using sizeof(seL4_BootInfo), so we can define the max number of
+ * elements according to the boot info frame size here. For userland the
+ * specification defines that the boot info frame is 4 KiB, which corresponds to
+ * the page size on x86, ARM and RISC-V, and extra boot info according to
+ * extraLen/extraBIPages follows afterwards.
+ */
+#define BI_FRAME_SIZE_BITS  12  /* 2^12 = 4096 */
+SEL4_COMPILE_ASSERT(seL4_BootInfo_exceeds_4KiB,
+                    sizeof(seL4_BootInfo) <= LIBSEL4_BIT(BI_FRAME_SIZE_BITS));
+#define CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS \
```
Indanz, line 88: I would call this something else than CONFIG_* to avoid confusion. True that that's not backward compatible, but I don't think seL4 should strive for backward compatibility when removing unnecessary config options.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -69,11 +70,24 @@ typedef struct seL4_BootInfo {
...
 
+/* The constant CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS has been removed from
+ * CMake configuration and a flexible array untypedList[] is used. It counts a
+ * 0 when using sizeof(seL4_BootInfo), so we can define the max number of
+ * elements according to the boot info frame size here. For userland the
+ * specification defines that the boot info frame is 4 KiB, which corresponds to
+ * the page size on x86, ARM and RISC-V, and extra boot info according to
+ * extraLen/extraBIPages follows afterwards.
+ */
+#define BI_FRAME_SIZE_BITS  12  /* 2^12 = 4096 */
```
Indanz, line 85: Why not keep it defined to PAGE_BITS as before? Or seL4_PageBits? It's only 4 kB because that's the smallest page size for all supported platforms.

Isn't this the place that defines the userland specification? It seems strange to refer to it here like this.

(Type "counts as 0".)

Indanz (approved):
(Github is confusing IMHO.)

libsel4/include/sel4/bootinfo_types.h
axel-h: Thanks, we can use `seL4_PageBits` here. I tried `PAGE_BITS` but that did no work out.

libsel4/include/sel4/bootinfo.h
axel-h: Actually, what about having a new `CONFIG_BI_FRAME_SIZE_BITS` instead that defaults to 12? And then have a 
```
SEL4_COMPILE_ASSERT(seL4_BootInfo_fits, CONFIG_BI_FRAME_SIZE_BITS >= seL4_PageBits)
#define BI_FRAME_SIZE       LIBSEL4_BIT(BI_FRAME_SIZE_BITS)
```
in the header. Rhe documentation would say that `CONFIG_BI_FRAME_SIZE_BITS` can be increased if `untypedList[]` is too small. 
That may break some useland code that has the hard-coded assumption the boot info frame is 4 KiB, but this needs to be adapted then. Unfortunately, there is no way to communicate this value to userland dynamically. But even for a pre-compiled kernel the headers provided with it make `CONFIG_BI_FRAME_SIZE_BITS` (and thus `BI_FRAME_SIZE`) available for the root task.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -69,14 +71,36 @@ typedef struct seL4_BootInfo {
     seL4_SlotRegion   schedcontrol; /* Caps to sched_control for each node */
 #endif
     seL4_SlotRegion   untyped;         /* untyped-object caps (untyped caps) */
-    seL4_UntypedDesc  untypedList[CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS]; /* information about each untyped */
```
yyshen, line 72: It might also depend on the root cnode size.

libsel4/include/sel4/bootinfo_types.h
axel-h: There is `MAX_NUM_BOOTINFO_UNTYPED_CAPS` now and `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS`is kept as an alias to avoid breaking existing code.

libsel4/include/sel4/bootinfo.h
axel-h: The boot info frame has to be one page, and changing this is not trivial. So no chance to make this trivially configurable for now. 240 untyped desciptors (500 on 32 bit) is really a hard limit, but it seems this is not hit anywhere besides some experimental configurations. As a follow-up change, we could allocate the boot info frame and the extra boot info together as a bunch of pages without special alignment needs. That would then allow having the the boot info take more than one page and the kernel boot memory allocation would get a bit simpler also. Seem there is no reason why extra boot info needs an alignment greater then one page either.

libsel4/include/sel4/bootinfo_types.h
axel-h: In `provide_cap()` there is a check about the space in the root C-Node and booting fails if it too small.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -66,17 +68,48 @@ typedef struct seL4_BootInfo {
...
     seL4_Domain       initThreadDomain; /* Initial thread's domain ID */
 #ifdef CONFIG_KERNEL_MCS
-    seL4_SlotRegion   schedcontrol; /* Caps to sched_control for each node */
+    seL4_SlotRegion   schedcontrol;    /* Caps to sched_control for each node */
 #endif
     seL4_SlotRegion   untyped;         /* untyped-object caps (untyped caps) */
-    seL4_UntypedDesc  untypedList[CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS]; /* information about each untyped */
-    /* the untypedList should be the last entry in this struct, in order
-     * to make this struct easier to represent in other languages */
+    seL4_UntypedDesc  untypedList[];   /* information about each untyped */
```
lsf37, line 74: This will fail struct size computation in verification, sorry. You have to provide a length for arrays in structs.

libsel4/include/sel4/bootinfo_types.h
lsf37: Ah, should have read this discussion first before commenting above. You can't have a non-fixed struct size anyway, you'll need to compute a size here for declaration. That is the reason it was a config parameter in the first place, because it's annoying to do that.

libsel4/include/sel4/bootinfo_types.h
yyshen: That is a runtime check. if we would like to calculate the size during compilation, we should consider the factors and give one that does not fail.

axel-h: We could ensure at compile time that the root C-Node has at least `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` slots. But it's likely that it has far more slots anyway. Furthermore, in most cases it's even unlikely all slots are really used. And then the exact number of slots to be used is not know at compile time, as that depends on the memory layout - now one can assume this is static practically, but technically it depends on parameters passed from a previous boot loader. 
So not sure if there is really a gain here doing this as a compile time check also.

yyshen: I appreciate the effort, but I am not sure if the calculated one is much better than the user-defined one. I think several compile time assertions to ensure that a user does not provide a number that is greater than one-page size or root cnode size should be enough?

axel-h: I tend to agree, all this was motivated by https://github.com/seL4/seL4/issues/716 basically. Seems we could improve the comments here a bit why this exists and what changing this implies. The value is not really meant to be changed beyond a certain limit, that is driven by the free space in the boot info page on 64 bit platforms. So that makes we wonder if this should relly be a config option and not an internal constants.

axel-h:
Thanks everybody for the feedback. I'm going to drop this PR for now, as it seems `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` has to exist in some form. I'l try again with another PR that make the size and handling of the fixed boot info frame flexible. So it can be more than one page, so it can adapt to large values of `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` or other fixed boot info data.

kent-mcleod:
I'm going to close this as the initial discussion seems to have reached a conclusion.

PR #722, opened 2021-12-26
boot: use correct bootinfo frame size

axel-h:
- The bootinfo size is `BI_FRAME_SIZE_BITS`, don't assume it is `PAGE_BITS`.
- Put compile time check in place to ensure the boot memory allocation can't get broken by mis-configuration.


kent-mcleod:
I'm running proof-test because I'm curious if `compile_assert` can be used inside a function. Otherwise these would need to be moved outside of the function scope.

lsf37 (approved):
Happy to merge this from the verification side. (Proof succeeded)

kent-mcleod (approved):

PR #726, opened 2021-12-27
add missing config.h include

axel-h:


kent-mcleod (approved):

GITHUBGUY12357999:
All tests are successfully implemented please review.

lsf37:
Just to confirm, we are adding the includes here to be consistent with normal C where each `.c` file would be a separate compilation unit?

I don't really have a problem with that, but I'd like to point out that I don't want a separate compilation workflow to emerge in the future that is different from the one verification sees.

Phrased differently: I do want the concatenation of all `.c` files in seL4 to remain the one single compilation unit, but I'm fine with making the includes explicit because that makes compilation more robust against reordering of `.c` files in that concatenation.

GITHUBGUY12357999:
Ok so u can go ahead.

GITHUBGUY12357999 (approved):
Approved.

GITHUBGUY12357999:
Just to confirm, we are adding the includes here to be consistent with
normal C where each .c file would be a separate compilation unit?---- YES
we are going to.

axel-h:
@lsf37: I have no intention to change the compilation flow. Just as you said, having proper includes is useful to avoid quirks or implicit (or better non trivially understandable) behavior.

lsf37 (approved):

PR #728, opened 2021-12-28
boot: improve memory region sanity checks

axel-h:
- Adjust free memory to kernel window, as well for the lower as the upper limit.
- Log more detailed boot message to help porting.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -869,17 +869,6 @@ BOOT_CODE void bi_finalise(void)
     };
 }
 
-BOOT_CODE static inline pptr_t ceiling_kernel_window(pptr_t p)
-{
-    /* Adjust address if it exceeds the kernel window
-     * Note that we compare physical address in case of overflow.
-     */
-    if (pptr_to_paddr((void *)p) > PADDR_TOP) {
-        p = PPTR_TOP;
-    }
-    return p;
-}
-
 BOOT_CODE static bool_t check_available_memory(word_t n_available,
                                                const p_region_t *available)
 {
@@ -889,28 +878,29 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
         return false;
     }
 
-    printf("available phys memory regions: %"SEL4_PRIu_word"\n", n_available);
+    printf("Available phys memory regions: %"SEL4_PRIu_word"\n", n_available);
     /* Force ordering and exclusivity of available regions. */
     for (word_t i = 0; i < n_available; i++) {
         const p_region_t *r = &available[i];
         printf("  [%"SEL4_PRIx_word"..%"SEL4_PRIx_word")\n", r->start, r->end);
 
-        /* Available regions must be sane */
+        /* Available regions must not wrap */
         if (r->start > r->end) {
             printf("ERROR: memory region %"SEL4_PRIu_word" has start > end\n", i);
             return false;
         }
 
-        /* Available regions can't be empty. */
+        /* Available regions must be non-empty. */
         if (r->start == r->end) {
             printf("ERROR: memory region %"SEL4_PRIu_word" empty\n", i);
             return false;
         }
 
         /* Regions must be ordered and must not overlap. Regions are [start..end),
-           so the == case is fine. Directly adjacent regions are allowed. */
+         * so the '==' case is fine. Directly adjacent regions are allowed.
+         */
         if ((i > 0) && (r->start < available[i - 1].end)) {
-            printf("ERROR: memory region %d in wrong order\n", (int)i);
+            printf("ERROR: memory region %"SEL4_PRIu_word" in wrong order\n", i);
             return false;
         }
     }
@@ -918,11 +908,61 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
     return true;
 }
 
+/* We can't declare arrays on the stack, so this is space for init_avail_reg()
+ * below to populate and init_freemem() to consume. */
+BOOT_BSS static region_t avail_reg[MAX_NUM_FREEMEM_REG];
+
+/**
+ * Initialise avail_reg with the parts of the available memory regions that are
+ * in the kernel window, converted to kernel pointers. Only this memory can be
+ * used for kernel objects. On 32-bit architectures it is not uncommon for
+ * regions to be completely or partially outside of the kernel window, because
+ * the available address space is small. Memory outside of the kernel window is
+ * discarded here, but it may be made available as device untyped memory instead
+ * if it is not also in a reserved region. Regions must have passed
+ * check_available_memory(). Returns the number of entries populated in
+ * avail_reg.
+ */
+BOOT_CODE static word_t init_avail_reg(word_t n_available,
+                                       const p_region_t *available)
+{
+    word_t cnt = 0;
+    for (word_t i = 0; i < n_available; i++) {
+        const p_region_t *r = &available[i];
+
+        if ((r->start >= PADDR_TOP) || (r->end <= PADDR_BASE)) {
+            printf("  region [%"SEL4_PRIx_word"..%"SEL4_PRIx_word") is outside the kernel"
+                   " window, only potentially available as device untypeds\n", r->start, r->end);
+        } else {
+            p_region_t usable_reg = {
+                .start = MAX(r->start, PADDR_BASE),
+                .end   = MIN(r->end, PADDR_TOP)
+            };
+
+            if (cnt >= ARRAY_SIZE(avail_reg)) {
+                /* No space left in the array, seems MAX_NUM_FREEMEM_REG should
+                 * be increased. Debug builds raise an assert here because this
+                 * is likely a porting issue that should be looked into. Release
+                 * builds will continue booting, but can't use this memory
+                 * region. This might still be sufficient to run.
+                 */
+                printf("  WARNING: cannot use region [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"),"
+                       " avail_reg[] is full\n", r->start, r->end);
+                assert(0);
+            } else {
+                avail_reg[cnt] = paddr_to_pptr_reg(usable_reg);
+                cnt++;
+            }
+        }
+    }
+    return cnt;
+}
+
 
 BOOT_CODE static bool_t check_reserved_memory(word_t n_reserved,
                                               const region_t *reserved)
 {
-    printf("reserved virt address space regions: %"SEL4_PRIu_word"\n",
+    printf("Reserved virt address space regions: %"SEL4_PRIu_word"\n",
            n_reserved);
     /* Force ordering and exclusivity of reserved regions. */
     for (word_t i = 0; i < n_reserved; i++) {
@@ -946,9 +986,6 @@ BOOT_CODE static bool_t check_reserved_memory(word_t n_reserved,
     return true;
 }
 
-/* we can't declare arrays on the stack, so this is space for
- * the function below to use. */
-BOOT_BSS static region_t avail_reg[MAX_NUM_FREEMEM_REG];
 /**
  * Dynamically initialise the available memory on the platform.
  * A region represents an area of memory.
@@ -957,8 +994,14 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
                               word_t n_reserved, const region_t *reserved,
                               v_region_t it_v_reg, word_t extra_bi_size_bits)
 {
-
     if (!check_available_memory(n_available, available)) {
+        return false;
+    }
+
+    /* After here, avail_reg[0..n_available) are all regions within the kernel window. */
+    n_available = init_avail_reg(n_available, available);
+    if (0 == n_available) {
+        printf("ERROR: no available memory within the kernel window\n");
         return false;
     }
 
@@ -968,13 +1011,6 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
 
     for (word_t i = 0; i < ARRAY_SIZE(ndks_boot.freemem); i++) {
         ndks_boot.freemem[i] = REG_EMPTY;
-    }
-
-    /* convert the available regions to pptrs */
-    for (word_t i = 0; i < n_available; i++) {
-        avail_reg[i] = paddr_to_pptr_reg(available[i]);
-        avail_reg[i].end = ceiling_kernel_window(avail_reg[i].end);
-        avail_reg[i].start = ceiling_kernel_window(avail_reg[i].start);
     }
 
     word_t a = 0;
--- a/src/arch/x86/kernel/boot_sys.c
+++ b/src/arch/x86/kernel/boot_sys.c
@@ -194,7 +194,7 @@ static BOOT_CODE bool_t add_mem_p_regs(p_region_t reg)
         // Some of the callers round down the end address to
         return true;
     }
-    if (reg.end > PADDR_TOP && reg.start > PADDR_TOP) {
+    if (reg.end > PADDR_TOP && reg.start >= PADDR_TOP) {
         /* Return true here as it's not an error for there to exist memory outside the kernel window,
          * we're just going to ignore it and leave it to be given out as device memory */
         return true;
@@ -204,7 +204,7 @@ static BOOT_CODE bool_t add_mem_p_regs(p_region_t reg)
         return false;
     }
     if (reg.end > PADDR_TOP) {
-        assert(reg.start <= PADDR_TOP);
+        assert(reg.start < PADDR_TOP);
         /* Clamp a region to the top of the kernel window if it extends beyond */
         reg.end = PADDR_TOP;
     }


kent-mcleod:
Are you able to explain more why you think the previous code was problematic?
`check_available_memory` checks if the declared memory regions are valid and if not returns an error that aborts the boot.  The new warnings diagnostics if memory regions are out of the kernel window will create many spurious warnings on 32-bit architectures that have lots of memory outside of the kernel window which shouldn't be a warning.  If you have different contexts where this should be a warning, then you could add warnings in the root tasks when it processes the initial untyped list.

axel-h:
This produces output that is helpful for porting - and thus falls into the category that we lack a mechanism to have more verbose output for this case and allow turning this off otherwise. I agree that on 32-bit this might be annoying due to the address space limitation.
Aside from this:
- `PADDR_BASE` is checked also, not just `PADDR_TOP`. Matters only when `PADDR_BASE` is not 0.
- catch the case where `n_available and `avail_reg[MAX_NUM_FREEMEM_REG]` size differs. This usually indicates a porting or configuration problem.
- things are consolidated in `check_available_memory()`.

axel-h:
Don't prefix the message with "WARNING" and add comments in the code that this may happen on 32-bit architectures.

src/kernel/boot.c
```diff
@@ -839,24 +834,63 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+            return 0;
+        }
+
+        /* Sanitize the available memory region, only such memory can be used
+         * that is visible in the kernel window. On 32-bit architectures this
+         * is not be uncommon, because the 32-bit virtual address space is too
+         * small to support a kernel windows that is big enough.
+         */
+        if ((r->start > PADDR_TOP) || (r->end <= PADDR_BASE)) {
+            printf("    can't use region, outside of kernel window\n");
```
lsf37, line 859: If I understand the boot code correctly, removing things from `freemem` will mean that they are used for device untyped. So the memory is still used, just not available for other kernel objects. 

```suggestion
            printf("    Region outside of kernel window, will use for device untypeds instead.\n");
```

src/kernel/boot.c
```diff
@@ -810,26 +810,21 @@ BOOT_CODE void bi_finalise(void)
...
-{
     /* The system configuration is broken if no region is available. */
     if (0 == n_available) {
         printf("ERROR: no memory regions available\n");
-        return false;
+        return 0;
     }
 
+    word_t cnt = 0;
     printf("available phys memory regions: %"SEL4_PRIu_word"\n", n_available);
```
lsf37, line 828: Not from this PR, but if we're changing the logging, we might keep the capitalisation consistent with the `Bootstrapping kernel` etc messages, which start with upper case.

```suggestion
    printf("Available phys memory regions: %"SEL4_PRIu_word"\n", n_available);
```

src/kernel/boot.c
```diff
@@ -839,24 +834,63 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+        p_region_t usable_reg = {
+            .start = MAX(r->start, PADDR_BASE),
+            .end   = MIN(r->end, PADDR_TOP)
+        };
+
+        if ((usable_reg.start != r->start) || (usable_reg.end != r->end)) {
+            /* Region is partial addressable only, this is also not uncommon on
+             * 32-bit architectures due to the limited virtual address space.
+             */
+            printf("    kernel window addressable:"
```
lsf37, line 872:
```suggestion
            printf("    Kernel window addressable:"
```

lsf37 (dismissed):
This looks Ok from my side, I only have a few wording tweaks to indicate that the memory is not fully lost, it's just used for device untypeds instead (+ making capitalisation consistent).

I was about to suggest to print `)` instead of `]` for the regions (see #1008), but it's probably better to do that consistently in a separate PR.

We should do a hardware run and check the logs to see how verbose the new output is. Will kick that off now.

lsf37:
Doesn't seem to be bad at all, the only one I randomly saw in the logs was for AM335X (beagle bone black):

```
Bootstrapping kernel
    [80000000..a0000000]
      kernel window addressable: [80000000..9ff00000]
```

Seeing it like that, the lower-level `kernel` in the log message does make sense, so I'd be fine with just leaving it as is.

src/kernel/boot.c
lsf37: (happy to leave that one as is, see below)

src/kernel/boot.c
```diff
@@ -839,31 +834,80 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+            return 0;
+        }
+
+        /* Sanitize the available memory region, only such memory can be used
+         * for kernel objects that is visible in the kernel window. On 32-bit
+         * architectures loosing some memory here is not uncommon, because the
+         * 32-bit virtual address space is too small to support a kernel window
+         * that is big enough. Memory outside of he kernel window is not list,
+         * it is made available as device untyped memory.
+         */
```
lsf37, line 860: This should be rephrased a bit. Maybe:

> Sanitize the available memory region: only memory in the kernel window can be used for kernel objects. On 32-bit architectures it is not uncommon for provided regions for provided regions to be only partially in the kernel window, because the available address space is small. Memory outside of the kernel window is made available as device untyped memory.

lsf37 (commented):
The protocol is usually to not make any more substantial changes after approval, but just to implement review feedback. New changes should go into a new PR.

In this case, I'm fine with the change, but have some tweaks for comments.

src/kernel/boot.c
```diff
@@ -839,31 +834,80 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
         /* Available regions can't be empty. */
         if (r->start == r->end) {
             printf("ERROR: memory region %"SEL4_PRIu_word" empty\n", i);
-            return false;
+            return 0;
         }
 
         /* Regions must be ordered and must not overlap. Regions are [start..end),
-           so the == case is fine. Directly adjacent regions are allowed. */
+         * so the '=='-case is fine. Directly adjacent regions are allowed.
```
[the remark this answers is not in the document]
axel-h: done

src/kernel/boot.c
axel-h: done

src/kernel/boot.c
```diff
@@ -896,31 +891,79 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+             * 32-bit architectures due to the limited virtual address space.
+             */
+            if (usable_reg.start != r->start) {
+                printf("    [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] device untyped memory\n",
+                       r->start, usable_reg.end);
+            }
+            printf("    [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] kernel window addressable\n",
+                   usable_reg.start, usable_reg.end);
+            if (usable_reg.end != r->end) {
+                printf("    [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] device untyped memory\n",
```
lsf37: We have since changed the printing to be explicit about the end of the region being exclusive, i.e. `[..)`. I'll fix that up.

src/kernel/boot.c
```diff
@@ -896,31 +891,79 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+
+        /* Sanitize the available memory region: only memory in the kernel
+         * window can be used for kernel objects. On 32-bit architectures it is
+         * not uncommon for provided regions to be only partially in the kernel
+         * window, because the available address space is small. Memory outside
+         * of the kernel window is made available as device untyped memory.
+         */
+        if ((r->start > PADDR_TOP) || (r->end <= PADDR_BASE)) {
+            printf("    region outside of kernel window, available as device untypeds.\n");
+            continue;
```
lsf37: `continue` is fine for the C parser, but is a problem for the binary verification tool chain. We might need to to slightly uglify this to avoid `continue`.

src/kernel/boot.c
```diff
@@ -896,31 +891,79 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+        if (cnt >= ARRAY_SIZE(avail_reg)) {
+            /* No space left in the array, seems MAX_NUM_FREEMEM_REG should be
+             * increased. Debug builds raise an assert here because this is
+             * likely a porting issue that should be looked into. Release builds
+             * will continue booting, but can't use this memory region. This
+             * might still be sufficient to run.
+             */
+            printf("    WARNING: can't use region, avail_reg[] is full\n");
+            assert(0);
+            continue;
```
lsf37: (same `continue` problem here)

src/kernel/boot.c
```diff
@@ -896,31 +891,79 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+            .end   = MIN(r->end, PADDR_TOP)
+        };
+
+        if ((usable_reg.start != r->start) || (usable_reg.end != r->end)) {
+            /* Region is partial addressable only, this is also not uncommon on
+             * 32-bit architectures due to the limited virtual address space.
+             */
+            if (usable_reg.start != r->start) {
+                printf("    [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] device untyped memory\n",
+                       r->start, usable_reg.end);
```
lsf37: Small bug:

```suggestion
                printf("    [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] device untyped memory\n",
                       r->start, usable_reg.start);
```

src/kernel/boot.c
```diff
@@ -896,31 +891,79 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+            return 0;
+        }
+
+        /* Sanitize the available memory region: only memory in the kernel
+         * window can be used for kernel objects. On 32-bit architectures it is
+         * not uncommon for provided regions to be only partially in the kernel
+         * window, because the available address space is small. Memory outside
+         * of the kernel window is made available as device untyped memory.
+         */
+        if ((r->start > PADDR_TOP) || (r->end <= PADDR_BASE)) {
```
lsf37: This can't be right -- a region `[PADDR_TOP, PADDR_TOP+0x10)` would become a `usable_reg = [PADDR_TOP, PADDR_TOP)` below, which is empty. (Hey, did I ever mention how cool inclusive-end regions are, they can't be empty..).

I think we want this:

```suggestion
        if ((r->start >= PADDR_TOP) || (r->end <= PADDR_BASE)) {
```

Still need to double-check if the end part is correct, but I think it is.

lsf37: (I think `add_mem_p_regs` has the same bug, but it is harmless to produce empty regions there. Here it will produce confusing output)

src/kernel/boot.c
```diff
@@ -867,27 +867,22 @@ BOOT_CODE void bi_finalise(void)
     };
 }
 
-BOOT_CODE static inline pptr_t ceiling_kernel_window(pptr_t p)
+BOOT_CODE static word_t check_available_memory(word_t n_available,
```
lsf37: The name `check_available_memory` is no longer right. It no longer just checks, it populates `avail_reg`. I would propose `init_avail_reg`.

lsf37: The more I think about it, the less I like that this function is no longer pure. I might have a go at splitting it into two parts (pure check and then init) and see if that looks any better. It would mean going through the regions twice, but these are not large arrays, so that should not be a problem, and thinking about verifying this in the future, having all of these checks separate from the state effects is probably nicer.

src/kernel/boot.c
lsf37: We should also call it "potential device untyped memory", because if it overlaps with a reserved region it will not become device untyped and the message will be confusing when it then doesn't appear as device untyped in the end.

src/kernel/boot.c
```diff
@@ -867,27 +867,22 @@ BOOT_CODE void bi_finalise(void)
...
+                                               const p_region_t *available)
 {
-    /* Adjust address if it exceeds the kernel window
-     * Note that we compare physical address in case of overflow.
-     */
-    if (pptr_to_paddr((void *)p) > PADDR_TOP) {
-        p = PPTR_TOP;
+    /* Ensure avail_reg is properly initialized. */
```
midnightveil: Is this actually necessary if `REG_EMPTY` is all 0?

lsf37: It's not and it would be semantically fine to leave out. I was even about to remove it, but then I remembered that I don't want to have to reason in verification that it's still all 0 when this function gets called, so I left it in.

lsf37:
I have pushed an additional commit that fixes up everything apart from `continue` (still need to do that), and it splits the function into a pure check and a stateful part (like it was before). Easiest to review end-to-end this time, because it mostly touches things the first commit has changed and should be squashed if we're happy with it.

src/kernel/boot.c
lsf37: Now that you made me think about it more, we can actually leave it out: we never need to know what the previous values of `avail_reg` are, we only write to it and we only care about the indices that we have written to. Will remove.

lsf37:
In terms of noise impact, now that hardware tests are working for a change, we have some data. We're getting a variation of a single region warning like this for exactly 4 platforms AM335X_BONEBLACK, OMAP3, TK1, and ZYNQMP (32 bit runs)

```
Bootstrapping kernel
Available phys memory regions: 1
  [80000000..a7f00000)
  region [80000000..a7f00000) is only partially in the kernel window:
    [80000000..9ff00000) kernel window addressable
    [9ff00000..a7f00000) potential device untyped memory
```

I'm not sure the second line adds much useful information, but it might be nice to know which part is in the kernel window. If we only report the part that's in the kernel window, then the additional output would be reduced to just one line, e.g. like this:

```
Bootstrapping kernel
Available phys memory regions: 1
  [80000000..a7f00000)
    only [80000000..9ff00000) in kernel window, rest is potential device untyped memory
```

What do people think?

lsf37 (approved):
I would now be happy with the outcome (after commits are squashed), but given that I tweaked a lot of this, it'd be good to get a second review.

midnightveil:
I think of the two you presented I prefer the first one.

If anything, I'd prefer reporting only that '[9ff00000..a7f00000) lies outside kernel window and will not appear as normal UT"; as that is the exceptional case.

I am worried that this feels like something is *wrong*, even if it's expected and OK. IDK.

lsf37:
> I think of the two you presented I prefer the first one.
> 
> If anything, I'd prefer reporting only that '[9ff00000..a7f00000) lies outside kernel window and will not appear as normal UT"; as that is the exceptional case.
> 
> I am worried that this feels like something is _wrong_, even if it's expected and OK. IDK.

It does have that niggle, which Kent was also worried about up there in the comments somewhere. We can also just drop this bit, since nothing is wrong and this really is the expected behaviour. The actual improvements in the PR are the additional checks, which are in the other function now anyway.

If there is a region passed in that is completely outside the kernel window it's probably still worth a warning (because you could just leave that out).

lsf37:
I have removed the additional printing for regions that are partially in the kernel window. I think it will lead to too much confusion whether this is a bad thing or not when it is not. Regions completely outside the window might be worth looking at, so I left that in.

There was some proof fallout from pointer name mangling. If people are otherwise happy with this PR now, I can have a look at that, it should be trivial. 

I have also added the off-by-one fix on x86 boot that is parallel to the one that was in this code.

midnightveil (approved):

src/kernel/boot.c
```diff
@@ -889,40 +878,92 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
             return false;
         }
     }
 
     return true;
 }
 
+/* We can't declare arrays on the stack, so this is space for init_avail_reg()
+ * below to populate and init_freemem() to consume. */
+BOOT_BSS static region_t avail_reg[MAX_NUM_FREEMEM_REG];
```
Indanz, line 913: Normally I would pass `avail_reg` as an argument to `init_avail_reg()` instead of sharing a global variable like this. I had to look twice to see what was going on, it's a bit unexpected.

src/kernel/boot.c
```diff
@@ -889,40 +878,92 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
...
+                printf("  WARNING: cannot use region [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"),"
+                       " avail_reg[] is full\n", r->start, r->end);
+                assert(0);
+            } else {
+                avail_reg[cnt] = paddr_to_pptr_reg(usable_reg);
+                cnt++;
+            }
+        }
+    }
+
```
Indanz: Stray empty new line.

Indanz (approved):
Just some nitpicks, otherwise looks fine.

src/kernel/boot.c
lsf37: Yes, there's not much we can do about this one. Arrays on the stack don't work on the verification side, because they create pointers into the stack (which the memory model disallows). Passing it as an argument just shifts it to the caller stack. I guess we could pass the global as an argument, but that also is a bit strange.

lsf37: (This is old, though, the commit just moves it up to the new function)

Indanz: > I guess we could pass the global as an argument, but that also is a bit strange.

Better than what we have now. Then at least it's conceptually on the stack of `init_freemem()` and the comment makes sense. Now you've split of `init_avail_reg()`, it's a true global variable and can't be on the stack of either function.

(If you put the array inside a struct you probably could put it on the stack, but it might be too big for that anyway.)

midnightveil: I do somewhat agree - several of these boot functions depend on global state (e.g. the ni_sdk_boot) and it could pass data around explicitly. But it's really not that bad, only when it goes cross-file.

Struct-with-array-on-stack has same problem as array-on-stack: kernel C subset does not permit taking refernces to locations on the stack.

Unless the proposal is to pass the struct-with-array by value, so that you don't have to take the address of it. But ew, that's a lot of data copying.

lsf37: Just tried this out and it's not great. The ARRAY_SIZE macro doesn't work on an array that is a parameter, because that just decays to a pointer. This is the same problem we'll have in verification when we get here, because we'd need to pass through precondtions about the size of this array and the fact that it actually is an array etc, which just complicates things over a single-use global array. Usually globals would be bad, because they can be modified from anywhere, but in this kind of scenario they are easy because we never read from this global again after init_free_mem.

Edit: this overlapped. "this" = my proposal to pass the global as a parameter.

lsf37: I agree that wrapping in a struct is not going to work.

midnightveil: Yeah, makes sense. It's fine.

PR #729, opened 2021-12-28
risc-v: remove unused/empty files

axel-h:
Seems these files were created as part of the initial port, but never received any content. Remove it to avoid implying anything here.


lsf37 (approved):
Happy with removing them.

PR #732, opened 2021-12-31
clean up compile_assert()

axel-h:
- use `_Static_assert()` in kernel builds (except verification) to have nicer error messages. Both gcc and clang support it even for c99.
- simplify macro `seL4_CompileTimeAssert()` to use `SEL4_COMPILE_ASSERT()`

### DIFF
--- a/include/assert.h
+++ b/include/assert.h
@@ -42,14 +42,28 @@ void _assert_fail(
 
 #endif /* DEBUG */
 
-/* Create an assert that will trigger a compile error if it fails. */
+/* Create an assert that triggers a compile error if the condition fails. We do
+ * not include sel4/macros.h that provides SEL4_COMPILE_ASSERT() for two
+ * reasons:
+ * - The kernel's source internals shall not have any unnecessary dependency on
+ *     the user interface headers.
+ * - The kernel user API headers aims to be compiler agnostic and stick to the
+ *     standard(s). As _Static_assert() is a c11 feature, the c99 used for
+ *     kernel compilation would use a helper macros. While this works, it
+ *     creates strange error messages then the condition fails.. Since kernel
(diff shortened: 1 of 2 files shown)


GITHUBGUY12357999 (changes requested):
Pls check all codes which are in red and re write them and then see if they are good and then request for approval thank you.

GITHUBGUY12357999 (changes requested):
Please check all codes again and re write them thank you.

GITHUBGUY12357999 (approved):
I approve

include/assert.h
```diff
@@ -43,8 +44,7 @@ void _assert_fail(
 #endif /* DEBUG */
 
 /* Create an assert that will trigger a compile error if it fails. */
-#define compile_assert(name, expr) \
-        typedef int __assert_failed_##name[(expr) ? 1 : -1] UNUSED;
+#define compile_assert(name, expr) SEL4_COMPILE_ASSERT(name, expr)
```
kent-mcleod, line 47: The kernel is only supposed to use type and constant definitions from libsel4 when they relate to communication between the kernel and user level and so it doesn't make sense for the kernel's private definition of compile_assert to be coupled to the implementation in libsel4.

axel-h: Ok, I will remove this change then. But I would like to change this to use `_Static_assert()` (or whatever native way the current compiler implements) whenever this exists. The error messages are just nicer there and it's a bit less hacky. Also, on the pragmatic side both gcc and clang support `_Static_assert()` even when `-std=c99` is used

to use the definition from libsel4 here then and see that whenever the compiler support a native compile time assert, this is preferred. That's a bit more convenient regarding the error messages

kent-mcleod (approved):

PR #733, opened 2021-12-31
Add support for generating syscall_stubs for different languages

kgugala:
This PR is related to discussion in the devel mailing list. It splits syscall_stubs generator into language specific backends. It also adds Rust and C backends.

Any comments are appreciated.

### DIFF
--- a/libsel4/tools/syscall_stub_gen_rs.py
+++ b/libsel4/tools/syscall_stub_gen_rs.py
@@ -1,0 +1,543 @@
+#!/usr/bin/env python3
+#
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2017, Ember Arlynx
+# Copyright (c) 2022 Google LLC
+# Copyright (c) 2022 Antmicro
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# This script is intended to work as a Rust-specific copy of syscall_stub_gen.py
+# Use it with sel4-sys crate to allow Rust applications inteface with the seL4 API.
+
+from argparse import ArgumentParser
+import sys
+import syscall_stub_gen
+
+
+TYPES = {
+    8:  "u8",
+    16: "u16",
+    32: "u32",
+    64: "u64"
+}
+
+
+def translate_expr(name):
+    if name == "type":
+        return "type_"
+    return name
+
+
+def translate_type(name):
+    type_trans = {
+        "int": "isize",
+        "seL4_Uint8": "u8",
+        "seL4_Uint16": "u16",
+        "seL4_Uint32": "u32",
+        "seL4_Uint64": "u64",
+        "seL4_Bool": "u8",
+        "seL4_CapData_t": "seL4_CapData",
+        "seL4_PrioProps_t": "seL4_PrioProps",
+        "seL4_CapRights_t": "seL4_CapRights",
+        "seL4_RISCV_Page_GetAddress_t": "seL4_RISCV_Page_GetAddress",
+    }
+    if name in type_trans:
+        return type_trans[name]
+    else:
+        return name
+
+
+def rust_render_parameter_name(param_type, name, mutable_ref=False):
+    """
+    Return a string of Rust code that would be used in a function
+    parameter declaration.
+    """
+    if type(param_type) in [syscall_stub_gen.Type, syscall_stub_gen.CapType, syscall_stub_gen.StructType, syscall_stub_gen.BitFieldType]:
+        name = translate_expr(name)
+        return "%s: %s" % (name, translate_type(param_type.name))
+    elif type(param_type) is syscall_stub_gen.PointerType:
+        return "%s: *%s %s" % (
+            translate_expr(name),
+            "mut" if mutable_ref else "const",
+            translate_type(param_type.name))
+    else:
+        raise Exception("Unsupported type")
+
+
+def rust_double_word_expression(param_type, var_name, word_num, word_size):
+
+    assert word_num == 0 or word_num == 1
+
+    if word_num == 0:
+        return "{1} as {0}".format(TYPES[param_type.size_bits], var_name)
+    elif word_num == 1:
+        return "{1}.wrapping_shr({2}) as {0}".format(TYPES[param_type.size_bits], var_name,
+                                                     word_size)
+
+
+def rust_expression(param_type, var_name, word_num=0, member_name=None):
+    """
+    Return code for an expression that gets word 'word_num'
+    of the provided type.
+    """
+    if type(param_type) is syscall_stub_gen.BitFieldType:
+        return "%s.words[%d]" % (var_name, word_num)
+    elif type(param_type) is syscall_stub_gen.StructType:
+        assert word_num < param_type.size_bits // param_type.wordsize
+
+        # Multiword structure.
+        assert param_type.pass_by_reference()
+        return "(*%s).%s" % (var_name, member_name[word_num])
+    elif type(param_type) is syscall_stub_gen.PointerType:
+        assert word_num == 0
+        return "unsafe { *%s }" % var_name
+    elif type(param_type) in [syscall_stub_gen.Type, syscall_stub_gen.CapType]:
+        assert word_num == 0
+        return "%s" % var_name
+    else:
+        raise Exception("Unsupported type")
+
+
+def generate_param_list(input_params, output_params):
+    # Generate parameters
+    params = []
+    for param in input_params:
+        t = param.type
+        if not t.pass_by_reference():
+            params.append(rust_render_parameter_name(t, param.name))
+        else:
+            params.append(rust_render_parameter_name(t.pointer(), param.name, mutable_ref=False))
+    for param in output_params:
+        # We drop params that aren't pass_by_reference deliberately
+        if param.type.pass_by_reference():
+            params.append(rust_render_parameter_name(
+                param.type.pointer(), param.name, mutable_ref=True))
+
+    return ", ".join(params)
+
+
+def generate_marshal_expressions(params, num_mrs, structs, wordsize):
+    """
+    Generate marshalling expressions for the given set of inputs.
+
+    We return a list of expressions; one expression per word required
+    to marshal all the inputs.
+    """
+
+    def generate_param_code(param, first_bit, num_bits, word_array, wordsize):
+        """
+        Generate code to marshal the given parameter into the correct
+        location in the message.
+
+        'word_array' is an array of the final contents of the message.
+        word_array[k] contains what should be placed in the k'th message
+        register, and is an array of expressions that will (eventually)
+        be bitwise-or'ed into it.
+        """
+
+        target_word = first_bit // wordsize
+        target_offset = first_bit % wordsize
+
+        # double word type
+        if param.type.double_word:
+            word_array[target_word].append(
+                rust_double_word_expression(param.type, param.name, 0, wordsize))
+            word_array[target_word +
+                       1].append(rust_double_word_expression(param.type, param.name, 1, wordsize))
+            return
+
+        # Single full word?
+        if num_bits == wordsize:
+            assert target_offset == 0
+            expr = rust_expression(param.type, param.name)
+            word_array[target_word].append(expr)
+            return
+
+        # Part of a word?
+        if num_bits < wordsize:
+            expr = rust_expression(param.type, param.name)
+            expr = "(%s & %#x)" % (expr, (1 << num_bits) - 1)
+            if target_offset:
+                expr = "(%s as seL4_Word).wrapping_shl(%d)" % (expr, target_offset)
+            word_array[target_word].append(expr)
+            return
+
+        # Multiword array
+        assert target_offset == 0
+        num_words = num_bits // wordsize
+        for i in range(num_words):
+            expr = rust_expression(param.type, param.name, i,
+                                   syscall_stub_gen.struct_members(param.type, structs))
+            word_array[target_word + i].append(expr)
+
+    # Get their marshalling positions
+    positions = syscall_stub_gen.get_parameter_positions(params, wordsize)
+
+    # Generate marshal code.
+    words = [[] for _ in range(num_mrs, syscall_stub_gen.MAX_MESSAGE_LENGTH)]
+    for (param, first_bit, num_bits) in positions:
+        generate_param_code(param, first_bit, num_bits, words, wordsize)
+
+    # Return list of expressions.
+    return [" | ".join(map(lambda x: "(" + translate_expr(x) + " as seL4_Word)", x))
+            for x in words if len(x) > 0]
+
+
+def generate_unmarshal_expressions(params, wordsize):
+    """
+    Generate unmarshalling expressions for the given set of outputs.
+
+    We return a list of list of expressions; one list per variable, containing
+    expressions for the words in it that must be unmarshalled. The expressions
+    will have tokens of the form:
+        "%(w0)s"
+    in them, indicating a read from a word in the message.
+    """
+
+    def unmarshal_single_param(first_bit, num_bits, wordsize):
+        """
+        Unmarshal a single parameter.
+        """
+        first_word = first_bit // wordsize
+        bit_offset = first_bit % wordsize
+
+        # Multiword type?
+        if num_bits > wordsize:
+            result = []
+            for x in range(num_bits // wordsize):
+                result.append("%%(w%d)s" % (x + first_word))
+            return result
+
+        # Otherwise, bit packed.
+        if num_bits == wordsize:
+            return ["%%(w%d)s" % first_word]
+        elif bit_offset == 0:
+            return ["(%%(w%d)s & %#x)" % (
+                first_word, (1 << num_bits) - 1)]
+        else:
+            return ["(%%(w%d)s.wrapping_shr(%d)) & %#x" % (
+                first_word, bit_offset, (1 << num_bits) - 1)]
+
+    # Get their marshalling positions
+    positions = syscall_stub_gen.get_parameter_positions(params, wordsize)
+
+    # Generate the unmarshal code.
+    results = []
+    for (param, first_bit, num_bits) in positions:
+        results.append((param, unmarshal_single_param(first_bit, num_bits, wordsize)))
+    return results
+
+
+def generate_result_struct(interface_name, method_name, output_params):
+    # Do we actually need a structure?
+    if not syscall_stub_gen.is_result_struct_required(output_params):
+        return None
+
+    # Generate the structure:
+    #
+    #   #[repr(C)]
+    #   #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    #   pub struct seL4_SchedContext_Consumed {
+    #       pub error: isize,
+    #       pub consumed: seL4_Time,
+    result = []
+    result.append("#[repr(C)]")
+    result.append("#[derive(Debug, Clone, Copy, PartialEq, Eq)]")
+    result.append("pub struct %s_%s {" % (interface_name, method_name))
+    result.append("\tpub error: isize,")
+    for i in output_params:
+        if not i.type.pass_by_reference():
+            result.append("\tpub %s," % rust_render_parameter_name(i.type, i.name))
+    result.append("}")
+    result.append("")
+    return "\n".join(result)
+
+
+def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_params, output_params, structs, use_only_ipc_buffer, comment, mcs):
+    result = []
+
+    if use_only_ipc_buffer:
+        num_mrs = 0
+    else:
+        if mcs and "%s-mcs" % arch in syscall_stub_gen.MESSAGE_REGISTERS_FOR_ARCH:
+            num_mrs = syscall_stub_gen.MESSAGE_REGISTERS_FOR_ARCH["%s-mcs" % arch]
+        else:
+            num_mrs = syscall_stub_gen.MESSAGE_REGISTERS_FOR_ARCH[arch]
+
+    # Split out cap parameters and standard parameters
+    standard_params = []
+    cap_params = []
+    for x in input_params:
+        if isinstance(x.type, syscall_stub_gen.CapType):
+            cap_params.append(x)
+        else:
+            standard_params.append(x)
+
+    # Determine if we are returning a structure, or just the error code.
+    returning_struct = False
+    results_structure = generate_result_struct(interface_name, method_name, output_params)
+    if results_structure:
+        return_type = "%s_%s" % (interface_name, method_name)
+        returning_struct = True
+    else:
+        return_type = "seL4_Result"
+
+    #
+    # Print doxygen comment.
+    #
+    result.append(comment)
+
+    #
+    # Print function header.
+    #
+    #   #[inline(always)]
+    #   #[macros::derive_test_wrapper]
+    #   pub unsafe fn seL4_Untyped_Retype(...) -> seL4_Result
+    #   {
+    #
+    result.append("#[inline(always)]")
+    result.append("pub unsafe fn %s_%s(%s) -> %s" % (interface_name, method_name,
+                                                     generate_param_list(input_params, output_params), return_type))
+    result.append("{")
+
+    #
+    # Get a list of expressions for our caps and inputs.
+    #
+    input_expressions = generate_marshal_expressions(standard_params, num_mrs,
+                                                     structs, wordsize)
+    cap_expressions = [x.name for x in cap_params]
+    service_cap = cap_expressions[0]
+    cap_expressions = cap_expressions[1:]
+
+    #
+    # Setup variables we will need.
+    #
+    if returning_struct:
+        result.append("\tlet mut result: %s = ::core::mem::zeroed();" % return_type)
+    result.append("\tlet tag = seL4_MessageInfo::new(InvocationLabel::%s as seL4_Word, 0, %d, %d);" % (
+        method_id, len(cap_expressions), len(input_expressions)))
+
+    for i in range(num_mrs):
+        result.append("\tlet mut mr%d: seL4_Word = 0;" % i)
+    result.append("")
+
+    #
+    # Copy capabilities.
+    #
+    #   /* Setup input capabilities. */
+    #   seL4_SetCap(i, cap);
+    #
+    if len(cap_expressions) > 0:
+        result.append("\t/* Setup input capabilities. */")
+        for i in range(len(cap_expressions)):
+            result.append("\tseL4_SetCap(%d, %s);" % (i, cap_expressions[i]))
+        result.append("")
+
+    #
+    # Copy in the inputs.
+    #
+    #   /* Marshal input parameters. */
+    #   mr0 = (start_offset as seL4_Word) as seL4_Word;
+    #   ...
+    #   seL4_SetMR(i, v);
+    #   ...
+    #
+    if max(num_mrs, len(input_expressions)) > 0:
+        result.append("\t/* Marshal and initialize parameters. */")
+        # Initialise in-register parameters
+        for i in range(num_mrs):
+            if i < len(input_expressions):
+                result.append("\tmr%d = %s as seL4_Word;" % (i, input_expressions[i]))
+            else:
+                result.append("\tmr%d = 0 as seL4_Word;" % i)
+        # Initialise buffered parameters
+        for i in range(num_mrs, len(input_expressions)):
+            expression = translate_expr(input_expressions[i])
+            result.append("\tseL4_SetMR(%d, %s);" % (i, expression))
+        result.append("")
+
+    #
+    # Generate the call.
+    #
+    if use_only_ipc_buffer:
+        result.append("\t/* Perform the call. */")
+        result.append("\tlet output_tag = seL4_Call(%s, tag);" % service_cap)
+    else:
+        result.append("\t/* Perform the call, passing in-register arguments directly. */")
+        result.append("\tlet output_tag = seL4_CallWithMRs(%s, tag," % (service_cap))
+        result.append("\t\t%s);" % ', '.join(
+            ("&mut mr%d" % i) for i in range(num_mrs)))
+
+    #
+    # Prepare the result.
+    #
+
+    if returning_struct:
+        result.append("\tresult.error = output_tag.get_label() as _;")
+    else:
+        result.append("\tlet error: seL4_Error = output_tag.get_label().into();")
+    result.append("")
+
+    if not use_only_ipc_buffer:
+        result.append("\t/* Unmarshal registers into IPC buffer on error. */")
+        result.append("\tif error != seL4_Error::seL4_NoError {")
+        for i in range(num_mrs):
+            result.append("\t\tseL4_SetMR(%d, mr%d);" % (i, i))
+        if returning_struct:
+            result.append("\t\treturn result;")
+        result.append("\t}")
+        result.append("")
+
+    #
+    # Generate unmarshalling code.
+    #
+    if len(output_params) > 0:
+        result.append("\t/* Unmarshal result. */")
+        source_words = {}
+        for i in range(syscall_stub_gen.MAX_MESSAGE_LENGTH):
+            if i < num_mrs:
+                source_words["w%d" % i] = "mr%d" % i
+            else:
+                source_words["w%d" % i] = "seL4_GetMR(%d)" % i
+        unmashalled_params = generate_unmarshal_expressions(output_params, wordsize)
+        for (param, words) in unmashalled_params:
+            param.name = translate_expr(param.name)
+            if param.type.pass_by_reference():
+                members = syscall_stub_gen.struct_members(param.type, structs)
+                for i in range(len(words)):
+                    result.append("\t(*%s).%s = %s;" %
+                                  (param.name, members[i], words[i] % source_words))
+            else:
+                if param.type.double_word:
+                    result.append("\tresult.%s = (%s as %s) + ((%s as %s).wrapping_shr(32));" %
+                                  (param.name, words[0] % source_words, TYPES[64],
+                                   words[1] % source_words, TYPES[64]))
+                else:
+                    for word in words:
+                        if type(param.type) is syscall_stub_gen.StructType:
+                            result.append("\tresult.%s = %s { words: [%s] };" % (
+                                param.name, param.type.name, word % source_words))
+                        else:
+                            result.append("\tresult.%s = %s as %s;" % (param.name, word %
+                                                                       source_words, translate_type(param.type.name)))
+
+    #
+    # }
+    #
+    result.append("\tresult" if returning_struct else "\terror.into()")
+    result.append("}")
+
+    return "\n".join(result) + "\n"
+
+
+def generate_stub_file(arch, wordsize, input_files, output_file, use_only_ipc_buffer, mcs):
+    """
+    Generate a header file containing system call stubs for seL4.
+    """
+    result = []
+
+    data_types = syscall_stub_gen.init_data_types(wordsize)
+    arch_types = syscall_stub_gen.init_arch_types(wordsize)
+
+    # Parse XML
+    methods = []
+    structs = []
+    for infile in input_files:
+        method, struct, _ = syscall_stub_gen.parse_xml_file(infile, data_types + arch_types[arch])
+        methods += method
+        structs += struct
+
+    # Print header.
+    result.append("""
+/*
+ * Automatically generated system call stubs.
+ */
+
+""")
+
+    #
+    # Generate structures needed to return results back to the user.
+    #
+    # We can not use pass-by-reference (except for really large objects), as
+    # the verification framework does not support them.
+    #
+    result.append("/*")
+    result.append(" * Return types for generated methods.")
+    result.append(" */")
+    for (interface_name, method_name, _, _, output_params, _, _) in methods:
+        results_structure = generate_result_struct(interface_name, method_name, output_params)
+        if results_structure:
+            result.append(results_structure)
+
+    #
+    # Generate the actual stub code.
+    #
+    result.append("/*")
+    result.append(" * Generated stubs.")
+    result.append(" */")
+    for (interface_name, method_name, method_id, inputs, outputs, condition, comment) in methods:
+        if condition != "":
+            if condition == "(!defined CONFIG_KERNEL_MCS) && CONFIG_MAX_NUM_NODES > 1":
+                # NB: CONFIG_MAX_NUM_NODES > 1 =>'s CONFIG_SMP_SUPPORT
+                condition = 'all(not(feature = "CONFIG_KERNEL_MCS"), feature = "CONFIG_SMP_SUPPORT")'
+            elif condition == "CONFIG_MAX_NUM_NODES > 1":
+                condition = 'feature = "CONFIG_SMP_SUPPORT"'
+            elif condition:
+                condition = condition.replace('defined', '')
+                condition = condition.replace('(', '')
+                condition = condition.replace(')', '')
+                if 'CONFIG_' in condition:
+                    condition = 'feature = "' + condition + '"'
+                if '!' in condition:
+                    condition = 'not(%s)' % condition.replace('!', '')
+            if condition:
+                result.append("#[cfg(%s)]" % condition)
+        result.append(generate_stub(arch, wordsize, interface_name, method_name,
+                                    method_id, inputs, outputs, structs, use_only_ipc_buffer, comment, mcs))
+
+    result.append("")
+
+    # Write the output
+    output = open(output_file, "w")
+    output.write("\n".join(result))
+    output.close()
+
+
+def process_args():
+    usage_str = """
+    %(prog)s [OPTIONS] [FILES] """
+    epilog_str = """
+
+    """
+    parser = ArgumentParser(description='seL4 System Call Stub Generator.',
+                            usage=usage_str,
+                            epilog=epilog_str)
+    parser.add_argument("-o", "--output", dest="output", default="/dev/stdout",
+                        help="Output file to write stub to. (default: %(default)s).")
+    parser.add_argument("-b", "--buffer", dest="buffer", action="store_true", default=False,
+                        help="Use IPC buffer exclusively, i.e. do not pass syscall arguments by registers. (default: %(default)s)")
+    parser.add_argument("-a", "--arch", dest="arch", required=True, choices=syscall_stub_gen.WORD_SIZE_BITS_ARCH,
+                        help="Architecture to generate stubs for.")
+    parser.add_argument("--mcs", dest="mcs", action="store_true",
+                        help="Generate MCS api.")
+
+    parser.add_argument("files", metavar="FILES", nargs="+",
+                        help="Input XML files.")
+
+    return parser
+
+
+def main():
+
+    parser = process_args()
+    args = parser.parse_args()
+
+    wordsize = syscall_stub_gen.WORD_SIZE_BITS_ARCH[args.arch]
+    # Generate the stubs.
+    generate_stub_file(args.arch, wordsize, args.files, args.output, args.buffer, args.mcs)
+
+
+if __name__ == "__main__":
+    sys.exit(main())


libsel4/tools/cgenerator.py
```diff
@@ -0,0 +1,598 @@
...
+                CapType("seL4_ARM_VSpace", wordsize),
+                CapType("seL4_ARM_ASIDControl", wordsize),
+                CapType("seL4_ARM_ASIDPool", wordsize),
+                CapType("seL4_ARM_VCPU", wordsize),
+                CapType("seL4_ARM_IOSpace", wordsize),
+                CapType("seL4_ARM_IOPageTable", wordsize),
+                StructType("seL4_UserContext", wordsize * 36, wordsize),
+            ] + arm_smmu,
+
+            "arm_hyp": [
```
axel-h, line 217: Since `arm_hyp` is basically `aarch32`, is there a chance to map this to `aarch32` to reduce the duplication and avoid setting a bad example?

libsel4/tools/cgenerator.py
```diff
@@ -0,0 +1,598 @@
...
+            else:
+                params.append(param.type.pointer().render_parameter_name(param.name))
+        for param in output_params:
+            if param.type.pass_by_reference():
+                params.append(param.type.pointer().render_parameter_name(param.name))
+
+        return ", ".join(params)
+
+    def _gen_file_header(self):
+        self.contents.append("#ifndef __LIBSEL4_SEL4_CLIENT_H")
```
axel-h, line 306: Can we change this to use "`#pragma once`"?

libsel4/tools/cgenerator.py
```diff
@@ -0,0 +1,598 @@
...
+#define assert_size_correct(type, expected_bytes) \\
+        typedef unsigned long __type_##type##_size_incorrect[ \\
+                (sizeof(type) == expected_bytes) ? 1 : -1]
+""")
+        for x in self.data_types + self.arch_types[self.arch]:
+            self.contents.append("assert_size_correct(%s, %d);" % (x.name, x.native_size_bits / 8))
+        self.contents.append("")
+
+    def _gen_file_footer(self):
+        self.contents.append("#endif /* __LIBSEL4_SEL4_CLIENT_H */")
```
axel-h, line 331: Drop this, see https://github.com/seL4/seL4/pull/733/files#r777258403

libsel4/tools/cgenerator.py
```diff
@@ -0,0 +1,598 @@
...
+        self.contents.append("""
+/*
+ * The following code generates a compile-time error if the system call
+ * stub generator has an incorrect understanding of how large a type is.
+ *
+ * If you receive a compile-time error here, you will need to adjust
+ * the type information in the stub generator.
+ */
+#define assert_size_correct(type, expected_bytes) \\
+        typedef unsigned long __type_##type##_size_incorrect[ \\
```
axel-h, line 323: A bit out of scope, but we could include `sel4/macros.h` and then simply use 
```
#define assert_size_correct(type, expected_bytes) \\
    SEL4_COMPILE_ASSERT(type##_size_incorrect, sizeof(type) == expected_bytes)`
```

libsel4/tools/cgenerator.py
```diff
@@ -0,0 +1,598 @@
...
+        self.contents.append("")
+
+    def _gen_file_footer(self):
+        self.contents.append("#endif /* __LIBSEL4_SEL4_CLIENT_H */")
+        self.contents.append("")
+
+    def _gen_conditional_compilation(self, condition) -> str:
+        if condition != '':
+            self.contents.append("#if %s" % condition)
+            return "#endif"
```
axel-h, line 337: Could make this 
```suggestion
            return "#endif /* %s */" % condition"
```

libsel4/tools/rustgenerator.py
```diff
@@ -0,0 +1,615 @@
+#
```
axel-h, line 1: Claim a copyright here?

libsel4/tools/rustgenerator.py
```diff
@@ -0,0 +1,615 @@
...
+            "int": "isize",
+            "seL4_Uint8": "u8",
+            "seL4_Uint16": "u16",
+            "seL4_Uint32": "u32",
+            "seL4_Uint64": "u64",
+            "seL4_Bool": "u8",
+            "seL4_CapData_t": "seL4_CapData",
+            "seL4_PrioProps_t": "seL4_PrioProps",
+            "seL4_CapRights_t": "seL4_CapRights",
+            "seL4_RISCV_Page_GetAddress_t": "seL4_RISCV_Page_GetAddress",
```
axel-h, line 27: It might be useful in the long run to have a way to hide architecture specific types, so they can't get in by quirky accidented eventually.

libsel4/tools/rustgenerator.py
```diff
@@ -0,0 +1,615 @@
...
+                CapType("seL4_ARM_VSpace", wordsize),
+                CapType("seL4_ARM_ASIDControl", wordsize),
+                CapType("seL4_ARM_ASIDPool", wordsize),
+                CapType("seL4_ARM_VCPU", wordsize),
+                CapType("seL4_ARM_IOSpace", wordsize),
+                CapType("seL4_ARM_IOPageTable", wordsize),
+                StructType("seL4_UserContext", wordsize * 36, wordsize),
+            ] + arm_smmu,
+
+            "arm_hyp": [
```
axel-h, line 250: Same here, might be worth just making this an alias of `aarch32` wit  a comment that this is a legacy from the past that somebody could clean up one day (but it may never happen)

axel-h:
- We have `cgenerator.py` and `rustgenerator.py` now, I think it would scale a bit nicer if this is `generator_c.py` (`Generator_C`) and `generator_rust.py` (`Generator_Rust`).
- My comment on the C generator cleanup are follow-up things mainly, because that is out of scope for this PR. Just making `arm_hyp` something on top of `aarch32` might be worth considering to avoid setting a bad example that encourages a "risc-v_hyp" hack

libsel4/tools/syscall_stub_gen.py
```diff
@@ -1069,6 +68,8 @@ def process_args():
     wsizegroup.add_argument("-c", "--cfile", dest="cfile",
                             help="Config file for Kbuild, used to get Word size.")
 
+    parser.add_argument("-l", "--lang", dest="lang", type=str.lower,
+                        choices=("c", "rust"), default="c")
```
axel-h, line 72: I wonder, can we not set a default here and force the caller to make a deliberate choice from now on? It might cause some inconvenience as some build workflow break, but the gain is less implicit behavior.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -1069,6 +68,8 @@ def process_args():
     wsizegroup.add_argument("-c", "--cfile", dest="cfile",
```
axel-h, line 68: The naming is a bit unfortunate now, because this is not related to "C". Can't we drop this, because Kbuild is no longer in use and CMake knows `KernelWordSize`/`WORD_SIZE`, so it could pass this here. Or pass `sel4arch` and there is an internal table then that deductes  the word size

libsel4/tools/syscall_stub_gen.py
```diff
@@ -1103,8 +104,10 @@ def main():
         print("Invalid word size.")
         sys.exit(2)
 
-    # Generate the stubs.
-    generate_stub_file(args.arch, wordsize, args.files, args.output, args.buffer, args.mcs)
+    pick_generator = {'c': CGenerator, 'rust': RustGenerator}
```
axel-h, line 107: Maybe we should delegate this to a a helper class to remove it from here. It's a long term bet that there might be more, The heller would know the available generators and and could also provide the list `choices` for the `--lang` above.

libsel4/tools/rustgenerator.py
```diff
@@ -0,0 +1,615 @@
...
+
+    @staticmethod
+    def translate_expr(name):
+        # TODO there are cases which implement this in-line, refactor to use this function instead
+        #      or get rid of this
+        if name == "type":
+            return "type_"
+        return name
+
+    TYPES = {
```
axel-h, line 42: I wonder if instead of using the plain numbers, having anther indirection here might one day pay off. It might be abstraction overkill also...

libsel4/tools/rustgenerator.py
```diff
@@ -0,0 +1,615 @@
...
+        Type = RustGenerator.Type
+        return [
+            # Simple Types
+            Type("int", 32, wordsize),
+            Type("long", wordsize, wordsize),
+
+            Type("seL4_Uint8", 8, wordsize),
+            Type("seL4_Uint16", 16, wordsize),
+            Type("seL4_Uint32", 32, wordsize),
+            Type("seL4_Uint64", 64, wordsize, double_word=(wordsize == 32)),
```
axel-h, line 189: Might also be a follow-up task, but it there a way to drop this `double_word` and handle this internally where needed?

libsel4/tools/cgenerator.py
wsipak: Yes, right now we have these lines instead:
```      
# legacy alias
arch_types["arm_hyp"] = arch_types["aarch32"]
```

libsel4/tools/cgenerator.py
wsipak: Yes, I've removed include guards and used `#pragma once` instead.

libsel4/tools/cgenerator.py
wsipak: Done.

libsel4/tools/cgenerator.py
wsipak: Indeed this looks much better, I've modified this part as suggested.

libsel4/tools/cgenerator.py
wsipak: I've modified this as suggested.

libsel4/tools/rustgenerator.py
wsipak: Yes, ditto as in C generator.

libsel4/tools/generator_c.py
```diff
@@ -0,0 +1,584 @@
...
+
+        # Generate the unmarshal code.
+        results = []
+        for (param, first_bit, num_bits) in positions:
+            results.append((param, unmarshal_single_param(first_bit, num_bits, wordsize)))
+        return results
+
+    def generate_marshal_expressions(self, params, num_mrs, structs):
+        wordsize = self.wordsize
+        WORD_CONST_SUFFIX_BITS = {
```
axel-h, line 526: `<kernel>include/basic_types.h` has the macro `SEL4_WORD_CONST(x)` to hide this. I wonder if we can simply use this here and keep the generator agnostic of this.

libsel4/tools/syscall_stub_gen.py
wsipak: I've removed the default value.
From what I see, in order to make it build properly, a change in `libsel4/CMakeLists.txt` was needed, I've modified it with `--lang=c`.

libsel4/tools/syscall_stub_gen.py
wsipak: As for now, I've renamed this to `"-c", "--conf-file", dest="conf_file"`.
The short option, which is used in `libsel4/CMakeLists.txt` is the same, so there was no need to change it.

libsel4/tools/syscall_stub_gen.py
wsipak: Yes, now we have `GeneratorProvider` class.
It stores information about available generators, provides a list of their names (for `choices`), and allows retrieving a generator using its name. Additionally there's a custom `Exception`, and in case wrong generator name was given, it prints available generators.

libsel4/tools/rustgenerator.py
wsipak: Yes, `double_word` was set as true when `(size_bits == wordsize*2)`, so I've moved it to `__init__` function of `Type`, and removed this from arguments of the function.

libsel4/tools/generator.py
```diff
@@ -0,0 +1,612 @@
...
+import itertools
+import operator
+import xml.dom.minidom
+
+
+# Maximum number of words that will be in a message.
+MAX_MESSAGE_LENGTH = 64
+
+# Number of bits in a standard word
+WORD_SIZE_BITS_ARCH = {
```
axel-h, line 17: Looking at this I wonder if it might scale better if there is an "arch object" that can tell the bit size, the number of message registers and maybe more.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -37,1012 +37,29 @@
 #     XML method descriptions are added.
 #
 
-import operator
-import itertools
-import xml.dom.minidom
 from argparse import ArgumentParser
 import sys
-from functools import reduce
+from generator import WORD_SIZE_BITS_ARCH
```
axel-h, line 42: Do we need this import here?

libsel4/tools/generator.py
```diff
@@ -0,0 +1,612 @@
...
+    def construction(self, expr, param):
+        raise NotImplementedError()
+
+    @staticmethod
+    def generate_param_list(input_params, output_params):
+        raise NotImplementedError()
+
+    def configure(self, arch, wordsize, xml_files, buffer, mcs):
+        # Ensure architecture looks sane.
+        if arch not in WORD_SIZE_BITS_ARCH.keys():
```
axel-h, line 301: This would look up the `arch` object then. Choices for `wordsize` and `mcs` might be something that the architecture has to support then or this fails.

libsel4/tools/generator.py
```diff
@@ -0,0 +1,612 @@
...
+        raise NotImplementedError()
+
+    def construction(self, expr, param):
+        raise NotImplementedError()
+
+    @staticmethod
+    def generate_param_list(input_params, output_params):
+        raise NotImplementedError()
+
+    def configure(self, arch, wordsize, xml_files, buffer, mcs):
```
axel-h, line 299: Seems `buffer`is a boolean, so we could rename this here to `use_only_ipc_buffer` to avoid confusion.

libsel4/tools/generator.py
```diff
@@ -0,0 +1,612 @@
...
+        """returns str with return type of a function
+        """
+        # TODO document
+        raise NotImplementedError()
+
+    def generate_stub(self, interface_name, method_name, method_id, input_params, output_params, structs, comment_lines):
+        if self.use_only_ipc_buffer:
+            num_mrs = 0
+        else:
+            if self.mcs and "%s-mcs" % self.arch in MESSAGE_REGISTERS_FOR_ARCH:
```
axel-h, line 488: This might also get simpler with an "arch object" that has a function to get number of registers depending on an mcs parameter.

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+
+    def _gen_file_header(self):
+        # nothing needed for Rust
+        return
+
+    def _gen_file_footer(self):
+        # nothing needed for Rust
+        return
+
+    def _gen_conditional_compilation(self, condition) -> str:
```
axel-h, line 335: Seems the follow up taks here could be making the xml a bit more C preprocessor agostic. It would just give a commpa-separated list of `[!]CONFIG_xxx` tokens. We only have "A and B" conditions. If `CONFIG_MAX_NUM_NODES > 1` is replaced by `CONFIG_SMP_SUPPORT` then also, parsing should be simple in each generator to create the Rust/C specific line.

axel-h: OK, seems this conditinal stuff is a rabbit hole. In https://github.com/axel-h/seL4/tree/patch-axel-74 I see that changing to `CONFIG_SMP_SUPPORT` causes some failures. And there are more script that use the xml's `conditional` statement, so using a C-preprocessors agnostic languate requires changing them also. Seems simpler to ensure this uses a common syntax everywhere and then build a small parser for the Rust generator that rejects everything it does not understand.

libsel4/tools/generator.py
wsipak: Yes, right now the information about architecture is handled differently, there's `Architecture` class for that.

libsel4/tools/syscall_stub_gen.py
wsipak: Ditto as above, I've created `Architecture` class aiming to solve this.

libsel4/tools/generator.py
wsipak: Yes, it's boolean now, but no longer in `configure`.
`configure` was called just after an empty init, so I decided to put this inside `__init__`, and now the Generator objects are created in proper state, without the need to configure them.

wsipak:
Hello @axel-h  thank you so much for your feedback!
The things I've changed are:
1. `configure` function of Generator wasn't really needed, these things are now done in `__init__`.
2. Generators now store an `Architecture` object. It's supposed to not only provide the name of the architecture, but also to handle things You mentioned in the comments.
3. The functions `init_data_types` and `init_arch_types` have the same body in both backends, they seem redundant.
I've implemented these in the generic `Generator`, which gets language-specific types as parameters from a backend here:
https://github.com/antmicro/seL4/blob/multilang-stub-gen/libsel4/tools/generator.py#L370
https://github.com/antmicro/seL4/blob/multilang-stub-gen/libsel4/tools/generator_c.py#L447
At the same time I'm thinking that maybe they should be treated separately.
These things seem to be language-specific, but the same for Rust and C, so I'm not sure how to handle this.
4. I've done refactoring the improve general readability and replaced some `TODO document` with docstrings.

libsel4/tools/generator_c.py
wsipak: This seems to be problematic. We've got `basic_types.h` implemented here:
https://github.com/seL4/seL4/blob/master/include/basic_types.h

`libsel4` doesn't include anything from the global kernel directory, but rather things from inside `libsel4` directory, and the paths for includes can be viewed here:
https://github.com/seL4/seL4/blob/master/libsel4/CMakeLists.txt#L71-L78

I expect that this is by design, and placing any include from `../` wouldn't be welcome inside `libsel4/CMakeLists.txt`
However, just to test if things would work, I added the missing path, and then path to every file that was reported in errors.
I ended up having these:
```
    "${CMAKE_CURRENT_SOURCE_DIR}/../include"
    "${CMAKE_CURRENT_SOURCE_DIR}/../include/${KernelWordSize}"
    "${CMAKE_CURRENT_SOURCE_DIR}/../include/arch/${KernelArch}"
    "${CMAKE_CURRENT_SOURCE_DIR}/../include/arch/${KernelArch}/arch/${KernelWordSize}"
```
It won't compile due to some unknown type names and conflicting types.

I think we would benefit from having the macro available from inside `libsel4`, but I expect that the solution would actually require changes in the header files, rather than including files from `../`, and bit a bit more complex than just an addition to this PR.

axel-h: Yes, this needs to be improved in the headers. I will create a PR for this.

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+        64: "u64"
+    }
+
+    class Type(object):
+        """
+        This class represents a Rust type (such as an 'u32', structure or
+        pointer.
+        """
+
+        def __init__(self, name, size_bits, wordsize, *, native_size_bits=None):
```
axel-h, line 43: Do we really need the `*` here?

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+                return "{1}.wrapping_shr({2}) as {0}".format(Generator_Rust.TYPES[self.size_bits], var_name,
+                                                             word_size)
+
+    class BitFieldType(Type):
+        """
+        A special 'struct' generated by the bitfield generator
+        """
+
+        def __init__(self, name, size_bits, wordsize):
+            Generator_Rust.Type.__init__(self, name, size_bits, wordsize)
```
axel-h, line 128: Pin Python3 we can use the generic `super().__init__(self, name, size_bits, wordsize)`

libsel4/tools/generator.py
axel-h: I'm a bit lost now. What I asked for is rename `buffer` in the parameter list to `use_only_ipc_buffer`. If `configure()` is gone now, then we can delete this function here also - or am I missing something?

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+        def __init__(self, base_type, wordsize, mutable):
+            Generator_Rust.Type.__init__(self, base_type.name, wordsize, wordsize)
+            self.base_type = base_type
+            self.mutable = mutable
+
+        def render_parameter_name(self, name):
+            if name == "type":
+                name = "type_"
+            if self.mutable:
+                return "%s: *mut %s" % (name, Generator_Rust.translate_type(self.name))
```
axel-h, line 169: Could merge this and avoid some redundancy (it's a bit less readable then, though):
```
                return "%s: *%s %s" % (
                    "type_" if name == "type" else name, 
                    "mut" if self.mutable else "const", 
                    Generator_Rust.translate_type(self.name))
```

If the constructor of the `Type` class does a `self.generator = Generator_Rust` in its constructor, then `Type` and all derived classes can use `self.generator.some_static_method()`. This avoids having absolute class references and we can do this here:
```
                return "%s: *%s %s" % (
                    "type_" if name == "type" else name, 
                    "mut" if self.mutable else "const", 
                    self.generator.translate_type(self.name))
```

libsel4/tools/generator_c.py
```diff
@@ -0,0 +1,584 @@
...
+            return self.size_bits > self.wordsize and not self.double_word
+
+        def render_parameter_name(self, name):
+            """
+            Return a string of C code that would be used in a function
+            parameter declaration.
+            """
+            return "%s %s" % (self.name, name)
+
+        def pointer(self):
```
axel-h, line 78: maybe call this `asPointerType()`

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+        def render_parameter_name(self, name):
+            """
+            Return a string of Rust code that would be used in a function
+            parameter declaration.
+            """
+            if name == "type":
+                name = "type_"
+            return "%s: %s" % (name, Generator_Rust.translate_type(self.name))
+
+        def pointer(self, mutable):
```
axel-h, line 73: Maybe it make es code a bit easier to understand if there are two functions here
- `asMutablePointerType()`
- `asImmutablePointerType()`

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+        # legacy alias
+        arch_types["arm_hyp"] = arch_types["aarch32"]
+        return arch_types
+
+    @staticmethod
+    def generate_param_list(input_params, output_params):
+        # Generate parameters
+        params = []
+        for param in input_params:
+            if not param.type.pass_by_reference():
```
axel-h: Could reduce the redundancy a bit:
```
for param in input_params:
    t = param.type
    if t.pass_by_reference(): 
        t = t.asConstPointerType()
    params.append(t.render_parameter_name(param.name))

for param in output_params:
    t = param.type
    if t.pass_by_reference(): 
        t = t.asMutablePointerType()
        params.append(t.render_parameter_name(param.name))

```

libsel4/tools/generator_rust.py
```diff
@@ -0,0 +1,604 @@
...
+        result = []
+        source_words = {}
+        for i in range(MAX_MESSAGE_LENGTH):
+            if i < num_mrs:
+                source_words["w%d" % i] = "mr%d" % i
+            else:
+                source_words["w%d" % i] = "seL4_GetMR(%d)" % i
+        unmashalled_params = self.generate_unmarshal_expressions(output_params)
+        for (param, words) in unmashalled_params:
+            param.name = Generator_Rust.translate_expr(param.name)
```
axel-h, line 341:
```suggestion
            param.name = self.translate_expr(param.name)
```

kent-mcleod:
Hi @kgugala and @wsipak, thanks for opening this PR.

[Here is the mailing list discussion][1] that this PR refers to.

There's already been a lot of implementation feedback from @axel-h, but before there's continued iteration on implementation detail feedback, it's important that we establish that we agree on the approach.

My understanding based on that discussion was that the goal is to upstream a rust-specific copy of the `syscall_stub_gen.py` script for generating rust language seL4 syscall stubs so that it could be kept up-to-date with any changes to the seL4 syscall API.  The discussion also mentioned that the C support could be left unchanged.

The current approach in this PR redesigns the `syscall_stub_gen.py` to support multiple languages with an expectation that future languages can be added via implementing a language specific python Generator class.  This approach is a more substantial change than I was expecting for adding rust support and would require more discussion about the design of the Generator class before I could see it getting merged. For example, this class wouldn't work for the [CakeML invocation bindings][3] which spread over multiple files and have both C and CakeML components.

It should also be possible for the rust support to be added via an additional script `rust_syscall_stub_gen.py` that reuses the xml portions of `syscall_stub_gen.py` to translate the XML files into the python representation and then used new additional code to generate the rust language bindings.  This would be a much simpler change to review and get merged because it's only adding functionality for Rust support and is how the XML invocation definitions are already reused by other scripts: [the manual][2], [CakeML][3].

It essentially comes down to whether this PR is trying to add support for one new language or for many new languages.
Instead of needing to design a new per-language interface that can satisfy the unknown constraints of future languages that may be added, implementing the additional Rust support via additional code only is easier to review and maintain as the implementation can be specialized for the language specific constraints.

Sorry I wasn't able to comment any sooner - I'm still on holiday leave until 10 Jan. 

[1]: https://lists.sel4.systems/hyperkitty/list/devel@sel4.systems/thread/AMUW4JXGTX5RU2PEHVHEC5CW3XFBBD4V/
[2]: https://github.com/seL4/seL4/blob/master/manual/tools/gen_invocations.py#L63-L67
[3]: https://github.com/seL4/cakeml_libs/blob/libsel4_cakeml/libsel4_cakeml/gen_cakeml_invocations.py#L346-L355

wsipak:
We definitely need to agree on a specific approach to handle this. Providing a rust-specific copy of `syscall_stub_gen.py` is of course possible. However, most of it would be in fact a copy of the C-specific script, and while it would be easier to bring in, the maintenance of it could become problematic.

We're aiming to provide a structure that will allow adding generators for other languages in a convenient way.

When it comes to the generator for the C language, the code is indeed changed here. However, please notice that most parts of it are moved into a class, and with a four spaces indent GitHub won't show these parts as the same code. Additionally, the code that actually generated strings of code and was placed in a single function (like in `generate_stub`) has been divided and moved into functions which generate things like function header, variable declarations etc. The output of the C code generator remains unchanged.

The XML parts contain things that are C-specific. For the purpose of making this more generic, we've replaced the part that created comments in `parse_xml_file`, so that it creates only the content of the comment, which is wrapped with `/* */` only later on. Another thing is conditions for conditional compilation. This would need a deeper rework in order to make it generic and I'd opt for leaving it for discussion.

In the matter of the `Generator` class, our goal is to provide a generic API that will work with generators for any language in general. It has a `generate` method which is supposed to generate the whole file in a generic way, using overridden language-specific methods. The functions which are language-specific and shall be overridden raise `NotImplementedError`.
Every function with name `_gen_*` is dedicated to generating a specific piece of code, say function header or a comment.

When it comes to integrating the script with the other parts of seL4 system, like CakeML and the manual: The first problem we've encountered is that the script is both executed and imported. It wasn’t an issue until we wanted to import from inside the script. A viable solution for this is to create a package instead.

As we're at the stage of discussion about what we really want here, We've provided a solution that makes minimal amounts of changes in other parts of seL4, and some backward-compatible code to make it work. Eventually It'd probably be better to adjust the other parts to the new generator, but for the purpose of discussion and showing what's needed, here's the part for backwards compatibility:
https://github.com/antmicro/seL4/blob/multilang-stub-gen/libsel4/tools/syscall_stub_gen/__init__.py#L13-L45

Patch needed for CakeML:
https://github.com/antmicro/seL4/blob/multilang-stub-gen/libsel4/tools/cakeml_libs.patch

ratmice:
> please notice that most parts of it are moved into a class, and with a four spaces indent GitHub won't show these parts as the same code.

In these cases it may be worth doing the changes without any re-indentation, and then doing it in a follow-up commit,
It can make review a lot easier.

kent-mcleod:
Sorry I was unable to reply earlier, I needed to find sufficient time to give this proper attention.

> We definitely need to agree on a specific approach to handle this. Providing a rust-specific copy of syscall_stub_gen.py is of course possible. However, most of it would be in fact a copy of the C-specific script, and while it would be easier to bring in, the maintenance of it could become problematic.

Providing the rust support in a separate script should make maintenance (and reviewing) easier.  Keeping the scripts separate makes it possible to hold the separate implementations to different standards of design and implementation quality. As a reviewer it is easier to quickly check whether the changes will have any effect on the C library without having to look through the code and as a developer making implementation/design changes to the rust code it is easier to quickly check if the changes will have any breaking effect on the C code. 

Of course it would be best if this didn't put too much additional burden on implementing the rust code generation. I did a quick experiment to see what a separate rust script would look like to give a better idea of what I was suggesting above: https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d.  You can test it with the following command: `./syscall_stub_gen_rs.py -a aarch32 ../sel4_arch_include/aarch32/interfaces/sel4arch.xml  ../arch_include/arm/interfaces/sel4arch.xml ../include/interfaces/sel4.xml  | less` which will generate a rust file for aarch32.

> When it comes to the generator for the C language, the code is indeed changed here. However, please notice that most parts of it are moved into a class, and with a four spaces indent GitHub won't show these parts as the same code. Additionally, the code that actually generated strings of code and was placed in a single function (like in generate_stub) has been divided and moved into functions which generate things like function header, variable declarations etc. The output of the C code generator remains unchanged.

I can see that you've intended to extend the generator so that it still outputs the same C source, but they way that this source is generated has changed a lot. It's hard to assess that these changes will produce the same C output for all of the different inputs it can be given without resorting to some level of manual inspection. Also considering that the libsel4 C library is a fairly core component of many seL4 user level applications, reviewing these changes has quite a lot of downside risk associated with them.  This would then be incurred across future changes also whenever rust-specific changes require making changes to any new shared code.

> We're aiming to provide a structure that will allow adding generators for other languages in a convenient way.

Currently adding support for other languages is hampered by many interface definitions only being defined in C header files instead of in XML.  Because of this, the existing generators are only generating a subset of the kernel interface.  An example of this is how both the Rust and C language implementations have type dependencies that assume external definitions that the respective libraries end up implementing. Different languages are going to need different subsets generated by their respective generators and likely won't end up fitting nicely into the single file structure defined by C and Rust's similar "structured-programming" structure.  Solving these problems ends up making the scope of these proposed changes a lot larger and would hold-up the rust support from being merged.

> The XML parts contain things that are C-specific. For the purpose of making this more generic, we've replaced the part that created comments in parse_xml_file, so that it creates only the content of the comment, which is wrapped with /* */ only later on. Another thing is conditions for conditional compilation. This would need a deeper rework in order to make it generic and I'd opt for leaving it for discussion.

I agree that there are places where the XML definitions are overly C specific.  The Class methods: `render_parameter_name`, 
`c_expression`, and `double_word_expression` are an example of where this doesn't end up being a big issue as the Rust script is able to provide its own implementations via a small ad-hoc dynamic dispatch approach and neither the C or Rust versions end up overly complex because of it.

I agree that the way that conditional compilation constraints are expressed in C preprocessor syntax within the XML could be changed to make it easier to generate in other languages but the conditions are evaluated according to CPP definitions provided by `autoconf.h` so these would also need to be provided in a different format.  It would be possible to call out to the CPP to evaluate the expressions during generation time, and so maybe that would be a suitable intermediate step.

Does what I'm suggesting make more sense with the example (https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d) as reference ?
Is the version of the seL4-sys library that uses the output rust file publicly available? This would make it easier to see how various things integrate.

libsel4/tools/syscall_stub_gen/generator_rust.py
```diff
@@ -0,0 +1,496 @@
...
+
+    def _gen_unmarshal_regs_into_ipc(self, num_mrs, returning_struct) -> str:
+        result = []
+        result.append("\tif error != seL4_Error::seL4_NoError {")
+        for i in range(num_mrs):
+            result.append("\t\tseL4_SetMR(%d, mr%d);" % (i, i))
+        if returning_struct:
+            result.append("\t\treturn result;")
+        else:
+            result.append("\t\terror.into()")
```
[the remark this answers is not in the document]
kent-mcleod: A separate meta comment: An example of how I personally find this structure more complex than a single script is that the scope of the function being generated no longer lines up with the scope of the functions doing the generation. This makes it hard to tell what will happen when things like return statements are encountered as they act more like `goto;`s within the meta scope.

kent-mcleod (commented):
Thanks for all of the work you've all done on this so far. It does go a long way to providing the components for implementing a more stable and maintainable user level rust library (with richer types too!). My feedback is essentially all relating to non-functional requirements that will make maintenance and review more manageable.

sleffler:
> Sorry I was unable to reply earlier, I needed to find sufficient time to give this proper attention.
> 
> > We definitely need to agree on a specific approach to handle this. Providing a rust-specific copy of syscall_stub_gen.py is of course possible. However, most of it would be in fact a copy of the C-specific script, and while it would be easier to bring in, the maintenance of it could become problematic.
> 
> Providing the rust support in a separate script should make maintenance (and reviewing) easier. Keeping the scripts separate makes it possible to hold the separate implementations to different standards of design and implementation quality. As a reviewer it is easier to quickly check whether the changes will have any effect on the C library without having to look through the code and as a developer making implementation/design changes to the rust code it is easier to quickly check if the changes will have any breaking effect on the C code.
> 
> Of course it would be best if this didn't put too much additional burden on implementing the rust code generation. I did a quick experiment to see what a separate rust script would look like to give a better idea of what I was suggesting above: [11315d4](https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d). You can test it with the following command: `./syscall_stub_gen_rs.py -a aarch32 ../sel4_arch_include/aarch32/interfaces/sel4arch.xml ../arch_include/arm/interfaces/sel4arch.xml ../include/interfaces/sel4.xml | less` which will generate a rust file for aarch32.

The initial python script that generated rust bindings was separate from the C version. It was out of date and had diverged and (IIRC) grown different dependencies which made updating painful. I re-based the code to the kernel script to simplify upstream integration but used a brute-force approach. Karol looked at that version and proposed overhauling the python to use modules. I prefer the modules to separate scripts but clearly both approaches can work. My experience makes me think Karol's approach is better but both require proper unit/regression testing to avoid bitrot--that is the purpose of the sel4test work that crafts C shims to the rust support to enable running sel4test (otherwise) unchanged.

> 
> > When it comes to the generator for the C language, the code is indeed changed here. However, please notice that most parts of it are moved into a class, and with a four spaces indent GitHub won't show these parts as the same code. Additionally, the code that actually generated strings of code and was placed in a single function (like in generate_stub) has been divided and moved into functions which generate things like function header, variable declarations etc. The output of the C code generator remains unchanged.
> 
> I can see that you've intended to extend the generator so that it still outputs the same C source, but they way that this source is generated has changed a lot. It's hard to assess that these changes will produce the same C output for all of the different inputs it can be given without resorting to some level of manual inspection. Also considering that the libsel4 C library is a fairly core component of many seL4 user level applications, reviewing these changes has quite a lot of downside risk associated with them. This would then be incurred across future changes also whenever rust-specific changes require making changes to any new shared code.

That you're suggesting inspecting the code says to me more/better testing is needed. As I mentioned above we use sel4test as a way to check the rust code. I also have a rust-based rootserver that exercises interfaces not used by sel4test.

> 
> > We're aiming to provide a structure that will allow adding generators for other languages in a convenient way.
> 
> Currently adding support for other languages is hampered by many interface definitions only being defined in C header files instead of in XML. Because of this, the existing generators are only generating a subset of the kernel interface. An example of this is how both the Rust and C language implementations have type dependencies that assume external definitions that the respective libraries end up implementing. Different languages are going to need different subsets generated by their respective generators and likely won't end up fitting nicely into the single file structure defined by C and Rust's similar "structured-programming" structure. Solving these problems ends up making the scope of these proposed changes a lot larger and would hold-up the rust support from being merged.

This is an initial version of sel4-sys; there are follow-on changes to to enable/fix MCS and my rust-based rootserver (+various bug fixes).

> 
> > The XML parts contain things that are C-specific. For the purpose of making this more generic, we've replaced the part that created comments in parse_xml_file, so that it creates only the content of the comment, which is wrapped with /* */ only later on. Another thing is conditions for conditional compilation. This would need a deeper rework in order to make it generic and I'd opt for leaving it for discussion.
> 
> I agree that there are places where the XML definitions are overly C specific. The Class methods: `render_parameter_name`, `c_expression`, and `double_word_expression` are an example of where this doesn't end up being a big issue as the Rust script is able to provide its own implementations via a small ad-hoc dynamic dispatch approach and neither the C or Rust versions end up overly complex because of it.
> 
> I agree that the way that conditional compilation constraints are expressed in C preprocessor syntax within the XML could be changed to make it easier to generate in other languages but the conditions are evaluated according to CPP definitions provided by `autoconf.h` so these would also need to be provided in a different format. It would be possible to call out to the CPP to evaluate the expressions during generation time, and so maybe that would be a suitable intermediate step.

There are many C dependencies in the seL4 interfaces that could be improved for portability. Chief among them is the use of C enum semantics.

> 
> Does what I'm suggesting make more sense with the example ([11315d4](https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d)) as reference ? Is the version of the seL4-sys library that uses the output rust file publicly available? This would make it easier to see how various things integrate.

We were waiting to resolve this PR so that code works out of the box. The best thing is to make sel4-sys + the sel4test changes available so you can see the end-to-end flow. The rootserver is harder to share atm 'cuz it's integrated with our devel environment.

#

kent-mcleod:
> I prefer the modules to separate scripts but clearly both approaches can work. My experience makes me think Karol's approach is better but both require proper unit/regression testing to avoid bitrot--that is the purpose of the sel4test work that crafts C shims to the rust support to enable running sel4test (otherwise) unchanged.

As we discussed in the developer call today, separate scripts are necessary for us to accept this PR at this stage.  https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d is an example of what we'd be happy to accept. This also keeps the door open for having separate scripts upstreamed for other language library bindings, potentially even scripts for different implementations within the same language with different designs.

> There are many C dependencies in the seL4 interfaces that could be improved for portability. Chief among them is the use of C enum semantics.

Longer term we do need to work to provide a more complete set of interface definitions in a portable format.

> That you're suggesting inspecting the code says to me more/better testing is needed. As I mentioned above we use sel4test as a way to check the rust code. I also have a rust-based rootserver that exercises interfaces not used by sel4test.

We'll always inspect code as it's changed for almost every seL4 repository. It's even more important when it's code that's expected to be present within systems running on a verified kernel.  That said, inspecting code isn't a substitute for adding/improving test cases and vice-versa.

wsipak:
Thank you for the commit with the suggested change.
Right now there's additional `syscall_stub_gen_rs.py` for Rust, with no other changes in any part of seL4.
Unfortunately we weren't able to reuse some parts of `syscall_stub_gen.py` that were imported from it with your suggested change, as for example `Type` classes need to implemented specifically for Rust. This makes the script a bit longer than in https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d .
As `init_data_types` and `init_arch_types` functions use the `Type` classes, we also need them reimplemented for Rust.

kent-mcleod:
Thanks for the updated changes. 

>  Unfortunately we weren't able to reuse some parts of syscall_stub_gen.py that were imported from it with your suggested change, as for example Type classes need to implemented specifically for Rust. This makes the script a bit longer than in 11315d4 .

Are you able to explain why you consider reuse isn't possible? I checked that the script without separate Rust definitions of the `Type` classes or `init_data_types` and `init_arch_types` is able to generate identical output as the one with the re-implemented type classes.  With some minor alterations to the smaller script (https://github.com/kent-mcleod/seL4/commit/5775851651ce01545aef9cd1a4cdfed9af24a0e9) it is able to generate the exact same output as https://github.com/seL4/seL4/commit/2703172175f6f2b0305861f1f3f675683e39db15 without requiring the `Type` classes to be reimplemented.  It's likely that we all agree that these type class definitions and implementation of `init_data_types` and `init_arch_types` are what would contribute to the bulk of the unwanted maintenance burden which is why it'd be preferred to not duplicate them.

libsel4/tools/syscall_stub_gen_rs.py
```diff
@@ -0,0 +1,776 @@
...
+    # Generate parameters
+    params = []
+    for param in input_params:
+        t = param.type
+        if t.pass_by_reference():
+            t = t.asPointerType(False)
+        params.append(t.render_parameter_name(param.name))
+    for param in output_params:
+        t = param.type
+        if t.pass_by_reference():
```
axel-h, line 325: make this an assert here?
```
assert t.pass_by_reference()
```

wsipak:
> Thanks for the updated changes.
> 
> > Unfortunately we weren't able to reuse some parts of syscall_stub_gen.py that were imported from it with your suggested change, as for example Type classes need to implemented specifically for Rust. This makes the script a bit longer than in [11315d4](https://github.com/seL4/seL4/commit/11315d4a1072158aa84ad29d8deb12a68d4c7e1d) .
> 
> Are you able to explain why you consider reuse isn't possible? I checked that the script without separate Rust definitions of the `Type` classes or `init_data_types` and `init_arch_types` is able to generate identical output as the one with the re-implemented type classes. With some minor alterations to the smaller script ([kent-mcleod@5775851](https://github.com/kent-mcleod/seL4/commit/5775851651ce01545aef9cd1a4cdfed9af24a0e9)) it is able to generate the exact same output as [2703172](https://github.com/seL4/seL4/commit/2703172175f6f2b0305861f1f3f675683e39db15) without requiring the `Type` classes to be reimplemented. It's likely that we all agree that these type class definitions and implementation of `init_data_types` and `init_arch_types` are what would contribute to the bulk of the unwanted maintenance burden which is why it'd be preferred to not duplicate them.

Yes, this indeed produces the same code for Rust. Now I can see that the methods of `Type` objects that generate C code aren't in fact called, but rather the fields are used in functions that generate Rust. Thank you!

libsel4/tools/syscall_stub_gen_rs.py
wsipak: When we have `if param.type.pass_by_reference():` the code is generated correctly.
If assert it here, the script will fail even though there's no need for that (We just need to skip unneeded arguments).
Here are some generated snippets to demonstrate what happens:

C code:
```
LIBSEL4_INLINE seL4_TCB_GetBreakpoint_t
seL4_TCB_GetBreakpoint(seL4_TCB _service, seL4_Uint16 bp_num)
{
	seL4_TCB_GetBreakpoint_t result;
	seL4_MessageInfo_t tag = seL4_MessageInfo_new(TCBGetBreakpoint, 0, 0, 1);
...
```

Rust generated as expected:
```
#[inline(always)]
pub unsafe fn seL4_TCB_GetBreakpoint(_service: seL4_TCB, bp_num: u16) -> seL4_TCB_GetBreakpoint
{
	let mut result: seL4_TCB_GetBreakpoint = ::core::mem::zeroed();
	let tag = seL4_MessageInfo::new(InvocationLabel::TCBGetBreakpoint as seL4_Word, 0, 0, 1);
...
```
Rust generated with output arguments that are `not pass_by_reference` added.
```
#[inline(always)]
pub unsafe fn seL4_TCB_GetBreakpoint(_service: seL4_TCB, bp_num: u16, vaddr: *mut seL4_Word, type_: *mut seL4_Word, size: *mut seL4_Word, rw: *mut seL4_Word, is_enabled: *mut u8) -> seL4_TCB_GetBreakpoint
{
    let mut result: seL4_TCB_GetBreakpoint = ::core::mem::zeroed();
    let tag = seL4_MessageInfo::new(InvocationLabel::TCBGetBreakpoint as seL4_Word, 0, 0, 1t);
```

Output arguments that are not `pass_by_reference` also appear in 4 other functions, and skipping them results in code that corresponds to the C code.
These lists of parameters are in fact obtained using `syscall_stub_gen.parse_xml_file`.

axel-h: Thanks very much to looking into this. Having a comment here would be helpful that dropping certain params is really intended (and no else-branch got lost somewhere).

wsipak: Sure, right now there's a comment informing that we're dropping these params intentionally.

kent-mcleod:
> Yes, this indeed produces the same code for Rust. Now I can see that the methods of `Type` objects that generate C code aren't in fact called, but rather the fields are used in functions that generate Rust. Thank you!

Ok great! From my perspective this is essentially ready to be merged.  2 more things:
- Can you amend the commit message to provide a body describing how this script is intended for a particular rust library that isn't part of this repository, possibly with information on where to find the library.  It might also be helpful to add that, although this script was derived from Robigalia's sel4-sys library, it's no longer compatible with that library due to some small interface changes.
- As this file was derived from Robigalia's sel4-sys, can you make sure that the correct copyright notices are present at the top of the file.

sleffler:
This is used by any external crate that defines seL4_Error & co.  Our
sel4-sys crate (an offshoot of the robigalia code) is such a crate. We're
working to make that code available.

ratmice:
I don't find that ideal, but the position that I find myself drifting towards is that if someone e.g. needs to modify the generated source, but cannot do so because of compatibility concerns with your crate, they aren't actually in any worse position than we were before this was added because they would still need to maintain their external own stub generator like before.

kent-mcleod:
> As this file was derived from Robigalia's sel4-sys, can you make sure that the correct copyright notices are present at the top of the file.

This should be `Copyright 2017, The Robigalia Developers`

> Can you amend the commit message to provide a body describing how this script is intended for a particular rust library that isn't part of this repository, possibly with information on where to find the library.

It would also be useful to add a similar explanation as a comment at the top of the file to avoid confusion. 

> Nice, I think it'd be good though to discuss the generated code though, perhaps I just don't understand the plan for it -- because it references modules and types which have no definition in the repository.

This PR is a result of a request on the mailing list that this script be upstreamed so that it doesn't get out of date as easily due to API changes to this repository.  Longer term plans for rust support require a fair amount of work to extend the XML interface definitions (or replace them with a different interface definition) so that its easier to have out-of-tree kernel interface libraries in other languages that are generated only from the XML interface. 


> they aren't actually in any worse position than we were before this was added because they would still need to maintain their external own stub generator like before.

They would also be able to make a request about upstreaming their script so that it would get updated at the same time as any changes to the kernel API.  Once there is better interface definition support, the scripts would likely go away again.

ratmice:
Thanks Kent, this alleviates my concenrns.  I just didn't want to get into a situation where we want to extend support to generate more rust support in tree (Like the InvocationLabels), but couldn't update the generated sources.

It sounds like this isn't going to actually be a problem then, and that I'm better off working towards that goal of out-of-tree consumption of the XML interfaces than trying to extend this current interim situation to the other generators so that it can be free standing.

wsipak:
In fact, a copyright line was present in the Robigalia project on GitLab (https://gitlab.com/robigalia/sel4-sys),
so I've added the same exact line here (`# Copyright 2017, Corey Richardson`).

Additionally, I've extended the commit body, and added a comment at the top of the file to explain what the script is for.

kent-mcleod:
> In fact, a copyright line was present in the Robigalia project on GitLab (https://gitlab.com/robigalia/sel4-sys), so I've added the same exact line here (`# Copyright 2017, Corey Richardson`).

I got in touch with the original author, and they confirmed it would be great if the name could be changed to Ember Arlynx to match their current details (https://gitlab.com/emberian).

> 
> Additionally, I've extended the commit body, and added a comment at the top of the file to explain what the script is for.

Thanks!

wsipak:
> > In fact, a copyright line was present in the Robigalia project on GitLab (https://gitlab.com/robigalia/sel4-sys), so I've added the same exact line here (`# Copyright 2017, Corey Richardson`).
> 
> I got in touch with the original author, and they confirmed it would be great if the name could be changed to Ember Arlynx to match their current details (https://gitlab.com/emberian).

Alright, done!

kent-mcleod (approved):

kent-mcleod:
Thanks for sticking with this until it got merged everyone.

axel-h:
I wonder is this script is broken now, see https://github.com/seL4/seL4/issues/1223
@kgugala @wsipak: are you using this script regularly now anywhere with the latest seL4 master brnach or was this just use once in the end?

PR #736, opened 2022-01-04
trivial: remove trailing whitespaces

axel-h:


axel-h:
Adding another commit that may explain the motivation and where this comes from. In the end it's all motivated by the discussion in https://github.com/seL4/seL4/pull/733

lsf37 (approved):
We don't usually do just style updates, but the xml is safe enough for conflict resolution, so I'm fine with merging this.

And I think making the condition syntax uniform for parsing is a good idea.

Issue #738, opened 2022-01-06
`SEL4_FORCE_LONG_ENUM` not C standard compatible


lsf37:
So I might be a bit late to the game, but this trick we're using to force the enum to be `long` does not seem to be legal C according to the standard. C99 and C11 say:

> 6.7.2.2 Enumeration specifiers
> 1)
> The expression that defines the value of an enumeration constant shall be an integer constant expression that has a value representable as an `int`.

Clause 3 later says:
> Each enumerated type shall be compatible with char, a signed integer type, or an unsigned integer type. The choice of type is implementation-defined but shall be capable of representing the values of all the members of the enumeration.

"an unsigned integer type" could be `long`, but the `int` in clause 1 is pretty clear. GCC and clang both seem to be doing this as we want, i.e. they satisfy the compile time asserts for the size of the enum type to be same as the size of `long`.

This came up because the verification C parser does not handle the `SEL4_FORCE_LONG_ENUM` in  #724 as expected (it forces `int` like the standard) and therefore later fails the proof in a pointer arithmetic expression down the road.

I can look into making the C parser more relaxed, but I wanted to put this out here for discussion first, because it doesn't seem to be officially supported by the standard, so could break later.

As long as binary verification runs, we will catch these problems, and even the compile time assert should already alert us to problems if they occur in future compiler versions, so I'm tending to try and make the C parser more relaxed. I'm not clear yet on how much work that is, because it will involve a new value analysis.

lsf37:
For context, this is what such an enum declaration looks like after preprocessing:
```c
typedef enum {
    SEL4_BOOTINFO_HEADER_PADDING = 0,
    SEL4_BOOTINFO_HEADER_X86_VBE = 1,
    SEL4_BOOTINFO_HEADER_X86_MBMMAP = 2,
    SEL4_BOOTINFO_HEADER_X86_ACPI_RSDP = 3,
    SEL4_BOOTINFO_HEADER_X86_FRAMEBUFFER = 4,
    SEL4_BOOTINFO_HEADER_X86_TSC_FREQ = 5,
    SEL4_BOOTINFO_HEADER_FDT = 6,
    SEL4_BOOTINFO_HEADER_NUM,
    _enum_pad_seL4_BootInfoID = ((1ULL << ((sizeof(long)*8) - 1)) - 1)
} seL4_BootInfoID;
```

Xaphiosis:
[...]

PR #724, opened 2021-12-27
use enum for SEL4_BOOTINFO_HEADER_xxx

axel-h:
- use enum for `SEL4_BOOTINFO_HEADER_xxx` to keep some semantic information
- Also add more comments.

PR #739, opened 2022-01-07
keep seL4_Word for seL4_BootInfoID

lsf37:
This reverts a small part of 1596aa5857 (#724) which unexpectedly broke the binary verification frontend, because the verification tool chain forces the enum to `int` instead of of `long`.

See also seL4 issue #738 -- when that issue is solved, we may be able to go back to the more specific type.

### DIFF
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -96,8 +96,8 @@ typedef enum {
 
 /* Common header for all additional bootinfo chunks to describe the chunk. */
 typedef struct seL4_BootInfoHeader {
-    seL4_BootInfoID  id;  /* identifier of the following blob */
-    seL4_Word        len; /* length of the chunk, including this header */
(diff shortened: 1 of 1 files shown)


axel-h (approved):

axel-h:
How did you find find this, did it slip because `proof-test` was not running on the PR?
You may want to fix the typo in commit comment, one "of" is enough in : "instead of of `long`."

lsf37:
> How did you find find this, 

The main regression test on the proof side broke

> did it slip because `proof-test` was not running on the PR?

Yes, I was too confident merging this without running the proof test. The type mismatch is only a problem in certain situations, so it doesn't always break the proofs when we do this and apparently all the other uses of that macro are benign.

> You may want to fix the typo in commit comment, one "of" is enough in : "instead of of `long`."

Ah, yes, thanks for spotting that. Will do.

lsf37:
(adjusted commit message)

PR #742, opened 2022-01-11
smmu+manual: typo and small sentence tune

Xaphiosis:
Came across these looking at SMMU. Not 100% on "and conducting a
TLB invalidation", but this is my guess. Cross-check?

It's possible that "invalidates the TLB" is wrong here, and "conducts a TLB invalidation" is what we want.


lsf37 (approved):
:+1:

PR #744, opened 2022-01-14
VM fault fastpath

alwin-joshy:
Implementation of the VM fault fastpath with some stubs implemented for other types of faults. Up to 62% faster for faulter -> handler direction and up to 67% faster for handler -> faulter direction. 

Includes

- Changed logic in c_traps files for AARCH64, x86_64, RISCV to redirect to fastpath.
- Newly implemented vm fault fastpath.
- Modified replyRecv fastpath to enable fastpathing of fault replies. 

Tests/Benchmarks

- seL4Test suite ran on AARCH64, AARCH32, x86, RISCV - All pass
- seL4bench suite + newly added benchmarks ran on AARCH64, AARCH32, x86. Could not run fault benchmarks on RISCV for some reason. 

EDIT: I said 62% and 67% faster but what I meant was 62% and 67% fewer cycles.


src/fastpath/fastpath.c
```diff
@@ -15,6 +15,289 @@
...
 #include <benchmark/benchmark_utilisation.h>
 
+
+#ifdef CONFIG_ARCH_ARM
+static inline
+FORCE_INLINE
+#endif
+void NORETURN fastpath_vm_fault(vm_fault_type_t type)
+{
+    cap_t handler_cap;
```
pingerino, line 25: nit: this declare before use is an old style from when we compiled the kernel with c90, it's not required anymore, and actually creates extra lines of proof. So new code doesn't need to use it.

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,289 @@
...
+    }
+#endif
+
+    /* Ensure the original caller is in the current domain and can be scheduled directly. */
+    if (unlikely(dest->tcbDomain != ksCurDomain && 0 < maxDom)) {
+        vm_fault_slowpath(type);
+    }
+
+#ifdef CONFIG_KERNEL_MCS
+    if (unlikely(dest->tcbSchedContext != NULL)) {
```
pingerino, line 132: So we're only fastpathing the case where the scheduling context is donated? does that make sense in the context of a VM fault fastpath?

alwin-joshy: I think when I discussed this with Gernot and Kevin, we agreed that passive servers make sense for a lot of fault handlers. Additionally, I based a lot of this logic off the seL4_Call fastpath, which also only supports calling to passive servers at the moment. I can see that it would be ideal to be able to support active servers as fault handlers too, but I would need to have a better understanding of why this is not an option for seL4_Call and what could be done to make it work for this fastpath.

src/fastpath/fastpath.c
alwin-joshy: I see, I started with a declare as used style but changed it to be more similar to the call and replyrecv fastpaths. I will change it soon.

include/arch/arm/arch/fastpath/fastpath.h
```diff
@@ -18,6 +18,15 @@ static inline
...
 
+
+void vm_fault_slowpath(vm_fault_type_t type)
+NORETURN;
+
+static inline
+void fastpath_vm_fault(vm_fault_type_t type)
+NORETURN;
+
+
```
lsf37, line 29: these should be guarded by `EXCEPTION_FASTPATH`

(also no need for additional blank lines, one is enough)

src/arch/arm/c_traps.c
```diff
@@ -62,29 +62,54 @@ void VISIBLE NORETURN c_handle_enfp(void)
 }
 #endif /* CONFIG_HAVE_FPU */
 
-static inline void NORETURN c_handle_vm_fault(vm_fault_type_t type)
+void NORETURN vm_fault_slowpath(vm_fault_type_t type)
```
lsf37, line 65: this function should still be inline -- the code should not change when `EXCEPTION_FASTPATH` is not enabled.

src/arch/arm/c_traps.c
```diff
@@ -127,8 +152,8 @@ void NORETURN slowpath(syscall_t syscall)
 void VISIBLE c_handle_syscall(word_t cptr, word_t msgInfo, syscall_t syscall)
 {
     NODE_LOCK_SYS;
-
```
lsf37, line 130: please avoid unrelated whitespace changes

src/arch/arm/c_traps.c
```diff
@@ -127,8 +152,8 @@ void NORETURN slowpath(syscall_t syscall)
...
     c_entry_hook();
+
```
lsf37, line 156: please avoid unrelated whitespace changes

src/arch/arm/c_traps.c
```diff
@@ -194,4 +219,4 @@ VISIBLE NORETURN void c_handle_vcpu_fault(word_t hsr)
     restore_user_context();
     UNREACHABLE();
 }
-#endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
+#endif // CONFIG_ARM_HYPERVISOR_SUPPORT
```
lsf37, line 222: please avoid style changes

src/fastpath/fastpath.c
pingerino: it doesn't make sense for seL4_Call as you need to pull in nearly the whole scheduler if you are switching SC's, as at minimum you need to reprogram the timer and check if you need to wake other tasks. There are subsets of that that could be pulled in, but you'd need to choose which scheduling scenario you want to fastpath. 

Also, the less we pull in the easier it is for verification to verify.

pingerino (commented):
Have you confirmed that this behaviour is indeed a complete subset (valid traversal) of a slow path? Otherwise it's impossible to verify, as I learned when modifying fastpaths myself.

src/fastpath/fastpath.c
```diff
@@ -472,19 +762,82 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
+#else
     /* Replies don't have a badge. */
     badge = 0;
 
     fastpath_copy_mrs(length, NODE_STATE(ksCurThread), caller);
 
     /* Dest thread is set Running, but not queued. */
-    thread_state_ptr_set_tsType_np(&caller->tcbState,
-                                   ThreadState_Running);
+    thread_state_ptr_set_tsType_np(&caller->tcbState, ThreadState_Running);
```
lsf37, line 835: please revert style changes, there should be zero code diff when the new fastpath is not configured.

pingerino (commented):
Thanks for the comment on what you tested - I notice it doesn't mention running a VMM. sel4test doesn't really test the virtualisation paths well, so you should try running linux on top at least.

src/fastpath/fastpath.c
```diff
@@ -324,9 +607,16 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
     /* Check that the caller has not faulted, in which case a fault
        reply is generated instead. */
     fault_type = seL4_Fault_get_seL4_FaultType(caller->tcbFault);
+
+#ifndef CONFIG_EXCEPTION_FASTPATH
     if (unlikely(fault_type != seL4_Fault_NullFault)) {
         slowpath(SysReplyRecv);
     }
+#else // TODO: Will be able to get rid of this once all exceptions are accounted for
```
lsf37, line 615: Please use issues not comments for todo lists.

src/fastpath/fastpath.c
```diff
@@ -472,19 +762,82 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
+#endif
+        default: {
+            restart = 1;
+        }
+        }
+
+        /* TODO: This works here right now because vm faults always restart - the alternative is that the thread becomes inactive,
+         * which would either be handled in the fastpath or redirected into the slowpath (preferred) - either way, this check
+         * will need to be moved above the point of no return to be handled to ensure we don't try and force a switch to an
+         * inactive thread. */
```
lsf37, line 798: This does not sound like a TODO you are planning to resolve. Adding an explanation here makes sense (esp on what is implicitly assumed), but it would be good to be slightly clearer that we expect the current implementation to work as is, and that more work will only be needed if we do not have `restart` set.

lsf37:
When you say

> Could not run fault benchmarks on RISCV for some reason.

do you mean there weren't any or they were failing?

lsf37:
The `rv32` simulation test is failing on `VMFault`, which I assume is related.

In general, what kind of testing has been done so far? Passing sel4test is good, but it does not say much about the fast path other than that it works in a few specific cases. Any thoughts an how to substantiate that the behaviour is the same as the slow path for all cases? Even if it works now, as long as there are no proofs, we need something that alerts us when the behaviour drifts from the slowpath (e.g. because there are slowpath changes an people have forgotten to update the fastpath accordingly).

In your tests, which configurations did you use? (SMP, MCS, HYP, verification/release/debug). You can switch on the hardware test suite on the PR by adding the `hw-test` label, but it will not configure the exception fastpath by default. 

Before we can merge this, we need to integrate it into CI. The main question for that is for which configurations we want the exception fastpath on. I don't think we want to double all tests (that would be easy enough to do, but it would cost a lot or test resources). It also can't be on for verification builds yet (the dependency for that still needs to be encoded in the `cmake` config). Any proposals for that?

To test the VMM with this, have a look at the setup running e.g. for [camkes-vm-examples](https://github.com/seL4/camkes-vm-examples/actions/runs/1525285140).

alwin-joshy:
> 

Enabling the fault benchmarks on RISCV didn't do anything - the fault benchmarks were just always skipped for some reason.

lsf37:
> Enabling the fault benchmarks on RISCV didn't do anything - the fault benchmarks were just always skipped for some reason.

Looking at he mainline sel4bench output it seems like those are not enabled at all for RISCV yet. Possibly because nobody has made them work yet. It'd be good to add them in general, but I'm not sure how much work that would be.

lsf37:
For the DCO and gitlint test failures, you need to add a [`Signed-off-by` line](https://docs.sel4.systems/processes/contributing.html) and provide [body messages](https://docs.sel4.systems/processes/git-conventions.html) (Ok to leave out for commits you are planning to squash).

src/fastpath/fastpath.c
```diff
@@ -254,7 +254,7 @@ void NORETURN fastpath_vm_fault(vm_fault_type_t type)
     }
 #endif
 
-#ifdef CONFIG_ARCH_RISCV64
+#ifdef CONFIG_ARCH_RISCV
     uint64_t addr;
```
axel-h, line 258: Is this is used for 32 and 64 bit now, using `word_t` might be better.

alwin-joshy: Thanks, I've only been using RISCV64 so far and thought that would fix it for 32. The slowpath code which I based this off (handleVMFault in arch/riscv/kernel/vspace.c) seems to use uint64_t regardless of whether it is 32 or 64 bit. Any idea why this could be the case?

lsf37: It looks like this is wrong in the slowpath as well -- `read_stval` produces a `word_t`, and `seL4_Fault_VMFault_new` consumes a 32 bit value on rv32.

It's not catastrophic in this case, because `read_stval` will never produce values `seL4_Fault_VMFault_new` cannot consume, but it's better to fix the type.

axel-h: The RISC-V port has some places where the code could be cleaned up, because they seems just copy/paste to make it built somehow. And usually u64 often works for 32-bit also...

lsf37:
(Have set the PR to `draft` as long as there is still work on making the tests pass)

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,339 @@
...
+    /* ensure only the idle thread or lower prio threads are present in the scheduler */
+    if (unlikely(dest->tcbPriority < NODE_STATE(ksCurThread->tcbPriority) &&
+                 !isHighestPrio(dom, dest->tcbPriority))) {
+
+        vm_fault_slowpath(type);
+    }
+
+    /* Ensure that the endpoint has has grant or grant-reply rights so that we can
+    * create the reply cap */
+    if (unlikely(!cap_endpoint_cap_get_capCanGrant(handler_cap) &&
```
yyshen, line 115: should this check be done earlier, when the handler_cap is available?

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,339 @@
...
+    /* Insert reply cap */
+    word_t replyCanGrant = thread_state_ptr_get_blockingIPCCanGrant(&dest->tcbState);;
+    cap_reply_cap_ptr_new_np(&callerSlot->cap, replyCanGrant, 0, TCB_REF(NODE_STATE(ksCurThread)));
+    mdb_node_ptr_set_mdbPrev_np(&callerSlot->cteMDBNode, CTE_REF(replySlot));
+    mdb_node_ptr_mset_mdbNext_mdbRevocable_mdbFirstBadged(&replySlot->cteMDBNode, CTE_REF(callerSlot), 1, 1);
+#endif
+
+
+
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
```
yyshen, line 328: It looks like VCPU exceptions are not included in this PR?

src/fastpath/fastpath.c
```diff
@@ -472,6 +811,71 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
     callerSlot->cteMDBNode = nullMDBNode;
 #endif
 
+#ifdef CONFIG_EXCEPTION_FASTPATH
+    if (unlikely(fault_type != seL4_Fault_NullFault)) {
```
yyshen, line 815: Why are these special handling for these fault types needed? The default is restart=1 for all other fault types?

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,339 @@
...
+#endif
+
+#ifdef ENABLE_SMP_SUPPORT
+    /* Ensure both threads have the same affinity */
+    if (unlikely(NODE_STATE(ksCurThread)->tcbAffinity != dest->tcbAffinity)) {
+        vm_fault_slowpath(type);
+    }
+#endif /* ENABLE_SMP_SUPPORT */
+
+#ifdef CONFIG_ARCH_AARCH64
```
yyshen, line 149: one general comment is there are many arch-specific ifdefs. I wonder if we could have a better way to organise the code.

src/fastpath/fastpath.c
```diff
@@ -15,6 +15,339 @@
...
+#ifdef CONFIG_ARCH_AARCH32
+    switch (type) {
+    case ARMDataAbort: {
+        word_t addr, fault;
+
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+        addr = getHDFAR();
+        addr = (addressTranslateS1CPR(addr) & ~MASK(PAGE_BITS)) | (addr & MASK(PAGE_BITS));
+        /* MSBs tell us that this was a DataAbort */
+        fault = getHSR() & 0x3ffffff;
```
yyshen, line 193: should give 0x3ffffff a name.

src/fastpath/fastpath.c
alwin-joshy: They have some extra logic as the other types of exceptions may replace parts of the register set of the faulting thread. Right now, these are not necessary as we are only handling vm faults but they're more like stubs to expand upon later down the line.

src/fastpath/fastpath.c
alwin-joshy: Yeah it's not very pretty, I was considering putting the logic in architecture dependent functions but for now I thought it might be easier to see everything if it is all together.

src/fastpath/fastpath.c
alwin-joshy: It could be, but this is where the equivalent condition is checked in the seL4_Call fastpath. I don't think it would make a difference logically if it was earlier, but I thought it would be better to keep it consistent with the existing structure of other fastpath implementations.

src/fastpath/fastpath.c
alwin-joshy: This is correct. I thought they were in a different category to normal VM faults. Extending this fastpath to support them would require some as they are a different class of fault in seL4. Do you think they belong in this PR or as future work?

alwin-joshy:
> The `rv32` simulation test is failing on `VMFault`, which I assume is related.
> 
> In general, what kind of testing has been done so far? Passing sel4test is good, but it does not say much about the fast path other than that it works in a few specific cases. Any thoughts an how to substantiate that the behaviour is the same as the slow path for all cases? Even if it works now, as long as there are no proofs, we need something that alerts us when the behaviour drifts from the slowpath (e.g. because there are slowpath changes an people have forgotten to update the fastpath accordingly).
> 
> In your tests, which configurations did you use? (SMP, MCS, HYP, verification/release/debug). You can switch on the hardware test suite on the PR by adding the `hw-test` label, but it will not configure the exception fastpath by default.
> 
> Before we can merge this, we need to integrate it into CI. The main question for that is for which configurations we want the exception fastpath on. I don't think we want to double all tests (that would be easy enough to do, but it would cost a lot or test resources). It also can't be on for verification builds yet (the dependency for that still needs to be encoded in the `cmake` config). Any proposals for that?
> 
> To test the VMM with this, have a look at the setup running e.g. for [camkes-vm-examples](https://github.com/seL4/camkes-vm-examples/actions/runs/1525285140).

I have run seL4test on both MCS and non-MCS for all architectures and added some new benchmarks to seL4bench , which I have also run on both MCS and non-MCS. The seL4bench ones were always done on release while the seL4test were done on debug. The updated benchmarks for vm faults can be found here https://github.com/seL4/sel4bench/pull/12. Note that unlike in seL4test, these benchmarks use seL4_replyRecv and hence go through the reply fastpath as well. I haven't done any testing for SMP. 

Regarding your first point, I'm not sure how this could be done without adding more comprehensive tests, though the existing ones do cover the different failure cases such as active vs passive fault handler, varying priorities for the server and the client. Otherwise, I don't how it can be validated besides from knowing that vm faults take much the same path as a regular seL4_Call, with the only notable differences being the setting of the tcbFault variable and the way in which the contents of the message is generated. Similarly, for the return path the notable differences are the way in which message contents are handled and how the badge and message info registers are not set. 

I have run most of the different applications from https://github.com/seL4/camkes-vm-examples and the behavior appears to be the same as in the master branch.

alwin-joshy:
Reviving this. Have fixed the style issues raised (moving stuff to arch-dependent functions instead of ifdefs) and fixed issues with the configuration. Have also brought it back up to date with the slowpath as there were changes to aarch64HYP and aarch32HYP and the other fastpaths.  Currently implemented for aarch64, aarch32, riscv and x86_64.

alwin-joshy:
@kent-mcleod Do you have some time to look at this?

kent-mcleod:
> @kent-mcleod Do you have some time to look at this?

Once the signal fastpath PR is resolved I'll start looking at this one.

alwin-joshy:
@kent-mcleod Could you take a look at this?

lsf37:
@alwin-joshy can you please rebase and resolve conflicts first before we look into it more closely? Just want to make sure the test are running on the latest version.

src/arch/arm/c_traps.c
```diff
@@ -62,29 +62,54 @@ void VISIBLE NORETURN c_handle_enfp(void)
...
+    ksKernelEntry.word = getRegister(NODE_STATE(ksCurThread), NextIP);
+#endif
+
+#ifdef CONFIG_EXCEPTION_FASTPATH
+    fastpath_vm_fault(seL4_DataFault);
+    UNREACHABLE();
+#else
+    vm_fault_slowpath(seL4_DataFault);
+    UNREACHABLE();
+#endif
```
kent-mcleod, line 93: Why are these changes manually in-lined from c_handle_vm_fault? If we're already expecting c_handle_vm_fault to be in-lined then resulting binary should already have the call in-lined.

src/fastpath/fastpath.c
```diff
@@ -502,7 +731,52 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
+
+        /* Note - this works as is for VM faults but will need to be changed when other faults are added. This is because
+         * vm faults always restart the faulting thread upon reply but this is not necessarily the case with other types
+         * of faults, which may instead become inactive. This can either be handled in the slowpath or redirected to the
+         * slowpath, but either way it must be changed so that we do not try and forcefully switch to a thread which is
+         * meant to be inactive. */
+        if (restart) {
+            word_t pc = getRestartPC(caller);
+            setNextPC(caller, pc);
+        }
```
kent-mcleod, line 751: This should probably be simplified to just:
```suggestion
        /* Note - this works as is for VM faults but will need to be changed when other faults are added. This is because
         * vm faults always restart the faulting thread upon reply but this is not necessarily the case with other types
         * of faults, which may instead become inactive. This can either be handled in the slowpath or redirected to the
         * slowpath, but either way it must be changed so that we do not try and forcefully switch to a thread which is
         * meant to be inactive. */
        word_t pc = getRestartPC(caller);
        setNextPC(caller, pc);
```

src/fastpath/fastpath.c
```diff
@@ -502,7 +731,52 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
+
+        /* Dest thread is set Running, but not queued. */
+        thread_state_ptr_set_tsType_np(&caller->tcbState, ThreadState_Running);
+        switchToThread_fp(caller, cap_pd, stored_hw_asid);
+
+        msgInfo = wordFromMessageInfo(seL4_MessageInfo_set_capsUnwrapped(info, 0));
+
+        fastpath_restore(badge, msgInfo, NODE_STATE(ksCurThread));
+    }
+#else
```
kent-mcleod, line 778: Is it possible to use the preprocessor defines to avoid the duplication of code here and conditionally place the original into an else branch?  Something like: 
```suggestion
    } else {
#else
```

kent-mcleod: And then at replace the `#endif` below with:
```
#ifdef CONFIG_EXCEPTION_FASTPATH
    }
#endif
```

kent-mcleod:
Similarly to the Signal Fastpath, are you able to propose this PR with just the AARCH64 configuration supported, also are you able to show the comparison in fault benchmark results with and without the fault fastpath?

src/arch/arm/c_traps.c
alwin-joshy: Sorry, I'm not 100% sure what you mean by this. `vm_fault_slowpath` is what I renamed `c_handle_vm_fault` to so that I could reuse it both here and in the fastpath.

alwin-joshy: Doing it like this avoids adding an extra function that is basically identical, but it violates the convention to have the kernel be the exact same if the config option is disabled, so maybe it's better to change it

kent-mcleod: The function bodies of `c_handle_data_fault` and `c_handle_instruction_fault` are identical apart from vm_fault_type_t. Previously, c_handle_vm_fault would be inlined into each of them.  If you make the CONFIG_EXCEPTION_FASTPATH changes in the original `c_handle_vm_fault` you get the fastpath without duplicating the C code I think

alwin-joshy: Oh yeah, of course. That's my bad

alwin-joshy:
I ran the benchmarks that I added to seL4bench (https://github.com/seL4/sel4bench/pull/18), which are based on the seL4test ways of testing vm faults, as well as the IPC benchmarks and obtained the following results

# Benchmarks - aarch64 (odroidc2) MCS, all measurements in cycles #

## Without the exception fastpath  ##

faulter -> fault handler:
	mean: 730.78
	std-dev: 7.44
	samples: 100

fault handler -> faulter:
	mean: 1135.27
	std-dev: 17.82
	samples: 100

normal reply fastpath (non-faulting):
	mean: 300.875
	std-dev: 4.13
	samples: 16

## With the exception fastpath ##

faulter -> fault handler
	mean: 288.26
	std-dev: 5.83
	samples: 100

fault handler -> faulter:
	mean: 334
	std-dev: 0
	samples: 100

normal reply fastpath (non-faulting):
	mean: 300
	std-dev: 0
	samples: 16

## Summary ##

faulter -> fault handler: ~2.5x faster
fault handler -> faulter: ~3.4x faster

kent-mcleod:
Reposting this comment from the last fastpath PR: https://github.com/seL4/seL4/pull/793#issuecomment-1191915511
> Has there been any discussion about the lack of checkBudgetRestart() on this fastpath? It's not present on any of the other fastpaths either which may be an existing oversight?

src/fastpath/fastpath.c
```diff
@@ -12,6 +12,192 @@
...
+#ifdef CONFIG_KERNEL_MCS
+    handler_cap = TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbFaultHandler)->cap;
+#else
+    cptr_t handlerCPtr;
+    handlerCPtr = NODE_STATE(ksCurThread)->tcbFaultHandler;
+    handler_cap = lookup_fp(TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbCTable)->cap, handlerCPtr);
+#endif
+
+    if (unlikely(!cap_capType_equals(handler_cap, cap_endpoint_cap) ||
+                 !cap_endpoint_cap_get_capCanReceive(handler_cap))) {
```
kent-mcleod, line 41: This cap right check needs to do different things for MCS and for non-MCS.  On MCS the cap rights are checked when the cap is added to the TCB, so checking that the type is cap_endpoint_cap is sufficient.
For non-mcs, the checks need to be that the cap is an endpoint cap AND the cap has send rights, AND the cap has (either grant rights or grantreply rights):  
```
        cap_endpoint_cap_get_capCanSend(handlerCap) &&
        (cap_endpoint_cap_get_capCanGrant(handlerCap) ||
         cap_endpoint_cap_get_capCanGrantReply(handlerCap)))
```

src/fastpath/fastpath.c
kent-mcleod: While this functionality is part of the call slowpath, it's not part of the fault slowpath, so it shouldn't be part of this fastpath.  (The check at the top of this function already establishes that endpoint cap has the correct rights).

src/fastpath/fastpath.c
```diff
@@ -502,21 +696,45 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
 
-    fastpath_copy_mrs(length, NODE_STATE(ksCurThread), caller);
+        /* Clear the tcbFault variable to indicate that it has been handled. */
+        caller->tcbFault = seL4_Fault_NullFault_new();
 
-    /* Dest thread is set Running, but not queued. */
-    thread_state_ptr_set_tsType_np(&caller->tcbState,
-                                   ThreadState_Running);
-    switchToThread_fp(caller, cap_pd, stored_hw_asid);
+        /* Dest thread is set Running, but not queued. */
```
kent-mcleod, line 712: Can you expand the comment here to mention that VMFaults set thread's to ThreadState_Restart in the slowpath, but this is inlining the part of activateThread() that moves a thread from Restart to Running while updating the PC to RestartPC?

src/fastpath/fastpath.c
```diff
@@ -37,9 +37,11 @@ void NORETURN fastpath_vm_fault(vm_fault_type_t type)
     handler_cap = lookup_fp(TCB_PTR_CTE_PTR(NODE_STATE(ksCurThread), tcbCTable)->cap, handlerCPtr);
 #endif
 
+    /* Check that the cap is an endpoint cap and on non-mcs, that you can send to it and create the reply cap */
     if (unlikely(!cap_capType_equals(handler_cap, cap_endpoint_cap)
 #ifndef CONFIG_KERNEL_MCS
-                 || !cap_endpoint_cap_get_capCanSend(handler_cap)
+                 || !cap_endpoint_cap_get_capCanSend(handler_cap) || !(cap_endpoint_cap_get_capCanGrant(handler_cap) ||
+                                                                       cap_endpoint_cap_get_capCanGrantReply(handler_cap))
```
kent-mcleod, line 44: I think this is easier to read as "IF there's no send rights OR (there's no grand rights and theres no grant reply rights)":
```
                  (!cap_endpoint_cap_get_capCanSend(handler_cap) || (!cap_endpoint_cap_get_capCanGrant(handler_cap) &&
                                                                       !cap_endpoint_cap_get_capCanGrantReply(handler_cap))
```

alwin-joshy: Alright

include/arch/arm/arch/64/mode/fastpath/fastpath.h
```diff
@@ -45,6 +45,42 @@ switchToThread_fp(tcb_t *thread, vspace_root_t *vroot, pde_t stored_hw_asid)
...
+            pc = GET_PAR_ADDR(addressTranslateS1(pc)) | (pc & MASK(PAGE_BITS));
+        }
+#endif
+        NODE_STATE(ksCurThread)->tcbFault = seL4_Fault_VMFault_new(pc, fault, true);
+        break;
+    }
+    }
+}
+
+
```
kent-mcleod, line 83: I think this needs to be surrounded by `CONFIG_EXCEPTION_FASTPATH`

kent-mcleod (approved):
Once the preprocess tests pass and assuming the hardware tests pass, I'm happy for this to be merged.

lsf37 (approved):
From the verification side, this is also fine to merge. I have not tried to properly reason through equivalence, but at least I didn't spot any obvious deviations and problems and I didn't see any obvious C subset problems either.

So, good to go when the tests have passed.

PR #745, opened 2022-01-17
github: use PR number instead of ref

lsf37:
`${{github.ref}}` is not what it seems for `pull request target`. It will point to the base branch instead of `refs/pulls/<num>` like the docs would suggest. Using the PR number directly should work.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -31,7 +31,7 @@ jobs:
           - arch: X64
             session: CRefine
     # test only most recent push to PR:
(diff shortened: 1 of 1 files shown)


lsf37:
This fixes the issue discovered in #709

axel-h:
In #709 the error is still "The proofs still do not run, error remains "Canceling since a higher priority waiting request for 'seL4-PR-C-proofs-refs/heads/master-1' exists" when removing and adding the proof-test label.

axel-h:
rebased  #709 on top of this to see if this fixed the issue.

lsf37:
Looks like this is functioning properly now.

axel-h (approved):

PR #746, opened 2022-01-18
config: AARCH64 config for verification

lsf37:
For ongoing verification of seL4 on AArch64. This config has SMMU and hypervisor support enabled. AArch64 also implies FPU support is on. The target board for now is the TX2, which has an SMMU.

When this is merged, I'm planning to add the config to preprocess and compile tests.

### DIFF
--- a/configs/AARCH64_verified.cmake
+++ b/configs/AARCH64_verified.cmake
@@ -1,0 +1,28 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2022, Proofcraft Pty Ltd
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+# This is a preliminary configuration to be used for developing functional
+# correctness proofs for the AArch64 architecture.
+
+# If this file is executed then build the kernel.elf and kernel_all_pp.c file
+include(${CMAKE_CURRENT_LIST_DIR}/../tools/helpers.cmake)
+cmake_script_build_kernel()
+
+set(KernelPlatform "tx2" CACHE STRING "")
+set(KernelSel4Arch "aarch64" CACHE STRING "")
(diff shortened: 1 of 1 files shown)


lsf37:
Style test fails on the `.cmake` file being executable, which in this case we do want to keep (in analogy with the other configs).

Xaphiosis:
Fine by me, but maybe someone other than us should review as well before merging?

Xaphiosis (approved):

configs/AARCH64_verified.cmake
```diff
@@ -0,0 +1,30 @@
...
+set(KernelMaxNumNodes "1" CACHE STRING "")
+set(KernelOptimisation "-O2" CACHE STRING "")
+set(KernelRetypeFanOutLimit "256" CACHE STRING "")
+set(KernelBenchmarks "none" CACHE STRING "")
+set(KernelDangerousCodeInjection OFF CACHE BOOL "")
+set(KernelFastpath ON CACHE BOOL "")
+set(KernelPrinting OFF CACHE BOOL "")
+set(KernelNumDomains 16 CACHE STRING "")
+set(KernelMaxNumBootinfoUntypedCap 166 CACHE STRING "")
+set(KernelRootCNodeSizeBits 19 CACHE STRING "")
```
kent-mcleod, line 29: Are you able to remove these lines? KernelRootCNodeSizeBits seems like an application specific increase and KernelMaxNumBootinfoUntypedCap isn't an option yet the one on line 30 is.

lsf37: yes, happy to

lsf37: we should probably check the other verified configs, where I copy/pasted these from.

lsf37:
Removed the two lines for `KernelMaxNumBootinfoUntypedCap` and `KernelRootCNodeSizeBits`, use defaults instead.

Issue #734, opened 2022-01-02
extra boot info pages should not be executable


axel-h:
Currently the boot info frame is mapped non-executable, but the extra boot info pages are mapped executable. There seems no good use case why the extra boot info should be executable by default. The root task could still remap this e.g. for exotic things like passing a firmware binary via a device tree...
Seems the practical reason why the pages are executable is, because `create_frames_of_region()` does not expose a parameter to specify the frames should be executable or not.

kent-mcleod:
The current boot code policy gives the initial task all the initial capabilities to system resources with unrestricted rights.  The bootinfo frame and the ipc frame are mapped non-executable and everything else is executable.  The boot code policy also maps all initial task code and data frames executable at the moment.

kent-mcleod:
The roottask could always remap the extended region non-executable if it wanted to prevent itself from accidentally executing code in there.

freddy123098:
So may I help in this issue?

lsf37:
I agree, the current policy is that init task is fully trusted and must be high assurance or a high assurance system. Making the kernel code more complex to guard against initialiser bugs is the wrong trade-off.

I'm closing this issue for now as wont-do, but feel free to re-open if there is strong disagreement.

Issue #714, opened 2021-12-23
define device memory explicitly


axel-h:
Currently device untypeds are created for all physical memory from 0 to `CONFIG_PADDR_USER_DEVICE_TOP` that is not marked as reserved. This approach simplifies platform specific address space management from the kernel's point of view, because there is no need to explicitly define device memory regions for a platform. The root task receives device memory caps for all physical memory and is expected to know where the actual platform devices are that can be accesses. The downside is, that accessing invalid device memory can lead to undefined behavior. In the best case the access is simply ignore and nothing happens (write is dropped, read reads zeros or f's or a random value. It may also causes an asynchronous trap that leads to a system halt - but the root cause is difficult to debug due to the asynchronous and delayed arrival. In the worst case the behavior is completely undefined.

Ideally, the device memory ranges are taken from the device tree. Such parsing might be not trivial, so an interim solution is requiring the seL4 specific overlay device tree to define the device memory regions. The Python scripts can create a proper header file from this, sp the boot process can use this easily. On x86 we don't have device tree, but this could be hard-coded there.

The disadvantage is, that all device memory must be really well defined in advance then. There are no longer any "maybe something is here on some boards" areas. Practically this may lead to more sub-platforms. However, we already have a way to pass custom overlays to the build process, there could handle board specific differences also. Overall, the additional board/platform management seem negligible compared to the stability gain that can be enforced at compile time.

Indanz:
> The downside is, that accessing invalid device memory can lead to undefined behavior.

Is this the only reason, or are there other advantages to change how it currently is?

I ask, because the above is also true for device memory in general, which usually have reserved registers with undefined behaviour.

axel-h:
It will not solve all issues. But I assume we create seL4 systems from "scratch" (my usecase is seL4 in ROM) with maximum stability in mind.

kent-mcleod:
> Currently device untypeds are created for all physical memory from 0 to CONFIG_PADDR_USER_DEVICE_TOP that is not marked as reserved.

This is intentional.  Historically, the kernel did try to create separate device untypeds for each defined device on a platform but this added a lot of additional complexity and overhead for the fine-grained book keeping it introduced without providing too much benefit.  It also created memory pressure in the bootinfo frame because many more UT caps would get created on some platforms with a large number of devices.

[...]

PR #751, opened 2022-01-25
make boot info allocation simpler and more flexible

axel-h:
This PR collects several commit that improve the boot info frame handling.

- allocate all boot info pages together using page alignment. There is no need to require a higher alignment.
- allow static boot info to use more than one page
- mark all boot info pages non-executable (https://github.com/seL4/seL4/issues/734)
- make `SEL4_BI_FRAME_SIZE` available to userland.
- code cleanup: create helper functions for boot info data 
- drop dead ARM/RISC-V code that was supposed to write a padding chunk. Seems this was a copy/paste from x86 but it never wrote any padding block practically. And on x86 this also seems a bit artificial given the valid data length is known.
- more sanity checks when populating the boot info data
- add compile assert for `seL4_UntypedDesc` size. This header file is shared by kernel and userland. We can control the kernel compiler setting, but userland might use an arbitrary setup. Put a safeguard in place that things works as expected.
- add compile time checks for `seL4_ASIDPoolBits`: The root server object allocation only works if `seL4_ASIDPoolBits` is exactly one page. Add a build breaker if this does not hold here.
- add more comments
- trivial: align comment formatting 

It's also follow up on the rejected PR https://github.com/seL4/seL4/pull/751 and provides a way to extend the hard limits of `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` for special porting configurations.

### DIFF
--- a/include/bootinfo.h
+++ b/include/bootinfo.h
@@ -15,11 +15,3 @@
 #define BI_REF(p) ((word_t)(p))
 
 #define S_REG_EMPTY (seL4_SlotRegion){ .start = 0, .end = 0 }
-
-/* The boot info frame takes at least one page, it must be big enough to hold
- * the seL4_BootInfo data structure. Due to internal restrictions, the boot info
- * frame size must be of the form 2^n. Furthermore, there might still be code
- * that makes the hard-coded assumption the boot info frame is always one page.
- */
-#define BI_FRAME_SIZE_BITS PAGE_BITS
-compile_assert(bi_size, sizeof(seL4_BootInfo) <= BIT(BI_FRAME_SIZE_BITS))
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -41,7 +41,7 @@ static inline bool_t is_reg_empty(region_t reg)
 
 bool_t init_freemem(word_t n_available, const p_region_t *available,
                     word_t n_reserved, const region_t *reserved,
-                    v_region_t it_v_reg, word_t extra_bi_size_bits);
+                    v_region_t it_v_reg, word_t num_bi_pages);
 bool_t reserve_region(p_region_t reg);
 void write_slot(slot_ptr_t slot_ptr, cap_t cap);
 cap_t create_root_cnode(void);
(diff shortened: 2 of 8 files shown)


kent-mcleod:
> It's also follow up on the rejected PR #751 and provides a way to extend the hard limits of CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS for special porting configurations.

What is causing you to exceed the limits of CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS?

axel-h:
> > It's also follow up on the rejected PR #751 and provides a way to extend the hard limits of CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS for special porting configurations.
> 
> What is causing you to exceed the limits of CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS?
We were specifying all device regions explicitly (https://github.com/seL4/seL4/issues/714, ) which is supposed to have the ROM kernel of our chip ensure nothing else can be accessed. Since this lead to a lot of small fragmented regions, is exhausted the array size.

kent-mcleod:
> We were specifying all device regions explicitly (#714, ) which is supposed to have the ROM kernel of our chip ensure nothing else can be accessed. Since this lead to a lot of small fragmented regions, is exhausted the array size.

This policy can already be implemented with a root task like the capDL loader with a CDL spec that can break those untypeds down into per device chunks. Implementing this in the kernel, when it can be implemented at user level, would be against the design principle relating to minimalism.

axel-h:
> > We were specifying all device regions explicitly (#714, ) which is supposed to have the ROM kernel of our chip ensure nothing else can be accessed. Since this lead to a lot of small fragmented regions, is exhausted the array size.
> 
> This policy can already be implemented with a root task like the capDL loader with a CDL spec that can break those untypeds down into per device chunks. Implementing this in the kernel, when it can be implemented at user level, would be against the design principle relating to minimalism.

Sure, but nevertheless, for development purposes this is useful until there is a root task that can do all this - and this is how I ran  into this. Note that the changes in this PR do not mainly address this, it just falls out more or less for free due to the way constants are used now.

lsf37:
I think we should close this PR, it refactors too much code without clear need for kernel-level changes.

There are some parts that would be valuable to include without the larger changes, in particular the additional compile time checks.

axel-h:
Closing this and create separate PRs:
- https://github.com/seL4/seL4/pull/999
- https://github.com/seL4/seL4/pull/1000

PR #752, opened 2022-01-25
Cleanup configuration related macros (SMP, MCS)

axel-h:
- trivial: comment end of conditional code parts 
- style: apply formatting rules
- define SMP conditional macros in config.h 
- provide helper macro MCS_TERNARY 
- use MCS_TERNARY to drop macro MCS_DO_IF_BUDGET()


include/api/syscall.h
```diff
@@ -16,19 +16,6 @@
...
 
-#ifdef CONFIG_KERNEL_MCS
-#define MCS_DO_IF_BUDGET(_block) \
-    updateTimestamp(); \
-    if (likely(checkBudgetRestart())) { \
-        _block \
-    }
-#else
-#define MCS_DO_IF_BUDGET(_block) \
-    { \
```
kent-mcleod, line 27: This was done this way because the if(1) statements by using a ternary approach required proof rework.

include/config.h
```diff
@@ -8,10 +8,16 @@
 
 #include <sel4/config.h>
 
-/* Set ENABLE_SMP_SUPPORT for kernel source files */
 #ifdef CONFIG_ENABLE_SMP_SUPPORT
```
pingerino, line 11: I don't see the benefit of this commit?

include/kernel/thread.h
```diff
@@ -280,20 +280,31 @@ static inline bool_t checkBudget(void)
...
- * state to ThreadState_Restart. To be called from kernel entries
- * where the operation should be restarted once the current thread
- * has budget again.
- */
 
-static inline bool_t checkBudgetRestart(void)
+/* This is called from kernel entries where operation are done only if the
+ * thread still has budget. Otherwise it needs to be restarted first.
+ */
+static inline bool_t checkBudgetWithRestart(void)
```
pingerino, line 287: This rename is unnecessary and invalidates the caches for anyone who is deeply familiar with the MCS branch. Please don't rename things unless they are factually wrong.

include/kernel/thread.h
```diff
@@ -280,20 +280,31 @@ static inline bool_t checkBudget(void)
...
 {
+    /* The assumption here is, that an operation was started by the current
+     * thread, thus this thread must be runnable. Otherwise something is wrong
+     * or the assumption does not hold and the function was called in a wrong
+     * context. It's not fatal if this assert is removed in release builds,
+     * because the current thread's state is check before it is set to
+     * ThreadState_Restart.
+     */
     assert(isRunnable(NODE_STATE(ksCurThread)));
-    bool_t result = checkBudget();
```
pingerino, line 292: this trivial rewrite will impact the proofs which creates far more work than it's worth.

include/kernel/thread.h
```diff
@@ -280,20 +280,31 @@ static inline bool_t checkBudget(void)
...
-static inline bool_t checkBudgetRestart(void)
+/* This is called from kernel entries where operation are done only if the
+ * thread still has budget. Otherwise it needs to be restarted first.
+ */
+static inline bool_t checkBudgetWithRestart(void)
 {
+    /* The assumption here is, that an operation was started by the current
+     * thread, thus this thread must be runnable. Otherwise something is wrong
+     * or the assumption does not hold and the function was called in a wrong
+     * context. It's not fatal if this assert is removed in release builds,
```
pingerino, line 292: The "it's not fatal" comment is true of all asserts? They are mostly documentation as well as an aid to anyone developing new features (break this assert and you will have a very bad time).

Happy to improve the docs, but please keep them short and avoid documenting obvious facts about asserts.

src/api/syscall.c
```diff
@@ -222,10 +224,11 @@ exception_t handleUnknownSyscall(word_t w)
 
 exception_t handleUserLevelFault(word_t w_a, word_t w_b)
 {
-    MCS_DO_IF_BUDGET({
```
pingerino, line 230: Again, this is not a notable improvement. Please don't merge.

pingerino:
Thanks for inviting my review, I do want to be part of MCS changes. 

That said, I don't have time to review stylistic changes and they should mostly be avoided on kernel code unless something is truly broken. It's a huge context switch for me to look at seL4 kernel code. One that is made *harder* if someone has gone and rearranged the code.  

I am happy to review comment changes to make sure they fit with the intent of the design.

pingerino (changes requested):

include/config.h
axel-h: The SMP conditional macros are use all over the code and thus seem generic, defining them in `statedata.h` seems the wrong place.

axel-h: I assume you comment refer to the commit https://github.com/seL4/seL4/pull/752/commits/37a0d3b5d34b0c4eaf7e0e836f64b9e5c3302020 (define SMP conditional macros in config.h). The SMP conditional macro are used in various place in the code, thus defining them here seems better place than `statedata.h`. Using `util.h` is another option, but thy seem quite closely related to the global configuration.

include/api/syscall.h
axel-h: Ok, that explans it. I will add this as a comment then instead.

include/kernel/thread.h
axel-h: It's not a rename, because the function does more now. It consolidates what the macro `MCS_DO_IF_BUDGET()` did and this was the only place where the remove function was used.

include/kernel/thread.h
axel-h: Ok, I was not aware that the proof also care about the local variable names. This was intended to clarify what this state variable actually holds.

src/api/syscall.c
axel-h: I think replacing the macro by a clearly visible if-statement with a condition is is an improvement in terms of making it easier to understand what the code actually does. I've spend some time understanding this, especially since this macro wraps big code blocks. Furthermore, it seem this macro prevents some code analyzers to see the code block, because it is just removed - at least this would explain why the style checker now complains. Now the style check is arguably not the most important thing, but there might be other tools also.

kent-mcleod:
I'm closing this PR as we are trying to avoid any changes to MCS code that isn't fixing a known design/implementation error while MCS verification is in progress. 

There's essentially a "compilation" step from the C sources that are checked into this repository into the single kernel C file that is used as the input for the CSpec in L4V.  Unlike processing elements that don't care about the names of functions or variables by the time they execute the code, the verification uses much more information from the input source. If you treat the two repositories seL4/seL4 and seL4/L4V as separate modules/projects, then the interface between them includes the entire kernel C file, variable names and all, for each kernel configuration that L4V verifies.  The `preprocess` tests detect changes in this interface.

Many of these "unclean" C preprocessor constructions in the C code are actually necessary and intentional to produce an output C file with a particular syntax tree.  This is required to support multiple different sets of Specs and Proofs for different verification configurations.  `MCS_DO_IF_BUDGET` is needed to stay consistent with both MCS and non-MCS verification configurations which have different designs (Eg in the Exec spec: https://github.com/seL4/l4v/blob/rt/spec/haskell/src/SEL4/API/Syscall.lhs#L85-L88, https://github.com/seL4/l4v/blob/master/spec/haskell/src/SEL4/API/Syscall.lhs#L81-L82). 

Before a target configuration is verified there is a lot more uncertainty and complexity that verification engineers have to deal with.  As verification projects sit on the critical path of the kernel's roadmap we almost always try to reduce verification complexity and trade-off more C preprocessor complexity to allow simpler verification specs and proofs (as the C code ends up being more specialized to a particular configuration).  This leads to situations where the C files may be more complex for a human to read, but they end up being less complex for the verification. So while it is possible to tweak the verification to try and handle more complicated C programs (eg when using `if (1) {}` or `if(0) {}` for including/excluding functionality at compile time), this is usually only done once the underlying configuration is already verified and can be tweaked/polished by someone on the verification team who has capacity to be able to make the required changes to support it.  

MCS is currently going through a verification process. While this is happening, the verification team manually keeps the C code (CSpec) consistent with the Abstract and Executable specs, it's not until the completion of CRefine and Refine steps that the tooling can automatically detect that the various specs become inconsistent.  Without this automatic detection, making changes to the C code like this would require someone on the verification team to manually review and update the Specs to check that they stay consistent. 

Changing function names requires changing the names in all the other specs and proofs that talk about them.  Changing the contents of functions require rewriting specs and proofs.  These changes would then require all other in progress verification branches to be rebased on top.

PR #759, opened 2022-02-01
RISC-V: reserve memory for SBI in device tree

kent-mcleod:
Instead of special handling for the SBI region in the kernel, which can
be platform specific, treat it as a reserved memory region in the device
tree which is sufficient to prevent the kernel from turning the reserved
region into kernel untyped caps.

This is potentially an alternative approach to https://github.com/seL4/seL4/pull/365.  This approach doesn't prevent the kernel from creating device untypeds for SBI regions but prevents the kernel from using any of the memory regions itself. This treatment is similar to how firmware memory is treated on x86 and Arm.

Test with: seL4/seL4_tools#135

### DIFF
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -83,12 +83,31 @@ def reserve_regions(regions: Set[Region], reserved: Set[Region]) -> Set[Region]:
     return ret
 
 
+def align_memory(regions: Set[Region], config: Config) -> List[Region]:
+    ''' Given a set of regions, sort them and align the first so that the
+    ELF loader will be able to load the kernel into it. Will return the
+    aligned memory region list, a set of any regions of memory that were
+    aligned out and the physBase value that the kernel will use. '''
+
+    ret = sorted(regions)
+    extra_reserved = set()
+
+    if config.get_kernel_phys_align() != 0:
+        new = ret[0].align_base(config.get_kernel_phys_align())
+        resv = Region(ret[0].base, new.base - ret[0].base)
+        extra_reserved.add(resv)
+        ret[0] = new
+
+    physBase = ret[0].base
+    return ret, extra_reserved, physBase
+
+
 def get_physical_memory(tree: FdtParser, config: Config) -> List[Region]:
     ''' returns a list of regions representing physical memory as used by the kernel '''
     regions = merge_memory_regions(get_memory_regions(tree))
     reserved = parse_reserved_regions(tree.get_path('/reserved-memory'))
     regions = reserve_regions(regions, reserved)
-    regions, extra_reserved, physBase = config.align_memory(regions)
+    regions, extra_reserved, physBase = align_memory(regions, config)
 
     return regions, reserved.union(extra_reserved), physBase
 
--- a/include/arch/riscv/arch/32/mode/hardware.h
+++ b/include/arch/riscv/arch/32/mode/hardware.h
@@ -31,7 +31,7 @@
 
 /* The first physical address to map into the kernel's physical memory
  * window */
-#define PADDR_BASE physBase()
+#define PADDR_BASE ROUND_DOWN(physBase(),22)
 
 /* The base address in virtual memory to use for the 1:1 physical memory
  * mapping */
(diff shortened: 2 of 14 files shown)


kent-mcleod:
This will depend on some changes to the elfloader.

kent-mcleod:
> This will depend on some changes to the elfloader.

https://github.com/seL4/seL4_tools/pull/135

axel-h:
> Do we document such changes somewhere other than digging into commit messages and code?

I'd argue that documenting things in the code is the minimal thing to aim for, as digging into commit messages turns out to be painful digging into PR comment is even more pain. I'm willing to contribute write things up into a document, would the seL4 manual be the proper place then?

yyshen:
> > Do we document such changes somewhere other than digging into commit messages and code?
> 
> I'd argue that documenting things in the code is the minimal thing to aim for, as digging into commit messages turns out to be painful digging into PR comment is even more pain. I'm willing to contribute write things up into a document, would the seL4 manual be the proper place then?

I guess the manual is probably not the right place to document such details. I am thinking about a wiki for the purpose. Also we could extract some previous questions and answers from the mailing list.

lsf37:
> guess the manual is probably not the right place to document such details. I am thinking about a wiki for the purpose. Also we could extract some previous questions and answers from the mailing list.

Could be a markdown page on the docs site somewhere. That's almost as easy to change as a wiki and wouldn't need new infrastructure.

yyshen:
> > guess the manual is probably not the right place to document such details. I am thinking about a wiki for the purpose. Also we could extract some previous questions and answers from the mailing list.
> 
> Could be a markdown page on the docs site somewhere. That's almost as easy to change as a wiki and wouldn't need new infrastructure.

That sounds a good start. Let us know which page to use and we can start to fill in contents gradually.

kent-mcleod:
> That sounds a good start. Let us know which page to use and we can start to fill in contents gradually.

The docsite is already structured into separate projects, so if you were to create a subdirectory in https://github.com/seL4/docs/tree/master/projects/sel4 for design documentation/commentary, that would be a good start.

tools/hardware/utils/memory.py
```diff
@@ -83,12 +83,31 @@ def reserve_regions(regions: Set[Region], reserved: Set[Region]) -> Set[Region]:
...
+    return ret, extra_reserved, physBase
+
+
 def get_physical_memory(tree: FdtParser, config: Config) -> List[Region]:
     ''' returns a list of regions representing physical memory as used by the kernel '''
     regions = merge_memory_regions(get_memory_regions(tree))
     reserved = parse_reserved_regions(tree.get_path('/reserved-memory'))
     regions = reserve_regions(regions, reserved)
-    regions, extra_reserved, physBase = config.align_memory(regions)
+    regions, extra_reserved, physBase = align_memory(regions, config)
```
axel-h, line 110: Wouldn't is make sense to inline `align_memory() here, because it is doing something really special that is only used here?

axel-h:
Nice work, that could be sufficient. I've rebased my pending python changes (from the dropped PR https://github.com/seL4/seL4/pull/572 on this now) and https://github.com/axel-h/seL4/pull/54 contain commit that would address  my comments above.

tools/hardware/utils/memory.py
kent-mcleod: I was just reverting it back to how it was before it was refactored into the Config class.  The other functions called by get_physical_memory are helpers defined in this file, so it doesn't seem like an issue for align_memory to continue to be one too.

yyshen (approved):

Ivan-Velickovic:
@kent-mcleod the Star64 and Cheshire DTS will need to be modified as well by the way, they were added after the PR was made.

kent-mcleod:
> @kent-mcleod the Star64 and Cheshire DTS will need to be modified as well by the way, they were added after the PR was made.

I think https://github.com/seL4/seL4_tools/pull/135 can be merged independently of this PR - I'm just running some hifive tests to see if there's any breakages.  (So I temporarily removed the commit from this branch)

Then this PR needs approval from the verification side because it changes the base address of the kernel and all the preprocess checks fail.

kent-mcleod:
Ultimately, the problem with loading on riscv is that the opensbi payload doesn't get to choose its address. so wanting to locate the kernel at the lowest available region of memory (right above the memory reserved by the openSBI firmware), would clash with the elfloader's loaded address by opensbi.
The sel4_tools PR updates the elfloader to be able to move itself to its goal load address above where the kernel and roottask will be loaded before loading them.

So with that change landed, it'd be possible to merge this change which takes out the special casing for riscv so that the kernel mapping addresses are chosen the same way as on arm -> the lowest aligned region of memory. And each platform can declare a reserved memory region for firmware in its device tree.

lsf37:
This does fail the proof. The failed goal looks like it might hold of `physBase` is sufficiently aligned. I can have a look at that if the PR is otherwise ready to go.

lsf37:
Proof update in https://github.com/seL4/l4v/pull/860 -- turns out that was trivial, it just needed the same definition of KERNEL_ELF_PADDR_BASE. When the proof tests have come back, we can merge here.

lsf37 (approved):

PR #760, opened 2022-02-02
Smmu v2 fixes improvements

Apave24:
Fixes to some issues that were found when attempting to use the smmu v2 driver on zynqmp/ultra96 platforms


include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -126,8 +126,8 @@ block sid_control_cap {
 
 block sid_cap {
 
-    padding              52
-    field capSID         12
+    padding              48
+    field capSID         16
```
lsf37, line 130: Can you please elaborate on why this is necessary and what it fixes?

include/arch/arm/arch/64/mode/kernel/vspace.h
```diff
@@ -142,7 +142,8 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
     return EXCEPTION_NONE;
 }
 
+#endif
+
 void increaseASIDBindCB(asid_t asid);
 void decreaseASIDBindCB(asid_t asid);
 
-#endif
```
lsf37, line 148: This does not look correct. I can see that the previous `ifdef` is wrong (if I have traced it correctly as `AARCH64_VSPACE_S2_START_L1`), but these functions should be under an SMMU guard and not be declared when SMMU is off.

src/arch/arm/64/object/objecttype.c
```diff
@@ -163,7 +163,7 @@ finaliseCap_ret_t Arch_finaliseCap(cap_t cap, bool_t final)
     case cap_page_upper_directory_cap:
 #ifdef AARCH64_VSPACE_S2_START_L1
 #ifdef CONFIG_ARM_SMMU
-        if (cap_page_upper_directory_cap_get_capPGDMappedCB(cap) != CB_INVALID) {
+        if (cap_page_upper_directory_cap_get_capPUDMappedCB(cap) != CB_INVALID) {
```
lsf37, line 166: Does the top-level not depend on configuration? This looks like it would now break on other configs.

src/drivers/smmu/smmuv2.c
```diff
@@ -547,13 +547,13 @@ void smmu_sid_bind_cb(word_t sid, word_t cb)
     reg = S2CR_PRIVCFG_SET(S2CR_PRIVCFG_DEFAULT);
     reg |= S2CR_TYPE_SET(S2CR_TYPE_CB);
     reg |= S2CR_CBNDX_SET(cb);
-    smmu_write_reg32(SMMU_GR0_PPTR, SMMU_S2CRn(sid), reg);
+    smmu_write_reg32(SMMU_GR0_PPTR, SMMU_S2CRn(cb), reg);
```
lsf37, line 550: I don't understand this change or the commit message, can you please elaborate? Surely it is not arbitrary whether this should be the `sid` or the `cb` here.

src/drivers/smmu/smmuv2.c
```diff
@@ -547,13 +547,13 @@ void smmu_sid_bind_cb(word_t sid, word_t cb)
...
     /* The number of stream-to-context mapping
      * is related to the stream indexing method.
      * We currently supports mapping one stream ID to one context bank.*/
     if (smmu_dev_knowledge.stream_match) {
         reg = SMR_VALID_SET(SMR_VALID_EN) | SMR_ID_SET(sid);
-        smmu_write_reg32(SMMU_GR0_PPTR, SMMU_SMRn(sid), reg);
+        smmu_write_reg32(SMMU_GR0_PPTR, SMMU_SMRn(cb), reg);
```
lsf37, line 556: same as above

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -144,8 +144,8 @@ block cb_control_cap {
 block cb_cap {
 
     padding               40
-    field capBindSID      12
-    field capCB           12
+    field capBindSID      16
+    field capCB           8
```
lsf37, line 148: You can't change the size of the CB field in only one cap type and structure -- other places in the kernel are expecting the CB field to be 12 bits wide, so if this is really necessary it would need a wider change and validation. There is padding available, so it looks like this could be left at 12, but I'm curious why this is being changed.

src/arch/arm/64/object/objecttype.c
lsf37: Ah, I see the `AARCH64_VSPACE_S2_START_L1` guard now, sorry. The change should be correct as is then, please ignore.

lsf37 (changes requested):
Thanks for upstreaming fixes for this config. I don't think we have any tests for it currently.

I'm concerned at the change in size of fields, though, which will affect all boards across the architecture. Can you please comment on the background of that?

include/arch/arm/arch/64/mode/object/structures.bf
yyshen: I guess that is probably the stream ID bit size is increased on the hardware so that more devices can be supported.

lsf37: I agree, probably, but the PR should really explain that not just change numbers ;-)

My concern is that if this is different between different boards, we either need to find a maximum that suits all and then check that we're not using values outside the range at runtime, or we need to use config information here and try to do this at compile time, adjusting padding and other fields as long as there is space.

I suspect we should the a static maximum over all boards for compile time and add the runtime check against board config in the decode functions where this is used.

lsf37: And I now found the commit message that actually says this, sorry, should have seen that earlier (although it would have helped to put that info into the PR description -- review time is a scarce resource).

So this is now a conflict between what the TX2 board provides and what the SMMUv2 spec seems to say. Do we still take the maximum?

yyshen: > I agree, probably, but the PR should really explain that not just change numbers ;-)
> 
> My concern is that if this is different between different boards, we either need to find a maximum that suits all and then check that we're not using values outside the range at runtime, or we need to use config information here and try to do this at compile time, adjusting padding and other fields as long as there is space.
> 
> I suspect we should the a static maximum over all boards for compile time and add the runtime check against board config in the decode functions where this is used.

Yes, I think this is a good approach: a static max number and a runtime probe to get the actual number. I hope the static max numbers are not too big, so we do not waste too much resources.

Apave24: My apologies, I will elaborate more in the PR description in the future. I appreciate the feedback. Yes, the goal of this change was to use a static maximum taken from the SMMUv2 spec so that it can more broadly support additional boards in the future. In this specific use case, the zynqmp utilizes 15 bits (MAX_SID 32767) as opposed the 7 implemented by the TX2.

include/arch/arm/arch/64/mode/kernel/vspace.h
Apave24: Ah yes, I will wrap these in a SMMU ifdef

src/drivers/smmu/smmuv2.c
Apave24: Looking back at this piece of the SMMUv2 driver, I think sid is probably the correct value to use
here for the TX2 platform, however, this causes issues on a zynqmp platform.

The TX2 platform implements 128 stream mapping groups and 7 bit stream ids therefore a max value
of 128 so using the sid here in that case is not a problem and probably the preferred method.

The zynqmp, however, implements only 48 stream mapping groups and allows stream ids all the way up 
to a max value of 32767. Therefore, using the sid value here will quickly overrun the stream mapping 
register space and start writing data to unintended locations.

In this current SMMUv2 implementation with the constraint of 1 stream id per context bank using the
context bank index here is more appropriate for the zynqmp, however, this would need to change to provide a smarter mapping of sid value to one of the 48 mapping groups in the future to support multiple stream ids per context bank.

All of that is to say, I think the following changes should be made to this commit:
 - Wrap these changes in a platform ifdef. One that uses the original method with the sid for the TX2 and one that uses the cb for the zynqmp.
 - Update the commit message to better describe the changes
 - Move this commit to the zynqmp platform specific PR seL4/seL4#761
 
 Do those changes seem appropriate?

src/drivers/smmu/smmuv2.c
Apave24: See above reply

include/arch/arm/arch/64/mode/object/structures.bf
Apave24: The reasoning here goes back to the SMMUv2 spec just like the bit changes for the SID cap. The spec
states a maximum of 128 context banks so 12 bits is in excess of this maximum. So this could be
dropped if need be. I am not seeing off hand where the dependency on the capCB field being 12 bits is
in the rest of the kernel. Could you point me to where it is that you are referencing?

Apave24: Ah I think I found it, are you referencing `capPUDMappedCB` and similar bit fields?

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
```diff
@@ -259,7 +259,7 @@ SEL4_SIZE_SANITY(seL4_PUDEntryBits, seL4_PUDIndexBits, seL4_PUDBits);
 #else /*CONFIG_ARM_SMMU*/
 /* 1 slot at the end of the vspace is used to hold the VMID assigned to the vspace */
 /* 1 slot at the end of the vspace is used to hold the number of SMMU CBs assigned to the vspace */
-#define seL4_VSpaceReservedSlots 2
+#define seL4_VSpaceReservedSlots 2u
```
kent-mcleod, line 262: Are you able to instead change the definiton on line 279 to:
`#define seL4_UserTop (0x000000ffffffffff - seL4_VSpaceReservedSlots * 0x40000000lu)`

include/arch/arm/arch/64/mode/object/structures.bf
kent-mcleod: There was an issue associated with extra changes that would be required when supporting 16bit StreamIDs: https://sel4.atlassian.net/browse/SELFOUR-2800

include/arch/arm/arch/64/mode/object/structures.bf
lsf37: Yes, thanks for looking, those are the ones I meant.

include/arch/arm/arch/64/mode/object/structures.bf
lsf37: Hm, with 2^15 ctes that'd be 2^20 bytes, i.e. 1MB. That is a lot, but supposedly if you have a board with such a large space you'd expect it to have a lot of memory as well. Would it really be that bad to allocate that much? Doing a two-level table as described in the issue would mean that the user has to dynamically manage it (like with ASID pools). It's doable, but it would be a fairly large API change and additional mechanism.

include/arch/arm/arch/64/mode/object/structures.bf
lsf37: (If we do the the approach that we follow the maximum allowed by the spec, we should identify and change all of these)

include/arch/arm/arch/64/mode/object/structures.bf
kent-mcleod: Well the difference is that TX2 allocates SIDs to devices in order.  ZCU102 is a lot more sparse as they use some part of the ID space to encode a tag that has meaning to the hardware (I think it affects which hardware TLB gets used).  So it may be possible to make some platform specific translation that compresses the SID back to a continuous range.

include/arch/arm/arch/64/mode/object/structures.bf
Apave24: Change for CB bits already in latest master

include/arch/arm/arch/64/mode/object/structures.bf
Apave24: Bumping the conversation. Not sure how to proceed here. Do we need to have it dynamically managed to get this through? Are there any repercussions for merging this as is? Sounds like a significant amount of work to get dynamic management working. This change at least gets it working on the zynqmp platforms in the mean time.

kent-mcleod: @Apave24, are you able to explain a bit more how the SID encodings work for the zynqmp boards? According to the documentaion: "The SMMU uses a 15-bit stream ID to perform address translations. This information is part of a transaction and indicates which master originated the request. Bits [9:0] of the stream ID are the master ID defined in Table 16-13. Bits [14:10] are the TBU number the master transaction passes through. For example, for the GEM0 PS master, its master ID is 10'h074 and its TBU number is 5'h02. Concatenating these fields gives GEM0 a stream ID of
15'h0874."

If you were to enumerate all of the possible SID values that a zynqmp board has is it 2^15 or closer to the number of master IDs (2^9)?

Apave24: The total number of possible SID values is probably closer to the number of master IDs (2^10), however, the zynqmp uses 15 bits for the IDs.

Although should this discussion be holding up this PR? These commits just allow for a potential ID value up to the maximum value per the SMMUv2 spec and fixes a build time error. The amount of memory allocated is based on the SMMU_MAX_SID which is set on a per platform basis which is added for the zynqmp in PR #761 . So maybe changes in that PR need to be reworked a bit and this one is good to go?

Apave24: Reduced the cnode size in PR #761 commit [3841eca](https://github.com/seL4/seL4/pull/761/commits/3841eca2ec2d6e2b9c31ba3e2410cec1c9d29b3e)

chrisguikema:
Bumping this PR due to the issue that I just created. Can this get another round of reviews? SMMU support for the ZCU102 platforms is something DornerWorks would really like to see mainlined :)

PR #761, opened 2022-02-02
Zynqmp ultra96 smmu

Apave24:
Changes to allow zynqmp/ultra96 platforms to use the smmu v2 driver


include/drivers/smmu/smmuv2.h
```diff
@@ -8,19 +8,23 @@
 
 #include <plat/machine/devices_gen.h>
 
+#if defined(CONFIG_PLAT_ZYNQMP)
+#define SMMU_PADDR           0xfd800000
+#elif defined(CONFIG_PLAT_TX2)
 /*the paddr address of the TX2 SMMU*/
-#define SMMU_TX2_PADDR                         0x12000000
```
lsf37, line 12: Thanks for that, this was on my todo list to rename :-)

include/drivers/smmu/smmuv2.h
```diff
@@ -8,19 +8,23 @@
...
+#define SMMU_PADDR                         0x12000000
```
lsf37, line 15: nitpick: could you please change the indent on this to align with `SMMU_PAGE_4KB` below (and the one for `CONFIG_PLAT_ZYNQMP`) as well

lsf37 (approved):
This looks good from my side, only a formatting nitpick.

This PR should probably only be merged after #760 is done.

lsf37:
Please also fix the gitlint and style check errors. You can see them by clicking on "Details" in the checks list at the bottom of the PR discussion.

Gitlint is complaining about a too long title, and neither commit has a body message. In this case I'd be fine without body, because there is really not much to say, but otherwise it generally [is required](https://docs.sel4.systems/processes/git-conventions.html).

The style check is triggering on the cmake file. The easiest for that is to do a `style-all.sh config.cmake` (using the script `style-all.sh` from [seL4_tools/misc/](https://github.com/seL4/seL4_tools/tree/master/misc)

include/arch/arm/arch/object/smmu.h
```diff
@@ -7,8 +7,13 @@
 #pragma once
 
 
+#if defined(CONFIG_PLAT_ZYNQMP)
+#define SMMU_SID_CNODE_SLOT_BITS    15
+#define SMMU_CB_CNODE_SLOT_BITS     4
+#elif defined(CONFIG_PLAT_TX2)
 #define SMMU_SID_CNODE_SLOT_BITS    8
 #define SMMU_CB_CNODE_SLOT_BITS     6
+#endif
```
yyshen, line 16: I wonder if we should move these constants to per-board config so we do not have more #if else when more boards are added.

include/drivers/smmu/smmuv2.h
```diff
@@ -8,19 +8,23 @@
 
 #include <plat/machine/devices_gen.h>
 
+#if defined(CONFIG_PLAT_ZYNQMP)
```
yyshen, line 11: We could potentially extract the SMMU address from DTS, just as we did for GIC.

kent-mcleod: It already is and should be generated as the constant `SMMU_PPTR`

kent-mcleod: Actually the mapping automatically extracted from the DTS for the SMMU is only the first 4k.

src/plat/zynqmp/config.cmake
```diff
@@ -77,6 +78,9 @@ if(KernelPlatformZynqmp)
         CLK_MAGIC 1374389535llu
         CLK_SHIFT 37u
         KERNEL_WCET 10u
+        SMMU drivers/smmu/smmuv2.h
+        MAX_SID 32767
```
axel-h, line 82: Can we use `SMMU_SID_CNODE_SLOT_BITS` and `SMMU_CB_CNODE_SLOT_BITS` here
and then have in the generc code 
```
#define SMMU_MAX_SID BIT(SMMU_SID_CNODE_SLOT_BITS)
#define SMMU_MAX_CI BIT(SMMU_CB_CNODE_SLOT_BITS)
```

ArthurW404:
Hi @Apave24 , I'm looking into getting this PR merged, I was wondering how you tested whether these changes worked on the Zynqmp ultra96?

PR #763, opened 2022-02-03
arm: consolidate stage 1 translation function use

Xaphiosis:
This commit introduces `addressTranslateS1` to be used on Arm platforms
with hypervisor enabled for stage 1 (vaddr to IPA) translation. On
AArch32 this is a rename from `addressTranslateS1CPR`, and on AArch64 it
wraps `ats1e1r`. This changes to ABI on AArch64 to report faulting
address as IPA.

Reasoning:

With hypervisor enabled, AArch64 defined `addressTranslateS1CPR` to do
nothing, while AArch32 defined it to do stage 1 translation. This
delivered VM faults to the user with the faulting address being either
an IPA or a vaddr depending on mode. This inconsistency is undesireable.

This commit proposes adjusting the inconsistency to match AArch32
behaviour, as it is one of the verified platforms.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -22,7 +22,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 Further information about [seL4 releases](https://docs.sel4.systems/sel4_release/) is available.
 
 ---
-Upcoming release: BINARY COMPATIBLE
+Upcoming release: BREAKING
 
 ## Changes
 
@@ -42,6 +42,8 @@ Upcoming release: BINARY COMPATIBLE
 * Removed obsolete define `HAVE_AUTOCONF`
 * Removed user address space reserved slots restriction on 40bit PA platforms when KernelArmHypervisorSupport is set.
   This change is reflected in the definition of the seL4_UserTop constant that holds the largest user virtual address.
+* aarch32 VM fault messages now deliver original (untranslated) faulting IP in a hypervisor context, matching
+  aarch64 behaviour.
 
 ## Upgrade Notes
 ---
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -375,4 +375,12 @@ static inline word_t ats2e0r(word_t va)
 void arch_clean_invalidate_caches(void);
 void arch_clean_invalidate_L1_caches(word_t type);
 
-
+static inline paddr_t addressTranslateS1(vptr_t vaddr)
+{
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    return ats1e1r(vaddr);
+#else
+    /* shouldn't be called on non-hyp, added for consistency with AArch32 */
+    return vaddr;
+#endif
+}
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -377,10 +377,5 @@ void arch_clean_invalidate_L1_caches(word_t type);
 
 static inline paddr_t addressTranslateS1(vptr_t vaddr)
 {
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
     return ats1e1r(vaddr);
-#else
-    /* shouldn't be called on non-hyp, added for consistency with AArch32 */
-    return vaddr;
-#endif
 }
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -1213,7 +1213,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
         addr = getHDFAR();
-        addr = (addressTranslateS1CPR(addr) & ~MASK(PAGE_BITS)) | (addr & MASK(PAGE_BITS));
+        addr = (addressTranslateS1(addr) & ~MASK(PAGE_BITS)) | (addr & MASK(PAGE_BITS));
         /* MSBs tell us that this was a DataAbort */
         fault = getHSR() & 0x3ffffff;
 #else
@@ -1244,7 +1244,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
         pc = getRestartPC(thread);
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-        pc = (addressTranslateS1CPR(pc) & ~MASK(PAGE_BITS)) | (pc & MASK(PAGE_BITS));
+        pc = (addressTranslateS1(pc) & ~MASK(PAGE_BITS)) | (pc & MASK(PAGE_BITS));
         /* MSBs tell us that this was a PrefetchAbort */
         fault = getHSR() & 0x3ffffff;
 #else
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -955,7 +955,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
         /* use the IPA */
         if (ARCH_NODE_STATE(armHSVCPUActive)) {
-            addr = GET_PAR_ADDR(ats1e1r(addr)) | (addr & MASK(PAGE_BITS));
+            addr = GET_PAR_ADDR(addressTranslateS1(addr)) | (addr & MASK(PAGE_BITS));
         }
 #endif
         current_fault = seL4_Fault_VMFault_new(addr, fault, false);
@@ -970,7 +970,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
         if (ARCH_NODE_STATE(armHSVCPUActive)) {
-            pc = GET_PAR_ADDR(ats1e1r(pc)) | (pc & MASK(PAGE_BITS));
+            pc = GET_PAR_ADDR(addressTranslateS1(pc)) | (pc & MASK(PAGE_BITS));
         }
 #endif
         current_fault = seL4_Fault_VMFault_new(pc, fault, true);
--- a/src/arch/arm/api/faults.c
+++ b/src/arch/arm/api/faults.c
@@ -37,7 +37,7 @@ word_t Arch_setMRs_fault(tcb_t *sender, tcb_t *receiver, word_t *receiveIPCBuffe
         if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
             word_t ipa, va;
             va = getRestartPC(sender);
-            ipa = (addressTranslateS1CPR(va) & ~MASK(PAGE_BITS)) | (va & MASK(PAGE_BITS));
+            ipa = (addressTranslateS1(va) & ~MASK(PAGE_BITS)) | (va & MASK(PAGE_BITS));
             setMR(receiver, receiveIPCBuffer, seL4_VMFault_IP, ipa);
         } else {
             setMR(receiver, receiveIPCBuffer, seL4_VMFault_IP, getRestartPC(sender));
--- a/src/arch/arm/api/faults.c
+++ b/src/arch/arm/api/faults.c
@@ -34,14 +34,7 @@ word_t Arch_setMRs_fault(tcb_t *sender, tcb_t *receiver, word_t *receiveIPCBuffe
 {
     switch (faultType) {
     case seL4_Fault_VMFault: {
-        if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
-            word_t ipa, va;
-            va = getRestartPC(sender);
-            ipa = (addressTranslateS1(va) & ~MASK(PAGE_BITS)) | (va & MASK(PAGE_BITS));
-            setMR(receiver, receiveIPCBuffer, seL4_VMFault_IP, ipa);
-        } else {
-            setMR(receiver, receiveIPCBuffer, seL4_VMFault_IP, getRestartPC(sender));
-        }
+        setMR(receiver, receiveIPCBuffer, seL4_VMFault_IP, getRestartPC(sender));
         setMR(receiver, receiveIPCBuffer, seL4_VMFault_Addr,
               seL4_Fault_VMFault_get_address(sender->tcbFault));
         setMR(receiver, receiveIPCBuffer, seL4_VMFault_PrefetchFault,
--- a/include/arch/arm/arch/32/mode/machine_pl2.h
+++ b/include/arch/arm/arch/32/mode/machine_pl2.h
@@ -118,7 +118,7 @@ static inline void invalidateHypTLB(void)
     isb();
 }
 
-static inline paddr_t PURE addressTranslateS1CPR(vptr_t vaddr)
+static inline paddr_t PURE addressTranslateS1(vptr_t vaddr)
 {
     uint32_t ipa0, ipa1;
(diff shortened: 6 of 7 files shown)


Xaphiosis:
I discussed this issue with Kent yesterday, and he suggested making AArch32 (arm_hyp) the conforming behaviour. This is what I was able to come up with. Gerwin: FYI.

I broke something for non-hyp, I'll go fix that.

Xaphiosis:
I am somewhat unclear why non-hyp ARM would call addressTranslateS1 :/

kent-mcleod:
> I am somewhat unclear why non-hyp ARM would call addressTranslateS1 :/

It's in a if (0) block from: `if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT))`

Xaphiosis:
> > I am somewhat unclear why non-hyp ARM would call addressTranslateS1 :/
> 
> It's in a if (0) block from: `if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT))`

Ahh, that's what the code meant by unguarded. Thank you!

lsf37 (approved):
Looks good from my side, but will need the corresponding tweaks in CRefine on ARM and ARM_HYP first, of course.

We should also add an entry to the [change log](https://github.com/seL4/seL4/blob/master/CHANGES), since it is ABI-visible.

Xaphiosis:
> Looks good from my side, but will need the corresponding tweaks in CRefine on ARM and ARM_HYP first, of course.
> 
> We should also add an entry to the [change log](https://github.com/seL4/seL4/blob/master/CHANGES), since it is ABI-visible and no longer binary or source compatible (which is fine)

Advice on a good phrasing welcome, after kernel engineers give this a thumbs up. I'll get started on the proofs in a bit.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -943,7 +943,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
         /* use the IPA */
         if (ARCH_NODE_STATE(armHSVCPUActive)) {
-            addr = GET_PAR_ADDR(ats1e1r(addr)) | (addr & MASK(PAGE_BITS));
+            addr = GET_PAR_ADDR(addressTranslateS1(addr)) | (addr & MASK(PAGE_BITS));
```
yyshen, line 946: The aarch64 code returns IPA previously, this is not the VA you mentioned in the commit?

Xaphiosis: Well spotted. Sigh. I intended to refer to `Arch_setMRs_fault` in `src/arch/arm/api/faults.c`.

Xaphiosis: On AArch64, that would not have given you an IPA before this commit, but my commit message is somewhat confused. @kent-mcleod : help?

src/arch/arm/api/faults.c
```diff
@@ -37,7 +37,7 @@ word_t Arch_setMRs_fault(tcb_t *sender, tcb_t *receiver, word_t *receiveIPCBuffe
         if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
             word_t ipa, va;
             va = getRestartPC(sender);
-            ipa = (addressTranslateS1CPR(va) & ~MASK(PAGE_BITS)) | (va & MASK(PAGE_BITS));
+            ipa = (addressTranslateS1(va) & ~MASK(PAGE_BITS)) | (va & MASK(PAGE_BITS));
```
yyshen, line 40: Do we usually only invoke this on doFaultTransfer, which is invoked by doIPCTransfer to handle a fault? If so, I doubt we would hit this path for a VM.

Xaphiosis: I can't really say much on that, need @kent-mcleod ... I'm fine to rewrite the commit message (will need it anyway), or even do a different code/proof update, but need kernel people to sort out what the desired behaviour is.

kent-mcleod: Data faults and pre fetch faults still cause a fault to be delivered to the VM's fault handler, but as a VMFault type, not a VCPU fault type. So this path does get hit for a VCPU. Currently the Fault address is translated to IPA, and on arm_hyp the PC is translated to IPA, but on aarch64 the PC is left as VA.   arm_hyp has 2 definitions of `addressTranslateS1CPR`, the non-hyp version just returns its argument while the arm-hyp one does a translation.  The aarch64 architecture was only providing a single definition of the function, the non-hyp version and so was not translating properly.  This difference from arm-hyp, hadn't been noticed by us because our user level vm libraries don't to anything with this value other than print it out to console.  Most VMM fault handling code would obtain the PC from TCB_ReadRegisters and update it via TCB_WriteRegisters which consistently returns the VA on both architectures.

Xaphiosis: That means I messed up yet another thing here, on aarch64 I made addressTranslateS1 always call `ats1e1r` whether hyp is on or not. From what you're saying addressTranslateS1 should have ifdefs to check for hyp, and return its argument in the no hyp case. Is that right?

kent-mcleod: Yea.  I missed it was defined twice on arm-hyp yesterday.  A 2nd definition shouldn't really be needed as the function isn't called in non-hyp configs (because of the `if(0)` guard and other ifdefs), but it makes it consistent with arm32.

yyshen: I missed that part: sendIPC calls the function when dealing a VM fault. I think the interesting question here is what the VMM needs. On aarch64, for data abort, the fault IP is the guest VA IP and the fault address is the IPA. For instruction abort, the fault IP is the guest VA IP and the fault address is the IPA of the faulting instruction. When running a VM, usually printing out the GVA IP helps to match the fault with the guest kernel disassembly, and the VMM usually does not need to decode a faulting instruction since the FSR gives info about the fault. This looks fine for me. Another thing here is that for native seL4 threads, the translation here should not be needed given that there is S1 is disabled. Just my thoughts, could be incorrect.

yyshen: > That means I messed up yet another thing here, on aarch64 I made addressTranslateS1 always call `ats1e1r` whether hyp is on or not. From what you're saying addressTranslateS1 should have ifdefs to check for hyp, and return its argument in the no hyp case. Is that right? This mess-up might be the cause of some of the hardware failures.

ats1e1r might generate translation exceptions, but fortunately "Except as described in this section, these faults are not taken as an exception for the address translation instructions, but instead the PAR_EL1.FST field holds the fault status information. In these cases the PAR_EL1.PA field does not hold the output address of the translation."

Xaphiosis: I'll go add it for consistency then. How would you improve the commit message to make sense to the target audience?

Xaphiosis: Ok, added non-hyp `addressTranslateS1` on aarch64 for consistency.

kent-mcleod: > For instruction abort, the fault IP is the guest VA IP and the fault address is the IPA of the faulting instruction. When running a VM, usually printing out the GVA IP helps to match the fault with the guest kernel disassembly, and the VMM usually does not need to decode a faulting instruction since the FSR gives info about the fault.

If you test it out at the moment, printing faults on the TK1 (arm-hyp) print an IPA for the faulting PC and printing faults on the TX2 (aarch64-hyp) print a VA for the faulting PC.  Using TCB_ReadRegister() for the PC gives you the VA on both and the Fault address is IPA on both.  So there's an argument that whatever way the PC is reported it should be the same for 32 and 64bit

yyshen: Another issue I vaguely remember here is that the translation assumes the fault-trigger's page tables. But it could be called from a fault handler's context, so the translation would be incorrect.

kent-mcleod: > Another thing here is that for native seL4 threads, the translation here should not be needed given that there is S1 is disabled. Just my thoughts, could be incorrect.

`handleVMFault` does check `armHSVCPUActive` on aarch64 but not on aarch32.  Checking the armHSVCPUACtive state and avoiding the traversal is probably faster right?

yyshen: I am fine with both (32 hyp or 64 hyp) as long as I know where to get the addresses :-).  Adding a check means we might get different addresses based on the runtime behaviours?

Xaphiosis:
looks like Simulation (armv8a, gcc) is failing once I added the latest commit (2f6f0f5) which makes non-hyp aarch64 `addressTranslateS1` do nothing (as we discussed, added for consistency with aarch32)... that's very confusing and not at all expected based on our discussion

yyshen:
> 

this happened for hyp or non-hyp configuration? could you point to the log?

Xaphiosis:
> > 
> 
> this happened for hyp or non-hyp configuration? could you point to the log?

I'm referring to this one: https://github.com/seL4/seL4/runs/5047568059?check_suite_focus=true

Xaphiosis:
OK, I changed the problem line (I used `#elif` instead of `#else` because my brain died, but then when I rebuilt for TX2 I got no errors, which is a mystery). Let's try again.

yyshen:
> > > 
> 
> > 
> 
> > this happened for hyp or non-hyp configuration? could you point to the log?
> 
> 
> 
> I'm referring to this one: https://github.com/seL4/seL4/runs/5047568059?check_suite_focus=true
> 
> 
> 
> Actually, I read it carefully and I'm an idiot and/or tired. Please hold while I fix it.

same here. I read the commit first, but until I read the build log I noticed the cause.

Xaphiosis:
OK, now it works fine. I will squash the commit, and then wait for a verdict on whether people are happy with the change and what you want me to change in the commit message (and/or changelog).

src/arch/arm/api/faults.c
kent-mcleod: Well the check shouldn't change the outcome.  Doing a stage 1 translation for non-vcpu threads would just return the IPA because there is no stage 1 translation.

yyshen: > Well the check shouldn't change the outcome. Doing a stage 1 translation for non-vcpu threads would just return the IPA because there is no stage 1 translation.

1. We probably need to confirm that the AT instruction does not return an exception when the stage 1 is disabled. I guess this is true.
 
2. What I meant is that the function could be called when the VCPU is not running. For instance, the VCPU triggers an fault but the fault handler is not already waiting on that endpoint. The VCPU blocks, and the fault handler later receives the fault, and thus the function is call when the fault handler is running. AT instruction depends on the current context, in this the fault handler's context is used to translate the address, other than the VCPU's context.

lsf37:
Too many threads of conversation on this one (also by email). If I understand correctly, what we settled on was:

1. unify AArch32 and AArch64 (done)

2. do the S1 translation in `handleVMFault`, but do not do the translation in `Arch_setMRs_fault` (needs adjustment and re-proof for ARM_HYP)

We wanted to do (2)  because in `handleVMFault` we know we are in the right context, i.e. the page tables that cause the fault are active, but we don't know that we are in the right page table context in `Arch_setMRs_fault`, because that could be called on a code path originating from a fault handler with different page tables.

(2) is small bug fix, so should go into the change long.

@yyshen, @kent-mcleod, @Xaphiosis agreed?

kent-mcleod:
> We wanted to do (2) because in `handleVMFault` we know we are in the right context, i.e. the page tables that cause the fault are active, but we don't know that we are in the right page table context in `Arch_setMRs_fault`, because that could be called on a code path originating from a fault handler with different page tables.

There's also 2 types of seL4_Fault_VMFault for each config, a Kernel type that's stored in a TCB object to save the fault, and there's the type that's sent in the IPC buffer to user level:

Kernel: `seL4_Fault_VMFault_new(uint64_t address, uint64_t FSR, uint64_t instructionFault)`

User: `seL4_Fault_VMFault_new(seL4_Uint64 IP, seL4_Uint64 Addr, seL4_Uint64 PrefetchFault, seL4_Uint64 FSR)`

There's always 2 addresses that can be optionally translated for a single VMFault.  The program counter, and the address that caused the fault. When the fault is an instruction fault (also called a PrefetchFault) the fault address and the program counter are the same.

What I believe we agreed on was that:
- we always translate the fault address in `handleVMFault()` for all cases.  In hypervisor contexts this requires a manual stage 1 translation.
- we never translate the IP address in Arch_setMRs_fault() or anywhere else. arm_hyp does translate currently, but it is a bug: https://sel4.atlassian.net/browse/SELFOUR-1602.  aarch64-hyp does no translation currently.  If we wanted to perform a translation and return a stage 2 address for IP we would need to change the kernel's definition of the seL4_Fault_VMFault type to include the IP in order to save the translated value in the TCB.

Xaphiosis:
I have added a commit that removes stage 1 translation from `Arch_setMRs_fault`. I am unsure if I need to modify `handleVMFault` on 32-bit Arm to also check `armHSVCPUActive` like on 64-bit (I thought this part worked so left it alone).

src/arch/arm/32/kernel/vspace.c
```diff
@@ -1213,7 +1213,7 @@ exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
         addr = getHDFAR();
-        addr = (addressTranslateS1CPR(addr) & ~MASK(PAGE_BITS)) | (addr & MASK(PAGE_BITS));
+        addr = (addressTranslateS1(addr) & ~MASK(PAGE_BITS)) | (addr & MASK(PAGE_BITS));
```
kent-mcleod, line 1216: @yyshen, do you think we should add `if (ARCH_NODE_STATE(armHSVCPUActive)) {` here to avoid a no-op with `addressTranslateS1`?

Xaphiosis:
Rebased and updated CHANGES; as agreed with @kent-mcleod this is `BREAKING`.

Xaphiosis:
Tweaked commit messages. Verification side is nearly done. Should we do another hardware test at this point?

CHANGES
```diff
@@ -42,6 +42,8 @@ Upcoming release: BINARY COMPATIBLE
 * Removed obsolete define `HAVE_AUTOCONF`
 * Removed user address space reserved slots restriction on 40bit PA platforms when KernelArmHypervisorSupport is set.
   This change is reflected in the definition of the seL4_UserTop constant that holds the largest user virtual address.
+* aarch32 VM fault messages now deliver original (untranslated) faulting IP address in a hypervisor context, matching
+  aarch64 behaviour.
```
lsf37, line 46: Potentially: 

```suggestion
* aarch32 VM fault messages now deliver original (untranslated) faulting IP in a hypervisor context, matching
  aarch64 behaviour.
```

(it sounded like a network IP address to me on first read, but it's the instruction pointer, of course)

lsf37 (approved):
Minor wording suggestion, but also happy for you to keep the original if you think it's clear enough. Just tripped me up on first read.

Content looks good.

CHANGES
Xaphiosis: Good point (ha). Will update after tests, thanks!

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -375,4 +375,12 @@ static inline word_t ats2e0r(word_t va)
 void arch_clean_invalidate_caches(void);
 void arch_clean_invalidate_L1_caches(word_t type);
 
-
+static inline paddr_t addressTranslateS1(vptr_t vaddr)
+{
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    return ats1e1r(vaddr);
+#else
+    /* shouldn't be called on non-hyp, added for consistency with AArch32 */
```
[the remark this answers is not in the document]
Xaphiosis: Will do.

Xaphiosis:
Hardware tests all succeeded. Incorporated Gerwin's and Alex's suggestions. We can now consider this round done. Would be nice to get an approval other than Gerwin's while I rebase the verification side.

kent-mcleod (approved):

PR #764, opened 2022-02-03
aarch64: verification VCPU tweaks

Xaphiosis:
In `vcpu_enable` and `vcpu_disable`, use `setHCR` to wrap setting
REG_HCR_EL2 (similar to AArch32).
Reorder `vcpu_enable` to match AArch32 operation order for easier proof.

### DIFF
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -402,6 +402,11 @@ static inline void writeVMPIDR_EL2(word_t reg)
     MSR(REG_VMPIDR_EL2, reg);
 }
 
+static inline void setHCR(word_t reg)
+{
+    MSR(REG_HCR_EL2, reg);
+}
+
 static word_t vcpu_hw_read_reg(word_t reg_index)
 {
     word_t reg = 0;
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Looks good from my side, but would appreciate an opinion from @yyshen or @kent-mcleod

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -619,7 +624,7 @@ static inline void vcpu_disable(vcpu_t *vcpu)
     /* Stage 1 MMU off */
     setSCTLR(SCTLR_DEFAULT);
     isb();
-    MSR(REG_HCR_EL2, HCR_NATIVE);
+    setHCR(HCR_NATIVE);
```
yyshen, line 627: the reorder should not have impact, but let see how HWrun goes.

kent-mcleod: If the reorder doesn't have an impact, then could the intermediate `isb()` be removed?

Xaphiosis: I can remove it, it's not present on arm_hyp, but don't know the implications

lsf37:
Hardware test has succeeded apart from the known-failing RISCV test and two instances of the known-flaky MULTICORE004 test which at least should be unrelated to this PR.

kent-mcleod (approved):

Xaphiosis:
Removed the extra isb and rebased.

Xaphiosis:
@yyshen after removing the isb, looks like this one failed https://github.com/seL4/seL4/runs/5059510776?check_suite_focus=true for an unrelated reason. advice? otherwise am ready to merge

yyshen (approved):

lsf37:
Can confirm that the IMX8MQ failure is unrelated.

yyshen:
> Can confirm that the IMX8MQ failure is unrelated.

yep, I looked at the log before clicked the button.

PR #766, opened 2022-02-05
remove usage of ANY in dtd

ratmice:
This patch restricts mixed-content elements from ANY
to the various leaf nodes which are used.

This is intended to allow for easier parsing in languages with
no recursive datatypes like rust.
Since they no longer have to handle things like
```
<description> text
   <description>more text</description>
<description>
```

which the dtd allowed for, but never used.

tested with:
```
for i in {libsel4/arch_include/*/interfaces/,\
libsel4/include/interfaces/,\
libsel4/sel4_arch_include/*/interfaces}/*.xml;
do
     xmllint --noout --dtdvalid libsel4/tools/sel4_idl.dtd $i;
done
```

### DIFF
--- a/libsel4/tools/sel4_idl.dtd
+++ b/libsel4/tools/sel4_idl.dtd
@@ -26,13 +26,13 @@
 <!ATTLIST method manual_name CDATA #IMPLIED>
 <!ATTLIST method manual_label CDATA #IMPLIED>
 
-<!ELEMENT param (#PCDATA|description|error)*>
+<!ELEMENT param (description?, error*)>
 <!ATTLIST param type CDATA #REQUIRED>
 <!ATTLIST param name CDATA #REQUIRED>
 <!ATTLIST param dir CDATA #REQUIRED>
 <!ATTLIST param description CDATA #IMPLIED>
 
-<!ELEMENT error (#PCDATA|description)*>
+<!ELEMENT error (description?)>
 <!ATTLIST error name CDATA #REQUIRED>
 <!ATTLIST error description CDATA #IMPLIED>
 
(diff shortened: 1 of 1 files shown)


libsel4/tools/sel4_idl.dtd
```diff
@@ -26,23 +26,23 @@
...
 <!ATTLIST error description CDATA #IMPLIED>
 
-<!ELEMENT brief ANY>
+<!ELEMENT brief (#PCDATA|texttt|docref|shortref|autoref|obj)*>
 
-<!ELEMENT description ANY>
+<!ELEMENT description (#PCDATA|texttt|docref|shortref|autoref|obj)*>
 
-<!ELEMENT return ANY>
+<!ELEMENT return (#PCDATA|texttt|docref|shortref|autoref|obj|errorenumdesc)*>
```
ratmice, line 43: Wasn't sure if `errorenumdesc` belonged in the other ones or not it wasn't used anywhere but `return`.

lsf37 (approved):
Looks good from my side, and reminds me that the xml lint check should also run on dtd changes..

lsf37:
Could you please a Signed-off-by for the commit?

ratmice:
added signed-off-by, and updated to the lastest.

ratmice:
So I noticed one more thing (I had accidentally added self-recursion in `docref`, updated patch removes that.

libsel4/tools/sel4_idl.dtd
lsf37: I think it makes sense to restrict it to this one. We can always add it elsewhere if we decide it is necessary.

lsf37 (approved):

ratmice:
Since i've made a few changes since starting this,
I wanted to confirm that I believe this now does everything I set out to:

I put my rust parser which relies on this formulation of the dtd here: https://github.com/ratmice/sel4_rust_tools
It could use some more testing, and isn't pretty, but should follow the dtd closely.

PR #770, opened 2022-02-11
Remove globals frame caveat

kent-mcleod:
There is no longer a globals frame and thus no longer a caveat about it.
The globals frame was removed when ARMv6 support was removed recently.


lsf37 (approved):
Thanks. Up-to-date documentation is hard :)

PR #771, opened 2022-02-14
use the loop.first builtin instead of namespaces

ratmice:
This allows the template to be parsed with minijinja which doesn't support
namespaces or set (at this time).

I thought it was a minor cleanup since it led to less imperative state during template generation.
The other jinja templates like syscalls (not touched here) are harder to port, so i've left that for another time.

thus, if it pleases you; if not we can close this.


tools/invocation_header_gen.py
```diff
@@ -14,8 +14,8 @@
 import sys
 import xml.dom.minidom
 import pkg_resources
-# We require jinja2 to be at least version 2.10 as we use the 'namespace' feature from
-# that version
+# We require jinja2 to be at least version 2.10,
+# so we can use use the 'namespace' feature from that version.
```
axel-h, line 18: That comment feels a bit wrong if namespaces are no longer used once the changes are merged. Maybe put the commit comment in here, it it's clear that being able to use `minijinja` is the reason certain features are not used.

ratmice: Yeah, its a bit of a weird state, other templates still use `namespace`,  In theory in the future I wanted to keep open that sel4 could add more namespace features in the future, even though `minijinja` doesn't support it there is no proposal to actually use minijinja in tree.

In theory this version of the template would allow us to downgrade the version requirements, I didn't want to do that in case there was a need for this feature in the future.

ratmice: I'll try and reword this so that intent is better reflected.

ratmice:
> Where do you use `minijinja`? The issue with this change is, if there is nothing in the default build flows that use this tool, any commit can break this again easily. Will there be follow-up PRs here or in other seL4 repos use this tool?

It is used in progress version of https://github.com/ratmice/sel4_rust_tools minijinja is a rust library by the author of jinja.
As such I have to copy/paste the template in and adjust for rust string constants etc, so changing upstream won't affect my usage (for better or worse).

My thoughts were I would get it working out of tree before we start discussing whether it belongs in tree, and/or whether it could be tested during sel4 CI.

At this point I just wanted to backport changes I have made to get the template working.

ratmice:
Would it be better if I just remove the reference to minijinja in the commit message/comment?
As far as I'm concerned that is an experiment which may or may not work out, and I'd rather that the commit just be judged on its own merits whether it is considered a worthwhile cleanup or not.

but still wanted to give some indication for why I was doing it, but this thread probably suffices.

axel-h:
Besides https://docs.sel4.systems/processes/git-conventions.html, my rationale usually is, that comments in the code are the first thing others will see in the long run. So this should explain the background and intentions, especially if there is a bigger plan behind something that is not obvious. Such a plan might also block short term improvement then that would break it.
Commit messages are something found when digging though the history - and this history may not be available in the worst case. So they should give a good hint what this code change is about. Commit messages can also better explain the delta, while code comments should be careful about explaining too much about how things were before the change, because that is gone and can be confusing in the long run.

I think in this case the PR discussion could be sufficient to clarify if this can be merged and what is the plan for Rust support. The PR can also be left in draft state for a while until other Rust support is about to be merged and there is a stronger need to ensure minijinja must be supported. In the practical side, I consider mentioning "minijinja" both in the commit message and a code comment as useful.

ratmice:
I somewhat hesitate to mark it as a draft, because I think i'm pretty much done here.
It would probably easier/simpler to use a list of dicts/structures instead of the lists of tuples in rust,
But i'm pretty set on keeping close to the current tools in this regard, if only so its easier to keep them up to date.

tools/invocation_header_gen.py
```diff
@@ -14,8 +14,11 @@
 import sys
 import xml.dom.minidom
 import pkg_resources
-# We require jinja2 to be at least version 2.10 as we use the 'namespace' feature from
-# that version
+# We require jinja2 to be at least version 2.10,
+# In the past we used the 'namespace' feature from that version.
+# other versions of jinja particularly `minijinja` doesn't namespace.
```
axel-h, line 19:
```suggestion
# other versions of jinja particularly `minijinja` don't support namespaces.
```

ratmice:
I addressed your comments, and pushed a first pass at the the rust implementations of header and `syscall_invocation_gen.py` using minijinja to https://github.com/ratmice/sel4_rust_tools.

The tools themselves could certainly used some cleanup though so don't expect anything pretty

kent-mcleod (approved):

tools/syscall_header_gen.py
```diff
@@ -201,18 +193,36 @@ def convert_to_assembler_format(s):
     return '_'.join(words).upper()
 
 
+def map_syscalls_neg(syscalls):
+    # This function will map a list of (condition, syscall_list) tuples
+    # From: [(cond1, ["SyscallOne", "SysCallTwo"]),
+    #        (cond2, ["SyscallThree", ...])]
+    # Into: [(cond1, [("SyscallOne", -1), ("SyscallTwo", -2)]),
+    #        (cond2, [(SyscallThree", -3), ...])]
+    r = range(1, sys.maxsize)
```
axel-h, line 202: Python has `itertools` that makes this easier. With array comprehensions there is no need to use map() and zip() either 
```
import itertools
....   

def map_syscalls_neg(syscalls):
    r = itertools.count(start = -1, step = -1)
    return [ (cond, [ (s, next(r)) for s in lst ]) for (cond, lst) in syscalls  ] 
```

tools/syscall_header_gen.py
```diff
@@ -201,18 +193,36 @@ def convert_to_assembler_format(s):
...
+        return (syscalls[0], list(variants))
+    return map(helper, syscalls)
+
+
 def generate_kernel_file(kernel_header, api, debug):
     template = Environment(loader=BaseLoader, trim_blocks=False,
                            lstrip_blocks=False).from_string(KERNEL_HEADER_TEMPLATE)
-    data = template.render({'assembler': api, 'enum': api + debug,
-                            'upper': convert_to_assembler_format})
+    data = template.render({'assembler': list(map_syscalls_neg(api)),
```
axel-h, line 215: With https://github.com/seL4/seL4/pull/771/files#r809640092 there is no need to use `list()` here to conver tthis.

tools/syscall_header_gen.py
```diff
@@ -201,18 +193,36 @@ def convert_to_assembler_format(s):
...
+                            'enum': list(map_syscalls_neg(api + debug)),
+                            'upper': convert_to_assembler_format,
+                            'syscall_min': sum(map(lambda l: len(l[1]), api)) * -1})
```
axel-h, line 218:
```suggestion
  'syscall_min': -sum(map(lambda l: len(l[1]), syscalls)) 
```
or use `-sum([ len(lst) for (cond, lst) in api ])` to avoid lambdas completely`

tools/syscall_header_gen.py
ratmice: Not much of a python person so thanks for these, they seem a bit cleaner.

tools/syscall_header_gen.py
axel-h: This still has the `list(...)` in there, do we really need this?

axel-h (approved):

axel-h:
Is there anything left that prevents this from merging?

lsf37 (approved):
Happy with it.

PR #773, opened 2022-02-17
riscv: fpu extension fixes

andybui01:
This includes a fix for the riscv F extension as well as a simplification of the riscv `config.cmake`.

### DIFF
--- a/include/arch/riscv/arch/machine/fpu.h
+++ b/include/arch/riscv/arch/machine/fpu.h
@@ -27,7 +27,7 @@ static inline void set_fs_off(void)
 
 #define FL "flw"
 #define FS "fsw"
-#define FP_REG_TYPES "4"
+#define FP_REG_BYTES "4"
 
 #endif
 
--- a/src/arch/riscv/config.cmake
+++ b/src/arch/riscv/config.cmake
@@ -68,8 +68,9 @@ elseif(KernelPTLevels EQUAL 4)
 endif()
 
 if(KernelRiscvExtD)
+    # The D extension depends on the base single-precision
+    # instruction subset F.
     set(KernelRiscvExtF ON)
-    set(KernelHaveFPU ON)
 endif()
 
 if(KernelRiscvExtF)


src/arch/riscv/config.cmake
```diff
@@ -67,12 +67,7 @@ elseif(KernelPTLevels EQUAL 4)
     math(EXPR KernelPaddrUserTop "1 << 56")
 endif()
 
-if(KernelRiscvExtD)
-    set(KernelRiscvExtF ON)
```
kent-mcleod, line 71: If D extension is enabled, then the F extension is also enabled should be intentional here?

andybui01: I don't think that has any intentional effect though, since the F extension ifdefs _only_ appear as an `#elif` to the D extension ifdefs in the kernel.

yyshen: This chapter describes the standard double-precision floating-point instruction-set extension, which
is named “D” and adds double-precision floating-point computational instructions compliant with
the IEEE 754-2008 arithmetic standard. The D extension depends on the base single-precision
instruction subset F

include/arch/riscv/arch/machine/fpu.h
```diff
@@ -27,7 +27,7 @@ static inline void set_fs_off(void)
 
 #define FL "flw"
 #define FS "fsw"
-#define FP_REG_TYPES "4"
+#define FP_REG_BYTES "4"
```
yyshen, line 30: this is a nice catch. Thanks.

src/arch/riscv/config.cmake
andybui01: To me, `CONFIG_HAVE_FPU` and `CONFIG_RISCV_EXT_F` are the same, that is, they both enable the most basic FPU instruction subset. Would it be useful to instead set `KernelRiscvExtF`_only_ (and not KernelHaveFpu) when we set `KernelRiscvExtD`, to be more in line with the spec?

A comment would explaining this might suffice too.

andybui01: > Seems currently no RISC-V platform enables anything there, it's all done manually? So we should even consider sanitizing things here and abort if we detect explicit combinations that we can't make sense of. That avoids building future legacy.

I'm a bit confused and not sure what you mean by stuff being done manually. Also what would we be sanitizing?

axel-h: I could not find any place that enabled the FPU on RISC-V, so the only way how to do this at the moment is specifying this as an explicit build parameter manually. Now is `KernelRiscvExtD` is enabled there, we need some sanitizing in CMake that also enabled `KernelRiscvExtF` then if this is not set - and it might better abort is it detects this is already set explicitly to "OFF", as this looks like a broken configuration then (or something that needs to be clarified, what this is supposed to achieve)

yyshen (commented):
Enabling EXT_D also enables EXT_F seems fine for me, since the hardware manual says EXT_D depends on EXT_F. Otherwise, this looks good to me.

lsf37:
@andybui01  one more minor request: when you say `spec` in the commit message, please clarify which spec. There are many specs in seL4 that could be relevant. Here it seems to be the RISC-V ISA spec.

axel-h (approved):
Looks fine from my side, good to be merged

andybui01:
https://github.com/seL4/seL4/blob/71c8707c6e4f2d04d116531e923a1f4ded460d1c/CMakeLists.txt#L88-L98

I was testing different combinations of 32/64 bit RISC-V with either F and D extensions enabled and my attention was drawn to these lines.

It seems as if we:

1. Assume that all floating point usage on RISCV64 is double precision.
2. Assume that there should be no floating point instructions compiled on RISCV32.

```cmake
if(KernelSel4ArchRiscV64)
        if(KernelRiscvExtD)
            string(APPEND common_flags " -march=rv64imafdc -mabi=lp64d")
        elseif(KernelRiscvExtF)
            string(APPEND common_flags " -march=rv64imafc -mabi=lp64f")
        else()
            string(APPEND common_flags " -march=rv64imac -mabi=lp64")
        endif()
    elseif(KernelSel4ArchRiscV32)
        # same but rv32
    else()
        message(FATAL_ERROR "unknown RISC-V KernelSel4Arch '${KernelSel4Arch}'")
endif()
```

However, I'm not sure introducing lines that look like giant switch-cases is the best way to go about this, maybe theres a cleaner way?

axel-h:
I'd propose to merge https://github.com/seL4/seL4/pull/783 and then merge this on top. Any objections?

andybui01:
Rebased and squashed `config.cmake` commits

yyshen (approved):

PR #776, opened 2022-02-24
RISC-V/CMake: Fix build when using GNU binutils >= 2.38

### LINKED ISSUE #749: GCC 11.3 RISC-V ISA spec change from 2.2/20190608 to 20191213 might affect seL4
In case we consider switching to GCC 11.3 or higher one day, we might run into this change causing problems:
- https://gcc.gnu.org/gcc-11/changes.html
- https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=98416dbb0a62579d4a7a4a76bab51b5b52fec2cd

There is a mail on the RISC-V SW-Dev mailinglist (sw-dev@groups.riscv.org) that describes the details (see also https://groups.google.com/a/groups.riscv.org/g/sw-dev/c/aE1ZeHHCYf4/m/v88QeovDAAAJ):

It’s Kito Cheng from the RISC-V GCC community, just sharing some news about the default ISA spec version that has been bumped to 20191213 on both RISC-V GCC and binutils recently, and that has one major incompatibility issue between current default ISA spec versions. The major incompatibility issue is the csr read/write (csrr*/csrw*) instructions and fence.i instruction has separated from the `I` extension, become two standalone extensions: Zicsr and Zifencei; so you might get error messages like that: unrecognized opcode `csrr`.
# How to resolve those issues?
Here is two solution to resolve:
- First one is what we recommend, adding zicsr and/or zifencei to your -march option, e.g. -march=rv64imac become -march=rv64imac_zicsr_zifencei
- The second one is a kind of workaround, using -misa-spec=2.2 to force RISC-V GCC using the older ISA spec version, however this is strongly  *NOT* recommended.
In case you want everything like before, you can add --with-isa-spec=2.2 build option for GCC and binutils, then the toolchain will default using the old version and work as an older GNU toolchain release.
# Why GCC/binutils bump that?
The default ISA spec version was set to 2.2 for long time, however the spec is released at May, 2017, which is 5 years ago, and RISC-V has at least two ISA spec release after that, so we believe that’s time to bump, and we also raise the discussion on the RISC-V BoF at Linux plumber, we reach a consensus is that we should moving forward even that having incompatible issue.
# Why bump that so late?
We worried about the incompatibility issue, but we realized that is unavoidable, and RISC-V is growing very quickly in the past year, so we believe it would be better to move forward now.
# Why is there an incompatibility issue there? Why can't you avoid that?
That is because the RISC-V ISA spec has changed, we believe the toolchain should honor what spec written.
# Which ISA spec versions are supported by GNU toolchain now?
Currently we support 2.2 20190608 and 20191213.
2.2: https://github.com/riscv/riscv-isa-manual/releases/tag/riscv-user-2.2
20190608: https://github.com/riscv/riscv-isa-manual/releases/tag/Ratified-IMFDQC-and-Priv-v1.11
20191213: https://github.com/riscv/riscv-isa-manual/releases/tag/Ratified-IMAFDQC
# Is there any newer ISA spec supported? It’s 2022!
20191213 is already 2 years ago, but RISC-V didn’t have further formal ISA spec release after that, what we can promise is that RISC-V GNU toolchain will support soon if ISA spec has newer release.

florianhofhammer:
It's not possible to build the RISC-V kernel anymore with GNU binutils version 2.38 since it requires specifying the Zicsr and Zifencei extensions explicitly.

This patch adds a check for the GNU binutils version used to build the kernel and adds the corresponding flags if necessary.

I'm not sure whether this is an elegant solution, any suggestions are welcome.

Edit:
Fixes #749

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,17 @@ elseif(KernelArchRiscV)
     # enabled
     string(APPEND _riscv_march "c")
 
+    # Determine if GNU toolchain is used and if yes, whether GCC version >= 11.3 (implies binutils version >= 2.38)
+    if(
+        CMAKE_ASM_COMPILER_ID STREQUAL "GNU"
+        AND CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "11.3"
+    )
+        # Manually enable Zicsr and Zifencei extensions
+        # This became necessary due to a change in the default ISA version in GNU binutils 2.38 which is the
+        # default binutils version shipped with GCC 11.3
+        string(APPEND _riscv_march "_zicsr_zifencei")
+    endif()
+
     string(APPEND common_flags " -march=${_riscv_march} -mabi=${_riscv_mabi}")
 
 else()


axel-h:
Thanks, is this addressing issue https://github.com/seL4/seL4/issues/749 ?

florianhofhammer:
Well... I totally missed that issue and just came across this problem when I updated my local RISC-V toolchain. Sorry for that!
But yes, it's targeting exactly that problem. However, I'll convert the PR into a draft PR for now, since I apparently missed quite some points. If you don't mind, I'll just ping you as soon as I'm confident that the PR is ready for review (i.e., tests pass).

axel-h:
It's good to know that somebody is working on this issue now, so there is some help for others that run into this.

florianhofhammer:
The tests should now all pass (at least they did in my forked repository).
The CMake configuration now adds the necessary extensions to the `-march` string if and only if GNU binutils >= 2.38 and GCC >= 10 are used.

I locally tested all combinations of GCC {8.3, 9.1, 10.2, 11.1} and GNU binutils {2.36, 2.38} and derived the compatibilities from those experiments.
The check for the GCC version should in practice never cause issues, since it's very unlikely that the newest binutils version is paired with a multiple years old GCC version.
Nevertheless, better safe than sorry.

axel-h:
Assuming we have to deal with RISC-V extensions in the future again, I wonder if we should make things a bit more generic.
In https://github.com/axel-h/seL4/tree/patch-axel-83 I've put one commit https://github.com/axel-h/seL4/commit/99a0fd03fba0cf97cf0b3acc41a2a998982b00ec on top if your changes that tries to avoid redundancies a bi, might be a try merging this into your changes

CMakeLists.txt
```diff
@@ -85,14 +85,46 @@ elseif(KernelArchARM)
...
+        )
+        # Check whether binutils >= 2.38
+        if(_gas_version_num VERSION_GREATER_EQUAL "2.38")
+            # Require gcc >= 10, otherwise it gets confused by the RISC-V Z extensions
+            if(CMAKE_C_COMPILER_VERSION VERSION_LESS "10")
+                message(
+                    FATAL_ERROR
+                        "Incompatible GCC and GNU binutils versions ${CMAKE_C_COMPILER_VERSION} and ${_gas_version_num}"
+                )
+            else()
```
kent-mcleod, line 109: As GCC 11.3 requires adding the extensions by default and would seem that using GCC 11.3 is required for using binutils >=2.38, could this compatibility check be reduced to just checking GCC version:
```
        if(
            (CMAKE_C_COMPILER_ID STREQUAL "GNU")
            AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "11.3.0")
        )
                string(APPEND KernelRiscVMachFeatureModifiers "_zicsr_zifencei")
        endif()
```

florianhofhammer: @kent-mcleod @axel-h I'm just replying here once since your comments about GCC versions seem to point into the same direction, I hope you don't mind! :)

As far as I understood the GCC and binutils changelogs correctly, it should technically be sufficient to check for GCC >= 11.3 because that's when it by default starts to ship with binutils 2.38. However, this combination cannot be guaranteed. In my local tests, I compiled multiple versions of GCC and the GNU binutils and checked their compatibility with regard to the extension flags. This is where my checks originate from: specifying the Zicsr and Zifenci extensions is only necessary from binutils 2.38 onwards but GCC already supports parsing/forwarding those flags to the assembler backend from version 10 onwards.
Even though it is unlikely that a combination of GCC 9.x and binutils 2.38 is encountered "in the wild", my checks explicitly make sure that the project configuration already fails and we don't encounter unexpected errors during the build process.

I agree that my checks are quite verbose, I just wanted to rule out as many error causes as possible. Do you think that a best-effort approach just checking the GCC version here is sufficient?

axel-h: What exactly happens when you use binutils 2.38 with an older GCC, do RISC-V builds raise an error due to some incompatibility eventually? What is the error? This seems be a general RISC-V issue then, that is no just seL4 specific?

florianhofhammer: Yes, this is a general RISC-V toolchain issue. E.g., GCC 9.1 in combination with binutils 2.38 errors out with `unsupported ISA subset 'z'`, it seems as if it simply cannot parse the ISA correctly and pass it on to the assembler.
Since this is not an officially supported combination, it makes sense that there are breaking changes/incompatibilities at some point.

axel-h: There are the "official" seL4 docker containers that contain the built environment that are supported and that are used for the verification. Saying that all other environment are "don't care" is a bit harsh, but adding too many check for tool combinations would also become overkill. Given this is a hard break in the whole GCC/RISC-V ecosystem, but it also yields an error message `unsupported ISA subset 'z` that one's favorite search engine will provide enough hints in the long run (e.g. https://www.mail-archive.com/ubuntu-bugs@lists.ubuntu.com/msg5993913.html), I think limiting this to checking GCC 11.3 and adding the extenstion "flags" is good enough. In the near future this might be a bit annoying, but in the long run I'd expect this to resolve itself. 

While we are at that topic, it seems the LLVM/clang world is totally unaffected of this?

CMakeLists.txt
```diff
@@ -91,29 +91,36 @@ elseif(KernelArchRiscV)
         # => can't change that easily without messing up the build system
         execute_process(COMMAND ${CROSS_COMPILER_PREFIX}as --version OUTPUT_VARIABLE _gas_version)
         # Extract version number
-        string(REGEX
-            MATCH
```
axel-h: nitpicking: can you make this syle change a separate commit?

CMakeLists.txt
florianhofhammer: I'm sorry for the late reply, I was (and unfortunately still am) quite busy this week.

Just checking for the GCC version works for me, it makes the checks way easier and shorter.
I'll rework and force-push to my branch.

CMakeLists.txt
florianhofhammer: As the binutils version check will be removed in favor of the GCC version check alone, this style change is also gonna disappear in the process.

kent-mcleod (approved):

florianhofhammer:
Should (hopefully) have been the last force-push now.
The last push basically just fixed the code style failure in your checking tools.

Given how small the changes in the end turned out, thanks for the extensive discussion and sorry for probably wasting quite a lot of your time on that!

lsf37 (approved):

lsf37:
No worries, it's always nice when things become simpler!

florianhofhammer:
Another force push for rebasing onto master (which changed in the meantime).

PR #777, opened 2022-02-25
doc: update comment about mask

axel-h:
Commit 9f2d5219 made this 32/64-bit agnostic, but did not update the comment. Demystify the magic number mentioned there.

I wonder, who was/is 'swj' ?


lsf37 (approved):
Thanks for that. sjw is Simon Winwood, which means this is a very old comment. It's Ok to remove the `sjw ---` bit, we don't use that convention any more.

axel-h:
removed `sjw ---`

PR #778, opened 2022-02-25
trivial: Convert UTF-8 character to ASCII

lbmeng:
When building seL4 on a host whose default encoding is not UTF-8,
tools/bitfield_gen.py complains, as with the following log from
Python 3.6 on a Ubuntu 18.04 host says:

```
    File "tools/bitfield_gen.py",
  line 2773, in <module>
      string = f.read()
    File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
      return codecs.ascii_decode(input, self.errors)[0]
  UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 607543: ordinal not in range(128)
```

As the Python 3 doc for open() [1] says the default encoding is
platform dependent, such build error may happen on some hosts.
We can either updating tools/bitfield_gen.py to call open() with
an explicit encoding="utf-8" parameter, or avoiding UTF-8 characters
in the source codes.

After inspecting the two places in current source tree that use UTF-8
characters, none of them is absolutely necessary. Let's convert
them to ASCII characters.

[1] https://docs.python.org/3.6/library/functions.html#open

Signed-off-by: Bin Meng <bmeng.cn@gmail.com>

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -13,7 +13,7 @@
 #ifdef __ASSEMBLER__
 
 /* Provide a helper macro to define integer constants that are not of the
- * default type 'ìnt', but 'unsigned long [long]'. When such constants are
+ * default type 'int', but 'unsigned long [long]'. When such constants are
  * shared between assembly and C code, some assemblers will fail because they
  * don't support C-style integer suffixes like 'ul'. Using a macro works around
  * this, as the suffix is only applied when the C compiler is used and dropped
--- a/src/machine/io.c
+++ b/src/machine/io.c
@@ -5,7 +5,7 @@
  *
  * Portions derived from musl:
  *
- * Copyright © 2005-2020 Rich Felker, et al.
+ * Copyright (c) 2005-2020 Rich Felker, et al.
  *
  * SPDX-License-Identifier: MIT
  */


axel-h:
Thanks, seems the two places you have found are really better off with sticking to ASCII, so this should be merged.
On the general topic, I wonder if explicitly using `encoding="utf-8"` is something that might make sense to make the script more generic?

include/util.h
```diff
@@ -13,7 +13,7 @@
 #ifdef __ASSEMBLER__
 
 /* Provide a helper macro to define integer constants that are not of the
- * default type 'ìnt', but 'unsigned long [long]'. When such constants are
+ * default type 'int', but 'unsigned long [long]'. When such constants are
```
lsf37, line 16: This one is a typo and should indeed be fixed

src/machine/io.c
```diff
@@ -5,7 +5,7 @@
  *
  * Portions derived from musl:
  *
- * Copyright © 2005-2020 Rich Felker, et al.
+ * Copyright (c) 2005-2020 Rich Felker, et al.
```
lsf37, line 8: We should usually not touch external copyright headers. I think in this case it is acceptable.

lbmeng:
Thank you for the quick comments!

> On the general topic, I wonder if explicitly using `encoding="utf-8"` is something that might make sense to make the script more generic?

I agree. We can fix the 2 identified places for now but we can't forbid UTF-8 characters in the source. One of the UTF-8 scenarios we can't avoid is people's names which may exist on the copyright block of the file header.

I can send another PR to fix the script if everyone thinks this is a good idea.

lsf37 (commented):
I don't object to the two edits as such, but I'm confused why these changes would help with things in bitfield-gen. I would not think that these files are read by that code.

lsf37:
> > On the general topic, I wonder if explicitly using `encoding="utf-8"` is something that might make sense to make the script more generic?
> 
> I agree. We can fix the 2 identified places for now but we can't forbid UTF-8 characters in the source. One of the UTF-8 scenarios we can't avoid is people's names which may exist on the copyright block of the file header.
> 
> I can send another PR to fix the script if everyone thinks this is a good idea.

Yes, I agree, it would be better to explicitly specify utf-8 as encoding for files in this repo.

lbmeng:
> I don't object to the two edits as such, but I'm confused why these changes would help with things in bitfield-gen. I would not think that these files are read by that code.

bitfield-gen actually reads kernel_all_pp_prune.c whose contents may contain source snippets.

lbmeng:
gitlint in the CI reports:

`13: B1 Line exceeds max length (104>72): "  UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 607543: ordinal not in range(128)"`

I did not insert a LR/CR in the commit message as I wanted to keep the original error message from Python.
Let me know if I should fix this in the commit message.

lsf37:
> > I don't object to the two edits as such, but I'm confused why these changes would help with things in bitfield-gen. I would not think that these files are read by that code.
> 
> bitfield-gen actually reads kernel_all_pp_prune.c whose contents may contain source snippets.

Right, of course, that makes perfect sense.

lsf37:
> gitlint in the CI reports:
> 
> `13: B1 Line exceeds max length (104>72): " UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 607543: ordinal not in range(128)"`
> 
> I did not insert a LF/CR in the commit message as I wanted to keep the original error message from Python. Let me know if I should fix this in the commit message.

Would prefer a LF in this case, since the message is still clear afterwards.

lsf37 (approved):

lbmeng:
> Would prefer a LF in this case, since the message is still clear afterwards.

Thanks. I pushed the updated commit.

lsf37:
Thanks for the contribution!

PR #780, opened 2022-02-25
tools/bitfield_gen.py: Ensure UTF-8 encoding for the open() call

lbmeng:
There are cases to have UTF-8 characters in the source codes, hence
tools/bitfield_gen.py must be able to handle them correctly. Let's
explicitly pass encoding="utf-8" to the open() call.

Signed-off-by: Bin Meng <bmeng.cn@gmail.com>

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -2768,7 +2768,7 @@ if __name__ == '__main__':
 
         pruned_names = set()
         for filename in options.prune_files:
-            f = open(filename)
+            f = open(filename, encoding="utf-8")
             string = f.read()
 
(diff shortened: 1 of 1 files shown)


axel-h:
Follow up from PR https://github.com/seL4/seL4/pull/778

lsf37:
We should probably specify utf-8 encoding for all input files (incl `.bf`), not only the ones in `options.prune_files`. I haven't looked through the source yet if there are other points in this file, but I suspect there are.

lsf37:
I wasn't able to rebase the branch, so I've opened another PR with slight updates in #866.

PR #782, opened 2022-02-26
use unreachable instead of returning something

axel-h:
The loop is always left by a direct return from the body, any code after the loop is not reachable.

### DIFF
--- a/src/kernel/cspace.c
+++ b/src/kernel/cspace.c
@@ -191,6 +191,5 @@ resolveAddressBits_ret_t resolveAddressBits(cap_t nodeCap, cptr_t capptr, word_t
         }
     }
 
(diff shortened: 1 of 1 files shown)


axel-h:
Looks like the proof are fine with this change. Is there any other reason to to keep this?

kent-mcleod:
> I don't see a strong need to change it, but it is generally better to document unreachable points, so I'm up for it.

Not changing it means avoiding introducing a potential issue for BV though. I asked for review from @mbrcknl

mbrcknl (approved):
This should be fine for binary verification. (Symbolic execution won't be able to find a path to the unreachable code, so it doesn't matter to BV what comes after loop.)

PR #783, opened 2022-02-26
make RISC-V parameter handling more generic

axel-h:
Making things a bit more generic is supposed to make using various RISC-V extensions easier. This also remove some redundancies also. Seem on rv32 we do not support the FPU at the moment because nobody has tested this?
 
This is supposed to make integration of https://github.com/seL4/seL4/pull/776 a bit easier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,17 +85,46 @@ elseif(KernelArchARM)
     endif()
 
 elseif(KernelArchRiscV)
-    if(KernelSel4ArchRiscV64)
-        if(KernelHaveFPU)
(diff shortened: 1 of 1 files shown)


axel-h:
Rebased and added a commit body that mentions this change adds supports for using the F-extension without the D-extension now also. See also discussion at https://github.com/seL4/seL4/pull/773

lsf37 (approved):
Happy with this, just want to hold off merging until I see the test on master succeeding. There was a build failure that looked suspect.

lsf37:
> Happy with this, just want to hold off merging until I see the test on master succeeding. There was a build failure that looked suspect.

Ok, that seems to have been a network fluke. Will merge this one now.

PR #784, opened 2022-02-26
benchmark/track: avoid ambiguous values

axel-h:
- Using unique values across all architectures avoids having ambiguous values. This allows adding more values without creating more ambiguity, which will become confusion for generic userland tools.
- The value for `Entry_VMExit` changes.
- Print entry path ID in case of unknown values.

The ambiguity has been introduced in https://github.com/seL4/seL4/commit/fc0f1eec76fbf623e34aeb0f774c777c3690223e, this also introduced `Entry_VMExit`. Unfortunately, there is no explanation if this was intended for some reason or if this was just done because the values are used on one architecture only.


pingerino:
I do not think this PR should be merged unless the values are actually used across more than one architecture.

PR #787, opened 2022-02-27
aarch64: Use asid_map_t bitfield to store vspace attributes to free mapping slots

kent-mcleod:
Use the asid_map bitfield from x86 for aarch64 to map from ASID to an intermediate asid_map object that can hold a vspace object reference plus additional related attributes.  This can free up reserved slots in the top level page table for use at user level. 

This PR uses some commits from https://github.com/seL4/seL4/pull/579 but is only targeting aarch64.

More background: https://sel4.atlassian.net/browse/SELFOUR-1840

### DIFF
--- a/include/arch/arm/arch/64/mode/kernel/vspace.h
+++ b/include/arch/arm/arch/64/mode/kernel/vspace.h
@@ -73,13 +73,11 @@ static const region_t BOOT_RODATA *mode_reserved_region = NULL;
 #define cap_vtable_root_ptr_set_mappedCB(_c, cb) \
     cap_page_upper_directory_cap_ptr_set_capPUDMappedCB(_c, cb)
 #define cap_vtable_cap_new(_a, _v, _m) cap_page_upper_directory_cap_new(_a, _v, _m, 0, CB_INVALID)
-#define vtable_invalid_new(_a, _v) pude_pude_invalid_new(_a, _v, 0)
-#define vtable_invalid_smmu_new(_cb) pude_pude_invalid_new(0, false, _cb)
+#define vtable_invalid_smmu_new(_cb) pude_pude_invalid_new(_cb)
 #define vtable_invalid_get_bind_cb(_v) \
     pude_pude_invalid_get_bind_cb(_v)
 #else
 #define cap_vtable_cap_new(_a, _v, _m) cap_page_upper_directory_cap_new(_a, _v, _m, 0)
-#define vtable_invalid_new(_a, _v) pude_pude_invalid_new(_a, _v)
 #endif  /*!CONFIG_ARM_SMMU*/
 
 #define vtable_invalid_get_stored_asid_valid(_v) \
@@ -90,7 +88,11 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 {
     cap_page_upper_directory_cap_ptr_set_capPUDMappedASID(&cte->cap, asid);
     cap_page_upper_directory_cap_ptr_set_capPUDIsMapped(&cte->cap, 1);
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap), 0, false);
+#else
     asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap));
+#endif
     poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
 #ifdef CONFIG_ARM_SMMU
     vspace_root_t *vtable = (vspace_root_t *)cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap);
@@ -116,14 +118,12 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
     cap_page_global_directory_cap_ptr_set_capPGDMappedCB(_c, cb)
 #define cap_vtable_cap_new(_a, _v, _m) \
     cap_page_global_directory_cap_new(_a, _v, _m, CB_INVALID)
-#define vtable_invalid_new(_a, _v) pgde_pgde_invalid_new(_a, _v, 0)
-#define vtable_invalid_smmu_new(_cb) pgde_pgde_invalid_new(0, false, _cb)
+#define vtable_invalid_smmu_new(_cb) pgde_pgde_invalid_new(_cb)
 #define vtable_invalid_get_bind_cb(_v) \
     pgde_pgde_invalid_get_bind_cb(_v)
 #else
 #define cap_vtable_cap_new(_a, _v, _m) \
     cap_page_global_directory_cap_new(_a, _v, _m)
-#define vtable_invalid_new(_a, _v) pgde_pgde_invalid_new(_a, _v)
 #endif /*!CONFIG_ARM_SMMU*/
 
 #define vtable_invalid_get_stored_asid_valid(_v) \
@@ -134,7 +134,11 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 {
     cap_page_global_directory_cap_ptr_set_capPGDMappedASID(&cte->cap, asid);
     cap_page_global_directory_cap_ptr_set_capPGDIsMapped(&cte->cap, 1);
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap), 0, false);
+#else
     asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap));
+#endif
     poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
 
 #ifdef CONFIG_ARM_SMMU
--- a/include/arch/arm/arch/64/mode/kernel/vspace.h
+++ b/include/arch/arm/arch/64/mode/kernel/vspace.h
@@ -14,22 +14,6 @@
 #define activate_global_pd activate_kernel_vspace
 #define MODE_RESERVED 0
 
-/* The VTABLE_VMID_SLOT in user-level applications's vspace root
- * is reserved for storing its allocated hardware 8-bit VMID
- * when running EL2. Note that this assumes that the IPA size for S2
- * translation and the VA size for the S1 translation do not use full
- * 48-bit. Please see the definition of seL4_UserTop for details.
- */
-#define VTABLE_VMID_SLOT   MASK(seL4_VSpaceIndexBits)
-
-/* The VTABLE_SMMU_SLOT in user-level applications's vspace root is reserved
- * for storing the number of context banks bound with this vspace when the
- * SMMU feature is enabled. This assumes the user-level address space do not
- * use the second last entry in the vspace root, which is preserved by the
- * seL4_UserTop.
- */
-#define VTABLE_SMMU_SLOT   (MASK(seL4_VSpaceIndexBits) - 1)
-
 /* ==================== BOOT CODE FINISHES HERE ==================== */
 
 bool_t CONST isVTableRoot(cap_t cap);
--- a/include/arch/arm/arch/64/mode/kernel/vspace.h
+++ b/include/arch/arm/arch/64/mode/kernel/vspace.h
@@ -49,6 +49,8 @@ void deleteASID(asid_t asid, vspace_root_t *vspace);
 hw_asid_t getHWASID(asid_t asid);
 #endif
 
+asid_map_t findMapForASID(asid_t asid);
+
 #ifdef __clang__
 static const region_t BOOT_RODATA mode_reserved_region[] = {};
 #else
@@ -88,10 +90,10 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 {
     cap_page_upper_directory_cap_ptr_set_capPUDMappedASID(&cte->cap, asid);
     cap_page_upper_directory_cap_ptr_set_capPUDIsMapped(&cte->cap, 1);
-    poolPtr->array[asid & MASK(asidLowBits)] =
-        PUDE_PTR(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap));
+    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap));
+    poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
 #ifdef CONFIG_ARM_SMMU
-    vspace_root_t *vtable = poolPtr->array[asid & MASK(asidLowBits)];
+    vspace_root_t *vtable = (vspace_root_t *)cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap);
     vtable[VTABLE_SMMU_SLOT] = vtable_invalid_smmu_new(0);
 #endif
     return EXCEPTION_NONE;
@@ -132,11 +134,11 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 {
     cap_page_global_directory_cap_ptr_set_capPGDMappedASID(&cte->cap, asid);
     cap_page_global_directory_cap_ptr_set_capPGDIsMapped(&cte->cap, 1);
-    poolPtr->array[asid & MASK(asidLowBits)] =
-        PGDE_PTR(cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap));
+    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap));
+    poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
 
 #ifdef CONFIG_ARM_SMMU
-    vspace_root_t *vtable = poolPtr->array[asid & MASK(asidLowBits)];
+    vspace_root_t *vtable = (vspace_root_t *)cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap);
     vtable[VTABLE_SMMU_SLOT] = vtable_invalid_smmu_new(0);
 #endif
     return EXCEPTION_NONE;
--- a/include/arch/arm/arch/64/mode/kernel/vspace.h
+++ b/include/arch/arm/arch/64/mode/kernel/vspace.h
@@ -73,9 +73,6 @@ static const region_t BOOT_RODATA *mode_reserved_region = NULL;
 #define cap_vtable_root_ptr_set_mappedCB(_c, cb) \
     cap_page_upper_directory_cap_ptr_set_capPUDMappedCB(_c, cb)
 #define cap_vtable_cap_new(_a, _v, _m) cap_page_upper_directory_cap_new(_a, _v, _m, 0, CB_INVALID)
-#define vtable_invalid_smmu_new(_cb) pude_pude_invalid_new(_cb)
-#define vtable_invalid_get_bind_cb(_v) \
-    pude_pude_invalid_get_bind_cb(_v)
 #else
 #define cap_vtable_cap_new(_a, _v, _m) cap_page_upper_directory_cap_new(_a, _v, _m, 0)
 #endif  /*!CONFIG_ARM_SMMU*/
@@ -88,16 +85,19 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 {
     cap_page_upper_directory_cap_ptr_set_capPUDMappedASID(&cte->cap, asid);
     cap_page_upper_directory_cap_ptr_set_capPUDIsMapped(&cte->cap, 1);
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap), 0, false);
-#else
-    asid_map_t asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap));
-#endif
-    poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
+    asid_map_t asid_map = asid_map_asid_map_vspace_new(
 #ifdef CONFIG_ARM_SMMU
-    vspace_root_t *vtable = (vspace_root_t *)cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap);
-    vtable[VTABLE_SMMU_SLOT] = vtable_invalid_smmu_new(0);
+                              /* bind_cb: Number of bound context banks */
+                              0,
 #endif
+                              /* vspace_root: reference to vspace root page table object */
+                              cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap)
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+                              /* stored_hw_vmid, stored_vmid_valid: Assigned hardware VMID for TLB. */
+                              , 0, false
+#endif
+                          );
+    poolPtr->array[asid & MASK(asidLowBits)] = asid_map;
     return EXCEPTION_NONE;
 }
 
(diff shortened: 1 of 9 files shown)


include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -65,8 +65,8 @@ block page_upper_directory_cap {
     field capPUDIsMapped             1
     field_high capPUDMappedAddress   10
 #if defined (CONFIG_ARM_SMMU)  && defined (AARCH64_VSPACE_S2_START_L1)
-    field capPUDMappedCB             12
-    padding                          36
+    field capPUDMappedCB             8
+    padding                          40
```
lsf37, line 69: This commit also changes the width of the CB field in the caps. Was that intentional? If yes, should probably be mentioned in the commit message. I haven't checked all other `cb` occurrences if the new width fits or may lead to overflow.

src/fastpath/fastpath.c
```diff
@@ -99,7 +99,23 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
...
+    asid_t asid = cap_vtable_root_get_mappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
+    if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
+                 VSPACE_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {
+        slowpath(SysCall);
+    }
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* Ensure the HWASID is valid. */
+    if (unlikely(!asid_map_asid_map_vspace_get_stored_vmid_valid(asid_map))) {
+        slowpath(SysCall);
```
lsf37, line 112:
```suggestion
    /* Ensure the vmid is valid. */
    if (unlikely(!asid_map_asid_map_vspace_get_stored_vmid_valid(asid_map))) {
        slowpath(SysCall);
```

lsf37: Given the difference in naming in the other configs (`hw_asid`), might need a comment that `vmid`s are the tags used in hyp mode.

src/fastpath/fastpath.c
```diff
@@ -352,7 +368,23 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
-    stored_hw_asid.words[0] = cap_vtable_root_get_mappedASID(newVTable);
+    /* Need to test that the ASID is still valid */
+    asid_t asid = cap_vtable_root_get_mappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
+    if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
+                 VSPACE_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {
+        slowpath(SysReplyRecv);
+    }
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* Ensure the HWASID is valid. */
```
lsf37, line 379:
```suggestion
    /* Ensure the vmid is valid. */
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -551,41 +551,51 @@ BOOT_CODE void activate_kernel_vspace(void)
...
 
 /* ==================== BOOT CODE FINISHES HERE ==================== */
 
-static findVSpaceForASID_ret_t findVSpaceForASID(asid_t asid)
+asid_map_t findMapForASID(asid_t asid)
 {
-    findVSpaceForASID_ret_t ret;
-    asid_pool_t *poolPtr;
-    vspace_root_t *vspace_root;
+    asid_pool_t        *poolPtr;
```
lsf37, line 574:
```suggestion
    asid_pool_t *poolPtr;
```

(at least as far as I can see in the diff, there are no other declarations to align with any more)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1071,45 +1081,36 @@ pude_t *pageDirectoryMapped(asid_t asid, vptr_t vaddr, pde_t *pd)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 
-static void invalidateASID(asid_t asid)
+static asid_map_t *findMapRefForASID(asid_t asid)
 {
-    asid_pool_t *asidPool;
+    asid_pool_t        *poolPtr;
```
lsf37, line 1086: is the additional indent intentional?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1071,45 +1081,36 @@ pude_t *pageDirectoryMapped(asid_t asid, vptr_t vaddr, pde_t *pd)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 
-static void invalidateASID(asid_t asid)
+static asid_map_t *findMapRefForASID(asid_t asid)
```
lsf37, line 1084: There is is this convention that the functions that expect things to maybe not be there are called `find` and return a potential failure. This one is more a `get`, because it asserts the presence of the ASID pool. I currently have the same `find` name in the prelim spec, based on Curtis' original commit and probably going back to x64, and it makes sense to mimic `findMapForASID`, but overall I think `getMapRefForASID` would be a better name.

lsf37: Actually, I lie, I don't have this function at all in the spec -- I need to unfold this in Haskell because there is no direct ref to a position in an array there. Anyway, still prefer the `get` name :-)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1175,66 +1176,63 @@ static void invalidateASIDEntry(asid_t asid)
...
 
-    vspace_root_t *vtable = asidPool->array[asid & MASK(asidLowBits)];
-    assert(vtable);
+    asid_map_t *asid_map = &asidPool->array[asid & MASK(asidLowBits)];
+    assert(asid_map_ptr_get_type(asid_map) == asid_map_asid_map_vspace);
 
-    stored_info = vtable[VTABLE_SMMU_SLOT];
-    vtable[VTABLE_SMMU_SLOT] = vtable_invalid_smmu_new(vtable_invalid_get_bind_cb(stored_info) + 1);
+    asid_map_asid_map_vspace_ptr_set_bind_cb(asid_map, asid_map_asid_map_vspace_ptr_get_bind_cb(asid_map) + 1);
+    return;
```
lsf37, line 1203:
```suggestion
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1175,66 +1176,63 @@ static void invalidateASIDEntry(asid_t asid)
...
 }
```
lsf37, line 1204:
```suggestion
    asid_map_asid_map_vspace_ptr_set_bind_cb(asid_map, asid_map_asid_map_vspace_ptr_get_bind_cb(asid_map) + 1);
}
```

src/arch/arm/64/kernel/vspace.c
kent-mcleod: Ok, get makes more sense.

src/arch/arm/64/kernel/vspace.c
kent-mcleod: It came in from the cherry-pick. I assumed the style tool would fix it but it seems not.

src/fastpath/fastpath.c
kent-mcleod: I went with the vmid naming because of the discussion on the draft PR, but to be more complete, all the other function names should probably be changed too...

src/arch/arm/64/kernel/vspace.c
```diff
@@ -2406,7 +2405,8 @@ exception_t decodeARMMMUInvocation(word_t invLabel, word_t length, cptr_t cptr,
 
         /* Find first free ASID */
         asid = cap_asid_pool_cap_get_capASIDBase(cap);
-        for (i = 0; i < (1 << asidLowBits) && (asid + i == 0 || pool->array[i]); i++);
+        for (i = 0; i < (1 << asidLowBits) && (asid + i == 0
+                                               || (asid_map_get_type(pool->array[i]) != asid_map_asid_map_none)); i++);
```
lsf37, line 2409: If for some reason the previous shorter version is preferred, I think it would be Ok to leave it, because it should be equivalent. My guess is that `-O2` will reduce it to the same assembly anyway, though, and this is slightly clearer.

src/fastpath/fastpath.c
lsf37: Yes, I agree, was just about to write that in a summary comment. Might make sense do that rename in a separate commit/PR.

lsf37 (approved):
Very nice, a whole lot of stuff is becoming clearer and most parts shorter.

src/fastpath/fastpath.c
lsf37: Probably even the type `hw_asid_t`, if that doesn't conflict with code that is shared between configs (might be a problem in the fastpath for instance).

lsf37:
Cool, I very much like this. Would be good to have another opinion from a systems person, maybe @yyshen if you have time?

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -239,19 +239,40 @@ block vm_attributes {
...
-    padding                         41
+    field bind_cb                   8
+    padding                         8
+#else
+    padding                         16
+#endif
+    field_high vspace_root          36
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    padding                         2
+    field stored_hw_vmid            8
```
yyshen, line 258: recent processors support 16-bit vmid, so we could just make it 16b-bit.

include/arch/arm/arch/64/mode/object/structures.bf
kent-mcleod: Intentional. According to the other PR and the SMMUv2 spec, there is a maximum number of context banks of 128 which requires 8 bits to include a valid bit. I tried to update the width in all of the other locations.  The asid_map_t is short of available bit space and I saw that there's already a push for changing them to 8bits in other PRs so figured to make this part of the change at the same time.

src/arch/arm/64/kernel/vspace.c
kent-mcleod: Yea, it came in with a cherry-picked commit and I assumed the style tool would fix it up.....

include/arch/arm/arch/64/mode/object/structures.bf
kent-mcleod: On a processor that supports 16bit VMID we would be able to just use the asid_t value which is also 16bits and not need a stored_hw_vmid slot.

yyshen:
> Cool, I very much like this. Would be good to have another opinion from a systems person, maybe @yyshen if you have time?

@lsf37 How did you do the review based on github? Sometimes I found it is hard to understand the changes by looking at the diffs on Github. Any suggestions?

include/arch/arm/arch/64/mode/object/structures.bf
yyshen: right, if we have enough hardware asid, we do not need the mapping. where do we specify that?

kent-mcleod:
> @lsf37 How did you do the review based on github? Sometimes I found it is hard to understand the changes by looking at the diffs on Github. Any suggestions?

Do you browse  the diffs commit by commit? I think that's the clearest you can get using the web client. If that's not good enough for me I manually clone and step through diffs locally.  Onetime I tried using GitHub's desktop app which was a bit better for quickly stepping through files and commits for a PR but you need more setup...

lsf37:
> Do you browse the diffs commit by commit? I think that's the clearest you can get using the web client. If that's not good enough for me I manually clone and step through diffs locally. Onetime I tried using GitHub's desktop app which was a bit better for quickly stepping through files and commits for a PR but you need more setup...

Yeah, I do all of these for large PRs, although the hurdle is pretty high for the GH desktop app. I find the diff view in VSCode pretty good (not always, but on average). For this specific one, I got by with looking at the single commits and long staring at the diff. It helps that I wrote a spec for this separately, so it was easy for me to recognise patterns.

yyshen:
> > Do you browse the diffs commit by commit? I think that's the clearest you can get using the web client. If that's not good enough for me I manually clone and step through diffs locally. Onetime I tried using GitHub's desktop app which was a bit better for quickly stepping through files and commits for a PR but you need more setup...
> 
> Yeah, I do all of these for large PRs, although the hurdle is pretty high for the GH desktop app. I find the diff view in VSCode pretty good (not always, but on average). For this specific one, I got by with looking at the single commits and long staring at the diff. It helps that I wrote a spec for this separately, so it was easy for me to recognise patterns.



> > Do you browse the diffs commit by commit? I think that's the clearest you can get using the web client. If that's not good enough for me I manually clone and step through diffs locally. Onetime I tried using GitHub's desktop app which was a bit better for quickly stepping through files and commits for a PR but you need more setup...
> 
> Yeah, I do all of these for large PRs, although the hurdle is pretty high for the GH desktop app. I find the diff view in VSCode pretty good (not always, but on average). For this specific one, I got by with looking at the single commits and long staring at the diff. It helps that I wrote a spec for this separately, so it was easy for me to recognise patterns.

Thanks for the tips. I found the split view works better for me.

src/arch/arm/64/kernel/vspace.c
lsf37: Yes, you're right, that doesn't really work. I tend to ignore C's type system because it doesn't guarantee anything :-) I was only looking at the representation, which hasn't changed for the False case.

include/arch/arm/arch/64/mode/object/structures.bf
lsf37: Ok, that makes sense. Would be good to mention in the commit message. I have checked the other occurrences, and at least the types are safe everywhere, so this is good from my side.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1404,13 +1422,17 @@ void deleteASID(asid_t asid, vspace_root_t *vspace)
...
-    if (poolPtr != NULL && poolPtr->array[asid & MASK(asidLowBits)] == vspace) {
-        invalidateTLBByASID(asid);
+    if (poolPtr != NULL) {
+        asid_map_t asid_map = poolPtr->array[asid & MASK(asidLowBits)];
+        if (asid_map_get_type(asid_map) == asid_map_asid_map_vspace &&
+            (vspace_root_t *)asid_map_asid_map_vspace_get_vspace_root(asid_map) == vspace) {
+            invalidateTLBByASID(asid);
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-        invalidateASIDEntry(asid);
+            invalidateASIDEntry(asid);
```
yyshen, line 1407: additional spaces?

include/arch/arm/arch/64/mode/kernel/vspace.h
```diff
@@ -91,7 +89,11 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
     asid_map_t asid_map;
     cap_page_upper_directory_cap_ptr_set_capPUDMappedASID(&cte->cap, asid);
     cap_page_upper_directory_cap_ptr_set_capPUDIsMapped(&cte->cap, 1);
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    asid_map = asid_map_asid_map_vspace_new(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap), 0, false);
```
yyshen: the change is mostly for arm64 hyp right? for the cores with 8-bit VMID. So the commit message could point out the VMID explicitly I guess.

src/fastpath/fastpath.c
```diff
@@ -99,7 +99,23 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
     stored_hw_asid.words[0] = 0;
 #endif
 #ifdef CONFIG_ARCH_AARCH64
-    stored_hw_asid.words[0] = cap_vtable_root_get_mappedASID(newVTable);
+    /* Need to test that the ASID is still valid */
+    asid_t asid = cap_vtable_root_get_mappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
```
yyshen, line 104: do we know how many cycles are added by this findMapForASID in fastpath?

include/arch/arm/arch/64/mode/object/structures.bf
yyshen: CONFIG_ARM_SMMU seems implying v2. We probably will need to support V3 soon.

yyshen (approved):
looks good to me. Thanks!

src/arch/arm/64/kernel/vspace.c
lsf37: it's now under an additional `if` so the new indent looks correct to me

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1183,55 +1186,43 @@ static vspace_root_t getASIDBindCB(asid_t asid)
...
-    assert(asid_map_get_type(asid_map) == asid_map_asid_map_vspace);
+    asid_map_t *asid_map = &asidPool->array[asid & MASK(asidLowBits)];
+    assert(asid_map_ptr_get_type(asid_map) == asid_map_asid_map_vspace);
 
-    vspace_root_t *vtable = (vspace_root_t *)asid_map_asid_map_vspace_get_vspace_root(asid_map);;
-    assert(vtable);
-
-    stored_info = vtable[VTABLE_SMMU_SLOT];
-    vtable[VTABLE_SMMU_SLOT] = vtable_invalid_smmu_new(vtable_invalid_get_bind_cb(stored_info) + 1);
+    asid_map_asid_map_vspace_ptr_set_bind_cb(asid_map, asid_map_asid_map_vspace_ptr_get_bind_cb(asid_map) + 1);
```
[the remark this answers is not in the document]
lsf37: Good point. This is independent of this PR, but I think there is a bug in this path, yes, because the decode path does not check that the increment is safe (if assertions are turned on, this would fail a bf generator assertion I think; if they are turned off it would wrap, which is not clear that we want here).

The check should be in decode, though, not here, because in here it is too late to do anything about it. It's actually unclear to me how this is supposed to work when the CB space is used up.

We should create an issue for this.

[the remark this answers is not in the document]
kent-mcleod: There is a maximum number of CBs allowed in an SMMU implementation defined by the spec to be 128 and CBs can only be assigned to a single VSpace. 8 bits was chosen to be large enough to support every CB being assigned to the same VSpace on any SMMUv2 implementation so it shouldn't be able to overflow due to existing checks in `decodeARMCBInvocation` and `decodeARMCBControlInvocation`.

src/fastpath/fastpath.c
kent-mcleod: There's a few changes here that affect the fastpath, so I'll try and break them down.

### Effect of asid_map_t and extra asid_t check on aarch64 non-hyp

aarch64 non-hyp was missing a check for the ASID validity. This means that if an asid_pool_t containing the target thread's VSpace was revoked and then a fastpath was taken the VSpace of the target thread would be switched to even though the VSpace would not have a valid ASID.  The correct behavior would be for an empty table to be assigned to the TTBR as part of the context switch which is what falling back to the slowpath achieves.

For this change I measured a 1 cycle increase in the published numbers (https://sel4.systems/About/Performance/) for default IPC call and reply on the TX1 and TX2. The program should be doing 2 extra memory reads and 2 extra comparisons. I didn't check any event counters yet though and am assuming that the CPU is just making the right magical optimizations and this isn't adding any new stalls.

Any cost increase should be acceptable here as it's adding a check that should be required.

### Effect of asid_map_t and extra asid_t check on aarch64 hyp

aarch64 non-hyp was also missing an asid validity check and would finally call `getHWASID` as part of the context switch. If the asid wasn't valid here then the kernel may dereference a null pointer. As part of these changes, `getHWASID` was removed from the fastpath and replacing it with the changes in the PR the fastpath ends up a few % faster.  I don't have the numbers written down, but can update before final merge. 

### Effect of asid_map_t and removing ASID from page table entry on aarch32 sabre

These changes aren't part of this PR, but I spent some time evaluating an aarch32 implementation assuming it would be most sensitive to costs of this change.  On the sabre, a memory look up is removed as the hw_asid is taken out of the reserved page table entry, but 2 lookups and a conditional branch are added for the ASID table lookups. Published numbers go from 317 cycles (call)  and 336 cycles (recv) on Sabre to 333 and 350 now.  So 16 and 14 more cycles which is 4-5% overhead.

include/arch/arm/arch/64/mode/kernel/vspace.h
kent-mcleod: The change is for aarch64 hyp and also SMMUv2 and eventually kernel image clone. Longer term there is a plan that every architecture will have the asid_map_t feature if it wants to support the kernel image clone feature (a vspace can have a different kernel image to avoid shared cache timings).

include/arch/arm/arch/64/mode/object/structures.bf
kent-mcleod: At the moment all of the supported platforms only have 8bit VMID. If there's a platform that supports 16bit VMID in the future then the kernel would get a config update to make this feature `CONFIG_ARM_HYPERVISOR_SUPPORT && CONFIG_8BIT_VMID`

include/arch/arm/arch/64/mode/kernel/vspace.h
```diff
@@ -86,14 +67,22 @@ static const region_t BOOT_RODATA *mode_reserved_region = NULL;
 
 static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *poolPtr, cte_t *cte)
 {
+    asid_map_t asid_map;
     cap_page_upper_directory_cap_ptr_set_capPUDMappedASID(&cte->cap, asid);
     cap_page_upper_directory_cap_ptr_set_capPUDIsMapped(&cte->cap, 1);
-    poolPtr->array[asid & MASK(asidLowBits)] =
-        PUDE_PTR(cap_page_upper_directory_cap_get_capPUDBasePtr(cte->cap));
+    asid_map = asid_map_asid_map_vspace_new(
```
axel-h, line 73: nitpicking: go for c99?
```suggestion
    asid_map_t asid_map = asid_map_asid_map_vspace_new(
```

include/arch/arm/arch/64/mode/kernel/vspace.h
```diff
@@ -130,15 +114,24 @@ static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *po
 
 static inline exception_t performASIDPoolInvocation(asid_t asid, asid_pool_t *poolPtr, cte_t *cte)
 {
+    asid_map_t asid_map;
     cap_page_global_directory_cap_ptr_set_capPGDMappedASID(&cte->cap, asid);
     cap_page_global_directory_cap_ptr_set_capPGDIsMapped(&cte->cap, 1);
-    poolPtr->array[asid & MASK(asidLowBits)] =
-        PGDE_PTR(cap_page_global_directory_cap_get_capPGDBasePtr(cte->cap));
-
+    asid_map = asid_map_asid_map_vspace_new(
```
axel-h, line 120: nitpicking: go for c99?
```suggestion
    asid_map_t asid_map = asid_map_asid_map_vspace_new(
```

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
```diff
@@ -273,10 +265,7 @@ SEL4_SIZE_SANITY(seL4_PUDEntryBits, seL4_PUDIndexBits, seL4_PUDBits);
 #if defined(CONFIG_ARM_PA_SIZE_BITS_44)
 #define seL4_UserTop 0x00000fffffffffff
 #elif defined(CONFIG_ARM_PA_SIZE_BITS_40)
-/* Currently other definitions of seL4_UserTop already have free slots at the end and don't need to subtract for seL4_VSpaceReservedSlots.
- * Each slot used requires subtracting 1GiB from the top address.
- */
-#define seL4_UserTop (0x000000ffffffffff - seL4_VSpaceReservedSlots * 0x40000000)
+#define seL4_UserTop 0x000000ffffffffff
```
axel-h, line 268: Could go for a more intuitive 
```suggestion
#define seL4_UserTop ((1 << 40) - 1)
```
now. Not sure of `(LIBSEL4_BIT(40) - 1)` also work, as this enforced 'ul' then.

kent-mcleod: Using a plain old constant is what all other definitions use.

lsf37: Would agree that it makes more sense to stay consistent with the rest here, although I do generally appreciate the form that tells me where the constant is coming from. It's reasonably easy to see here, though, so not a big deal.

src/arch/arm/64/kernel/vspace.c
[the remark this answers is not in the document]
lsf37: In any case it is a separate issue, happy to leave it for another PR or an issue (still need to double check that decode path -- the first time I looked I didn't see a check that would constrain the overflow)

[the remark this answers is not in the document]
kent-mcleod: It's constrained by the upper bound on number of CBs that can exist in the system which is the number of slots in the static CB table: `extern bool_t smmuStateCBTable[SMMU_MAX_CB];`  SMMU_MAX_CB cannot be greater than 128.

kent-mcleod:
Just waiting for this to pass hw tests before merging

Issue #775, opened 2022-02-21
replace check for CONFIG_MAX_NUM_NODES > 1 with CONFIG_ENABLE_SMP_SUPPORT


axel-h:
Would it make sense to replace the checks `CONFIG_MAX_NUM_NODES > 1` with a check for `CONFIG_ENABLE_SMP_SUPPORT` instead - unless there is a strong reason for certain code parts to keep exactly this check?  That seem to be helpful in some cases (see e.g. https://github.com/seL4/seL4/issues/774#issuecomment-1046532618) and allows separating general SMP conditional parts completely from the number of configured nodes. 
Practically, running an SMP configuration with just one core might have a use case for testing purposes only. Or maybe for AMP use cases.

PR #789, opened 2022-02-28
Fix Issue774

### LINKED ISSUE #774: C-Preprocessor conditionals in XML
Before I jump too far down the rabbit hole on this, I wanted to sync with the community,
It was mentioned during one of the developer meetings that the xml currently contains some c-isms in the form of c-preprocessor declarations, 

Most of these are fairly simple, an example of the most complex being:
```
condition="!(defined(CONFIG_ARM_HYPERVISOR_SUPPORT) &amp;&amp; defined(CONFIG_ARM_PA_SIZE_BITS_40))"
```
in [sel4_arch_include/aarch64/interfaces/sel4arch.xml](https://github.com/seL4/seL4/blob/master/libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml#L212)

If it weren't for these couple of complex conditions we could use something simple like:
 `condition_var="CONFIG_..." condition_defined=true|false`

I tried to come up with a schema to cope with these definitions in a language neutral fashion:
#### Examples
```
<condition>
        <config var="CONFIG_FOO"/>
</condition>
```

```
<condition>
        <not>
         <and>
             <or>
                <config var="CONFIG_A"/>
                <config var="CONFIG_B"/>
             </or>
             <not><config var="CONFIG_C"/></not>
         </and>
        </not>
</condition>
```
#### Schema
The schema for the above xml uses an xsd (below), while the current encoding is using a [dtd](https://github.com/seL4/seL4/blob/master/libsel4/tools/sel4_idl.dtd#L25)

As far as I can tell trying to do this with just a dtd leads to a much more verbose xml encoding.

```
<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <xsd:element name="condition" type="Unary" />
    <xsd:group name="ConfigGroup">
        <xsd:choice>
            <xsd:element name="and" type="Binary" />
            <xsd:element name="or" type="Binary" />
            <xsd:element name="not" type="Unary" />
            <xsd:element name="config">
                <xsd:complexType>
                        <xsd:attribute name="var" type="xsd:string" use="required"/>
                </xsd:complexType>
            </xsd:element>
        </xsd:choice>
    </xsd:group>

  <xsd:complexType name="Unary">
        <xsd:group ref="ConfigGroup"/>
  </xsd:complexType>

  <xsd:complexType name="Binary">
    <xsd:sequence>
            <xsd:group ref="ConfigGroup" minOccurs="2" maxOccurs="2"/>
    </xsd:sequence>
  </xsd:complexType>
</xsd:schema>
```

#### Questions

* Should we include the `CONFIG_` contents of the `var` attribute?
* Does this need an RFC even though it isn't really changing any visible aspects or behavior, but just refining existing behavior?
* Is there any opposition to this or the approach taken here, or a better approach, or scheduling concerns if it could impact ongoing projects?

#### checklist

- [ ]  Switch the existing dtd to use an xsd
- [ ]  Move the `condition=` attributes to `condition` elements in the sel4 xml.
- [ ]  Write python code converting the new condition elements to c-preprocessor `#if` conditions.

ratmice:
This should be basically done and fixes #774, I need to check #775 further, but it does resolve the xml parts.

This probably needs some squashing and rewriting,
Because it adds the new element and removes the old attribute separate commits,
It includes a test script which compares the old attribute to the new generated pre-processor condition.
Then removes that script when the old attributes are removed.

I imagine you'll probably want to squash that all down.

While I used the test script to compare all the conditions in the XML,
I didn't actually go through building every configuration/architecture locally.


ratmice:
I have no idea how to fix the manual, seems like if I fix manual/libsel4_tools it breaks tools, and libsel4/tools and vice versa.
I'll have to try and figure it out tomorrow, unless someone more familiar with python might know.

canarysnort01:
In addition to the module import issue, it looks to me like `manual/tools/gen_invocations.py` is still trying to validate against the deleted DTD file.

ratmice:
Haven't figured out how to build the manual locally yet, so here's to hoping that that dtd validation also support an xsd. :crossed_fingers:

ratmice:
So despite some Vogonesque hacks to get the `manual` + style check working,
I feel like this is probably ready for review, I've tried to clean up the history into logical chunks,

Although I haven't compiled each patch in the series individually, I believe each commit should compile and not mess up a bisect.
But still feel free to squash or ask me to, if that is preferred.

lsf37:
First time I hear the word Vogonesque fitting naturally into a sentence. Kudos! :-) I hope to have time to review this later today.

libsel4/include/api/syscall.xsd
```diff
@@ -35,11 +35,34 @@
...
+        </xsd:choice>
+    </xsd:group>
+
+  <xsd:complexType name="Unary">
+        <xsd:group ref="ConfigGroup"/>
+  </xsd:complexType>
+
+  <xsd:complexType name="Binary">
+    <xsd:sequence>
+            <xsd:group ref="ConfigGroup" minOccurs="2" maxOccurs="2"/>
```
axel-h, line 62: Is it really necessary to limit `maxOccurs` to 2?

libsel4/tools/condition.py
```diff
@@ -0,0 +1 @@
+../../tools/condition.py
```
axel-h, line 1: What is this, so we really need this indirection and the ".."? Don't we have the search path somewhere?

libsel4/tools/loader/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+# Copyright 2022, seL4 Project a Series of LF Projects, LLC
+#
+# SPDX-License-Identifier: BSD-2-Clause or GPL-2.0-only
+
+import os
+import sys
+import pathlib
+
+# Add the parent module to the sys.path, do so in this __init__.py
+# So that the code producing the side-effect can occurr after import
```
axel-h, line 11:
```suggestion
# so that the code producing the side-effect can occur after import
```

libsel4/tools/loader/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+# statements.  Otherwise autopep8 would want to reformat this code
+# so the side-effect occurs after all import statements.
```
axel-h, line 13:
```suggestion
# so the side-effect occurs after all import statements.
```
Seems "occurred" is the only time where two 'r' are used.

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+        if child.nodeType == node.COMMENT_NODE or (child.nodeType == node.TEXT_NODE and len(child.data.strip()) == 0):
+            node.removeChild(child)
+
+
+def condition_to_cpp(conditions):
+    n = len(conditions)
+    assert n <= 1
+    if n == 0:
+        return ""
+    else:
```
axel-h, line 17: Since the if-block returns, we could save a level of indention if we drop the "else"

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+        remove_ws_comments(conditions[0])
+        children = conditions[0].childNodes
+        if children == None or len(children) == 0:
```
axel-h, line 20: Do we really have to check for `None` explicitly?
```suggestion
        if not children or len(children) == 0:
```

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+
+def remove_ws_comments(node):
+    for child in list(node.childNodes):
+        if child.nodeType == node.COMMENT_NODE or (child.nodeType == node.TEXT_NODE and len(child.data.strip()) == 0):
+            node.removeChild(child)
+
+
+def condition_to_cpp(conditions):
+    n = len(conditions)
+    assert n <= 1
```
axel-h, line 14: See https://github.com/seL4/seL4/pull/789/files#r821356538, we could support more then two also.

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+
+        def helper(expr):
+            remove_ws_comments(expr)
+            if expr.tagName == "config":
+                return "defined({})".format(expr.getAttribute("var"))
+            elif expr.tagName == "not":
+                child = expr.firstChild
+                if child.tagName == "and" or child.tagName == "or":
+                    return "!({})".format(helper(child))
+                else:
```
axel-h, line 33: What is `child.tagName` in this case? Could avoid unconditional else and reject anything unknown here to ensure this raises an exception if somebody every extends this mechanism and forgets to update this

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+                    return "!{}".format(helper(child))
+            elif expr.tagName == "and":
+                return "{} && {}".format(helper(expr.childNodes[0]), helper(expr.childNodes[1]))
```
axel-h, line 36: Something like `' && '.join([helper(e) for e in expr.childNodes])` would work for an arbitrary number of elements

tools/condition.py
```diff
@@ -0,0 +1,41 @@
...
+            elif expr.tagName == "or":
+                return "{} || {}".format(helper(expr.childNodes[0]), helper(expr.childNodes[1]))
+            else:
+                return ""
+        return helper(children[0])
```
axel-h, line 41: please add an empty line before this for readabilty

axel-h:
Nice work!

libsel4/tools/condition.py
ratmice: This is making a symlink to condition.py, I'm not sure why it shows the path as its contents.
I just went with how it seemed to be done in libsel4/tools, where the rest of these files are all symlinks.

libsel4/include/api/syscall.xsd
ratmice: Probably not, but there wasn't anything that used more than 2.
I'm usually pretty conservative about adding things that aren't in use (Note I added `<or>` though despite nothing using it).
I can definitely willing to change it though if that is the consensus.

tools/condition.py
axel-h: Ah sorry, I misinterpreted that. Maybe add a comment

tools/condition.py
ratmice: I'm going to split this bit into two patches, and do a follow up after the comparison to the existing conditional `attribute` which moves the parentheses here.

libsel4/include/api/syscall.xsd
ratmice: So, I went ahead and did this change as it may make the (currently non-existent rust output) easier so you don't have to encode and as `all(a, all(b, c))` in theory.

Thank you for the review Axel, I believe i've covered all of your points now.

axel-h (approved):
Look ok from my side.

libsel4/tools/condition.py
axel-h: Thanks for the explanation. If this just follows existing stuff, then it's fine. And symlink within the same repo will always work.

Improving the python usage would be a different PR indeed. You could create an issue to track this ;)

libsel4/tools/__init__.py
```diff
@@ -8,4 +8,6 @@
 # This file allows the the libsel4 tools directory to be imported as
 # a python module.
 
+from . import loader
```
kent-mcleod, line 11: Why is the new loader module needed?

libsel4/tools/syscall_stub_gen_rs.py
```diff
@@ -14,6 +14,7 @@
 from argparse import ArgumentParser
 import sys
 import syscall_stub_gen
+from condition import condition_to_cpp
```
kent-mcleod, line 17: I can't see `condition_to_cpp` being needed in this file?

kent-mcleod (commented):
Looks good to me! I don't think you should to squash the intermediate commits as it's more convincing that the migration is correct.

libsel4/tools/__init__.py
ratmice: I added a large comment explaining it

libsel4/tools/__init__.py
```diff
@@ -8,4 +8,27 @@
...
+# The following import of `loader` avoids a python module loading error
+# When this is used as a module in manual/tools/libsel4_tools.
+# Traceback (most recent call last):
+#  File "/path/to/syscall_stub_gen.py", line 46, in <module>
+#    from condition import condition_to_cpp
+# ModuleNotFoundError: No module named 'condition'
+#
+# Imports for modules differ from imports for scripts
+# Such that when run as a script in tools/ it includes the current
+# directory in the search path.  But not when imported as a module.
```
axel-h, line 20:
```suggestion
# directory in the search path.  But not when imported as a module
```

libsel4/tools/__init__.py
axel-h: In `syscall_stub_gen.py`, can't  we do 
```
sys.path.append(str(pathlib.Path(os.path.dirname(os.path.realpath(__file__))).parent))
```
to fix this?

libsel4/tools/__init__.py
```diff
@@ -8,4 +8,27 @@
...
+# `loader` will add the current directory to the search path.
```
axel-h, line 21:
```suggestion
# Importing `loader` will add the current directory to the search path.
```

libsel4/tools/__init__.py
ratmice: Presumably, that would work but I think it would still run into the autopep8 reordering problem.
We would also need to get rid of the `.parent` stuff.

axel-h: Can you create an issue for this, then it's kept track of?

ratmice: Sure, but prefer to do it after merge, so we don't end up with dangling issues.

lsf37 (approved):

ratmice:
Hmm, the timeout I presume was running `MULTICORE0005` seems suspicious, given the whole `CONFIG_MAX_NUM_NODES > 1` -> `CONFIG_ENABLE_SMP_SUPPORT` aspect of this change.

lsf37:
> Hmm, the timeout I presume was running `MULTICORE0005` seems suspicious, given the whole `CONFIG_MAX_NUM_NODES > 1` -> `CONFIG_ENABLE_SMP_SUPPORT` aspect of this change.

I doubt it's that, but let me re-run. This specific test is pretty flaky (marked as such in the test driver), and there is an [open issue](https://github.com/seL4/sel4test/issues/44) for it on sel4test.

lsf37:
> > Hmm, the timeout I presume was running `MULTICORE0005` seems suspicious, given the whole `CONFIG_MAX_NUM_NODES > 1` -> `CONFIG_ENABLE_SMP_SUPPORT` aspect of this change.
> 
> I doubt it's that, but let me re-run. This specific test is pretty flaky (marked as such in the test driver), and there is an [open issue](https://github.com/seL4/sel4test/issues/44) for it on sel4test.

Ok, re-run succeeded, it's definitely not a config option problem.

The xmllint install in the GitHub action is failing for some reason. It looked like a network problem, but maybe I'm wrong about that (if it is, then it is persistent). Still need to look into that.

ratmice:
> The xmllint install in the GitHub action is failing for some reason. It looked like a network problem, but maybe I'm wrong about that (if it is, then it is persistent). Still need to look into that.

Looking around it looks like that is a symptom of the apt package cache being out of date with regards to the latest,
this kind of makes sense in that the failure looks to be a security issue which would cause the deletion of a package.

Hopefully `apt-get update` fixes, and seems to be what others have done.  I didn't see any other `apt-get install` calls in other ci files here.

lsf37:
> > The xmllint install in the GitHub action is failing for some reason. It looked like a network problem, but maybe I'm wrong about that (if it is, then it is persistent). Still need to look into that.
> 
> Looking around it looks like that is a symptom of the apt package cache being out of date with regards to the latest, this kind of makes sense in that the failure looks to be a security issue which would cause the deletion of a package.

Yes, that makes perfect sense.

> Hopefully `apt-get update` fixes, and seems to be what others have done. 

Which indeed fixed it. Thanks!

libsel4/tools/__init__.py
ratmice: I filed #811 for this now.

PR #791, opened 2022-03-01
trivial: fix examples explaining IRQT_TO_IDX macro

Ivan-Velickovic:
This PR just fixes examples given in a comment for how an IRQ gets converted to a CNode index.

Example 2: The IRQ number is not a PPI so the index is `(irq.target_core)*NUM_PPI + (_irq.irq)`, which means it should be `+ 16` not `+ 12`.

Example 3: The IRQ number is a PPI so the index is `(CONFIG_MAX_NUM_NODES-1)*NUM_PPI + (_irq.irq)`, which means it should be `(3 * 32) + 33` since there's 4 cores, not sure where the `(33 - 32)` came from.

### DIFF
--- a/include/arch/arm/arch/machine/gic_common.h
+++ b/include/arch/arm/arch/machine/gic_common.h
@@ -37,7 +37,7 @@
  * The layout of IRQs into the CNode are all of PPI's for each core first, followed
  * by the global interrupts.  Examples:
  *   core: 0, irq: 12 -> index 12.
- *   core: 2, irq: 16 -> (2 * 32) + 12
+ *   core: 2, irq: 16 -> (2 * 32) + 16
  *   core: 1, irq: 33, (4 total cores) -> (4 * 32) + (33-32).
  */
 #define IRQ_IS_PPI(_irq) (HW_IRQ_IS_PPI(_irq.irq))


include/arch/arm/arch/machine/gic_common.h
```diff
@@ -37,8 +37,8 @@
  * The layout of IRQs into the CNode are all of PPI's for each core first, followed
  * by the global interrupts.  Examples:
  *   core: 0, irq: 12 -> index 12.
- *   core: 2, irq: 16 -> (2 * 32) + 12
- *   core: 1, irq: 33, (4 total cores) -> (4 * 32) + (33-32).
+ *   core: 2, irq: 16 -> (2 * 32) + 16
```
axel-h, line 40: You are right, this should be 16. Seems it's a copy/paste bug, form above.

include/arch/arm/arch/machine/gic_common.h
```diff
@@ -37,8 +37,8 @@
...
+ *   core: 1, irq: 33, (4 total cores) -> (3 * 32) + 33.
```
yyshen, line 41: The previous example is clear to me. 4 * 32 clearly expresses that the first 4*32 slots are for the PPI and SGI of the 4 cores.

axel-h, line 41: Sees the `(33-32)` comes from seeing it as `(CONFIG_MAX_NUM_NODES * NUM_PPI + (_irq.irq - NUM_PPI))`, but at some point this was simplified to `(CONFIG_MAX_NUM_NODES - 1) * NUM_PPI + (_irq.irq)` then in the code.

Ivan-Velickovic:
The third example is correct either way, so I've decided to leave it and just fix the mistake in the second example.

include/arch/arm/arch/machine/gic_common.h
kent-mcleod: Axel's right that the implementation was simplified.  The example is trying to explain the design and should be illustrating that the SPI interrupts start from 0 at the end of the PPI/SGI of all the cores.

kent-mcleod (approved):

lsf37 (approved):

PR #792, opened 2022-03-03
bitfield_gen: allow _ in integer literals

lsf37:
This allows optional ignored `_` at any position in any integer literals (dec, oct, hex, bin) for better readability. For example

`0xFFFF_FFFF_FFFF_FF00 = 0xFFFFFFFFFFFFFF00`
`0b0000_0000_0001_0000 = 0b0000000000010000`
`1_000_000 = 1000000`

This is a syntax addition only, no semantic changes.

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -108,8 +108,8 @@ def t_IDENTIFIER(t):
 
 
 def t_INTLIT(t):
-    r'([1-9][0-9]*|0[oO]?[0-7]+|0[xX][0-9a-fA-F]+|0[bB][01]+|0)[lL]?'
-    t.value = int(t.value, 0)
+    r'([1-9][0-9_]*|0[oO]?[0-7_]+|0[xX][0-9a-fA-F_]+|0[bB][01_]+|0)[lL]?'
+    t.value = int(t.value.replace('_', ''), 0)
     return t
 
 


lsf37:
In terms of concrete syntax, C doesn't have any such separators, C++ has `'` which I find terrible, and Rust has `_` which seems unambiguous and Ok. I have also seen `.`, which I find easier to read (if we exclude decimals), but it could be too confusing with a decimal point/fraction.

tools/bitfield_gen.py
```diff
@@ -108,8 +108,8 @@ def t_IDENTIFIER(t):
 
 
 def t_INTLIT(t):
-    r'([1-9][0-9]*|0[oO]?[0-7]+|0[xX][0-9a-fA-F]+|0[bB][01]+|0)[lL]?'
-    t.value = int(t.value, 0)
+    r'([1-9][0-9_]*|0[oO]?[0-7_]+|0[xX][0-9a-fA-F_]+|0[bB][01_]+|0)[lL]?'
```
axel-h, line 111: Maybe add a comment here
```
# This allows optional ignored _ at any position in any integer literalsexample
# 0xFFFF_FFFF_FFFF_FF00 = 0xFFFFFFFFFFFFFF00
# 0b0000_0000_0001_0000 = 0b0000000000010000
# 1_000_000 = 1000000
```

axel-h:
Cool, that is a nice feature. As a demonstration that this is working as expected, do you have a PR somewhere that uses this?

axel-h (approved):

lsf37:
> Cool, that is a nice feature. As a demonstration that this is working as expected, do you have a PR somewhere that uses this?

I did use it, but then I changed what I was doing and didn't need it any more for that :-)

I should have a look through the existing .bf files if there are areas where it would make sense to use directly.

kent-mcleod:
> I did use it, but then I changed what I was doing and didn't need it any more for that :-)
> 
> I should have a look through the existing .bf files if there are areas where it would make sense to use directly.

I don't think there is anything.  The largest constant used I could find is 64 and there's no usages of `0b` constants either.

lsf37:
Yes, you're right, this only was needed for the large tag value I had for the compound-tag work. Which now will also disappear, because they will be tuples.

So it's not really necessary. Should we still keep it or abandon?

kent-mcleod:
> abandon

lsf37:
> abandon

Agreed, we can resurrect it if we ever do need large values, but we really shouldn't.

PR #795, opened 2022-03-09
avoid multiple definitions of vspace architecture specific function interface

axel-h:
The generic code uses these functions, thus there should be one common interface definition only.

I have not added these functions
- `bool_t CONST isValidVTableRoot(cap_t cap)`
- `void setVMRoot(tcb_t *tcb)`

because they exists for all architectures, but there is no generic code that uses them. Just the architecture specific code seem to follow a common pattern here.

### DIFF
--- a/include/arch/arm/arch/kernel/vspace.h
+++ b/include/arch/arm/arch/kernel/vspace.h
@@ -29,20 +29,12 @@ void write_it_asid_pool(cap_t it_ap_cap, cap_t it_pd_cap);
 /* need a fake array to get the pointer from the linker script */
 extern char arm_vector_table[1];
 
-word_t *PURE lookupIPCBuffer(bool_t isReceiver, tcb_t *thread);
-exception_t handleVMFault(tcb_t *thread, vm_fault_type_t vm_faultType);
 pde_t *pageTableMapped(asid_t asid, vptr_t vaddr, pte_t *pt);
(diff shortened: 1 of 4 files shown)


lsf37:
You're right that functions called from generic code should have their interface defined generically. Should not be proof-breaking, but I'm double-checking just in case.

kent-mcleod (changes requested):
I don't think this needs changing.  Sometimes these functions end up getting implemented as static inline functions on some architectures and so trying to make them live in a shared header file when they don't have a single implementation isn't necessarily better.

axel-h:
> Sometimes these functions end up getting implemented as static inline functions

You are right, this would get difficult then. But in the case here I'd argue that it's not inlined , so we can be a bit more strict on the architecture interfaces definition to improve the code quality. We can worry about the inline thing if this happens.

lsf37:
Let's close this one for now. I'm fine with the idea as such, but Kent is right, it's not high value-to-effort ratio and we'll probably see further conflicts with aarch64 work etc.

axel-h:
Well, I can't really follow the argumentation actually and it lacks an example. The compiler might inline things at a certain optimization level anyway. The main intention of this PR is enforcing some interface between generic code and arch code that is easier to find (and harder to break by accident across architectures). I find it sad that there is not much interest in going that way.

PR #799, opened 2022-03-10
bitfield_gen: allow non-contiguous tag fields

### LINKED ISSUE #788: allow non-contiguous tag field in bitfield generator
Context: on AArch64, the encoding for PTE types is context dependent: on level 3, a block PTE is signified by bottom bits `0b11`, but on levels 0-2, bottom bits `0b11` instead mean table PTE. 

For verification, we need a context-independent encoding of tagged unions. (And for general sanity).

Luckily there are 4 bits in PTEs that are reserved for software, so we can use one of these to indicate which kind of PTE we are talking about. The problem is that these bits live at positions `[58:55]`, and the type bits at `[1:0]`, so they don't form a contiguous field, which the current bitfield generator assumes.

With a relatively small extension, this could be made possible.  The extension would contain roughly the following steps/features:

- [ ] extend parser to `tagged_union <type> <field>[:<field>]*`
- [ ] each of these fields has to be in the same position and width in all blocks/alternatives
- [ ] for simplicity, non-contiguous tags cannot be mixed with class-mask tags
- [ ] tag mask = `or` of the masks of all tag fields
- [ ] enum = plain numbers (no shifting), e.g. `0b00010000011` where the first `1` is in a first field (e.g. one of those reserved bits) and the bottom `11` are in a separate field (e.g. the hardware type field)
- [ ] creation = as now, tag is a plain number, needs no masking or shifting
- [ ] reading the tag = new combined function, works like normal field read, but with the overall tag mask
- [ ] should not generate field functions (esp not `set`) for any tag field

Making the tag the `or` of all tag fields (masked) means there is no performance impact compared to the current bitfield code -- the same number of shifting/masking operations is happening after constant folding. It does mean that all tag fields have to be within the same word (e.g. can't be spread over `word[0]` and `word[1]` of a larger bitfield).

The bitfield proofs should be almost unchanged, but we will see more when we have a few examples.

Adding multiple words would be a performance impact (but only when they are used), and it would add a fair bit of complexity, including proof change, so this is best left for a separate stage if it is needed.

lsf37:
A tagged union can now optionally use multiple fields to indicate the tag. These are called "sliced" tags in the code. The tag fields have to be at the same position and width in each block of the tagged unions, and all tag fields have to be within the same word.

When sliced tags are used, tag class masks cannot be used at the same time for this tagged union.

In the `tagged_union` declaration, the values of such sliced tags become tuples. 

Example:
```
block ex1 {
    padding 16
    field x 1
    padding 16
    padding 29
    field y 2
}

block ex2 {
    padding 16
    field x 1
    padding 16
    field b 5
    padding 20
    field a 4
    field y 2
}

tagged_union ex exType(x,y) {
    tag ex1 (1,0)
    tag ex2 (1,3)
}
```

In the existing bitfield generation almost nothing changes: some redundant assertions were dropped, but no assertion strength lost, rest of the code should be equal.

## Testing

I've tested compilation and proofs for small examples as well as for some otherwise transparent changes in the seL4 bitfield structures. These all pass. As usual, error checking can probably be improved, and there may well be combinations of base, tags, tag positions/offset/sizes that for some reason don't work. In those cases, the proofs will fail, of course, but you have to actually run them.

Implements/fixes #788

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 #
+# Copyright 2022, Proofcraft Pty Ltd
 # Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
 #
 # SPDX-License-Identifier: BSD-2-Clause
@@ -226,10 +227,30 @@ def p_fields_padding(t):
     t[0] = t[1] + [(None, t[3], False)]
 
 
+def p_tag_name_parts_one(t):
+    """tag_name_parts : IDENTIFIER"""
+    t[0] = [t[1]]
+
+
+def p_tag_name_parts(t):
+    """tag_name_parts : tag_name_parts COMMA IDENTIFIER"""
+    t[0] = t[1] + [t[3]]
+
+
+def p_tag_slices_empty(t):
+    """tag_slices : """
+    t[0] = []
+
+
+def p_tag_slices(t):
+    """tag_slices : LPAREN tag_name_parts RPAREN"""
+    t[0] = t[2]
+
+
 def p_tagged_union(t):
-    """tagged_union : TAGGED_UNION IDENTIFIER IDENTIFIER""" \
+    """tagged_union : TAGGED_UNION IDENTIFIER IDENTIFIER tag_slices""" \
         """ LBRACE masks tags RBRACE"""
-    t[0] = TaggedUnion(name=t[2], tagname=t[3], classes=t[5], tags=t[6])
+    t[0] = TaggedUnion(name=t[2], tagname=t[3], tag_slices=t[4], classes=t[6], tags=t[7])
 
 
 def p_tags_empty(t):
@@ -237,8 +258,28 @@ def p_tags_empty(t):
     t[0] = []
 
 
+def p_tag_values_one(t):
+    """tag_values : INTLIT"""
+    t[0] = [t[1]]
+
+
+def p_tag_values(t):
+    """tag_values : tag_values COMMA INTLIT"""
+    t[0] = t[1] + [t[3]]
+
+
+def p_tag_value(t):
+    """tag_value : LPAREN tag_values RPAREN"""
+    t[0] = t[2]
+
+
+def p_tag_value_one(t):
+    """tag_value : INTLIT"""
+    t[0] = [t[1]]
+
+
 def p_tags(t):
-    """tags : tags TAG IDENTIFIER INTLIT"""
+    """tags : tags TAG IDENTIFIER tag_value"""
     t[0] = t[1] + [(t[3], t[4])]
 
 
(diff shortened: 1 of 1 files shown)


lsf37:
(rebased; also running full hw build to make sure I didn't overlook any obscure non-proof case)

kent-mcleod:
> When sliced tags are used, tag class masks cannot be used at the same time for this tagged union.
 
What does this mean? That tag class masks cannot be used for bitfield types that have sliced tags?  What uses tag class masks currently?

lsf37:
> > When sliced tags are used, tag class masks cannot be used at the same time for this tagged union.
> 
> What does this mean? That tag class masks cannot be used for bitfield types that have sliced tags? 

Yes, it's one or the other, but not both.

> What uses tag class masks currently?

Only caps.

Xaphiosis:
I think this looks good. Some questions (mostly on error handling and presenting it to user):
- Does the bitfield generator have any documentation where you could add what you wrote for this PR?
- I remember a limitation to a single word, will it error out sanely if more than one word is needed?
- I haven't found an error for when class masks and sliced tags occur together
- What happens (in the python) when the position/size don't match between blocks of a union?

lsf37:
> I think this looks good. Some questions (mostly on error handling and presenting it to user):
> 
> * Does the bitfield generator have any documentation where you could add what you wrote for this PR?

Unfortunately not, and I think we should start writing some, because the source is getting more and more complex over time.

> * I remember a limitation to a single word, will it error out sanely if more than one word is needed?

Good question. It'll error if it computes a different tag index for any tag field (which happens when you go into another word), but I don't know what happens when you provide a tag field that is larger than the word base size. Will check.

> * I haven't found an error for when class masks and sliced tags occur together

I thought I explicitly tested for it, but I can't find it either, let me check that.

> * What happens (in the python) when the position/size don't match between blocks of a union?

There are a bunch of cases in the `resolve` and `set_base` methods that will throw an exception when they detect size or position mismatch. This happened already before the patch, so it it's probably not very visible in the diff. The only difference between now and before is that this check is in a loop for all tag slices (of which there is one if we're in the old case).


- [ ] make an issue for bitfieldgen docs
- [ ] check what happens when tag size > word size
- [ ] check what happens when class mask and compound tags are used at the same time

lsf37:
> *  make an issue for bitfieldgen docs
> *  check what happens when tag size > word size
> *  check what happens when class mask and compound tags are used at the same time

When tag size > word size we get an error message, because no field is allowed to be larger than the word size anyway. So that one is fine. When compound tags and class masks are used simultaneously, the class mask is currently ignore silently, which is not good. I'll add an explicit error check for that.

lsf37:
> When compound tags and class masks are used simultaneously, the class mask is currently ignore silently, which is not good. I'll add an explicit error check for that.

Done. From my side this is now ready to merge.

Xaphiosis (approved):

PR #801, opened 2022-03-11
RFC: AArch64: remove VSpace object types, seL4_ARM_PageDirectory and seL4_ARM_PageUpperDirectory

kent-mcleod:
This PR proposes to remove two AArch64 VSpace object types, seL4_ARM_PageDirectory and seL4_ARM_PageUpperDirectory. The functionality previously provided by these types will be provided by the existing seL4_ARM_PageTable object type.  This allows for a simpler API and enables a smaller kernel implementation that will be easier to verify.

- [x] Resolve https://github.com/seL4/seL4/issues/1075

Test with: seL4/seL4_libs#59
Test with: https://github.com/seL4/sel4test/pull/100


Indanz:
The code changes seem like a big improvement overall. Could this be also done for other architectures and have a unified page table API for all supported architectures?

Could you please explain how this affects user space? Specifically it is unclear to me how to map a PD after these changes.

Currently my code allocates and maps a PD when seL4_ARM_PageTable_Map() fails with seL4_FailedLookup. How does seL4_ARM_PageTable_Map() know whether to do a PT mapping or PD mapping now?

lsf37:
> The code changes seem like a big improvement overall. Could this be also done for other architectures and have a unified page table API for all supported architectures?

In the longer-term future that might be an option yes, although there are no current plans for this. It'd certainly help with proof maintainability and keeping down proof size. There are a number of things that need to be solved on the proof side first, i.e. AArch32 uses additional page table encodings (contiguous entries) that the current generic code can't deal with yet, and it's unclear if generalising that will make sense.
 
> Could you please explain how this affects user space? Specifically it is unclear to me how to map a PD after these changes.

From the mapping operation itself as before.

> Currently my code allocates and maps a PD when seL4_ARM_PageTable_Map() fails with seL4_FailedLookup. How does seL4_ARM_PageTable_Map() know whether to do a PT mapping or PD mapping now?

Page directory and page table are now the same kind of object (same as in RISC-V), and you can map that object at either level. If the lookup for the provided address stops at a slot that can map a page directory, it will map the object as a page directory. If there already is a page directory mapped, the lookup will proceed to the next level and it will map a page table.

Indanz:
> From the mapping operation itself as before.

Before I used seL4_ARM_PageDirectory_Map(), but that is removed in this change.

> If the lookup for the provided address stops at a slot that can map a page directory, it will map the object as a page directory. If there already is a page directory mapped, the lookup will proceed to the next level and it will map a page table.

This would break my code, as it won't know at which level the mapping has been done. It will try to do a PT mapping and silently get a PD mapping instead. As no error is returned, it assumes the PT mapping succeeded, while the program will crash when it tries to access the memory.

Forcing user space to keep a shadow bookkeeping on top of managing page tables seems an unreasonable burden and a regression to the current code.

At least add a seL4_ARM_VMAttributes flag to map it at any level, that way the behaviour is backwards compatible. Or retain seL4_ARM_PageDirectory_Map(), as doing two different operations with one functions depending on kernel state seems strange.

axel-h:
Following up on the discussion during hangout call:
- there should be no need to have book keeping in userland, if the kernel has this information anyway and could make it available
- need to document the API breakages
- consider adding a query API that can be used to determine what is mapped there. The information is not secret and this does not reveal any details userland would be able to find out anyway. 
- could add a flag, so one can explicitly control the behavior.

pingerino:
> there should be no need to have book keeping in userland, if the kernel has this information anyway and could make it available

I wasn't on the call, but this is contrary to the design of the kernel which keeps the code as small as possible.

kent-mcleod:
> Or retain seL4_ARM_PageDirectory_Map(), as doing two different operations with one functions depending on kernel state seems strange.

Just to confirm, are you suggesting that one option would be to keep separate invocations for mapping a page table to a specific level in the page table hierarchy which would fail if this wasn't the case, while still only having a single page table object type?

kent-mcleod:
What is the difference between a page directory and a page table in the terminology you are using?  If the final mapping is a page mapping, via seL4_ARM_Page_Map(), it will keep returning seL4_FailedLookup with the level information required to know what intermediate page table structures are needed until it succeeds. This allows you to continuously map PageTable objects (using seL4_ARM_PageTable_Map()) until the seL4_ARM_Page_Map finally.

In the main branch, trying to map a PageTable into a VSpace without a PageDirectory already mapped will also return a seL4_FailedLookup with the level information. However with the proposed change, instead the PageTable will be successfully mapped at an earlier mapping level. 

> As no error is returned, it assumes the PT mapping succeeded, while the program will crash when it tries to access the memory.

In the scenario you describe, the program would crash anyway because a Page still hasn't been mapped. Can you confirm that by "PT mapping succeeded" you are referring to an operation mapping a page table object into another page table (previously referred to as a page directory) and not an operation mapping a Frame into a page table?

Indanz:
I'm terribly sorry, I confused "page mapping" with "page table mapping". I thought you also unified page caps with page table caps.

So everywhere I said page table/PT I meant page mapping. Please ignore all my comments above, the current changes are fine as they are and nothing else is needed.

lsf37:
> I'm terribly sorry, I confused "page mapping" with "page table mapping". I thought you also unified page caps with page table caps.
> 
> So everywhere I said page table/PT I meant page mapping. Please ignore all my comments above, the current changes are fine as they are and nothing else is needed.

Cool, glad this is cleared up.

We still should make sure to describe the now relaxed error condition (the RFC already does that; we should add it to the CHANGES section here), but I'm happy that we have not introduced need for additional user-level book-keeping.

lsf37:
[RFC-10] is now approved, so this PR can go out of draft mode and be reviewed for merge.

libsel4/sel4_arch_include/aarch64/interfaces/sel4arch.xml
```diff
@@ -207,158 +207,4 @@
             </error>
         </method>
     </interface>
-    <interface name="seL4_ARM_PageUpperDirectory" manual_name="Page Upper Directory"
```
kent-mcleod, line 210: The CHANGES file will need updating for these changes

axel-h:
What is the plan about getting this merged?

yyshen:
> What is the difference between a page directory and a page table in the terminology you are using? If the final mapping is a page mapping, via seL4_ARM_Page_Map(), it will keep returning seL4_FailedLookup with the level information required to know what intermediate page table structures are needed until it succeeds. This allows you to continuously map PageTable objects (using seL4_ARM_PageTable_Map()) until the seL4_ARM_Page_Map finally.
> 
Just to confirm that for 2MiB and 1GiB page mapping, it works the same way.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/objecttype.h
```diff
@@ -8,7 +8,6 @@
 
 typedef enum _mode_object {
     seL4_ARM_HugePageObject = seL4_NonArchObjectTypeCount,
-    seL4_ARM_PageUpperDirectoryObject,
-    seL4_ARM_PageGlobalDirectoryObject,
+    seL4_ARM_VSpaceObject,
```
yyshen, line 11: It looks like a mode-agnostic object.

src/arch/arm/object/smmu.c
```diff
@@ -272,7 +272,7 @@ exception_t decodeARMCBInvocation(word_t label, unsigned int length, cptr_t cptr
         cb = cap_cb_cap_get_capCB(cap);
         cbSlot = smmuStateCBNode + cb;
         setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
-        smmu_tlb_invalidate_cb(cb, cap_vtable_root_get_mappedASID(cbSlot->cap));
+        smmu_tlb_invalidate_cb(cb, cap_vspace_cap_get_capMappedASID(cbSlot->cap));
```
yyshen, line 275: so this means smmuv2 + aarch64 only?

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/objecttype.h
lsf37: The concept could indeed be generalised to other architectures, but I would leave that for a later stage.

kent-mcleod:
> We still should make sure to describe the now relaxed error condition (the RFC already does that; we should add it to the CHANGES section here), but I'm happy that we have not introduced need for additional user-level book-keeping.

Done

src/arch/arm/object/smmu.c
kent-mcleod: The change applies to all aarch64 configurations.  These changes here are because the cap_vtable_root_get_mappedASID CPP macro was removed as it's not needed anymore (https://github.com/seL4/seL4/pull/801/files#diff-6916e963c01ff1638d47ccd93c325cb39419e814a584fc0b59e966958dbda834L46-L115)

include/arch/arm/arch/64/mode/object/structures.h
```diff
@@ -208,16 +159,10 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
...
-    case cap_page_directory_cap:
-        return PT_PTR(cap_page_directory_cap_get_capPDBasePtr(cap));
-
-    case cap_page_upper_directory_cap:
-        return PUD_PTR(cap_page_upper_directory_cap_get_capPUDBasePtr(cap));
-
-    case cap_page_global_directory_cap:
-        return PGD_PTR(cap_page_global_directory_cap_get_capPGDBasePtr(cap));
+    case cap_vspace_cap:
+        return PT_PTR(cap_vspace_cap_get_capPTBasePtr(cap));
```
lsf37, line 165: we're returning a `void*` in the end, so this probably doesn't matter much, but should this be `VSPACE_PTR` instead of `PT_PTR` here?

```suggestion
    case cap_vspace_cap:
        return VSPACE_PTR(cap_vspace_cap_get_capPTBasePtr(cap));
```

include/arch/arm/arch/64/mode/object/structures.h
```diff
@@ -236,60 +181,29 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
...
 {
-    return (pde_t) {
-        {
-            0
-        }
-    };
+    uint64_t ret;
+    ret = (pt->words[0] & 0xfffffffff000ull);
+    return ret;
 }
```
lsf37, line 209: I probably knew this at some point, but is there a reason why we're not using the bitfield generator function for this one? `page_base_address` is a proper field, so should work, no?

Also, would we not be catching the `UXN` field with this mask? (`pte_sw_type` is always 0 for `page_pte`, but `UXN`, I'm less sure)

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
```diff
@@ -6,4 +6,40 @@
...
+
+#else
+
+#define seL4_PGDBits 12
+#define seL4_PGDEntryBits 3
+#define seL4_PGDIndexBits    9
+
+#define seL4_PUDBits 12
+#define seL4_PUDIndexBits 9
+#define seL4_ARM_PageGlobalDirectoryObject seL4_ARM_VSpaceObject
```
lsf37, line 39: The build failure in `libsel4vka` looks like `seL4_ARM_PageGlobalDirectoryObject` was declared as a name even in hyp configs with `CONFIG_ARM_PA_SIZE_BITS_40` defined. Not sure if it is better to do that again so less code breaks, or if it's better to break that code, because it shouldn't have been using that name in the first place.

src/arch/arm/64/object/objecttype.c
```diff
@@ -415,41 +341,23 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
...
                );
 #else
 
-        return cap_page_global_directory_cap_new(
+        return cap_vspace_cap_new(
                    asidInvalid,           /* capPGDMappedASID   */
                    (word_t)regionBase,    /* capPGDBasePtr      */
                    0                      /* capPGDIsMapped     */
                );
 #endif /*!CONFIG_ARM_SMMU*/
```
lsf37, line 360: the comments in this bit still are referring to `capPGD`. Should probably update.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -433,20 +407,20 @@ static BOOT_CODE cap_t create_it_pd_cap(cap_t vspace_cap, pptr_t pptr, vptr_t vp
...
 }
 
 static BOOT_CODE cap_t create_it_pud_cap(cap_t vspace_cap, pptr_t pptr, vptr_t vptr, asid_t asid)
 {
     cap_t cap;
-    cap = cap_page_upper_directory_cap_new(
+    cap = cap_page_table_cap_new(
               asid,               /* capPUDMappedASID */
               pptr,               /* capPUDBasePtr */
               1,                  /* capPUDIsMapped */
```
lsf37, line 426: The comments should now probably be `capPTMappedASID` etc

src/arch/arm/64/kernel/vspace.c
```diff
@@ -474,37 +448,40 @@ BOOT_CODE cap_t create_it_address_space(cap_t root_cnode_cap, v_region_t it_v_re
     seL4_SlotPos slot_pos_after;
 
     /* create the PGD */
-    vspace_cap = cap_vtable_cap_new(
+    vspace_cap = cap_vspace_cap_new(
                      IT_ASID,        /* capPGDMappedASID */
                      rootserver.vspace, /* capPGDBasePtr   */
                      1               /* capPGDIsMapped   */
```
lsf37, line 454: (comment update)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1251,134 +977,66 @@ static inline void invalidateTLBByASIDVA(asid_t asid, vptr_t vaddr)
...
-        cleanByVA_PoU((vptr_t)pudSlot, pptr_to_paddr(pudSlot));
-        invalidateTLBByASID(asid);
+    for (word_t i = 0; i < UPT_LEVELS - 1 && pt != target_pt; i++) {
+        ptSlot = pt + GET_UPT_INDEX(vptr, i);
+        if (unlikely(!pte_pte_table_ptr_get_present(ptSlot))) {
+            /* couldn't find it */
+            return;
+        }
+        pt = paddr_to_pptr(pte_pte_table_ptr_get_pt_base_address(ptSlot));
     }
```
lsf37, line 998: We might need this loop in a separate function for verification in the future, but let's leave it here for now and see how far we get.

lsf37 (approved):
Cool, this looks very nice, only some minor tweaks in the comments.

We still need to get sel4test passed before we can merge, but that also looks close.

include/arch/arm/arch/64/mode/object/structures.h
lsf37: Actually I think I miscounted, the mask is fine, should be 36 bits, which fits and does mask out the higher parts (UXN etc).

I'd still go for the bitfield function unless there is a specific reason not to.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -44,43 +44,15 @@ block page_table_cap {
...
-#else 
-    padding                          48
-#endif 
-}
-
--- First-level page table (page global directory)
-block page_global_directory_cap {
-    field capPGDMappedASID           16
-    field_high capPGDBasePtr         48
+-- First-level page table (vspace_root)
```
yyshen, line 47: it looks like that the Arm arch manual uses level 0, 1, 2, 3 to refer the page tables. Should we follow the same pattern?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -647,290 +624,80 @@ exception_t checkValidIPCBuffer(vptr_t vptr, cap_t cap)
...
-        pd = paddr_to_pptr(pude_pude_pd_ptr_get_pd_base_address(pudSlot.pudSlot));
-        pdSlot = pd + pdIndex;
-
-        ret.status = EXCEPTION_NONE;
-        ret.pdSlot = pdSlot;
-        return ret;
-    }
-}
-
 static lookupPTSlot_ret_t lookupPTSlot(vspace_root_t *vspace, vptr_t vptr)
```
yyshen, line 627: How do we represent a lookup error in general? When

src/arch/arm/64/kernel/vspace.c
```diff
@@ -647,290 +624,80 @@ exception_t checkValidIPCBuffer(vptr_t vptr, cap_t cap)
...
-                   0,                          /* Ignored - Outter shareable */
-                   APFromVMRights(vm_rights),
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-                   S2_DEVICE_nGnRnE
-#else
-                   DEVICE_nGnRnE
-#endif
-               );
+    if (page_size == ARMSmallPage) {
+        ret.words[0] |= 0x3;
```
yyshen, line 698: So, this is the only difference between making 4K page and making 2MiB and 1GiB pages? We could use a bit comment here?

yyshen (approved):
Great work. It simplified the code a lot.

include/arch/arm/arch/64/mode/object/structures.bf
lsf37: I would slightly prefer the names, because the numbers can be read either way around (0 being bottom or 0 being top), and you have to know which way around the Arm manual uses them to read the code. With the names that possibility of confusion is gone. You do need to know which name maps to which level, so the gain is not much, but I'm still slightly in favour of that.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -341,38 +291,27 @@ block pde_large {
...
-tagged_union pde pde_type {
-    tag pde_large                   1
-    tag pde_small                   3
+-- There is two page types because the 4k page size
+-- uses a different hardware encoding. We use bit 58
+-- which is reserved for software use to encode this
+-- difference in the tag for these types.
+tagged_union pte pte_type(pte_hw_type, pte_sw_type) {
+    tag pte_table               (3, 0)
+    tag pte_page                (1, 0)
```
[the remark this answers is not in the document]
lsf37: We don't need to -- they have the same encoding (the rest of the code knows which level the PTE is on, so knows the page size from that). It's only bottom-level pages that have a different hardware PTE encoding.

src/arch/arm/64/kernel/vspace.c
lsf37: Yeah, you're right, we should document that.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -309,28 +270,17 @@ block pude_1g {
...
-    tag pude_invalid                0
-    tag pude_1g                     1
-    tag pude_pd                     3
+    field pte_hw_type               2
 }
 
-block pde_large {
-    padding                         9
+-- The level 3 page pte structure
+block pte_4k_page {
```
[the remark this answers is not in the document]
kent-mcleod: Probably something we could support as a static config option initially.  (Do you know what the ISA says about whether the granule size can be changed on a regular program context switch? Or whether it's only expected to be set once for the entire system during boot?)

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -271,34 +241,25 @@ tagged_union asid_map type {
...
-    padding                         62
-    field pgde_type                 2
-}
 
-block pgde_pud {
-    padding                         16
-    field_high pud_base_address     36
+block pte_table {
+    padding                         5
+    field pte_sw_type               1
```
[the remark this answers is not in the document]
kent-mcleod: There is an explanation about this on the definition of  `pte_type`:
There is two page type fields because the 4k page size uses a different hardware encoding. We use bit 58 which is reserved for software use to encode this difference in the tag for these types.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -271,34 +241,25 @@ tagged_union asid_map type {
...
     padding                         10
-    field pgde_type                 2 -- must be 0b11
-}
-
-tagged_union pgde pgde_type {
-    tag pgde_invalid                0
-    tag pgde_pud                    3
+    field_high pt_base_address      36
+    padding                         10
+    field pte_hw_type               2
```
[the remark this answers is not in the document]
kent-mcleod: See https://github.com/seL4/seL4/pull/801/commits/eaf38ea18d6c04c45e25acac8cee083ec1e7778e#r1282524459

include/arch/arm/arch/64/mode/machine/hardware.h
```diff
@@ -11,6 +11,28 @@
...
+#ifdef AARCH64_VSPACE_S2_START_L1
+#define UPT_LEVELS 3
+#define ULVL_FRM_ARM_PT_LVL(n) ((n)-1)
+#else
+#define UPT_LEVELS 4
+#define ULVL_FRM_ARM_PT_LVL(n) (n)
+#endif
+#define KPT_LEVELS 4
+#define KLVL_FRM_ARM_PT_LVL(n) (n)
+#define GET_KPT_INDEX(addr, n)  (((addr) >> (((PT_INDEX_BITS) * (((KPT_LEVELS) - 1) - (n))) + seL4_PageBits)) & MASK(PT_INDEX_BITS))
```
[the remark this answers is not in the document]
kent-mcleod: I updated the comments at the top of these definitions to explain n counts from the top level. (ie n=0 is vspace root table)

src/arch/arm/64/kernel/vspace.c
kent-mcleod: The old lookup functions try to return a reference to the slot which can be used to inspect the state of the slot (empty, page table entry or page entry), however if the slot doesn't exist in the provided vspace because it's container pagetable doesn't exist, the reference returned is null.  A status variable is also returned to indicate whether the reference is valid or not.

The new lookup function lookupPTSlot returns the most resolved slot reference and the number of remaining bits that weren't resolved.  This way a slot reference will always be returned (because it's required to call the function with a valid vspace_root_t and a vptr_t that's contained within the vspace), and the caller is able to use the number of resolved bits return value to know how deep the slot is in the table hierarchy.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -236,55 +209,57 @@ BOOT_CODE void map_kernel_window(void)
...
+    assert(GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
     assert(IS_ALIGNED(PPTR_TOP, seL4_HugePageBits));
 
     /* place the PUD into the PGD */
-    armKSGlobalKernelPGD[GET_PGD_INDEX(PPTR_BASE)] = pgde_pgde_pud_new(
-                                                         addrFromKPPtr(armKSGlobalKernelPUD));
+    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] = pte_pte_table_new(
+                                                                                 addrFromKPPtr(armKSGlobalKernelPUD));
 
     /* place all PDs except the last one in PUD */
```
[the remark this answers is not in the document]
kent-mcleod: I'm tempted to leave them as is for now as it still seems useful for this function context where the kernel vspace is built up level by level.

include/arch/arm/arch/64/mode/object/structures.bf
```diff
@@ -176,9 +148,7 @@ tagged_union cap capType {
     -- 5-bit tag arch caps
     tag frame_cap                   1
     tag page_table_cap              3
-    tag page_directory_cap          5
-    tag page_upper_directory_cap    7
-    tag page_global_directory_cap   9
+    tag vspace_cap                  9
```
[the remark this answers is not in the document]
kent-mcleod: I don't think there's a requirement that they're consecutive (maybe the proofs have some requirement here?), but there's the seL4 debug call  that user level can use to identify a captype from the cap number, so leaving the numbers the same avoids some user level breakage here since we can.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
kent-mcleod: I added a fix to my PR in seL4_libs for this change

include/arch/arm/arch/64/mode/object/structures.h
kent-mcleod: > I probably knew this at some point, but is there a reason why we're not using the bitfield generator function for this one? `page_base_address` is a proper field, so should work, no?

Because the asserts that check the type bits line up fail if using the generated functions.

kent-mcleod:
Now just waiting for tests to pass.

Ivan-Velickovic:
@kent-mcleod do you plan to update the kernel's manual?

kent-mcleod:
> @kent-mcleod do you plan to update the kernel's manual?

The API sections should update automatically from the changed XML definitions .  I'll make a quick update of https://sel4.systems/Info/Docs/seL4-manual-latest.pdf#subsubsection.7.1.1.4 though.  Thanks for the reminder.

PR #802, opened 2022-03-15
cmake: remove unused code

axel-h:
Clean up the CMake script:
- `SLOTS` cannot be less than `MAX_IRQ` (I'm not sure why this was added initially, seems some safe-guard that was never really needed)
- remove helper variable `MAX_IRQ` and use only `MAX_NUM_IRQ` everywhere  (should have happened at the last refactoring already)

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -48,10 +48,9 @@ set_property(
 
 # These options are now set in seL4Config.cmake
 if(DEFINED CALLED_declare_default_headers)
-    # calculate the irq cnode size based on MAX_IRQ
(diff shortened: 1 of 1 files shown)


kent-mcleod (approved):
Unused is probably a better word to use in this context than useless.

axel-h:
> Unused is probably a better word to use in this context than useless.

Agree, sounds a bit nicer. 

@szhuang: I've looked though the code history and it seems this check never really needed. However, I'm not sure if I miss something there why this was added. Do you happen remember anything?

PR #803, opened 2022-03-16
small python config handling improvements

axel-h:
- reduce redundancy in config selection
- reverse config class names. The naming is more intuitive this way and the classes show up together in the IDE outline.

### DIFF
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -111,9 +111,12 @@ class RISCVConfig(Config):
 
 def get_arch_config(sel4arch: str, addrspace_max: int) -> Config:
     ''' Return an appropriate Config object for the given architecture '''
-    if sel4arch in ['aarch32', 'aarch64', 'arm_hyp']:
-        return ARMConfig(sel4arch, addrspace_max)
-    elif sel4arch in ['riscv32', 'riscv64']:
-        return RISCVConfig(sel4arch, addrspace_max)
-    else:
-        raise ValueError('Unsupported sel4arch "{}" specified.'.format(sel4arch))
+
+    for (ctor, arch_list) in [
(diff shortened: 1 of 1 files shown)


lsf37:
Can't really say I'm in favour of this one.

> * reduce redundancy in config selection

I don't see any reduction in redundancy. The name `sel4arch` is mentioned one times less, but I find the previous structure clearer. The new one is not wrong or anything, but it's not better.

> * reverse config class names. The naming is more intuitive this way and the classes show up together in the IDE outline.

My IDE already shows them next to each other. In fact, there are only three classes, so I don't think this is a good argument for code churn.

axel-h:
> I don't see any reduction in redundancy. The name sel4arch is mentioned one times less, but I find the previous structure clearer. The new one is not wrong or anything, but it's not bette

Basically, this change come from some experimenting with adding more architectures. It scales a bit better, because new classes can be added as a line to the array then.

lsf37:
> Basically, this change come from some experimenting with adding more architectures. It scales a bit better, because new classes can be added as a line to the array then.

I don't really see that, to me it scales exactly the same: you either add an additional `ifelse` branch or you add a line.

axel-h:
It's more "pythonic" ;)

kent-mcleod:
> * reduce redundancy in config selection
> 
> * reverse config class names. The naming is more intuitive this way and the classes show up together in the IDE outline.

A PR wouldn't be blocked from being merged based on these requirements. As these changes aren't fixing an existing problem or adding new functionality I don't think they should be merged. We aren't about to add any new architectures so trying to make the code ready for something we aren't actively planning for isn't a good use of limited attention spans.

axel-h:
Closing this based on the feedback

PR #804, opened 2022-03-19
riscv-he: Merge branch 'master' into riscv_he_v0.6

Ivan-Velickovic:
The goal of this PR is to merge in the latest kernel into the RISC-V HE branch used for the RISC-V VMM. See  SEL4PROJ/sel4-riscv-vmm-manifest#1 for details.

There were merge conflicts in the following files, not sure if this is helpful, but I've noted my resolution for each, they are all fairly minor:

* `include/arch/riscv/arch/machine.h`
    * Merged in write_sip. Not sure whether to get rid of this or not, I can't find anywhere it's being used.
    * Removed `Arch_finaliseInterrupt` and helper function declarations below it, since they were removed.
* `src/arch/riscv/config.cmake`
    * Removed `halt.s` since it no longer exists.
* `src/arch/riscv/machine/hardware.c`
    * Rename `INTERRUPT_CORE_TIMER` to `KERNEL_TIMER_IRQ`.
    * `map_kernel_devices`, same as master but with the PLIC mapped in explicitly.
    * `isIRQPending` stays the same with the interrupt flags renamed to match the current ones.
    * `getNewActiveIRQ` was merged into `getActiveIRQ`, the logic for finding the active IRQ is the same before the merge.
* `src/arch/riscv/platform_gen.h.in`
     * Rename `INTERRUPT_CORE_TIMER` to `KERNEL_TIMER_IRQ`.
* `src/plat/spike/config.cmake`
     * Removed `LLU` from `TIMER_FREQUENCY` to match the current style.

The only other change I made was to add a condition for the spike platform in `riscv_plic0.h`, otherwise the VMM doesn't build.


axel-h:
Why are you not simply rebasing the branch? Who owns it, @yyshen?

yyshen:
@Ivan-Velickovic Rebasing the branch with master branch would be better? I hope rebasing will not introduce more work.

lsf37:
> How much time are you going top spend on this topic? I've looked though the commits the branch and wonder if they should be squashed a bit. especially when the deal with older version of the H extension.
> I've also see this uses spike and not the QUEM RISC-V virt platform (https://www.qemu.org/docs/master/system/riscv/virt.html), would it make sense to switch to it?

lsf37:
Sorry, accidentally hit the "close" button instead of commenting. If squashing helps with future rebases that would be worth considering, yes. Probably better for @yyshen and @Ivan-Velickovic to work this out between themselves if they are the main people working on this branch.

If the branch is rebased, do we still need a PR? There seems to be one additional commit.

axel-h:
I think we can close the PR, this is still a dev branch that does no give much guarantees at the moment. The next step for a PR could be upstreaming parts of this to `master`, not sure if there is something alreadd - refactoring maybe?

Ivan-Velickovic:
Closed since it's not necessary for this to be a PR, it's only the rebase with one extra commit.

Ivan-Velickovic:
> How much time are you going top spend on this topic? I've looked though the commits the branch and wonder if they should be squashed a bit. especially when the deal with older version of the H extension. I've also see this uses spike and not the QUEM RISC-V virt platform (https://www.qemu.org/docs/master/system/riscv/virt.html), would it make sense to switch to it?

The VMM runs using the QEMU `virt` platform. I assume they're similar enough that adding kernel support for the RISC-V `virt` platform isn't necessary, for now.

axel-h:
I have qemu "virt" somehow building now and will start upstreaming this, see https://github.com/seL4/seL4/pull/818

PR #806, opened 2022-03-21
Follow common CMAKE usage

locriacyber:
Instead of hard coding compiler and compile targets, let CMake figure that out itself.


configs/seL4Config.cmake
```diff
@@ -227,11 +227,18 @@ if(KernelArchARM)
...
 endif()
 
-if("${TRIPLE}" STREQUAL "")
+set(LLVM_TOOLCHAIN ON CACHE BOOL "if using clang as compiler (auto detected)")
+if (CMAKE_ASM_COMPILER_ID MATCHES "GNU")
+    set(LLVM_TOOLCHAIN OFF)
+endif()
+
+
+if("$LLVM_TOOLCHAIN" STREQUAL "ON")
```
axel-h, line 236: nitpicking:
```suggestion
if("${LLVM_TOOLCHAIN}" STREQUAL "ON")
```
We could also simply use `if(LLVM_TOOLCHAIN)`, as this is sufficient given the if-evaulation rules from https://cmake.org/cmake/help/latest/command/if.html#command:if

configs/seL4Config.cmake
```diff
@@ -244,16 +251,6 @@ if(
     configure_file(
         "${CMAKE_CURRENT_LIST_DIR}/../${toolchain_file}" "${toolchain_outputfile}.temp" @ONLY
     )
-    if(EXISTS "${toolchain_outputfile}")
```
axel-h, line 247: @kent-mcleod: what do you think about removing this check? I assume it there for a reason mainly coming from using the CMake GUI to change the settings. So we might better keep this to avoid broken builds then.

locriacyber: I edit CMakeCache.txt directly. For some reason this don't allow me to compile.

To avoid broken builds, it's better to add protection in CMake scripts directly.

lsf37: It sounds to me like the check did exactly what it was intended to do. Editing `CMakeCache.txt` directly is not supported by the rest of the build system and will lead to inconsistent results.

lsf37:
I see no reason for making this change. Please provide a rationale why this fixing anything or makes something better.

configs/seL4Config.cmake
kent-mcleod: > It sounds to me like the check did exactly what it was intended to do. Editing `CMakeCache.txt` directly is not supported by the rest of the build system and will lead to inconsistent results.

The error should have only been raised if the changes to the CMakeCache.txt changed the contents of the generated toolchain file. It shouldn't raise an error for simply editing the CMakeCache.txt file as this is what the `ccmake` program essentially does - which is supported.  However making config changes that affect the generated toolchain file (which should be restricted to changes in the following cache variables: `KernelSel4Arch`, `KernelArch`, `KernelWordSize`, `CROSS_COMPILER_PREFIX`) will generate this error as changing the toolchain file requires re-initializing the build because CMake caches a lot of toolchain discovery it does and this doesn't get reset if the toolchain file changes.

kent-mcleod:
> Instead of hard coding compiler and compile targets, let CMake figure that out itself.

The default toolchain files have a fair amount of policy around selecting the right compilers based on the target platform and assumptions about the developer environment.  We don't want to change this if it means having to also always pass in a toolchain option with a platform option.

The existing scripts are set up to ignore the default toolchain files if `CMAKE_TOOLCHAIN_FILE` is set on the CMake command line. If you want different toolchain selection policy then this may be a better option for you.

locriacyber:
It's better to rely on CMake to find the right compiler, and the project should only reject the compiler if it's not good. The project should never choose the compiler itself, not hard-code the path of compiler either.

lsf37:
> It's better to rely on CMake to find the right compiler, and the project should only reject the compiler if it's not good. The project should never choose the compiler itself, not hard-code the path of compiler either.

I disagree entirely. seL4 makes assumptions on the specific compiler and flags being used and will not work correctly with just any compiler. So choosing the compiler is precisely what it should do.

kent-mcleod:
> It's better to rely on CMake to find the right compiler, and the project should only reject the compiler if it's not good. The project should never choose the compiler itself, not hard-code the path of compiler either.

We are following CMake's documented way of setting cross compilers using toolchain files based on the documentation here: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling.

> The [CMAKE_\<LANG\>_COMPILER](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER.html#variable:CMAKE_%3CLANG%3E_COMPILER) variables may be set to full paths, or to names of compilers to search for in standard locations. For toolchains that do not support linking binaries without custom flags or scripts one may set the [CMAKE_TRY_COMPILE_TARGET_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_TRY_COMPILE_TARGET_TYPE.html#variable:CMAKE_TRY_COMPILE_TARGET_TYPE) variable to STATIC_LIBRARY to tell CMake not to try to link executables during its checks.

locriacyber:
@kent-mcleod thanks! I'll use CMAKE_TOOLCHAIN_FILE.

PR #808, opened 2022-03-22
python: use '#paragma once' include guard

axel-h:
using the '#paragma once' include guard simplifies things a bit and it's the modern way to guard includes.

### DIFF
--- a/libsel4/tools/syscall_stub_gen.py
+++ b/libsel4/tools/syscall_stub_gen.py
@@ -981,8 +981,7 @@ def generate_stub_file(arch, wordsize, input_files, output_file, use_only_ipc_bu
  * Automatically generated system call stubs.
  */
 
-#ifndef __LIBSEL4_SEL4_CLIENT_H
-#define __LIBSEL4_SEL4_CLIENT_H
+#pragma once
 """)
 
     # Emit the includes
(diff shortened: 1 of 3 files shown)


pingerino:
IIRC I didn't do this as it didn't work for kernel headers (#pragma once does not work if we concatenate all of the c files together).

axel-h:
> IIRC I didn't do this as it didn't work for kernel headers (#pragma once does not work if we concatenate all of the c files together).

This generates header files that are included. As far as I understand things,just the c-files are concatenated. The headers are still kept for the compiler to get pulled in, where it honors the `#pragma once` then. I can't see any traces from the header file contents in `kernel_all.c`. The compiler intermediate output `kernel_all.i` contains the header content eventually - but then exactly once, as expected.

ratmice:
Not a officially recognized reviewer, but I believe that Axel is correct, and that perhaps the confusion is that while syscall_stub_gen.py generates functions, and function bodies.  It does so for inline functions which are then written to header files, so it should not impact the concatenation of source files.

lsf37:
Axel is right, this one does work here and should be unified. We're already using `#pragma once` for the rest of the kernel headers and the preprocessed concatenated output is identical after this change.

lsf37 (approved):

kent-mcleod (approved):

PR #809, opened 2022-03-22
fix config includes

axel-h:
The kernel files usually simply include `config.h`, which then handles further details. It's unclear why these two files do things differently, there seems no specific reason and it might be just legacy. The file `config.h` will include `sel4/config.h` anyway, which does include `autoconf.h` then.
Changing this avoids setting a bad example. If there is reason, e.g. why `sel4/config.h` must be hidden from the GIC header file code that might be worth documenting. Due to the way the kernel build works on a single file, this header might have been included already anyway.

### DIFF
--- a/include/arch/arm/arch/machine/gic_common.h
+++ b/include/arch/arm/arch/machine/gic_common.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include <autoconf.h>
+#include <config.h>
 #include <stdint.h>
 #include <machine/interrupt.h>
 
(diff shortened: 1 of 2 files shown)


kent-mcleod:
> Changing this avoids setting a bad example.

It used to be more important that config.h was included and not autoconf.h because there originally was no autoconf.h for verification builds.  Now that there's always an autoconf.h there's no issue with files including it as it's always guaranteed to be there and contain configs defined by the build.

kent-mcleod (approved):

axel-h:
As with https://github.com/seL4/seL4/pull/825, the goal of these changes is a cleaner and common include structure. it also avoids using dynamically generated headers when there is a fixed header under source control also.

Issue #813, opened 2022-03-23
Remove deprecated libsel4 headers


axel-h:
I wonder if we can finally remove some deprecated libsel4 headers:

- `sel4/messages.h`
  - Seems deprecated since 2016 already. 
  - In https://github.com/seL4/seL4_projects_libs/pull/36 another legacy usage has been removed . 
  - The only usage that remains is in `refos`, see also https://github.com/seL4/refos/issues/15

- `sel4/arch/pfIPC.h`, `sel4/arch/exIPC.h`
  - Seems deprecated since 2016 already.
  - https://github.com/seL4/seL4_libs/pull/61 removes usage.
  - no other usage known

Is anybody else still using this?

kent-mcleod:
I think this is fine.  The upcoming release is already API breaking and migration only requires deleting the files.  A more complicated question is whether the old contents of those header files should be removed from sel4/deprecated.h

axel-h:
I was wondering if we can remove the include from `sel4/sel4 h` at least in the next release, so it must be included manually. The next release woul then remove the contents, but could still keep the empty files

kent-mcleod:
I was more specifically meaning removing the definitions such as `EXCEPT_IPC_SYS_MR_R0` which is what sel4/arch/exIPC.h headers used to contain.

axel-h:
Closing, https://github.com/seL4/seL4/pull/814 got merged.

PR #814, opened 2022-03-25
Remove deprecated libsel4 headers (address #813)

axel-h:
Remove deprecated libsel4 headers and cleanup the `deprecated.h` files.

Addresses https://github.com/seL4/seL4/issues/813


pingerino:
Can we keep the deprecated files, even if they are empty? I'm sure we will deprecate more things in future, and it's good to have an established pattern/placeholder to put things.

axel-h:
> Can we keep the deprecated files, even if they are empty? I'm sure we will deprecate more things in future, and it's good to have an established pattern/placeholder to put things.

I've kept the `deprecated.h` files in the arch-specific folders, as they appear to be the place where legacy/deprecation would be moved before things get dropped eventually. I have deleted the `pfIPC.h` and `exIPC.h` files, because they appear to be real legacy by now, the code contents were removed by you in 2016 already. I can see your point, that they serve as an example, but I wonder if it's better for the code base to cut off some legacy eventually to avoid bloating. I'd argue, that the code base is not so complex that this is something to keep alive, one could reference this in the history (or this PR), if the issue comes up again. Actually, I found it a bit confusing that these files exist and was wondering if I'm missing something that is still attached here.

pingerino:
> > Can we keep the deprecated files, even if they are empty? I'm sure we will deprecate more things in future, and it's good to have an established pattern/placeholder to put things.
> 
> I've kept the `deprecated.h` files in the arch-specific folders, as they appear to be the place where legacy/deprecation would be moved before things get dropped eventually. I have deleted the `pfIPC.h` and `exIPC.h` files, because they appear to be real legacy by now, the code contents were removed by you in 2016 already. I can see your point, that they serve as an example, but I wonder if it's better for the code base to cut off some legacy eventually to avoid bloating. I'd argue, that the code base is not so complex that this is something to keep alive, one could reference this in the history (or this PR), if the issue comes up again. Actually, I found it a bit confusing that these files exist and was wondering if I'm missing something that is still attached here.

Ah, Looks like you had already done what I was thinking and my morning brain couldn't put two and two together.

Originally I kept the pfIPC etc headers because some projects used them and would break. It's probably ok to delete them now.

axel-h:
Could this be merged?

kent-mcleod (approved):

PR #817, opened 2022-03-28
CAVEATS-generic: ASID caveat applies all arches

kent-mcleod:
Remove x86 and Arm from heading of the "Re-using address spaces" caveat
because it's also present on RISCV and isn't actually arch specific.


lsf37 (approved):

lsf37:
Yes, should have done that ages ago.

PR #818, opened 2022-03-29
support qemu-riscv-virt platform

axel-h:
- add `qemu-riscv-virt` support to match what exists for `qemu-arm-virt`. This lifts the dependency on using `spike`, which may still be very similar practically
- sel4test passes for non-MCS only,  MCS configuration fails with timeout error
- this still needs some small changes in other repos, so it's a draft for now
  - see working branch https://github.com/axel-h/util_libs/commits/patch-axel-5
  - updating OpenSBI should be considered https://github.com/seL4/sel4test-manifest/pull/13

Test with: seL4/util_libs#143, seL4/seL4_tools#154, seL4/sel4test#82


src/plat/qemu-riscv-virt/config.cmake
```diff
@@ -0,0 +1,103 @@
...
+        declare_seL4_arch(riscv64)
+    elseif("${KernelSel4Arch}" STREQUAL riscv32)
+        # This is still untested
+        declare_seL4_arch(riscv32)
+    else()
+        fallback_declare_seL4_arch_default(riscv64)
+    endif()
+
+    config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "generic")
+    config_set(KernelPlatformFirstHartID FIRST_HART_ID 1)
```
Ivan-Velickovic, line 24: The Spike uses 0 as the first hart ID, is there a reason the QEMU one is 1? Just curious, I couldn't get the RISC-V VMM working without changing this to 0.

axel-h: Yes, you are right. Thanks. I've been testing with 2 cores, thus this was working for me.

src/plat/qemu-riscv-virt/config.cmake
```diff
@@ -0,0 +1,103 @@
...
+        RESULT_VARIABLE error
+    )
+    if(error)
+        message(FATAL_ERROR "Failed to convert DTB to DTS (${QEMU_DTB})")
+    endif()
+
+    list(APPEND KernelDTSList "${QEMU_DTS}" "${CMAKE_CURRENT_LIST_DIR}/overlay-qemu-riscv-virt.dts")
+
+    # QEMU emulates SiFive PLIC/CLINT by default with a few interrupts, using
+    # 64 here seems a safe value. The 10 MHz timer clock is just a dummy value.
```
kent-mcleod, line 97: Is it intentional that this is in sync with the default timer value in the simulator definition? https://github.com/qemu/qemu/blob/2c3e83f92d93fbab071b8a96b8ab769b01902475/include/hw/intc/riscv_aclint.h#L75 It might be worth mentioning that in the comment.

kent-mcleod (approved):

kent-mcleod:
Looks good to me.  Would you be interested in adding this as a platform to the ci-actions when this is merged?

lsf37:
That'd be very cool. I can help if you want.

src/plat/qemu-riscv-virt/config.cmake
axel-h: It's no really intentional, might be just lucky here. Anyway, thanks for the note, I've extended the comment.

axel-h:
> Looks good to me. Would you be interested in adding this as a platform to the ci-actions when this is merged?

Yes, actually I'm trying to get the CI action to work on my branches, but so far it's not doing anything...

Ivan-Velickovic:
Which test fails in the MCS configuration? I get all tests passing with the following configs:

`../init-build.sh -DPLATFORM=qemu-riscv-virt -DRISCV64=1 -DSIMULATION=1 -DQEMU_MEMORY=4095M -DMCS=1`
`../init-build.sh -DPLATFORM=qemu-riscv-virt -DRISCV64=1 -DSIMULATION=1 -DQEMU_MEMORY=4095M`

axel-h:
@lsf37 : What is the trick to no longer run the sel4test-sim with `SPIKE64_debug_gcc_64` but use my new RISCVVIRT64 then?

lsf37:
> @lsf37 : What is the trick to no longer run the sel4test-sim with `SPIKE64_debug_gcc_64` but use my new RISCVVIRT64 then?

I can have a look at that. Does it make sense to test both platforms (spike and qemu-riscv-virt) or do we want to fully remove spike?

lsf37:
> @lsf37 : What is the trick to no longer run the sel4test-sim with `SPIKE64_debug_gcc_64` but use my new RISCVVIRT64 then?

So to get sel4test-sim to run on GitHub with the new `platforms.yml` definition you need to deploy it as a docker container somewhere and update [sel4test-sim/action.yml](https://github.com/Hensoldt-Cyber/ci-actions/blob/e072b7f481062018fbe001d8cce4aa465b54b011/sel4test-sim/action.yml#L39) to that deployed location.

axel-h:
> Does it make sense to test both platforms (spike and qemu-riscv-virt) or do we want to fully remove spike?

I think the virt platform is enough, spike just seems a subset of it nowadays.

lsf37:
> > Does it make sense to test both platforms (spike and qemu-riscv-virt) or do we want to fully remove spike?
> 
> I think the virt platform is enough, spike just seems a subset of it nowadays.

Ok, I'm happy with that. Are you fine with me adding all this to CI or did you want to first continue the experiment on your branch?

lsf37:
> > Ok, I'm happy with that. Are you fine with me adding all this to CI or did you want to first continue the experiment on your branch?
> 
> Do you refer to the changed I tried in [seL4/ci-actions#209](https://github.com/seL4/ci-actions/pull/209) ?

Yes, that's the one I meant. If you drop the experimental `.github/workflow` changes in the PR here, then we can merge the seL4 PR first, and then the ci-actions PR directly after.

lsf37 (approved):

lsf37:
Ok, so if we want to make one more attempt at properly testing this before we merge (I'm fine with this PR as is, but we might need additional changes in sel4test etc):

I've deployed the updated docker container as `sel4test-sim:testing` under the TS account, and added a local branch https://github.com/seL4/ci-actions/tree/patch-axel-6 in the ci-actions repo that now points at this deployment location.

That means, @axel, if you update the `.github/workflows` change in this PR here from
```
 - uses: Hensoldt-Cyber/ci-actions/sel4test-sim@patch-axel-6
 ```
to
```
 - uses: seL4/ci-actions/sel4test-sim@patch-axel-6
 ```
we should see it running the simulation tests with the new platform.

Ivan-Velickovic:
The test `FRAMEEXPORTS0001` requires more than 2GiB of RAM on RISC-V 64-bit and so will currently fail with the following build command: `../init-build.sh -DPLATFORM=qemu-riscv-virt -DRISCV64=1 -DSIMULATION=1`. One way to fix this is to change the default QEMU memory to be 4GiB, like the Spike.

axel-h:
> The test `FRAMEEXPORTS0001` requires more than 2GiB of RAM on RISC-V 64-bit and so will currently fail with the following build command: `../init-build.sh -DPLATFORM=qemu-riscv-virt -DRISCV64=1 -DSIMULATION=1`. One way to fix this is to change the default QEMU memory to be 4GiB, like the Spike.

I will change this then.

Ivan-Velickovic:
I think the tests are failing since `-bios none` is missing from the QEMU command.

Ivan-Velickovic:
> > I think the tests are failing since `-bios none` is missing from the QEMU command.
> 
> You mean `QEMU_CMD` in the cmake change here or should should there be something in the invocation on the test side?

The patch to seL4_tools needs a change.

axel-h:
> I think the tests are failing since `-bios none` is missing from the QEMU command.

Thanks. Yes, that would explain it. I'll have a look into the python scripts.

Ivan-Velickovic:
Cool, the `virt` tests pass now.

axel-h:
> Cool, the `virt` tests pass now.

Yes. I've removed the "wip" patch from this PR now, so this can be merged.

axel-h:
I have added `-bios none` to the command line that is used to get the device tree. This avoids that QEMU tries to load the bundled OpenSBI image. This avoid having a build failure if the image files does not exists.

PR #821, opened 2022-03-31
FPU as an object

andybui01:
### Brief summary
Currently, seL4 treats the FPU as part of the thread context, saved in the TCB on preemption. As FPU state is large, the kernel uses lazy switching for the FPU, meaning on a context switch FPU access is disabled, resulting in an exception when a thread attempts to use it; the kernel will then save and restore the FPU state. This introduces a timing channel (e.g. [cve-2018-3665](https://access.redhat.com/security/cve/cve-2018-3665)). The eager FPU design is chosen to remedy this, however, eager loading the FPU on every context switch in seL4 is impractical as it introduces high overheads to the context switch. The FPU is thus represented as an object so that the eager load only occurs if a thread has an FPU capability bound to it.

### API
The current API for the FPU object takes inspiration from the notification binding process:
`seL4_TCB_BindFPU(seL4_TCB tcb, seL4_CPtr fpu)`
`seL4_TCB_UnbindFPU(seL4_TCB tcb)`

### odroidc2 benchmarks
_(MCS off, inter-vspace, length 0, same prio)_
Call performance:
| Benchmark              | Mean | Min | Max | Std Dev | Iterations |
|------------------------|------|-----|-----|---------|------------|
| Old IPC call           | 234  | 234 | 234 | 0       | 16         |
| Eager FPU - No FPU     | 236  | 236 | 236 | 0       | 16         |
| Eager FPU - Both FPU   | 381  | 381 | 381 | 0       | 16         |
| Eager FPU - Client FPU | 250  | 250 | 250 | 0       | 16         |
| Eager FPU - Server FPU | 252  | 249 | 255 | 3       | 16         |

ReplyRecv performance:
| Benchmark              | Mean | Min | Max | Std Dev | Iterations |
|------------------------|------|-----|-----|---------|------------|
| Old IPC                | 244  | 244 | 244 | 0       | 16         |
| Eager FPU - No FPU     | 258  | 258 | 258 | 0       | 16         |
| Eager FPU - Both FPU   | 405  | 403 | 412 | 4       | 16         |
| Eager FPU - Client FPU | 270  | 270 | 270 | 0       | 16         |
| Eager FPU - Server FPU | 278  | 271 | 280 | 4       | 16         |

### Implementation status
| Arch    | Status        |
|---------|---------------|
| aarch64 | semi-complete |
| aarch32 | semi-complete/requires debugging |
| riscv   | semi-complete |
| x86     | WIP           |

- Semi-complete implementations currently suffer a problem in testing that cannot be explained. seL4test has plenty of uts to work with, however, after some uts have been split in testing by the utspace split allocator, they can't be untype_retype'd again, and the kernel complains that there is no memory left to allocate in the untyped when there should be. This is being investigated. Apart from this issue the semi-complete implementations are functional on ALL tests (the problem doesn't arise when running each test individually).

- The exact fault which is triggered when a thread does not have an FPU and attempts to perform an FPU op is left unimplemented and is left as a question to the developer community on what they prefer.

- The x86 WIP implementation can be viewed here: https://github.com/andybui01/seL4/tree/andy/wip-fpu-x86port.

- The aarch32 implementation is buggy when running with both SMP and MCS enabled in _debugging_ mode. My guess here is there's some memory overflow into the debugging structures that's causing this or vice versa.

### Using it
The following changes are necessary to run a build with the FPU object: https://github.com/andybui01/seL4_libs/commit/c45675827401f3943c5dee6cbd6354a784d10835.

If you are compiling for ARM, the flag `-mgeneral-regs-only` must be added to the muslcsys cmakelist. Furthermore, an FPU object must be allocated for the rootserver early on, to prevent any FPU faults from occurring where the compiler uses a FPU register. Again, this is only for ARM.

### DIFF
--- a/include/arch/arm/arch/32/mode/fastpath/fastpath.h
+++ b/include/arch/arm/arch/32/mode/fastpath/fastpath.h
@@ -123,7 +123,7 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
 #endif
 
 #ifdef CONFIG_HAVE_FPU
-    lazyFPURestore(NODE_STATE(ksCurThread));
+    eagerFPURestore(NODE_STATE(ksCurThread));
 #endif /* CONFIG_HAVE_FPU */
 
     register word_t badge_reg asm("r0") = badge;
--- a/include/arch/arm/arch/32/mode/machine/fpu.h
+++ b/include/arch/arm/arch/32/mode/machine/fpu.h
@@ -88,9 +88,10 @@ static void setEnFPEXC(void)
     VMSR(FPEXC, fpexc);
 }
 /* Store state in the FPU registers into memory. */
-static inline void saveFpuState(user_fpu_state_t *dest)
+static inline void saveFpuState(tcb_fpu_t *dest)
 {
     word_t fpexc;
+    fpu_t *fpu = dest->tcbBoundFpu;
 
     /* Fetch FPEXC. */
     VMRS(FPEXC, fpexc);
@@ -106,7 +107,7 @@ static inline void saveFpuState(user_fpu_state_t *dest)
     }
 #endif
 
-    dest->fpexc = fpexc;
+    fpu->fpexc = fpexc;
 
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
         /* before touching the regsiters, we need to set the EN bit */
@@ -114,10 +115,10 @@ static inline void saveFpuState(user_fpu_state_t *dest)
     }
 
     /* We don't support asynchronous exceptions */
-    assert((dest->fpexc & BIT(FPEXC_EX_BIT)) == 0);
+    assert((fpu->fpexc & BIT(FPEXC_EX_BIT)) == 0);
 
     if (isFPUD32SupportedCached) {
-        register word_t regs_d16_d31 asm("ip") = (word_t) &dest->fpregs[16];
+        register word_t regs_d16_d31 asm("ip") = (word_t) &fpu->regs_d16_d31[0];
         asm volatile(
             ".word 0xeccc0b20        \n"    /*  vstmia  ip, {d16-d31} */
             :
@@ -126,12 +127,14 @@ static inline void saveFpuState(user_fpu_state_t *dest)
         );
     }
 
-    register word_t regs_d0_d15 asm("r2") = (word_t) &dest->fpregs[0];
+    register word_t regs_d0_d14 asm("r2") = (word_t) &fpu->regs_d0_d14[0];
     asm volatile(
-        /* Store d0 - d15 to memory */
-        ".word 0xec820b20       \n" /* vstmia  r2, {d0-d15}" */
-        :
-        : "r"(regs_d0_d15)
+        /* Store d0 - d14 to memory */
+        ".word 0xec820b1e       \n" /* vstmia  r2, {d0-d14}" */
+
+        /* Store d15 to memory */
+        "vstr d15, [%1]         \n"
+        :: "r"(regs_d0_d14), "r"(&dest->d15)
     );
 
     /* Store FPSCR. */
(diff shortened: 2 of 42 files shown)


lsf37:
See also the [pre-RFC discussion](https://sel4.discourse.group/t/pre-rfc-fpu-as-an-object/475) that belongs to this. 

The conceptual side in the pre-RFC discussion on alternative designs etc should be resolved first before people spend a lot of time reviewing details.

andybui01:
> The conceptual side in the pre-RFC discussion on alternative designs etc should be resolved first before people spend a lot of time reviewing details.

Yep, just wanted to get the (partial at this point) implementation out there with some basic benchmarks. I'm happy to provide more benchmarking results if it advances the discussion.

lsf37:
> Yep, just wanted to get the (partial at this point) implementation out there with some basic benchmarks. I'm happy to provide more benchmarking results if it advances the discussion.

The question is not if this design is faster than what is there now, but more if an alternative design would not be overall better and faster than both, e.g. reducing TCB size without requiring a separate object. Eager/lazy is a separate dimension to that and could be implemented in either.

The background is that notification binding was an absolute nightmare to verify (took a lot longer that the most pessimistic estimate which was already significantly higher than our actual estimate at the time), so this does not exactly inspire confidence that this will be cheap. Avoiding having to deal with yet another object would also be nicer from the user side. But it's hard to say if that other idea would actually work without trying it.

gernotheiser:
> The question is not if this design is faster than what is there now, but more if an alternative design would not be overall better and faster than both, e.g. reducing TCB size without requiring a separate object. Eager/lazy is a separate dimension to that and could be implemented in either.

In theory, yes, in practice, I don't think so. If there is no explicit authorisation for FPU use, then eager switching would have to switch the FPU state in *every* context switch, and our fastest IPC times would blow out. The point is that OS services should not need the FPU, and thus invoking them should not require dealing with FPU state. So, eager switching and authorisation of FPU access are inherently connected. So it needs a capability and hence an object type.

In contrast, TCB size is more of an annoyance than a practical problem, but its reduction would be an ancillary benefit of an FPU object.

gernotheiser:
Having said that, we'll need to let this issue rest for now, as we have higher priorities short-term

lsf37:
> In theory, yes, in practice, I don't think so. If there is no explicit authorisation for FPU use, then eager switching would have to switch the FPU state in _every_ context switch, and our fastest IPC times would blow out. 

Like I said, switching is completely orthogonal to where the storage is. This includes authorisation. You can easily have cap-authorised FPU switching without a new object.

> The point is that OS services should not need the FPU, and thus invoking them should not require dealing with FPU state. So, eager switching and authorisation of FPU access are inherently connected. So it needs a capability and hence an object type.

I agree up until the last step. Cap authorisation does not need a new object type. It could be a right in the TCB cap, it could be its own FPU cap, and wether to switch or not needs only 1 bit of state, doesn't have to be an object. 

> In contrast, TCB size is more of an annoyance than a practical problem, but its reduction would be an ancillary benefit of an FPU object.

If TCB size is not a problem then we should not do the FPU object, because it introduces a lot of complexity for little gain. I'm up for working on reducing TCB size if it does turn out to be a problem, but we should explore other reasonable options first.

DemiMarie:
> The point is that OS services should not need the FPU

This is likely true on some embedded platforms.  It is _not_ true in general, as SIMD registers are used in many different places for cryptography and compression.  Preventing OS services from using optimized cryptographic routines would severely degrade disk encryption and VPN performance, among other things.

lsf37:
> > The point is that OS services should not need the FPU
> 
> This is likely true on some embedded platforms. It is _not_ true in general, as SIMD registers are used in many different places for cryptography and compression. Preventing OS services from using optimized cryptographic routines would severely degrade disk encryption and VPN performance, among other things.

The point of making this decision available at runtime is that user-level can set this up as required. If it is a component that benefits from FPU, you can switch it on, if it doesn't need it you can switch it off. I.e. the encryption component would have it on, many other OS components would probably have it off, application-level components would probably have it on, etc.

I do agree that it would be much nicer to be able to talk about actual systems and requirements here and measure instead of thinking about hypotheticals, but that is unfortunately unrealistic for this one. Current seL4 systems will do what the kernel currently supports and other systems that you could build don't exist yet or can't exist yet.

lsf37:
Closing this PR since there is no current work on it. Can be re-opened if needed, and the PR branch will stay for reference.

PR #824, opened 2022-03-31
cmake: extend macro declare_seL4_arch() to support multiple architectures

axel-h:
Provide a convenience function for the seL4 architecture setup. This remove a bit of redundancy from the code and things get easier for platforms.


axel-h:
Is this acceptable as CMake simplification?

kent-mcleod:
> Is this acceptable as CMake simplification?

Could this change also lead to less confusing user experience when setting config options? The current behavior when specifying a platform with an unsupported arch is that the arch gets overridden to one the platform supports, while now it would return an error and stop processing.

axel-h:
> Could this change also lead to less confusing user experience when setting config options? The current behavior when specifying a platform with an unsupported arch is that the arch gets overridden to one the platform supports, while now it would return an error and stop processing.

Well,yes. I was assuming the old behavior was not really a graceful feature, but just a sloppy implementation - that is why I made this a bit more strict now. If this breaks some build flows, it seems the builds flow should be fixed. Unless there is a good reason why this is not possible.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -12,18 +12,18 @@ declare_platform(qemu-arm-virt KernelPlatformQEMUArmVirt PLAT_QEMU_ARM_VIRT Kern
...
-    endif()
-    if(KernelSel4ArchArmHyp)
-        set(qemu_user_top 0xe0000000)
-    endif()
+    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "")
     set(QEMU_ARCH "arm")
     set(KernelArchArmV7a ON)
+    if(ARMV7_OPTIONS_ve)
+        declare_seL4_arch(aarch32 arm_hyp)
+        set(KernelArchArmV7ve ON)
```
Indanz, line 20: I don't see any users of either `KernelArchArmV7ve` or `ARCH_ARM_V7VE`, so it could be removed altogether if we want.

Indanz (approved):
Looks good to me.

configs/seL4Config.cmake
```diff
@@ -18,12 +18,31 @@ set(KernelDTSList "")
...
+
+# This macro is used by platforms to declare which seL4 architecture(s) they
+# support. It takes a list and sets up the one selected by KernelSel4Arch. If
+# KernelSel4Arch is not set, the architecture specified by the first list
+# element is used.
+# Usage example: declare_seL4_arch("aarch64" "aarch32")
+macro(declare_seL4_arch)
+    # Since this is a macro and not a function, ARGV is not a real variable. One
+    # must be created to be able iterate over it.
+    set(_arch_list "${ARGV}")
```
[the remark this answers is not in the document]
axel-h: If I remember correctly, this needs some more refactoring then due to the way how variable scoping works in CMake. I think calling `config_choice()` from a function does not work trivially, we'd have to manually put all variables it sets into the parent scope again.

src/plat/qemu-arm-virt/config.cmake
axel-h: It seems we could clean up some things here, but that should be a separate PR then.

Indanz: Agreed.

axel-h: It turns out that `ARCH_ARM_V7VE` seem unused in the C code, but `KernelArchArmV7ve` is used implicitly when we invoke the compiler with `-march=...` then. We might be able to clean up things a bit given that ARMv7 is superseded by now and things will never change beyond what we currently do. But we can't remove this completely.

lsf37:
This change has broken the [webserver example](https://github.com/seL4/sel4webserver/actions/runs/6969318254). @axel-h could you please have a look at that?

lsf37:
Does this mean that something should be documented in the change log, because it has effects on upstream `cmake` files?

axel-h:
I think the failure comes from
https://github.com/seL4/sel4webserver/blob/3d52bdc937bbebe39bb591e6ba706e8764ab4acd/settings.cmake#L64 
This should not be set there unconditionally, but just for ARMv7 platforms. Will run some more test.

Indanz:
The error is "KernelSel4Arch 'arm_hyp' not in 'aarch64'". I'd say failing to build if an unsupported `KernelSel4Arch` is set is an improvement. It doesn't seem necessary to document that a wrong config that happens to build will now fail to build, assuming that that's the case here.

lsf37:
> The error is "KernelSel4Arch 'arm_hyp' not in 'aarch64'". I'd say failing to build if an unsupported `KernelSel4Arch` is set is an improvement. It doesn't seem necessary to document that a wrong config that happens to build will now fail to build, assuming that that's the case here.

I think it'd be worth an entry to in change log (but not elsewhere in the docs). I agree that it's not a bad change, but if you're hitting this error as a user after the release, it'd be nice to get a hint for what you have to fix (even if this combination was wrong in the first place).

kent-mcleod:
This breaks camkes-vm-examples too.

Previously this initialization line would work and now it causes an error:
```
$ ../init-build.sh -DPLATFORM=qemu-arm-virt -DCAMKES_VM_APP=vm_minimal  -DCROSS_COMPILER_PREFIX=aarch64-none-linux-gnu-
loading initial cache file /tmp/tmp.5Kv410Y6vE/projects/vm-examples/settings.cmake
-- Set platform details from PLATFORM=qemu-arm-virt
--   KernelPlatform: qemu-arm-virt
-- Found seL4: /tmp/tmp.5Kv410Y6vE/kernel  
-- ARM_CPU not set, defaulting to cortex-a53
CMake Error at /tmp/tmp.5Kv410Y6vE/kernel/configs/seL4Config.cmake:43 (message):
  KernelSel4Arch 'arm_hyp' not in 'aarch64'
  /tmp/tmp.5Kv410Y6vE/kernel/src/plat/qemu-arm-virt/config.cmake:30 (declare_seL4_arch)
  /tmp/tmp.5Kv410Y6vE/kernel/src/plat/qemu-arm-virt/config.cmake:49 (setup_qemu_armv8)
  /tmp/tmp.5Kv410Y6vE/kernel/configs/seL4Config.cmake:170 (include)
  /tmp/tmp.5Kv410Y6vE/kernel/FindseL4.cmake:21 (include)
  settings.cmake:72 (sel4_configure_platform_settings)


-- Configuring incomplete, errors occurred!
```

You now have to pass in -DAARCH64=ON which shouldn't be required when selecting a 64-bit platform (the default of qemu-arm-virt when -DARM_CPU is not set)

Indanz:
`arm_hyp` is only used for aarch32 with hypervisor support, right? So when building for aarch64 you wouldn't expect to also set `arm_hyp`.

But the commit doesn't change much for cortex-a53, all it adds is checking whether `KernelSel4Arch` is valid. In this case, something sets it to `arm_hyp`, probably the camkes-vm config. It seems to be [this line](https://github.com/seL4/camkes-vm-examples/blob/6656ebe20269383a2e6ca51b698eac1e9814dd9d/settings.cmake#L41). That seems to be genuinely wrong if we end up building aarch64.

axel-h:
> It seems to be [this line](https://github.com/seL4/camkes-vm-examples/blob/6656ebe20269383a2e6ca51b698eac1e9814dd9d/settings.cmake#L41). That seems to be genuinely wrong if we end up building aarch64.

Yes, this causes it. Thanks for looking into that. I have improved the QEMU CPU selection in https://github.com/seL4/seL4/pull/1182 a bit now. It seem the breaking change is actually, that the default for QEMU was AARCH32 before and now it is AARCH64. I assume you want me to modify the PR, that it implements this behavior again?

axel-h:
OK, just to clarify what we have/want for QEMU/arm
- The seL4 default behavior is AARCH64
- The camkes-vm-examples setup defaults to AARCH32 if nothing is given - because it sets `arm_hyp` then. Then vm_minimal in AARCH32 does build, but it will not run. We have 64-bit Linux images only, they crash on ARMv7 emualtions.

kent-mcleod:
What I believe we had: 
- qemu-system-arm defaults to cortex-a53. Which selects aarch64 unless KernelSel4Arch has been initialized to aarch32 already. 
- in camkes-vm-examples Not providing AARCH64 would still result in an aarch64 build selected. 

So I agree that camkes-vm-examples shouldn't be testing for AARCH64. It was an old variable used for older versions of the gcc.cmake toolchain file. But it should still default to aarch64

axel-h:
Ok, then I think https://github.com/seL4/seL4/pull/1182 should fix the problem.

Indanz:
> > It seems to be [this line](https://github.com/seL4/camkes-vm-examples/blob/6656ebe20269383a2e6ca51b698eac1e9814dd9d/settings.cmake#L41). That seems to be genuinely wrong if we end up building aarch64.
> 
> Yes, this causes it. Thanks for looking into that. I have improved the QEMU CPU selection in #1182 a bit now. It seem the breaking change is actually, that the default for QEMU was AARCH32 before and now it is AARCH64.

I don't see how that got changed in this PR, maybe I'm too tired, but I totally missed that (and still do).

> I assume you want me to modify the PR, that it implements this behavior again?

Just to be clear, with "genuinely wrong" I meant that line I referred, not this PR.

axel-h:
> I don't see how that got changed in this PR, maybe I'm too tired, but I totally missed that (and still do).

All good, I got this one wrong. The PR did not change the default - but it made it too string. It changed the handling and did not properly take into account all setting that have been made by projects. Bottom line is again, that we need more in CI that cover all cases that exist.

PR #825, opened 2022-04-04
libsel4: use sel4/config.h instead of autoconf.h

axel-h:
Clarify libsel4 include structure and use `sel4/config.h` everywhere. Just `sel4/config.h` will eventually include the generated `autoconf.h` then. This reduces the dependency on the usage of the `autoconf.h` file and provide a central config file for libsel4 that is under git control.

### DIFF
--- a/libsel4/arch_include/arm/sel4/arch/constants_cortex_a15.h
+++ b/libsel4/arch_include/arm/sel4/arch/constants_cortex_a15.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include <autoconf.h>
+#include <sel4/config.h>
 
 #if !defined(CONFIG_ARM_CORTEX_A15)
 #error CONFIG_ARM_CORTEX_A15 is not defined
--- a/libsel4/arch_include/arm/sel4/arch/constants_cortex_a35.h
+++ b/libsel4/arch_include/arm/sel4/arch/constants_cortex_a35.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include <autoconf.h>
+#include <sel4/config.h>
 
 #if !defined(CONFIG_ARM_CORTEX_A35)
 #error CONFIG_ARM_CORTEX_A35 is not defined
--- a/libsel4/arch_include/arm/sel4/arch/constants_cortex_a53.h
+++ b/libsel4/arch_include/arm/sel4/arch/constants_cortex_a53.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include <autoconf.h>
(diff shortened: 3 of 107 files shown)


Indanz (approved):
Looks good.

I would also change the `#include <autoconf.h>` in sel4/config.h with `#include <kernel/gen_config.h>`, as that's all that autoconf.h contains.

Is it time to remove autoconf.h altogether and also remove generate_autoconf from tools/helpers.cmake?

axel-h:
If I remember correctly, this did not work because libsel4 code is used by the kernel directly and by userland. In these cases the actual `autoconf.h` used at compile time is slightly different. So dropping this might be more complex any maybe not even desired.

Indanz:
You're right, libsel4/autoconf/autoconf.h also includes sel4/gen_config.h.

kent-mcleod:
@axel-h, is there a grand plan here?  Given the length of time autoconf.h has been around and how many downstream header files include it (https://github.com/search?p=2&q=org%3AseL4+autoconf.h&type=Issues), I don't think it's going to get removed easily. 

We could consider updating the build of libsel4 to copy the kernel/gen_config.h into the sel4/ namespace and allow libsel4 to remove all include path dependencies for `kernel_Config sel4_Config sel4_autoconf`? This would allow newer libraries and updated libraries to move away from using autoconf and deleting it can be an easier option in the future.

axel-h:
@kent-mcleod: I have no intention to remove ` autoconf.h`, as it works quite well for what it is doing. I just want to limit the places where it is included because `sel4/config.h` is what is supposed to be used. It will handle the rest internally, but that's just on exactly one place then.

What you suggest sounds an interesting way towards a more "release package" orientated build then also. I will look into this a bit more.

Ben-PH:
In the context of [this pre-RFC](https://sel4.discourse.group/t/pre-rfc-to-effortlessly-develop-sel4-systems-in-rust/539), I think getting rid of `autoconf.h` becomes more pressing?

the complexity described here:
> If I remember correctly, this did not work because libsel4 code is used by the kernel directly and by userland. In these cases the actual autoconf.h used at compile time is slightly different. So dropping this might be more complex and maybe not even desired.

...makes the question of having a rust environment more complex as well.

lsf37:
I don't think the presence of `autoconf.h` (or not) make a difference for Rust builds. If we capture the config already in json, e.g. like in #893, the Rust build doesn't need to interact with it. It's just a C build output like other C header files.

kent-mcleod (approved):

axel-h:
rebased and applies change to newer files.

PR #834, opened 2022-04-27
benchmark/arm64: write-through kernel log buffer

malus-brandywine:
Set Write-through non-transient, No-Write-allocate, Read-allocate
attributes for kernel log buffer.

Signed-off-by: Nataliya Korovkina <malus.brandywine@gmail.com>

Test with: seL4/seL4_tools#146

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -35,6 +35,7 @@
  *  - GRE Unused Device Gathering, Reordering, Early write acknowledgement
  *  - NORMAL_NC Normal Memory, Inner/Outer non-cacheable
  *  - NORMAL Normal Memory, Inner/Outer Write-back non-transient, Write-allocate, Read-allocate
+ *  - NORMAL_WT Normal Memory, Inner/Outer Write-through non-transient, No-Write-allocate, Read-allocate
  * Note: These should match with contents of MAIR_EL1 register!
  */
 enum mair_types {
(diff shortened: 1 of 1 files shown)


malus-brandywine:
Important: to be merged after [elfloader PR #146](https://github.com/seL4/seL4_tools/pull/146)

axel-h:
I wonder if the aarch64 boot code of the kernel should do a sanity check in the MAIR_EL1 to ensure the contents are as expected. The kernel might also be loaded by something else instead of the ElfLoader and this loader must be updated then also. Can't be bad to have a safeguard in place - even if it's just a breaking change for a debug feature now.

malus-brandywine:
I'm also looking for help from verification people, please. Preprocess check failed.

axel-h:
> I'm also looking for help from verification people, please. Preprocess check failed.

That failure is expected, because the sources have changed. There should not be a direct impact on the verification, because you just added a new value that is used by debug code only. No existing or non-debug should use this.

malus-brandywine:
> > I'm also looking for help from verification people, please. Preprocess check failed.
> 
> That failure is expected, because the sources have changed. There should not be a direct impact on the verification, because you just added a new value that is used by debug code only. No existing or non-debug should use this.

Thank you, @axel-h !

kent-mcleod (approved):

lsf37:
I think there is actually minor verification impact, because the enum changed, which we're modelling. No big problem, though. I'll report back when the spec is updated.

lsf37 (approved):
This is good to go from the verification side, we don't currently model all of the enum, so no change needed.

PR #839, opened 2022-05-10
qemu-arm-virt: Add support for 32bit arm hyp mode configurations

kent-mcleod:
Add support for the qemu-arm-virt simulation platform to run seL4 in hyp mode.

### DIFF
--- a/include/arch/arm/arch/32/mode/hardware.h
+++ b/include/arch/arm/arch/32/mode/hardware.h
@@ -85,6 +85,10 @@
 #ifndef __ASSEMBLER__
 /* It is required that USER_TOP must be aligned to at least 20 bits */
 compile_assert(USER_TOP_correctly_aligned, IS_ALIGNED(USER_TOP, 20));
+/* It is required on arm_hyp that USER_TOP isn't lower than the top GiB */
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+compile_assert(USER_TOP_top_gb, USER_TOP >= 0xC0000000);
+#endif
 
 #include <plat/machine/hardware.h>
 #endif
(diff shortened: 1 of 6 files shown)


include/arch/arm/arch/32/mode/hardware.h
```diff
@@ -85,6 +85,10 @@
 #ifndef __ASSEMBLER__
 /* It is required that USER_TOP must be aligned to at least 20 bits */
 compile_assert(USER_TOP_correctly_aligned, IS_ALIGNED(USER_TOP, 20));
+/* It is required on arm_hyp that USER_TOP isn't lower than the top GiB */
```
Indanz, line 88: Where does this requirement come from?

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -60,7 +71,7 @@ if(KernelPlatformQEMUArmVirt)
     config_set(KernelARMPlatform ARM_PLAT qemu-arm-virt)
     set(DTBPath "${CMAKE_BINARY_DIR}/virt.dtb")
     set(DTSPath "${CMAKE_BINARY_DIR}/virt.dts")
-    if(KernelArmHypervisorSupport)
+    if(KernelArmHypervisorSupport OR KernelSel4ArchArmHyp)
```
Indanz, line 74: Why have both? Is there a difference in meaning between the two?

Indanz (approved):
Looks good to me.

include/arch/arm/arch/32/mode/hardware.h
kent-mcleod: https://github.com/seL4/seL4/blob/master/src/arch/arm/32/kernel/vspace.c#L316.  The kernel has it's own PL2 address space, but only uses the top 1 GiB.

src/plat/qemu-arm-virt/config.cmake
kent-mcleod: `KernelArmHypervisorSupport` may not be set at this point as it's not declared until a later CMake file is processed.

yyshen (approved):
Good work!

PR #843, opened 2022-05-17
x86_64: Fix x86_64 gcc compile error with sysexitq

kent-mcleod:
sysexit requires a 64-bit operand to remain in 64-bit mode when
switching to user level. gcc versions older than 12 encode this as
"REX.W SYSEXIT" while clang encodes this as "SYSEXITQ". Both compilers
don't support the alternate encoding. We instead use ".byte
0x48,0x0F,0x35" which is the actual opcode of the required instruction
which will work with both compilers.

### DIFF
--- a/include/arch/x86/arch/64/mode/fastpath/fastpath.h
+++ b/include/arch/x86/arch/64/mode/fastpath/fastpath.h
@@ -203,7 +203,7 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
 #endif /* CONFIG_KERNEL_SKIM_WINDOW */
 #endif /* defined(ENABLE_SMP_SUPPORT) && defined(CONFIG_KERNEL_SKIM_WINDOW) */
             "sti\n"
-            "sysexitq\n"
+            SYSEXITQ"\n"
             :
             : "c"(&cur_thread->tcbArch.tcbContext.registers[RAX]),
             "D"(badge),
--- a/include/arch/x86/arch/64/mode/machine.h
+++ b/include/arch/x86/arch/64/mode/machine.h
@@ -13,6 +13,22 @@
 #include <arch/model/smp.h>
 #include <arch/machine.h>
 
+/*
+ * SYSEXIT  0F 35     ; Return to compatibility mode from fast system call.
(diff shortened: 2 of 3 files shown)


include/arch/x86/arch/64/mode/fastpath/fastpath.h
```diff
@@ -203,7 +203,15 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
...
-            "sysexitq\n"
+            /* Return to user.
+             *
+             * SYSEXIT  0F 35     ; Return to compatibility mode from fast system call.
+             * SYSEXITQ 48 0F 35  ; Return to 64-bit mode from fast system call.
+             *
+             * clang uses "sysexitq" and gcc uses "rex.w sysexit" to generate 0x48,0x0F,0x35
+             * only at gcc 12 or later will sysexitq work on both compilers
+             */
+            ".byte 0x48,0x0F,0x35 \n"
```
axel-h, line 214: We could put this in `include/arch/x86/arch/64/mode/hardware.h`
```
/*
 * SYSEXIT  0F 35     ; Return to compatibility mode from fast system call.
 * SYSEXITQ 48 0F 35  ; Return to 64-bit mode from fast system call.
 *
 * clang uses "sysexitq" and gcc uses "rex.w sysexit" to generate 0x48,0x0F,0x35
 * only at gcc 12 or later "sysexitq" works on both compilers.
 */
#if (defined(__clang__) || (defined(__GNUC__) && (__GNUC__ > 12)))
    #define SYSEXITQ "sysexitq"
#elif defined(__GNUC__)
    #define SYSEXITQ ".byte 0x48,0x0F,0x35"
#else
    #error "unsupported compiler"
#endif
```

and then just use this here (and also in src/arch/x86/64/c_traps.c);
```
            "sti\n"
            SYSEXITQ "\n" /* return to user */
            :
```

kent-mcleod: That should at least allow the CParser to work without modifications

lsf37: Yes, I like that much better!

lsf37 (approved):

Issue #844, opened 2022-05-17
Call updateTimestamp() before taking kernel lock for IRQs?


Indanz:
System calls are initiated by the currently running task, so charging them makes sense. The same is not true for interrupts, especially not the scheduling timer interrupt causing a context switch.

It may seem it does not matter whether a task is charged for the context switch at the beginning of its period or at the end, but it does matter for periodic tasks that call seL4_yield(): Counter intuitively, such tasks run less often if they are not charged at the beginning. It is because they will run every period + kernel lock take time (~= WCET), instead of every period.

This introduces an accumulating error between the system clock and timers managed by such tasks. Practically it means any timers handled by such task will sometimes trigger one period later than expected, if they use absolute time based on the system clock.

On the system I use the error is about 1 ms per four seconds. It is a problem because I have a polling driver that wants to poll every millisecond, now it's sometimes triggers twice after two milliseconds, introducing unnecessary jitter into the system.

In my case there are ways of working around this issue, but the fundamental question is whether such drift for periodic tasks is unwanted or not.

Below a patch that eliminates drift for periodic tasks on my system (Aarch64). I also need the -1 in setDeadline(), but sadly getTimerPrecision() uses usToTicks(TIMER_PRECISION), with TIMER_PRECISION in microseconds instead of ticks.
```
diff --git a/src/api/syscall.c b/src/api/syscall.c
index 40f6f0690280..9155503de9dd 100644
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -41,7 +41,6 @@ exception_t handleInterruptEntry(void)
     irq = getActiveIRQ();
 #ifdef CONFIG_KERNEL_MCS
     if (SMP_TERNARY(clh_is_self_in_queue(), 1)) {
-        updateTimestamp();
         checkBudget();
     }
 #endif
diff --git a/src/arch/arm/c_traps.c b/src/arch/arm/c_traps.c
index 740d00313d9a..550bf78bbaf4 100644
--- a/src/arch/arm/c_traps.c
+++ b/src/arch/arm/c_traps.c
@@ -89,6 +89,7 @@ void VISIBLE NORETURN c_handle_instruction_fault(void)

 void VISIBLE NORETURN c_handle_interrupt(void)
 {
+    updateTimestamp();
     NODE_LOCK_IRQ_IF(IRQT_TO_IRQ(getActiveIRQ()) != irq_remote_call_ipi);
     c_entry_hook();

[...]

PR #847, opened 2022-05-18
updateTimestamp before taking kernel lock for IRQs

Indanz:
Otherwise periodic tasks that run out of budget can be postponed
by low priority tasks, effectively increasing their period with
the average lock take time. This will cause periodic tasks to
drift in regard to the system clock and each other, making the
system less deterministic. See also seL4 issue #844.


include/kernel/thread.h
```diff
@@ -307,5 +307,7 @@ void awaken(void);
 /* Place the thread bound to this scheduling context in the release queue
  * of periodic threads waiting for budget recharge */
 void postpone(sched_context_t *sc);
+#else
+static inline void updateTimestamp(void) {}
```
axel-h, line 311: Given we need the function in MSC and non-MCS now (even if it is just a dummy for non-MSC to avoid to much conditional parts), I suggest to move it out of the condition block now and instead create a conditional block for the function body. Having just one definition in the code makes it easier to find the actual implementation.
The compiler should be able to optimize this out, but it may have an impact on verification. In the declined PR https://github.com/seL4/seL4/pull/752 I added a macro `MCS_TERNARY()` to wrap this and avoid impacting verification.

axel-h:
I wonder, what are the rules concerning the `c_entry_hook()` actually. Seems this nicely get's along everywhere, so there is n immediate issues. 
Calling `updateTimestamp()` everywhere also seems handy for the logging/profiling framework, as no separate hook is needed then any longer.

include/kernel/thread.h
Indanz: Agreed and done.

Indanz:
> I wonder, what are the rules concerning the c_entry_hook() actually. Seems this nicely get's along everywhere, so there are no immediate issues.

Yes, I noticed that too. I've been sidetracked enough already, so I've willfully ignored it.

> Calling updateTimestamp() everywhere also seems handy for the logging/profiling framework, as no separate hook is needed then any longer.

If you mean `ksEnter` and `timestamp()`, I wouldn't be so sure. Before I agree it was pretty much the same as `ksCurTime`, but now ksEnter is used to measure the time spent while holding the lock.

(Except in case of FPU faults on arm, which updates ksEnter without holding the lock, but that's just a bug: I think `ksEnter` must become a per node variable. Wasn't someone busy with getting rid of lazy FPU loading anyway?)

Also, to get rid of `ksEnter`, the system clock must be precise enough for benchmarking. Both `timestamp()` and `getCurrentTime()` are architecture specific and may not be compatible. E.g. `x86_rdtsc()` is quite different than x86's `timestamp()`, even though both use `rdtsc`.

include/kernel/thread.h
```diff
@@ -309,3 +285,28 @@ void awaken(void);
 void postpone(sched_context_t *sc);
 #endif
 
+/* Update the kernels timestamp and stores in ksCurTime.
```
axel-h, line 288:
```suggestion
/* Update the kernel's timestamp and stores it in ksCurTime.
```

include/kernel/thread.h
```diff
@@ -309,3 +285,28 @@ void awaken(void);
...
+ * The difference between the previous kernel timestamp and the one just read
+ * is stored in ksConsumed.
+ *
+ * Should be called on every kernel entry where threads can be billed.
+ * May be called without holding the kernel lock!
```
axel-h, line 293:
```suggestion
 * Can be called without holding the kernel lock because no global data is written.
```

include/kernel/thread.h
```diff
@@ -309,3 +285,28 @@ void awaken(void);
...
+ */
+static inline void updateTimestamp(void)
+{
+#ifdef CONFIG_KERNEL_MCS
+    time_t prev = NODE_STATE(ksCurTime);
+    NODE_STATE(ksCurTime) = getCurrentTime();
```
axel-h, line 299: I wonder, should we read this into a local variable and then have just one write back to memory? I'd trust the compiler to figure out it can do thus, but doing it explicitly in the source might play save and clearly state we want this.

include/kernel/thread.h
Indanz: The whole point of the comment was to warn people who may want to change this function to be very careful. Perhaps I should change the "May" into "Will".

Currently global data is written, because although it's per node, it may be read cross-node. Hopefully that will get fixed soon though (#854).

I don't like the "because no global data is written" because that implies that that is the only condition before this is thread-safe, which is not true.

include/kernel/thread.h
Indanz: Agreed, good moment for this. I also fixed the ksConsumed comment.

include/kernel/thread.h
axel-h: ah ok, missed that. Maybe make it even a compile time assert then, to this place it found easily once this changes.

axel-h: I wonder, now that we call `updateTimestamp()` on every kernel entry, would it be worth making the `numDomain` check compiled in conditionally at non-SMP only? Might save a few instructions then.

include/kernel/thread.h
```diff
@@ -309,3 +285,29 @@ void awaken(void);
...
+    assert(NODE_STATE(ksCurTime) < MAX_RELEASE_TIME);
+    time_t consumed = (NODE_STATE(ksCurTime) - prev);
+    NODE_STATE(ksConsumed) += consumed;
+    /* Not supported for SMP: */
```
axel-h, line 303: I wonder, now that we call `updateTimestamp()` on every kernel entry, would it be worth making the `numDomain` check compiled in conditionally at non-SMP only? Might save a few instructions everytime.

kent-mcleod:
> (Except in case of FPU faults on arm, which updates ksEnter without holding the lock, but that's just a bug: I think `ksEnter` must become a per node variable. Wasn't someone busy with getting rid of lazy FPU loading anyway?)

It does seem like `c_handle_enfp` not acquiring a lock on SMP is a bug as it does access a lot of state that could be being torn down by another core with the lock. 

> I wonder, what are the rules concerning the c_entry_hook() actually. Seems this nicely get's along everywhere, so there are no immediate issues.

c_entry_hook() is called inside the lock so doesn't seem usable here.

I also agree with keeping both timestamp() and getCurrentTime() separate.

Indanz:
> It does seem like c_handle_enfp not acquiring a lock on SMP is a bug as it does access a lot of state that could be being torn down by another core with the lock.

A currently running task can't suddenly disappear, that's synchronised with IPIs. Other than `ksCurThread`, only local CPU state is accessed and other task's `user_fpu_state_t`. That is fixed by `fpuThreadDelete()` on task deletion, so everything other than `c_entry_hook()` seems safe to me. No need to take the lock.

Actually, `c_exit_hook()` is always called without holding the kernel lock, so another reason all benchmarking variables should be per node. `benchmark_track_exit` is not thread-safe either. I'll make a new issue for this.

kent-mcleod:
> A currently running task can't suddenly disappear, that's synchronised with IPIs. Other than `ksCurThread`, only local CPU state is accessed and other task's `user_fpu_state_t`. That is fixed by `fpuThreadDelete()` on task deletion, so everything other than `c_entry_hook()` seems safe to me. No need to take the lock.

Comparing with all other arches, aarch64 is the only one that doesn't take the lock before handling the fpu fault. But it seems you're right that it's safe.  The only race that I can see is where a spurious IPI to set the current FPU owner to NULL is generated when the remote core reads an out of date value of ksActiveFPUState.

include/kernel/thread.h
```diff
@@ -309,3 +285,29 @@ void awaken(void);
...
+/* Update the kernel's timestamp and store it in ksCurTime.
+ * The difference between the previous kernel timestamp and the one just read
+ * is added to ksConsumed.
+ *
+ * Should be called on every kernel entry where threads can be billed.
+ * Will be called without holding the kernel lock!
+ */
+static inline void updateTimestamp(void)
+{
+#ifdef CONFIG_KERNEL_MCS
```
lsf37, line 297: This change makes the `updateTimestamp` function visible (and called) in all configs, and would require proof update for everything. The declaration should stay under the `CONFIG_KERNEL_MCS` guard, and the calls as well.

Might make sense to add a macro for the call that is empty (fully empty, not semicolon) when `CONFIG_KERNEL_MCS` is off to avoid lots of `ifdef` at the call sites.

lsf37 (approved):
I'm probably going to regret this when we verify SMP, but I think this looks fine from the verification side (modulo my CONFIG_KERNEL_MCS comment). And I do agree with the intention that the time should be read as early as possible at kernel entry.

Calling `getCurrentTime()` outside the lock will make things more nondeterministic for verification, because reading the time will change the global state (at least in the model we're thinking of now, maybe it can be made core-local), but it is supposed to be non-deterministic already. 

Otherwise we seem to be accessing only core-local state and locals which should not be able to see any interference from elsewhere. We can only be competing for the lock with code from other cores, which would not touch the state we're writing, so I can see no new data races.

Can someone assure me that calling `getCurrentTime()` outside the lock is safe on all architectures? I have looked at a few, and they seem all read-only as they should be, but I'm not sure if any of the lower-level instructions have side effects.

include/kernel/thread.h
lsf37: Sorry about the redo, it would add a lot of noise to the proofs. Yes, a conditionally empty macro at the call site and a conditionally included declaration/definition should work for verification (the preprocess test should then be passing for the non-MCS configs).

Indanz:
For proper SMP verification I think it should be proven that anything outside the global lock only touches node-local data, with the exception of bootup. For non-formal proof the compiler and linker could be used to check this, by compiling the code outside of global locks separately.

> Can someone assure me that calling getCurrentTime() outside the lock is safe on all architectures? I have looked at a few, and they seem all read-only as they should be, but I'm not sure if any of the lower-level instructions have side effects.

The lower-level system instructions to get time like used on x86, aarch64 and riscv don't have side effects. The arm32 global and Exynos timers read clock registers and those also don't have any side effects. am335x and omap3430 are single core. So I'm convinced that `getCurrentTime()` is safe outside the global lock.

I'm not convinced that `setDeadline()` is SMP-safe except for the one using architectural system timers though: arm32, x86 APIC and Exynos seem suspect to me. But that's an unrelated SMP MCS issue.

include/kernel/thread.h
Indanz: I added the macro, but I didn't go for fully empty, without even semicolons added, as that seems a bit excessive. Do the proofs require updates for semicolons only? Can't those be just filtered out automatically?

Indanz:
> Interestingly, a test I added recently triggers an assertion with this branch:
> 
> ```
> Running test SCHED_CONTEXT_0014 (Test deleting a scheduling context while it runs out of budget on another core)
> seL4 failed assertion 'refill_sufficient(NODE_STATE(ksCurSC), NODE_STATE(ksConsumed))' at 
> /home/iz/sel4test/kernel/include/kernel/thread.h:130 in function commitTime
> ```
> 
> Which is the call on line 75 of src/smp/ipi.c just after switchToIdleThread() in ipiStallCoreCallback().

Okay, turns out that before this change, `ksCurTime` wasn't updated at all for IPIs, so time accounting is incorrect in master.

I believe some assertions in `commitTime` are wrong.

- There is no guarantee that `ksCurSC` has `MIN_BUDGET` budget left, as `commitTime` may be called at any time. This invalidates the `assert(refill_sufficient(NODE_STATE(ksCurSC), NODE_STATE(ksConsumed)));`. It probably should be `assert(refill_capacity(NODE_STATE(ksCurSC), NODE_STATE(ksConsumed)));`, which matches the comment.

- The first `refill_ready` assertion is correct, as otherwise an unready SC would be running and that would be weird.
- The second `refill_sufficient` assertion is correct, as the head has just been updated.
- The second `refill_ready` assertion is incorrect, as `refill_budget_check` may move `rTime`  into the future.

Why do we have both slightly different `commitTime` and `chargeBudget` functions?

pingerino:
>  There is no guarantee that ksCurSC has MIN_BUDGET budget left, as commitTime may be called at any time. This invalidates the assert(refill_sufficient(NODE_STATE(ksCurSC), NODE_STATE(ksConsumed)));. It probably should be assert(refill_capacity(NODE_STATE(ksCurSC), NODE_STATE(ksConsumed)));, which matches the comment.

One of the invariants of the MCS kernel is that any thread in the kernel has to have enough budget to get out again (kernel WCET). If this isn't sufficient, we don't do whatever the thread wanted to do, but simluate a timer tick instead. commitTime should not ever be called "at any time". 

IIRC commitTime is called when we *know* there's enough budget to charge the SC without having to do any shuffling of sporadic replenishments or simulating a timer tick. chargeBudget is for the other case. 

Going on memory here though ;)

Indanz:
> IIRC commitTime is called when we know there's enough budget to charge the SC without having to do any shuffling of sporadic replenishments or simulating a timer tick. chargeBudget is for the other case.

Thanks, that explains everything. I'll check if this is true for all callers of commitTime. It seems that the commitTime in the IPI path I added with the bugfix for the user-after-free of SC's was wrong and should have been a chargeBudget instead.

include/kernel/thread.h
```diff
@@ -307,5 +306,7 @@ void awaken(void);
 /* Place the thread bound to this scheduling context in the release queue
  * of periodic threads waiting for budget recharge */
 void postpone(sched_context_t *sc);
+#else
+#define updateTimestamp()
```
axel-h, line 310: Can you leave a comment that we have this as a macro define to nothing because this avoids having to update the proofs. Otherwise this will be another place in the code that looks a bit odd when having a pure C view.

axel-h: @lsf37: can't we make this `#define updateTimestamp() ((void)(0))` (or `do {} while false`) because the proofs can handle this as a no-op? Then we'd have statement here that requires/consume a `;` and looks like a function. It's also what we are doing with `printf()` in https://github.com/seL4/seL4/blob/812c0c2a395e3179e7119d7b9df4f153a19132a2/include/machine/io.h#L110-L117

Indanz: I still don't know if semicolons need manual proof updates for non-MCS or not, so I went with @kent-mcleod's suggestion and stuffed everything in the `config_ternary`.

Indanz:
I added a test for this, see https://github.com/seL4/sel4test/pull/76.

src/smp/ipi.c
```diff
@@ -35,7 +35,8 @@ void ipiStallCoreCallback(bool_t irqPath)
         SCHED_ENQUEUE_CURRENT_TCB;
         switchToIdleThread();
 #ifdef CONFIG_KERNEL_MCS
-        commitTime();
+        /* This is safe because the Idle thread is not schedulable: */
+        chargeBudget(NODE_STATE(ksConsumed), false);
```
kent-mcleod, line 39: I can't convince myself that this change is correct either. chargeBudget potentially has to deliver a timeout fault to the thread associated with the ksCurSC if it's run out of budget.  However changing to the IdleThread in the line before means that the ksCurSC isn't the SC of the ksCurThread and so chargeBudget cannot deliver a timeout fault to the correct thread's fault handler.

Indanz: `chargeBudget` is called with `canTimeoutFault == false` here, so it wil never deliver a timeout fault anyway. And as we're handling an IPI I don't think we want to do a timeout fault, as this schedule has nothing to do with timing out.

The whole second if-block will be skipped because the idle thread is not schedulable. Then the only difference with commitTime is a slight detail on how round-robin tasks are charged (ksConsumed versus whole rAmount). If it would go into the second if-block it would trigger `assert(NODE_STATE(ksCurThread)->tcbSchedContext == NODE_STATE(ksCurSC))`.

kent-mcleod: > chargeBudget is called with canTimeoutFault == false here, so it wil never deliver a timeout fault anyway. And as we're handling an IPI I don't think we want to do a timeout fault, as this schedule has nothing to do with timing out.

In the case where the SC is out of budget, the thread it's bound to needs to either be placed in the release queue or generate a timeout fault. Instead, because the thread is added back to the scheduler (on line 35) it could be scheduled again without sufficient budget to be able to leave the kernel and reenter which violates the invariant that all thread's in the scheduler queue must have sufficient budget.

kent-mcleod: It may be worth moving this commit to a separate PR if it's unrelated to moving `updateTimestamp()`?  Is the new test in [this comment](https://github.com/seL4/seL4/pull/847#issuecomment-1136100190) only failing with the `updateTimestamp()` changes, or does it fail for MCS in the main branch too?

PR #849, opened 2022-05-19
MCS, ARM: Introduce TIMER_OVERHEAD_TICKS

Indanz:
For ARM currently TIMER_PRECISION exists, but that is in microseconds
and not fine-grained enough.

This is needed to make periodic tasks synchronous with the system clock.
If this value is zero every period will be extended with the overhead of
taking an interrupt and reading the system clock. To avoid this drift,
the configured value should be set to at least the average overhead.

See also issue #844.


kent-mcleod (approved):

lsf37:
No problems with this patch from the MCS verification side, the preprocess failure is benign.

axel-h:
Does is make sense to support `TIMER_PRECISION` for x86 and RISC-V also, as the same code seems to exist there? The default would be 0 anyway, so there is no impact (except for verification). Seems ARM is the platform this is most tested on. Having a general test system that shows the behavior would be nice - can you share a stripped down version of your code?

I assume this is a minimal invasive fix for now that can be merged. Otherwise addressing the concerns you raise could be addressed in a larger change. Since `setNextInterrupt()` is the only user of this, a dedicated value to be applied here could also be given directly to reduce the overall complexity?

kent-mcleod:
> I assume this is a minimal invasive fix for now that can be merged. Otherwise addressing the concerns you raise could be addressed in a larger change. Since `setNextInterrupt()` is the only user of this, a dedicated value to be applied here could also be given directly to reduce the overall complexity?

Probably an alternative fix to consider would be changing TIMER_PRECISION into ticks instead of `us` for more granularity.

axel-h:
> Probably an alternative fix to consider would be changing TIMER_PRECISION into ticks instead of `us` for more granularity.

I wonder if doing this is better than having a second configuration value now, how big would the impact be here in verification?

Indanz:
> Probably an alternative fix to consider would be changing TIMER_PRECISION into ticks instead of us for more granularity.

I did consider that, but that would break backward compatibility. Adding TIMER_OVERHEAD_TICKS to x86 and riscv and removing TIMER_PRECISION  effectively achieves this.

The current `getTimerPrecision` is a bit misleading in how it's used and I wonder if the x86 and riscv hard-coded 1 us is correct.

(I don't think any of this impacts verification because it's in hardware specific platform code.)

axel-h:
> The current `getTimerPrecision` is a bit misleading in how it's used and I wonder if the x86 and riscv hard-coded 1 us is correct.

I have been wondering about this also and assumed on RISC-V it could be just copy/paste. That is why I asked if you could helper me with setting up a test system, so I can run a test on the RISC-V hardware I have.

Indanz:
> That is why I asked if you could helper me with setting up a test system, so I can run a test on the RISC-V hardware I have.

Ah, sorry, missed this.

>  can you share a stripped down version of your code?

Sadly not, there is too much setup and system code involved. I can write a test for sel4test though. It will consist of a periodic task with highest priority being started first and stopped last after all other tests have run, and seeing if it ran the number of expected periods compared to the configured period. All the task would do is increment a counter and call seL4_Yield().

Issue #498, opened 2021-08-20
sc_sporadic should imply sc_active


mktnk3:
#476 bundles up the `sc_sporadic` check and `sc_active` together, to shortcut the verification process temporarily, with the understanding that `sc_sporadic` should imply `sc_active`. However, there seems to be some edge cases where this does not hold (see #496). This probably suggests potential bug somewhere. One place that `scRefillMax` is set to 0 is deletion in `finaliseCap` where we don’t modify the sporadic flag, but I am not too sure if it is actually relevant.

Apparently, adding the assert fails debug_MCS configuration, but not verification_MCS or release_MCS.

lsf37:
The corresponding sel4test log for this failure is

```
  Running test SCHED_CONTEXT_0003 (Basic api_sc_bind/UnbindObject testing)
  <<seL4(CPU 0) [decodeSchedContext_Bind/126 T0xffffff807c23b400 "46" @43c7d0]: SchedContext_Bind: invalid cap.>>
  seL4 failed assertion 'sc == NULL || sc_active(sc)' at /github/workspace/kernel/include/kernel/sporadic.h:164 in function sc_sporadic
  halting...
  Kernel entry via Syscall, number: 1, Call
  Cap type: 22, Invocation tag: 31
```

See also [on the test itself](https://github.com/seL4/seL4/runs/3378662043#step:5:3842)

mktnk3:
Curtis agrees that `finaliseCap` is the most likely culprit. We need to set `scSporadic` to 0 before setting `scRefillMax` to 0 here:
https://github.com/seL4/seL4/blob/master/src/object/objecttype.c#L231-L232

kent-mcleod:
> Apparently, adding the assert fails debug_MCS configuration, but not verification_MCS or release_MCS.

asserts are only asserted in debug builds.  Release or verification builds don't compile with asserts active.

lsf37:
What's the status of this one? The failing assertion was removed in #496, because the invariant conjectured there is currently not true.

Do we need it to be true? What happens if it is not?

Has anyone tried the proposed fix of resetting `scSporadic` in `finaliseCap`? It does seem to make sense to reset it in any case.

lsf37:
[...]

PR #852, opened 2022-05-20
Remove redundant sc_active checks

Indanz:
sc_sporadic() already calls sc_active.

### DIFF
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -91,7 +91,7 @@ void restart(tcb_t *target)
         cancelIPC(target);
 #ifdef CONFIG_KERNEL_MCS
         setThreadState(target, ThreadState_Restart);
-        if (sc_sporadic(target->tcbSchedContext) && sc_active(target->tcbSchedContext)
+        if (sc_sporadic(target->tcbSchedContext)
             && target->tcbSchedContext != NODE_STATE(ksCurSC)) {
             refill_unblock_check(target->tcbSchedContext);
         }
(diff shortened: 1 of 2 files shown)


axel-h:
Should there be asserts then to ensure/state this always hold?

Indanz:
> Should there be asserts then to ensure/state this always hold?

I don't think that's necessary, `sc_sporadic` does:
```
static inline bool_t sc_sporadic(sched_context_t *sc)
{
    return sc != NULL && sc_active(sc) && sc->scSporadic;
}
```

axel-h:
ok, looks like a good cleanup then.

lsf37:
@corlewis could you have a look if updating the specs for this would be much proof update work? `sc_active` is indeed already called in `sc_sporadic`, so there shouldn't really be much proof state change from this.

axel-h (approved):

kent-mcleod (approved):

corlewis:
> @corlewis could you have a look if updating the specs for this would be much proof update work? `sc_active` is indeed already called in `sc_sporadic`, so there shouldn't really be much proof state change from this.

I don't think that this will require any proof update work. Currently, the specifications have actually gone in the opposite direction, explicitly separating the `sc_sporadic` and `sc_active` tests. However, this should be easily shown to be equivalent during refinement.

I don't think that it is worth refactoring the specifications to better match the C. The long-term goal is still to remove the `sc_active` test, by proving a new invariant and resolving https://github.com/seL4/seL4/issues/498.

corlewis (approved):

Issue #855, opened 2022-05-23
Is CDT a global tree in kernel or per thread tree in kernel?


FinnZhang:
There are some words below in the seL4 manual.
> that seL4 maintains a capability derivation tree (CDT) in which it tracks the relationship between these copied capabilities and the originals

As the words says, **seL4** maintains a CDT.  That means there is only one global CDT in the seL4 kernel. But I can't find any global CDT exist or maybe I misunderstand meaning of CDT.
Someone says CDT is a tree at per thread level but I disagree. So I am so confused about the CDT without more document.
Could you please answer my question or give me more references about CDT? Thanks a lot.

Indanz:
From how I understand it, it's a single, global tree, with the root server's TCB having the root node of this tree as its associated CSpace,

FinnZhang:
> From how I understand it, it's a single, global tree, with the root server's TCB having the root node of this tree as its associated CSpace,

Thanks for your reply. Would you mind providing me some references about CDT data structure principle ?

Indanz:
I've got my information from the [seL4 manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf), see figure 3.1. But it helps to actually use seL4 to write programs, because then you encounter all the details and you get an overall view of how it all ties together. The CDT is about object creation and lifetime.

Keep in mind that seL4 delegates _all_ memory management to user space (with the only exception being the startup task), so there is nothing else than what is described in the manual. Every object reference you have to configure yourself, so you know exactly how everything ties together.

Although the capability derivation tree is a tree structure, all CSpaces together are one big graphs and not necessarily a tree. The starting node (or "root") of that graph is defined per TCB as its associated CSpace. How far each task can traverse the graph is defined by the graph's structure and the CNodes' guard sizes.

Conceptually the CDT is managed as a subtree of the global CSpace graph.

This is introduces limitations that are not always documented, like you're not allowed to put a cap to an untyped region into a CNode derived from that untyped region.

FinnZhang:
> I've got my information from the [seL4 manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf), see figure 3.1. But it helps to actually use seL4 to write programs, because then you encounter all the details and you get an overall view of how it all ties together. The CDT is about object creation and lifetime.
> 
[...]

PR #859, opened 2022-05-25
Allow copy of untyped regions with children

Indanz:
The new cap is copied from the old one and thus it has the correct
FreeIndex. Nothing else is needed for correctness, as long as the
source cap gets marked full so that free space accounting is only
done on the newest cap.

### DIFF
--- a/src/object/objecttype.c
+++ b/src/object/objecttype.c
@@ -79,12 +79,8 @@ deriveCap_ret_t deriveCap(cte_t *slot, cap_t cap)
         break;
 
     case cap_untyped_cap:
-        ret.status = ensureNoChildren(slot);
-        if (ret.status != EXCEPTION_NONE) {
-            ret.cap = cap_null_cap_new();
-        } else {
-            ret.cap = cap;
-        }
+        ret.status = EXCEPTION_NONE;
+        ret.cap = cap;
         break;
 
 #ifndef CONFIG_KERNEL_MCS


Indanz:
Minor improvement, feel free to drop it if it complicates proofs.

pingerino:
I am fairly sure we don't allow this in order to limit WCET and the depth of the derivation tree.

mbrcknl:
My [comment on #855](https://github.com/seL4/seL4/issues/855#issuecomment-1138029800) might also be relevant.

Indanz:
> I am fairly sure we don't allow this in order to limit WCET of revoke by limiting the depth of the derivation tree.

I don't see how, as it doesn't change the derivation tree as far as I can tell. Is the CDT stored in the caps or in the untyped region itself? If it's stored in the caps then it would make sense, but then why are moves allowed?

> I don't quite follow the situation that requires you to copy, then allocate, then move. Can you give a little more detail about what step you think should be unnecessary?

What I want is:
- Allocate objects from an untyped region.
- Copy the UT cap to pass it on to the just created process.
- When closing down, revoke the UT to clean everything up.

Now I need to:
- Make a temporary copy of the UT cap.
- Allocate objects from the temp cap.
- Move the temporary cap to pass it on to the just created process.
- When closing down, revoke the UT to clean everything up.

As far as I can tell the end result is exactly the same and the limitation is artificial.

> A retype from the copy would reset the copy's free index without deleting the existing children of the original untyped, and you'd end up using the same memory twice. To avoid that, you'd need the copy's base pointer to start from the free region of the original untyped

But all the information is copied from the old cap to the new one, how can the base pointer be wrong?

mbrcknl:
> What I want is:
> 
> * Allocate objects from an untyped region.
> * Copy the UT cap to pass it on to the just created process.
> * When closing down, revoke the UT to clean everything up.
> 
> Now I need to:
> 
> * Make a temporary copy of the UT cap.
> * Allocate objects from the temp cap.
> * Move the temporary cap to pass it on to the just created process.
> * When closing down, revoke the UT to clean everything up.
> 
> As far as I can tell the end result is exactly the same and the limitation is artificial.
> 
> > A retype from the copy would reset the copy's free index without deleting the existing children of the original untyped, and you'd end up using the same memory twice. To avoid that, you'd need the copy's base pointer to start from the free region of the original untyped
> 
> But all the information is copied from the old cap to the new one, how can the base pointer be wrong?

Oh, wow, I think I see what you're saying.

I was imagining that you would get a CDT in which the caps allocated from the original UT were *siblings* of the copy of the UT cap. If that were the case, then the check for CDT children on the next retype would falsely conclude that the memory can be reclaimed.

But I think what actually happens is that the caps allocated from the original UT become CDT *children* of the copy of the UT cap. In other words, the UT copy gets inserted between the original UT and its existing CDT children. In that case, I tentatively agree that everything should be fine. That's actually kinda neat.

However, the proof update might not be simple as the C update. The simplicity of the C patch comes from the way the CDT is implemented as a linked list in depth-first order, with the tree structure inferred as necessary from various other bits of information. However, the spec models the CDT as a partial mapping from child to parent, so it would need to be explicit about reparenting the children. We'd then need to prove that this is refined by the simple list insertion in the C. I'm actually wondering whether the `ensureNoChildren` check was added specifically to avoid that proof.

I'll leave this open for a bit in case anyone has further thoughts, but I suspect the proof work won't be a high priority.

lsf37:
I don't think the change achieves the same result as the sequence of operations described here:

> - Make a temporary copy of the UT cap.
> - Allocate objects from the temp cap.
> - Move the temporary cap to pass it on to the just created process.
> - When closing down, revoke the UT to clean everything up.

In that sequence, the temp UT cap has the caps to the allocated objects as direct descendants, and the original UT cap is the parent of the temp cap. Revoking the parent revokes everything.

This change would make a child of the half-allocated cap, and IIUC inherit the descendants of the source cap, i.e. inserting itself in the tree between the source and its children. This is problematic for revocation, because revocation of the child doesn't reset the free counter of the parent. This would introduce an inconsistency between what the parent cap thinks is allocated and the child cap thinks is allocated. If you now delete the child, you have an inconsistency between (missing) descendants of the source cap and its allocation status. (If you always revoke the parent, that is fine, but you can't rely on the user doing that).

More generally, making a copy of an UT cap when already allocating would either have to create a parent, a child, or a sibling. In the current implementation UT caps can't have siblings, and it wouldn't help for the scenario described. Creating a parent would be contrary to what a copy operation is supposed to do, and creating a child creates problems with revocation consistency.

I think the operations you are doing now express fairly nicely what the derivation relationship is supposed to achieve. You do have one syscall more than you might want, but you're not creating unnecessary caps or other overhead.

Indanz:
I needed to read the source code more to understand what you all are saying. When doing the pull request I missed that `cte_t` is part of each cap and not part of the object the cap refer to. All life-time management of both objects and capabilities is handled with the `mdb_node_t` linked list in the `cte_t`. And this is the main thing that implements the CDT.

So, like @mbrcknl observed, that copying of non-empty UT actually works is just a coincidence of how the CDT is implemented and not trivial to proof. I assumed the proofs would not need any updating, but that's clearly not true because of the implicit reparenting happening.

I think it makes sense to just close this pull request, as the change is trivial and the real work in on the verification side, all for minimal gain.

mbrcknl:
I was curious, so I just spent a couple of hours looking at what proof update would be needed, and got very much stuck in the mud. It would take quite some effort, so yes, I think we'll leave it be. :-)

Issue #854, opened 2022-05-23
Do not use cross-node ksCurTime


Indanz:
Replace `NODE_STATE_ON_CORE(ksCurTime)` calls with `NODE_STATE(ksCurTime)`.

The code using ksCurTime assumes that `ksCurTime` is up-to-date, but this assumption is wrong for `ksCurTime` of other CPU cores.

refill_new(), refill_ready(), refill_update() and refill_unblock_check() all use ksCurTime via `NODE_STATE_ON_CORE`.

The implications of using `NODE_STATE(ksCurTime)` is that all clocks must be synchronous:
- Riscv is okay: The specification states: "The real-time clocks of all hardware threads in a single user application should be synchronized to within one tick of the real-time clock."
- x86 okay if not ancient when Invariant TSC is supported. (Constant TSC is probably also good enough for seL4 if cores are not shut off via C/T/P-states.)
- aarch64 is okay.
- arm32: arm_global.h is okay. Exynos timer seems okay. am335x and omap3430 are single-core.

See also issue #844.

kent-mcleod:
This seems like a necessary fix.

Indanz:
Fix is merged.

Issue #513, opened 2021-08-25
boot printing leads to assertion failure on SMP+MCS


lsf37:
Commit 43667a0 from PR #438 which looks entirely harmless, leads to an assertion failure immediately after dropping to user space for SMP + MCS on the sabre board. It's not a fluke, the assertion triggers reliably, and did not trigger in the commit before that.

See also the [test logs](https://github.com/seL4/seL4/runs/3422369363#step:5:6007):
```
  Booting all finished, dropped to user space
  seL4 failed assertion 'refill_ready(thread->tcbSchedContext)' at /github/workspace/kernel/src/kernel/thread.c:447 in function switchToThread
  halting...
  Kernel entry via Interrupt, irq 0
```

There is strange looking test output before this snippet, but that is just all CPU cores printing
```
DIDRv: 3, armv 70, coproc baseline only? no.
CPU is in secure mode. Enabling debugging in secure user mode.
```
at the same time and unrelated to this commit.

Indanz:
Does this still happen?

midnightveil:
> Kent points out that sabre is using a global timer, which is a problem for SMP + MCS in general. It's possible that cores overwrite each others deadlines this way. Should not be a problem if the new deadline is shorter, but might be a problem if the new deadline is later than the old one needed to be.

I'll just add a comment here: this shouldn't be the issue; the 'Global Timer'

> The global timer is accessible to all Cortex-A9 processors. Each Cortex-A9 processor has a 64-bit comparator that is used to assert a private interrupt when the global timer has reached the comparator value. All the Cortex-A9 processors in a design use the banked ID, ID27, for this interrupt. ID27 is sent to the Interrupt Controller as a Private Peripheral Interrupt.
>
> — [About the Global Timer; Cortex-A9 MPCore Technical Reference Manual r2p0](https://developer.arm.com/documentation/ddi0407/e/global-timer--private-timers--and-watchdog-registers/about-the-global-timer?lang=en)

The comparator registers and interrupts are banked per CPU, but share the same counter registers.

PR #863, opened 2022-05-30
Do not use cross-node ksCurTime

Indanz:
The code using ksCurTime assumes that ksCurTime is up-to-date,
but this assumption is wrong for ksCurTime of other CPU cores.
Those can be quite some time in the past.

The implications of using NODE_STATE(ksCurTime) is that clocks
on all cores must be synchronous:

- Riscv is okay: The specification states: "The real-time clocks
  of all hardware threads in a single user application should be
  synchronized to within one tick of the real-time clock."
- x86 okay if not ancient when Invariant TSC is supported.
- aarch64 is okay.
- arm32: arm_global.h is okay. Exynos timer seems okay. am335x and
  omap3430 are single-core.

See also #854.


kent-mcleod (approved):

pingerino (approved):

src/kernel/sporadic.c
```diff
@@ -167,7 +167,7 @@ void refill_new(sched_context_t *sc, word_t max_refills, ticks_t budget, ticks_t
     /* full budget available */
     refill_head(sc)->rAmount = budget;
     /* budget can be used from now */
-    refill_head(sc)->rTime = NODE_STATE_ON_CORE(ksCurTime, core);
+    refill_head(sc)->rTime = NODE_STATE(ksCurTime);
```
axel-h, line 170: Now that `core` is not used here any longer, It seems there is no reason to pass it to `refill_new()` at all? Removing this could happen in an additional commit, will also remove some of SMP-conditional compilation from the code

lsf37 (approved):
I'm fine with this. Feels a bit strange that we can assume time to be in sync over all cores, but if the hardware achieves that, no reason not to use it. (And I agree that reading the time from a different core can be way out of date, so this is definitely an improvement).

src/kernel/sporadic.c
Indanz: Well spotted. I also removed the REFILL_NEW macro.

Indanz:
> Feels a bit strange that we can assume time to be in sync over all cores, but if the hardware achieves that, no reason not to use it.

Agreed. I think we should add a test for this, either in sel4test or in debug builds. I'll have a look if there is a good spot for this somewhere during bootup.

src/arch/x86/kernel/smp_sys.c
```diff
@@ -126,6 +136,11 @@ VISIBLE void boot_node(void)
         fail("boot_node failed for some reason :(\n");
     }
 
+#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
+    clock_sync_test = getCurrentTime();
+    /* Make sure clock_sync_test is set before smp_aps_index */
+    asm volatile("" ::: "memory");
```
axel-h, line 142: Use `__atomic_signal_fence(__ATOMIC_ACQ_REL)`, because this is not just a compile barrier, but also a CPU/Pipeline barrier?

axel-h: Then better leave this explicit barrier out and change the comment to say: x86 guarantees global order, volatile qualifier prevents compiler reordering.

src/arch/arm/kernel/boot.c
```diff
@@ -305,6 +313,18 @@ BOOT_CODE static void release_secondary_cpus(void)
...
         /* perform a memory release+acquire to get new values of ksNumCPUs */
         __atomic_signal_fence(__ATOMIC_ACQ_REL);
+#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
+        /* Test whether clocks are synchronised across nodes: */
+        ticks_t t = getCurrentTime();
+        for (int i = 1; i < CONFIG_MAX_NUM_NODES; ++i) {
+            if (clock_sync_test[i] != 0) {
+                ticks_t clock_diff = t - clock_sync_test[i];
+
+                assert(clock_diff >= 0 && clock_diff <= getTimerPrecision());
```
axel-h, line 323: Does this wok, with `ticks_t` an aliad of `uint64_t` this can't be negative? Maybe use:
```suggestion
                assert((t >=clock_sync_test[i]) && (t - clock_sync_test[i] <= getTimerPrecision()));
```

src/arch/x86/kernel/smp_sys.c
Indanz: Done.

Indanz:
If the changes are otherwise approved, could someone trigger the SMP HW tests to verify if everything is okay?

src/arch/riscv/kernel/boot.c
```diff
@@ -159,6 +162,11 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(word_t hart_id, word_t co
     init_cpu();
     NODE_LOCK_SYS;
 
+#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
+    clock_sync_test[getCurrentCPUIndex()] = getCurrentTime();
+    /* Make sure clock_sync_test is set before ksNumCPUs */
+    __atomic_signal_fence(__ATOMIC_ACQ_REL);
```
axel-h, line 168: If I get the documentation right,  `__atomic_signal_fence()`does not guarantee a pipeline barrier instruction is created, it's just about the compiler's instruction generation order. So this must be `__atomic_thread_fence(__ATOMIC_ACQ_REL)` then, as we have to ensure the new value in `clock_sync_test` is visible before (or at the same time) of the change in `ksNumCPUs`.

Another observation is, that currently the check runs once we have obtained the BKL (from `NODE_LOCK_SYS` above) I think we could run the clock sync test even before trying to get the BKL, because it's completely unrelated to it and intended to run in parallel on all cores. Having the BKL It does not harm, just adds some more delay before the core's clock value is available.

Indanz: I think you are right about that `__atomic_signal_fence` should be replaced with `__atomic_thread_fence`. However, this should probably also happen for the existing call. I'll replace all calls to be on the safe side.

Currently the check only runs on core 0 without the BKL held, the other cores only store their timestamp (with the BLK held, but that does not matter). The test is done while core 0 is busy looping and waiting for the other cores to finish startup. The time between changing `ksNumCPUs` and for that change to propagate across nodes is the only measuring overhead.

I tried doing the test the other way round, with core 0 making a timestamp and then doing the compare in the launched cores, but that adds too much cpu init overhead for correct time measurements.

src/arch/arm/kernel/boot.c
```diff
@@ -274,6 +277,11 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
     NODE_LOCK_SYS;
 
+#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
+    clock_sync_test[getCurrentCPUIndex()] = getCurrentTime();
+    /* Make sure clock_sync_test is set before ksNumCPUs */
+    __atomic_thread_fence(__ATOMIC_ACQ_REL);
```
axel-h, line 283: A barrier that ensures writes must finish should sufficient here,
```suggestion
        /* Ensure the update of clock_sync_test is visible before the update of ksNumCPUs */
        __atomic_thread_fence(__ATOMIC_RELEASE);
```

src/arch/arm/kernel/boot.c
```diff
@@ -304,7 +312,19 @@ BOOT_CODE static void release_secondary_cpus(void)
     /* Wait until all the secondary cores are done initialising */
     while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
         /* perform a memory release+acquire to get new values of ksNumCPUs */
-        __atomic_signal_fence(__ATOMIC_ACQ_REL);
+        __atomic_thread_fence(__ATOMIC_ACQ_REL);
```
axel-h, line 315: Can you make the change `__atomic_signal_fence`-> `__atomic_thread_fence` in a separate commit, because it fixes a potential `ksNumCPUs` update problem (ensure barriers instruction are created), which is unrelated to the clock sync test.

axel-h: Seems a `__ATOMIC_ACQUIRE` is sufficient then also, as we just have to ensure that every load is executed

Indanz: Normally I would agree that it should have been a separate commit. I considered that, but I deemed the problem to be too small to warrant it's own commit.

Because `ksNumCPUs` is a global variable there won't be any chance of optimising the access away, and the only difference is a missing barrier. The barrier is only necessary if there is data ordering, but there isn't any other data involved to order against until I introduced `clock_sync_test`.

Indanz: > Seems a __ATOMIC_ACQUIRE is sufficient then also

Probably, but that change does not warrant the slight chance of introducing any future bugs caused by other code changes IMHO.

I think the code should be rewritten to C11 standard atomics anyway, if you want to really improve it. `ksNumCPUs` is being used as a sort of lock, if the store and loads were done with acquire and release, then the `__atomic_thread_fence` after setting `clock_sync_test` could disappear altogether.

src/arch/arm/kernel/boot.c
```diff
@@ -304,7 +312,19 @@ BOOT_CODE static void release_secondary_cpus(void)
...
+#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
+        /* Test whether clocks are synchronised across nodes: */
+        ticks_t t = getCurrentTime();
+        for (int i = 1; i < CONFIG_MAX_NUM_NODES; ++i) {
+            ticks_t tc = clock_sync_test[i];
+
+            if (tc != 0) {
+                assert(tc <= t && t <= tc + getTimerPrecision());
```
axel-h, line 323: I have some doubt that there is a 100% guarantee that `tc <= t`, because this sequence could happen
```
    Core 1                              Core 0
                                     t = getCurrentTime()
t1 = getCurrentTime() (= t + n)
clock_sync_test[1] = t1
                                     tc = clock_sync_test[1] (= t + n)
                                     assert, since tc is higher than t
```
This might be unlikely to happen, but can't be rules out. Should this check better allow `t` to be in the range `tc - getTimerPrecision()/2` to `tc + getTimerPrecision()/2`? Another way might be taking a snapshot, then take the time and then check it against the snapshot. The assumption here is, that a timestamp taken after making the snapshot must be greater or equal than anything in the snapshot 
```
#if defined(CONFIG_KERNEL_MCS) && defined(CONFIG_DEBUG_BUILD)
        ticks_t snapshot[CONFIG_MAX_NUM_NODES];
        for (int i = 1; i < CONFIG_MAX_NUM_NODES; ++i) {
            snapshot[i] = clock_sync_test[i];
        }
        /* ensure the snapshot is taken before we get the time below */
        __atomic_thread_fence(__ATOMIC_ACQUIRE);
        ticks_t t = getCurrentTime(); /* this cannot be less than an snapshot value */
        for (int i = 1; i < CONFIG_MAX_NUM_NODES; ++i) {
            ticks_t tc = snapshot[i];
            if (tc != 0) {
                assert(tc <= t && t <= tc + getTimerPrecision());
                clock_sync_test[i] = 0;
            }
        }
```

Indanz: Well spotted. It's a fairly theoretical problem though: As `ksNumCPUs` is only modified while holding the BLK, node 0 has time equal to the BLK hold time to do the check. But it does mean that the data ordering with `ksNumCPU`  is forced and unnecessary, so all the barriers could go away if this race is fixed.

A simpler fix would be to just always take a timestamp within the `if (tc != 0)`. I'll have a look at improving the code.

axel-h: > A simpler fix would be to just always take a timestamp within the if (tc != 0). I'll have a look at improving the code.

I have also considered. The downside of this would be, that the timestamp is "drifting" away with each call, but practically that delta might be negligible if `getTimerPrecision()`is  large enough. On the other side it has to be large enough to cope with the situation that we miss an update of `clock_sync_test`, reach the end of the loop and then start over again, which calls `getCurrentTime()` again then anyway. Bottom line would be that we assume calling `getCurrentTime()` is fast and cheap, so it does no matter how often we do it.

src/arch/arm/kernel/boot.c
```diff
@@ -303,8 +304,12 @@ BOOT_CODE static void release_secondary_cpus(void)
 
     /* Wait until all the secondary cores are done initialising */
     while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
-        /* perform a memory release+acquire to get new values of ksNumCPUs */
-        __atomic_signal_fence(__ATOMIC_ACQ_REL);
+        /* perform a memory acquire to get new values of ksNumCPUs, release for ksCurTime */
+        __atomic_thread_fence(__ATOMIC_ACQ_REL);
+#if defined(CONFIG_DEBUG_BUILD) && defined(CONFIG_KERNEL_MCS)
+        /* Test whether clocks are synchronised across nodes: */
+        NODE_STATE(ksCurTime) = getCurrentTime();
```
axel-h, line 311: It's nitpicking, but there must be a `__atomic_thread_fence(__ATOMIC_RELEASE)` afterwards if we want to ensure the value becomes visible exactly here in the last loop iteration also. Otherwise, the last iteration has no pipeline barrier, as it just executed the check in the loop header and then leaves the function

Indanz: That's why `clock_sync_test()` is called before incrementing `ksNumCPUs`, to guarantee that `ksCurTime` on node 0 will keep changing.

axel-h: I'd still recommend using this order, ie insert the new code before the barrier.

```
while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
#if defined(CONFIG_DEBUG_BUILD) && defined(CONFIG_KERNEL_MCS)
        /* Test whether clocks are synchronised across nodes: */
        NODE_STATE(ksCurTime) = getCurrentTime();
#endif
        /* ensure new values of ksNumCPUs are read and ksCurTime is written */
        __atomic_thread_fence(__ATOMIC_ACQ_REL);
}

File: src/kernel/boot.c [hunk 9]
```diff
@@ -553,6 +553,24 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
     return tcb;
 }
 
+#ifndef clock_sync_test
+BOOT_CODE void clock_sync_test(void)
+{
+    ticks_t t, t0;
+
+    t = NODE_STATE_ON_CORE(ksCurTime, 0);
```
axel-h, line 561: Maybe add an `assert(0 != getCurrentCPUIndex())` here, as this must be called on secondary cores only, otherwise it deadlocks.

include/kernel/boot.h
```diff
@@ -135,3 +135,10 @@ static inline BOOT_CODE pptr_t it_alloc_paging(void)
 
 /* return the amount of paging structures required to cover v_reg */
 word_t arch_get_n_paging(v_region_t it_veg);
+
+#if defined(CONFIG_DEBUG_BUILD) && defined(ENABLE_SMP_SUPPORT) && defined(CONFIG_KERNEL_MCS)
+/* Test whether clocks are synchronised across nodes */
+BOOT_CODE void clock_sync_test(void);
+#else
+#define clock_sync_test()
```
axel-h, line 143: @lsf37: can't we make this #define clock_sync_test() ((void)(0)) (or do {} while false) because the proofs can handle this as a no-op? Then we'd have statement here that requires/consume a ; and looks like a function

src/arch/arm/kernel/boot.c
Indanz: Okay, I think I missed that I could expand the individual test results.

src/kernel/boot.c
```diff
@@ -553,6 +553,24 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
     return tcb;
 }
 
+#ifndef clock_sync_test
```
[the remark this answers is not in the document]
axel-h: Effectively this seems to be something that can be enabled or disabled. So, why not do this in `include/kernel/boot.h`:
```
#if defined(CONFIG_DEBUG_BUILD) && defined(ENABLE_SMP_SUPPORT) && defined(CONFIG_KERNEL_MCS)
#define ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
#endif
```
And then it would simply be `#ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT` everywhere. One day, this could be made a config option in CMake even, so one could also enable this for release kernels have have them halt if this fails.

Indanz:
Why aren't there more HW Run test that tests SMP + MCS + DEBUG? Some do SMP + MCS + RELEASE, but not debug. Or only SMP or MCS, but not both.

- HIFIVE_debug_SMP_MCS_gcc_64 fails with:
```
  clock_sync_test[3]: t0 = 23715653, t = 23715655, td = 2
  seL4 failed assertion 't0 <= t && t <= t0 + usToTicks(1)' at /github/workspace/kernel/src/kernel/boot.c:571 in function clock_sync_test
```
The system clock is 1 MHz, so it took at least 1 us for the `ksCurTime` value to propagate from one core to the next, which seems ridiculous. But if the clock is allowed to be almost 1 us out-of-sync, there isn't much margin left for the measurement.

- ODROID_XU4_release_hyp_clang_32 fails with `[[Timeout]]`, which seems unrelated to this pull request.

- [PC99_debug_SMP_MCS_gcc_32 ](https://github.com/seL4/seL4/runs/6689978186?check_suite_focus=true) failed once with:
```
  Starting node #1 with APIC ID 2
  clock_sync_test[1]: t0 = 74135544547, t = 74135544564, td = 17
  Starting node #2 with APIC ID 4
  clock_sync_test[2]: t0 = 74164562456, t = 74164562456, td = 0
  Starting node #3 with APIC ID 6
  clock_sync_test[3]: t0 = 74193279059, t = 74193279045, td = 18446744073709551602
  seL4 failed assertion 't0 <= t && t <= t0 + usToTicks(1)' at /github/workspace/kernel/src/kernel/boot.c:570 in function clock_sync_test
```
Which means we also have to apply the margin to t0 (except if the latest code changes fixed this somehow).

Pushed hopefully the final version...

kent-mcleod:
> Why aren't there more HW Run test that tests SMP + MCS + DEBUG? Some do SMP + MCS + RELEASE, but not debug. Or only SMP or MCS, but not both.

Because MCS + SMP kept failing tests I expect. I wonder if in cases where RELEASE is enabled but debug not it's because of the asserts not being present:
- https://github.com/seL4/seL4/issues/513 
- https://github.com/seL4/ci-actions/pull/146

lsf37:
> > Why aren't there more HW Run test that tests SMP + MCS + DEBUG? Some do SMP + MCS + RELEASE, but not debug. Or only SMP or MCS, but not both.
> 
> Because MCS + SMP kept failing tests I expect. 

I'm not sure that is the reason. The current set of tests is just what was there in Bamboo at the point of transition, and the Bamboo code looked more like we just hadn't gotten around to adding more to the test yet, i.e. it's just what people were focusing on at the time.

> I wonder if in cases where RELEASE is enabled but debug not it's because of the asserts not being present:

That is entirely possible. I could make a branch of the CI actions where everything is enabled and we have a look at what works and what fails.

kent-mcleod:
> HIFIVE_debug_SMP_MCS_gcc_64 fails with:

2us does seem pretty high here.  I just checked and the CPU should be operating at 1GHz based on the register settings[1] before the kernel is started.  The kernel reads the time by performing an SBI call which invokes an SBI handler. So maybe there's more than 1000 cycles between 2 separate reads but this is unlikely. It's probably worth following up on some more. 


[1]: 
```
RISC-V # md 0x10000000
10000000: c0000000 82110ec0 00000000 82110dc0    ................
10000010: 80000000 00000000 00000000 82128ec0    ................
10000020: 80000000 00000000 0000002f 00000004    ......../.......
```
"For example, to setup COREPLL for 1 GHz operation, program divr = 0 (x1), divf
= 59 (4000 MHz VCO), divq = 2 (/4 Output divider)."

Indanz:
> The kernel reads the time by performing an SBI call which invokes an SBI handler.

Is the `rdtime` instruction handled in software??? That would explain the 2 us, but seems unlikely to me.

I verified that the system clock is 1 MHz according to the datasheet.

kent-mcleod:
> Is the `rdtime` instruction handled in software???

Actually I'm remembering an old version.  It is just an instruction and not an SBI call (https://github.com/seL4/seL4/blob/773ec49d2b6e87f880c2ad6e2df297e3d5f11807/include/arch/riscv/arch/64/mode/machine.h#L16).

Indanz:
> rdtime can be silently emulated by SBI. And this seems to be the case on the current SiFive board (Unleashed/540, Unmatched/740) , the behavior has been copied by others (e.g. Ariane) then also.

This is incredibly stupid for a new platform. Anyway, riscv's `getTimerPrecision` should return 2 then: 1 us for the clock granularity plus 1 us for the SBI overhead.

kent-mcleod:
> This is incredibly stupid for a new platform. Anyway, riscv's `getTimerPrecision` should return 2 then: 1 us for the clock granularity plus 1 us for the SBI overhead.

Is this an intermittent failure? Only failing tests for the recent run is Preprocess which I believe are fine. This ready to merge if we're happy to independently enable other SMP + MCS configurations?

Indanz:
> Is this an intermittent failure?

No, I fixed it by setting the threshold to fixed 1 us + `getTimerPrecision`. The check is relaxed enough that all tests should pass now.

It's ready to merge, except if we want to test more platforms with SMP + MCS + DEBUG.

axel-h:
> This is incredibly stupid for a new platform.

I assume they just decided to implement the cycle counter in hardware for the first silicon(s), but no dedicated timer to reduce complexity. There are some discussion about a timer/RTC peripheral that could come in here eventually, for now partly the `CLINT.time` and `CLINT.mtimecmp[hart]` does this. So, on platforms known to emulate `rdtime` (register `CSR_TIME` actually, `rdtime` is just a psedudo-op that reads it), seL4 could use the cycle counter or the time source directly. That would bypass the SBI, might work unless the SBI is doing further timer "virtualization" here or uses the timer for it's own purposes.

Indanz:
> seL4 could use the cycle counter or the time source directly.

seL4 can't use the cycle count on riscv, as it misses the guarantees that rdtime/CSR_TIME give, which is that it's synchronised across cores and won't overflow. _"The rate at which the cycle counter advances will depend on the implementation and operating environment."_ so it's pretty useless for timekeeping.

axel-h:
> seL4 can't use the cycle count on riscv, as it misses the guarantees that rdtime/CSR_TIME give, which is that it's synchronised across cores and won't overflow.

That is true, but we could still use `CLINT.mtime` directly, as this is what SBI does anyway. It would be a RISC-V SoC specific solution after all might give better realtime behavior at the price of sacrificing generic API usage. It's an option to evaluate if somebody really cares.

Indanz:
> but we could still use CLINT.mtime directly, as this is what SBI does anyway. It would be a RISC-V SoC specific solution after all might give better realtime behavior at the price of sacrificing generic API usage. It's an option to evaluate if somebody really cares.

That seems like the best solution, if that's possible and solves all problems. Anyone who cares about performance on riscv should care about this.

axel-h:
What is the state of this PR now, can it be merged. Seem it bringe more benefit than harm, if specific platform have issues these could be solved individually then. Or is there a desire to hide this behind a config option in case there are platforms/architectures where the cores time is not synchronized?

kent-mcleod:
> What is the state of this PR now, can it be merged. Seem it bringe more benefit than harm, if specific platform have issues these could be solved individually then. Or is there a desire to hide this behind a config option in case there are platforms/architectures where the cores time is not synchronized?

I think that @lsf37 had added a test PR with more SMP + MCS configurations enabled to see if they could be permanently enabled to give this PR more platforms to run tests on.

Indanz:
Specifically, I added a debug test that checks if clocks are indeed synchronous enough at bootup, but it only runs for MCS+SMP+DEBUG and not many platforms have that combination enabled.

lsf37:
The increased test set has passed, so this is ready to merge from my side. Any last objections before I push the button?

kent-mcleod (approved):

PR #866, opened 2022-06-02
tools/bitfield_gen: ensure UTF-8 encoding

lsf37:
There are cases that reasonably require non-ASCII characters in the source code, hence tools/bitfield_gen.py must be able to handle them correctly. This commit makes sure the bitfield generator consistently opens all files with UTF-8 encoding.

This is a slightly tweaked version of #780 that applies UTF-8 encoding to all files the bitfield generator handles.

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -2807,7 +2807,7 @@ class OutputFile(object):
             global temp_output_files
             temp_output_files.append(self)
         else:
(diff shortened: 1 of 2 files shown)


lsf37:
@lbmeng are you still happy with my slightly adjusted commit?

lbmeng:
> @lbmeng are you still happy with my slightly adjusted commit?

Yes, thanks for taking care of this!

Issue #856, opened 2022-05-23
Use ARMv8.1 swap instruction for acquiruing the big kernel lock on SMP


axel-h:
`clh_lock_acquire()` eventually calls `try_arch_atomic_exchange_rlx()`. On ARM this does not use `__atomic_exchange_n()` but a manual exclusive load/store sequence. After some digging, I finally found the explanation here: https://sel4.discourse.group/t/why-does-the-clh-lock-not-use-gcc-for-arm-atomics/103

- Add some comment in the code that explains this implementation choice, so one does not have to dig this up.
- ARM v8.1 added a atomic swap instruction again, somewhere around GCC 9.4 support for this was added, too. We should consider giving `__atomic_exchange_n()` another chance now. Keep the manual solution just for ARMv7 and v8.0 systems then. There is a list on https://en.wikipedia.org/wiki/Comparison_of_ARMv8-A_processors that claims A55, A65, A75+ are v8.2 already, the older ones are v8.0

Indanz:
If compiling with "-march=armv8.1-a" or "-march=armv8-a+lse", `__atomic_exchange_n()` will [automatically](https://godbolt.org/z/9e4q6b4zx) use `swpal`, so I think platforms with v8.1 should add "lse" to `KernelArmMachFeatureModifiers` instead.

yyshen:
https://mysqlonarm.github.io/ARM-LSE-and-MySQL/

Indanz:
seL4 probably wants to pass the `-mno-outline-atomics` GCC option and set the cpu flags correctly.

axel-h:
It's done for GCC >=10 and See https://github.com/seL4/seL4/pull/909 also does it for Clang >=12

PR #867, opened 2022-06-07
smp: Address MCS timekeeping issues in ipiStallCoreCallback

kent-mcleod:
Update (5.Mar.2025): With https://github.com/seL4/seL4/pull/867 being recently merged, this PR needs to be rebased. I think the top 5 commits are still worth reapplying so I'll keep the PR open. 
Because it's been so long since doing this work, I can't find the issue it was addressing. But essentially when a TCB with an SC on one core performs an operation that requires remotely stalling another core, the time consumed by this stall isn't properly accounted. But there's also some correctness bugs where some operations are performed outside of the kernel lock.

### DIFF
--- a/src/smp/ipi.c
+++ b/src/smp/ipi.c
@@ -29,7 +29,14 @@ void ipiStallCoreCallback(bool_t irqPath)
     word_t cpu = getCurrentCPUIndex();
     clh_node_t *node = &big_kernel_lock.node[cpu];
 
-    if (clh_is_self_in_queue() && !irqPath) {
+    // There's 3 ways that this function can be called:
+    // - On the IRQ path where the lock isn't acquired
+    // - On the IRQ path where the lock is attempted to be acquired
+    // - On the non-IRQ path where the lock is attempted to be acquired
+    // !irqPath implies clh_is_self_in_queue()
+    assert(clh_is_self_in_queue() || irqPath);
+
+    if (!irqPath) {
         /* The current thread is running as we would replace this thread with an idle thread
          *
          * The instruction should be re-executed if we are in kernel to handle syscalls.
--- a/src/smp/ipi.c
+++ b/src/smp/ipi.c
@@ -36,62 +36,53 @@ void ipiStallCoreCallback(bool_t irqPath)
     // !irqPath implies clh_is_self_in_queue()
     assert(clh_is_self_in_queue() || irqPath);
 
-    if (!irqPath) {
-        /* The current thread is running as we would replace this thread with an idle thread
-         *
-         * The instruction should be re-executed if we are in kernel to handle syscalls.
-         * Also, thread in 'ThreadState_RunningVM' should remain in same state.
-         * Note that, 'ThreadState_Restart' does not always result in regenerating exception
-         * if we are in kernel to handle them, e.g. hardware single step exception. */
-        if (thread_state_ptr_get_tsType(&NODE_STATE(ksCurThread)->tcbState) == ThreadState_Running) {
-            setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
-        }
-
-        SCHED_ENQUEUE_CURRENT_TCB;
-        switchToIdleThread();
-#ifdef CONFIG_KERNEL_MCS
-        commitTime();
-        NODE_STATE(ksCurSC) = NODE_STATE(ksIdleThread)->tcbSchedContext;
-#endif
-        NODE_STATE(ksSchedulerAction) = SchedulerAction_ResumeCurrentThread;
-
-        /* Let the cpu requesting this IPI continue while we wait on the lock */
-        node->ipi = 0;
-#ifdef CONFIG_ARCH_RISCV
-        ipi_clear_irq(irq_remote_call_ipi);
-#endif
-        ipi_wait();
-
-        /* Continue waiting on lock */
-        while (node->watch->state != CLHState_Granted) {
-            __atomic_thread_fence(__ATOMIC_ACQUIRE);
-            if (clh_is_ipi_pending(cpu)) {
-                /* Multiple calls for similar reason could result in stack overflow */
-                assert(big_kernel_lock.ipi.remoteCall != IpiRemoteCall_Stall);
-                handleIPI(CORE_IRQ_TO_IRQT(cpu, irq_remote_call_ipi), irqPath);
-            }
-            arch_pause();
-        }
-
-        /* make sure no resource access passes from this point */
-        asm volatile("" ::: "memory");
-
-        /* Start idle thread to capture the pending IPI */
-        activateThread();
-        restore_user_context();
-    } else {
-        /* We get here either without grabbing the lock from normal interrupt path or from
-         * inside the lock while waiting to grab the lock for handling pending interrupt.
-         * In latter case, we return to the 'clh_lock_acquire' to grab the lock and
-         * handle the pending interrupt. Its valid as interrups are async events! */
-        SCHED_ENQUEUE_CURRENT_TCB;
-        switchToIdleThread();
-#ifdef CONFIG_KERNEL_MCS
-        commitTime();
-        NODE_STATE(ksCurSC) = NODE_STATE(ksIdleThread)->tcbSchedContext;
-#endif
-        NODE_STATE(ksSchedulerAction) = SchedulerAction_ResumeCurrentThread;
+    /* The current thread is running as we would replace this thread with an idle thread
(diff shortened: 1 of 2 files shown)


include/smp/lock.h
```diff
@@ -71,14 +71,6 @@ static inline void *sel4_atomic_exchange(void *ptr, bool_t
     while (!try_arch_atomic_exchange_rlx(&big_kernel_lock.head,
                                          (void *) big_kernel_lock.node_owners[cpu].node,
                                          (void **) &prev)) {
-        if (clh_is_ipi_pending(cpu)) {
-            /* we only handle irq_remote_call_ipi here as other type of IPIs
-             * are async and could be delayed. 'handleIPI' may not return
-             * based on value of the 'irqPath'. */
-            handleIPI(CORE_IRQ_TO_IRQT(cpu, irq_remote_call_ipi), irqPath);
```
kent-mcleod, line 78: It's actually incorrect to call `handleIPI` here the value of `big_kernel_lock.node_owners[cpu].next` hasn't been correctly initialized until `sel4_atomic_exchange` returns and so `ipiStallCoreCallback` called by `handleIPI` will poll the incorrect lock address.

lsf37: I agree that it is incorrect, but do we have any data on why we were trying to do `handleIPI` in here? Were we trying to assure progress of something that otherwise wouldn't happen?

axel-h: The commit history says something that it can take up to seconds until the atomic exchange succeeds. It does not say which platform, maybe one that was quite bad at SMP support. Modern platform should be better, ARMv8.1 even brings the swap instruction back.

kent-mcleod: @axel-h do you have a link for that message? The commit that introduced the handleIPI call doesn't have a body: https://github.com/seL4/seL4/commit/3d2ae69f9cb184c34b6b56d1365e57693e832db4#

axel-h: I was wondering about this code and eventually I found this discussion that has some background: https://sel4.discourse.group/t/why-does-the-clh-lock-not-use-gcc-for-arm-atomics/103, then I created https://github.com/seL4/seL4/issues/856

Indanz: This those long delays may be fixed by commits [Repair barriers in sel4_atomic_exchange](https://github.com/seL4/seL4/commit/8ba22dcdc5fa6) and [Repair barriers in clh_lock_acquire](https://github.com/seL4/seL4/commit/fe1278326f664b).

include/smp/lock.h
```diff
@@ -71,14 +71,6 @@ static inline void *sel4_atomic_exchange(void *ptr, bool_t
...
-        }
-
-        arch_pause();
```
Indanz, line 81: Why also remove `arch_pause()`? That call has nothing to do with handling IPIs.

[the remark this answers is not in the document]
Indanz: Only ARM's `try_arch_atomic_exchange_rlx` may return false, and ARM's `arch_pause()` is empty, so it's fine to remove. Worth mentioning though.

I propose to rename `try_arch_atomic_exchange_rlx` to `arch_atomic_exchange_rlx` and to move the while loop into ARM's implementation, `clh_lock_acquire` is the only caller.

kent-mcleod: I was essentially trying to revert https://github.com/seL4/seL4/commit/3d2ae69f9cb184c34b6b56d1365e57693e832db4 which also added arch_pause().  Before the commit the x86 variant directly called __atomic_exchange_n and continued to call __atomic_exchange_n in it's implementation of try_arch_atomic_exchange.  Meanwhile, arch_pause is an empty static inline on arm. 

This part of the lock is supposed to have minimal bus contention as it's just coordinating the insertion into the FIFO queue. If the pause is needed for x86 then it should either be handled by the compilers implementation of __atomic_exchange_n, or introducing it should have been in a separate commit explaining why just having __atomic_exchange_n was insufficient.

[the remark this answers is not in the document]
kent-mcleod: See https://github.com/seL4/seL4/pull/867#discussion_r891212765

axel-h: Seem you are aiming at soething like what I was experimenting with in https://github.com/axel-h/seL4/pull/73/

kent-mcleod: > Seem you are aiming at something like what I was experimenting with in [axel-h#73](https://github.com/axel-h/seL4/pull/73)

I'm trying to resolve the issue with ipiStallCoreCallback on MCS.

src/smp/ipi.c
```diff
@@ -21,62 +21,69 @@
...
-        ipi_clear_irq(irq_remote_call_ipi);
+    ipi_clear_irq(irq_remote_call_ipi);
 #endif
-        ipi_wait(totalCoreBarrier);
+    ipi_wait(totalCoreBarrier);
 
-        /* Continue waiting on lock */
-        while (big_kernel_lock.node_owners[getCurrentCPUIndex()].next->value != CLHState_Granted) {
-            if (clh_is_ipi_pending(getCurrentCPUIndex())) {
+    /* Continue waiting on lock */
```
kent-mcleod, line 69: I think I accidentally reverted a comment I had here:
- Because there's no way to remove a node from the lock queue until it has acquired the lock and can then release it, in the syscall context, we need to wait for the lock here and then immediately return.  We can't return to higher up the stack because the syscall context is no longer valid since the thread was changed.

src/smp/ipi.c
```diff
@@ -42,14 +42,18 @@ void ipiStallCoreCallback(bool_t irqPath)
     if (!irqPath && thread_state_ptr_get_tsType(&NODE_STATE(ksCurThread)->tcbState) == ThreadState_Running) {
         setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
     }
-
-    SCHED_ENQUEUE_CURRENT_TCB;
-    switchToIdleThread();
 #ifdef CONFIG_KERNEL_MCS
-    commitTime();
-    NODE_STATE(ksCurSC) = NODE_STATE(ksIdleThread)->tcbSchedContext;
```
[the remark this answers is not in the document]
Indanz: Never mind, you added a `switchSchedContext` call, which updates `ksCurSC`.

src/smp/ipi.c
```diff
@@ -21,62 +21,72 @@
...
+#ifdef CONFIG_KERNEL_MCS
+    updateTimestamp();
+#endif
+    if (config_ternary(CONFIG_KERNEL_MCS, checkBudget(),1)) {
         SCHED_ENQUEUE_CURRENT_TCB;
-        switchToIdleThread();
+    }
+    switchToIdleThread();
+    NODE_STATE(ksSchedulerAction) = SchedulerAction_ResumeCurrentThread;
+    doMaskReschedule(ARCH_NODE_STATE(ipiReschedulePending));
```
kent-mcleod, line 53: Triggering an IPI within an IPI handler is getting pretty sketchy... I believe this is currently needed because a timeoutFault generated by the current thread being at the end of its budget can make a thread runnable on a different node. As doMaskReschedule is a non-blocking IPI it only delivers an interrupt to the other nodes which only act on it when they independently get the lock.  (Compared with blocking IPIs which are executed by the target nodes without acquiring the lock). I expect that this means that if the reschedule IPI is sent back to the original node that triggered this synchronous IPI it will be delivered once that node releases the lock.

Indanz: Why not let the future schedule call do the `doMaskReschedule`? Sending reschedule IPIs is non-blocking, but receiving isn't. As the reschedule IPI is handled with the kernel lock taken, doing it here will force a third remote core to block on the kernel lock immediately.

Indanz:
I am wondering if it's not easier to go back to the old code which didn't switch to the idle thread's SC and to do the switch only when the SC is being deleted. That, or do the actual deletion on the remote core at the end of IPICoreStallback. Then all timekeeping issues are bypassed: Either the interrupted SC is charged as it used to be, or the SC is gone so time accounting doesn't matter any more.

What started as a simple bugfix opened a can of worms it seems.

kent-mcleod:
After talking a bit to @Indanz , we're starting to consider that it might be better to remove the stallTCB ipi and replace with case-specific operations to be performed on the remote core instead of stalling. This way the time accounting for the stall time problem goes away.

kent-mcleod:
Closing this change for now. I extracted the last commit that fixes a non-mcs SMP bug: https://github.com/seL4/seL4/pull/1421

PR #869, opened 2022-06-09
fix arm64 address space graphic in comment

FinnZhang:
Reason: There is a error in the comment about arm64 address space layout. 
Do: Fix the comment error.


axel-h:
Thanks. 

One minor thing, please change the e-mail address in the sign-off message from `14357454+FinnZhang@users.noreply.github.com` to a valid working one.

FinnZhang:
@axel-h Done. Please review. Thanks.

lsf37 (approved):
Thank for fixing that.

I think there might be another error in the picture for hypervisor mode. The picture puts PPTR_BASE at `2^48 - 2^39` (`0xFF80_0000_0000`), but the code puts it at `2^39` (`0x80_0000_0000`). @kent-mcleod I assume the code is correct here and the picture should be updated?

Issue #635, opened 2021-11-02
static data defined in smp/ipi.h


Indanz:
include/smp/ipi.h not only declares data, but also defines all IPI related global variables, while the file is included in multiple places. It works because all kernel code is combined into kernel_all_pp_prune.c, but it freaks me out. Is this intentional?

axel-h:
I agree, this should be avoided and it violates a lot of common C best practice rules also. Since there is a `src/smp/ipi.c`, putting them there should easy and we can keep an extern declaration in the header. Can you make a patch an then we can see if the proof run accepts this.
Putting them in `src/model/statedata.c` might also be an option, but it seem this file is not really intended for all global variables per se. Seems it just collected a few things over time that did not have a better place because there was no *.c file.

lsf37:
`statedata.c` is intended for the global data of the kernel, it is definitely not a random accumulation place, and should be the first place to look at for the main global state.

The exception would implementation details of some mechanims. For IPI this is a bit borderline, `ipi.c` might make sense. @kent-mcleod do I have this right?

kent-mcleod:
`ipi.c` is probably best in this case.  statedata.c is usually for state that is also present in the executable spec.  Currently these variables are for implementing the IPI mechanisms in the C kernel and so placing them in `ipi.c` makes sense. Similar to how `gic_v2.c` contains globals related to the GIC C implementation.

PR #871, opened 2022-06-15
SMP: Improve kernel lock

Indanz:
The main thing this achieves is reverting commit [3d2ae69f](https://github.com/seL4/seL4/commit/3d2ae69f9cb184c34b6b56d1365e57693e832db4) and fixing the underlaying delay problem properly.

CAS can take very long to finish on ARM if the exclusive reservation granule (ERG) is large for the platform, as any writes on other cores within up to 2Kb could make the CAS fail. The correct fix is to add padding around the global lock equal to ERG to make CAS fast, making the fudgy extra IPI check redundant.

This incorporates @kent-mcleod and @axel-h recent improvements to the kernel locking code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
 * riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
   Except for RISCV32 with LLVM clang enabled will default KernelRiscvExtD to OFF.
+* Multiple SMP related bugs fixed.
 
 ### Platforms
 


include/arch/arm/arch/machine/hardware.h
```diff
@@ -30,3 +30,11 @@ typedef word_t vm_fault_type_t;
 #define L1_CACHE_LINE_SIZE_BITS CONFIG_L1_CACHE_LINE_SIZE_BITS
 #define L1_CACHE_LINE_SIZE BIT(L1_CACHE_LINE_SIZE_BITS)
 
+/* Used to align the big kernel lock to the exclusive reservation granule size.
+ * Without this nearby writes can delay atomic operations implemented with looping
+ * exclusive load/store instructions for an undefined time.
+ *
+ * Usually equal to L1_CACHE_LINE_SIZE, but 2k is the maximum for SMP systems.
```
axel-h, line 37: Can you refer to any source here, is this ARMv8 or ARMv7 also?

axel-h:
See also:
- https://github.com/seL4/seL4/pull/867
- https://github.com/seL4/seL4/issues/856

include/arch/arm/arch/machine/hardware.h
Indanz: Good idea, done.

include/smp/lock.h
```diff
@@ -22,86 +22,48 @@
...
+    return big_kernel_lock.node[cpu].ipi == 1;
 }
 
-static inline void *sel4_atomic_exchange(void *ptr, bool_t
-                                         irqPath, word_t cpu, int memorder)
+static inline void FORCE_INLINE clh_lock_acquire(bool_t irqPath)
 {
-    clh_qnode_t *prev;
+    word_t cpu = getCurrentCPUIndex();
+    clh_node_t *node = &big_kernel_lock.node[cpu];
```
axel-h, line 58: This is generic code, so let's add a save guard that catches development errors with the CPU index mapping.
```
assert(cpu < ARRAY_SIZE(big_kernel_lock.node));
```

include/smp/lock.h
```diff
@@ -120,27 +82,30 @@ static inline void FORCE_INLINE clh_lock_acquire(word_t cpu, bool_t irqPath)
...
+    /* Pass lock to successor */
+    node->myreq->state = CLHState_Granted;
+    /* Take ownership of watched request, to use next time we take the lock */
+    node->myreq = node->watch;
 }
 
 static inline bool_t FORCE_INLINE clh_is_self_in_queue(void)
 {
-    return big_kernel_lock.node_owners[getCurrentCPUIndex()].node->value == CLHState_Pending;
+    return big_kernel_lock.node[getCurrentCPUIndex()].myreq->state == CLHState_Pending;
```
axel-h, line 100: Same save guard here:
```
word_t cpu = getCurrentCPUIndex();
assert(cpu < ARRAY_SIZE(big_kernel_lock.node));
return big_kernel_lock.node[cpu].myreq->state == CLHState_Pending;
```

include/smp/lock.h
```diff
@@ -22,86 +22,48 @@
...
-typedef struct clh_qnode_p {
-    clh_qnode_t *node;
-    clh_qnode_t *next;
-    /* This is the software IPI flag */
+} clh_req_state_t;
+
+/* Lock request */
+typedef struct clh_req {
+    clh_req_state_t state;
+} ALIGN(L1_CACHE_LINE_SIZE) clh_req_t;
```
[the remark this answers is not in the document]
Indanz: It's not necessary, but it matches what the paper does. Having it as a structure makes it clearer that it's an object that's passed around, so I think I prefer it as it is. It's a very clever algorithm and making it as easy to understand as possible is the goal here.

To me it would be more logical to have the update of `node->myreq` together with the `node->myreq->state = CLHState_Pending`, but the paper also does it this way.

include/smp/lock.h
Indanz: No one else checks the return value of `getCurrentCPUIndex()`. If that's broken you will never get to an assertion here.

If you want to check it, check `getCurrentCPUIndex() < CONFIG_MAX_NUM_NODES` during boot.

include/smp/lock.h
```diff
@@ -22,86 +22,48 @@
...
+    clh_req_t *watch; // Used by predecessor to grant the lock to us.
+    clh_req_t *myreq; // Used to grant the lock to our successor.
+    /* This is the software blocking IPI flag */
     word_t ipi;
-
-    PAD_TO_NEXT_CACHE_LN(sizeof(clh_qnode_t *) +
-                         sizeof(clh_qnode_t *) +
-                         sizeof(word_t));
-} clh_qnode_p_t;
+} ALIGN(L1_CACHE_LINE_SIZE) clh_node_t;
```
axel-h, line 38: I wonder, does this struct really have an alignment requirement? The member have one that should impose a certain placement of the struct.

Indanz: Not really, it avoids some cache line bouncing. All data in `clh_node_t` is per-node, the alignment is to avoid cache line sharing between nodes.

So it's not required for correctness, only for performance, and the alignment applies to the whole of `clh_node_t` and not individual members.

src/smp/lock.c
```diff
@@ -4,15 +4,19 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+#include <assert.h>
```
axel-h, line 7: please leave `config.h` has the first include and include `assert.h` afterwards

Indanz: Done.

Indanz:
IPI had the same undefined delay because of unrelated writes within the exclusive reservation granule problem as the global lock. I fixed it by moving all IPI state into ` big_kernel_lock`. This also resolves #635.

Indanz:
A next step could be to replace all the compiler built-in atomics with C11 standard atomics. Then we could add models of the algorithms used and check them with [CppMem](http://svr-pes20-cppmem.cl.cam.ac.uk/cppmem/).

Indanz:
With these lock/IPI fixes, #867 can proceed, and when that's done, then #847 (minus the MCS, SMP: Fix ipiStallCoreCallback commit) can finally be merged.

include/smp/lock.h
kent-mcleod: Interesting. I didn't know that the aligned attribute padded the internal size of structs.  @lsf37, how will this play with the CParser?

Indanz: Makes sense.

I used ALIGNED because I really don't like `PAD_TO_NEXT_CACHE_LN`, which is also very error prone. I can undo that part of the change if necessary of course.

src/arch/riscv/smp/ipi.c
```diff
@@ -82,7 +65,7 @@ void ipi_send_target(irq_t irq, word_t hart_id)
     assert(core_id < CONFIG_MAX_NUM_NODES);
 
     assert((ipiIrq[core_id] == irqInvalid) || (ipiIrq[core_id] == irq_reschedule_ipi) ||
-           (ipiIrq[core_id] == irq_remote_call_ipi && big_kernel_lock.node_owners[core_id].ipi == 0));
+           (ipiIrq[core_id] == irq_remote_call_ipi && clh_is_ipi_pending(core_id)));
```
kent-mcleod, line 68: This assert appears to have swapped from !pending to pending.

src/smp/ipi.c
```diff
@@ -41,18 +41,17 @@ void ipiStallCoreCallback(bool_t irqPath)
...
+        big_kernel_lock.node[getCurrentCPUIndex()].ipi = 0;
 #ifdef CONFIG_ARCH_RISCV
         ipi_clear_irq(irq_remote_call_ipi);
 #endif
-        ipi_wait(totalCoreBarrier);
+        ipi_wait();
 
         /* Continue waiting on lock */
-        while (big_kernel_lock.node_owners[getCurrentCPUIndex()].next->value != CLHState_Granted) {
+        while (clh_is_self_in_queue()) {
```
kent-mcleod, line 51: This isn't the same check. Previously the loop is checking if the next lock holder has granted (released).  Now it's just checking whether the current node holder is waiting. clh_is_self_in_queue() returns true if the current node is in the queue for the lock and also returns true if it is the current lock holder.  It only returns false when the current node has released the lock and hasn't entered the queue yet.

kent-mcleod (commented):
This seems to be a solid improvement. Currently some of the SMP tests are failing but I think this is just because of a couple conditional check errors that I commented on.

include/smp/lock.h
lsf37: > Interesting. I didn't know that the aligned attribute padded the internal size of structs. @lsf37, how will this play with the CParser?

I don't think it will be able to deal with that. It might accept it, but if we actually want to verify anything about them, we will need structs to be packed (i.e. no compiler-generated padding). Since this is not a verified config, it's probably fine to use this for now and deal with making padding explicit at some later point.

src/smp/ipi.c
Indanz: Indeed it isn't, I was confused by the renames and by the comment, `clh_is_self_in_queue` is more subtle than I expected. Fixed.

include/smp/lock.h
Indanz: Mind that the only critical alignment and padding in the code is the one for EXCL_RES_GRANULE_SIZE, the others are only for performance. Manual padding structures does not guarantee alignment for over-aligned data either, that can only be verified on the binary, after linking.

src/smp/ipi.c
```diff
@@ -79,10 +78,45 @@ void ipiStallCoreCallback(bool_t irqPath)
...
+{
+    word_t cores = big_kernel_lock.totalCoreBarrier;
+    word_t *ipiSyncBarrier_count = &big_kernel_lock.ipiSyncBarrier_count;
+
+    if (__atomic_fetch_add(ipiSyncBarrier_count, 1, __ATOMIC_ACQ_REL) == cores) {
+        *ipiSyncBarrier_count = 0;
+        __atomic_thread_fence(__ATOMIC_RELEASE);
+    }
+
+    while (*ipiSyncBarrier_count != 0) {
```
yyshen, line 91: If the core1 resets *ipiSyncBarrier_count to 0, but before  core1 observes the change, another core3 starts another round of ipi_wait, the *ipiSyncBarrrier_count becomes greater than 0. Thus, core1 might stay in the while of the last round ipi_wait?

Indanz: > another core3 starts another round of ipi_wait

That's not possible I think. Only the core holding the kernel lock can initiate IPIs, and only cores receiving that IPI will call ipi_wait. IPIs are effectively serialised by the big kernel lock.

Anyway, if this would happen, it wouldn't work with the old code either, which did:
```
ipiSyncBarrier.globalsense = ~ipiSyncBarrier.globalsense;
```
and doing that twice also doesn't give a difference.

yyshen: the old code works since it does not wait on the counter.

Indanz: I'm not sure why the remote cores need to wait for all other cores to have handled the IPI instead of just returning to user space. Those `ip_wait` calls could probable be replaced with an `ipi_ack` instead. Perhaps it makes sense to split `ipi_wait` into `ipi_ack` and `ipi_wait`. To me it seems that the IPI sender only needs to wait for the other cores to ack the IPI. Now ipi_wait also works as a lock in case multiple CPUs want to do an IPI at the same time, but that seems unnecessary, as that's not possible anyway because of overwritten global function and args data.

Indanz: I don't see how it could have worked:
```
core0		core1		core2		core3
---------------------------------------------------------
sense=A
IPI for 1 & 2
count=1
		count=2
				count=3
				sense=~A
				count=0
						IPI for 1
						count=1
		count=2
		sense=A
		count=0
still sees A
```
The race window (if there is one) is smaller, but it's not gone.

yyshen: Sorry, I did not follow the example you gave. Maybe a SPIN model could confirm the new barrier code is correct.

Indanz: If sending blocking IPIs is not serialised by the big kernel lock, multiple IPI senders will re-use the same global data, and it will be broken, agreed? So in my example above core3 would not be able to send an IPI until core0 is done, so disregard that.

After my change there is a theoretical race window where:
```
core0		core1		core2
---------------------------------------------------------
take bkl
IPI for c1
		count=1
count=2
count=0
release bkl
				take bkl
				IPI for 1
				count=1
		still sees
		count != 0
		deadlock
```
But before my change that would have been:
```
sense=A

core0		core1		core2
---------------------------------------------------------
take bkl
IPI for c1
		count=1
count=2
sense=~A
count=0
release bkl
				take bkl
				IPI for 1
				count=1
		sees A != ~A
		no deadlock
```
However, if core 2 would send an IPI to core 3 and 3 would invert globalsense before core1 checked it, there is still a race, although no deadlock.

So I agree the original code was more robust against races. I'll have a look at how to fix this, maybe for now just add back the sense thing.

kent-mcleod: > However, if core 2 would send an IPI to core 3 and 3 would invert globalsense before core1 checked it, there is still a race, although no deadlock.

Instead of a toggle, it could be an incrementing counter

kent-mcleod: It'd probably be better to take out the ipi_wait in a separate PR so that the behavior change from this PR is smaller.

Indanz: > It'd probably be better to take out the ipi_wait in a separate PR so that the behavior change from this PR is smaller.

Agreed. Especially if that causes the IPC1001 failure (running the test in a loop here, but I can't reproduce it).

Indanz:
@kent-mcleod: Should I add your singed-off-by to commit e7322128697208c7 "smp: Don't check for IPIs in sel4_atomic_exchange", or should I squash it with the revert of commit 3d2ae69f9cb184c34b6b5, commit bb7b956f13 "Remove try_arch_atomic_exchange_rlx"?

kent-mcleod:
> @kent-mcleod: Should I add your singed-off-by to commit [e732212](https://github.com/seL4/seL4/commit/e7322128697208c76fc478c0334fd5305cd2727c) "smp: Don't check for IPIs in sel4_atomic_exchange", or should I squash it with the revert of commit [3d2ae69](https://github.com/seL4/seL4/commit/3d2ae69f9cb184c34b6b56d1365e57693e832db4), commit [bb7b956](https://github.com/seL4/seL4/commit/bb7b956f13f442af223bb30440c75a9b2d7f2d72) "Remove try_arch_atomic_exchange_rlx"?

Squashing is fine with me.

Indanz:
I thought the benchmarking code took the kernel entry timestamp after taking the kernel lock, so I'm curious how you measured a change there?

kent-mcleod:
Most of the sel4bench benchmarks measure from user level and don't use the kernel log buffer.

Indanz:
Thinking about it a bit more, 5-10% performance difference seems crazy.  Are you sure it's not just noise?

kent-mcleod:
> Thinking about it a bit more, 5-10% performance difference seems crazy. Are you sure it's not just noise?

I still haven't had a chance to sit down and run these again sorry. I could run sel4bench on specific hardware platforms and upload the logs if you wanted to go through them @Indanz ?

Indanz:
> I still haven't had a chance to sit down and run these again sorry.

Same here.

> I could run sel4bench on specific hardware platforms and upload the logs if you wanted to go through them

I can run sel4bench myself, that would save you time. But I only have access to one platform which isn't out-of-order, so it's not the best one to test locking code with.

If the gain is caused by less inline code, making `clh_lock_acquire` out-of-line may improve performance further, but that might very well depend on the compiler and platform.

I reviewed the code again and all seems fine to me. The main concern with unexpected performance gains in locking code is that it's caused by a bug. But I only strengthened the barriers, I didn't relax anything.

Indanz:
The machine got restarted halfway the ODROID_XU4_debug_hyp_clang_32 test run, confusing the result parser. The tests itself passed. The TX2 machine maybe turned off currently.

kent-mcleod:
@lsf37, would it be possible to add an option to run sel4bench actions on seL4 kernel PRs via a label?

kent-mcleod:
> @lsf37, would it be possible to add an option to run sel4bench actions on seL4 kernel PRs via a label?

Added https://github.com/seL4/sel4bench/pull/30 for now to allow for smp benchmarks to be more easily compared against mainline

lsf37:
> @lsf37, would it be possible to add an option to run sel4bench actions on seL4 kernel PRs via a label?

Yes that could be done. I'm on leave this week, but I can have a look next week.

Indanz:
After the update, SABRE_release_SMP_clang_32 hangs in MULTICORE0004, which does cross-core task cleanup.

kent-mcleod:
(Just a note that there was more discussion about this PR here: https://github.com/seL4/sel4bench/pull/30)

Indanz:
Pull #936 also fails MULTICORE0004 test for `ODROID_C4_release_SMP_MCS_clang_64`, so perhaps something else is going on and it's not related to my changes. Need to spend some time with hardware and running that test a lot to see what's up with that.

Ivan-Velickovic:
> Pull #936 also fails MULTICORE0004 test for `ODROID_C4_release_SMP_MCS_clang_64`, so perhaps something else is going on and it's not related to my changes. Need to spend some time with hardware and running that test a lot to see what's up with that.

Not sure if it's useful to you, but when initially adding the OdroidC4 to the CI, I saw failures for SMP+MCS+HYP, more details here: https://github.com/seL4/seL4/issues/928. I suspect something else is going on, independent of both this PR and #936.

Indanz:
MULTICORE_0004 sometimes failing for clang is a known open issue in [sel4test](https://github.com/seL4/sel4test/issues/44), so not caused by my changes. I'd like to get to the bottom of that and commit the fix in this pull request though.

Indanz:
@yyshen You mentioned at the summit that you had two bugfixes for the kernel locking code, one for IPis and one for the lock itself?

Indanz:
Interesting to see that the SABRE_verification_SMP_clang_32 test fails exactly the same way as @yyshen's version in his PR:
```
  
  DIDRv: 3, armv 70, coproc baseline only? no.
  CPU is in secure mode. Enabling debugging in secure user mode.
  Bootstrapping kernel
    [10000000..2ff00000]
  reserved virt address space regions: 4
    [e0000000..e0050000]
    [e0050000..e005a2f1]
    [e005b000..e0491000]
    [ff000000..ff100000]
  DDDIIIDDDRRRvvv:::   333,,,   aaarrrmmmvvv   777000,,,   cccoooppprrroooccc   bbbaaassseeellliiinnneee   ooonnnlllyyy???   nnnooo...
  
  
  
  
  CCCPPPUUU   iiisss   iiinnn   ssseeecccuuurrreee   mmmooodddeee...   EEEnnnaaabbbllliiinnnggg   dddeeebbbuuuggggggiiinnnggg   iiinnn   ssseeecccuuurrreee   uuussseeerrr   mmmooodddeee...
  
  
  
  
  
  
  KERNEL PREFETCH ABORT!
  Faulting instruction: 0x0
  IFSR: 0x5
  halting...
```

TX2_verification_hyp_gcc_64 failed SERSERV_CLI_PROC_004: Not SMP, so not related to any of my changes, but I've seen this fail in other CI runs too now, so perhaps another issue to look into.

Indanz:
Okay, looks like the weird SABRE problems are fixed with commit 58cedea29. I've ran the tests about ten times now, no failures so far. All looks good from my end. Does anyone want to do one last review before we finally merge this?

Indanz:
Now that  #1372 is merged and following the advice there, I can build with clang and reproduce it with clang 21. The fault happens inside the call to schedule by core 3, so way at the end of init_kernel. If I add NODE_UNLOCK_IF_HELD to kernelPrefetchAbort sel4test continues until it hangs in the BIND0001 test. I'll continue debugging tomorrow.

Indanz:
Requesting new reviews. For the people who reviewed before, everything before commit "SMP: Fix IPI Race" is unchanged.

include/arch/arm/arch/32/mode/smp/smp.h
```diff
@@ -8,29 +8,14 @@
...
-    asm("mov %[stack_address], %[currStackAddress]" : [stack_address] "=r"(stack_address) : [currStackAddress] "r"(&stack_address):);
-    return stack_address;
-}
-
 static inline CONST cpu_id_t getCurrentCPUIndex(void)
 {
-    cpu_id_t cpu_id;
-    word_t sp = getCurSP();
+    char sp[8];
+    word_t idx = (word_t)(sp - kernel_stack_alloc[0]);
```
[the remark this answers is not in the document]
Indanz: Should be valid C, you're allowed to take the pointer of data on the stack. It's as weak as the original, just rewritten in C instead of inline assembler, so less chance of subtle problems and more readable. The compiler is allowed to optimize inline assembler out if it thinks it can, and making it atomic doesn't always prevent that either.

I could change it to use TPIDRPRW (and HTPIDR with HYP) I suppose, but the change would be bigger. I assumed there was some reason to use SP, like accessing TPIDRPRW being slow or something.

llvm.cmake
```diff
@@ -42,3 +42,5 @@ if(NOT ("${CCACHE}" STREQUAL CCACHE-NOTFOUND))
     set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ${CCACHE})
 endif()
 mark_as_advanced(CCACHE)
+
+string(APPEND common_flags " -Wno-unused-command-line-argument")
```
kent-mcleod, line 46: Can this be declared with KernelCommonFlags() in the kernel's CMakeLists.txt with a clang guard instead.

Indanz: I didn't really meant this commit to be part of this PR, just had it for convenience and originally planned creating a separate PR for that. I'll drop it from here and create a new PR with your suggestion.

src/arch/arm/machine/l2c_310.c
```diff
@@ -340,6 +340,7 @@ BOOT_CODE void initL2Cache(void)
     /* Direct register access */
     l2cc->control.control |= CTRL_CTRL_EN;
 #endif /* TI_MSHIELD */
+    dsb();
```
kent-mcleod, line 343: What do you think about adding line comments on each of these mem barriers explaining their assumptions/guarantees?
Eg:
"The write to the control register needs to complete before the following write to the aux control register.
The dsb guarantees that the setACTLR() doesn't start executing until the completion of the l2cc control register update".

kent-mcleod: I think we should try and aim for comments on all mem barriers used throughout the kernel if we want to be able to keep improving their application.

kent-mcleod:
Nice find! Is it just that 1 commit that fixes the sabre tests, or are the l2cc membarriers needed also?

If we're all happy with this set of changes to be merged, would you be happy to merge them in stages (smp lock refactor, sabre stack cpuid fixes, l2cc membarriers, arm gic init)?

Indanz:
> Nice find! Is it just that 1 commit that fixes the sabre tests, or are the l2cc membarriers needed also?

I'm not entirely sure to be honest. I thought the gcc crash had the same cause as the clang one. The l2cc fixes made it work for gcc, but perhaps that was coincidental, and just that one commit fixes all problems for both gcc and clang. I did verify that just that one commit makes all the difference for clang.

[...goes off testing...]

I just tried gcc and clang with e47e8a85e and a436c5d0e reverted, both work fine now, so it was pure luck and it really was just that SP bug.

Was it only a problem on the Sabre board because that is the only Aarch32 platform with SMP enabled in CI? The other 32-bit Arm configs with SMP enabled are Aarch64 I think.

The l2cc patches are still necessary for correctness though. As those operations only run rarely, some extra barrier overhead won't be noticeable anyway, so doesn't hurt.

> If we're all happy with this set of changes to be merged, would you be happy to merge them in stages (smp lock refactor, sabre stack cpuid fixes, l2cc membarriers, arm gic init)?

I tried ordering the commits in a sensible order, is that good enough, or do you prefer separate PRs for those commits because it isn't part of improving kernel locking any more? (I could rename this PR slightly. ;-)

For me the reason to keep them all here is because then it can all be tested together and because without the SP fix the lock changes will trigger the SP bug on Aarch32 SMP. And similarly, without the streamlining of the locking code it's hard to trigger the SP bug, so it's hard to verify whether the fix is effective or not. If you split it off from this PR you lose the "tested together" part.

It's getting big though, so if I do split it, the SP fix must be merged before this PR. let me know what you prefer. I'm a bit exhausted after spending more than 50 hours on this, I just want to get it merged one way or the other at this point.

kent-mcleod:
> It's getting big though, so if I do split it, the SP fix must be merged before this PR. let me know what you prefer. I'm a bit exhausted after spending more than 50 hours on this, I just want to get it merged one way or the other at this point.

I think having a big PR for reviewing and testing everything together makes sense to me in this case.

My argument for merging separately is that the smaller each atomic change through the CI, the easier for someone to come back in the future if there's any issues.

Accordingly, if you want, I'm happy to shepherd the mechanical git + pull request + merging stuff  for splitting it apart and checking the tests pass

lsf37:
> Would be good to test currently disabled configuration options with this PR applied. With luck, these are fixed too now:
> 
> * [OdroidC4 with debug+SMP+MCS on clang fails MULTICORE0002  #1333](https://github.com/seL4/seL4/issues/1333)
> * [support MCS + SMP config on imx8mm #929](https://github.com/seL4/seL4/issues/929)
> * [TX1, TX2, ODROID_C4 fail `SCHED_CONTEXT_0014` on MCS+hyp+MC on clang #928](https://github.com/seL4/seL4/issues/928)
> * [test MULTICORE_0004 sometimes hangs sel4test#44](https://github.com/seL4/sel4test/issues/44)

That would be truly epic, thanks for all the hard work on this!

> Is there any way to kick off those ci-actions for this PR, or do I need to manually run them one by one?

These are all globally disabled now, so unfortunately no easy way. I could enable them, but then things will start failing everywhere (apart from hopefully here). Need to think about if there is a good way to make a branch on ci-actions where those are on in a way that we can use here. Should be possible.

Indanz:
> My argument for merging separately is that the smaller each atomic change through the CI, the easier for someone to come back in the future if there's any issues.

I will re-order the commits so that hopefully at each commit everything still builds and works. That's good to do, no matter how we merge this.

> Accordingly, if you want, I'm happy to shepherd the mechanical git + pull request + merging stuff for splitting it apart and checking the tests pass

The annoying thing of doing that is that it's a serial operation if later commits depend on earlier ones (which is the case with the SP fix). So it seems like a lot of effort with not that much gain.

I'll re-order the commits and compile and test them. Probably won't test all 18 commits one by one, but at least compile each of them for both aarch32 and aarch64.

Indanz:
> > Is there any way to kick off those ci-actions for this PR, or do I need to manually run them one by one?
> 
> These are all globally disabled now, so unfortunately no easy way. I could enable them, but then things will start failing everywhere (apart from hopefully here). Need to think about if there is a good way to make a branch on ci-actions where those are on in a way that we can use here. Should be possible.

Let's do this after this PR is merged, then we can test on master instead of complicating things further. Any additional fixes will be independent anyway, so could go in a new PR.

llvm.cmake
Indanz: Dropped it from this PR.

Indanz:
Okay, I reordered the commits so the SP fix comes early. I compiled all and tested almost all commits on the sabre for both gcc and clang. They all compiled fine, but hang in `MULTICORE0004`, until [SMP: Barrier in ipiStallCoreCallback](https://github.com/seL4/seL4/pull/871/commits/e1ce44cdef2790dcafad755b0d7f56f647f8691e).

It all looks good to me and I think it's ready for merging. @kent-mcleod, if you want to do any further PR splitting, you can do it now, but I think it won't add much and will increase the chance of making a mistake somewhere. I am very leery of such last second changes.

src/arch/arm/kernel/boot.c
```diff
@@ -206,9 +206,14 @@ BOOT_CODE static bool_t init_cpu(void)
      * On ARM SMP, the array index here is the CPU ID
      */
     word_t stack_top = ((word_t) kernel_stack_alloc[CURRENT_CPU_INDEX()]) + BIT(CONFIG_KERNEL_STACK_BITS);
-#if defined(ENABLE_SMP_SUPPORT) && defined(CONFIG_ARCH_AARCH64)
+#ifdef ENABLE_SMP_SUPPORT
+#ifdef CONFIG_ARCH_AARCH64
     /* the least 12 bits are used to store logical core ID */
     stack_top |= getCurrentCPUIndex();
+#elif defined(CONFIG_ARCH_AARCH32)
```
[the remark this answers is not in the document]
Indanz: [Magic](https://github.com/seL4/seL4/blob/475f0911fcd5263b181ab999e75f8bec51df76c9/src/arch/arm/64/head.S#L65)... ;-)

lsf37:
I have retargeted the PR to the branch `smp-tests` which should contain all the test that were previously disabled. I do hope nothing has changed in the review interface, at least there shouldn't be any visible code changes from this.

kent-mcleod (approved):

lsf37 (approved):
I agree with Kent that this looks like a solid improvement. I can't say that I understand the details of the lock here, but it is at least clearer to me than it was before and it fixes tests that failed before (unfortunately not all of them, but some).

From the verification side, there is no proof impact from the things that spill out from the SMP config, so no problem from there.

kent-mcleod:
Stage 1 https://github.com/seL4/seL4/pull/1413

Indanz:
I didn't know that SCHED_CONTEXT_0014 was disabled for some platforms, it not passing is a sign of either a bug in the test, or some SMP issue ("Test deleting a scheduling context while it runs out of budget on another core"). Perhaps it's fixed if I use atomics instead of volatiles in the while loop.

MULTICORE0001 and MULTICORE0002 probably have the same while loop on volatile counter issue as SCHED_CONTEXT_0014. All in all it seems that "volatile" doesn't give us the guarantees we assume. I'll open a PR for sel4test.

kent-mcleod:
Part 2 pr: https://github.com/seL4/seL4/pull/1414

Indanz:
> Perhaps it's fixed if I use atomics instead of volatiles in the while loop.

Trying that out at https://github.com/seL4/sel4test/pull/134.

kent-mcleod:
This PR was merged in 2 stages:
- https://github.com/seL4/seL4/pull/1413
- https://github.com/seL4/seL4/pull/1414

Indanz:
I don't dare pressing the "Rebase and merge button", who knows what's going to happen.

Would've probably been better if stage 2 happened via this PR, now it might look like it didn't got merged and won't show up in merged PRs list. Let's hope no one goes through that list for any reason...

lsf37:
Need to retarget first to master again before we merge. One sec.

lsf37:
If you rebase the branch now onto master, and push here, we should see just the commits that are missing.

Indanz:
There are no differences left, can you "merge" this PR without the tests finishing?

lsf37:
Yes, I think so, if I rebase this directly. Let me see.

lsf37:
It still shows it just as closed if I push the rebased branch, sorry. But let it be recorded here that it was merged successfully via #1413 and #1414.

Indanz:
I could add something to the changelog and we merge that as a placeholder? E.g:

```diff
diff --git a/CHANGES.md b/CHANGES.md
index 5f6fc7eea..a77699044 100644
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
 * riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
   Except for RISCV32 with LLVM clang enabled will default KernelRiscvExtD to OFF.
+* Multiple SMP related bugs fixed.
 
 ### Platforms
 
```

lsf37:
> I could add something to the changelog and we merge that as a placeholder? E.g:
> 
> ```diff
> diff --git a/CHANGES.md b/CHANGES.md
> index 5f6fc7eea..a77699044 100644
> --- a/CHANGES.md
> +++ b/CHANGES.md
> @@ -28,6 +28,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
>  * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
>  * riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
>    Except for RISCV32 with LLVM clang enabled will default KernelRiscvExtD to OFF.
> +* Multiple SMP related bugs fixed.
>  
>  ### Platforms
>  
> ```

Happy for that if you want to, yes. We should have a changelog entry for it.

Indanz:
Added a very sparse changelog. Was afraid that it would get too long if I go into details, but probably went too far the other way. n the other hand, it has all the relevant info and the exact details don't really matter. Open for suggestions.

kent-mcleod (approved):

lsf37 (approved):
That's about as short as it gets :-)

I guess the real change is going to be removing some of the configs from the CAVEATS file and declaring them as supported (e.g. SMP and clang in general). I'd leave that until we have updated the tests properly.

Indanz:
I didn't know that file existed! CAVEATS probably needs updating anyway. But I'd rather do that after seeing no SMP related problems in CI for a few months before promising too much now.

I think there is only one outstanding MCS+SMP bug remaining, #941, but that is a logical bug, not a multithreading bug.

I'll "merge" this PR and close it that way, except if anyone comes with additional changelog requests soon.

kent-mcleod:
In https://github.com/seL4/seL4/pull/867 I was waiting on this PR. Now that this is merged, I plan to update the other one.

kent-mcleod:
I wonder if we should eventually go through and set the maximum exclusive reservation granule size by microarchitecure. The cortex-a53 defines the max to be 16 words.

lsf37 (approved):

PR #872, opened 2022-06-20
mcs: add AUXUPD and GHOSTUPD comments for verification

michaelmcinerney:
This adds AUXUPD and GHOSTUPD comments to aid verification in
reasoning about the length and type of the refill buffer
associated with scheduling contexts

This also add a AUXUPD comment for replies

Signed-off-by: Michael McInerney <michael.mcinerney@proofcraft.systems>

### DIFF
--- a/src/object/objecttype.c
+++ b/src/object/objecttype.c
@@ -573,10 +573,24 @@ cap_t createObject(object_t t, void *regionBase, word_t userSize, bool_t deviceM
 
 #ifdef CONFIG_KERNEL_MCS
     case seL4_SchedContextObject:
+        /** AUXUPD:
+            "(True,
+              ptr_retyps (refills_len (unat \<acute>userSize)
+                                      (size_of TYPE(sched_context_C))
+                                      (size_of TYPE(refill_C)))
+                         (Ptr ((ptr_val \<acute>regionBase) +
+                               word_of_nat (size_of TYPE(sched_context_C))) :: refill_C ptr)
+              \<circ> ptr_retyp (Ptr (ptr_val \<acute>regionBase) :: sched_context_C ptr))" */
+        /** GHOSTUPD:
+            "(True, gs_new_refill_buffer_length (ptr_val \<acute>regionBase)
+                                                (refills_len (unat \<acute>userSize)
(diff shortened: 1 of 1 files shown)


src/object/objecttype.c
```diff
@@ -573,10 +573,17 @@ cap_t createObject(object_t t, void *regionBase, word_t userSize, bool_t deviceM
 
 #ifdef CONFIG_KERNEL_MCS
     case seL4_SchedContextObject:
+        /** AUXUPD: "(True, ptr_retyps (refills_len (unat \<acute>userSize) (size_of TYPE(sched_context_C)) (size_of TYPE(refill_C)))
+          (Ptr ((ptr_val \<acute>regionBase) + word_of_nat (size_of TYPE(sched_context_C))) :: refill_C ptr)
+          \<circ> ptr_retyp (Ptr (ptr_val \<acute>regionBase) :: sched_context_C ptr))" */
```
lsf37, line 578: I realise the layout is copied from/consistent with the other ones, but line break is at 100 so we have a bit of space and it would be nicer to indent properly, i.e. the second line as a parameter to `ptr_retyps`, and the third line on the same semantic level as `ptr_retyps`.

lsf37 (approved):
Content is good, only indentation nitpicks.

michaelmcinerney:
I'm not sure whether the newer indentation is an improvement. There are still a couple lines that go to 110 characters or so within the plain text editor. Within Isabelle this should be shortened a little bit.

lsf37:
> I'm not sure whether the newer indentation is an improvement. There are still a couple lines that go to 110 characters or so within the plain text editor. Within Isabelle this should be shortened a little bit.

That looks much more readable, thanks! We could break after `AUXUPD:` itself or the `True` part of the tuple if that makes it fit within 100.

michaelmcinerney:
The breaking is quite aggressive now, but it does all fit within 100 characters

kent-mcleod (approved):

lsf37:
> The breaking is quite aggressive now, but it does all fit within 100 characters

I'm up for it. Ready to merge from my side (after squashing).

corlewis (approved):

PR #874, opened 2022-07-04
MCS: Restrict reply grant rights

mbrcknl:
This PR represents the first alternative discussed in [RFC-13](https://sel4.atlassian.net/browse/RFC-13). It ensures that in MCS configurations, the grant bit on a server's endpoint capability can be used to control the server's ability to grant capabilities when replying to a call. Without this, it will be difficult to construct a reasonable formal model of system integrity.

The second alternative is given in https://github.com/seL4/seL4/pull/945. It incorporates this PR, and additionally removes the grant bit from the MCS reply object capability. The second PR is motivated by the observation that there is no clear use for the reply object capability grant right once this PR is implemented.

-----------

In MCS configurations, a server's ability to grant capabilities when
replying to a call were previously determined entirely by the grant
right on the reply capability the server supplied when calling
`receiveIPC`. If the server had sufficient untyped memory, it could
create arbitrary reply capabilities, giving itself the ability to grant
to its clients. Consequently, the only way to restrict a server's
ability to grant to its clients was to avoid giving it untyped memory.
This would break the expected authority confinement property.

Contrast this with non-MCS configurations, where the grant rights on a
server's receive endpoint control the server's ability to grant
capabilities when replying to calls. This makes it much easier to
restrict a server's ability to grant when replying.

This commit makes MCS configurations behave more like non-MCS
configurations: In order to grant capabilities when replying to a call,
a server must have both a grant right on the receive endpoint capability
when it calls `receiveIPC` and also a grant right on the reply
capability when it replies.

This came up in a conversation with @kent-mcleod. This is the first
alternative discussed in seL4 RFC-13:

https://sel4.atlassian.net/browse/RFC-13

[RFC-13]: https://sel4.atlassian.net/browse/RFC-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/object/endpoint.c
```diff
@@ -108,8 +110,6 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
         }
         possibleSwitchTo(dest);
 #else
-        bool_t replyCanGrant = thread_state_ptr_get_blockingIPCCanGrant(&dest->tcbState);;
```
kent-mcleod, line 111: It looks like the removal of the extra semicolon here is breaking several of the preprocess checks that should pass.

gernotheiser:
I was at first a bit confused about "want MCS to behave more like non-MCS" (we introduced MCS because we want a different behaviour ;-) I take it you mean "in terms of rights propagation", which I agree with. The present model does indeed seem to have distasteful consequences.
It does seem though that the grant right on the reply cap is fairly redundant (but probably required for consistency).

src/object/endpoint.c
mbrcknl: Yeah. I think this should be fine for the non-MCS proofs, since the `ccorres_rewrite` machinery should normalise to the same thing.

lsf37:
Well spotted, I think the change makes sense, otherwise you always have authority leakage through replies.

There might be one or two more of these tweaks as we start thinking through the access control proofs, but the earlier we catch and fix them, the better.

> It does seem though that the grant right on the reply cap is fairly redundant (but probably required for consistency).

I think it makes sense: we do want a policy-level distinction between reply messages that can send only content and reply messages that can send caps, because it allows us to constrain authority while still allowing info flow in both directions.

lsf37:
I'd be fine with leaving it in for now. It's easier to do a second round that removes it than a second round that adds it back in because we do run into a case where we need it after all.

kent-mcleod:
> I'd be fine with leaving it in for now. It's easier to do a second round that removes it than a second round that adds it back in because we do run into a case where we need it after all.

It's likely easier to take it out than leave it in.  The right hasn't been added to capDL or CAmkES yet, so leaving it in would still require updating them to allow it to be set to allow cap transfer on reply.

mbrcknl:
> > I'd be fine with leaving it in for now. It's easier to do a second round that removes it than a second round that adds it back in because we do run into a case where we need it after all.
> 
> It's likely easier to take it out than leave it in. The right hasn't been added to capDL or CAmkES yet, so leaving it in would still require updating them to allow it to be set to allow cap transfer on reply.

There are also the existing invariant/refinement proofs to consider. In any case, the existing proofs would need to be updated for either choice, so we should perhaps revisit this question when I (or someone else) gets around to doing that update.

lsf37:
> > I'd be fine with leaving it in for now. It's easier to do a second round that removes it than a second round that adds it back in because we do run into a case where we need it after all.
> 
> It's likely easier to take it out than leave it in. The right hasn't been added to capDL or CAmkES yet, so leaving it in would still require updating them to allow it to be set to allow cap transfer on reply.

That is a fair point.

For Matt's example, can we achieve the same policy restriction by doing something to the receive endpoint cap for the lower-trust component? I think not, because replies shortcut directly to the TCB, that's why the PR is putting a flag into the reply object. In fact, the lower-trust component doesn't even need to have any endpoint cap, it'd just invoke the reply.

Which means removing it is an actual reduction in what we can express. Not really a problem, just wanted to make sure I understand what I'm agreeing to :-)

Ok, so if you want to remove Grant for reply caps, I'm on board.

src/object/endpoint.c
lsf37: Yes, the proof test succeeded.

lsf37:
Just as another note: storing rights in objects is usually a red flag, because rights are usually associated with caps and storing them anywhere else but caps opens up time-of-check/time-of-use problems. In this case, the time of check is the receive, and time of use is the reply. By the time we reply the original authorisation might have been revoked or changed on the sender side or on the receiver side (Matt's example is one where rights change on the receiver side).

I don't think this is a problem here, if the user has the correct model in mind, which is that a reply can always happen any time later as soon as one is authorised by a send. The change here doesn't really make this any worse or better than before.

kent-mcleod:
> By the time we reply the original authorisation might have been revoked or changed on the sender side or on the receiver side (Matt's example is one where rights change on the receiver side).

This can occur currently in non-mcs right? I think the TOCTOU issues are inherent with endpoints. We already expect cancelBadgeSends to be called explicitly if a badged endpoint is revoked.  And this only revokes operations in the send queue, it wouldn't cancel operations that were already waiting for a reply.

mbrcknl:
> > This can occur currently in non-mcs right? I think the TOCTOU issues are inherent with endpoints. We already expect cancelBadgeSends to be called explicitly if a badged endpoint is revoked. And this only revokes operations in the send queue, it wouldn't cancel operations that were already waiting for a reply.
> 
> Yes, that is correct, this is not new. I was mainly trying to document that adding the flag doesn't introduce a new problem.

Does this strengthen the case for keeping the grant right on the MCS reply cap? In the sense that it restores the ability to mask the rights after making the receive call? I still can't decide whether it's useful.

With regards to capDL/CAmkES, what's the behaviour now? Do they always give reply-grant rights, or always deny them?

lsf37:
After conversation with @kevinelp and @gernotheiser just now on access control in MCS in general, I think we should bump this up and merge. (We arrived at the same conclusion as Matt here.)

The remaining open question is whether to remove the grant right on Reply objects in general (equivalent to assuming you always have it). I have not managed to come up with any example where removing the right would be a restriction. AFAICS the system is always better modelled by having/not having Grant on the original receive cap. The additional right on the reply cap doesn't add anything.

> Does this strengthen the case for keeping the grant right on the MCS reply cap? In the sense that it restores the ability to mask the rights after making the receive call? I still can't decide whether it's useful.

I don't think it'd add something fundamental: the use case would be that you want to down-grade the rights of a communication that is currently in flight (i.e. client has sent to server, server is about to reply). To do that downgrade successfully, you'd need to be able to do surgery on the server's reply caps, knowing exactly which one it is going to use, and the server would need to be Ok with that and produce a useful result with potentially a computation that was done with different policy assumptions. I don't think that makes much sense. If you change policy, you'd usually want to either abort current transactions and restart or let them go through and apply the policy to new ones, but not expect that in-flight transactions will dynamically adapt to the new policy.

> With regards to capDL/CAmkES, what's the behaviour now? Do they always give reply-grant rights, or always deny them?

They currently don't model the right, which means it will by default be present after retype.

mbrcknl:
@lsf37 I tuned out at a crucial moment in this morning's developer hangout, when you were talking about making an RFC for this and #940.

Was the intention to make separate RFCs for each of this and #940, or a single RFC for both? Separate RFCs might be easier to write, since we can just take one each.

lsf37:
> @lsf37 I tuned out at a crucial moment in this morning's developer hangout, when you were talking about making an RFC for this and #940.
> 
> Was the intention to make separate RFCs for each of this and #940, or a single RFC for both? Separate RFCs might be easier to write, since we can just take one each.

I was initially thinking just one, but you're right, let's do one each. This will also help keep the discussion focussed.

kevinelp:
I find the terminology that has stuck surrounding reply objects confusing especially with the light of the comparisons the previous model (and resume caps in the KeyKOS/EROS literature).  I think we should refer to reply object caps as "reply object caps", not "reply caps".

Pre-MCS, reply caps were synthesized via call and gave the right to reply to the caller. They were stored in either a cnode or the tcb (an optimisation), but they were a reply cap, i.e. a cap that gave the right to reply.

Reply objects are different, they are effectively containers used to contain the right to reply (which is no longer an explicit reply cap). I can squint an see reply objects as a custom-sized 1-slot cnodes that contain a right of reply that can't be moved to a cnode. In theory, one could have an API that moved the right of reply between different reply objects (note: not espousing such an API addition).

So what's my point, besides a vent about terminology? The grant right on a reply object cap is not too dissimilar to adding a hypothetical grant right on a cnode cap in the old model where if the reply cap was invoked via a cnode cap with grant, it would magically get more power than it originally had. That makes little sense to me as it implies the reply right conveyed can be increased based on where it is stored, not what it innately conveys. Note: I don't have an issue with downgrades, but that is not the behaviour (and I'm not proposing downgrading rights via reply object cap rights).

I support the removal of rights on reply object caps. Is there anything more to invoking reply objects other than exercise the right it contains if it contains a right to reply?

mbrcknl:
> I find the terminology that has stuck surrounding reply objects confusing especially with the light of the comparisons the previous model (and resume caps in the KeyKOS/EROS literature). I think we should refer to reply object caps as "reply object caps", not "reply caps".
> 
> Pre-MCS, reply caps were synthesized via call and gave the right to reply to the caller. They were stored in either a cnode or the tcb (an optimisation), but they were a reply cap, i.e. a cap that gave the right to reply.
> 
> Reply objects are different, they are effectively containers used to contain the right to reply (which is no longer an explicit reply cap). I can squint an see reply objects as a custom-sized 1-slot cnodes that contain a right of reply that can't be moved to a cnode. In theory, one could have an API that moved the right of reply between different reply objects (note: not espousing such an API addition).
> 
> So what's my point, besides a vent about terminology? The grant right on a reply object cap is not too dissimilar to adding a hypothetical grant right on a cnode cap in the old model where if the reply cap was invoked via a cnode cap with grant, it would magically get more power than it originally had. That makes little sense to me as it implies the reply right conveyed can be increased based on where it is stored, not what it innately conveys. Note: I don't have an issue with downgrades, but that is not the behaviour (and I'm not proposing downgrading rights via reply object cap rights).
> 
> I support the removal of rights on reply object caps. Is there anything more to invoking reply objects other than exercise the right it contains if it contains a right to reply?

I read your comments here and on #945 in the wrong order, but I think I agree.

Just to be clear, though: The only thing this PR does is make the grant bit on the receiver's endpoint capability effective as an upper bound on the receiver's ability to grant via reply. Previously, that bit was entirely ignored. This PR keeps the grant bit on the reply object capability, but that bit can now only be used to mask any grant rights conferred by the receive endpoint capability. It cannot be used to confer a right to grant via reply when the receive endpoint capability does not confer that right.

The other PR (#945) takes the further step to remove the grant bit on the reply object capability, if we agree there is no use for it. I had thought there might be a use for a per-capability mask, but [as noted in the other PR](https://github.com/seL4/seL4/pull/945#issuecomment-1301609528), I've now convinced myself that my imagined use case is not realistic.

mbrcknl:
I've posted an RFC at https://sel4.atlassian.net/browse/RFC-13. I'd welcome any feedback on it before (or after) I post it to the mailing list.

lsf37 (approved):
The change looks good to me (together with #945) as implementation of the RFC.

This will need updates in MCS specs, AInvs, and Refine, so I'm adding @michaelmcinerney to have a look as well, but the interference does look very small.

If we have actually covered everything correctly will come out in the access control proofs, which are still a bit in the future, but we're getting there.

michaelmcinerney (approved):
I haven't kept up with the motivation here, but considering only the impact it should have on the MCS proofs, I do think it will be very small, so I'm happy with that

lsf37:
(rebased)

Indanz:
Should we close this one in favour of https://github.com/seL4/seL4/pull/945? I think the decision was to remove the grant right from reply caps.

lsf37:
> Should we close this one in favour of #945? I think the decision was to remove the grant right from reply caps.

The choice was for #945, but both are necessary, #945 stacks on top of this one.

src/object/reply.c
```diff
@@ -30,6 +30,9 @@ void reply_push(tcb_t *tcb_caller, tcb_t *tcb_callee, reply_t *reply, bool_t can
 
     /* link caller and reply */
     reply->replyTCB = tcb_caller;
+
+    /* canGrant should have already been set according to the grant
+       right on the receiver's endpoint capability. */
```
Indanz, line 35: This seems like a random comment to have here (also, comment style is non-standard).

Indanz:
> > Should we close this one in favour of #945? I think the decision was to remove the grant right from reply caps.
> 
> The choice was for #945, but both are necessary, #945 stacks on top of this one.

Okay, I missed that. This is quite confusing. Looking at the PRs in isolation does not give a clear overview of what the end result will be, so in future I would prefer normal PRs against master, as that is easier to review. Will the proofs also be done in two steps?

Issue #497, opened 2021-08-20
bingup the new platform not in Linux repository


neil1899:
hi @lsf37 
about porint doc：
“You may need to add device nodes depending on what Linux uses and what is required by seL4. For example, in the port of the Rockpro64 a memory node and an extra timer node were needed so these had to be defined manually.”

how can I get the required by seL4? have any doc about it?

lsf37:
Sorry, no idea, I don't know much about Linux. I'm not the right person for that kind of question. Maybe someone on the [mailing list](https://lists.sel4.systems/postorius/lists/devel.sel4.systems/) or [discourse](https://sel4.discourse.group) can help.

neil1899:
OK thanks

PR #876, opened 2022-07-07
ci: add MCS to preprocess-deploy

mbrcknl:
This commit will cause the preprocess-deploy workflow to fail if there
are changes to the preprocessed seL4 sources in MCS configurations
covered by the proofs. This means that changes to proof-relevant MCS
configurations will require a manual verification manifest update.

Previously, only non-MCS configurations were covered, which meant that
new verification manifests would be automatically deployed even if there
were proof-relevant changes to MCS code.

### DIFF
--- a/.github/workflows/preprocess-deploy.yml
+++ b/.github/workflows/preprocess-deploy.yml
@@ -31,11 +31,19 @@ jobs:
     strategy:
       matrix:
         arch: [ARM, ARM_HYP, AARCH64, RISCV64, X64]
-        # no MCS here, auto-updating mcs.xml should be a separate job.
+        feature: ["", MCS]
+        exclude:
+          - arch: ARM_HYP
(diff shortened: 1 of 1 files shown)


mbrcknl:
Switching to draft while we think a bit more about seL4/l4v#497.

lsf37:
Maybe I should phrase this differently: the current setup is already too unified and makes the assumption that in the verification the `rt` and `master` branches verify the same revision of the seL4 kernel. That assumption was fine before the C proofs started, but it is not true any more. Of course they should eventually converge and ideally remain at the same one, but that's the goal state, not the necessarily the current state.

mbrcknl:
Closing for now. Will open new PRs when I have a proper solution to #497, i.e. automatic CPP-compatible updates in a separate stream for MCS.

PR #877, opened 2022-07-11
testing: do not merge

lsf37:
Testing build + hardware runs for further MCS platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
      SPDX-License-Identifier: GPL-2.0-only
 -->
 
+# Testing
+
 The seL4 microkernel
 ====================
 
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
      SPDX-License-Identifier: GPL-2.0-only
 -->
 
-# Testing
-
 The seL4 microkernel
 ====================
 


Indanz:
Except for time-outs, IMX8MQ_EVK fails with:
`seL4 failed assertion 'NODE_STATE(ksCurTime) < MAX_RELEASE_TIME' at /github/workspace/kernel/include/kernel/thread.h:235 in function updateTimestamp`
Perhaps we should add a check early at boot to catch any `ksCurTime` close to `MAX_RELEASE_TIME` (e.g. within 10 years).

Some TX1, TX2 and PC99 tests can't find seL4 images.

lsf37:
> Some TX1, TX2 and PC99 tests can't find seL4 images.

I enabled some more configs after the build had already started and it looks like the hardware-run matrix picked those up already. I've restarted it to get the more complete picture.

Not sure why we were getting so many timeouts on the machine queue. There were two boards that people held locks for and seem to have forgotten to release. That should now be gone. Let's see what the new run does.

Indanz:
[HW Run (ODROID_XU4, armv7a, clang)](https://github.com/seL4/seL4/runs/7293526359?check_suite_focus=true#logs) fails with:
>   Running test CACHEFLUSH0001 (Test a cache maintenance on pages)
>  		<failure type="failure">*ptrc == 0xBEEFCAFE at line 104 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/cache.c</failure>

I'm not convinced that `test_page_flush` is correct: It relies on no cache miss or prefetching happening between the `seL4_ARM_Page_Invalidate_Data` call and the `*ptr` write. I don't think that's guaranteed.

In case the invalidation performs an implicit clean, the result is the same as for `seL4_ARM_Page_CleanInvalidate_Data` above, which is already tested.

Anyway, this has nothing to do with MCS, nor even really with SMP, except if the system calls are implemented differently.

[HW Run (TX2, armv8a, clang)](https://github.com/seL4/seL4/runs/7293526691?check_suite_focus=true#logs) fails with:
>   Running test SCHED_CONTEXT_0014 (Test deleting a scheduling context while it runs out of budget on another core)
>  
>  [[Timeout]]
>  None

Not sure what's going on here.

PC99 nehalem clang 64 started Linux instead of seL4.

lsf37:
Ok, so after a few re-runs, failures for these boards seem to be real: `IMX8MQ_EVK`, `IMX8MM_EVK`, `TX2`, `ODROID_XU4`

I agree that the cache flush failure for Odroid doesn't seem to be an MCS issue as such, but it is only triggered on MCS currently. I might disable that that one config an open an issue in sel4test for that.

Similarly for the `TX2`, the failing combination for `SCHED_CONTEXT_0014` is MCS+hyp+MC. It doesn't seem to get stuck for non-hyp configs, which is interesting and might point to a bug there.

`IMX8MQ_EVK` seems to be failing `SCHED0012` on all MCS configs, so I'll leave that board for disabled for MCS for now. We can add it again when we have figured out what is going on there. I didn't spot any assertion failure on `MAX_RELEASE_TIME` any more, but we might still want to add the check that @Indanz mentioned.

`IMX8MM_EVK` is failing for MULITCORE001 and MULITCORE002 in some configs, and it seems to be hung at kernel boot for `IMX8MM_EVK_debug_SMP_MCS_gcc_64` (timeout after "starting kernel"), so I'll leave MCS+SMP disabled for that for now.

lsf37:
I'll kick off another run which should now exclude the failing platforms.

lsf37:
Ok, with the current exclusions in seL4/ci-actions#224 the hardware tests are now all succeeding.

If there are no further concerns, I'd close the PR here and merge the one in ci-actions.

lsf37:
This seems to be working now and the ci-actions PR has been merged.

PR #878, opened 2022-07-11
Add Initial draft of oxidised manual

Ben-PH:
Copied over from my original [personal project](https://github.com/Ben-PH/seL4_rust-doc/tree/42ce1e43f45ed4b7654efd02a09a19e4d6444150) with some minor edits appropriate to it now existing in the upstream codebase

I've marked this as a draft. Up to this point, this has been mostly a self-study exercise to help me grok the API. I would be happy to contribute some of my time freely to bring it up to standard for the upstream code-base, if it's considered valuable. Some things I'm particularly interested in

- A process to make maintaining this documentation have low overhead 
- Extending to other architectures
- Integrate feedback about how to interpret the API in an idiomatic rust manner

You can see the rendered documentation (minus the edits mentioned) [here](https://ben-ph.github.io/seL4_rust-doc/sel4_rust_doc/index.html)


oxidised_manual/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "sel4_rust-doc"
+version = "0.0.1"
+edition = "2018"
```
axel-h, line 4: make this 2022?

Ben-PH:
It's what I used as a reference for putting this together. I wanted to avoid having a moving target for a reference.

At some point, I would like to generate this manual from the same templates as the manual that's currently being referred to.

oxidised_manual/Cargo.toml
Ben-PH: Making it 2021 (2022 not supported)

Ben-PH:
appologies. Forgot to sign off on the commit. Will do that now...

lsf37:
While I think this is pretty cool, there is a fundamental problem: if taken to the end, this essentially duplicates the manual and we're already behind with just keeping one manual up to date.

So we either need to drive this to the point where it replaces the old manual (which is entirely up for debate), or we need to generate something from a common source. Or something in between, e.g. there could be a manual that describes concepts which is shared, and a separate part that is API reference which is generated.

In all cases, we need to make sure that both APIs (C libsel4 and Rust) are documented. 

Do you have any thoughts on how that could be achieved without a large amount of duplication?

lsf37:
For clarification: is this attempting to be documentation only or is it describing one of the existing seL4 Rust libraries (e.g. from Robigalia or Icecap)?

Ben-PH:
It's been too long since I've had a dig into the codebase to look at generation.

I could see myself having some fun to extend the generation to build this documentation...

lsf37:
> It started off as a close to direct transcription of the manual. At various points, I started fiddling around with making it more "rusty" (such returning `Result<T, Sel4Error>`, and some other experiments).
> 
> seL4 was the primary basis. If it's an important point, I'll have to double check by going through the commit history (happy to do so), but I don't recall any direct influence from existing seL4 rust libraries (I last worked on this about a year ago, so memory is getting vague), beyond picking up ideas on making the rust interpretation of the api "more rusty".

I like the idea of a rusty seL4 library on top of the generated bindings, I think that could be really nice. If we do that, it'd be good if that library actually existed to make sure it all makes sense and works together properly. Icecap has done a lot in that direction already, but it's specific to AArch64. I don't quite know where Robigalia is at, and there is a Google team that promised to publish their seL4 Rust work at some point. Such a library would probably live in its own repository, not the kernel repo itself, but that can be a separate point of discussion.

My feeling is that it would make sense for the foundation to consolidate and start maintaining that library layer at some point and that this documentation would best fit there. There'd still be the duplication problem with the generated libsel4 bindings, but if there is an actual higher-level library, the duplication would be more conceptual than literal as it is now. I.e. it'd be fine to diverge and only be loosely coupled. Of course this doesn't fix the problem that we'd need people to actually maintain it.

Ben-PH:
I'm really happy to hear that!

It's getting late for me (paris time). I'd welcome any ideas/suggestions you have on areas of focus I can work on to add rust support to this project (e.g. grok the manual generation, work on generating code from the bitfields, etc).

I'll try to join the dev hangouts as well.

lsf37:
I'll close the PR here, but I think this work could be used in the upcoming work on a "standard" seL4 Rust library outside of the seL4 repo itself.

PR #880, opened 2022-07-12
RISC-V PTE compliance.

ahmedcharles:
The RISC-V Instruction Set Manual - Volume II: Privileged Architecture:

For non-leaf PTEs, the D, A, and U bits are reserved for future standard
use. Until their use is defined by a standard extension, they must be
cleared by software for forward compatibility.

Signed-off-by: Ahmed Charles <acharles@outlook.com>

### DIFF
--- a/src/arch/riscv/kernel/vspace.c
+++ b/src/arch/riscv/kernel/vspace.c
@@ -70,10 +70,10 @@ static pte_t pte_next(word_t phys_addr, bool_t is_leaf)
 
     return pte_new(ppn,
                    0,     /* sw */
-                   1,     /* dirty */
-                   1,     /* accessed */
+                   is_leaf ? 1 : 0,     /* dirty (leaf)/reserved (non-leaf) */
+                   is_leaf ? 1 : 0,     /* accessed (leaf)/reserved (non-leaf) */
                    1,     /* global */
-                   0,     /* user */
+                   is_leaf ? 0 : 0,     /* user (leaf)/reserved (non-leaf) */
                    exec,  /* execute */
                    write, /* write */
                    read,  /* read */
@@ -180,10 +180,10 @@ BOOT_CODE void map_it_pt_cap(cap_t vspace_cap, cap_t pt_cap)
     *targetSlot = pte_new(
                       (addrFromPPtr(pt) >> seL4_PageBits),
                       0, /* sw */
-                      1, /* dirty */
-                      1, /* accessed */
+                      0, /* dirty (reserved non-leaf) */
+                      0, /* accessed (reserved non-leaf) */
                       0,  /* global */
-                      0,  /* user */
+                      0,  /* user (reserved non-leaf) */
                       0,  /* execute */
                       0,  /* write */
                       0,  /* read */
@@ -207,10 +207,10 @@ BOOT_CODE void map_it_frame_cap(cap_t vspace_cap, cap_t frame_cap)
     *targetSlot = pte_new(
                       (pptr_to_paddr(frame_pptr) >> seL4_PageBits),
                       0, /* sw */
-                      1, /* dirty */
-                      1, /* accessed */
+                      1, /* dirty (leaf) */
+                      1, /* accessed (leaf) */
                       0,  /* global */
-                      1,  /* user */
+                      1,  /* user (leaf) */
                       1,  /* execute */
                       1,  /* write */
                       1,  /* read */
@@ -531,10 +531,10 @@ void unmapPageTable(asid_t asid, vptr_t vptr, pte_t *target_pt)
     *ptSlot = pte_new(
                   0,  /* phy_address */
                   0,  /* sw */
-                  0,  /* dirty */
-                  0,  /* accessed */
+                  0,  /* dirty (reserved non-leaf) */
+                  0,  /* accessed (reserved non-leaf) */
                   0,  /* global */
-                  0,  /* user */
+                  0,  /* user (reserved non-leaf) */
                   0,  /* execute */
                   0,  /* write */
                   0,  /* read */
@@ -656,10 +656,10 @@ static pte_t CONST makeUserPTE(paddr_t paddr, bool_t executable, vm_rights_t vm_
         return pte_new(
                    paddr >> seL4_PageBits,
                    0, /* sw */
-                   1, /* dirty */
-                   1, /* accessed */
+                   1, /* dirty (leaf) */
+                   1, /* accessed (leaf) */
                    0, /* global */
-                   1, /* user */
+                   1, /* user (leaf) */
                    executable, /* execute */
                    RISCVGetWriteFromVMRights(vm_rights), /* write */
                    RISCVGetReadFromVMRights(vm_rights), /* read */
@@ -773,10 +773,10 @@ static exception_t decodeRISCVPageTableInvocation(word_t label, word_t length,
                         PTE_PTR(cap_page_table_cap_get_capPTBasePtr(cap)));
     pte_t pte = pte_new((paddr >> seL4_PageBits),
                         0, /* sw */
-                        1, /* dirty */
-                        1, /* accessed */
+                        0, /* dirty (reserved non-leaf) */
+                        0, /* accessed (reserved non-leaf) */
                         0,  /* global */
-                        0,  /* user */
+                        0,  /* user (reserved non-leaf) */
                         0,  /* execute */
                         0,  /* write */
                         0,  /* read */


kent-mcleod (approved):
@lsf37, this will require a proof update.

src/arch/riscv/kernel/vspace.c
```diff
@@ -180,8 +180,8 @@ BOOT_CODE void map_it_pt_cap(cap_t vspace_cap, cap_t pt_cap)
     *targetSlot = pte_new(
                       (addrFromPPtr(pt) >> seL4_PageBits),
                       0, /* sw */
-                      1, /* dirty */
-                      1, /* accessed */
+                      0, /* dirty */
```
axel-h, line 183:
```suggestion
                      0, /* reserved */
```
I think the comment should be update to make clear these bit mean something different now

src/arch/riscv/kernel/vspace.c
```diff
@@ -180,8 +180,8 @@ BOOT_CODE void map_it_pt_cap(cap_t vspace_cap, cap_t pt_cap)
     *targetSlot = pte_new(
```
axel-h, line 180: It might be a pain for the proofs, but having a `pte_new_leaf()` and `pte_new_table()` might be useful to make the code more clear.

src/arch/riscv/kernel/vspace.c
```diff
@@ -70,10 +70,10 @@ static pte_t pte_next(word_t phys_addr, bool_t is_leaf)
 
     return pte_new(ppn,
                    0,     /* sw */
-                   1,     /* dirty */
-                   1,     /* accessed */
+                   is_leaf ? 1 : 0,     /* dirty leaf/reserved non-leaf */
+                   is_leaf ? 1 : 0,     /* accessed leaf/reserved non-leaf */
```
lsf37, line 74: Does anyone know why these were set to 1 in the first place? We don't usually used dirty/accessed in the kernel.

src/arch/riscv/kernel/vspace.c
```diff
@@ -207,10 +207,10 @@ BOOT_CODE void map_it_frame_cap(cap_t vspace_cap, cap_t frame_cap)
     *targetSlot = pte_new(
                       (pptr_to_paddr(frame_pptr) >> seL4_PageBits),
                       0, /* sw */
-                      1, /* dirty */
-                      1, /* accessed */
+                      1, /* dirty leaf/reserved non-leaf */
+                      1, /* accessed leaf/reserved non-leaf */
```
lsf37, line 211: Does this not have to be 0 for non-leaf?

lsf37:
I will have to look into this for verification. This could be a major problem, because PTE encoding is currently uniform for RISCV (that was on of the nice things about the architecture) and there is not necessarily enough information to tell what level a PTE is on.

One way out could be to always have the dirty and accessed bits on 0 for new PTEs. Did we have any specific reasons to set them to 1?

src/arch/riscv/kernel/vspace.c
axel-h: That was done by @heshamelmatary in the initial port. So maybe it has no particular reason as the spec was not saying much.

src/arch/riscv/kernel/vspace.c
lsf37: We could wrap `pte_new`, I don't think it would be too catastrophic, but I'd first like to explore if we can rescue the uniform encoding. If not this is going to be months of work, because it kills one of the fundamental formalisation assumptions.

src/arch/riscv/kernel/vspace.c
ahmedcharles: For leaf tables, they have to be set to 1 if you want the OS to ignore them.

The privileged spec allows two implementations:

- When a virtual page is accessed and the A bit is clear, or is written and the D bit is clear, a
  page-fault exception is raised.
- When a virtual page is accessed and the A bit is clear, or is written and the D bit is clear, the
  implementation sets the corresponding bit(s) in the PTE. The PTE update must be atomic
  with respect to other accesses to the PTE, and must atomically check that the PTE is valid
  and grants sufficient permissions. Updates of the A bit may be performed as a result of
  speculation, but updates to the D bit must be exact (i.e., not speculative), and observed in
  program order by the local hart. Furthermore, the PTE update must appear in the global
  memory order no later than the explicit memory access, or any subsequent explicit memory
  access to that virtual page by the local hart. The ordering on loads and stores provided by
  FENCE instructions and the acquire/release bits on atomic instructions also orders the PTE
  updates associated with those loads and stores as observed by remote harts.
  The PTE update is not required to be atomic with respect to the explicit memory access that
  caused the update, and the sequence is interruptible. However, the hart must not perform
  the explicit memory access before the PTE update is globally visible.

Ideally, seL4 would expose these to user space but it seems that this is not currently the case.

src/arch/riscv/kernel/vspace.c
ahmedcharles: I decided to make all of the comments the same, though I can change that if desired. Let me know.

ahmedcharles:
With the caveat that I've never verified anything, I think the easy thing for verification would be to always let them be specified by the user for leaf entries. `seL4_RISCV_PageTable_Map` would always set them to `0` and `seL4_RISCV_Page_Map` would let them be specified by the user. `seL4_RISCV_Page_GetAddress` would also have to return them.

src/arch/riscv/kernel/vspace.c
lsf37: It confused me in this one, so it's probably better to change. Maybe `/* accessed (leaf) */` would work, to indicate that we haven't forgotten about the leaf/no-leaf distinction and know which case we are in.

lsf37:
> With the caveat that I've never verified anything, I think the easy thing for verification would be to always let them be specified by the user for leaf entries. `seL4_RISCV_PageTable_Map` would always set them to `0` and `seL4_RISCV_Page_Map` would let them be specified by the user. `seL4_RISCV_Page_GetAddress` would also have to return them.

It might be much less problematic than I thought initially, because I just checked and contrary to what I remembered we do already distinguish leaf from non-leaf PTEs in the formalisation. This means it is just a question of formulating what the encoding has to be and while that is some work that needs to be done, it should not be a big deal.

src/arch/riscv/kernel/vspace.c
lsf37: Thanks for looking these up.

Ok, this means we do want to set them to 1 for now. Exporting the mechanism to user space would be an option, but it would be an API change that we should do separately if we decide to do it, and it should go through an RFC.

ahmedcharles:
Thanks, that's good to know. I wouldn't mind trying to do that but I've never even run the proofs so far, so perhaps not the best idea.

What about exposing it to user space? Along with RSW (bits 8-9) in the PTE?

lsf37:
> Thanks, that's good to know. I wouldn't mind trying to do that but I've never even run the proofs so far, so perhaps not the best idea.

That's Ok, I'll have a look into it (might take a few days)

> What about exposing it to user space? Along with RSW (bits 8-9) in the PTE?

In terms of the proof I don't think it would be a big change, but it'd be an API change that needs an RFC. Could be done, but should be done separately. If we do it for RISCV, there is the immediate question whether something similar should be done for Arm and x86.

src/arch/riscv/kernel/vspace.c
axel-h: Would be something like https://github.com/axel-h/seL4/commit/060b0cdee851a4aecc49b5f7d9f9a6ac31b67428 then that I had in mind

lsf37:
Thanks for the update. From my side this is done for C, and I'll have a look at updating the proofs before we merge.

ahmedcharles:
Thanks.

src/arch/riscv/kernel/vspace.c
heshamelmatary: We always want to leave them as 1s for the kernel, for leaf pages.

WRT user pages, I didn't see a good reason to expose D/A to userspace back then as no project/lib (even on Arm and x86) needed to make use of them for page placement policies; seL4/userspace don't do (demand) paging/swapping to disk anyway.

ahmedcharles: I think the primary benefit of exposing it to user space will be for people writing dynamic systems in the future. I agree that there are currently no systems that use seL4 in that way, but I'd certainly like there to be.

pingerino: The UNSW AOS course has students implement demand paging :)

lsf37:
Ok, proofs are ready to go, so I'll merge both.

PR #882, opened 2022-07-14
Small common_riscv.lds improvements

kent-mcleod:
- Move the .boot.bss section into the range of memory that can be reclaimed after boot.
- Mark the idle thread symbol as NOLOAD to avoid it taking up space in the ELF file.

### DIFF
--- a/src/arch/arm/common_arm.lds
+++ b/src/arch/arm/common_arm.lds
@@ -59,7 +59,7 @@ SECTIONS
         *(.data)
     }
 
-    .bss . : AT(ADDR(.bss) - KERNEL_OFFSET)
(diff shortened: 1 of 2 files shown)


lsf37:
All relevant tests have passed. `haswell3` still has boot problems, and @andybui01 is holding the lock on `odroidxu4_1`, so these are all unrelated (and not RISCV anyway).

heshamelmatary:
The idle thread section might make use of a comment stating it's not expected to contain any code, otherwise this might be confusing.

Also, I have seen some bugs with llvm/lld where it allocates space in the ELF for .bss. You might want to mark .bss.* as `NOLOAD` as well.

PR #885, opened 2022-07-19
structures: fix MCS 64-bit notification obj size

lsf37:
The kernel expects object sizes to be powers of two for size and alignment computations.

- add missing padding for MCS 64-bit configurations for notifications (other configs were already fine)
- add missing padding for reply object struct
- strengthen compile time assertion to catch discrepancies in the future.

This most likely did not trigger any actual bugs because we don't tend to do much computation with the object size directly for these specific types, but verification predicates do need to do those computations and rely on them lining up.

### DIFF
--- a/include/object/structures.h
+++ b/include/object/structures.h
@@ -396,25 +396,28 @@ struct reply {
      * (the last caller before the server) or another reply object. 0 if no scheduling
      * context was passed along the call chain */
     call_stack_t replyNext;
+
+    /* Unused, explicit padding to make struct size the correct power of 2. */
+    word_t padding;
 };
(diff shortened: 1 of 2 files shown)


lsf37:
Just saw that there are already assertions, but they are not strong enough. Will adjust.

lsf37:
Which leads me to discover that the size of reply objects is also wrong.

lsf37:
Ok, now ready for review with compile time assertions adjusted for the other relevant object types as well.

kent-mcleod (approved):

michaelmcinerney (approved):

Issue #827, opened 2022-04-04
Use CMake file(SIZE ...) instead of stat


axel-h:
Follow-up from https://github.com/seL4/seL4/pull/826

Avoid the stat utility altogether, by using the built-in file (SIZE ...) command that became available in CMake version 3.14.

lsf37:
The point in #826 was not to do that yet, because we don't want to require CMake 3.14. We should revisit this in a year or two.

axel-h:
Ubuntu 20.4/LTS has CMake 3.16, see https://packages.ubuntu.com/focal/cmake
Debian Bullseye has CMake 3.18 https://packages.debian.org/bullseye/cmake

Given that a switch from Ubuntu 18.4 to 20.4 is planned anyway, this seems the only practical blocker.

lsf37:
I don't think we're planning to give up support for 18.04, we're just switching the default to 20.04 because it's more common (although 22 is coming out, so maybe that should be the target instead, there hasn't been any discussion about this yet that I'm aware of).

End of life for Ubuntu 18.04 is Apr 2023, so that would be a good time to drop support for older tool chains, I think.

ratmice:
I don't know how much it matters for seL4 proper, but it looks like ubuntu 22 is the first one above that has `> cmake 3.19`, which added cmake builtin support for json.  It could perhaps be nice if some things were driven by JSON like lists of architectures.

The cmake <-> cargo integration recently gained support for this, so on such versions it no longer requires preinstalled **host** tools, as cmake can just directly communicate with cargo via json.

So thats a :+1: from me if you want to make the leap to 22.

axel-h:
In CMake 3.17 `CMAKE_CURRENT_FUNCTION_LIST_DIR` was added, which comes handy in some cases also.

lsf37:
So, this is complicated. The user experience of getting the right build toolchain installed is already pretty complex. If we add to that a step to manually upgrade `cmake`, I don't think that is a good idea. We might do that for Ubuntu 18.04 in 2023, but Ubuntu 20.04 is going to be around and extremely common for quite a while, as will be Debian Bullseye, so going beyond cmake 3.16 will make things harder for a whole lot of people.

To upgrade manually, you roughly have to
 - uninstall current version
 - either download and compile from source or
 - add kitware apt repo, install their keys, and install from there

It's not impossible, but I can see it going wrong plenty.

[...]

Issue #411, opened 2021-06-21
What needs to be exported when installing a kernel CMakeLists project?


kent-mcleod:
https://github.com/seL4/seL4/pull/410 and https://github.com/seL4/seL4/pull/408 are PR candidates for adding basic support for installing the result of a kernel build to some external location so that it can be consumed by some other set of build tools.  This basic support only installs the kernel.elf binary, and user level header and source files that make up the kernel API. 

This issue is for tracking which other files should be installed and what changes would be required for supporting that.

## Files potentially relevant for non-CMake user builds
- KernelDTBPath: The device tree used by the kernel when generating platform headers on Arm and RISC-V

## Files needed by the elfloader project
- platform_gen.yaml: A description of how the platform's physical address space is partitioned into Ram and device memory.  Used by the elfloader when calculating absolute addresses when linking.
- KernelDTBPath: See above
- hardware_gen.py tool: A python tool for parsing a device tree at compile time and generating C Header files containing extracted properties about the target platform.  This tool includes: tools/hardware/*, tools/hardware.yml, tools/hardware_gen.py and tools/hardware_schema.yml

## Files needed for importing into another CMake project
- CMAKE_TOOLCHAIN_FILE: The toolchain file that describes the cross-compilation settings for the target platform
- libsel4/CMakeLists.txt: CMake file for building libsel4 from the already exported headers and source files.
- seL4KernelExport.cmake: CMake export file ([info](https://cmake.org/cmake/help/latest/command/install.html#export)) for defining CMake imported targets corresponding to installed targets (kernel.elf, kernel_Config, sel4_Config, sel4_autoconf).
- Kernel CMake helper modules (tools/helpers.cmake, tools/flags.cmake): These two files can be used by user level projects for sharing some common CMake functionality, and also for importing additional build flags defined by the kernel but aren't present in the toolchain.cmake file. This is typically `-march` and `-mabi` and related flags that encapsulate which hardware functionality that the kernel is configured to support or not. IE if the FPU is disabled, then these flags would define an march that doesn't use the FPU.
- Exported CMake cache settings: For config settings defined by our config_string, config_choice, config_option functions, their option definitions end up as a `#define` in a kernel_Config library and also as an entry in the CMake config cache. In order to continue accessing these config options via the CMake cache it would be useful to get the add_config_library() method to generate option files in formats other than C header files. This could then be used to import all of the CMake-level kernel options into another CMake project.
[...]

PR #893, opened 2022-07-27
Add json file with config values

ahmedcharles:
Signed-off-by: Ahmed Charles <acharles@outlook.com>

### DIFF
--- a/tools/helpers.cmake
+++ b/tools/helpers.cmake
@@ -204,30 +204,32 @@ function(GenProofsBFTarget target_name target_file pbf_path pbf_target prunes de
     )
 endfunction(GenProofsBFTarget)
 
-macro(cfg_str_add_disabled cfg_str name)
+macro(cfg_str_add_disabled cfg_str cfg_json name)
     # include in liner files can't use double slash comments
     list(APPEND ${cfg_str} "/* disabled: CONFIG_${name} */")
+    string(JSON ${cfg_json} SET ${${cfg_json}} "${name}" "null")
 endmacro()
 
-macro(cfg_str_add_define cfg_str name value comment)
+macro(cfg_str_add_define cfg_str cfg_json name value comment)
     set(cfg_define_str "#define CONFIG_${name}  ${value}")
     if(NOT "${comment}" STREQUAL "")
         # include in liner files can't use double slash comments
         string(APPEND cfg_define_str "  /* ${comment} */")
     endif()
     list(APPEND ${cfg_str} ${cfg_define_str})
+    string(JSON ${cfg_json} SET ${${cfg_json}} ${name} "\"${value}\"")
 endmacro()
 
-macro(cfg_str_add cfg_str name value)
+macro(cfg_str_add cfg_str cfg_json name value)
     if("${value}" STREQUAL "")
-        cfg_str_add_define(${cfg_str} ${name} "" "empty")
+        cfg_str_add_define(${cfg_str} ${cfg_json} ${name} "" "empty")
     else()
-        cfg_str_add_define(${cfg_str} ${name} "${value}" "")
+        cfg_str_add_define(${cfg_str} ${cfg_json} ${name} "${value}" "")
     endif()
 endmacro()
 
-macro(cfg_str_add_as_1 cfg_str name var)
-    cfg_str_add_define(${cfg_str} ${name} "1" "${var}=${${var}}")
+macro(cfg_str_add_as_1 cfg_str cfg_json name var)
+    cfg_str_add_define(${cfg_str} ${cfg_json} ${name} "1" "${var}=${${var}}")
 endmacro()
 
 # config_option(cmake_option_name c_config_name doc DEFAULT default [DEPENDS deps] [DEFAULT_DISABLE default_disabled])
@@ -289,12 +291,14 @@ function(config_option optionname configname doc)
         set(${optionname}_DISABLED TRUE CACHE INTERNAL "" FORCE)
     endif()
     set(local_config_string "${configure_string}")
+    set(local_config_json "${configure_json}")
     if(${optionname})
-        cfg_str_add_as_1(local_config_string ${configname} ${optionname})
+        cfg_str_add_as_1(local_config_string local_config_json ${configname} ${optionname})
     else()
-        cfg_str_add_disabled(local_config_string ${configname})
+        cfg_str_add_disabled(local_config_string local_config_json ${configname})
     endif()
     set(configure_string "${local_config_string}" PARENT_SCOPE)
+    set(configure_json "${local_config_json}" PARENT_SCOPE)
 endfunction(config_option)
 
 # Set a configuration option to a particular value. This value will not appear in
@@ -304,13 +308,13 @@ macro(config_set optionname configname value)
     set(${optionname} "${value}" CACHE INTERNAL "" FORCE)
     set(c_define "CONFIG_${configname}")
     if("${value}" STREQUAL "OFF")
-        cfg_str_add_disabled(configure_string ${configname})
+        cfg_str_add_disabled(configure_string configure_json ${configname})
     else()
         if("${value}" STREQUAL "ON")
-            cfg_str_add_as_1(configure_string ${configname} ${optionname})
+            cfg_str_add_as_1(configure_string configure_json ${configname} ${optionname})
         else()
             # we have to quote ${value} here because it could be empty
-            cfg_str_add(configure_string ${configname} "${value}")
+            cfg_str_add(configure_string configure_json ${configname} "${value}")
         endif()
     endif()
 endmacro(config_set)
@@ -344,6 +348,7 @@ function(config_string optionname configname doc)
     endif()
     set(valid ON)
     set(local_config_string "${configure_string}")
+    set(local_config_json "${configure_json}")
     if(NOT "${CONFIG_DEPENDS}" STREQUAL "")
         # Check the passed in dependencies. This loop and logic is inspired by the
         # actual cmake_dependent_option code
@@ -398,9 +403,10 @@ function(config_string optionname configname doc)
         else()
             set(quote "\"")
         endif()
-        cfg_str_add(local_config_string ${configname} "${quote}@${cfg_tag_option}@${quote}")
+        cfg_str_add(local_config_string local_config_json ${configname} "${quote}@${cfg_tag_option}@${quote}")
     endif()
     set(configure_string "${local_config_string}" PARENT_SCOPE)
+    set(configure_json "${local_config_json}" PARENT_SCOPE)
 endfunction(config_string)
 
 # Defines a multi choice / select configuration option
@@ -426,6 +432,7 @@ function(config_choice optionname configname doc)
     # Cannot use ARGN because each argument itself is a list
     math(EXPR limit "${ARGC} - 1")
     set(local_config_string "${configure_string}")
+    set(local_config_json "${configure_json}")
     # force_default represents whether we need to force a new value or not. We would need
     # to force a new value for example if we detect that the current selected choice is
     # no longer (due to conditions) a valid choice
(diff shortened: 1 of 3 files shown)


ahmedcharles:
Note: this has some rough edges since it fails if `configure_json` is set incorrectly, but I think it just needs some cmake polish.

ratmice:
Are the tests failing because this requires a newer minimum cmake version I believe v3.19 added `string(JSON ...`
which is greater than the currently required minimum json version (3.7.2 it looks like from the patch context)?

So I think there needs to be a bump of the minimum version here, and the discussion around doing so,
I haven't looked at the patch in depth, but in regards using json I approve, and think it will help cargo/rust integration.
So I'm totally on board with the idea, one we can settle on an approach/timeframe for version bumping.

ahmedcharles:
Oh, I hadn't considered when JSON was added or what the current minimum is. I suppose that's part of the cmake polish required. :)

tools/helpers.cmake
```diff
@@ -204,30 +204,32 @@ function(GenProofsBFTarget target_name target_file pbf_path pbf_target prunes de
     )
 endfunction(GenProofsBFTarget)
 
-macro(cfg_str_add_disabled cfg_str name)
+macro(cfg_str_add_disabled cfg_str cfg_json name)
     # include in liner files can't use double slash comments
     list(APPEND ${cfg_str} "/* disabled: CONFIG_${name} */")
+    string(JSON ${cfg_json} SET ${${cfg_json}} "${name}" "null")
```
axel-h, line 210: It's really sad that JSON does not allow comments, because I consider this something that should not show up in the "code" to avoid bloating and confusion. Is there any gain adding `null` explicitly when you could also handle this when parsing? My long intention when making this visible as comment in the header was having a way to see whgat potential configuration quirks exists (like setting some ARM values when not targeting ARM at all)  that should be looked into one day.

tools/helpers.cmake
```diff
@@ -204,30 +204,32 @@ function(GenProofsBFTarget target_name target_file pbf_path pbf_target prunes de
...
     endif()
     list(APPEND ${cfg_str} ${cfg_define_str})
+    string(JSON ${cfg_json} SET ${${cfg_json}} ${name} "\"${value}\"")
 endmacro()
 
-macro(cfg_str_add cfg_str name value)
+macro(cfg_str_add cfg_str cfg_json name value)
     if("${value}" STREQUAL "")
-        cfg_str_add_define(${cfg_str} ${name} "" "empty")
+        cfg_str_add_define(${cfg_str} ${cfg_json} ${name} "" "empty")
```
axel-h, line 225: Can't JASON use an empty string here to avoid a potential conflict if `"empty"` is ever used somewhere?

axel-h: Ah now, ignore this comment.

axel-h:
> So I think there needs to be a bump of the minimum version here, and the discussion around doing so [...]

 See e.g. https://github.com/seL4/seL4/issues/827 for a discussion about which version to use.

tools/helpers.cmake
ahmedcharles: When developing it, it was useful to know that all of the data was handled, but it doesn't have to be like this long term, no. Most parsers don't let you determine the difference between something in json being null vs not existing. Given that humans don't need to look at this file (in fact, them looking at it only temps them to change it, which defeats the goal), I think not having comments is fine.

tools/helpers.cmake
ahmedcharles: Perhaps if you're unfamiliar with the cmake, this is confusing. The `"empty"` is passed to the comment parameter, not the value, so it's not semantically relevant and doesn't end up in the json.

ahmedcharles:
It's possible to add this support optionally so that people without 3.19 don't get it, but the build still works.

lsf37:
> It's possible to add this support optionally so that people without 3.19 don't get it, but the build still works.

Since it's fairly simple JSON output, would it be possible to implement the function "natively" as a cmake function for versions < 3.19?

kent-mcleod:
This discussion might be relevant: https://github.com/seL4/seL4/issues/411 

Is this PR a result of discussion on https://sel4.discourse.group/t/pre-rfc-to-effortlessly-develop-sel4-systems-in-rust/539/? 
Was there any discussion in the dev call yesterday?

lsf37:
> This discussion might be relevant: #411

Thanks for the pointer, I was looking for that one.

> Is this PR a result of discussion on https://sel4.discourse.group/t/pre-rfc-to-effortlessly-develop-sel4-systems-in-rust/539/? Was there any discussion in the dev call yesterday?

Yes, plenty. Ben promised to write up a few short notes.

Relevant for this PR: we were looking for a way to make the `cmake` config info from a kernel build available to the lowest-level rust crate (e.g. `libsel4-sys`). With that info, the rest of the Rust build steps to create libs/binaries can be independent. (Doesn't actually have to be a full kernel build, just needs the `camke` step.)

Assembling the Rust build output into a final system image can be a separate step that would either pick up existing `cmake` state from the kernel build system and use existing `cmake` tooling for it (maybe invoked by cargo, maybe invoked via `cmake`), or use whatever else is appropriate for that system.

It turns out this info is useful in multiple ways: @Ivan-Velickovic mentioned just now he had a yaml config export in progress for work on the core platform, and I'm also parsing the cmake output for verification to import config options there. So defining a single format and file for easily accessing that info will be useful in a whole bunch of places. 

Looking at this PR, it seems exporting this part of the config info directly from `cmake` could be achieved fairly easily (if we can write json or yaml from `cmake` without bumping the version, but that sounds doable). @Ivan-Velickovic's approach is slightly different. If I understood him correctly, he's parsing the cached `cmake` output in python and dumping the file from there. If generating the file directly from `cmake` turns out to be too fiddly, we could also do that. 

In either case, one thing that is important (and correctly included in this PR) is that we should include all disabled/unset config options in the output, so downstream tools don't have to know which of these exist.

We probably should include version info in the output, so that tools can later tell which kernel and config version they are building against.

tools/helpers.cmake
lsf37: I think we do need to output unset options for consumers of this file. Otherwise you have to add explicit logic for options that don't exist in the consumer, and it'd be extremely easy to introduce bugs by misspelling option names. I.e. it'd be good if a missing option could be an error. Similarly, an option that shows up but is not expected could be a warning that something is not right.

lsf37:
Sample output is here: https://sel4.discourse.group/t/json-kernel-config-output/549

ahmedcharles:
Is there a cmake variable with the version of the kernel in it?

ratmice:
> Is there a cmake variable with the version of the kernel in it?

I don't see anything, I'm not sure exactly what the right path is, but presumably you could just use something like:

`file(STRINGS ${CMAKE_CURRENT_LIST_DIR}/VERSION SEL4_KERNEL_VERSION)`
or maybe `${CMAKE_CURRENT_SOURCE_DIR}/VERSION` or some such?

axel-h:
> Since it's fairly simple JSON output, would it be possible to implement the function "natively" as a cmake function for versions < 3.19?

I wonder if the most scalable solution would eventually be moving this completely to Python and have pluggable output modules the that could generate arbitrary language bindings. Then CMake is not the limitation at all and a custom workflow based on a kernel release package could just invoke the Python script again with a custom output module. It's a bit overkill for just C and Rust now, though. And it might require shipping a generic config file that can be processed, which is in - JSON? XML? ...?

lsf37:
> > Since it's fairly simple JSON output, would it be possible to implement the function "natively" as a cmake function for versions < 3.19?
> 
> I wonder if the most scalable solution would eventually be moving this completely to Python and have pluggable output modules the that could generate arbitrary language bindings. 

When you say "this" do you mean python parsing the cmake cache and producing config option output? (i.e. @Ivan-Velickovic's approach)

That is certainly an option. I liked this one here, because it's fairly direct, but if the json output is finicky to do before version 3.19, I don't think python would be a problem. It's a build dependency already anyway.

ahmedcharles:
I think there should be one place to track the config. My biggest issue with most of the rust integrations currently is that, for example, sel4.toml and seL4's cmake files don't have to be in sync. I think having a single source of truth is the most important factor. I think it's important that the header used by the kernel is generated the same way as the json file (or any other files with configuration).

lsf37:
> I think there should be one place to track the config. 

Completely agree with that. AFAICT the python approach still satisfies this, i.e. the single source of truth is the cmake input. It just arrives at the output in a different way.

kent-mcleod:
> Relevant for this PR: we were looking for a way to make the `cmake` config info from a kernel build available to the lowest-level rust crate (e.g. `libsel4-sys`). With that info, the rest of the Rust build steps to create libs/binaries can be independent. (Doesn't actually have to be a full kernel build, just needs the `camke` step.)

The set of config options that are currently emitted as C header files is different from the set of config options stored in CMakeCache.txt. So you would need the union of these two sets to be exported if you wanted your external tool to have access to all possible config options.  There are a few approaches for resolving this that seem to be currently concurrently discussed:
- Try and close this gap by ensuring the header files have all "public" config options and then write different generators for different encodings that could be JSON/YAML/make or whatever else.
  - We could potentially go further and stop storing any kernel configuration options in CMakeCache.txt and require that all options are passed in as arguments to the kernel build.  
- Parse the CMakeCache.txt and generate options from there mirroring the execution of the CMake scripts.
- Move the configuration processing out of CMake scripts to a configuration tool and have it generate a single config artifact that the kernel and user level has to use. (Kind of like we used to do when we used KConfig...)

ahmedcharles:
My understanding of the code is that all of the cmake configuration options use the make functions that I modified. Are there options which do not? (I.e. I thought this change was comprehensive, but if it is not, could you provide a link to an option that isn't covered but does impact the kernel/userspace ABI?)

kent-mcleod:
There might not be anything that affects the ABI in a way that you care about actually: 
- [KernelDomainSchedule](https://github.com/seL4/seL4/blob/79eff033aeb4e25ae2783f44cb4f72e5cf04fd57/config.cmake#L290-L299) is almost an example, in that it definitely configures the way that the kernel handles scheduling of threads in different domains.
- KernelDTBPath, KernelCustomDTSOverlay and KernelCustomDTS affect the devices and untyped layout that the kernel provides to userlevel 
- KernelArmHaveGenericTimer doesn't have a C header equivalent, but it may be irrelevant due to the dependent options KernelArmExportPCNTUser, KernelArmExportVCNTUser, KernelArmExportPTMRUser, KernelArmExportVTMRUser which are all in the C headers.

tools/helpers.cmake
Ben-PH: Would it be worthwhile to use a markup that allows for comments instead? I think yaml seemed to be met with interest/approval during the hangouts.

Ben-PH:
> Yes, plenty. Ben promised to write up a few short notes.

https://hackmd.io/@eqFu3cQ2RbyIaZC0igYCLQ/Bkvyq0p2c/edit

tools/helpers.cmake
axel-h: @lsf37 made the remark that it is intended for machine processing in the first place. Given people look at this still to understand what is happening under the hood, it has some benefit to allow comments. Making  debuggigng and prototyping easier is another use case.

ahmedcharles: 1. This is intended for output, not input, therefore, users aren't supposed to change it. Having comments would only encourage people to read it and that would only encourage them to change it. If CBOR were as easy to produce as JSON, I'd just as soon use that so when people opened it, it was just as useful as opening kernel.elf, cause realistically, the need to have comments in both files are equivalent.

lsf37:
The plan is to discuss Nick's proposal, and if we're happy with going with that, to work on this PR to make the JSON export robust. In particular, we probably don't want to increase the minimum cmake version requirement too much yet, so we may need to tweak how the export is implemented.

nspin:
I've opened https://github.com/seL4/seL4/pull/975, which implements something similar to this PR.

lsf37:
Closing this PR as superseded by #975. Thanks @ahmedcharles for starting this discussion!

PR #894, opened 2022-07-27
RPi3/4 updates and fixes

JoonasOnatsu:
This PR adds tested working DTS memory map for RPi4,
and DTS overrides for RPi3/4 so that the timer component
for TImeServer can be created directly from device tree.

Also system timer interrupts are fixed to avoid spurious
interrupts.

Test with: seL4/util_libs#140

### DIFF
--- a/src/plat/bcm2711/overlay-rpi4.dts
+++ b/src/plat/bcm2711/overlay-rpi4.dts
@@ -23,7 +23,7 @@
 	 * https://github.com/raspberrypi/firmware/issues/1374 for
 	 * reference.
 	 *
-	 * RPi4B seems to be in "low peripheral" mode by default.
+	 * RPi4B runs in "low peripheral" mode by default.
 	 * These values were observed from running
 	 * Raspbian instance from /proc/iomem and /sys/firmware/devicetree/base
 	 *
@@ -37,36 +37,18 @@
 	 * 0x01 0x00000000 0x80000000  2048MB
 	 * 0x01 0x80000000 0x80000000  2048MB
 	 *
-	 * Disjunct memory areas seem to work only
-	 * when declared in different memory DT nodes.
-	 *
-	 * If all of the declarations below are in the
-	 * same node, the kernel build system seems
-	 * to only understand the second 3008MB area
-	 * and ignores the rest.
 	 */
 
 	/delete-node/ memory;
 
 	memory@0 {
 		device_type = "memory";
-		reg = < 0x00000000 0x00000000 0x3b400000 >;
-	};
-
-	memory@40000000 {
-		device_type = "memory";
-		reg = < 0x00000000 0x40000000 0xbc000000 >;
-	};
-
-	memory@100000000 {
-		device_type = "memory";
-		reg = < 0x00000001 0x00000000 0x80000000
+		reg = < 0x00000000 0x00000000 0x3b400000
+			0x00000000 0x40000000 0xbc000000
+			0x00000001 0x00000000 0x80000000
 			0x00000001 0x80000000 0x80000000 >;
 	};
 
-	/* TODO: add DMA dedicated ranges
-	 * here also?
-	 */
 	reserved-memory {
 		#address-cells = <0x02>;
 		#size-cells = <0x01>;
@@ -93,7 +75,7 @@
 		 * by declaring it as reserved memory.
 		 */
 		reserved-memory@3b400000 {
-			reg = < 0x00000000 0x3b400000 0x04C00000 >;
+			reg = < 0x00000000 0x3b400000 0x04c00000 >;
 			no-map;
 		};
 	};
--- a/src/plat/bcm2711/overlay-rpi4.dts
+++ b/src/plat/bcm2711/overlay-rpi4.dts
@@ -1,6 +1,7 @@
 /*
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright (C) 2021, Hensoldt Cyber GmbH
+ * Copyright 2021, Technology Innovation Institute
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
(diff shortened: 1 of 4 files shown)


src/plat/bcm2711/overlay-rpi4.dts
```diff
@@ -17,8 +18,83 @@
 		    &{/timer};
 	};
 
+	/*
+	 * See BCM2711 TRM section 1.2 and
+	 * https://github.com/raspberrypi/firmware/issues/1374 for
+	 * reference.
+	 *
+	 * RPi4B seems to be in "low peripheral" mode by default.
```
axel-h, line 26: Good to have this written down finally, just remove the "seems" and say "runs" because it looks like this is a fact.

JoonasOnatsu: Comment fixed to state this as a fact.

src/plat/bcm2711/overlay-rpi4.dts
```diff
@@ -18,35 +18,60 @@
...
-          *
-          * 0x00 0x00000000 0x3c000000  960MB
-          * 0x00 0x40000000 0xbc000000  3008MB
-          * 0x01 0x00000000 0x80000000  2048MB
-          * 0x01 0x80000000 0x80000000  2048MB
-          */
-        reg = < 0x00 0x00000000 0x3c000000 /* 960MB */
-                0x00 0x40000000 0xbc000000 /* 3008MB */
-                0x01 0x00000000 0x80000000 /* 2048MB */
-                0x01 0x80000000 0x80000000 /* 2048MB */ >;
```
kent-mcleod: > Second fix is separating the disjunct memory areas
to separate nodes, as the generator Python scripts
in 'kernel/tools/' etc discarded the first area below
1GB limit if all areas were defined in the same node.
The effect of different variations can be observed
during kernel build in the generated
'kernel/gen_headers/plat/machine/devices_gen.h' header.

I don't think I understand what the error here is? When I generate the header file using this region definition the output I see generated appears to contain all of the regions (using b8ef00e09a1273d9f436dc67db4761cd4711295f):
```
static const p_region_t BOOT_RODATA avail_p_regs[] = {
    /* /memory@0 */
    {
        .start = 0x0,
        .end   = 0x3c000000
    },
    /* /memory@0 */
    {
        .start = 0x40000000,
        .end   = 0xfc000000
    },
    /* /memory@0 */
    {
        .start = 0x100000000,
        .end   = 0x200000000
    },
};
```

JoonasOnatsu: Seems like you are right about that the `avail_p_regs` is generated correctly without having the memory areas in separate nodes. I had the issue occurring at some point in time, from where this observation and workaround was born from, but honestly it's been so long that I don't recall the exact circumstances for it. From what I recall, the last area

 `    /* /memory@0 */
    {
        .start = 0x100000000,
        .end   = 0x200000000
    }`

was being discarded by the build system for some reason.
Obviously, the behavior noted in the comment is no longer true, I will update the file accordingly.

src/plat/bcm2711/overlay-rpi4.dts
```diff
@@ -21,4 +22,25 @@
 		/* This is configurable in the Pi's config.txt, but we use 128MiB of RAM by default. */
 		reg = <0x00 0x00000000 0x08000000>;
 	};
+
+	reserved-memory {
+		#address-cells = <0x01>;
```
kent-mcleod: The upstream definition of this node in tools/dts/rpi4.dts uses a different definition of this property:
```
	reserved-memory {
		#address-cells = <0x02>;
```
It would probably be safer to use the same encoding here to avoid the child nodes from the original file from being misinterpreted?

src/plat/bcm2711/overlay-rpi4.dts
```diff
@@ -21,4 +22,25 @@
 		/* This is configurable in the Pi's config.txt, but we use 128MiB of RAM by default. */
 		reg = <0x00 0x00000000 0x08000000>;
 	};
+
+	reserved-memory {
```
kent-mcleod, line 55: I think hardware_gen.py should eventually be updated to properly handle `/memreserve/` statements.

kent-mcleod: I made an issue to track this: https://github.com/seL4/seL4/issues/899

src/plat/bcm2711/overlay-rpi4.dts
```diff
@@ -18,35 +18,60 @@
...
-                0x01 0x80000000 0x80000000 /* 2048MB */ >;
+	memory@100000000 {
+		device_type = "memory";
+		reg = < 0x00000001 0x00000000 0x80000000
+			0x00000001 0x80000000 0x80000000 >;
 	};
 
+	/* TODO: add DMA dedicated ranges
+	 * here also?
+	 */
```
kent-mcleod, line 54: What are the dedicated DMA ranges this comment refers to? Can this comment be removed?

src/plat/bcm2711/overlay-rpi4.dts
JoonasOnatsu: Thanks for noticing, this is just a typing error. Will fix

src/plat/bcm2711/overlay-rpi4.dts
JoonasOnatsu: This comment seems to be a leftover note for myself which can be removed.
I was thinking if all of the reserved memory definitions should be in the same place
for ease of maintaining, as now the DMA pool definition is in `tools/dts/rpi4.dts`:

```
    reserved-memory {
		#address-cells = <0x02>;
		#size-cells = <0x01>;
		ranges;

		linux,cma {
			compatible = "shared-dma-pool";
			size = <0x4000000>;
			reusable;
			linux,cma-default;
			alloc-ranges = <0x00 0x00 0x40000000>;
		};
	};
```

src/plat/bcm2711/overlay-rpi4.dts
JoonasOnatsu: Just noticed that this definition is updated on cb8c87c to be correct.

src/plat/bcm2711/overlay-rpi4.dts
JoonasOnatsu: Resolved in 979a513.

src/plat/bcm2711/overlay-rpi4.dts
JoonasOnatsu: Fixed in 59be165.

kent-mcleod:
Thanks, this looks ready to merge.  Can you please clean up the history so that the fixup commits are merged with the ones making the changes.

JoonasOnatsu:
> Thanks, this looks ready to merge. Can you please clean up the history so that the fixup commits are merged with the ones making the changes.

Will do. I will also rebase this on the latest master in the process.

axel-h:
Seems the HW builds are failing for RasPi3, could you have a look into this?

JoonasOnatsu:
> Seems the HW builds are failing for RasPi3, could you have a look into this?

Will do, thanks for informing.

JoonasOnatsu:
@axel-h The failing builds for RPi3 should be fixed by https://github.com/seL4/util_libs/pull/140

I also rebased the changes on top of current master. Because the builds are triggered, I suppose
they will fail before that PR gets merged. Could someone manually retrigger the build steps for this
after ```util_libs``` gets fixed?

axel-h:
Thanks. I've added the `Test with: https://github.com/seL4/util_libs/pull/140` line to the description of this PR, so the GitHub CI should pick up the other PR and run the builds with that one. The `hw-build` label is also set, so in a few minutes we should see if things pass now.

lsf37 (approved):
Happy with this. Will merge here and in `util_libs`.

PR #896, opened 2022-07-29
Only access IOAPIC registers that are actually present

wom-bat:
IOAPICS can have varying numbers of lines attached.
The actual number can be accessed in the top 16 bits of the
version register.

Rather than assuming fixed 24 lines per IRQ, read the actual number
and use that.

### DIFF
--- a/src/plat/pc99/machine/ioapic.c
+++ b/src/plat/pc99/machine/ioapic.c
@@ -15,6 +15,7 @@
 #define IOAPIC_WINDOW 0x10
 
 #define IOAPIC_REG_IOAPICID 0x00
+#define IOAPIC_REG_IOAPICVER 0x01
 #define IOAPIC_REG_IOREDTBL 0x10
 
 #define IOREDTBL_LOW(reg) (IOAPIC_REG_IOREDTBL + (reg) * 2)
@@ -32,9 +33,17 @@
 
 #define IOREDTBL_HIGH_RESERVED_BITS 24
 
-/* Cache what we believe is in the low word of the IOREDTBL. This
- * has all the state of trigger modes etc etc */
+/*
+ * Cache what we believe is in the low word of the IOREDTBL. This
+ * has all the state of trigger modes etc etc.
+ * IOAPIC_IRQ_LINES is the upper bound on the number of lines
+ * per IOAPIC.
+ */
 static uint32_t ioredtbl_state[IOAPIC_IRQ_LINES * MAX(1, CONFIG_MAX_NUM_IOAPIC)];
+/*
+ * The number of IRQ lines for each IOAPIC
+ */
+static uint8_t ioapic_nirqs[MAX(1, CONFIG_MAX_NUM_IOAPIC)];
 
 /* Number of IOAPICs in the system */
 static uint32_t num_ioapics = 0;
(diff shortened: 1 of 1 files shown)


src/plat/pc99/machine/ioapic.c
```diff
@@ -118,8 +134,8 @@ exception_t ioapic_decode_map_pin_to_vector(word_t ioapic, word_t pin, word_t le
         current_syscall_error.rangeErrorMax = num_ioapics - 1;
         return EXCEPTION_SYSCALL_ERROR;
     }
-    if (pin >= IOAPIC_IRQ_LINES) {
-        userError("Invalid IOAPIC pin %ld, there are %d pins", (long)pin, IOAPIC_IRQ_LINES);
+    if (pin >= ioapic_nirqs[ioapic]) {
+        userError("Invalid IOAPIC pin %ld, there are %d pins", (long)pin, ioapic_nirqs[ioapic]);
         current_syscall_error.type = seL4_RangeError;
         current_syscall_error.rangeErrorMin = 0;
         current_syscall_error.rangeErrorMax = IOAPIC_IRQ_LINES - 1;
```
kent-mcleod, line 141: Should this rangeErrorMax return value also be updated?

kent-mcleod (approved):
This improvement seems reasonable to me.  Maybe a comment should be added to `ioredtbl_high` to explain that `IOAPIC_IRQ_LINES` is used as an upper bound of how much memory to reserve for book keeping.

src/plat/pc99/machine/ioapic.c
wom-bat: I think you're right.  Will adjust.

src/plat/pc99/machine/ioapic.c
```diff
@@ -52,12 +57,23 @@ static uint32_t ioapic_read(uint32_t ioapic, word_t reg)
...
+
+    ioapic_write(ioapic, IOAPIC_REGSEL, IOAPIC_REG_IOAPICVER);
+    nirqs = (ioapic_read(ioapic, IOAPIC_WINDOW) >> 16) + 1;
+    ioapic_nirqs[ioapic] = nirqs;
+
+    /*
+     * All current implementations have 24 or fewer lines
+     * Protect against the future one that may have more
+     */
+    assert(nirqs <= IOAPIC_IRQ_LINES);
```
lsf37, line 70: `assert` will not be running in release configs, and verification does need to know that all values in that array are bounded. It looks like this is kernel init code, so if there is a condition where the hardware does not meet expectations, it should properly abort initialisation and bail out, not assert.

lsf37 (commented):
This will need a small proof + spec update, but looks doable with reasonable effort. @mbrcknl might remember more about those x64 proofs than I.

src/plat/pc99/machine/ioapic.c
wom-bat: OK, I'll do something different.  I think it's important at this point that we do bail out.

wom-bat: OK I think this is fixed now.

src/plat/pc99/machine/ioapic.c
wom-bat: Done.

wom-bat:
> Maybe a comment should be added to ioredtbl_high to explain that IOAPIC_IRQ_LINES is used as an upper bound of how much memory to reserve for book keeping.


Fixed.

kent-mcleod:
Is this blocked on a verification spec update?

lsf37:
> Is this blocked on a verification spec update?

Spec and proof, yes.

wom-bat:
Problem I'm trying to solve is machine hang when attempting to access non-existent registers.

kent-mcleod:
> Spec and proof, yes.

@lsf37, are proof updates scheduled or in progres?

> Problem I'm trying to solve is machine hang when attempting to access non-existent registers.

@wom-bat, is this urgent? I don't think we can merge it until proofs are ready. If proof updates aren't scheduled and the changes are urgent, we could put it behind a temporary compile-time config? Otherwise the PR is blocked until proofs are done AFAIK.

lsf37:
> > Spec and proof, yes.
> 
> @lsf37, are proof updates scheduled or in progres?

Unless someone has funding for it it'll have to wait until I have spare time (with a bit of luck in the week after next) or someone else picks it up.

src/plat/pc99/machine/ioapic.c
```diff
@@ -118,11 +143,11 @@ exception_t ioapic_decode_map_pin_to_vector(word_t ioapic, word_t pin, word_t le
...
         return EXCEPTION_SYSCALL_ERROR;
     }
-    if (pin >= IOAPIC_IRQ_LINES) {
-        userError("Invalid IOAPIC pin %ld, there are %d pins", (long)pin, IOAPIC_IRQ_LINES);
+    if (pin >= ioapic_nirqs[ioapic]) {
+        userError("Invalid IOAPIC pin %ld, there are %d pins", (long)pin, ioapic_nirqs[ioapic]);
         current_syscall_error.type = seL4_RangeError;
         current_syscall_error.rangeErrorMin = 0;
-        current_syscall_error.rangeErrorMax = IOAPIC_IRQ_LINES - 1;
+        current_syscall_error.rangeErrorMax = ioapic_nirqs[ioapic] + 1;
```
lsf37, line 150: Should this still be `- 1`?

Indanz: Looks like it, yes.

lsf37:
Rebased and squashed in preparation for the proof update.

lsf37 (approved):
Verification is done, this is now ready for merge.

PR #898, opened 2022-07-29
Add newer Skylake model ID

wom-bat:
At least some server-class Skylake processors use 0x55 as
their model ID.
These are Skylake X processors.

### DIFF
--- a/include/arch/x86/arch/machine.h
+++ b/include/arch/x86/arch/machine.h
@@ -55,6 +55,7 @@
  */
 #define SKYLAKE_1_MODEL_ID      0x4E
 #define SKYLAKE_2_MODEL_ID      0x5E
+#define SKYLAKE_X_MODEL_ID      0x55
 #define BROADWELL_1_MODEL_ID    0x4D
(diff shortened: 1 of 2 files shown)


lsf37 (approved):
This is fine to merge from the verification side (no proof update required). I haven't checked the actual number itself, though.

kent-mcleod (approved):

wom-bat:
> This is fine to merge from the verification side (no proof update required). I haven't checked the actual number itself, though.

There's one failing test ... can I ignore that?  it's testing that there is no change to the preprocessed output which of course there will be)

lsf37:
> > This is fine to merge from the verification side (no proof update required). I haven't checked the actual number itself, though.
> 
> There's one failing test ... can I ignore that? it's testing that there is no change to the preprocessed output which of course there will be)

It's checking that the preprocessed output that the theorem provers sees does not change. It does in this case, but the function is in boot code, which is not in the verified set, so in this case it's fine to merge (we can't ignore it completely, because we will need to manually establish sync with the proof again after it's merged).

Issue #875, opened 2022-07-05
`array-bounds` error


tsoutsman:
When compiling the kernel with `-DKernelVerificationBuild:BOOL=OFF`, my builds failing with the following error:
```
[46/47] Building C object CMakeFiles/kernel.elf.dir/kernel_all.c.obj
FAILED: CMakeFiles/kernel.elf.dir/kernel_all.c.obj 
/usr/bin/gcc --sysroot=/home/klim/Downloads/seL4/build  -I/home/klim/Downloads/seL4/include -I/home/klim/Downloads/seL4/include/64 -I/home/klim/Downloads/seL4/include/arch/x86 -I/home/klim/Downloads/seL4/include/arch/x86/arch/64 -I/home/klim/Downloads/seL4/include/plat/pc99 -I/home/klim/Downloads/seL4/include/plat/pc99/plat/64 -I/home/klim/Downloads/seL4/libsel4/include -I/home/klim/Downloads/seL4/libsel4/arch_include/x86 -I/home/klim/Downloads/seL4/libsel4/sel4_arch_include/x86_64 -I/home/klim/Downloads/seL4/libsel4/sel4_plat_include/pc99 -I/home/klim/Downloads/seL4/libsel4/mode_include/64 -I/home/klim/Downloads/seL4/build/gen_config -I/home/klim/Downloads/seL4/build/autoconf -I/home/klim/Downloads/seL4/build/gen_headers -I/home/klim/Downloads/seL4/build/generated -m64  -D__KERNEL_64__ -march=nehalem -O2 -g -DNDEBUG -std=c99 -Wall -Werror -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Wmissing-declarations -Wundef -Wpointer-arith -Wno-nonnull -nostdinc -ffreestanding -fno-stack-protector -fno-asynchronous-unwind-tables -fno-common -O2 -nostdlib -fno-pic -fno-pie -DDEBUG -g -ggdb -mcmodel=kernel -mno-mmx -mno-sse -mno-sse2 -mno-3dnow -MD -MT CMakeFiles/kernel.elf.dir/kernel_all.c.obj -MF CMakeFiles/kernel.elf.dir/kernel_all.c.obj.d -o CMakeFiles/kernel.elf.dir/kernel_all.c.obj -c /home/klim/Downloads/seL4/build/kernel_all.c
[...]

PR #904, opened 2022-08-31
Fix gcc-12 spurious array bounds warning

wom-bat:
Issue #875 shows how gcc-12 is much more aggressive in
array bounds checking.  By lying to gcc and telling it the
page size is zero, the warning can be avoided.


wom-bat:
So the new parameter doesn't exist in earlier GCC.   Need to think about how to add it conditionally.

kent-mcleod (approved):
Thanks for fixing this.  Are you able to add a comment explaining why the diagnostic change is needed.

wom-bat:
> Thanks for fixing this. Are you able to add a comment explaining why the diagnostic change is needed.

Done.

kent-mcleod:
Are you also able to clean up the git history before merging

PR #909, opened 2022-09-14
Fix compilation with CLANG > 11

wom-bat:
LLVM 12 makes -moutline-atomics on by default;
turn it off.


Indanz:
I verified that clang 12.0.0 is the first version supporting `-mno-outline-atomics`.

lsf37 (approved):
Looks good from my side, and doesn't break the current builds, but it'd be good to get another set of eyes from a systems person.

axel-h (approved):
Looks ok to me. Please squash the two commits before merging.

some notes:
- See aec0f5a3f00147815e35290819cb2e008a916402 for change we made for GCC 10
- See Clang release notes at https://releases.llvm.org/12.0.0/tools/clang/docs/ReleaseNotes.html#new-compiler-flags
- this PR is loosely related to https://github.com/seL4/seL4/issues/856

wom-bat:
> @wom-bat: out of couriousity, how did you run into this issues? Was it just general review or did this trigger an actual issue somewhere?

I was trying to reproduce the last HW run failure on my machine, which has tip-of-branch toolchains.  You get link errors without the -mno-outline-atomics because the compiler-rt lib doesn't yet have the right support.

PR #911, opened 2022-09-19
aarch64: Use isb after changing FPU control registers

yyshen:
Signed-off-by: Yanyan Shen <yshen@hybridkernel.com>

### DIFF
--- a/include/arch/arm/arch/64/mode/machine/fpu.h
+++ b/include/arch/arm/arch/64/mode/machine/fpu.h
@@ -88,6 +88,7 @@ static inline void enableTrapFpu(void)
     MRS("cptr_el2", cptr);
     cptr |= (BIT(10) | BIT(31));
     MSR("cptr_el2", cptr);
+    isb();
 }
 
 /* Disable trapping FPU instructions to EL2 */
@@ -97,6 +98,7 @@ static inline void disableTrapFpu(void)
     MRS("cptr_el2", cptr);
     cptr &= ~(BIT(10) | BIT(31));
     MSR("cptr_el2", cptr);
+    isb();
 }
 
 /* Enable FPU access in EL0 and EL1 */
@@ -106,6 +108,7 @@ static inline void enableFpuEL01(void)
     MRS("cpacr_el1", cpacr);
     cpacr |= (3 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
+    isb();
 }
 
 /* Disable FPU access in EL0 */
@@ -116,6 +119,7 @@ static inline void disableFpuEL0(void)
     cpacr &= ~(3 << CPACR_EL1_FPEN);
     cpacr |= (1 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
+    isb();
 }
 
 /* Enable the FPU to be used without faulting.


yyshen:
@lsf37 The preprocess step needs some changes.

kent-mcleod:
I thought that these register system control register updates only affected execution at lower execution levels and so a return-from-exception would already be sufficient to synchronize the updates to these registers?

kent-mcleod:
To elaborate, looking in the ARM Cortex-A programmers guide for ARMv8, there's a section on `isb`s: [13.2.2 ISB in more detail](https://developer.arm.com/documentation/den0024/a/Memory-Ordering/Barriers/ISB-in-more-detail)

It actually uses updating `CPACR_EL1` as an example where an isb is required after, but I assume that this is when EL1 is the current EL.

yyshen:
> I thought that these register system control register updates only affected execution at lower execution levels and so a return-from-exception would already be sufficient to synchronize the updates to these registers?

The EL2 code enables/disables access to FPU before/after it calls saveFpuState/loadFpuState. On some processors, without the ISB, accessing the FPU triggers an kernel exception.

mbrcknl (commented):
The proposed changes should not affect any proofs, since these functions would be below the level of our machine model. In any case, I suspect aarch64 verification has not yet reached C refinement. @lsf37 should be able to confirm.

If you and @kent-mcleod decide that this should be merged, you would need to ignore the preprocess failure. Once this is merged, I can bump the verification manifest.

yyshen:
> The proposed changes should not affect any proofs, since these functions would be below the level of our machine model. In any case, I suspect aarch64 verification has not yet reached C refinement. @lsf37 should be able to confirm.
> 
> If you and @kent-mcleod decide that this should be merged, you would need to ignore the preprocess failure. Once this is merged, I can bump the verification manifest.

Thank you for checking.

lsf37:
Yes, can confirm that this is fine for verification (and will need a preprocess bump)

yyshen:
@kent-mcleod I would like to merge it if you have no objections.

kent-mcleod:
> The EL2 code enables/disables access to FPU before/after it calls saveFpuState/loadFpuState. On some processors, without the ISB, accessing the FPU triggers a kernel exception.

Ok, I can see that it's needed on updates to cptr_el2 that change access from EL2. Why is the ISB needed for the updates to cpacr_el1 also?

yyshen:
> > The EL2 code enables/disables access to FPU before/after it calls saveFpuState/loadFpuState. On some processors, without the ISB, accessing the FPU triggers a kernel exception.
> 
> Ok, I can see that it's needed on updates to cptr_el2 that change access from EL2. Why is the ISB needed for the updates to cpacr_el1 also?

https://github.com/seL4/seL4/blob/d57b2cd0f8a09bf6168f2db39b261651559422a9/include/arch/arm/arch/64/mode/machine/fpu.h#L123

enableFpuEL01 is also called before saving/restoring FPU state without an exception level change.

kent-mcleod:
> enableFpuEL01 is also called before saving/restoring FPU state without an exception level change.

Ok.  disableFpu() doesn't disable the FPU for EL1 when the kernel is operating at EL1.  But there's no explicit function that enables the FPU during kernel boot, so the first time that enableFPU is called could be enabling the FPU state at the same exception level?

kent-mcleod (approved):

yyshen:
> > enableFpuEL01 is also called before saving/restoring FPU state without an exception level change.
> 
> Ok. disableFpu() doesn't disable the FPU for EL1 when the kernel is operating at EL1. But there's no explicit function that enables the FPU during kernel boot, so the first time that enableFPU is called could be enabling the FPU state at the same exception level?

Good point, the I was not inspecting the EL0/1 case carefully enough. It looks like we could have the EL1 access enabled all the time and only disable/enable EL0 access for lazy fpu switch, since we know the kernel itself does not use the FPU.

yyshen:
@kent-mcleod I will merge the current one as it is, and see if I can come up a tuned version for EL1. Thanks

PR #914, opened 2022-09-25
boot: Check that predicted page tables were used

kent-mcleod:
During boot, all dynamic memory allocated by the kernel is precalculated and then allocated in one chunk. The precalculation is supposed to match the allocation logic precisely which means any left over objects indicated an inconsistency between the precalculation and reality. The kernel should print an error and potentially abort if it encounters this situation.

I noticed that there's a page table calculation error in vtd_get_n_paging for calculating page tables required to map iommu reserved regions that over estimates the number of page tables required quite significantly (for example 14337 extra tables on the haswell test machines we use).  This check should detect any other similar calculation errors in the startup paths that we test.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -832,6 +832,12 @@ BOOT_CODE bool_t create_untypeds(cap_t root_cnode_cap)
 
 BOOT_CODE void bi_finalise(void)
 {
+
+    if (rootserver.paging.start != rootserver.paging.end) {
+        printf("WARNING: internal book keeping errror. Less pagetables allocated than predicted: "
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
I like the idea!

lsf37:
And it seems to be catching a bunch of divergences. If they are small enough those would be good to fix in the same PR.

axel-h (approved):
You could merge this with printing a WARNING and drop the assert(), so there is marker in place that this issue is well known. You could also add a comment in the code that explains the background a bit, that would be sufficient in case others run into this and wonder if it's a bug or not.

axel-h:
@kent-mcleod Are you planning to get this merged?

PR #918, opened 2022-09-28
trivial: fix typo

axel-h:
The preprocess/AARCH64 error seem to come from the changes of https://github.com/seL4/seL4/pull/911, where an update is still missing.


lsf37 (approved):
:+1: Will do the preprocess bump, so the failure should disappear on re-run.

PR #919, opened 2022-09-28
make macro less obscure

axel-h:
- explicitly mention the parameter to make macro less obscure.
- add brackets to make the expression an atom.

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -18,7 +18,7 @@
 typedef cte_t  slot_t;
 typedef cte_t *slot_ptr_t;
 #define SLOT_PTR(pptr, pos) (((slot_ptr_t)(pptr)) + (pos))
-#define pptr_of_cap (pptr_t)cap_get_capPtr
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

lsf37:
Interesting that this only seems to be used on AARCH64 according to the preprocess test, but I don't think it will conflict with #801.

lsf37:
The pre-process test checks the AST, so if it's only parentheses and they didn't change the structure of the term, it will not complain even if there is a text diff.

(it failed the first time, because it also still picked up the diff of https://github.com/seL4/seL4/commit/b5b5b6ebddebd2d1885ef5d66770631f8fbd8c02 -- not quite sure where that test log went, I'll have a look)

lsf37:
Can't find the first failed run either, it's probably not kept, because it was attached to revision [d956ac1](https://github.com/seL4/seL4/commit/d956ac12e383c85464fd5d00749add499d9c3de1) which was force-pushed over.

PR #921, opened 2022-09-29
boot: remove unused return values

axel-h:
Remove the return value from `configure_sched_context()`, because it never fails. As a consequence, `create_idle_thread()` also never fails and does not need a return value.


axel-h:
Proofs do not complain, but they don't cover the boot code anyway.

lsf37 (approved):
No problem from the verification side, and I think this one is worth it.

It makes things shorter and clearer, and I don't think we'll run into situations later where we do have cases where creating the idle thread or SC for the idle thread might fail (it's static and we're not interacting with untypeds at this point).

axel-h:
Could this get merged?

kent-mcleod (approved):

axel-h:
rebased to align with changes from merge of https://github.com/seL4/seL4/pull/863

PR #922, opened 2022-09-29
use types time_t and tickts_t properly

axel-h:
- doc: improve comments
- use types time_t and tickts_t properly, currently the semantics are incorrect
- print period and budget with units

The PR in seL4_libs is not necessary, it's just another bit of semantic cleanup.

### DIFF
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -20,6 +20,11 @@
 /* seL4_CapRights_t defined in mode/api/shared_types.bf */
 
 typedef word_t prio_t;
+
+/* The kernel uses ticks_t internally to represent time to make it easy to
+ * interact with hardware timers. The userland API uses time in micro seconds,
+ * which is represented by time_t in the kernel.
+ */
 typedef uint64_t ticks_t;
 typedef uint64_t time_t;
 
--- a/include/arch/arm/arch/32/mode/machine/timer.h
+++ b/include/arch/arm/arch/32/mode/machine/timer.h
@@ -16,7 +16,9 @@
  * CLK_MAGIC and TIMER_CLOCK_MHZ -- these definitions might need to move
  * if we come across an arm platform that does not suit this model */
 
-/* get the max value ticksToUs can be passed without overflowing */
+/* Get the max. ticks_t value that can be expressed in time_t (time in us). This
+ * is the max. value ticksToUs() can be passed without overflowing.
+ */
 static inline CONST ticks_t getMaxTicksToUs(void)
 {
 #if USE_KHZ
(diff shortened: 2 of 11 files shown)


include/api/types.h
```diff
@@ -20,6 +20,12 @@
 /* seL4_CapRights_t defined in mode/api/shared_types.bf */
 
 typedef word_t prio_t;
+
+/* The userland API uses ticks_t, which is micro seconds, while the kernel uses
```
pingerino, line 24: this comment is a bit confusing - I had to read it multiple times to understand it - and I wrote the api.

Did you consider something simpler eg:

```
ticks_t: time in ticks at the frequency of the hardware timer
time_t: time in microseconds
```

if you really think it helps, you could add

```
User level api functions all use `time_t`.  
```

include/arch/arm/arch/32/mode/machine/timer.h
```diff
@@ -16,7 +16,7 @@
  * CLK_MAGIC and TIMER_CLOCK_MHZ -- these definitions might need to move
  * if we come across an arm platform that does not suit this model */
 
-/* get the max value ticksToUs can be passed without overflowing */
+/* Get the max value ticksToUs() can handle without overflowing. */
```
pingerino, line 19: this change is a bit pedantic - what's wrong with 'be passed`? and I don't want to encourage rewriting all of the comments to use correct capitalisation - too hard to enforce, and way too little benefit to bother with.

pingerino (changes requested):

include/api/types.h
axel-h: I have reworded it, what do you think now?

axel-h:
@pingerino Ist this ok now? My main goal is to get the type fixes in, as this is semantically broken. The comment changes are just a nice to have thing.

include/api/types.h
```diff
@@ -20,6 +20,11 @@
 /* seL4_CapRights_t defined in mode/api/shared_types.bf */
 
 typedef word_t prio_t;
+
+/* The kernel uses ticks_t internally to represent the time, this is driven by
+ * the frequency of the hardware timer. The userland API uses time in micro
+ * seconds, which is represented by time_t in the kernel.
```
lsf37, line 26:
```suggestion
/* The kernel uses ticks_t internally to represent time to make it easy to interact with hardware timers.
 * The userland API uses time in micro seconds, which is represented by time_t in the kernel.
```

axel-h: changed comment

kent-mcleod (approved):

PR #931, opened 2022-10-14
fastpatch: use thread passed as parameter

axel-h:
Use the thread passed as parameter instead of making assumption that this is in sync with the global state.

Using `NODE_STATE(ksCurThread)` feels semantically wrong when `cur_thread` is passed to the function. It seems this function is not supposed to ever be called with `cur_thread` being different to `NODE_STATE(ksCurThread)`, so I wonder if an explicit assert in the code should be added then to state this (and catch potential copy/paste porting errors).


lsf37 (approved):
I agree, this change also brings these in line with the rest of the fast path code.

axel-h:
Could this get merged?

kent-mcleod (approved):

PR #932, opened 2022-10-16
user proper field from thread context structure

axel-h:
The thread context structure layout is explicitly designed to have the register context first, as this simplifies saving the user context on entry in assembly code. On the exit path there is no need to hard-code this assumption, stick to the semantics and leave details to the compiler.

### DIFF
--- a/include/arch/arm/arch/32/mode/fastpath/fastpath.h
+++ b/include/arch/arm/arch/32/mode/fastpath/fastpath.h
@@ -128,7 +128,7 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
 
     register word_t badge_reg asm("r0") = badge;
     register word_t msgInfo_reg asm("r1") = msgInfo;
-    register word_t cur_thread_reg asm("r2") = (word_t)cur_thread;
+    register word_t cur_thread_reg asm("r2") = (word_t)cur_thread->tcbArch.tcbContext.registers;
 
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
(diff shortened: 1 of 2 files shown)


include/arch/arm/arch/32/mode/fastpath/fastpath.h
```diff
@@ -128,7 +128,7 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
 
     register word_t badge_reg asm("r0") = badge;
     register word_t msgInfo_reg asm("r1") = msgInfo;
-    register word_t cur_thread_reg asm("r2") = (word_t)cur_thread;
+    register word_t cur_thread_reg asm("r2") = (word_t)cur_thread->tcbArch.tcbContext.registers;
```
kent-mcleod, line 131: @lsf37: How come this commit doesn't trigger a preprocessor fail for armv7?

lsf37: Both functions have a `/** DONT_TRANSLATE */` tag, so the AST nodes in the parser will have no content.

lsf37 (approved):
I'm fine with this change, it's the same semantics and a bit safer should the layout ever change.

lsf37:
It would really be nice to have that binary comparison now to check that this really is no change in the binary. I might have a look at that manually.

lsf37:
So the binaries do differ, which is interesting. I'd have thought that we'd get the same after optimisations.

lsf37:
But the disassembly appears to be the same. The only difference is a symbol name  `CSWTCH.1200` vs  `CSWTCH.1199` which appears in the string table.

My faith in optimising compilers is restored..

PR #933, opened 2022-10-16
risc-v: remove K constraint in fcsr access wrapper

axel-h:
The 'K' constraint is for 5-bit unsigned integer immediate operands only in both GCC and LLVM. It does not apply to the value written.

See manuals:
- https://llvm.org/docs/LangRef.html#supported-constraint-code-list
-  https://gcc.gnu.org/onlinedocs/gcc/Machine-Constraints.html

In the Linux kernel there is this the commit https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/riscv/include/asm/csr.h?id=5d8544e2d0075a5f3c9a2cf27152354d54360da1 that adds the 'K' constraint, which might explain the origin. Even if this has never been change, I still think this is incorrect usage.

### DIFF
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -141,12 +141,12 @@ static inline void clearMemory(void *ptr, unsigned int bits)
 
 static inline void write_satp(word_t value)
 {
-    asm volatile("csrw satp, %0" :: "rK"(value));
+    asm volatile("csrw satp, %0" :: "r"(value));
 }
 
 static inline void write_stvec(word_t value)
 {
-    asm volatile("csrw stvec, %0" :: "rK"(value));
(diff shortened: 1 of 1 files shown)


kent-mcleod:
> Even if this has never been change, I still think this is incorrect usage.

The K allows the compiler optimizer to do constant propagation and turn the csrw into an immediate.  The following assembly is generated from the C code `write_stvec(0);` with compiler optimizations enabled. The first assembly is with `rK` and the second is just with `r`.

```
520:   10505073                csrwi   stvec,0
```

```
 520:   4781                    li      a5,0
 522:   10579073                csrw    stvec,a5
```

axel-h:
Good point, compiler is smarter than expected. Closing this then.

PR #935, opened 2022-10-19
gcc: add x86 cross-compiler for arm host

chrisguikema:
When compiling in the Docker container on an M1 MacBook, or any other ARM host, the CMAKE_HOST_APPLE variable is not set. Therefore, no CROSS_COMPILER_PREFIX is set and x86 applications will not compile.

This commit adds another check to see if the host processor is an ARM chipset, and sets the appropriate CROSS_COMPILER_PREFIX variable.

M1 docker issues described here: https://github.com/seL4/seL4-CAmkES-L4v-dockerfiles/issues/40

### DIFF
--- a/gcc.cmake
+++ b/gcc.cmake
@@ -109,6 +109,9 @@ if("${CROSS_COMPILER_PREFIX}" STREQUAL "")
         if(CMAKE_HOST_APPLE)
             # CMAKE_HOST_APPLE is a CMake variable that evaluates to True on a Mac OSX system
             FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-" "x86_64-unknown-linux-gnu-")
+        elseif("${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
+            # If we're on an aarch64 host, gcc won't support x86. Set an x86 compatible toolchain
+            FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-" "x86_64-unknown-linux-gnu-")
         endif()
     endif()
 endif()


gcc.cmake
```diff
@@ -109,6 +109,9 @@ if("${CROSS_COMPILER_PREFIX}" STREQUAL "")
         if(CMAKE_HOST_APPLE)
             # CMAKE_HOST_APPLE is a CMake variable that evaluates to True on a Mac OSX system
             FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-" "x86_64-unknown-linux-gnu-")
+        elseif("${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
+            # If we're on an aarch64 host, gcc won't support x86. Set an x86 compatible toolchain
+            set(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-")
```
axel-h, line 114: why not merge the checks and avoid some redundancy
```suggestion
        # If we haven't set a target above we assume x86_64/ia32 target, even
        # if running on non-x86 architectures, e.g. Apple M1. CMAKE_HOST_APPLE
        # is set when running natively on Mac OS, but it's missing when running
        # inside a docker container. So check for the host architecture also.
        if(CMAKE_HOST_APPLE OR "${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "aarch64")
            FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-" "x86_64-unknown-linux-gnu-")
```
I assume we always want to default to x86 and not the host's native architecture, as forcing users to set `CROSS_COMPILER_PREFIX` costs too much convenience or break too much legacy scripts? Could we take the target platform into account here and make a guess based on this? Do we know `PLATFORM` here already?

lsf37: I'm not sure we should merge these, they are two different cases. They happen to lead to the same outcome right now, but that might change in the future. One case is that we're compiling on a Mac, the other case is that we're compiling on an AArch64 machine.

lsf37: > I assume we always want to default to x86 and not the host's native architecture, as forcing users to set CROSS_COMPILER_PREFIX costs too much convenience or break too much legacy scripts? Could we take the target platform into account here and make a guess based on this? Do we know PLATFORM here already?

The default, when we reach this place, is to build for `x86` -- the other architectures are checked and set before and all set the tool chain explicitly. I don't think it would be a good idea to make the default target dependent on the host, that sounds like a recipe for confusion.

lsf37 (approved):
I'm happy with this as is, but will leave some time for others to respond.

gcc.cmake
axel-h: Ok, in this case I thing the comment should explicitly say that. And also, that this is the special case on Mac OS where we run in a docker container on ARM silicon. The commit comment already states this, but I think it would be good to also says this in the code comment explicitly.

What about `set(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-")` vs. `FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-")`? Seem the 2nd one would be better to catch problems early in case the docker container contents change.

lsf37: But that's the point, it's not just a special case for docker on M1. It's any AArch64 host that will need this, so a comment about M1 would be misleading. Currently docker on M1 is the most likely to reproduce the problem, because it's common, but if you want to build on a normal Linux Aarch64 machine you'll need the same change. It really is exactly what the condition says: if the host is an AArch64 machine we need a cross compiler for x86. That sounds perfectly clear to me.

I do agree that `FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-"`) or even FindPrefixedGCC(CROSS_COMPILER_PREFIX "x86_64-linux-gnu-" "x86_64-unknown-linux-gnu-") would be slightly better, esp in the general Linux case.

axel-h: So, is the check for `CMAKE_HOST_APPLE` still needed then? Seem regardless of docker or not, the host architecture will be "aarch64". And on an intel Mac there was no issue. 

Also, (nitpicking a bit), seems the check could actually be something like "NOT "${CMAKE_HOST_SYSTEM_PROCESSOR}" STREQUAL "x86_64"`, as we could also be on e.g. RISC-V silicon. However, given running on M1 silicon is the only practical case for the near future and given that `CMAKE_HOST_SYSTEM_PROCESSOR` can return some other values even on x86, checking for "aarch64" and adding other architectures case by case might have less side effects.

chrisguikema: So the codebase I was using was a bit outdated and wasn't using the `FindPrefixedGCC` function. I can update to use that easily. 

I agree with @axel-h, instead of checking for `APPLE`, checking against the host processor architecture does the same thing with one fewer check. 

Also, is compiling seL4 without the Docker containers even feasible at this point? It was the first thing I tried on my M1 MacBook, and it had a host of small issues that made the Docker container the easier route. And when you're using an ARM based Docker container, `APPLE` is not set.

lsf37: The check for `CMAKE_HOST_APPLE` is definitely still necessary for Intel-based Macs, I don't think anything has changed there or is likely to change in the future. That check is a host OS check, whereas the new check is a host architecture check. Both make sense separately to me and might need different settings in the future.

Compiling without Docker is not only feasible but I would think pretty widespread. I only use docker where really necessary, including on the M1 (although I don't think I have done the full setup there, just what I need for verification, which is compiling seL4, but not CAmkES etc). Docker is good for reproducibility and getting started quickly, and I'm not arguing against it in any way, but it's definitely not the only environment.

lsf37:
Ok, thanks for the update with `FindPrefixedGCC`. From my side we're ready to merge!

axel-h (approved):

kent-mcleod (approved):

PR #936, opened 2022-10-20
aarch64, fpu: update hardware fpu support check

chrisguikema:
Cortex-A55 processors have additional support for half-precision floating point arithmetic, which is defined as 0b0001 in id_aa64pfr0_el1. This value includes all support for single and double precision floating point types. This commit adds a macro checking the register for 0b1111, which is universally defined as "Floating-point is not implemented". All other values are reserved.


lsf37:
The preprocess failure is fine from the verification side.

Just to double-check, though, since this is in non-verified code: is the rest of the kernel code fine with a half-precision FPU? Skimming briefly I can see treatment for 16 or 32 registers, so I'm guessing yes, but someone with more knowledge of this code should probably have a look at that.

@yyshen do you know more about this?

chrisguikema:
> Just to double-check, though, since this is in non-verified code: is the rest of the kernel code fine with a half-precision FPU?

From what I remember, we successfully ran sel4test on an SDM670 Qualcomm platform, and I think that includes FPU tests. But that was a _long_ time ago. I'll let people with more expertise chime in.

src/arch/arm/64/machine/fpu.c
```diff
@@ -30,10 +33,10 @@ BOOT_CODE bool_t fpsimd_HWCapTest(void)
 
     /* Check if the hardware has FP and ASIMD support... */
     MRS("id_aa64pfr0_el1", id_aa64pfr0);
-    if (((id_aa64pfr0 >> ID_AA64PFR0_EL1_FP) & MASK(4)) == MASK(4) ||
-        ((id_aa64pfr0 >> ID_AA64PFR0_EL1_ASIMD) & MASK(4)) == MASK(4)) {
-        return false;
+    if (IS_FPU_ENABLED(id_aa64pfr0, ID_AA64PFR0_EL1_FP) ||
+        IS_FPU_ENABLED(id_aa64pfr0, ID_AA64PFR0_EL1_ASIMD)) {
+        return true;
```
Indanz, line 38: The logic is changed from returning false when either FP or ASIMD is not supported to only returning false when both are unsupported.

If intentional, perhaps change the commit message to make this explicit and explain why it's okay? The current message does not match the change made, as the old code already checked for 0xF and there is no explicit check for 0x1 added.

The original check seems to assume that either both or none are present, and that assumption might be propagated elsewhere. What will happen now if FP is supported, but ASIMD isn't?

src/arch/arm/64/machine/fpu.c
```diff
@@ -8,6 +8,9 @@
 #include <arch/machine/fpu.h>
 #include <mode/model/statedata.h>
 
+#define FPU_NOT_IMPLEMENTED (0xF)
+#define IS_FPU_ENABLED(id_aa64pfr0, shift) (((id_aa64pfr0 >> shift) & MASK(4)) != FPU_NOT_IMPLEMENTED)
```
axel-h, line 12: macro parameters must be put in brackets to make them atoms.
```suggestion
#define IS_FPU_ENABLED(id_aa64pfr0, shift) ((((id_aa64pfr0) >> (shift)) & MASK(4)) != FPU_NOT_IMPLEMENTED)
```

kent-mcleod:
> From what I remember, we successfully ran sel4test on an SDM670 Qualcomm platform, and I think that includes FPU tests. But that was a _long_ time ago. I'll let people with more expertise chime in.

The kernel has support for the odroid-c4 which has an Cortex-A55, so we can use that platform for validating these changes at least..

yyshen:
> The preprocess failure is fine from the verification side.
> 
> Just to double-check, though, since this is in non-verified code: is the rest of the kernel code fine with a half-precision FPU? Skimming briefly I can see treatment for 16 or 32 registers, so I'm guessing yes, but someone with more knowledge of this code should probably have a look at that.
> 
> @yyshen do you know more about this?

The kernel saves/restores FPU registers by their max length, and it does not care about how the user code uses the registers. So I think the kernel part should be fine.

src/arch/arm/64/machine/fpu.c
yyshen: I think the new commit makes sense since as I recall, VFP and ASIMD registers are mapped to the same underlying hardware registers. So we need to switch FPU registers when either of them (most of the time, both) is implemented.

yyshen (approved):
axel-h made a good suggestion.

chrisguikema:
Alright, I've addressed Axel's comment

yyshen:
@kent-mcleod @axel-h I think we can merge it. any other improvements?

lsf37 (approved):
fine from the proof side

kent-mcleod:
> @kent-mcleod @axel-h I think we can merge it. any other improvements?

I still don't follow how the proposed change is necessary for Cortex-A55 support. 
> This commit adds a macro checking the register for 0b1111, which is universally defined as "Floating-point is not implemented". All other values are reserved.

This is how the current code already checks the register: `(id_aa64pfr0 >> ID_AA64PFR0_EL1_FP) & MASK(4)) == MASK(4)`.

The OdroidC4 is a Cortex-A55 and the FPU tests pass for it: https://github.com/seL4/seL4/actions/runs/5389315496/job/14589928227#step:4:1413

The logic that's changing is as [Indan's comment](https://github.com/seL4/seL4/pull/936#discussion_r1001612778) describes: 
> The logic is changed from returning false when either FP or ASIMD is not supported to only returning false when both are unsupported.

It seems reasonable that the FPU support should be enabled if either the AdvSIMD or FP fields in ID_AA64PFR0_EL1 are != 0b1111 instead of requiring that both are not equal.  But if this is the case, then the commit message should be updated to reflect that is what's being changed (which is unrelated to Cortex-A55 support).

lsf37:
It seems to me that the only bit missing here is updating the commit message. @chrisguikema are you happy to do that?

chrisguikema:
Yeah no problem.

src/arch/arm/64/machine/fpu.c
```diff
@@ -8,6 +8,9 @@
 #include <arch/machine/fpu.h>
 #include <mode/model/statedata.h>
 
+#define FPU_NOT_IMPLEMENTED (0xF)
+#define IS_FPU_ENABLED(id_aa64pfr0, shift) ((((id_aa64pfr0) >> (shift)) & MASK(4)) != FPU_NOT_IMPLEMENTED)
```
axel-h, line 12: It is a bit confusing if we use `MASK(4)` and `0xf` and `FPU_NOT_IMPLEMENTED`. Maybe simply say
```suggestion
#define IS_FPU_ENABLED(id_aa64pfr0, shift) ((((id_aa64pfr0) >> (shift)) & MASK(4)) != MASK(4))
```

axel-h: Or, looking at the ARM arch manual, make this really generic:

#define GET_ARM_CORE_FEATURE(id_aa64pfr, shift) ((id_aa64pfr) >> (shift)) & MASK(4))

#define GET_ARM_CORE_FEATURE_FPU(id_aa64pfr0) GET_ARM_CORE_FEATURE(id_aa64pfr0, ID_AA64PFR0_EL1_FP)
#define GET_ARM_CORE_FEATURE_ADVSIMD(id_aa64pfr0) GET_ARM_CORE_FEATURE(id_aa64pfr0, ID_AA64PFR0_EL1_ASIMD)

/* 0b1111 means "not implemented". All other values indicate the implementation type */
#define IS_FPU_ENABLED(id_aa64pfr0) (GET_ARM_CORE_FEATURE_FPU(id_aa64pfr0) != MASK(4))
#define IS_ADVSMDI_ENABLED(id_aa64pfr0) (GET_ARM_CORE_FEATURE_ADVSIMD(id_aa64pfr0) != MASK(4))
...
 if (IF_FPU_ENABLED(id_aa64pfr0) ||
        IS_ADVSMDI_ENABLED(id_aa64pfr0) {
        return true;
```

kent-mcleod:
Going off of https://developer.arm.com/documentation/ddi0487/maa/-Part-D-The-AArch64-System-Level-Architecture/-Chapter-D24-AArch64-System-Register-Descriptions/-D24-1-About-the-AArch64-System-registers/-D24-1-3-Principles-of-the-ID-scheme-for-fields-in-ID-registers, the way that the kernel currently handles the feature detection for FPU on aarch64 is sufficient.

We don't support any microarchitectures yet that don't support FPU and SIMD and don't have any plans to currently considering that the architecture points out that support is almost mandatory as it is a mandatory requirement of the ARM Procedure Call Standard for AArch64.

In the event that there are CPUs that have either value set to 0xffff, we don't really have a picture for what the code should do here. Really the calling function should probably fail to boot if this feature test function returns false.

PR #937, opened 2022-10-20
zynqmp: remove hypervisor overlay

chrisguikema:
This was only required since the PCI bus was generating devices that were too large when running in hypervisor mode. That behavior has been fixed, so the overlay is not needed.


axel-h:
For reference, do you know the commit/PR that fixed this?

chrisguikema:
https://github.com/seL4/seL4/blob/master/tools/hardware/device.py#L195

This is the check that prevents the incorrect untyped objects. Looking at the history of that file, it was basically rewritten from the old hardware_gen.py file. Its first iteration had the PCI logic:

https://github.com/seL4/seL4/commit/75f2c54b42c2dbea56ca91ae6b029c3d3ad8eed4#diff-59d56b684bc443878f2895c0dccdd3cffe64793e0a8f9680d1edf47cb59aedb2R178

Browsing the history at `6e3606d056a3e24dcfc48ae0f1b7644e9010f6d0` (the commit before the rewrite) shows different logic for PCI:

https://github.com/seL4/seL4/blob/6e3606d056a3e24dcfc48ae0f1b7644e9010f6d0/tools/hardware_gen.py#L464

axel-h (approved):
Give you added this overlay file in https://github.com/seL4/seL4/commit/c4f947994f7a1143480e20cda048d36fd22dada0, I assume you are the expert on this topic anyway. So, looks fine to me to remove it again then.

PR #944, opened 2022-10-28
Polarfire fixes

Apave24:
A few small commits fixing the timer node in the device tree and the TIMER_FREQUENCY in the config for the polarfire platform.

### DIFF
--- a/tools/dts/mpfs_icicle.dts
+++ b/tools/dts/mpfs_icicle.dts
@@ -465,7 +465,8 @@
 
 	timer0: timer@20125000 {
 		compatible = "timer";
-                reg = <0x0 20125000 0x0 0x400>;
+                status = "okay";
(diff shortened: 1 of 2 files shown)


tools/dts/mpfs_icicle.dts
```diff
@@ -463,9 +463,10 @@
 		};
 	};
 
-	timer0: timer@20125000 {
```
axel-h, line 466: Do we really have to drop the "timer0" name here?

axel-h (approved):
Looks ok to me. Please address the comment I made, then it's good to merge.

tools/dts/mpfs_icicle.dts
Apave24: Nope. Undone.

kent-mcleod (approved):

PR #945, opened 2022-11-02
MCS: Remove grant right from reply cap

mbrcknl:
This PR adds to #874, and represents the second alternative discussed in [RFC-13](https://sel4.atlassian.net/browse/RFC-13):
- #874 restricts the ability to grant via a reply according to the grant right on the receive endpoint cap, but does not remove the grant right from reply caps.
- This PR additionally removes the grant right from reply caps.

This PR is relative to #874. If you want to see this alternative in full, use a [branch-based comparison against master](https://github.com/seL4/seL4/compare/master...mbrcknl/mcs-grant-via-reply-alt?expand=1).

------

This commit removes the grant right from reply caps on MCS. This is the
second alternative discussed in seL4 RFC-13:

https://sel4.atlassian.net/browse/RFC-13

Grant rights on reply object caps in MCS have limited use, because any
server with sufficient untyped memory can create new reply objects with
full grant rights.

A previous change restricted a server's ability to grant to its clients
by requiring that the server has grant rights on both the reply object
cap and the endpoint cap it used to receive a call. This change removes
grant rights from reply object caps, so that only the grant rights on
the receive endpoint matter.

This second change is motivated by the observation that there is no
clear use for a grant right on reply object caps once the first change
is implemented.

We did consider a hypothetical use case for this grant right. Suppose
there is a server T that is trusted to grant to its clients, but which
may delegate some calls to another server U that is *not* trusted to
grant to its clients. One way to perform the delegation would be for T
to pass the reply object capability to U, so that U can reply directly.
T would need grant rights on its receive endpoint, but would need to be
able to diminish grant rights on any reply object capability before
passing it to U.

However, we concluded that the protocol that would be needed to manage
reply objects between T and U would be more complex and less efficient
than simply proxying the reply via T.

Additionally, it's not even possible to construct such a system in a
non-MCS configuration, because the non-MCS kernel does not provide an
API to diminish the rights on a reply capability, other than using the
grant right on the receiver's endpoint capability.

[RFC-13]: https://sel4.atlassian.net/browse/RFC-13?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kevinelp:
It's tempting to say this is an unrealistic use case. Instead, I'll propose a different mechanism to achieve what you describe. Reply objects caps could have a right reduction mask that could be used to downgrade rights contained in the reply object upon invocation. The trusted server T has a non-downgrading reply object cap to the reply object, the untrusted server U has a grant masking reply object cap to the reply object. This delegation can be achieved via downgrading of existing rights instead of upgrading potentially non-existent rights.

mbrcknl:
> It's tempting to say this is an unrealistic use case. Instead, I'll propose a different mechanism to achieve what you describe. Reply objects caps could have a right reduction mask that could be used to downgrade rights contained in the reply object upon invocation. The trusted server T has a non-downgrading reply object cap to the reply object, the untrusted server U has a grant masking reply object cap to the reply object. This delegation can be achieved via downgrading of existing rights instead of upgrading potentially non-existent rights.

Correct me if I'm wrong @kevinelp, but I think you're describing exactly what's in #874. In #874, the grant rights in a reply object are set by the kernel every time the receiver calls `seL4_Recv` or `seL4_ReplyRecv`, to be the same as the grant bit on the receiver's endpoint cap. The grant bit on the reply cap is set by whoever retyped the reply object, or whoever since did a CNode operation to mask the rights. So the two grant bits are set independently, and you need both to be true to actually grant via a reply.

In any case, I think I'm coming to the view that the use case is contrived/unrealistic. The idea of the scenario is that T is some kind of trusted proxy/multiplexer, which receives calls and delegates them to other components, depending on some partial decoding of the call. However, if U is not trusted to grant via reply, then it is almost certainly not trusted to reply to every request that T receives, so T will only want to give U a reply cap after T has decoded enough of the call to determine that it is a call that U should reply to. So T would need to call U, granting U the reply cap for the call that T received. Confused yet? To make matters worse, T can only reuse the reply object after U performs the reply on the delegated call, but how does T know when U is done with the reply object? It has to wait for U to reply to the call that T made to U to delegate the original reply cap! In other words, U has to perform two reply operations, firstly on the reply that T delegated to it, and secondly on the call that T made to U for the delegation. And then since T does not fully trust U, T has to revoke the reply it delegated to U before it reuses it for the next receive. At this point, it's so complicated that T might as well just proxy the reply phase of the original call as well as the receive phase. That way, U never gets access to the reply object that T is using, and there are fewer operations involved.

kevinelp:
> > It's tempting to say this is an unrealistic use case. Instead, I'll propose a different mechanism to achieve what you describe. Reply objects caps could have a right reduction mask that could be used to downgrade rights contained in the reply object upon invocation. The trusted server T has a non-downgrading reply object cap to the reply object, the untrusted server U has a grant masking reply object cap to the reply object. This delegation can be achieved via downgrading of existing rights instead of upgrading potentially non-existent rights.
> 
> Correct me if I'm wrong @kevinelp, but I think you're describing exactly what's in #874. In #874, the grant rights in a reply object are set by the kernel every time the receiver calls `seL4_Recv` or `seL4_ReplyRecv`, to be the same as the grant bit on the receiver's endpoint cap. The grant bit on the reply cap is set by whoever retyped the reply object, or whoever since did a CNode operation to mask the rights. So the two grant bits are set independently, and you need both to be true to actually grant via a reply.
> 

I'm going to be a pedant, you mean bits in the reply object cap :-)

Not exactly, I'm saying that if reply object caps do anything with rights they should only diminish reply rights - it's effectively a mask and while we are fiddling, it should be applicable to any rights associated with replying. It may be grant is the only relevant one - I'd need to trawl code.  

> In any case, I think I'm coming to the view that the use case is contrived/unrealistic. The idea of the scenario is that T is some kind of trusted proxy/multiplexer, which receives calls and delegates them to other components (including U), depending on some partial decoding of the call. T gives U the reply cap, so U can reply directly to the original call received by T.
> 
> However, if U is not trusted to grant via reply, then it is almost certainly not trusted to reply to every request that T receives, so T will only want to give U a reply cap after T has decoded enough of the call to determine that it is a call that U should reply to. So T would need to call U, granting U the reply cap for the call that T received. Confused yet?

No, I'm following, what you propose is something like minting a new reply object cap with less rights and for U's use, or provide an API to move a reply right to a different reply object accessible to U with less reply object cap rights. 

> 
> To make matters worse, T can only reuse the reply object after U performs the reply on the delegated call, but how does T know when U is done with the reply object? It has to wait for U to reply to the call that T made to U to delegate the original reply cap! In other words, U has to perform two reply operations, firstly on the reply that T delegated to it, and secondly on the call that T made to U for the delegation. And then since T does not fully trust U, T has to revoke the reply it delegated to U before it reuses it for the next receive.
> 
> At this point, it's so complicated that T might as well just proxy the reply phase of the original call as well as the receive phase. That way, U never gets access to the reply object that T is using, and there are fewer operations involved.

Yep, there are two separate issues here - what can you express in a model (i.e. potentially unrealistic examples), and whether we can express some use case we are building or envisage building.  Not convinced this example is in the second class :-)

mbrcknl:
> I'm going to be a pedant, you mean bits in the reply object cap :-)

Haha, yes, I read your comments in the wrong order, so I'm now on board with your terminology. :-)

> Not exactly, I'm saying that if reply object caps do anything with rights they should only diminish reply rights - it's effectively a mask and while we are fiddling, it should be applicable to any rights associated with replying. It may be grant is the only relevant one - I'd need to trawl code.

Right, I think we're basically agreeing. Some clarifications of what's in the PRs, just to be sure:
- In #874, the grant bit on the reply object capability indeed only acts as a mask against the grant bit in the receive endpoint capability. So to grant via reply, the receiver needs both bits set.
- In this PR, the grant bit on the reply object capability has been removed, so the ability to grant via reply can only be controlled by the grant bit on the receive endpoint capability.
- I think grant is the only right relevant to replying. There's nothing to read, and without write, replying makes little sense.

> > In any case, I think I'm coming to the view that the use case is contrived/unrealistic. The idea of the scenario is that T is some kind of trusted proxy/multiplexer, which receives calls and delegates them to other components (including U), depending on some partial decoding of the call. T gives U the reply cap, so U can reply directly to the original call received by T.
> > However, if U is not trusted to grant via reply, then it is almost certainly not trusted to reply to every request that T receives, so T will only want to give U a reply cap after T has decoded enough of the call to determine that it is a call that U should reply to. So T would need to call U, granting U the reply cap for the call that T received. Confused yet?
> 
> No, I'm following, what you propose is something like minting a new reply object cap with less rights and for U's use, or provide an API to move a reply right to a different reply object accessible to U with less reply object cap rights.

:+1:

> > To make matters worse, T can only reuse the reply object after U performs the reply on the delegated call, but how does T know when U is done with the reply object? It has to wait for U to reply to the call that T made to U to delegate the original reply cap! In other words, U has to perform two reply operations, firstly on the reply that T delegated to it, and secondly on the call that T made to U for the delegation. And then since T does not fully trust U, T has to revoke the reply it delegated to U before it reuses it for the next receive.
> > At this point, it's so complicated that T might as well just proxy the reply phase of the original call as well as the receive phase. That way, U never gets access to the reply object that T is using, and there are fewer operations involved.
> 
> Yep, there are two separate issues here - what can you express in a model (i.e. potentially unrealistic examples), and whether we can express some use case we are building or envisage building. Not convinced this example is in the second class :-)

Agreed, I have talked myself out of caring about this case. :-)

mbrcknl:
While discussing this at the recent developer hangout, @lsf37 asked whether it was possible to remove grant rights from a non-MCS reply capability. I said, yes, of course it's possible, because there's this case in `maskCapRights`:

```
    case cap_reply_cap: {
        cap_t new_cap;

        new_cap = cap_reply_cap_set_capReplyCanGrant(
                      cap, cap_reply_cap_get_capReplyCanGrant(cap) &
                      seL4_CapRights_get_capAllowGrant(cap_rights));
        return new_cap;
    }
```

But it turns out that code is useless, because there's also this case in `deriveCap`:

```
#ifndef CONFIG_KERNEL_MCS
    case cap_reply_cap:
        ret.status = EXCEPTION_NONE;
        ret.cap = cap_null_cap_new();
        break;
#endif
```

So in non-MCS, a copy/mint of a reply cap with reduced rights won't fail, but won't actually give you a copy. I'm not sure why, but we seem to want to ensure that there is only a single non-master reply cap, since `CNode_SaveCaller` acts as a move. Mutate and rotate don't allow changing rights, so IIUC, there is no way to remove the grant right from a non-MCS reply cap.

That makes the decision between the alternatives easier, since we don't need to worry about existing non-MCS users wanting to migrate code that masks reply grant rights.

mbrcknl:
I've posted an RFC at https://sel4.atlassian.net/browse/RFC-13. I'd welcome any feedback on it before (or after) I post it to the mailing list.

kevinelp:
The RFC seems good. I'll make our discussion more complicated by mentioning an alternative for setting the grant right associated with the synthesised reply right (in the reply object). One could introduce "take" rights in the original sender cap and set the grant for reply accordingly. Of the present options under discussion, a server can either grant to all clients or none. Client specific "take" could provide more flexibility, but I don't see much push for it. Hence, this is more a mention. The current grant scheme is a simplification of the more general take/grant, so I was always expecting a discussion about providing "take" at some point, however we don't seem to be quite there yet.

lsf37 (approved):
Happy with this!

Indanz (approved):

Indanz:
Is this waiting for verification before this can be merged?

lsf37:
> Is this waiting for verification before this can be merged?

Yes, that's correct.

corlewis:
Rebased the branch, so that it can be tested with the verification changes.

Indanz:
> Rebased the branch, so that it can be tested with the verification changes.

You made a mistake with the rebase and modified all merged commits in the process. You want your commits to be at the end of the rebase so you retain the commits from master.

corlewis:
> > Rebased the branch, so that it can be tested with the verification changes.
> 
> You made a mistake with the rebase and modified all merged commits in the process. You want your commits to be at the end of the rebase so you retain the commits from master.

Ah, I think I see the problem, the `mcs-grant-via-reply` branch needs to be rebased as well since it's the target of this PR.

lsf37:
(rebased)

midnightveil:
Not sure if I've followed the RFC rabiithole far enough here, but given seL4/l4v#808 is done against this alternative, does it make sense to retarget this branch onto mainline instead of having this diff against the first alternative? (and rework commits as appropriate?).

Or have I missed something and the choice is still up in the air?

lsf37:
> Not sure if I've followed the RFC rabiithole far enough here, but given [seL4/l4v#808](https://github.com/seL4/l4v/pull/808) is done against this alternative, does it make sense to retarget this branch onto mainline instead of having this diff against the first alternative? (and rework commits as appropriate?).
> 
> Or have I missed something and the choice is still up in the air?

The choice was for this one, but they stack nicely and are largely orthogonal in what code they touch. Basically this option just goes one step further than the first option in #874. We could make it one PR, but the discussion on both is useful in addition to the RFC, I think.

PR #948, opened 2022-11-04
boot/arm: remove activate_global_pd usage

axel-h:
`activate_global_pd` is just an alias for `activate_kernel_vspace` nowadays. Seems the C parts is trivial, not sure about the impact on the proofs.


axel-h:
Looks like the proofs don't have an issue also.

lsf37 (approved):
Happy with that, there's not point having a separate name for it.

PR #949, opened 2022-11-10
zynq7000: fix DTS overlay on MCS

axel-h:
Apply changes from commit 0f619780 for MCS also. Without them there is no output visible when running e.g. sel4test (tested on QEMU only)

### DIFF
--- a/src/plat/zynq7000/mcs-overlay-zynq7000.dts
+++ b/src/plat/zynq7000/mcs-overlay-zynq7000.dts
@@ -6,6 +6,9 @@
 
 / {
     chosen {
+        seL4,elfloader-devices =
+            "serial0",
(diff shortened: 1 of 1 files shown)


axel-h:
I wonder, am I the only one running into this?

lsf37:
There's currently no MCS on for qemu tests and I don't think we have a `zynq7000` in the machine queue. So it's entirely possible that we haven't seen the problem yet.

I did notice that `zynq7000` didn't work for MCS on qemu when I tried last week, although I don't think it was related to not producing output.

lsf37 (approved):
No objections from my side.

axel-h:
The ltimer still does not work (in qemu real time) on this platform, but my understanding is, that this should not block MCS in general.

PR #951, opened 2022-11-15
risc-v: support rv32 targets for qemu-riscv-virt

axel-h:
Fix a build breaker.

Test with: seL4/ci-actions#234

### DIFF
--- a/include/drivers/irq/riscv_plic0.h
+++ b/include/drivers/irq/riscv_plic0.h
@@ -69,12 +69,12 @@
 
 
 
-static inline uint32_t readl(uint64_t addr)
+static inline uint32_t readl(word_t addr)
 {
     return *((volatile uint32_t *)(addr));
 }
 
-static inline void writel(uint32_t val, uint64_t addr)
+static inline void writel(uint32_t val, word_t addr)
 {
     *((volatile uint32_t *)(addr)) = val;
 }
@@ -136,7 +136,7 @@ static inline void plic_complete_claim(irq_t irq)
 
 static inline void plic_mask_irq(bool_t disable, irq_t irq)
 {
-    uint64_t addr = 0;
+    word_t addr = 0;
     uint32_t val = 0;
     uint32_t bit = 0;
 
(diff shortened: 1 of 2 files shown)


axel-h:
I wonder, is anybody using seL4 on RV32 at the moment? If so please let me know which platforms are out there that you work on. Thanks

kent-mcleod (approved):

lsf37 (approved):

Ivan-Velickovic:
seL4test will build/pass with this patch: https://github.com/Ivan-Velickovic/seL4/commit/5bc7ad983bd1d04ccf3c41994a995af85109ca00 (in addition to this PR).

lsf37:
Cool! @axel-h , could you add that commit to this PR? (assuming you're fine with that @Ivan-Velickovic)

lsf37 (approved):

Ivan-Velickovic:
Giving QEMU 2GiB of RAM doesn't work, it needs to be less than 2GiB since `avail_p_regs` will refer to the end of memory, `0x10000000` in this case, which doesn't fit in a 32-bit word.

axel-h:
Ture. Sorry, this slipped my tests. So currently we still get the error
```
kernel/gen_headers/plat/machine/devices_gen.h:49:18: error: unsigned conversion from 'long long int' to 'long unsigned int' changes value from '4294967296' to '0' [-Werror=overflow]
   49 |         .end   = 0x100000000
      |                  ^~~~~~~~~~~
```
instead of 
```
   49 |         .end   = 0x140000000
      |                  ^~~~~~~~~~~
```

Seems the most trivial fix really is just using '2047' in thge CMake scripts. But that does not solve the general problem that QEMU might be configures to have more memory, and seL4 should just ignore the space it can't handle. And even more general, any 32-bit platform should be allowd to have memory up to the end. Seems this would be addresses best in the Python scripts. For the RISC-V/ARM QEMU platforms, I'm looking into reserving all memory above '0xffffffff' in a dedicated 32-bit overlay DTS for the kernel. Then the scripts you adjust the usable regions to end a '0xffffffff'.

PR #958, opened 2022-12-16
mcs: remove redundant memzero from createObject

michaelmcinerney:
This removes a memzero call from the seL4_SchedContextObject and seL4_ReplyObject cases of createObject. The memory is now cleared via clearMemory within resetUntypedCap

Signed-off-by: Michael McInerney <michael.mcinerney@proofcraft.systems>


lsf37 (approved):
Yup, at this point in the code, the memory has already been cleared.

kent-mcleod (approved):

PR #961, opened 2023-01-07
manual: Fix a typo in threads.tex

nuta:
👋 In the following table, an undefined field `seL4_VMFault_SP` is referenced despite `seL4_VMFault_Addr` is not documented here. I guess it's just a typo.

Screenshot 2023-01-07 at 21 28 22


axel-h:
Given the existing description, that makes sense.

kent-mcleod (approved):

kent-mcleod:
Thanks! You're right that this does appear to be a typo.

PR #968, opened 2023-02-02
mcs: remove `getKernelWcetTicks` term from `refill_ready`

michaelmcinerney:
Please see the commit message for a brief explanation of this change. I'll elaborate on that message a little here, but I should point out that I have made this change to the abstract specification, and the proofs for all the invariants we have (which are very detailed regarding the refill lists) all work very nicely.

`refill_ready` currently says that an sc is ready if its head `rTime` is at most the current time plus the kernel WCET. I think there are a couple problems with this. 

First, `refill_unblock_check` might be called on an sc that is ready, but whose head `rTime` is somewhere strictly between the current time, and the current time plus the kernel WCET. This would result in `refill_unblock_check` bringing the head refill forward. This breaks an invariant that the verification team have (we call it `window`, within` valid_refills`) that states that the `rTime` of the last refill in the list must be at most the period from the `rTime` of the head refill. This also seems to violate a general principle that the kernel functions should never bring a refill's time forward (but may take its time back).

More generally speaking, again consider a sc whose whose head `rTime` is somewhere strictly between the current time, and the current time plus the kernel WCET. Since the kernel WCET is the worst case execution time, the time to exit the kernel must be strictly less than the kernel WCET, and so we could end up in a situation where we have a thread running on an sc whose `rTime` is still in the future, which seems to be a violation of the timing model for MCS.

Removing the `getKernelWcetTicks` term from `refill_ready` solves both these issues.

### DIFF
--- a/include/arch/x86/arch/machine/timer.h
+++ b/include/arch/x86/arch/machine/timer.h
@@ -62,7 +62,6 @@ static inline PURE time_t ticksToUs(ticks_t ticks)
 
 static inline void setDeadline(ticks_t deadline)
 {
-    assert(deadline > NODE_STATE(ksCurTime));
     if (likely(x86KSapicRatio == 0)) {
         x86_wrmsr(IA32_TSC_DEADLINE_MSR, deadline);
     } else {
--- a/include/arch/x86/arch/machine/timer.h
+++ b/include/arch/x86/arch/machine/timer.h
@@ -65,9 +65,10 @@ static inline void setDeadline(ticks_t deadline)
     if (likely(x86KSapicRatio == 0)) {
         x86_wrmsr(IA32_TSC_DEADLINE_MSR, deadline);
     } else {
-        /* convert deadline from tscKhz to apic khz */
-        deadline -= getCurrentTime();
-        apic_write_reg(APIC_TIMER_COUNT, div64(deadline, x86KSapicRatio));
+        /* Must not underflow */
+        deadline -= MIN(deadline, getCurrentTime());
+        /* Convert deadline from tscKhz to apic khz. Must be at least 1 tick. */
+        apic_write_reg(APIC_TIMER_COUNT, MAX(1, div64(deadline, x86KSapicRatio)));
     }
 }
 #else
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -585,6 +585,13 @@ void setNextInterrupt(void)
         next_interrupt = MIN(refill_head(NODE_STATE(ksReleaseHead)->tcbSchedContext)->rTime, next_interrupt);
     }
 
+    /* We should never be attempting to schedule anything earlier than ksCurTime */
+    assert(next_interrupt >= NODE_STATE(ksCurTime));
+
+    /* Our lower bound ksCurTime is slightly in the past (at kernel entry) and
+       we are further subtracting getTimerPrecision(), so we may be setting a
+       deadline in the past. If that is the case, we assume the IRQ will be
+       raised immediately after we leave the kernel. */
     setDeadline(next_interrupt - getTimerPrecision());
 }
 
--- a/include/arch/riscv/arch/machine/timer.h
+++ b/include/arch/riscv/arch/machine/timer.h
@@ -62,7 +62,6 @@ static inline ticks_t getCurrentTime(void)
 /* set the next deadline irq - deadline is absolute */
 static inline void setDeadline(ticks_t deadline)
 {
-    assert(deadline > NODE_STATE(ksCurTime));
     /* Setting the timer acknowledges any existing IRQs */
     sbi_set_timer(deadline);
 }
--- a/include/drivers/timer/arm_generic.h
+++ b/include/drivers/timer/arm_generic.h
@@ -24,7 +24,6 @@ static inline ticks_t getCurrentTime(void)
 /** DONT_TRANSLATE **/
 static inline void setDeadline(ticks_t deadline)
 {
-    assert(deadline >= NODE_STATE(ksCurTime));
     SYSTEM_WRITE_64(CNT_CVAL, deadline);
 }
 
--- a/include/drivers/timer/arm_global.h
+++ b/include/drivers/timer/arm_global.h
@@ -70,7 +70,9 @@ static inline void setDeadline(ticks_t deadline)
     globalTimer->comparatorUpper = (uint32_t)(deadline >> 32llu);
     /* enable cmp */
     globalTimer->control |= BIT(COMP_ENABLE);
-    /* if this fails PRECISION is too low */
+    /* Assert that either the deadline is in the future or that the IRQ has already been raised.
+       This should be guaranteed by hardware from r2p0 on of the Cortex-A9 MPCore Technical
+       Reference Manual (r2p0 published in 2009) */
     assert(getCurrentTime() < deadline || globalTimer->isr == 1u);
 }
 
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -126,7 +126,7 @@ static inline bool_t refill_sufficient(sched_context_t *sc, ticks_t usage)
  */
 static inline bool_t refill_ready(sched_context_t *sc)
 {
-    return refill_head(sc)->rTime <= (NODE_STATE(ksCurTime) + getKernelWcetTicks());
+    return refill_head(sc)->rTime <= NODE_STATE(ksCurTime);
 }
 
 /*
--- a/include/machine/timer.h
+++ b/include/machine/timer.h
@@ -16,7 +16,9 @@
 /* Read the current time from the timer. */
 /** MODIFIES: [*] */
 static inline ticks_t getCurrentTime(void);
-/* set the next deadline irq - deadline is absolute */
+/* Set the next deadline irq - deadline is absolute and may be slightly in
+   the past. If it is set in the past, we expect an interrupt to be raised
+   immediately after we leave the kernel. */
 /** MODIFIES: [*] */
 static inline void setDeadline(ticks_t deadline);
 /* ack previous deadline irq */


lsf37:
Test results show failures for MCS tests now. On closer look, apart from the two currently dodgy boards (imx8mm and zcu102), all failures are in `debug` configurations.

Is it possible that this change makes the kernel more sensitive to the configured kernel WCET being correct? The default config value is likely too low when the kernel is printing things.

lsf37:
To be fair, in the failing test
```
2023-02-02T08:29:50.8239157Z Running test SCHED0012 (Test one periodic thread)
2023-02-02T08:29:50.8239718Z Tick
2023-02-02T08:29:50.8240110Z Tock 1
2023-02-02T08:29:51.8254222Z Tick
2023-02-02T08:29:52.0044619Z Tock 2
2023-02-02T08:29:52.8239919Z Tick
2023-02-02T08:29:53.0027874Z Tock 3
2023-02-02T08:29:54.0046179Z seL4 failed assertion 'deadline > NODE_STATE(ksCurTime)' at /github/workspace/kernel/include/arch/x86/arch/machine/timer.h:59 in function setDeadline
2023-02-02T08:29:54.0046784Z halting...
2023-02-02T08:29:54.0047109Z Kernel entry via Interrupt, irq 157
```
assuming the `Tick` messages are user-level printing not kernel-level printing, there wouldn't be any influence. So it might just be that assertions are off in the other builds.

lsf37:
Ok, so after some debugging and discussion with @michaelmcinerney, we think we know what is going on.

Consider the function `setNextInterrupt` with a few more `asserts` added:

```c
void setNextInterrupt(void)
{
    time_t next_interrupt = NODE_STATE(ksCurTime) +
                            refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount;

    if (numDomains > 1) {
        next_interrupt = MIN(next_interrupt, NODE_STATE(ksCurTime) + ksDomainTime);
    }

    if (NODE_STATE(ksReleaseHead) != NULL) {
        assert(refill_head(NODE_STATE(ksReleaseHead)->tcbSchedContext)->rTime >= NODE_STATE(ksCurTime) + MIN_BUDGET);
        next_interrupt = MIN(refill_head(NODE_STATE(ksReleaseHead)->tcbSchedContext)->rTime, next_interrupt);
    }

    assert(refill_head(NODE_STATE(ksCurThread)->tcbSchedContext)->rAmount >= MIN_BUDGET);
    setDeadline(next_interrupt);
}
```

What is now failing is the `assert` for the `rTime` of `ksReleaseHead`, which is still strictly after `ksCurTime`, but not sufficiently late for it to satisfy `MIN_BUDGET`. In fact, in the test is less than `getTimerPrecision()` after `ksCurTime` (if I remove the timer precision term, all tests pass without the new assertions).

With the new assertions, we get

```
unning test SCHED0012 (Test one periodic thread)
Tick
Tock 1
seL4 failed assertion 'refill_head(NODE_STATE(ksReleaseHead)->tcbSchedContext)->rTime >= NODE_STATE(ksCurTime) + MIN_BUDGET' at /host/test/kernel/src/kernel/thread.c:585 in function setNextInterrupt
```

The `rAmount` assertion for the current thread does not fail, which is good, because that follows directly from one of the invariants we have on refills (could currently still be out of sync with the code, but we don't think it is). For the release queue, we only know that `rTime` is after `ksCurTime`, but not necessarily by `MIN_BUDGET`.

So we either have to establish that invariant, or not take the release queue into account when we set the next deadline (which would then mean we no longer know that `rTime` is always > `ksCurTime` in the release queue).

michaelmcinerney:
To add to Gerwin's comment...

I think it's fine if the threads in the release queue are within `MIN_BUDGET` of the current time. The release queue will indeed consist of threads that aren't ready (once `awaken` has run). We have not proved this property, but it will follow easily from the loop condition in `awaken`. 

But I don't think we should modify `next_interrupt` if the release queue is nonempty. At the point in the kernel where `next_deadline` is called, we have chosen the current thread and will want to run it with its current sc. So we are committed to running that sc for its head amount (maybe minus the timer precision). Now, the amount in this head refill might be quite large, and we could run for quite a while before we reenter the kernel. Why should we prematurely end this thread running if there just so happens to be a thread in the release queue which has an sc that would become ready during this time? It seems to me that the thread at the head of the release queue will just have to wait its turn and perhaps run after the current thread has finished running on its sc. 

I'm not sure whether this line of reasoning would have any impact on whether we reprogram the timer when we modify the release queue. I think that is done at some point currently

lsf37:
The head of the release queue could be at higher priority than the current thread, so when the head becomes potentially ready, we need to check. I suspect that is why we're currently taking the interrupt earlier.

lsf37:
Trying to draw a diagram of the situation:

```
---- t ---][--- k ---][---- t2 ---------
           ^a         ^b          ^c
```
where `t` is the previous thread, `k` is the kernel, `t2` is the thread we're scheduling, `a` is the time at kernel entry (=`ksCurTime`), `b` is the time at kernel exit, and `c` is the `rTime` of the head of the release queue, i.e. the time at which another thread of maybe higher priority might become ready to run again.

There are multiple issues here, but the main one is mostly about how close we want `c` to come to `b`.

From a higher-level perspective, these are the properties we think should hold:

1. no thread should run before its `rTime` (release time) has occured
2. the highest priority thread that is ready should always be running
3. the kernel can postpone an `rTime` (increase it), but cannot decrease it (make it earlier)
4. the next timer interrupt must be in the future
5. (?) if we schedule a thread, we want it to run for a minimum time before the next timer interrupt comes in
6. `t2`'s budget is at least `MIN_BUDGET`, i.e. left to its own devices and without other interrupts, it would run for at least `MIN_BUDET` (which is `2*WCET`)

These cannot all be satisfied at the same time. Currently the kernel is violating property (1), because it declares threads as ready when they are within `WCET` of their `rTime`, so the get scheduled too early. @michaelmcinerney's change was trying to fix that because it leads to other problems in the proofs. The fix violates (4), because the rest of the computation applies a fudge factor (`getTimerPrecision()`) to the deadline to schedule it slightly before `rTime` (not entirely sure why we do this, but it's intended behaviour). If the `rTime` is close to the current time, we the fudge factor means we set the interrupt for the past. This didn't happen before, because we were always at least WCET away from the next deadline.

Before getting into what to do, one problem I have with all of that computation inside `setNextInterrupt` is that it is based on `ksCurTime`, which is the current time at kernel entry (updated at preemption points), not the current time at when we compute the deadline. This might be necessary to correctly account kernel time to threads (@kent-mcleod or @pingerino can you confirm?), but I'm not sure we are covering all the cases.

In particular, this means the `assert` inside most of the `setDeadline` implementations is still fairly relaxed: it doesn't check that we're setting the deadline in the future, it only checks that we're not setting it before kernel entry. There's probably a good reason for this, not sure if reading the timer inside the `assert` has too much side effect or is itself not that fast or something like that. In any case, we might be failing the condition we actually want more often than we think. @pingerino do you remember if/why adding an assert for `deadline > getCurrentTime()` would be problematic? (like in the arm global timer implementation)

It is actually not quite clear to me what the usual relationship between `a`, `b`, `WCET`, and `timerPrecision` is. `timerPrecision` is currently mostly set to 1us, and WCET to 10us. On a 1GHz machine that means 1000 cycles can happen within `timerPrecision` and 10k cycles within `WCET`. So for a fast syscall `b` might actually be less than `timerPrecision` from `a` and setting a deadline of `a + x - timerPrecision` might really set that deadline in the past before kernel entry (this is the assert we're seeing fail). This was not actually much better in the old version (before the change), just for a different case: if we're in a long-running call so that the current time in `setNextInterrupt` is slightly less than `a + WCET`, and we're setting the new deadline to `a + x - timerPrecision` where `x` is only slightly larger than WCET, we are still setting a deadline in the past, but we're not catching it in the `assert`.

So either case can actually violate property (4). It'd be good to try to make a stress test that gets this to happen so that we can get some confidence we're not triggering that any more.

(This also means that a thread with MIN_BUDGET doesn't really get MIN_BUDGET to run in this case, but only MIN_BUDGET - WCET, because the deadline is `a + MIN_BUDGET`, but the current time is `a + WCET` already,  which was probably clear to everyone but me :-))

I'm unclear on how important property (5) is or if it is actually a property we are aiming to guarantee. If it is not, we could just set the next IRQ deadline to the maximum of something like `getCurrentTiime() + getTimerPrecision()` and the deadline we have computed before. This would mean we're getting an interrupt very soon after we've exited the kernel. It'd potentially be slightly later than we wanted it, but that would be consistent with (3), and I think kernel entry would account for the time correctly. It could violate (2) for a `timerPrecision` amount of time.

If we do want something like (5), we need to decide on how much buffer we want. WCET? MIN_BUDGET? If yes, why? We would then potentially violate (2) for that amount of time. I'm less clear on how to achieve that, although forcing the deadline to be at least `getCurrentTiime() + bufferTime` in `setDeadline` like in the previous case might work.

Any opinions?

Indanz:
(Although I changed my email address in Github, notifications were still sent to my previous work address, so I missed all notifications sent to me the last month. Should be okay now.)

I'll look into this in more detail later, but here a few comments:

My feeling is that WCET is sometimes added arbitrarily, this seems one of those places.  WCET is maximum system call execution time, but sometimes what is actually wanted is the maximum kernel entry time. I think that is/should be already covered by `timerPrecision`.

> The fix violates (4), because the rest of the computation applies a fudge factor (getTimerPrecision()) to the deadline to schedule it slightly before rTime (not entirely sure why we do this, but it's intended behaviour).

Otherwise tasks would run up to `getTimerPrecision()` too long (on top of the HW timer precision). `getTimerPrecision()` should be close to the time to takes the timer IRQ and to get the current time, in addition to the actual timer precision (see also #849). Not doing this would violate (2).

> If the rTime is close to the current time, the fudge factor means we set the interrupt for the past.

As long as the timer interrupt triggers when this happens, this is harmless. The alternative would be to add an explicit check for this, and if it happens, do the timer IRQ handling code immediately. But this would increase WCET.

So I think (4) could be relaxed to that the kernel should set a new deadline equal or higher to `ksCurTime`.

> Before getting into what to do, one problem I have with all of that computation inside setNextInterrupt is that it is based on ksCurTime, which is the current time at kernel entry (updated at preemption points), not the current time at when we compute the deadline. This might be necessary to correctly account kernel time to threads

Yes, that is the case. The current kernel entry time is billed to the current thread, so using `ksCurTime` is appropriate, otherwise no one would be billed for the time between `ksCurTime` and the `setNextInterrupt()` call.

And because all timing is relative to `ksCurTime`, not subtracting the time between the timer IRQ triggering and the update of `ksCurTime` for the next timer update time would lead to accumulating timing errors.

> I'm unclear on how important property (5) is or if it is actually a property we are aiming to guarantee. 

I don't think (5) is necessary with (6) in place.

lsf37:
Ok, so if (5) is not something we want, I think it'd be perfectly fine to force the deadline to be at most at `ksCurTime` in the `rTime` case, i.e. taking the max of what the `rTime` gives us and `ksCurTime + getTimePrecision()`, because we then deduct `getTimerPrecision()` again and we get precisely `ksCurTime`.

What then happens is that we exit the kernel, we immediately take a timer interrupt and wake up the head of the release queue. That means it's at most `getTimerPrecision()` late compared to what we do now, where it might instead be almost `WCET` early.

lsf37:
Can we get into a situation where we have many threads in the release queue and they all have `rTime`s of `ksCurTime + 1`, `+2`, `+3`, etc?

Because that could cause a cascade of such interrupts. Although that's not quite right, because more than one tick passes even if user space gets no time at all, so we'd wake a few of them at the same time.

Ok, but imagine they are staggered perfectly at something like `ksCurTime + timerPrecision`, `ksCurTime + 2 * timerPrecision`, etc. That *would* lead to very little user space work and a lot of kernel spinning on waking up threads. Would we be able to construct such a system? Would it be possible to get into that state without wanting to?

Indanz:
Yes, but only with many sporadic tasks with different priorities. The only way to avoid that is to let tasks run too early.

If it happens, the interrupted tasks will be delayed by the highest priority task and after that everything is ready to run and sorted by priority and rTime, so it's unlikely to happen again. It can only stick in this state if they are periodic, call yield and are launched unfortunately, but that's worked around easily by launching highest priority tasks first.

Indanz (approved):
All in all this seems to reduce the timing errors, with no great downsides. Would be good to compare sel4bench before and after this change, to see if it makes a difference there.

lsf37:
Rebased and added the check/`getTimerPrecision` update from the discussion above so that we never set a timer IRQ earlier than `ksCurTime`.

We will still potentially be setting timer interrupts that are slightly in the past, because `ksCurTime` is the kernel entry time, not the kernel exit time. This will only work if we can reliably assume that timer interrupts set in the past will be raised immediately.

Does anyone know if that is guaranteed by the hardware? @kent-mcleod @Indanz maybe?

Indanz:
> This will only work if we can reliably assume that timer interrupts set in the past will be raised immediately.

There is always a race between the current time and when setting the next timeout, because the current time can increment just before setting the next interrupt. So either hardware needs to be designed to deal with it (easy with 64-bit timers), or the `setDeadline()` function should check that no race happened.

However, if current implementations of `setDeadline()` assume the next timeout will be far enough in the future that a race is not possible, then changing this may break things, if hardware can't deal with timeouts in the past.

- x86 TSC deadline seems fine.
- x86 APIC TIMER COUNT subtracts `getCurrentTime()` from the deadline and can't deal with negative offsets. To me this seems broken, even for the current implementation.
- RISC-V does an SBI call so is should be safe.
- ARM generic timer is fine.
- ARM global timer seems okay, except it has an assert that may trigger.

lsf37:
Excellent, thanks for that survey. And I think I know what is going wrong in my patch: all of the computed times can be just after `ksCurTime`, not just the one in the one branch I'm adjusting. So I'll move the check further down, which should also make it simpler.

You're right about x86 APIC TIMER COUNT, that does sound broken in any case. We should look into that.

lsf37:
Ok, I've addressed the potential APIC TIMER COUNT underflow and tweaked the other `setDeadline` assertion that assumed `deadline > ksCurTime` instead of only `deadline >= ksCurTime`, because that's the only thing we can guarantee from the rest of the code.

I have also moved the timer precision check directly to where it's relevant, so hopefully that is a bit clearer.

include/arch/x86/arch/machine/timer.h
```diff
@@ -62,13 +62,15 @@ static inline PURE time_t ticksToUs(ticks_t ticks)
...
         x86_wrmsr(IA32_TSC_DEADLINE_MSR, deadline);
     } else {
-        /* convert deadline from tscKhz to apic khz */
-        deadline -= getCurrentTime();
-        apic_write_reg(APIC_TIMER_COUNT, div64(deadline, x86KSapicRatio));
+        /* Must not underflow, and getCurrentTime() will be later than ksCurTime,
+           so the assert above does not protect from underflow here. */
+        deadline -= MIN(deadline, getCurrentTime());
+        /* Convert deadline from tscKhz to apic khz. Must be at least 1 tick. */
+        apic_write_reg(APIC_TIMER_COUNT, MIN(1, div64(deadline, x86KSapicRatio)));
```
Indanz, line 73: This should be MAX(1, ...).

src/kernel/thread.c
```diff
@@ -585,7 +585,15 @@ void setNextInterrupt(void)
...
-    setDeadline(next_interrupt - getTimerPrecision());
+    assert(next_interrupt >= NODE_STATE(ksCurTime));
+
+    /* Either of the rTime values above might sit just after ksCurTime (less
+       than getTimerPrecision() in the future), so we need to make sure that
+       subtracting getTimerPrecision() does not put us into the past. We are
+       fine with scheduling an interrupt at ksCurTime, even though that is
+       strictly speaking also in the past (at kernel entry), because we assume
+       it will be raised immediately after we leave the kernel. */
+    setDeadline(MAX(NODE_STATE(ksCurTime), next_interrupt - getTimerPrecision()));
```
Indanz, line 596: I don't see why this distinction is made. If we assume setDeadline can handle times in the past, why not let it just set a time before ksCurTime? That is, why not just remove all the asserts instead of updating them?

If it makes verification easier, we can move the `- getTimerPrecision()` part into the setDeadline() functions.

Indanz:
> I have also moved the timer precision check directly to where it's relevant, so hopefully that is a bit clearer.

I don't see this change.

include/arch/x86/arch/machine/timer.h
lsf37: Indeed, thank you.

src/kernel/thread.c
lsf37: We could do that, we'd just have to remove the assertions about `ksCurTime` from the insides of the various `setDeadline` implementations. I didn't want to add that much code disturbance, but it would simplify things a bit. I'll give it a shot.

Apart from `ackDeadlineIRQ` in `arm_generic.h`, this is the only code that calls `setDeadline`.

lsf37:
On the `tx1` board, the SCHED0021 test fails with this patch now:

```xml
  	<testcase classname="sel4test" name="SCHED0021">
  Running test SCHED0021 (Test for pre-emption during running of many threads with equal prio)
  		<error>Check preemption_thread_data[thread](2) == 1(1) failed. at line 1620 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
  		<error>Check preemption_thread_data[thread](2) == 1(1) failed. at line 1620 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
  		<error>Check preemption_thread_data[thread](2) == 1(1) failed. at line 1620 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
  		<error>Check preemption_thread_data[thread](2) == 1(1) failed. at line 1620 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
  		<error>Check duration(40027000) < ((((5 * (1000UL * 1000UL)) * (4 + 1)) * (102)) / (100))(25500000) failed. at line 1631 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
  Test SCHED0021 failed
  		<failure type="failure">result == SUCCESS at line 296 of file /github/workspace/projects/sel4test/apps/sel4test-driver/src/testtypes.c</failure>
  	</testcase>
  	<testcase classname="sel4test" name="SCHED_CONTEXT_0001">
```

Will need to look into the test tomorrow to see what's going on. I have not seen it fail this way before, so this is new behaviour at least.

lsf37:
> > I have also moved the timer precision check directly to where it's relevant, so hopefully that is a bit clearer.
> 
> I don't see this change.

Sorry for the force-push, I shouldn't do that. The `MAX` term moved from one of the `if` branches in `setNextInterrupt` to the `setDeadline` invocation.

Indanz:
I don't think the test is robust, it should do another yield call before

```
test_simple_preempt_start = 1;
```
otherwise there is a chance it will be scheduled after the write, but before the actual yield system call is executed.

Indanz:
> > > I have also moved the timer precision check directly to where it's relevant, so hopefully that is a bit clearer.
> > 
> > 
> > I don't see this change.
> 
> Sorry for the force-push, I shouldn't do that. The `MAX` term moved from one of the `if` branches in `setNextInterrupt` to the `setDeadline` invocation.

Force-push is fine, but in this case I assumed you meant a change compared to mainline, not a change compared to your previous push, hence the confusion.

src/kernel/thread.c
lsf37: I have addressed these now (still need to squash things into coherent commits). In `arm_global_timer.h` we still have the following:

```c
    /* enable cmp */
    globalTimer->control |= BIT(COMP_ENABLE);
    /* if this fails PRECISION is too low */
    assert(getCurrentTime() < deadline || globalTimer->isr == 1u);
```

but I think that holds, because it basically says just after enabling the interrupt that it is either in the future or it has been raised just now, which is what we want.

Indanz: Yes, you're right.

It seems it depends on the hardware revision whether ARM global timer can handle times in the past:

https://developer.arm.com/documentation/ddi0407/g/Global-timer--private-timers--and-watchdog-registers/About-the-Global-Timer

> In r2p0 the comparators for each processor with the global timer fire when the timer value is greater than or equal to. In previous revisions the comparators fired when the timer value was equal to.

Revision 2 came out in 2009, only one year after Cortex-A9 was released, so it seems unlikely to be used by actual hardware released years later. The imx6 and zynq7000 platforms use the ARM global timer, both were released in 2011.

lsf37: - [ ] I should probably add a comment to that effect to the `assert`

lsf37:
sel4bench results are now [there](https://github.com/seL4/seL4/actions/runs/6713040391?pr=968), the corresponding run for the ancestor on the master branch is [here](https://github.com/seL4/sel4bench/actions/runs/6672012855).

As always, I find these hard to interpret. All changes seem to be within the noise (i.e. the means less than std dev on each side apart). Some look a few cycles faster, some a few cycles slower, I didn't see anything conclusive.

I'll start rebasing and cleaning up the commits.

lsf37:
(rebase and commit squash now done)

lsf37 (approved):
From my side, these changes are now ready to merge. @Indanz are you also fine with the updated version?

Indanz (approved):
All looks good to me.

I won't expect any significant changes in benchmarking numbers with these kind if changes.

lsf37:
Ok, rebased again and added that final comment that was still open.

PR #969, opened 2023-02-02
Rocketchip bug fixes + zcu102 softcore

chrisguikema:
This PR does two things:

1. Differentiates the base Rocketchip platform with a version generated for the ZCU102 FPGA fabric
2. Fixes bugs we encountered running on the softcore implementation

We tested this with the VMM created by @yyshen

### DIFF
--- a/src/arch/riscv/common_riscv.lds
+++ b/src/arch/riscv/common_riscv.lds
@@ -2,6 +2,7 @@
  * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
  * Copyright 2015, 2016 Hesham Almatary <heshamelmatary@gmail.com>
  * Copyright 2021, HENSOLDT Cyber
+ * Copyright 2023, DornerWorks
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -90,6 +91,12 @@ SECTIONS
 
         /* large data such as the globals frame and global PD */
         *(.bss.aligned)
+#if defined(CONFIG_PLAT_ROCKETCHIP_ZCU102)
+        /* softcore instantiation on ZCU102 needs these bytes to
+         * avoid userspace issues
+         */
+        . = . + 8;
+#endif
     }
 
     . = ALIGN(4K);
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -233,6 +233,13 @@ static BOOT_CODE bool_t try_init_kernel(
     /* If a DTB was provided, pass the data on as extra bootinfo */
     p_region_t dtb_p_reg = P_REG_EMPTY;
     if (dtb_size > 0) {
+#ifdef CONFIG_PLAT_ROCKETCHIP_ZCU102
+        /* The softcore rocketchip instantiation doesn't work well when this
+         * page isn't reserved. Round up so the whole page is reserved to
+         * avoid the problem
+         */
+        dtb_size = ROUND_UP(dtb_size, PAGE_BITS);
+#endif
         paddr_t dtb_phys_end = dtb_phys_addr + dtb_size;
         if (dtb_phys_end < dtb_phys_addr) {
             /* An integer overflow happened in DTB end address calculation, the
--- a/src/arch/riscv/machine/hardware.c
+++ b/src/arch/riscv/machine/hardware.c
@@ -102,6 +102,12 @@ static inline irq_t getActiveIRQ(void)
          * has claimed it in a multicore system.
          */
         irq = plic_get_claim();
+#ifdef CONFIG_PLAT_QEMU_RISCV_VIRT
+        /* QEMU bug requires external interrupts to be immediately claimed. For
+         * other platforms, the claim is done in invokeIRQHandler_AckIRQ.
+         */
+        plic_complete_claim(irq);
+#endif
 #ifdef ENABLE_SMP_SUPPORT
     } else if (sip & BIT(SIP_SSIP)) {
         sbi_clear_ipi();
--- a/src/object/interrupt.c
+++ b/src/object/interrupt.c
@@ -134,8 +134,15 @@ exception_t decodeIRQHandlerInvocation(word_t invLabel, irq_t irq)
 void invokeIRQHandler_AckIRQ(irq_t irq)
 {
 #ifdef CONFIG_ARCH_RISCV
+#if !defined(CONFIG_PLAT_QEMU_RISCV_VIRT)
+    /* QEMU has a bug where interrupts must be
+     * immediately claimed, which is done in getActiveIRQ. For other
+     * platforms, the claim can wait and be done here.
+     */
     plic_complete_claim(irq);
+#endif
 #else
+
 #if defined ENABLE_SMP_SUPPORT && defined CONFIG_ARCH_ARM
     if (IRQ_IS_PPI(irq) && IRQT_TO_CORE(irq) != getCurrentCPUIndex()) {
         doRemoteMaskPrivateInterrupt(IRQT_TO_CORE(irq), false, IRQT_TO_IDX(irq));
--- a/src/plat/rocketchip/config.cmake
+++ b/src/plat/rocketchip/config.cmake
@@ -1,6 +1,7 @@
 #
 # Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
 # Copyright 2021, HENSOLDT Cyber
+# Copyright 2023, DornerWorks
 #
 # SPDX-License-Identifier: GPL-2.0-only
 #
@@ -9,12 +10,36 @@ cmake_minimum_required(VERSION 3.7.2)
 
 declare_platform(rocketchip KernelPlatformRocketchip PLAT_ROCKETCHIP KernelArchRiscV)
 
+set(c_configs PLAT_ROCKETCHIP_BASE PLAT_ROCKETCHIP_ZCU102)
+set(cmake_configs KernelPlatformRocketchipBase KernelPlatformRocketchipZCU102)
+
+set(plat_lists rocketchip-base rocketchip-zcu102)
+foreach(config IN LISTS cmake_configs)
+    unset(${config} CACHE)
+endforeach()
+
 if(KernelPlatformRocketchip)
     declare_seL4_arch(riscv64)
-    config_set(KernelRiscVPlatform RISCV_PLAT "rocketchip")
+
+    check_platform_and_fallback_to_default(KernelRiscVPlatform "rocketchip-base")
+    list(FIND plat_lists ${KernelRiscVPlatform} index)
+    if("${index}" STREQUAL "-1")
+        message(FATAL_ERROR "Which rocketchip platform not specified")
+    endif()
+    list(GET c_configs ${index} c_config)
+    list(GET cmake_configs ${index} cmake_config)
+    config_set(KernelRiscVPlatform RISCV_PLAT ${KernelRiscVPlatform})
+    config_set(${cmake_config} ${c_config} ON)
+
     config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)
     config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "generic")
     list(APPEND KernelDTSList "tools/dts/rocketchip.dts")
+    # The Rocketchip-ZCU102 is a softcore instantiation that runs on the ZCU102's
+    # FPGA fabric. Information on generating and running seL4 on the platform can
+    # be found at https://docs.sel4.systems/Hardware/
+    if(KernelPlatformRocketchipZCU102)
+        list(APPEND KernelDTSList "src/plat/rocketchip/overlay-rocketchip-zcu102.dts")
+    endif()
     # This is an experimental platform that supports accessing peripherals, but
     # the status of support for external interrupts via a PLIC is unclear and
     # may differ depending on the version that is synthesized. Declaring no
--- a/src/plat/rocketchip/config.cmake
+++ b/src/plat/rocketchip/config.cmake
@@ -44,18 +44,23 @@ if(KernelPlatformRocketchip)
         # repo is added as a remote to the tools/opensbi project in the seL4 codebase
         config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "rocket-fpga-zcu104")
         list(APPEND KernelDTSList "src/plat/rocketchip/overlay-rocketchip-zcu102.dts")
+        # The zcu102 instantiation supports the PLIC and external interrupts
+        declare_default_headers(
+            TIMER_FREQUENCY 10000000 PLIC_MAX_NUM_INT 128
+            INTERRUPT_CONTROLLER drivers/irq/riscv_plic0.h
+        )
     else()
         config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "generic")
+        # This is an experimental platform that supports accessing peripherals, but
+        # the status of support for external interrupts via a PLIC is unclear and
+        # may differ depending on the version that is synthesized. Declaring no
+        # interrupts and using the dummy PLIC driver seems the best option for now
+        # to avoid confusion or even crashes.
+        declare_default_headers(
+            TIMER_FREQUENCY 10000000 PLIC_MAX_NUM_INT 0
+            INTERRUPT_CONTROLLER drivers/irq/riscv_plic_dummy.h
+        )
     endif()
-    # This is an experimental platform that supports accessing peripherals, but
-    # the status of support for external interrupts via a PLIC is unclear and
-    # may differ depending on the version that is synthesized. Declaring no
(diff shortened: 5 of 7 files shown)


src/arch/riscv/common_riscv.lds
```diff
@@ -80,6 +79,7 @@ SECTIONS
 
         /* large data such as the globals frame and global PD */
         *(.bss.aligned)
+        . = . + 8;
```
axel-h, line 82: Can you add a more verbose explanation here, why this padding is needed on your platforms. I have removed this, because it was completely unclear, why this is needed. Bringing it back is fine, but this time it must have a proper explanation, what is going on on this platform.

src/arch/riscv/kernel/boot.c
```diff
@@ -234,6 +234,13 @@ static BOOT_CODE bool_t try_init_kernel(
     /* If a DTB was provided, pass the data on as extra bootinfo */
     p_region_t dtb_p_reg = P_REG_EMPTY;
     if (dtb_size > 0) {
+#ifdef CONFIG_PLAT_ROCKETCHIP
+        /* The softcore rocketchip instantiation doesn't work well when this
+         * page isn't reserved. Round up so the whole page is reserved to
+         * avoid the problem
+         */
+        dtb_size = ROUND_UP(dtb_size, PAGE_BITS);
```
axel-h, line 242: Can you be more specific, what is the problem then? For the user mode mapping, we will have to use full pages anyway. Maybe we should run up for all platforms?

chrisguikema: I didn't go super in-depth here to figure out _why_ this was happening, but I can tell you the behavior we were seeing.

We ran yanyan's VMM just fine with an older codebase. When rebasing onto this branch, we ran into a couple issues. One was the 8 byte padding, the other was that user-space generated an "virtual address already mapped" error. By reverting a few old commits, things would work properly, but that's not a good solution for larger commits in my opinion.

I was able to get past the virtual address error by increasing the static allocator mempool size, but I'd immediately get a RAM fault when trying to run the VM. 

So I looked at what was happening in the commits we needed to revert, and one of them removed the rounding for the device tree. Like you mentioned, userspace needs to map in full pages anyways, so if a portion of a page is reserved, the entire page should be. My best guess is that there was an overlap somewhere there shouldn't have been, but like I said, I didn't grab GDB and look at every memory map, I just found the solution.

src/arch/riscv/common_riscv.lds
chrisguikema: I just tested it again, and its the same behavior I described with the DTB round up. The rootserver gets an "Virtual address already mapped" exception from the kernel.

chrisguikema:
By the way, we did all of our testing of the riscv_he_v0.6 branch, but these changes should also apply to `master` if discussion warrants it

src/object/interrupt.c
```diff
@@ -134,8 +134,15 @@ exception_t decodeIRQHandlerInvocation(word_t invLabel, irq_t irq)
 void invokeIRQHandler_AckIRQ(irq_t irq)
 {
 #ifdef CONFIG_ARCH_RISCV
-    return;
+#if !defined(CONFIG_PLAT_SPIKE)
+    /* QEMU has a bug where interrupts must be
```
Ivan-Velickovic, line 138: 1. Do you know if this apply to all versions of QEMU?
2. This should also apply to `CONFIG_PLAT_QEMU_RISCV_VIRT`

chrisguikema: @Furao do you know the answer to this? We can update the ifdef easy enough.

src/arch/riscv/machine/hardware.c
```diff
@@ -167,6 +168,10 @@ static irq_t handleVTimer(void)
 
 static irq_t active_irq[CONFIG_MAX_NUM_NODES];
 
+static inline void sbi_clear_ipi(void)
```
axel-h, line 171: Please remove this, this function is already defined in `include/arch/riscv/arch/sbi.h`, see https://github.com/seL4/seL4/blob/master/include/arch/riscv/arch/sbi.h#L101-L104

src/object/interrupt.c
axel-h: It's better to use explicit platform checks, so it is clear where this applies. Please check for `CONFIG_PLAT_QEMU_RISCV_VIRT` for now, as this is the official QEMU platform. 
There are other platforms (HiFive, PolarFire) where a QEMU simulation also exists. They may have the same problem, but here we should consider the real hardware as the default target. An additional kernel config define should be added to explicitly enable the fix for this specific QEMU bug.

src/arch/riscv/machine/hardware.c
chrisguikema: Done.

src/object/interrupt.c
chrisguikema: I agree that we should do an explicit check for QEMU_RISCV_VIRT, but that platform doesn't exist in `riscv_he_v0.6`, hence the check for `spike`.

@axel-h @Ivan-Velickovic do you think this PR should be rebased onto master?

chrisguikema:
@yyshen I think a rebase may have conflicted our commit `91f35b9b789cebb72d7d75eea8c9e3ff5d8b6d69` with your commit `bd660ffb71b756791e0b26ebfbbcbe558dae5944`. I'll do some testing to see what changes actually need to be made for the system to run on real hardware.

axel-h:
@chrisguikema: I have a PR pending to update OpenSBI, see https://github.com/seL4/seL4/pull/383
What version are you using?

chrisguikema:
> What version are you using?

We're using this branch of an OpenSBI fork: https://github.com/bao-project/opensbi/tree/bao/rocket

src/object/interrupt.c
Ivan-Velickovic: Ah, my bad. I forgot that this branch wouldn't have support for virt yet. Yes, I think it's a good idea to rebase the branch on master.

chrisguikema: Just to clarify, you're saying I should rebase `rocketchip_bug_fixes_zcu102` onto master, not `riscv_he_0.6`?

axel-h: It seems @Ivan-Velickovic ist trying to rebase `riscv_he_0.6`, so you can rebease on it then.

chrisguikema: I rebased this branch onto `master`

src/object/interrupt.c
```diff
@@ -134,8 +134,16 @@ exception_t decodeIRQHandlerInvocation(word_t invLabel, irq_t irq)
 void invokeIRQHandler_AckIRQ(irq_t irq)
 {
 #ifdef CONFIG_ARCH_RISCV
+#if !defined(CONFIG_PLAT_QEMU_RISCV_VIRT)
+    /* QEMU has a bug where interrupts must be
+     * immediately claimed, which is done in getActiveIRQ. For other
+     * platforms, the claim can wait and be done here
```
axel-h, line 140:
```suggestion
     * platforms, the claim can wait and be done here.
```

chrisguikema:
Once this PR is approved, I can link the resources we've used to run on the ZCU102 softcore rocketchip. It would probably be best to update the docs page, right?

src/arch/riscv/common_riscv.lds
chrisguikema: is it possible to wrap this change around the CONFIG_PLAT_ROCKETCHIP_ZCU102 ifdef? That way its only used for the platform its needed.

axel-h: In https://github.com/seL4/seL4/blob/master/CMakeLists.txt#L657-L669 the linker script run through the preprocessor, so an `#ifdef` could work.

src/object/interrupt.c
chrisguikema: Done.

src/arch/riscv/machine/hardware.c
```diff
@@ -102,6 +102,10 @@ static inline irq_t getActiveIRQ(void)
          * has claimed it in a multicore system.
          */
         irq = plic_get_claim();
+#ifdef CONFIG_PLAT_QEMU_RISCV_VIRT
+         // unmatched writes to claim are ignored by hardware
```
axel-h, line 106: Can you extend the comment to mention the counterpart in `invokeIRQHandler_AckIRQ()`, like the comment there mentions this. Then we have the connection, in case QEMU's behavior changes in future version.

chrisguikema: Done.

src/arch/riscv/common_riscv.lds
axel-h: I've created https://github.com/seL4/seL4/issues/972 as a follow-up task to investigate this. So it's not necessarily a merge blocker to have this conditional code for now.

src/arch/riscv/kernel/boot.c
axel-h: I've created https://github.com/seL4/seL4/issues/972 as a follow-up task to investigate this. So it's not necessarily a merge blocker to have this conditional code for now.

src/arch/riscv/common_riscv.lds
chrisguikema: Thanks, Axel!

src/arch/riscv/kernel/boot.c
chrisguikema: Thanks, Axel!

chrisguikema:
Just wanted to ping this PR to make sure it didn't get stale. @Ivan-Velickovic have you had a chance to take another look?

axel-h:
This looks mostly ok from my side. I just wonder where `ROCKETCHIP_ZCU102` actually comes from. Is this running against a released version from https://github.com/chipsalliance/rocket-chip that we could mention in a comment in `rocketchip/config.cmake` or `overlay-rocketchip-zcu102.dts` That would at least give some pointers until somebody add some details how to build and run this are at https://docs.sel4.systems/Hardware/rocketchip.html

src/plat/rocketchip/config.cmake
```diff
@@ -9,21 +10,49 @@ cmake_minimum_required(VERSION 3.7.2)
...
-    # the status of support for external interrupts via a PLIC is unclear and
-    # may differ depending on the version that is synthesized. Declaring no
-    # interrupts and using the dummy PLIC driver seems the best option for now
-    # to avoid confusion or even crashes.
-    declare_default_headers(
-        TIMER_FREQUENCY 10000000 PLIC_MAX_NUM_INT 0
-        INTERRUPT_CONTROLLER drivers/irq/riscv_plic_dummy.h
-    )
+    if(KernelPlatformRocketchipZCU102)
+        config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "rocket-fpga-zcu102")
```
axel-h, line 37: Mabye add the comment here that needs the OpenSBI fork from https://github.com/bao-project/opensbi/tree/bao/rocket then, as it seem not mainlines yet.

chrisguikema: Good suggestion; I'll add a link.

chrisguikema:
> This looks mostly ok from my side. I just wonder where `ROCKETCHIP_ZCU102` actually comes from. 

I didn't generate the bitstream myself, but one of my co-workers did. I think @Furao or @FTGStudio may have more insight. Basically, we added the h-extensions to the mainline rocketchip and generated a bitstream compatible with the ZCU102. My plan is to release the raw bitstream with instructions such that anyone with a ZCU102 can grab the bitstream and get it up-and-running quickly.

Because there were differences with the hardware resources (memory location, interrupt controller, etc), I created the "new platform" to keep the existing rocketchip support from changing.

Furao:
Chris is pretty close. We are using the `hyp` branch of this fork of rocket-chip which has hyp extensions applied: https://github.com/josecm/rocket-chip. This branch is at some point after v1.6. We also apply some patches to the mainline chipyard to add the ZCU102 as a supported platform: https://github.com/ucb-bar/chipyard/tree/64632c8aeef5386b4840e324be4429c0ceaac635 

We basically replicated the steps here: https://github.com/bao-project/bao-demos

chrisguikema:
My plan is to update the docs site with a PR once this is approved that points to a pre-built image, as well as the steps required to replicate the image.

axel-h:
Thanks @Furao and @chrisguikema. Can you add all these comment into the code/script, so we have this documented at least in a place where somebody picking up this patch will find it easily.
I consider the QEMU fix is something to merge in an case. It might be worth bringing this into a separate PR to merge this quickly (assuming there is an agreement to merge this). CI testing should also show that this works fine with the RISC-V hardware that is there.
For the rest, seems it's stuck where some PRs that add new platform get stuck. Needs some input from @lsf37 now on how to proceed.

chrisguikema:
I've added comments in the code such that the discussion points made here aren't lost. I've also created a rocketchip-zcu102.md file which I can make a PR for in the docs repo. This document overviews how we generated the bitstream such that others can reproduce it.

chrisguikema:
I created a docs PR here: https://github.com/seL4/docs/pull/175

src/object/interrupt.c
```diff
@@ -134,8 +134,16 @@ exception_t decodeIRQHandlerInvocation(word_t invLabel, irq_t irq)
...
 {
 #ifdef CONFIG_ARCH_RISCV
+#if !defined(CONFIG_PLAT_QEMU_RISCV_VIRT)
+    /* QEMU has a bug where interrupts must be
+     * immediately claimed, which is done in getActiveIRQ. For other
+     * platforms, the claim can wait and be done here.
+     */
     plic_complete_claim(irq);
+#endif
+    return;
```
kent-mcleod, line 144: Can this extra return statement please be removed as it's changing the AST of verified code and causing the preprocess tests to fail.

kent-mcleod:
> Differentiates the base Rocketchip platform with a version generated for the ZCU102 FPGA fabric

The description for the base platform is also for a ZCU FPGA fabric (https://docs.sel4.systems/Hardware/rocketchip.html)? Why is this a different platform? Is it just a different version of the bitstream with particular extensions? If this is the case, then maybe the config names could be changed to better reflect this?

chrisguikema:
> The description for the base platform is also for a ZCU FPGA fabric

The ZC706 is a zynq7000 development kit. The ZCU102 is a zynqmp development kit. I'm not an FPGA guy, but I assume there are differences in the FPGA fabric which make the two incompatible

src/plat/rocketchip/config.cmake
```diff
@@ -9,21 +10,57 @@ cmake_minimum_required(VERSION 3.7.2)
...
-    # to avoid confusion or even crashes.
-    declare_default_headers(
-        TIMER_FREQUENCY 10000000 PLIC_MAX_NUM_INT 0
-        INTERRUPT_CONTROLLER drivers/irq/riscv_plic_dummy.h
-    )
+    # The Rocketchip-ZCU102 is a softcore instantiation that runs on the ZCU102's
+    # FPGA fabric. Information on generating and running seL4 on the platform can
+    # be found at https://docs.sel4.systems/Hardware/
+    if(KernelPlatformRocketchipZCU102)
+        # The rocket-fpga-zcu104 platform can be found at the following git repo:
```
axel-h, line 40: So, is it zcu102 or zcu104 now?

```suggestion
        # The rocket-fpga-zcu102 platform can be found at the following git repo:
```

src/plat/rocketchip/overlay-rocketchip-zcu102.dts
```diff
@@ -0,0 +1,35 @@
...
+	serial@ff010000{
+		compatible = "cdns,uart-r1p12";
+		status = "okay";
+		interrupts = <2>;
+		interrupt-parent = <&L0>;
+		reg = <0x0 0xff010000 0x0 0x1000>;
+		clock-names = "uart_clk", "pclk";
+		clocks = <&uartclk>, <&uartclk>;
+	};
+};
```
axel-h, line 35: Could you add a linebreak in the last line?

src/plat/rocketchip/config.cmake
chrisguikema: The repo we forked has it as `rocket-zcu-104`. 

https://github.com/dornerworks/opensbi/tree/bao/rocket/platform/rocket-fpga-zcu104

chrisguikema: But it works without issues on the ZCU102 hardware

src/plat/rocketchip/overlay-rocketchip-zcu102.dts
chrisguikema: Done.

src/plat/rocketchip/config.cmake
axel-h: Ok, maybe add this also to the text, so this is clear.

kent-mcleod (approved):

PR #970, opened 2023-02-06
support loongarch and loongson 3A5000 platform

ChenYangng:
Port seL4 to loongarch refer to risc-v version and run on the loongson 3A5000 platform.

Test with: sel4test
Run on the qemu and real machines.Passed the same 112 test cases as risc-v/spike.

[This is an environment for running loongarch bios and OS on X86 machines](https://github.com/foxsen/qemu-loongarch-runenv)
Corresponding pull requests for sel4runtime, musllibc, seL4_libs, util_libs, seL4_tools and sel4test will also be provided.


.github/workflows/compilation-checks.yml
```diff
@@ -19,15 +19,18 @@ jobs:
...
           - arch: RISCV64
             compiler: llvm
+          # llvm Loongarch64 compilation is not currently supported
+          - arch: Loongarch64
+            compiler: llvm
     steps:
-    - uses: actions/checkout@v3
-    - uses: seL4/ci-actions/standalone-kernel@master
+    - uses: actions/checkout@v2
+    - uses: tyyteam/la-seL4-ci-actions/standalone-kernel@master
```
lsf37, line 33: We cannot approve workflow runs with actions from other repositories. Please revert those change before we look closer at this PR (also in cparser.yml and push.yml).

lsf37 (changes requested):

lsf37:
Thanks for publishing this massive contribution. Supporting an entire new architecture will require a discussion and decision of the technical steering committee of the seL4 foundation. The [RFC process](https://docs.sel4.systems/processes/rfc-process.html) doesn't specifically mention new architectures, but this change is definitely large enough to warrant an RFC and wider community discussion first before we start reviewing technical details.

The RFC template is geared towards API changes, but the main concerns I can see that the committee might have are:
 - What does support for the new architecture add to the value of seL4?
 - What are the plans for verification for this architecture? If there are none, should the architecture still be supported by the foundation? Why?
 - What maintenance and other burdens does the contribution add to the code and the resources of the foundation? In particular:
   - What are the plans for longer-term maintenance? Who will fix things when they break and are there enough people knowledgable about loongarch in the seL4 community to keep maintaining the code once the original authors have moved on?
   - Will the authors of the contribution be around for a longer time (5+ years) to answer community questions and help people get up and running?
 - What hardware is available for the architecture and how can it be supported in the CI infrastructure?

If you could address these in the RFC, it'll help the discussion.

axel-h:
Since there is QEMU support for this architecture in the latest version, there is an option to add this to CI at least, so we could have a run to see sel4test passing. If we get CI into a state, where this could run for forks and PRs (if explicitly enabled) this might  make a nice Github CI workflow improvement for the seL4 ecosystem. Given the RFC process to add the architecture in seL4 mainline might take longer, this CI work could be a practical technical step towards it. But then again, it takes somebody to drive this task.

I have to admit that this view is biassed by my (naive) attempt to unify the boot/trace across architectures. ARM and RISC-V are already similar, RISC-V is basically an ARM fork. Loongarch seems another copy/paste fork of RISC-V now. We could avoid a lot of code duplication in sone places...

lsf37:
> Since there is QEMU support for this architecture in the latest version, there is an option to add this to CI at least, so we could have a run to see sel4test passing. If we get CI into a state, where this could run for forks and PRs (if explicitly enabled) this might make a nice Github CI workflow improvement for the seL4 ecosystem. 

Difficult. The way it is done here does not work, because pointing to different actions gives those actions access to secrets of the seL4 github org.

Pointing to different actions in your own fork is completely fine, of course, and pointing to different actions from an in-repo branch is also fine, because if you can push to an in-repo branch you already have access to those secrets. Those are the mechanisms I currently use to test extensions to the CI infrastructure.

In full generality, I can't see a way around this -- changing CI infrastructure need a very high trust bar, because it has high privilege access.

That doesn't mean we couldn't figure out a few specific mechanisms like adding to the simulation action. I don't have any bright ideas, though, esp since these run from docker, so you actually need to be able to deploy the docker container to change anything that the seL4 org runs.

xen0n:
@ChenYangng I think in this case you should simply upstream the CI support first to https://github.com/seL4/ci-actions, then resume work on this PR.

lsf37:
> @ChenYangng I think in this case you should simply upstream the CI support first to https://github.com/seL4/ci-actions, then resume work on this PR.

That is probably the best order once we have decided to go forward, but I would not recommend more PRs before going through the RFC process.

xen0n:
> > @ChenYangng I think in this case you should simply upstream the CI support first to https://github.com/seL4/ci-actions, then resume work on this PR.
> 
> That is probably the best order once we have decided to go forward, but I would not recommend more PRs before going through the RFC process.

Of course. I'm just a passer-by that happen to know and have done some LoongArch work. Let's wait for the original author to submit the RFC first then...

lsf37:
@ChenYangng I'm closing this PR for now, because it needs an RFC first. We can reopen it when the RFC process is finished.

PR #974, opened 2023-02-28
Expose more files for external tools via `cmake --install`

nspin:
When the kernel is built as a top-level CMake project, the kernel and libsel4 can be installed using `cmake --install`. This PR adds some more files to that installation which can be useful to components surrounding the kernel, such as userspace bindings to the seL4 API in languages other than C.

Specifically, these files are:

- `sel4{,-arch,-sel4arch}.xml` and `syscall.xml`: Syscall and object interface definition files.
- `*.pbf`: Preprocessed `.bf` files.
- `kernel.dtb`: The device tree generated and used for the kernel build.
- `platform_gen.yaml`: Generated from `kernel.dtb` and used to generate `platform_gen.h`.

This PR includes the addition of a new subdirectory of the installation directory named `support/` which includes `kernel.dtb` and `platform_gen.yaml`. Perhaps there is a more descriptive name for this directory. Alternatively, these files could just be installed to the top-level of the installation directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -737,6 +737,23 @@ if("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_CURRENT_SOURCE_DIR}")
         DESTINATION libsel4/include
         FILES_MATCHING
         PATTERN "*.h"
+        PATTERN "*.pbf"
+        PATTERN "api/syscall.xml"
+        PATTERN "api/syscall.xsd"
+        PATTERN "interfaces/sel4.xml"
+    )
+    # Manually resolve conflict between the two files name sel4arch.xml
+    install(
+        FILES
+            "${CMAKE_CURRENT_SOURCE_DIR}/libsel4/arch_include/${KernelArch}/interfaces/sel4arch.xml"
+        DESTINATION libsel4/include/interfaces
+        RENAME sel4-arch.xml
+    )
+    install(
+        FILES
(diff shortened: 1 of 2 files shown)


nspin:
This PR is related to https://github.com/seL4/seL4/pull/975.

kent-mcleod:
> `sel4{,-arch,-sel4arch}.xml` and `syscall.xml`: Syscall and object interface definition files.

What would the impact be of preserving the source names of these files as they currently are, and using the install step of the build to rename them to the new naming scheme you are proposing? This would avoid unintentionally breaking any existing tools that are already trying to parse the xml files remotely.  I don't have a particularly strong sense of how to reason through the cost-tradeoffs of breaking these links though.

ratmice:
I do have some tools which currently have to hard code these paths, I don't mind updating them (in general the tools need to evolve as the specification does).
I haven't looked too closely at the timing/dependencies involved between the .yaml output in your other patch, and here.
I assume that the contents of the yaml output depends upon the contents of the source xml, as such it wouldn't be a good place to store paths to the XML.  I wonder if something (perhaps) a spec.yaml containing the static information about files in the repository in a more neutral way.  Anyhow that is one alternative that comes to mind here.

nspin:
> What would the impact be of preserving the source names of these files as they currently are, and using the install step of the build to rename them to the new naming scheme you are proposing?

I've modified this PR to do that instead. I think this is a good place to start, as it makes the XML available with `cmake --install` without breaking anything. I can make a separate PR with the breaking change once this one has landed. How does that sound?

lsf37 (approved):
This looks good from my side. @kent-mcleod are you happy as well?

kent-mcleod (approved):

PR #975, opened 2023-02-28
Generate `gen_config.yaml` before `gen_config.h`

nspin:
With this PR, instead of directly accumulating the contents of `gen_config.h` , `helpers.cmake` accumulates the contents of a YAML file called `gen_config.yaml`, from which it then derives `gen_config.h`. The benefit of this approach is that `gen_config.yaml`  can then serve a language-independent means of accessing the kernel configuration.

This PR includes ideas from https://github.com/seL4/seL4/pull/893.

I chose YAML over JSON because generating a YAML dictionary one entry at a time is easier than doing so in JSON, and the build system already depends on PyYAML.


nspin:
This PR is related to https://github.com/seL4/seL4/pull/974.

Ivan-Velickovic:
Example config for those interested:
```
ARM_HIKEY_OUTSTANDING_PREFETCHERS: "0"
ARM_HIKEY_PREFETCHER_STRIDE: "0"
ARM_HIKEY_PREFETCHER_NPFSTRM: "0"
ARM_HIKEY_PREFETCHER_STBPFDIS: false
ARM_HIKEY_PREFETCHER_STBPFRS: false
PLAT_IMX7: false
USER_TOP: "0xa0000000"
ARCH_AARCH32: false
ARCH_AARCH64: true # KernelSel4ArchAarch64=ON
ARCH_ARM_HYP: false
ARCH_RISCV32: false
ARCH_RISCV64: false
ARCH_X86_64: false
ARCH_IA32: false
SEL4_ARCH: "aarch64"
ARCH_ARM: false
ARCH: "arm"
ARCH_ARM: true # KernelArchARM=ON
WORD_SIZE: "64"
ARM_PLAT: "tx2"
MACH: "nvidia"
PLAT_BCM2711: false
PLAT_BCM2837: false
PLAT_FVP: false
PLAT_HIKEY: false
PLAT_IMX8MQ_EVK: false
PLAT_IMX8MM_EVK: false
PLAT_MAAXBOARD: false
PLAT_ODROIDC2: false
PLAT_ODROIDC4: false
PLAT_QEMU_ARM_VIRT: false
PLAT_QUARTZ64: false
PLAT_ROCKPRO64: false
PLAT_TQMA8XQP1GB: false
PLAT_TX1: false
PLAT_TX2: true # KernelPlatformTx2=ON
PLAT_ZYNQMP: false
PLAT: "tx2"
ARM_CORTEX_A7: false
ARM_CORTEX_A8: false
ARM_CORTEX_A9: false
ARM_CORTEX_A15: false
ARM_CORTEX_A35: false
ARM_CORTEX_A53: false
ARM_CORTEX_A55: false
ARM_CORTEX_A57: true # KernelArmCortexA57=ON
ARM_CORTEX_A72: false
ARCH_ARM_V7A: false
ARCH_ARM_V7VE: false
ARCH_ARM_V8A: true # KernelArchArmV8a=ON
ARM_SMMU: true # KernelArmSMMU=ON
AARCH64_SERROR_IGNORE: true # KernelAArch64SErrorIgnore=ON
ARM_MACH: "nvidia"
KERNEL_MCS: false
ARM_PA_SIZE_BITS_40: false
ARM_PA_SIZE_BITS_44: true # KernelArmPASizeBits44=ON
ARM_ICACHE_VIPT: ""
DEBUG_DISABLE_L2_CACHE: false
DEBUG_DISABLE_L1_ICACHE: false
DEBUG_DISABLE_L1_DCACHE: false
DEBUG_DISABLE_BRANCH_PREDICTION: false
ARM_HYPERVISOR_SUPPORT: true # KernelArmHypervisorSupport=ON
ARM_GIC_V3_SUPPORT: false
AARCH64_VSPACE_S2_START_L1: false
ARM_HYP_ENABLE_VCPU_CP14_SAVE_AND_RESTORE: false
ARM_ERRATA_430973: false
ARM_ERRATA_773022: false
ARM_SMMU: true # KernelArmSMMU=ON
TK1_SMMU: false
ENABLE_A9_PREFETCHER: false
EXPORT_PMU_USER: false
DISABLE_WFI_WFE_TRAPS: false
SMMU_INTERRUPT_ENABLE: false
AARCH32_FPU_ENABLE_CONTEXT_SWITCH: false
AARCH64_SERROR_IGNORE: false
L1_CACHE_LINE_SIZE_BITS: "6"
EXPORT_PCNT_USER: false
EXPORT_VCNT_USER: false
EXPORT_PTMR_USER: false
EXPORT_VTMR_USER: false
VTIMER_UPDATE_VOFFSET: true # KernelArmVtimerUpdateVOffset=ON
HAVE_FPU: true # KernelHaveFPU=ON
PADDR_USER_DEVICE_TOP: "17592186044416"
ROOT_CNODE_SIZE_BITS: "12"
TIMER_TICK_MS: "2"
TIME_SLICE: "5"
RETYPE_FAN_OUT_LIMIT: "256"
MAX_NUM_WORK_UNITS_PER_PREEMPTION: "100"
RESET_CHUNK_BITS: "8"
MAX_NUM_BOOTINFO_UNTYPED_CAPS: "50"
FASTPATH: true # KernelFastpath=ON
NUM_DOMAINS: "16"
SIGNAL_FASTPATH: false
NUM_PRIORITIES: "256"
MAX_NUM_NODES: "1"
ENABLE_SMP_SUPPORT: false
KERNEL_STACK_BITS: "12"
FPU_MAX_RESTORES_SINCE_SWITCH: "64"
VERIFICATION_BUILD: true # KernelVerificationBuild=ON
BINARY_VERIFICATION_BUILD: false
DEBUG_BUILD: false
HARDWARE_DEBUG_API: false
PRINTING: false
KERNEL_INVOCATION_REPORT_ERROR_IPC: false
NO_BENCHMARKS: true # KernelBenchmarksNone=ON
KERNEL_BENCHMARK: "none"
ENABLE_BENCHMARKS: false
KERNEL_LOG_BUFFER: false
MAX_NUM_TRACE_POINTS: "0"
IRQ_REPORTING: false
COLOUR_PRINTING: false
USER_STACK_TRACE_LENGTH: "0"
KERNEL_OPT_LEVEL_O2: true # KernelOptimisationO2=ON
KERNEL_OPT_LEVEL_OS: false
KERNEL_OPT_LEVEL_O0: false
KERNEL_OPT_LEVEL_O1: false
KERNEL_OPT_LEVEL_O3: false
KERNEL_OPT_LEVEL: "-O2"
KERNEL_OPTIMISATION_CLONE_FUNCTIONS: true # KernelOptimisationCloneFunctions=ON
KERNEL_FWHOLE_PROGRAM: false
DANGEROUS_CODE_INJECTION: false
DEBUG_DISABLE_PREFETCHERS: false
SET_TLS_BASE_SELF: false
CLZ_32: false
CLZ_64: false
CTZ_32: false
CTZ_64: false
CLZ_NO_BUILTIN: false
CTZ_NO_BUILTIN: false
```

lsf37:
Cool, this will also be useful for verification where I currently parse this info out of generated headers. Much nicer to consume yaml directly.

There seems to be an issue in the sel4test simulation build for riscv32 -- can you have a look at that?

ratmice:
I think the only preference I have for JSON over YAML here is the hope that eventually with newer versions of cmake, we could read JSON sources directly without the PyYAML dependency but that is kind of a minor thing I suppose and this is definitely an improvement either way.

tools/config_header_gen.py
```diff
@@ -0,0 +1,61 @@
...
+        type=argparse.FileType("r"),
+        default=sys.stdin,
+        help="Input YAML file.",
+    )
+    parser.add_argument(
+        "-o",
+        dest="out_file",
+        type=argparse.FileType("w"),
+        default=sys.stdout,
+        help="Output C header file.",
```
axel-h, line 27: What about a `--lang=<c|.....>¸ parameter?

nspin:
> There seems to be an issue in the sel4test simulation build for riscv32 -- can you have a look at that?

It seems that the bug causing that failure was triggered rather than caused by this PR. I've added a commit with the fix to this PR:

https://github.com/seL4/seL4/pull/975/commits/d619f13f6538dfe88b9df1846e43029a916a0dfc

nspin:
> I think the only preference I have for JSON over YAML here is the hope that eventually with newer versions of cmake, we could read JSON sources directly without the PyYAML dependency

Do you mean using phrases like `string(JSON user_top GET configure_string USER_TOP)` in CMake code? Are there cases that you have in mind where something like that would be more useful that just `${KernelUserTop}`?

tools/config_header_gen.py
nspin: Would that be to leave room for growth, or do you have specific alternative languages or formats in mind?

axel-h: Well, both actually. I have nothing specific really in mind, this would just come handy for some experimenting. There are some places already where tools implicitly assume it's C (or ARM), and I think it's good practice to make such choices explicit.

ratmice:
> > I think the only preference I have for JSON over YAML here is the hope that eventually with newer versions of cmake, we could read JSON sources directly without the PyYAML dependency
> 
> Do you mean using phrases like `string(JSON user_top GET configure_string USER_TOP)` in CMake code? Are there cases that you have in mind where something like that would be more useful that the usual `${KernelUserTop}`?

Yes, that is `string(JSON ...)` is exactly what I meant.  In my current external tools I don't actually have any python dependency, just cmake and a rust compiler for the host system pre-installed, while I could definitely add a crate dependency for yaml in rust it is something that then needs to be compiled and run before kernel build.  So it really just is that newer cmake already have a JSON parser built in while YAML needs an additional parser.

tools/config_header_gen.py
lsf37: Not sure that makes sense here -- the yaml file is supposed to be the interface for other languages, and this tool is the one that re-establishes the seL4 C headers from that. If we're adding support for other languages, it's not clear that we'd want to change this tool. We'd usually instead provide a separate tool that parses yaml (as we're doing for Rust), and it would not be in the kernel cmake build.

We could rename it slightly to make explicit that it is for C, but the help text already does say that.

lsf37:
Would it make sense to emit both `yaml` and `json` at some later point when it is easy for us to raise the minimum `cmake` version?

tools/helpers.cmake
```diff
@@ -396,9 +390,9 @@ function(config_string optionname configname doc)
         if(CONFIG_UNQUOTE)
             set(quote "")
         else()
-            set(quote "\"")
+            set(quote "\\\\\\\\\\\\\"")
```
kent-mcleod, line 393: Can you add a comment explaining the escaping calculation that's going on here?

tools/config_header_gen.py
```diff
@@ -0,0 +1,61 @@
...
+                entry = f"#define {macro}  1"
+            else:
+                entry = f"/* disabled: {macro} */"
+        elif isinstance(value, str):
+            if value:
+                entry = f"#define {macro}  {value}"
+            else:
+                entry = f"#define {macro}  /* empty */"
+        else:
+            raise Exception(
```
kent-mcleod, line 52: Is this supposed to result in the exact same transformation from original CMake config definitions and values to C header definitions?

tools/config_header_gen.py
axel-h: Point taken. I'd argue that it might be a chance to enforce unification of tools. But it's hard to argue given the python script can already be used amost as a library. Maybe just rework it, so the `main()` calls a function with parameters then?

tools/helpers.cmake
axel-h: Second to that, I had a similar thought counting the slashes.

tools/config_header_gen.py
nspin: Yes. I chose to preserve the same per-line format of `gen_config.h`.

kent-mcleod (approved):
I'm happy with this change

nspin:
>  In my current external tools I don't actually have any python dependency, just cmake and a rust compiler for the host system pre-installed

@ratmice Ah, I thought you were referring to the PyYAML dependency in the kernel's build system itself. I think you're right that JSON might be a better choice for the final language-independent output format. `gen_config.json` is essential enough that it might be read in all sorts of circumstances, including directly by build system code itself. Nix is another example of a build system DSL with built-in support for JSON but not YAML.

One consideration is that JSON doesn't allow comments, and the current `gen_config.h` has some generated comments on the CamlCase CMake variables related to certain entries, but the consensus from https://github.com/seL4/seL4/pull/893 seems to be that those comments are not only unnecessary but perhaps also undesirable.

> Would it make sense to emit both yaml and json at some later point when it is easy for us to raise the minimum cmake version?

What do you think about a change to this PR where the Python script emits C and JSON, and the YAML version of the file is relegated to a kernel build system implementation detail? The install target would install the JSON and not the YAML.

lsf37:
> What do you think about a change to this PR where the Python script emits C and JSON, and the YAML version of the file is relegated to a kernel build system implementation detail? The install target would install the JSON and not the YAML.

I'd be fine with that.

In general, I do prefer yaml to json because I find it more readable and I like comments, but it is true that this is intended for tool input, not for human input.

> When string(JSON ...) becomes more widely available, we could consider using that internally instead of YAML. A note for the future is that, if we decided to do that, then I think we should do so in a way that does not require every downstream project to change set(configuration_string "") to set(configuration_string "{}") .

Yes, that would be good.

ratmice:
> > What do you think about a change to this PR where the Python script emits C and JSON, and the YAML version of the file is relegated to a kernel build system implementation detail? The install target would install the JSON and not the YAML.
> 
> I'd be fine with that.

Seems reasonable to me.

tools/helpers.cmake
nspin: I came up with a better solution that involves less fighting with CMake:

https://github.com/seL4/seL4/pull/975/files#diff-67838857cc14e477be77851c70d3611d8c215fe0861678ec948e1b0859aade77R393
https://github.com/seL4/seL4/pull/975/files#diff-67838857cc14e477be77851c70d3611d8c215fe0861678ec948e1b0859aade77R393

tools/config_header_gen.py
nspin: The change mentioned in https://github.com/seL4/seL4/pull/975#issuecomment-1449088299 has resulted in a structure for the script that matches what you are describing, @axel-h.

tools/helpers.cmake
kent-mcleod: Is the second URL supposed to be to a different change line?

nspin: Oops, thanks! Now it points to the correct line.

kent-mcleod: Thanks, now I understand what the change is accomplishing!

axel-h (approved):

lsf37 (approved):
Also happy with this.

lsf37:
It looks like we did introduce a problem with this PR, although I don't quite understand yet how.

Some (but not all) builds are not finding the file `gen_config.h` any more. For instance, in the verification build I get:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/test-runner/ver/l4v/spec/cspec/c/config-build/ARM/gen_config/kernel/gen_config.h'
```

In the [rumprum repo](https://github.com/seL4/rumprun-sel4-demoapps/actions/runs/4308571361/jobs/7515001758#step:3:1721):

```
/github/workspace/projects/rumprun/platform/sel4/pci/rumpdma.c:14:10: fatal error: rumprun/gen_config.h: No such file or directory
     14 | #include <rumprun/gen_config.h>
        |          ^~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
  nbmkdep: compile failed.
```

As far as I could see from the diff, we wouldn't expect the file to have moved, and we'd still expect it to be generated if we compile the kernel. So I'm not quite sure where the failure comes from.

@nspin or @kent-mcleod any insight?

On the verification side, I want to change the setup anyway to read the json file instead, so we don't need to worry about that too much, but it'd still be good to understand what is going on.

nspin:
> It looks like we did introduce a problem with this PR, although I don't quite understand yet how.

My guess is that it has to do with undeclared dependency relationships that went unnoticed before due to chance. While I was preparing this PR, CI reported a similar looking issue, which I fixed with this commit (part of this PR):

https://github.com/seL4/seL4/pull/975/commits/7d86920d29d0f7df25d3867d9b2e572ad4a298bc

I will investigate further.

lsf37:
I should add that the verification config build just invokes `cmake`, i.e.:

```
 cmake -C ${CONFIG} \
                -DCMAKE_TOOLCHAIN_FILE=${CSPEC_DIR}/c/no-compiler.cmake \
                ${KERNEL_CMAKE_EXTRA_OPTIONS} \
                -G Ninja ${SOURCE_ROOT}
```

which produces `yaml` files, but not the C header and the json file.

lsf37:
If I run `ninja gen_config/kernel/gen_config.json` afterwards I do get what I need.

nspin:
> The main difference appears to be that the .h file used to be available after cmake directly, because it was produced from cmake, whereas now that needs an additional build step. I think that's fine.

Ah, I see. Before, that code was accessing `gen_config.h` before any ninja invocation. A few thoughts on this:

- If `string(JSON ...)` ever replaces the YAML implementation detail, then the JSON will be available before any ninja invocation. Furthermore, it would seem that this CMake feature would make it easy to also produce `gen_config.h` within CMake itself, but `string(JSON ...)` does not appear to provide a way to enumerate keys. Anyways, it sounds like the code in question will switch away from using `gen_config.h` anyways.
- Right now, `gen_config.yaml` is available before any CMake invocation. However, if it were to be used outside of the kernel project, then it becomes more than an implementation detail. I think it would be a little bit weird to expose both YAML and JSON. Could this be a reason to switch back to just YAML?

kent-mcleod:
> My guess is that it has to do with undeclared dependency relationships that went unnoticed before due to chance. While I was preparing this PR, CI reported a similar looking issue, which I fixed with [this commit](https://github.com/seL4/seL4/pull/975/commits/7d86920d29d0f7df25d3867d9b2e572ad4a298bc) (part of this PR):

The rumprun project does depend on the header files existing after CMake generation. https://github.com/seL4/rumprun/blob/master/CMakeLists.txt#L164-L171.  Further down in that file it should be possible to add a target dependency.

kent-mcleod:
> Right now, `gen_config.yaml` is available before any CMake invocation. However, if it were to be used outside of the kernel project, then it becomes more than an implementation detail. I think it would be a little bit weird to expose both YAML and JSON. Could this be a reason to switch back to just YAML?

I thought about the consequence of .h files not existing before ninja invocations while reviewing and concluded that it's probably not an actual requirement but only an implementation detail.   CAmkES is the main tool in the ecosystem that does a lot of processing during CMake invocations and I think even it doesn't need to depend on config header files being generated already.

nspin:
> The rumprun project does depend on the header files existing after CMake generation. https://github.com/seL4/rumprun/blob/master/CMakeLists.txt#L164-L171. Further down in that file it should be possible to add a target dependency.

This does this trick:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -364,6 +364,8 @@ CreateRumprunBuildCommand(
     ${RUMPRUN_INCLUDE_FILES}
     ${RUMPRUN_SEL4LIBS}
     muslc
+    sel4_autoconf
+    rumprun_Config
     OUTPUT_FILES
     ${CMAKE_CURRENT_BINARY_DIR}/${rumprun_sel4_arch}/sel4-obj/rumprun-intermediate.o
 )
@@ -384,6 +386,8 @@ CreateRumprunBuildCommand(
     ${SRC_NETBSD_FILES}
     ${RUMPRUN_SEL4LIBS}
     muslc
+    sel4_autoconf
+    rumprun_Config
 )
 
 CreateRumprunBuildCommand(
@@ -413,6 +417,8 @@ CreateRumprunBuildCommand(
     ${SRC_NETBSD_FILES}
     ${RUMPRUN_SEL4LIBS}
     muslc
+    sel4_autoconf
+    rumprun_Config
 )
```

I'll open a PR.

lsf37:
Ok, all tests are green now. It would probably be good to add a log entry to `CHANGES` that points out that you now need a `ninja gen_config/kernel/gen_config.h` after `cmake` to get the generated headers if your build relied on them being available after just `cmake`.

PR #977, opened 2023-03-01
Unclobber interface XML files

nspin:
The following groups of files conflict when the `arch_include` and `sel4_arch_include` directories are overlaid:

- `arch_include/*/interfaces/sel4arch.xml`
- `sel4_arch_include/*/interfaces/sel4arch.xml`

This PR renames them to:

- `arch_include/*/interfaces/sel4-arch.xml`
- `sel4_arch_include/*/interfaces/sel4-sel4arch.xml`

Now, when the `include`, `arch_include`, and `sel4_arch_include` directories are overlaid, we are left with:

- `interfaces/sel4.xml`
- `interfaces/sel4-arch.xml`
- `interfaces/sel4-sel4arch.xml`

This PR breaks other projects which expect to find these files at their current locations. In https://github.com/seL4/seL4/pull/974, we decided to split this breaking change into its own PR.


ratmice:
For the record the code that I'll need to update for this is at https://github.com/ratmice/sel4_rust_tools/blob/main/sel4_xml_types/src/invocations.rs#L416-L427 this is just tests, but generally the call sites also will need updating.  I'm happy to do so if/when this patch is merged.

axel-h:
From my side changing this is ok. Please add a note in https://github.com/seL4/seL4/blob/master/CHANGES about this breaking change.

kent-mcleod:
I don't think we need to make this change as in my opinion the motivation for this change, (simplifying the kernel XML installation automation steps), doesn't justify the costs (moving files that may or may not have external references we can't be aware of until they break for someone/something). 

Are there any other benefits I may be missing?

nspin:
I'm not sure that this counts towards justifying the costs, but I'm thinking about external tools that might be run both inside and outside of a CMake context. If tools could rely on finding these files by searching for them in the libsel4 include directories, then they could work the same way in either case:

- `-I$<JOIN:$<TARGET_PROPERTY:sel4,INCLUDE_DIRECTORIES>, -I>` inside CMake
- `-I$SEL4_INSTALL_PREFIX/libsel4/include` outside of CMake

As a concrete example, the `sel4-sys` crate I've been working on finds these XML files at build time by searching the colon-separated list of include directories passed via the environment variable `SEL4_INCLUDE_DIRS`. With a change like this, this crate can work in a consistent way both inside and outside of a CMake context:

- `SEL4_INCLUDE_DIRS=$<JOIN:$<TARGET_PROPERTY:sel4,INCLUDE_DIRECTORIES>,:> cargo ...` inside CMake
- `SEL4_INCLUDE_DIRS=$SEL4_INSTALL_PREFIX/libsel4/include cargo ...` outside of CMake

Without a change like this, I would resort to something like allowing the expected locations of the XML files to be overridden by environment variables like `SEL4_INTERFACE_XML`, `SEL4_INTERFACE_ARCH_XML`, and `SEL4_INTERFACE_SEL4_ARCH_XML`. This override feature would be used inside of CMake but not outside of CMake.

Obviously this isn't a big problem, but I just thought I'd illustrate why I thought to propose this change.

nspin:
Updated to use improved filenames merged in https://github.com/seL4/seL4/pull/1106.

I also added a note in CHANGES.

lsf37:
@kent-mcleod what do you think of the additional motivation Nick gave?

kent-mcleod:
The original mistake was made when `sel4_arch_include` was added and we didn't choose a different name than `sel4arch.xml`.

Are there any options that could be considered for creating symlinks under the old names that redirect to the new names, and then deleting the old symlinks at some future date after a reasonable deprecation period?

nspin:
I've just added the symlinks, in a separate commit.

nspin:
I think the `PR / Style (pull_request)` failure is due to this PR being the first to change certain files since their formatting standards have changed.

lsf37:
> I think the `PR / Style (pull_request)` failure is due to this PR being the first to change certain files since their formatting standards have changed.

Correct, I'll push a separate commit for that.

lsf37 (approved):

PR #979, opened 2023-03-02
ARM_SMMU: Fix configuration handling for SMMU opt

kent-mcleod:
The TX2 is currently the only platform that supports CONFIG_ARM_SMMU and wants it on by default, however the CMake scripts wouldn't allow it to be disabled. Now it should work normally:
- When TX2 is set the config is enabled unless it's preconfigured to OFF.
- When the platform is changed to not TX2, the SMMU option is turned off.


Xaphiosis:
I'm seeing a hiccup here. If you compare other platforms against tx2 when building with hyp and smmu both off, you do indeed get this, so the cmake part seems to go ok:
```
imx8mq-evk/CMakeCache.txt
751:KernelArmSMMU:INTERNAL=OFF
752:KernelArmSMMU_DISABLED:INTERNAL=TRUE

tx2/CMakeCache.txt
285:KernelArmSMMU:BOOL=OFF
```

Note tx2 doesn't get the `KernelArmSMMU_DISABLED:INTERNAL`, but I don't know if that means anything.

However, the smmu code is still looks to be generated:
```
    seL4_CapSMMUSIDControl = 12,                                                 
    seL4_CapSMMUCBControl = 13,
    
etc.
```

kent-mcleod:
> However, the smmu code still looks to be generated:

Are there more definition examples? The definitions in the example you provide don't appear to be #ifdef gated and appear on all configurations at the moment.

lsf37 (approved):
Looks good.

kent-mcleod:
Added a commit adding #ifdef guards for an smmuv2.c file if CONFIG_ARM_SMMU isn't set. This is necessary as the hardware_gen.py script picks out the compatibility string for `/iommu@12000000` that's referenced in the fdt by seL4,kernel-devices for the tx2 platform. This results in the smmuv2.c file being included into the set of input C files for the kernel unconditionally. This follows the same pattern as for the platform serial drivers that are still selected for a release build, but #ifdef'd out during preprocessing if CONFIG_PRINTING is not enabled.

lsf37:
Thanks, that explains why Raf was still seeing SMMU code in there.

Xaphiosis (approved):
Thank you, that was really fast.

Xaphiosis:
@lsf37 why aren't we seeing a preprocess failure? The code definitely changed

kent-mcleod:
> Can confirm the only SMMU lines on all platforms from sel4test with SMMU disabled now look the same:

We should consider removing these too... I left them alone for now as that's a public header.  These constants reserve slots in the system's initial CSpace. Several slots are already arch/feature specific, but this is inconsistent with how other constants are allocated such as for invocation labels.

kent-mcleod:
> @lsf37 why aren't we seeing a preprocess failure? The code definitely changed

You probably want to turn CONFIG_ARM_SMMU off in the verified config now

Xaphiosis:
> > @lsf37 why aren't we seeing a preprocess failure? The code definitely changed
> 
> You probably want to turn CONFIG_ARM_SMMU off in the verified config now

Cool, I'll make that a separate PR after this one goes in.

Xaphiosis:
Preprocess failure looks good! I'll leave the systems side to you.

lsf37 (approved):
Cool, thanks for the adjustments. 

As Raf says, the preprocessor failure is what we want. The style failure is also fine to ignore, that's just the fact that the verification `.cmake` files are executable, which we do want.

Will merge once the build has finished.

PR #980, opened 2023-03-02
Include change to MinSchedContextBits in CHANGES

Ivan-Velickovic:
The motivation for this is that not everything uses the the seL4 API C headers for constants like `MinSchedContextBits` (e.g. seL4 Core Platform). Anything depending on the value may want to be aware of a change in the constant.


lsf37 (approved):
Thanks!

Ivan-Velickovic:
(just fixing DCO)

PR #938, opened 2022-10-24
mcs: correct MinSchedContextBits and add compile assert

michaelmcinerney:
Signed-off-by: Michael McInerney <michael.mcinerney@proofcraft.systems>

PR #981, opened 2023-03-02
more seL4_MinSchedContextBits docs

lsf37:
The `seL4_MinSchedContextBits` size change in #938 is breaking and can be hard to debug. Provide some guidance.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -36,7 +36,7 @@ Upcoming release: BREAKING
   compliant.
 * Remove imx31/kzm platform support. This platform is being removed as it is sufficiently old and unused.
 * Remove ARM1136JF_S and ARMv6 support. This architecture version is being removed as it is sufficiently old and
-  unused. See <https://sel4.atlassian.net/browse/RFC-8.>
+  unused. See [RFC-8](https://sel4.atlassian.net/browse/RFC-8).
 * Remove ARMv6 specific configs: KernelGlobalsFrame and KernelDangerousCodeInjectionOnUndefInstr. This removes the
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -26,17 +26,17 @@ Upcoming release: BREAKING
...
+* Added support for the Avnet MaaXBoard
+* Added support for arm_hyp on qemu-arm-virt platfrom with cortex-a15 CPU
+* Added support for qemu-riscv-virt
+* Rename libsel4 config option ENABLE_SMP_SUPPORT to CONFIG_ENABLE_SMP_SUPPORT to be namespace compliant.
+* Rename libsel4 config option AARCH64_VSPACE_S2_START_L1 to CONFIG_AARCH64_VSPACE_S2_START_L1 to be namespace
+  compliant.
 * Remove imx31/kzm platform support. This platform is being removed as it is sufficiently old and unused.
 * Remove ARM1136JF_S and ARMv6 support. This architecture version is being removed as it is sufficiently old and
-  unused. See https://sel4.atlassian.net/browse/RFC-8.
+  unused. See <https://sel4.atlassian.net/browse/RFC-8.>
```
axel-h, line 39: Including the final full stop seem wrong. Also, why apply the `<...>` style here for links, shouldn't this be something like `[RFC-8](https://sel4.atlassian.net/browse/RFC-8).` to match the markdown style for the other links in this file?

lsf37: Yes, you're right, thanks for spotting that, the `.` should definitely be outside.

And `[RFC-8](https://sel4.atlassian.net/browse/RFC-8)` is probably nicer, I just removed the linter warnings I got. Will update.

kent-mcleod (approved):

PR #982, opened 2023-03-02
configs: Ensure compatibility with binary verification

mbrcknl:
For verified configurations on which we perform binary verification, ensure that we disable certain whole-program optimisations that are incompatible with binary verification.


lsf37 (approved):
Looks good.

lsf37:
@mbrcknl you might need to kick this with a dummy push to get the actions running. They don't seem to automatically re-trigger after the outage today, and since they were never created there is not manual thing I can do to them in the UI.

lsf37:
(Ok to ignore the gitlint failure on file mode -- we do want these to be executable)

PR #983, opened 2023-03-03
make physBase a static inline function

Xaphiosis:
(note that this PR will need some work on the verification side as well, but I want to get it past kernel engineers first, as I hold out a last gasp of hope there may be a less painful way to do this, or at least one less offensive to human life)

For verification flexible w.r.t kernel placement in physical memory, we need to relate physBase as a named constant to its abstract equivalent. Unfortunately, apart from enums, the C programming language does not have real constants. The C parser follows the C standard and requires enums constants to be storable as int, meaning without major overhaul enums are not sufficient for storing word_t-sized memory addresses.

Since the linker scripts can't deal with static inline functions in the constants they need (KERNEL_ELF_BASE and KERNEL_ELF_PADDR_BASE), we provide the following preprocessor definitions for the linker specifically:
* physBase_raw (the numerical value returned by physBase())
* KERNEL_ELF_BASE_RAW
* KERNEL_ELF_PADDR_BASE_RAW


tools/hardware/outputs/c_header.py
```diff
@@ -27,7 +27,7 @@
 
 #pragma once
 
-#define physBase {{ "0x{:x}".format(physBase) }}
+#define physBase_raw {{ "0x{:x}".format(physBase) }}
```
yyshen, line 30: shall we make it uppercase, as the usual macros?

yyshen (approved):

tools/hardware/outputs/c_header.py
axel-h: Yes, that's a good idea.

include/arch/arm/arch/32/mode/hardware.h
```diff
@@ -54,7 +54,7 @@
 
 /* The first physical address to map into the kernel's physical memory
  * window */
-#define PADDR_BASE physBase
+#define PADDR_BASE physBase()
```
axel-h, line 57: Can't we drop `PADDR_BASE`then completely and use `physBase()` (or better `get_physBase()`) everywhere. It's very confusing in the C code when thing that look like macro constants turn out to be function. And they can't be used in macros that build values needed at compile time.

tools/hardware/outputs/c_header.py
```diff
@@ -36,6 +36,14 @@
 #include <linker.h>         /* for BOOT_RODATA */
 #include <basic_types.h>    /* for p_region_t, kernel_frame_t (arch/types.h) */
 
+/* wrap raw physBase location constant to give it a symbolic name in C that's
```
axel-h, line 39: I think it would be helpful if this comment is much more verbose and contains some of the things you wrote in the pull request comment. Then it is more clear what's behind this without digging it up in the commit history.

tools/hardware/outputs/c_header.py
```diff
@@ -36,6 +36,14 @@
...
+ * visible to verification
+ */
+static inline word_t physBase(void)
```
axel-h, line 42: We could state in the interface explicitly that this is supposed to return a constant.
```suggestion
static inline CONST word_t physBase(void)
```
The compiler should be smart enough to figure this out from the code in the body, so I'm not sure if there is much difference. However, it allows the compiler to raise some warning if for whatever reason the code in the body does not produce a constant.

lsf37: adding CONST makes sense, yes

tools/hardware/outputs/c_header.py
Xaphiosis: - [ ] rename to PHYS_BASE_RAW

include/arch/arm/arch/32/mode/hardware.h
Xaphiosis: `get_physBase` does not meet naming seL4 conventions. `getPhysBase` indicates that we are obtaining it from some state, not really a fan of this option

tools/hardware/outputs/c_header.py
Xaphiosis: - [ ] expand comment

tools/hardware/outputs/c_header.py
Xaphiosis: - [ ] add CONST

include/arch/arm/arch/32/mode/hardware.h
lsf37: I agree, `getPhysBase()` sounds like it is *not* a constant.

Xaphiosis: regarding PADDR_BASE, the concept appears on most (all?) arches, shows up in memory map comments and in verification. I'd be reluctant to mess with that as by-the-way to this PR

Xaphiosis:
Addressed most issues, responded to others, rebased. Will squash once people are satisfied with the result.

lsf37:
I've been looking at the assembly code that gets generated out of this change, just to confirm that it does actually get inlined and we don't get unwanted function calls etc.

The result is: the inlining does work, and I get the same immediate values that are loaded. Unexpectedly, it does actually seem to lead to some completely unrelated optimisations. It eliminates a few more lds/str in locations that have nothing to do with this change. I don't think it'll be observable in speed anywhere (I saw most of these in boot code), but it's good to see that if at all the binary is a tiny bit smaller.

kent-mcleod (approved):

Xaphiosis:
Squashed.

lsf37 (approved):
Also good from my side.

tools/hardware/outputs/c_header.py
```diff
@@ -36,6 +36,15 @@
 #include <linker.h>         /* for BOOT_RODATA */
 #include <basic_types.h>    /* for p_region_t, kernel_frame_t (arch/types.h) */
 
+/* Wrap raw physBase location constant to give it a symbolic name in C that's
+ * visible to verification. This is necessary as there are no real constants
+ * in C except enums, and enums constants must fit in an int.
+ */
+static inline CONST word_t physBase(void)
```
axel-h, line 43: So, can't we use this?
```
const word_t physBase = PHYS_BASE_RAW;
```

lsf37: No, that will put a symbol in memory. The function will instead be compiled out like before.

axel-h: Ok, thanks for confirming.

tools/hardware/outputs/c_header.py
```diff
@@ -36,6 +36,15 @@
...
+{
+    return PHYS_BASE_RAW;
+}
+
```
axel-h, line 47: It seems all this affects 32-bit platforms only. On 64-bit architectures we define PADDR_BASE to be UL_CONST(0), so this is not used at all. Should we have `#define PADDR_BASE physBase()' here to ensure this is consistent everywhere? Or is this all something to be buried in the 32-bit world?

lsf37: For PADDR_BASE, it does seem to affect 32 bit only, but it affects KERNEL_ELF_PADDR_BASE and with that KERNEL_ELF_BASE and `addrFromKPPtr` on 64 bit platforms as well.

kent-mcleod: physBase comes from defining the base physical address of the kernel virtual mapping. If the kernel virtual mapping domain is greater or equal to the range of physical addresses, then having the base of the virtual domain map to the lowest physical address seems simplest.

When the physical address range is larger than the kernel virtual mapping (which on 32bit it is) then we can only map to a subset of the physical address range. 

If we expect that on all 64bit architectures the physical address range will fit within the virtual mapping space then physBase would only ever need to be 0. But who knows if this will always be the case as sufficiently complex devices start requiring larger address ranges for encoding more information than just a spatial address into the physical address.

lsf37: Ok, got it, that makes perfect sense, yes. I guess we can change it when we get to a device that needs it. For now it seems reasonable to keep 0 here.

PR #984, opened 2023-03-03
ci proof: Upload build artifacts

mbrcknl:
For proof workflow runs on seL4 PRs, upload kernel build artifacts generated by the aws-proofs action. These can be used to run binary verification, although we currently still require binary verification to be manually triggered.


lsf37 (approved):
👍

Issue #941, opened 2022-10-27
MCS+SMP: Missing remoteQueueUpdate somewhere?


Indanz:
Setup:
- Passive server with bound notification.
- The notification has a scheduling context bound to CPU core 2. (This is the only task running on core 2.)
- All IPC callers are on core 1.

The passive server gets stuck when it is woken up by a signal while it was on core 1. I suspect when it just replied to a call.

A register dump shows the server and all callers are stuck in the system call, the server in `ReplyRecv`, the callers in `Call`.

Adding a `remoteQueueUpdate()` before the `break` in the `SysReplyRecv` case of `handleSyscall()` solves the problem for me (with fastpath disabled for easier debugging).

I'll continue debugging tomorrow and update this issue when more is known.

Anyone any idea what might be up?

kent-mcleod:
> Anyone any idea what might be up?

In `receiveSignal` when `NtfnState_Active`, `maybeDonateSchedContext()` will call migrateTCB, but the TCB will never be inserted into any scheduler queues on the other core. So there's missing calls to SCHED_APPEND for triggering the schedule on the other core, and also to rescheduleRequired() for rescheduling on the current core. It doesn't appear that this scenario is being handled by the kernel at the moment.  We need to add a test case for this: when ksCurThread is passive, receiving a endpoint or notification with a sc for a different core will migrate and reschedule the thread on the other core.

Indanz:
Thanks for looking into this.

`maybeDonateSchedContext()` calls:
- `schedContext_donate`, which assigns the SC to the TCB (and does migrateTCB).
- `schedContext_resume`, which may postpone the SC if it can't run now.

It may make more sense to do the migrateTCB in the resume instead of the donate, as well as possibly any new check.

Though I don't think my case comes via `receiveSignal`, but via `receiveIPC`, which calls `completeSignal` if a signal is pending, but from there on the code is almost the same (there is a subtle difference between the sporadic check of `NtfnState_Active` compared to the one in `completeSignal`).

Indanz:
The code in `schedContext_bindTCB` is relevant too. It bugs me that all callers of `schedContext_resume` handle this case slightly differently (like calling enqueue and rescheduleRequired in different orders).

Should we add the following check somewhere or is something else needed?
```
if (isSchedulable(target)) {
    possibleSwitchTo(target);
}
```

Indanz:
Perhaps:

```
diff --git a/src/object/notification.c b/src/object/notification.c
[...]

PR #986, opened 2023-03-09
MCS: Fix passive server core migration during ReplyRecv

JE-Archer:
Fixes an issue where passive servers are not scheduled after migrating to another core during ReplyRecv. If during the handling of the receive part of ReplyRecv, the current thread migrates to a remote core, it will be inserted into the scheduling queues of the remote core, but no reschedule IPI will be sent to the remote core. In the worst case, if the remote core is running the idle thread, the passive server may never be scheduled. This commit fixes this by sending a reschedule IPI after the current thread is enqueued, if necessary.

Note that while handling the receive part of the syscall, while the current thread will receive the caller's scheduling context and have its tcbAffinity set to the remote core, it will not be enqueued on the remote core. It is only during the call to schedule() after handling the syscall that the current thread, the affinity of which is now the remote core, will be enqueued on the remote core. However, this will only insert the TCB into the queue, it will not send a reschedule IPI to the remote core. If the remote core is running the idle thread, it will continue running the idle thread even if the kernel is entered via an interrupt, and even though the passive server's TCB is in the core's ready queues, because its scheduler action will not change from ResumeCurrentThread (the idle thread). This behaviour can be reliably reproduced with seL4test test IPC0028 (which was disabled for me, so double-check that it actually runs).

There are multiple ways that this issue could be fixed. For example, the enqueue+IPI could be done during the handling of the syscall. However this would probably necessitate changing the scheduling code anyway, as it will still call SCHED_ENQUEUE_CURRENT_TCB. For now I have chosen the simple fix of adding an IPI where the enqueue is done.

See also #941 for more discussion.

Test with: https://github.com/seL4/sel4test/pull/123

### DIFF
--- a/include/object/tcb.h
+++ b/include/object/tcb.h
@@ -69,13 +69,24 @@ void remoteTCBStall(tcb_t *tcb);
     remoteQueueUpdate(_t);          \
 } while (0)
 
+
+#define SCHED_ENQUEUE_CURRENT_TCB do {                                    \
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex()); \
+    tcbSchedEnqueue(NODE_STATE(ksCurThread));                             \
+} while (0)
+#define SCHED_APPEND_CURRENT_TCB do {                                     \
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex()); \
+    tcbSchedAppend(NODE_STATE(ksCurThread));                              \
+} while (0)
+
 #else
 #define SCHED_ENQUEUE(_t)           tcbSchedEnqueue(_t)
 #define SCHED_APPEND(_t)            tcbSchedAppend(_t)
-#endif /* ENABLE_SMP_SUPPORT */
 
 #define SCHED_ENQUEUE_CURRENT_TCB   tcbSchedEnqueue(NODE_STATE(ksCurThread))
 #define SCHED_APPEND_CURRENT_TCB    tcbSchedAppend(NODE_STATE(ksCurThread))
+#endif /* ENABLE_SMP_SUPPORT */
+
 
 #ifdef CONFIG_KERNEL_MCS
 /* Add TCB into the priority ordered endpoint queue */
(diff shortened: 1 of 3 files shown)


gernotheiser:
Can the kernel check what prio is running on the destination core? It would be good to only interrupt if the migrating thread should actually preempt what's currently running

Indanz:
The problem I have with this solution is that conceptually it doesn't make sense: `ksCurThread` should always be running on the current core. This fix relies on `SCHED_ENQUEUE_CURRENT_TCB`/`SCHED_APPEND_CURRENT_TCB` being called before `ksCurThread` gets updated. It introduces subtle ordering assumptions which can be easily broken in the future when innocuous unrelated changes are made.

I am also not convinced it solves all corner cases that trigger this bug. I don't have access to MEP's hardware any more where this could be easily triggered, so can't test if this solves the problem I originally ran into, but I doubt it.

I think we have to make a table of all permutations of when tasks can be woken up cross-core and make sure there are tests for all cases. Any fix we come up with can then be properly tested.

Indanz:
> I agree that it is weird conceptually that a thread on a remote core is enqueued on the remote core during a call to schedule() on the current core, and especially weird that it is done through a call to `SCHED_ENQUEUE_CURRENT_TCB`, where the "current TCB" is understood to mean the TCB running on the current core. However that is the way the code currently works with SC donation across cores during ReplyRecv.

If that's indeed what happens now, then that it works at all seems pure coincidence.

> Maybe we could enqueue the current thread on the other core during handleRecv and then set NODE_STATE(ksCurThread) to NULL, and adapt schedule() to handle NULL ksCurThread, or something along those lines?

I believe that `schedule` should only be concerned with scheduling on the current core, and should not be involved with cross-core scheduling at all, neither directly or indirectly.

We should not forget to update the fastpath code too, including the signal one that was added recently and is default off.

> If you can describe the scenarios I can try to reproduce them. What sort of hardware are we talking about?

See #941 for the scenario. The hardware is the tqma board, which has an i.MX8 Soc, but I meant more the software stack than the actual hardware itself. With MEP's software it was easy to do load testing with many tasks and different scenario's. sel4test is good for testing specific scenario's, but not good for SMP load testing where lots of things happen concurrently to shake out any concurrency bugs.

lsf37:
> with some invariant like NODE_STATE(ksCurThread) == NULL || NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex().

If at all possible `ksCurThread` should never be NULL and always point to a valid TCB. Everything else would seriously mess with a lot of invariants we currently have about the current thread.

PR #976, opened 2023-03-01
aarch32 bit support for rpi4

kent-mcleod:
Add support for rpi4b running in 32bit mode.

PR #990, opened 2023-03-22
qemu: avoid QEMU build/run crashing for 32-bit systems.

axel-h:
Use device tree overlay as a smart hack to avoid crashes if 32-bit QEMU is misconfigured or has >=4 GiB RAM.

This cope with the current inability of the seL4 build scripts to handle a full 4 GiB address space without overflows. It handles configurations where a 32-bit QEMU has RAM beyond the 4 GiB border (by accident or design) by simply reserving all memory beyond `0xffffffff`. It avoids that the seL4 build scripts generate regions addresses >= `0x100000000`, which break compilation, because such values don't fit into 32-bit variables used for physical addresses on 32-bit systems.
The long term solution should be a fix in the builds scripts, so a full 4 GiB address space is usable and everything beyond 4 GiB is ignored.

### DIFF
--- a/src/plat/qemu-arm-virt/config.cmake
+++ b/src/plat/qemu-arm-virt/config.cmake
@@ -135,6 +135,13 @@ if(KernelPlatformQEMUArmVirt)
                 # 2 GiB, which seems a good trade-off nowadays. It's sufficient
                 # for test/demo systems, but still something the host can
                 # provide without running short on resources.
+                # The memory starts at 1 GiB (0x40000000), so up to 3 GiB can be
+                # accessed before exceeding the 32-bit address space. For 32-bit
+                # systems, using memory beyond this point is non-trivial. While
+                # the LPAE MMU model supports accessing up to a 1 TiB (40-bit)
+                # physical address space even on 32-bit systems, the 32-bit
+                # version of seL4 can access physical addresses in the 32-bit
+                # range only.
                 set(QEMU_MEMORY "1024")
             endif()
 
@@ -218,6 +225,10 @@ if(KernelPlatformQEMUArmVirt)
     endif()
 
     list(APPEND KernelDTSList "${QEMU_DTS}" "${CMAKE_CURRENT_LIST_DIR}/overlay-qemu-arm-virt.dts")
+
+    if(KernelSel4ArchAarch32)
+        list(APPEND KernelDTSList "${CMAKE_CURRENT_LIST_DIR}/overlay-qemu-arm-virt32.dts")
+    endif()
 
     if(KernelArmHypervisorSupport OR KernelSel4ArchArmHyp)
         list(APPEND KernelDTSList "${CMAKE_CURRENT_LIST_DIR}/overlay-reserve-vm-memory.dts")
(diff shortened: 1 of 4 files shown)


lsf37 (commented):
In general I'm fine with this change, but the comments are a bit misleading.

It  is not the build scripts that prevent accessing physical memory > 4GB on 32 bit platforms, it is seL4 itself because it uses 32 bit values for physical addresses, i.e. it currently doesn't support the 40 bit extensions. Of course the build scripts also don't support larger physical addresses, but the root cause is not that (the build scripts are probably much easier to fix).

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -135,6 +135,13 @@ if(KernelPlatformQEMUArmVirt)
...
                 # for test/demo systems, but still something the host can
                 # provide without running short on resources.
+                # The memory starts at 1 GiB (0x40000000), so up to 3 GiB can be
+                # accessed before exceeding the 32-bit address space. For 32-bit
+                # systems, using memory beyond this point is non-trivial. While
+                # the LPAE MMU model supports accessing up to a 1 TiB (40-bit)
+                # physical address space even on 32-bit systems, the 32-bit
+                # version of seL4 can access physical addresses in the 32-bit
+                # range only.
                 set(QEMU_MEMORY "1024")
```
kent-mcleod, line 145: Is it possible to instead generate an error at build time if QEMU_MEMORY >= 3072 && KernelSel4ArchAarch32?

axel-h: That can be done when changing the scripts. But I think it would be nice to have 3072 working, so it's more intuitive what can be given there. Purpose of this PR really is not touching the scripts.

kent-mcleod: Isn't it more likely that the new overlay does unexpected things with reserving regions outside of the 32bit region.  It would be simpler to just reject a Qemu memory configuration that the kernel doesn't support when CMake is processing.

kent-mcleod: Actually, I think it might be a good idea to try and use the overlay method you suggest.  I can reuse the same approach for https://github.com/seL4/seL4/pull/976/files

Ivan-Velickovic:
What is the status of this?

axel-h:
> What is the status of this?

Could be merged to address the issue I have described.

kent-mcleod (approved):

kent-mcleod:
> > What is the status of this?
> 
> Could be merged to address the issue I have described.

I think @lsf37's initial feedback still needs to be addressed.

axel-h:
> I think @lsf37's initial feedback still needs to be addressed.

Changes the wording. It is better now?

Indanz (approved):

Ivan-Velickovic:
@axel-h can you rebase this? We can merge this and then https://github.com/seL4/ci-actions/pull/234 as well.

lsf37 (approved):

lsf37:
> > I think @lsf37's initial feedback still needs to be addressed.
> 
> Changed the wording. It is better now?

Yes, happy with the wording now!

axel-h:
> @axel-h can you rebase this? We can merge this and then [seL4/ci-actions#234](https://github.com/seL4/ci-actions/pull/234) as well.

Done

Issue #564, opened 2021-09-09
VSpace vs PageDirectory / GlobalPageDirectory


bennoleslie:
I'm sure I won't get the terminology exactly right here, but let's try anyway.

There seems to be a bit of confusion (for me at least) between the top-level paging structure (PageDirectory on aarch32, GlobalPageDirectory on aarch64) and VSpace.

It seems like these are, more or less, the same thing. And/or that VSpace is an alias for the other.

But things are a bit confusing when it comes to the documentation. For example, on aarch32 the cache operations are documented as being methods on the seL4_ARM_PageDirectory object, while for aarch64 the cache operations are documented as being on the seL4_ARM_VSpace object.

Is this just a documentation thing, or are these actually different between 32 vs 64-bit?

kent-mcleod:
`seL4_ARM_VSpace` is aarch64 specific where different kernel configurations can result in either 4 levels or 3 levels of page table hierarchy.  Because the kernel objects are all named differently for different levels, seL4_ARM_VSpace was added as an alias of whatever object is at the top level. This was only made for aarch64.

lsf37:
In general, I would tend to call the top-level page table structure that is currently associated with a TCB the VSpace object of that TCB. 

I think this mostly needs cleanup up in the docs (i.e. maybe just adding a proper definition).For some architectures (RISCV) being "top-level" only happens when it is assigned to an ASID pool, and for others it can be different object types.

PR #994, opened 2023-03-25
manual: improvements to VSpace section

lsf37:
- clearer distinction between VSpace and VSpace object
- give example and consistent definition for VSpace object (addresses #564) -- it was defined before already, but a bit hidden
- more consistent terminology in the rest of VSpace section, esp wrt to cap vs object
- use `\obj` consistently when referring to kernel objects (previously mix of `\ob` and `\texttt`)

### DIFF
--- a/manual/parts/vspace.tex
+++ b/manual/parts/vspace.tex
@@ -6,50 +6,52 @@
 
 \chapter{\label{ch:vspace}Address Spaces and Virtual Memory}
 
-A virtual address space in seL4 is called a VSpace. In a similar
-way to a CSpace (see \autoref{ch:cspace}), a VSpace is composed of objects
-provided by the microkernel. Unlike CSpaces, these objects for managing
-virtual memory largely correspond to those of the hardware. Consequently,
-each architecture defines its own objects for the top-level VSpace and further intermediate paging structures.
-Common to every architecture is the \obj{Page}, representing a frame of physical memory.
-The kernel also includes \obj{ASID Pool} and
-\obj{ASID Control} objects for tracking the status of address spaces.
+A virtual address space in seL4 is called a VSpace. Similarly to a CSpace (see \autoref{ch:cspace}),
+a VSpace is composed of objects provided by the kernel. Unlike CSpaces, objects for managing virtual
+memory correspond to those of the hardware and each architecture defines its own object types for
+paging structures. Also unlike CSpaces, we call only the top-level paging structure a VSpace object.
+It provides the top-level authority to the VSpace.
 
-These VSpace-related objects are sufficient to implement the
-hardware data structures required to create, manipulate, and destroy
-virtual memory address spaces. It should be noted that, as usual, the
-manipulator of a virtual memory space needs the appropriate
-capabilities to the required objects.
+Common to all architectures is the \obj{Frame}, representing a frame of physical memory. Frame
+objects are manipulated via \obj{Page} capabilities, which represents both authority to the frame,
+as well as to the virtual memory mapping, i.e.\ the page, when mapped. The kernel also provides
+\obj{ASID Pool} objects and \obj{ASID Control} invocations for tracking the status of address space
+identifiers for VSpaces.
+
+These VSpace-related objects are sufficient to implement the hardware data structures required to
+create, manipulate, and destroy virtual memory address spaces. As usual, the manipulator of a
+virtual memory space needs the appropriate capabilities to the required objects.
 
 \section{Objects}
 
 \subsection{Hardware Virtual Memory Objects}
 
 Each architecture has a top-level paging structure (level 0) and a number of intermediate levels.
-The top-level paging structure corresponds directly to the higher-level concept of a VSpace in seL4.
-For each architecture, the VSpace is realised as a different object, as determined by the
-architectural details.
+When referring to it generically, we call this top-level paging structure the VSpace object. The
+seL4 object type that implements the VSpace object is architecture dependent. For instance on
+AArch32, a VSpace is represented by the PageDirectory object and on x64 by a PML4 object.
 
-In general, each paging structure at each level contains slots where the next level paging structure,
-or a specifically sized frame of memory, can be mapped. If the previous level is not mapped,
-a mapping operation will fail. Developers need to manually create and map all paging structures.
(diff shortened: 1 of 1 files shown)


manual/parts/vspace.tex
```diff
@@ -100,8 +99,8 @@ \subsubsection{AArch32}
 
 \subsubsection{AArch64}
 
-Arm AArch64 processors have a four-level page-table structure, where the VSpace is realised as a
-\texttt{PageGlobalDirectory}. All paging structures are index by 9 bits of the virtual address.
+Arm AArch64 processors have a four-level page-table structure. The VSpace object is implemented by the
+\obj{PageGlobalDirectory} object. All paging structures are indexed by 9 bits of the virtual address.
```
kent-mcleod, line 103: This is configuration dependent. On some aarch64 hypervisor configurations, the top level object is the PageUpperDirectory and there is no PageGlobalDirectory.  To help with this libsel4 defines the seL4_ARM_VSpaceObject alias that is a preprocessor macro that expands to seL4_ARM_PageUpperDirectoryObject or seL4_ARM_PageGlobalDirectoryObject.

lsf37: Excellent point, will add that. (I had only looked at what was already in the manual)

lsf37: Ok, done. Coincidentally, the reference to PageGlobalDirectory didn't actually exist in the genereated API docs, but is `group__aarch64__seL4__ARM__VSpace` instead. Added that change as well.

lsf37:
(rebased)

manual/parts/vspace.tex
```diff
@@ -86,12 +86,9 @@ \subsubsection{x64}
...
 
-Like IA-32, Arm AArch32 realise the VSpace as a \texttt{PageDirectory}, which covers the entire
-4\,GiB address range, and a second-level \texttt{PageTable}. The second-level structures on AArch32
-cover 1\,MiB address ranges.
-
-Arm AArch32 processors have a two-level page-table structure.
-The top-level page directory covers a range of 4\,GiB and each page table covers a 1\,MiB range.
+Like IA-32, Arm AArch32 implements the VSpace object with a \texttt{PageDirectory} object which
+covers the entire 4\,GiB address range.  The second-level structures on AArch32 are
+\texttt{PageTable} objects and cover 1\,MiB address ranges.
```
kent-mcleod: Technically on aarch32 hyp Page table objects cover 2MiB instead of 1MiB.

manual/parts/vspace.tex
```diff
@@ -257,11 +255,11 @@ \subsection{ASID Control}
...
-An \obj{ASID Pool} confers the right to create a subset of the available
-maximum applications. For a VSpace to be usable by an application, it
-must be assigned to an ASID. This is done using a capability to an
-\obj{ASID Pool}. The \obj{ASID Pool} object has a single method, \texttt{Assign}, for each
-architecture:
+An \obj{ASID Pool} confers the right to use a subset of the globally available address space
+identifiers. The size of this subset is architecture dependent. For a VSpace object to be usable by
+a thread, it must be assigned to an ASID via an \obj{ASID Pool} capability. Each ASID can be
+assigned to at most one VSpace. The \obj{ASID Pool} capability has a single invocation,
+\texttt{Assign}, for each architecture.
```
kent-mcleod, line 271: Is there a good reason why assigning a vspace to an asid pool doesn't require you to specify the slot index to use? The kernel can spend a long time searching for a free asid slot if it's always trying to find the next free one. And it's easy to reuse ASIDs unintentionally if you revoke an address space and then create a new one, it'll use the same slot I believe.

manual/parts/vspace.tex
```diff
@@ -316,14 +314,14 @@ \section{Mapping Attributes}
...
-address space. Attempting to map the same capability
-twice will result in an error.
+The seL4 kernel does not allow intermediate paging structures (e.g.\ \obj{PageTable} objects) to be
+shared, but it does allow pages to be shared between VSpaces, and VSpaces to be shared by threads.
+
+To share a page, the capability to the \obj{Page} must first be duplicated using the
+\apifunc{seL4\_CNode\_Copy}{cnode_copy} method and the copy must be used in the Map invocation
+(e.g.\ \apifunc{seL4\_ARM\_Page\_Map}{arm_page_map} \ifxeightsix or
+\apifunc{seL4\_x86\_Page\_Map}{x86_page_map}\fi) that maps the page into the second address space.
+Attempting to map the same capability twice will result in an error.
```
kent-mcleod, line 333: unless it's overwriting it's previous paging slot which is how page mappings are updated I think

kent-mcleod (approved):

manual/parts/vspace.tex
lsf37: > Is there a good reason why assigning a vspace to an asid pool doesn't require you to specify the slot index to use? 

Probably not, tbh. The topic has come up in the past in [SELFOUR-2363](https://sel4.atlassian.net/browse/SELFOUR-2364) and [SELFOUR-2364](https://sel4.atlassian.net/browse/SELFOUR-2363)

> The kernel can spend a long time searching for a free asid slot if it's always trying to find the next free one. And it's easy to reuse ASIDs unintentionally if you revoke an address space and then create a new one, it'll use the same slot I believe.

Yes, agreed. I don't think it would be terribly hard to verify that solution, but it would probably break a lot of user-space code. Then again, you could implement the same policy in a library that simulates the old interface.

manual/parts/vspace.tex
lsf37: You're right, even un-technically. I think this is just wrong and should be updated.

manual/parts/vspace.tex
lsf37: Have addressed that now by specifying the target (different tables or vspaces).

We should probably add a section/paragraph about remap somewhere, but I'll leave that for a separate PR.

manual/parts/vspace.tex
lsf37: Added the distinction between hyp and non-hyp.

lsf37:
PR feedback is addressed, will merge this now. There is more work that could/should be done on this section (mention/explain remapping), but that is for a future PR.

PR #998, opened 2023-03-27
clean up build of manual

lsf37:
- fix broken references in the manual (it looks like the build scripts are hiding these)
- make latex complain a bit less
- more recent doxygen versions declare some options obsolete that we're not using anyway. Remove these.
- prevent unnecessary rebuilds in the Makefile


lsf37:
Looks like I messed up in the Makefile tweaks. Setting to draft until I figure it out.

lsf37:
> Looks like I messed up in the Makefile tweaks. Setting to draft until I figure it out.

Ok, got it working smoothly now. Ready for review.

PR #999, opened 2023-03-27
introduce constant seL4_BootInfoFrameSize 

axel-h:
- Provide `seL4_BootInfoFrameSize` for userland, to there is no longer a need to hard-code the 4 KiByte assumption.
- Remove `SEL4_BI_FRAME_SIZE`

For usage examples see:
- seL4/seL4_libs#73
- seL4/capdl#49

### DIFF
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -8,6 +8,7 @@
 
 #include <sel4/config.h>
 #include <sel4/macros.h>
+#include <sel4/sel4_arch/constants.h>
 
 /* caps with fixed slot positions in the root CNode */
 enum {
@@ -75,12 +76,27 @@ typedef struct seL4_BootInfo {
      * to make this struct easier to represent in other languages */
 } seL4_BootInfo;
 
-/* If extraLen > 0, then 4K after the start of bootinfo there is a region of the
- * size extraLen that contains additional boot info data chunks. They are
- * arch/platform specific and may or may not exist in any given execution. Each
- * chunk has a header that contains an ID to describe the chunk. All IDs share a
- * global namespace to ensure uniqueness.
+/* The boot info frame must be large enough to hold the seL4_BootInfo data
+ * structure. Due to internal restrictions, the size must be of the form 2^n and
+ * the minimum is one page.
  */
+#define seL4_BootInfoFrameBits  seL4_PageBits
+#define seL4_BootInfoFrameSize  LIBSEL4_BIT(seL4_BootInfoFrameBits)
+
+SEL4_COMPILE_ASSERT(
+    invalid_seL4_BootInfoFrameSize,
+    sizeof(seL4_BootInfo) <= seL4_BootInfoFrameSize)
+
+/* If seL4_BootInfo.extraLen > 0, this indicate the presence of additional boot
+ * information chunks starting at the offset seL4_BootInfoFrameSize. Userland
+ * code often contains the hard-coded assumption that the offset is 4 KiByte,
+ * because the boot info frame usually is one page, which is 4 KiByte on x86,
+ * Arm and RISC-V.
+ * The additional boot info chunks are arch/platform specific, they may or may
+ * not exist in any given execution. Each chunk has a header that contains an ID
+ * to describe the chunk. All IDs share a global namespace to ensure uniqueness.
+ */
+
 typedef enum {
     SEL4_BOOTINFO_HEADER_PADDING            = 0,
     SEL4_BOOTINFO_HEADER_X86_VBE            = 1,
--- a/manual/parts/bootup.tex
+++ b/manual/parts/bootup.tex
@@ -118,14 +118,15 @@ of slots in the initial thread's CNode, starting with CPTR \texttt{start} and wi
   \end{center}
 \end{table}
 
-Depending on the architecture and platform there might be additional pieces of boot
-information. If \texttt{extraLen} is greater than zero, then 4K after the start of bootinfo
-is a region of extraLen bytes containing additional bootinfo structures. Each chunk starts
-with a \texttt{seL4\_BootInfoHeader}, described in \autoref{tab:bi_header_struct}, that
-describes what the chunk is and how long it is, where the length includes the header. The
-length can be used to skip over chunks that you do not understand. The only generally
-defined chunk type is \texttt{SEL4\_BOOTINFO\_HEADER\_PADDING} and describes an empty
-chunk that has no data, any other types are platform or architecture specific. The
+The size of the fixed Boot Info Frame is \texttt{seL4\_BootInfoFrameSize}. In the standard
+configuration, this is one page, which is 4 KiByte on x86, ARM and RISC-V. Depending on the
+architecture and platform, there might be additional pieces of variable boot information
+following afterwards. The overall size of this data is \texttt{extraLen}, it contains a
+sequence of blobs, where each one start with a \texttt{seL4\_BootInfoHeader} described in
+\autoref{tab:bi_header_struct}. This header describes what the blob is and how long it is,
+where the length includes the header. Thus, the length can be used to skip over unknown
+chunks. The only generally defined chunk type is \texttt{SEL4\_BOOTINFO\_HEADER\_PADDING}
+and describes a blob where any payload data exists for padding only. The
 \texttt{extraBIPages} slot region gives the frames capabilities for the pages that make up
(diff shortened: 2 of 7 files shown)


manual/parts/bootup.tex
```diff
@@ -118,14 +118,15 @@ \section{\label{ch:bootup:bootinfo}BootInfo Frame}
...
-Depending on the architecture and platform there might be additional pieces of boot
-information. If \texttt{extraLen} is greater than zero, then 4K after the start of bootinfo
-is a region of extraLen bytes containing additional bootinfo structures. Each chunk starts
-with a \texttt{seL4\_BootInfoHeader}, described in \autoref{tab:bi_header_struct}, that
-describes what the chunk is and how long it is, where the length includes the header. The
-length can be used to skip over chunks that you do not understand. The only generally
-defined chunk type is \texttt{SEL4\_BOOTINFO\_HEADER\_PADDING} and describes an empty
-chunk that has no data, any other types are platform or architecture specific. The
+The size of the fixed Boot Info Frame is \texttt{SEL4\_BI\_FRAME\_SIZE}. In the standard
+configuration, this is one page, which is 4 KiByte on x86, ARM and RISC-V. Depending on the
```
lsf37, line 122:
```suggestion
configuration, this is 4 KiByte on x86, Arm and RISC-V. Depending on the
```

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -75,11 +76,23 @@ typedef struct seL4_BootInfo {
...
 
-/* If extraLen > 0, then 4K after the start of bootinfo there is a region of the
- * size extraLen that contains additional boot info data chunks. They are
- * arch/platform specific and may or may not exist in any given execution. Each
- * chunk has a header that contains an ID to describe the chunk. All IDs share a
- * global namespace to ensure uniqueness.
+/* The fixed boot info frame size is one page. As the page size is 4 KiByte on
+ * x86, ARM and RISC-V, lots of userland code has a hard coded assumption that
+ * the boot info frame size is 4 KiByte, but it should use SEL4_BI_FRAME_SIZE
+ * actually.
```
lsf37, line 82:
```suggestion
/* The fixed boot info frame size, per default 4 KiByte on x86, Arm and RISC-V.
 * Instead of assuming the default, userland code should use the SEL4_BI_FRAME_SIZE constant.
```

lsf37 (approved):
I'd be happy with that. Two minor tweaks below.

manual/parts/bootup.tex
axel-h: Mentioning that it's a page was intentional so have an explanation for the magic 4 KiB value. We don't have any other architecture with different page sizes, where all this might get more interesting then...

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -75,11 +76,26 @@ typedef struct seL4_BootInfo {
...
- * size extraLen that contains additional boot info data chunks. They are
- * arch/platform specific and may or may not exist in any given execution. Each
- * chunk has a header that contains an ID to describe the chunk. All IDs share a
- * global namespace to ensure uniqueness.
+/* The boot info frame must be large enough to hold the seL4_BootInfo data
+ * structure. Due to internal restrictions, the size must be of the form 2^n and
+ * the minimum is one page.
+ */
+#define BI_FRAME_SIZE_BITS  seL4_PageBits
+#define SEL4_BI_FRAME_SIZE  LIBSEL4_BIT(BI_FRAME_SIZE_BITS)
```
kent-mcleod, line 84: This name should probably be seL4_BootInfoFrameSize to match the naming convention for other constants in libsel4. `BI_FRAME_SIZE_BITS` is a private kernel constant, and probably shouldn't be declared in libsel4 unless it was also named suitably.

axel-h: I will rename it to `seL4_BootInfoFrameSize`. And put another commit on top then that removes `BI_FRAME_SIZE_BITS` from the kernel code, as there seems no benefit keeping it as an alias.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -79,11 +80,26 @@ typedef struct seL4_BootInfo {
...
-/* If extraLen > 0, then 4K after the start of bootinfo there is a region of the
- * size extraLen that contains additional boot info data chunks. They are
- * arch/platform specific and may or may not exist in any given execution. Each
- * chunk has a header that contains an ID to describe the chunk. All IDs share a
- * global namespace to ensure uniqueness.
+/* The boot info frame must be large enough to hold the seL4_BootInfo data
+ * structure. Due to internal restrictions, the size must be of the form 2^n and
+ * the minimum is one page.
+ */
+#define seL4_BootInfoFrameSizeBits  seL4_PageBits
```
kent-mcleod, line 87: I think this name should just be `seL4_BootInfoFrameBits` to follow the Size/Bits naming pattern that the other constants follow. (Thoughts @lsf37?)

axel-h: Yes, makes sense. Will change this.

kent-mcleod: Thanks!

kent-mcleod (approved):

axel-h:
fixed typo

axel-h:
Can we merge this?

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -79,11 +80,26 @@ typedef struct seL4_BootInfo {
...
+
+/* If seL4_BootInfo.extraLen > 0, this indicate the presence of additional boot
+ * information chunks starting at the offset seL4_BootInfoFrameSize. Note that
+ * since x86, Arm and RISC-V use 4 KiByte pages and the boot info frame usually
+ * is one page, legacy userland code often contains the hard-coded assumption,
+ * that this offset is 4 KiByte. Unfortunately, seL4_BootInfo currently has no
+ * field that allows passing this offset. The additional boot is arch/platform
+ * specific, it may or may not exist in any given execution. Each chunk has a
+ * header that contains an ID to describe the chunk. All IDs share a global
+ * namespace to ensure uniqueness.
```
kent-mcleod, line 102: I don't see why this needed to be reworded to declare existing applications as legacy. The bootinfo region size is defined as 4096 as part of the boot ABI changing the size would be a breaking change.  User code should be able to keep using a page size assumption as we aren't going to change it. (The extraLen mechanism is for allowing expanding out past the 4k size limit).

axel-h: I will keep the wording then, there are compile time checks in place now that `CONFIG_MAX_NUM_BOOTINFO_UNTYPED_CAPS` does not get a value that breaks the size. And besides corner case experimental configurations that oage size assumption should always hold practically.

axel-h:
I thin I have addressed all remarks, can we merge this, and https://github.com/seL4/seL4_libs/pull/73, 
https://github.com/seL4/capdl/pull/49 now?

manual/parts/bootup.tex
lsf37: Ok, fair enough.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -75,12 +76,27 @@ typedef struct seL4_BootInfo {
...
+#define seL4_BootInfoFrameBits  seL4_PageBits
+#define seL4_BootInfoFrameSize  LIBSEL4_BIT(seL4_BootInfoFrameBits)
+
+SEL4_COMPILE_ASSERT(
+    invalid_seL4_BootInfoFrameSize,
+    sizeof(seL4_BootInfo) <= seL4_BootInfoFrameSize)
+
+/* If seL4_BootInfo.extraLen > 0, this indicate the presence of additional boot
+ * information chunks starting at the offset seL4_BootInfoFrameSize. Userland
+ * code often contains the hard-coded assumption, that the offset is 4 KiByte,
```
lsf37, line 92:
```suggestion
 * code often contains the hard-coded assumption that the offset is 4 KiByte,
```

lsf37 (approved):
Looking good to merge now, one final super minor comma tweak left.

libsel4/include/sel4/bootinfo_types.h
axel-h: done

axel-h:
> Looking good to merge now, one final super minor comma tweak left.

donr

PR #1001, opened 2023-03-31
machine: compile time checks for multikernel

lsf37:
Add compile time checks for conditions on physBase that are necessary for verification of multikernel builds to succeed -- if these fail, the proofs will fail.

If these succeed, and nothing else has changed compared to a verified kernel other than physBase, then the proofs will succeed. This does not mean that all platform requirements are validated, it just means that all requirements for the proofs to be consistent are met.

The conditions correspond to those in

    spec/machine/*/Arch_Kernel_Config_Lemmas.thy

in the verification repository. E.g. see [here for Aarch32](https://github.com/seL4/l4v/blob/master/spec/machine/ARM/Arch_Kernel_Config_Lemmas.thy).

### DIFF
--- a/include/arch/arm/arch/64/mode/hardware.h
+++ b/include/arch/arm/arch/64/mode/hardware.h
@@ -210,6 +210,7 @@
 /* The log buffer is placed before the device region */
 #define KS_LOG_PPTR (KDEV_BASE - UL_CONST(0x200000))
 
+#ifndef __ASSEMBLER__
 /* All PPTR addresses must be canonical to be able to be stored in caps or objects.
    Check that all UTs that are created will have valid address in the PPTR space.
    For non-hyp, PPTR_BASE is in the top part of the address space and device untyped
@@ -218,7 +219,6 @@
    overflow without going into address ranges that are non-canonical.  These static
    asserts check that the kernel config won't lead to UTs being created that aren't
    representable. */
-#ifndef __ASSEMBLER__
 compile_assert(ut_max_less_than_cannonical, CONFIG_PADDR_USER_DEVICE_TOP <= BIT(47));
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 compile_assert(ut_max_is_cannonical, (PPTR_BASE + CONFIG_PADDR_USER_DEVICE_TOP) <= BIT(48));
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -16,6 +16,24 @@
 #include <arch/sbi.h>
 #include <mode/machine.h>
 
+/* PPTR_BASE must be smaller than KERNEL_ELF_BASE. */
+compile_assert(pptr_base_less_elf_base, PPTR_BASE < KERNEL_ELF_BASE_RAW);
+/* physBase must be aligned to a page for verification to succeed. */
+compile_assert(phys_base_page_aligned, IS_ALIGNED(PHYS_BASE_RAW, seL4_PageBits));
+
+/* The following compile time checks are verification artifacts. Not necessarily
+   all systems need to satisfy these, but proofs will need to be changed
+   manually if they are not satisfied. These particular checks also are not
+   necessarily sufficient for all real platforms, but they are good sanity
+   checks to have always on. */
+
+/* PPTR_BASE must be at least twice as far away from KERNEL_ELF_BASE as a LargePage. */
+compile_assert(pptr_base_distance, PPTR_BASE + BIT(seL4_LargePageBits + 1) < KERNEL_ELF_BASE_RAW);
+/* Kernel ELF window must have at least 64k space */
+compile_assert(kernel_elf_distance, KERNEL_ELF_BASE_RAW + BIT(16) < KDEV_BASE);
+/* End of kernel ELF window must not overflow as a word_t */
+compile_assert(kernel_elf_no_overflow, KERNEL_ELF_BASE_RAW < KERNEL_ELF_BASE_RAW + BIT(16));
+
 /* Bit flags in CSR MIP/SIP (interrupt pending). */
 /* Bit 0 was SIP_USIP, but the N extension will be dropped in v1.12 */
 #define SIP_SSIP   1 /* S-Mode software interrupt pending. */
--- a/include/arch/arm/arch/32/mode/hardware.h
+++ b/include/arch/arm/arch/32/mode/hardware.h
@@ -89,10 +89,15 @@
 #ifndef __ASSEMBLER__
 /* It is required that USER_TOP must be aligned to at least 20 bits */
 compile_assert(USER_TOP_correctly_aligned, IS_ALIGNED(USER_TOP, 20));
+
 /* It is required on arm_hyp that USER_TOP isn't lower than the top GiB */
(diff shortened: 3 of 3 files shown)


lsf37:
For Arm, I managed to put the compile time assertions close to where the constants are defined in `xx/mode/hardware.h`, but for RISC-V, this didn't work, because `hardware.h` does not include enough other headers to express the checks.

If the discrepancy is too annoying, we could instead try to include the missing headers in `xx/mode/hardware.h` for RISC-V. Actually, now that I think of it, this is probably better, because I'm not sure the conditions make sense for 32 bit mode in RISC-V.

include/arch/riscv/arch/machine.h
```diff
@@ -16,6 +16,24 @@
...
+   manually if they are not satisfied. These particular checks also are not
+   necessarily sufficient for all real platforms, but they are good sanity
+   checks to have always on. */
+
+/* PPTR_BASE must be at least twice as far away from KERNEL_ELF_BASE as a LargePage. */
+compile_assert(pptr_base_distance, PPTR_BASE + BIT(seL4_LargePageBits + 1) < KERNEL_ELF_BASE_RAW);
+/* Kernel ELF window must have at least 64k space */
+compile_assert(kernel_elf_distance, KERNEL_ELF_BASE_RAW + BIT(16) < KDEV_BASE);
+/* End of kernel ELF window must not overflow as a uint64 */
+compile_assert(kernel_elf_no_overflow, KERNEL_ELF_BASE_RAW < KERNEL_ELF_BASE_RAW + BIT(16));
```
lsf37, line 35: Since this is currently included for both modes (32+64), the default word size is not necessarily 64, i.e. the comment needs to be updated if we leave this here. The check still makes sense on 32 bit systems (checking against 32 bit overflow), so if are happy leaving the check at this location, only the comment needs to change.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -218,9 +219,13 @@
...
    asserts check that the kernel config won't lead to UTs being created that aren't
    representable. */
-#ifndef __ASSEMBLER__
 compile_assert(ut_max_less_than_cannonical, CONFIG_PADDR_USER_DEVICE_TOP <= BIT(47));
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 compile_assert(ut_max_is_cannonical, (PPTR_BASE + CONFIG_PADDR_USER_DEVICE_TOP) <= BIT(48));
 #endif
+
+/* PPTR_BASE must be smaller than KERNEL_ELFBASE. KERNEL_ELF_BASE is unfolded below
+   because the compiler refuses to do so automatically. */
```
axel-h, line 228: Maybe "refuse" is too strong. `KERNEL_ELF_BASE` is built using `KERNEL_ELF_PADDR_BASE`, which isince 65a1b457a4ba3e72838434a68c5ce9889d3ce406 is the result of a function call `physBase()`. While this is a inlined const function, sadly in C such things are still not evaluated at compile time to be used in expressions then (zig can do this, maybe there is a gcc/clang switch also?). What we could do here is using `KERNEL_ELF_BASE_RAW` and state it the comment, why this is the better choice..

axel-h: Seems with c23 we could do change https://github.com/seL4/seL4/blob/master/tools/hardware/outputs/c_header.py#L43-L46 to use `constexpr word_t physBase = PHYS_BASE_RAW;`. Unfortunately `constexpr` for function is still not in the standard.

lsf37: I tried using `KERNEL_ELF_BASE_RAW` initially, but it was complaining about `PPTR_BASE_OFFSET` which is used in both. It's possible that the definition of `PPTR_BASE_OFFSET` just wasn't included yet at this point. Maybe I should have another look at that.

I don't think we want to switch to C23 yet at this point, although some features might be nice to cherry pick. In any case, not for this PR, though.

include/arch/riscv/arch/machine.h
```diff
@@ -16,6 +16,24 @@
 #include <arch/sbi.h>
 #include <mode/machine.h>
 
+/* PPTR_BASE must be smaller than KERNEL_ELFBASE. */
+compile_assert(pptr_base_less_elf_base, PPTR_BASE < KERNEL_ELF_BASE_RAW);
+/* physBase must be aligned to a page for verification to succeed. */
+compile_assert(phys_base_page_aligned, IS_ALIGNED(PHYS_BASE_RAW, seL4_PageBits));
```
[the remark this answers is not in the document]
lsf37: The specific condition is because the init state example in verification uses a page for the kernel ELF window, so in that sense a verification artifact. But I don't think it would make much sense to try to define a virtual address window that is not at least page aligned in any case, so IMO this condition should be checked regardless. (You could theoretically set the constant to a non-aligned value, but you're still going to have to map at least a page)

include/arch/arm/arch/64/mode/hardware.h
lsf37: Ok, yes including `<hardware.h>` does fix the problem and `KERNEL_ELF_BASE_RAW` then works (see additional commit). The generic `<hardware.h>` is the place where `PPTR_BASE_OFFSET` is defined. It is a bit strange to include it here, although it does appear to compile fine.

I'd be fine either way (unfolding `KERNEL_ELF_BASE_RAW` and `PPTR_BASE_OFFSET` here or including generic  `<hardware.h>` at the end).  @kent-mcleod do you have any preferences for this?

lsf37:
> Odroid_C4 and the two ZYNQ boards do not seem to fulfil these requirements (everything else does). Will have a look at what is going on there.

Ok, so what's happening at least on Odroid_C4 is that `KERNEL_ELF_BASE` ends up equal to `PPTR_BASE`, because `physBase` is 0. The diagram here:  https://github.com/seL4/seL4/blob/3161e7a563883ece423a61b5ea521c9337776031/include/arch/arm/arch/64/mode/hardware.h#L137-L154
is not quite clear on whether that is a situation that we allow or not.

There is probably nothing spectacularly wrong with having the ELF window at the very start of the kernel window, assuming that memory isn't used for objects. Will have to dig into the verification to figure out where exactly we needed that condition and if the proof can be changed to allow it.

lsf37:
>  Will have to dig into the verification to figure out where exactly we needed that condition and if the proof can be changed to allow it.

We can allow it, the condition can be dropped. Will update.

include/arch/arm/arch/64/mode/hardware.h
lsf37: The point has become irrelevant, because the check is not actually necessary for AArch64 (at least for the verified hyp config), so I'll remove the check and mark this as resolved.

lsf37:
> Odroid_C4 and the two ZYNQ boards do not seem to fulfil these requirements (everything else does). Will have a look at what is going on there.

Confirmed that the issue for all 3 boards is that `physBase == 0` and therefore `PPTR_BASE == KERNEL_ELF_BASE`, so nothing problematic in any of them.

kent-mcleod (approved):

include/arch/riscv/arch/machine.h
lsf37: Have updated the comment to `word_t`.

Issue #1002, opened 2023-04-02
In the RISC-V platform, there is a risk of integer overflow.


CtrlZ233:
https://github.com/seL4/seL4/blob/ec170f4c21cdd6c9e163c0da681c912a52b97b8a/src/kernel/boot.c#L682

In the seL4test project, using the configuration parameter for the Spike architecture, this line of code caused an integer overflow.

Printed code and results are shown below：
```c
   if (start < CONFIG_PADDR_USER_DEVICE_TOP) {
        region_t reg = paddr_to_pptr_reg((p_region_t) {
            start, CONFIG_PADDR_USER_DEVICE_TOP
        });
        printf("untyped_region: 0x%lx ... 0x%lx\n", reg.start, reg.end);
        if (!create_untypeds_for_region(root_cnode_cap, true, reg, first_untyped_slot)) {
            printf("ERROR: creation of untypeds for top device region"
                   " [%"SEL4_PRIx_word"..%"SEL4_PRIx_word"] failed\n",
                   reg.start, reg.end);
            return false;
        }
    }
```

```
untyped_region: 0xffffffc17ff00000 ... 0x4000000000
```

lsf37:
Thanks for spotting that. While `CONFIG_PADDR_USER_DEVICE_TOP ` and the other region data that is input to these functions are trusted input, it would still be good to sanity check it and fail early here instead of producing the overflow.

lsf37:
Ok, so looking further at this (thanks to @Ivan-Velickovic for pointing me in this direction), the overflow as such is actually completely fine, because we are trying to create device Untypeds which don't have to have valid pointers, they just need to store regions.

The only problem is that we don't want one contiguous region `0xffffffc17ff00000 ... 0x4000000000`. I think it should be creating two regions `0xffffffc17ff00000 ... 0xffffffffffffffff` and `0x0 ... 0x4000000000`. @kent-mcleod am I seeing this correctly?

lsf37:
Sorry, should have remembered this from the beginning: this is in fact completely fine, because the function this is passed to will deal gracefully with such regions and divide them up into aligned non-overflowing parts.

The thing that is missing is documentation for this fact.

axel-h:
Looks like we've been there already, see https://github.com/seL4/seL4/issues/389

CtrlZ233:
> Looks like we've been there already, see #389

Okay, I think I understand what happened. Thanks.

PR #1003, opened 2023-04-03
sanity checks for memory regions

lsf37:
Systematically add assertions for validity (`start <= end`) of `paddr_to_pptr_reg` return values in boot code.

Memory regions `region_t`, `p_region_t`, and `v_region_t` are generally assumed to satisfy `start <= end`. This condition can be violated after address translation.

The correct response may be to fail or to split the region into `[start .. -1`], `[0 .. end]`. While overflow and even invalid kernel pointers are fine for device untyped where these are never dereferenced, the code still assumes `start <= end`, so doing nothing in cases where this is violated is incorrect.

### DIFF
--- a/include/basic_types.h
+++ b/include/basic_types.h
@@ -46,16 +46,19 @@ enum _bool {
 };
 typedef word_t bool_t;
 
+/** A region of kernel virtual memory. Empty if start == end. Must satisfy start <= end. */
 typedef struct region {
     pptr_t start;
     pptr_t end;
 } region_t;
 
+/** A region of physical addresses. Empty if start == end. Must satisfy start <= end. */
 typedef struct p_region {
     paddr_t start;
     paddr_t end;
 } p_region_t;
 
+/** A region of user virtual addresses. Empty if start == end. Must satisfy start <= end. */
 typedef struct v_region {
     vptr_t start;
     vptr_t end;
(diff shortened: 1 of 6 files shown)


lsf37:
This does not really fix issue #1002, but it will flag it earlier. Ideally we should split device regions where the end overflows so that this memory can be used.

lsf37:
Potentially a better (or additional?) way to check this would be to add an overall check in `check_available_memory` and demand that physical regions are split up already such that partial overflow does not happen.

Then we don't have to deal with that complexity in the boot code.

lsf37:
Just saw that there are a bunch more region conversions that don't use `paddr_to_pptr_reg`, but `paddr_to_pptr` directly. I can add asserts for those as well, but I'll wait for feedback on this first.

lsf37:
Coo, and it turns out we're doing this everywhere and things were working just fine. Which bit am I missing? I don't see how this line

https://github.com/seL4/seL4/blob/ec170f4c21cdd6c9e163c0da681c912a52b97b8a/src/kernel/boot.c#L630

works correctly if not `start <= end`.

lsf37:
Right, I figured it out, it all ends up working because the loop forces alignment. Thank you for joining me on my journey of self discovery ;-)

We seriously need to document this. Will put up a different PR for this.

Issue #1004, opened 2023-04-03
document why it is Ok for some virtual kernel memory regions to overflow


lsf37:
In #1002 and #389 we both stumbled over the fact that sometimes the function `create_untypeds_for_region` is being supplied with a region that overflows, i.e. where `end < start`.

This can happen when a valid physical memory region is translated to kernel virtual addresses. The overflow is harmless and is being handled gracefully by `create_untypeds_for_region` for device untypeds.

This is because the loop in that function splits the region into size-aligned chunks, and size-aligned chunks cannot themselves overflow, i.e. they will always satisfy `start <= end` (and because it checks for minimum size, actually `start < end`).

The overflow might also be harmless for normal untypeds as long as the resulting virtual addresses are in the kernel window.

All of this should be documented at least in the code, and since it is user visible possibly also somewhere in the manual.

PR #1005, opened 2023-04-03
boot: strengthen distinctness check

lsf37:
Strengthen the distinctness checks for memory regions in `check_available_memory` and `check_reserved_memory`.

We want the property `reg[i-1].end < reg[i].start`, because if these values are equal, `reg[i].start` is part of both regions. Directly adjacent regions have `reg[i].start == reg[i-1].end + 1`.

The negation of `reg[i-1].end < reg[i].start` is `reg[i].start >= reg[i-1].end`.

Caught this while looking at documenting the overflow issue (#1004).

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -765,7 +765,7 @@ BOOT_CODE static bool_t check_available_memory(word_t n_available,
         }
 
         /* Regions must be ordered and must not overlap. */
-        if ((i > 0) && (r->start < available[i - 1].end)) {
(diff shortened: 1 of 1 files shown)


lsf37:
Not my best days understanding the region code. Of course the end is exclusive, otherwise an emptiness check of `start == end` doesn't make sense. That means the existing test is correct.

But then again calling it `end` is then just plain wrong. It's one past the end, and our printing of regions as intervals is also wrong, it really shouldn't use `]` which means inclusive. More doc issues to open..

PR #1006, opened 2023-04-03
boot: sanity checks for provide_untyped_cap

lsf37:
Check arguments for alignment, size, and kernel window (if not device untyped). This provides sanity checks in case any of the memory region computations are wrong.

Since this code is not performance critical and can return failure, these are not assertions, but normal conditions with user feedback. I.e. they are on in release and verified configurations.

(One part of #1004)

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -655,7 +655,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
        need to assume that the kernel window is aligned up to potentially
        seL4_MaxUntypedBits. */
     if (!device_memory && !pptr_in_kernel_window(pptr + MASK(size_bits))) {
-        printf("Kernel init: End of non-device untyped at %p outside kernel window (size %lu)",
+        printf("Kernel init: End of non-device untyped at %p outside kernel window (size %"SEL4_PRIu_word")",
                (void *)pptr, size_bits);
         return false;
     }
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -634,7 +634,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
 
     /* Bounds check for size parameter */
     if (size_bits > seL4_MaxUntypedBits || size_bits < seL4_MinUntypedBits) {
-        printf("Kernel init: Invalid untyped size %lu", size_bits);
+        printf("Kernel init: Invalid untyped size %"SEL4_PRIu_word, size_bits);
         return false;
     }
 
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -640,7 +640,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
 
     /* All cap ptrs must be aligned to object size */
     if (!IS_ALIGNED(pptr, size_bits)) {
-        printf("Kernel init: Unaligned untyped pptr %p (alignment %lu)", (void *)pptr, size_bits);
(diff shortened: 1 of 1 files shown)


src/kernel/boot.c
```diff
@@ -599,6 +628,29 @@ BOOT_CODE static bool_t provide_untyped_cap(
...
+        printf("Kernel init: Unaligned untyped paddr %p (alignedment %lu)", (void *)pptr, size_bits);
+        return false;
+    }
+
+    /* All cap ptrs apart from device untypeds must be in the kernel window */
+    if (!device_memory && !pptr_in_kernel_window(pptr)) {
+        printf("Kernel init: Non-device untyped paddr %p outside kernel window",
+               (void *)pptr);
+        return false;
+    }
```
kent-mcleod, line 652: Does this also need to check that the end of the region isn't outside of the kernel window? I don't think there's a constraint that the kernel window is power-of-2 sized.

lsf37: Yes, good point, will add that check.

lsf37: Done.

lsf37:
I think this is now in good shape and could be merged. HW tests are passing, so there are no violations of these properties so far which is good :-)

kent-mcleod (approved):

src/kernel/boot.c
```diff
@@ -599,6 +628,38 @@ BOOT_CODE static bool_t provide_untyped_cap(
 {
     bool_t ret;
     cap_t ut_cap;
+
+    /* Since we are in boot code, we can do extensive error checking and
+       return failure if anything unexpected happens. */
+
+    /* Bounds check for size parameter */
+    if (size_bits > seL4_MaxUntypedBits || size_bits < seL4_MinUntypedBits) {
+        printf("Kernel init: Invalid untyped size %lu", size_bits);
```
axel-h, line 637: Nitpicking
```suggestion
        printf("Kernel init: Invalid untyped size %"SEL4_PRIx_word, size_bits);
```

src/kernel/boot.c
```diff
@@ -599,6 +628,38 @@ BOOT_CODE static bool_t provide_untyped_cap(
...
+        return false;
+    }
+
+    /* All cap ptrs must be aligned to object size */
+    if (!IS_ALIGNED(pptr, size_bits)) {
+        printf("Kernel init: Unaligned untyped pptr %p (alignment %lu)", (void *)pptr, size_bits);
```
axel-h, line 643:
```suggestion
        printf("Kernel init: Unaligned untyped pptr %p (alignment %"SEL4_PRIx_word")", (void *)pptr, size_bits);
```

src/kernel/boot.c
```diff
@@ -599,6 +628,38 @@ BOOT_CODE static bool_t provide_untyped_cap(
...
+        printf("Kernel init: Non-device untyped pptr %p outside kernel window",
+               (void *)pptr);
+        return false;
+    }
+
+    /* Check that the end of the region is also in the kernel window, so we don't
+       need to assume that the kernel window is aligned up to potentially
+       seL4_MaxUntypedBits. */
+    if (!device_memory && !pptr_in_kernel_window(pptr + MASK(size_bits))) {
+        printf("Kernel init: End of non-device untyped at %p outside kernel window (size %lu)",
```
axel-h, line 658:
```suggestion
        printf("Kernel init: End of non-device untyped at %p outside kernel window (size %"SEL4_PRIx_word")",
```

Issue #842, opened 2022-05-15
discoverability of bitfield functions


ChenYangng:
How do functions in structures_gen.h get generated?

Indanz:
They are generated by tools/bitfield_gen.py from structures.bf files, e.g. include/arch/arm/arch/64/mode/object/structures.bf. Search for "_gen.h" in CMakeLists.txt for more details.

ratmice:
It'd perhaps be good to add a comment to the generator/generated code indicating the source.

axel-h:
PRs are always welcome ;)
Otherwise greping through the ninja files and looking at the CMake builtd targets graph often yields such information.

Indanz:
In this case it's far from obvious how it's generated and grepping the source directory doesn't really work (I tried). Grepping the build directory works slightly better.

What CMake build target graph? The one you get with `cmake --graphviz=test.dot .`?

For automatically generated files it is very common to say in the header how they are generated and from which input files. But if I add:

```
diff --git a/tools/bitfield_gen.py b/tools/bitfield_gen.py
index 31da3a92346a..7d2af948598e 100755
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -3057,6 +3057,8 @@ if __name__ == '__main__':

                 print("end", file=out_file)
     else:
+        if not in_filename is None:
+            print("/* Generated from '%s' */" % in_filename, file=out_file)
         print("#pragma once\n", file=out_file)
         print('\n'.join(map(lambda x: '#include <%s>' % x,
                             INCLUDES[options.environment])), file=out_file)
```
I get `/* Generated from '/home/iz/vmm-mep/build3/kernel/generated/arch/object/structures.bf.pbf' */`, which is the pre-processed output. And this is mentioned in the build.ninja log already: `DESC = Generating from generated/arch/object/structures.bf.pbf`. But there the trails ends, it's hard to find how/why the bf.pbf files get generated. That magic happens at CMakeLists.txt lines 554-555, in the "Bitfield header generation" section. Following the trail all the way back to the correct structures.bf file is almost impossible.

All in all the original question was a very good one.

kent-mcleod:
We could generate a summarized "Generated from " line in CMake and pass it through the script so that the context is preserved?

Indanz:
Sounds good, but that seems far from trivial because of the many layers of indirection.

[...]

PR #1009, opened 2023-04-06
add manual for bitfield generator

lsf37:
Add a user manual for the bitfield generator that documents syntax, semantics and basic concepts. Should also be able to serve as main spec for what the tool is supposed to do.

This is not fully complete yet (see TODO list in comment at the bottom), but I think it is already useful and can be merged.

One thing I'm less sure about is where it should go. I've currently put it into the `tools` directory so it is next to the generator itself and purposefully kept it to a single `.md` file. An alternative would be to create a `docs/` directory and put it there (maybe making the manual also a subdir of that `docs/` directory). Happy with either.

### DIFF
--- a/tools/bitfield_gen.md
+++ b/tools/bitfield_gen.md
@@ -5,6 +5,25 @@
 
 # Bitfield Generator Manual
 
+- [Introduction](#introduction)
+- [Bit Field Generator by Example](#bit-field-generator-by-example)
+  - [Blocks](#blocks)
+  - [Field High](#field-high)
+  - [Visible Field Order](#visible-field-order)
+  - [Tagged Unions](#tagged-unions)
+  - [Advanced Tag Fields](#advanced-tag-fields)
+    - [Tag Slices](#tag-slices)
+    - [Masks](#masks)
+  - [Raw Value Access](#raw-value-access)
+  - [Architecture Parameters and Canonical Pointers](#architecture-parameters-and-canonical-pointers)
+- [Syntax Reference](#syntax-reference)
+  - [Lexical Structure](#lexical-structure)
+  - [Syntactic Structure](#syntactic-structure)
+  - [Restrictions](#restrictions)
+- [Correctness](#correctness)
+- [Limitations](#limitations)
+
+
 ## Introduction
 
 The tool `bitfield_gen.py` in the `seL4/tools/` directory generates bitfield
--- a/tools/bitfield_gen.md
+++ b/tools/bitfield_gen.md
@@ -1,0 +1,673 @@
+<!--
+  Copyright 2023 Proofcraft Pty Ltd
+  SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Bitfield Generator Manual
+
+## Introduction
+
+The tool `bitfield_gen.py` in the `seL4/tools/` directory generates bitfield
+accessor code in C with formal specifications and proofs in Isabelle/HOL.
+
+It takes specifications of bitfield layouts in memory such as the following
+
+```bf_gen
+block VMFault {
+    field     address           32
+
+    field     FSR               5
+    padding                     7
+    field     instructionFault  1
+    padding                     15
+    field     seL4_FaultType    4
+}
+```
+
+and generates a C struct with constructor and getter/setter functions with
+call-by-value and call-by-pointer semantics. It also, separately, generates a
+corresponding data type definition in Isabelle/HOL with corresponding functions
+in HOL for that data type, as well as proofs that the generated C code correctly
+implements these functions. This includes absence of undefined behaviour in C.
+
+The purpose of the tool is to provide bit-level access to structures in memory
+with precise, deterministic and reliable layout semantics, with proofs of
+correctness, and without the need for manual masking and shifting operations.
+The C standard leaves C bitfields too underspecified to be fit for that purpose.
+
+In addition to [basic blocks](#blocks) of bitfields such as above, the tool
+supports [tagged unions](#tagged-unions) and various layout options for these
+blocks.
+
+## Bit Field Generator by Example
+
+This section walks through the features of the bitfield specification languages
+by example. The [full syntax](#syntax-reference) is described in the next
+section for reference.
+
+### Blocks
+
+The basic concepts of the specification languages are blocks which contain a
+list of fields. The example used in the introduction declares a block with
+name `VMFault`, which contains four fields and two padding entries:
+
+```bf_gen
+block VMFault {
+    field     address           32
+
+    field     FSR               5
+    padding                     7
+    field     instructionFault  1
+    padding                     15
+    field     seL4_FaultType    4
+}
+```
+
+This will generate a type `VMFault_t` in C. Assuming that the word size of the
+underlying architecture is 32 bits, this `VMFault_t` type will have a size of
+two machine words. The sizes of the content and padding fields in a block must
+add up to a multiple of the machine word size.
+
+The bits in the block are laid out in memory in the order they are mentioned in
+the specification. That is, the left-most 32 bits (i.e. bits 63 to 32) of
+`VMFault` will correspond to the `address` field, the next 5 bits to `FSR`, the
+next 7 bits are inaccessible padding which is set to 0 on creation of the
+bitfield, then one bit for `instructionFault`, 15 bits of padding, and finally 4
+bits (bits 0 to 3 of the entire type) for `seL4_FaultType`.
+
+The following diagram corresponds to the block specification above.
+
+```bf_dia
+   │------------------------  address  ----------------------------│
+   ┌─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┐
+   │                                                               │ ..
+   └─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┴─┘
+   63                                                             32
+
+                     instruction fault                  seL4_FaultType
+   │-- FSR --│             │-|                             │-------|
+   ┌─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┬─┐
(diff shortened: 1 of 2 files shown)


tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+
+```bf_gen
+base 32
+```
+
+selects a 32 bit native word size.
+
+Many 64 bit architectures have the concept of canonical pointers where not all
+64 bits of the available space are usable for pointers, but only for example 48
+bits. In this setting a pointer value would canonical if and only if all top
```
corlewis, line 510:
```suggestion
bits. In this setting a pointer value would be canonical if and only if all top
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+bits are set when bit 47 (counting from 0) is set. This is equivalent to sign
+extension, i.e. treating bit 47 as the sign bit and extend that sign bit up to
```
corlewis, line 512:
```suggestion
extension, i.e. treating bit 47 as the sign bit and extending that sign bit up to
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+The tool will not access padding fields, but is not guaranteed to preserve them
+when copying. Writing/reading padding fields is not guaranteed to be stable.
+
+### Architecture Parameters and Canonical Pointers
+
+The tool supports architectures with 32 and 64 bit word sizes. The command
+
+```bf_gen
+base 32
+```
```
corlewis, line 504: It might help to say where this command goes and what it's scope is. This is probably connected to unwritten sections involving the `.bf` file extension and how the bitfield generator is used.

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+means word size is 64, pointers are at most 48 bits, and no sign extension. This
+is currently equivalent to just `base 64`.
+
+## Syntax Reference
+
+### Lexical Structure
+
+The generator distinguishes identifiers, integer literals, comments, white
+space, keywords, and operators. Keywords and operators are shown inline in the
+grammar in the next section. The rest are define below.
```
corlewis, line 544:
```suggestion
grammar in the next section. The rest are defined below.
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+
+A multi-letter ident must start with a letter or underscore, followed by
+letters, underscores or numbers. A one-letter ident must consist of one
+letter. Letters are Latin letters `[a-z]` and `[A-Z]`, digits are `[0-9]`.
```
corlewis, line 548: I am assuming that this is what you meant here. If it is, then I think it would be nicer to distinguish letters and characters.
```suggestion
A multi-character ident must start with a letter or underscore, followed by
letters, underscores or numbers. A one-character ident must consist of one
letter. Letters are Latin letters `[a-z]` and `[A-Z]`, digits are `[0-9]`.
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+
+To gain confidence that the generated specification is meaningful and itself has
+the right properties it needs human inspection or it needs to be used in further
+proofs. In the seL4 verification both are the case. In particular, the generated
+specification together with the generated implementation proofs are the
+interface that the C verification uses when it encounters bitfield functions. If
+the generated specifications are trivial, not useful, wrong, or do not express
+the behaviour of the generated C (e.g. if the generated correctness proof was
+trivial), these proofs could not succeed, because the specification would then
+not provide the need semantics of the code (e.g. that loading a value after
```
corlewis, line 648:
```suggestion
not provide the needed semantics of the code (e.g. that loading a value after
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+storing it does not lose information). If the specification provides strong
+enough properties to make the rest of the C verification succeed, we know that
+the specification is strong enough for our needs, even though it might still be
+missing additional facts or properties the generated C code might have. This is
```
corlewis, line 652:
```suggestion
missing additional facts or properties that the generated C code might have. This is
```

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+```c
+VMFault_t VMFault(uint32_t address, uint32_t FSR, uint32_t instructionFault,
+                  uint32_t seL4_FaultType);
+```
+
+Values that do not fit in the bit size of the target field will be cut off
+accordingly, i.e., as if using `val & mask(size)`.
+
+To read the value of each field, the generator will provide the following
+functions:
```
corlewis, line 122: This will be covered by a future pruning section, but for now it might be worth mentioning somewhere around here that the generator only provides functions that are used.

corlewis (approved):
This is really good, and I agree that it is already useful and could be merged. It might be worth making an explicit note that it isn't complete yet though, in case someone is reading it in a form where they aren't able to see the TODO comment.

I can only comment from the perspective of someone who has used and read `.bf` files many times before without ever looking at the generator itself, but I think that everything so far is nicely written and relatively understandable. I did have to go through the `masks` sections a few times to properly understand it, but I think that was mostly due to needing to get the bit patterns straight in my mind, and I don't have any semantic suggestions for how it could be improved.

tools/bitfield_gen.md
```diff
@@ -0,0 +1,672 @@
...
+
+The sub mask inclusion means that the generator can check mask sizes
+consecutively. I.e. if `tag_value & 0xe != 0xe`, we always know it is a 4 bit
+tag, no matter how many other tag sizes are declared.
+
+The tag mask value declared for each alternative in the tagged union must fit
+the mask for the size of the tag field in the block. Remember that with `mask 4
+ 0x0e` the `0x0e` bit pattern means that size 4 does *not* apply, and instead
+the next larger size applies. So in the example, because `content_cap` has a
+4-bit tag field, `content_cap` must have a tag value where the bits `0x0e` are
```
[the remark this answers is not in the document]
lsf37: well spotted, thanks!

tools/bitfield_gen.md
lsf37: Good point. I should make sure that is actually true, but I think you can have it anywhere in the spec.

tools/bitfield_gen.md
lsf37: Yes, reading #842 I think you're right. The problem is that it's a bit more complicated (it's more the build system that does this than the generator), but I'll figure something out.

tools/bitfield_gen.md
lsf37: You mean because `_` and `[0-9]` are not letters? Yes, I think your change makes sense.

tools/bitfield_gen.md
lsf37: So it turns out I was wrong and you can have multiple base commands and everything will work just fine. The base command covers all subsequent blocks.

tools/bitfield_gen.md
lsf37: Now done a bit further down in the document.

tools/bitfield_gen.md
lsf37: Have fixed that now and removed the "exactly one" condition later in the restrictions section as well.

lsf37:
> I did have to go through the `masks` sections a few times to properly understand it, but I think that was mostly due to needing to get the bit patterns straight in my mind, and I don't have any semantic suggestions for how it could be improved.

It is the most mind bending part of the generator. A diagram would probably help, but I'm not entirely sure what to put into the diagram :-)

tools/bitfield_gen.md
```diff
@@ -0,0 +1,653 @@
...
+    padding                     15
+    field     seL4_FaultType    4
+}
+```
+
+and generates a C struct with constructor and getter/setter functions with
+call-by-value and call-by-pointer semantics. It also, separately, generates a
+corresponding data type definition in Isabelle HOL with corresponding functions
+in HOL for that data type, as well as proofs that the generated C code correctly
+implements these functions. This includes absence of undefined behaviour in C.
```
kent-mcleod: It might be useful to explain why this tool was made. My understanding is that it's because the C standard doesn't completely specify how bit-fields are laid out, and originally the compiler implementations tended to be inconsistent or have bugs?

tools/bitfield_gen.md
```diff
@@ -0,0 +1,653 @@
...
+
+```c
+VMFault_t VMFault_set_address(VMFault_t vm_fault, uint32_t address);
+VMFault_t VMFault_set_FSR(VMFault_t vm_fault, uint32_t FSR);
+VMFault_t VMFault_set_instructionFault(VMFault_t vm_fault, uint32_t instructionFault);
+VMFault_t VMFault_set_seL4_FaultType(VMFault_t vm_fault, uint32_t seL4_FaultType);
+```
+
+All of the above (constructor, getter, setter) are pure functions without side
+effects or dependency on global state. The generator declares all of them as
```
kent-mcleod, line 141: Is this true? I think on some compiler configurations, there are `assert`s inserted that call abort if some arguments have invalid values...

tools/bitfield_gen.md
```diff
@@ -0,0 +1,653 @@
...
+See the [Architecture Parameters][] section below for details.
+
+For instance, if the platform is declared to have base 64 and 48 significant
+bits, and we have a pointer that we know is aligned to 12 bits, we need to store
+only 48-12 = 36 bits to recreate the correct original pointer value. That means
+if we declare
+
+```bf_gen
+block store_more {
+    field       some_info  26
```
kent-mcleod, line 228: Was this supposed to add up to `64` total bits? in which case some_info should probably be `16`.  I don't think it breaks the example though it's just odd to be declaring a block with 74 total bits.

tools/bitfield_gen.md
kent-mcleod: One other thought is that by knowing and controlling the exact bitfield allocations, the fastpath implementations are allowed to rely on layout optimizations.

kent-mcleod (approved):

tools/bitfield_gen.md
lsf37: Yes, totally was. Don't trust verification people with adding numbers :-)

tools/bitfield_gen.md
lsf37: Yes, it does insert `asserts`, will mention those. They only have side effects when the code is incorrect ;-)

tools/bitfield_gen.md
lsf37: Good point. I'll leave out the compiler bugs from back when on gcc for Arm, because I suspect they have been fixed by now and that is mostly of historical interest, but the precise layout semantics are still needed over standard C bitfields.

PR #1010, opened 2023-04-06
bitfield_gen: comment for original source file

lsf37:
Add a `--from_file <file>` option to the bitfield generator for printing a `/* generated from <file> */` message in a comment.

Use this option in cmake to provide the original source .bf file before preprocessing so it's easier to find out where the corresponding definitions are.

This should address the concerns in #842.


axel-h (approved):
Nice

kent-mcleod (approved):

PR #1011, opened 2023-04-06
gic_v3: fix offset for gicd iroutern mem-mapped reg

JorgeMVP:
GICD IROUTERN starts from the offset 0x6100 for the SPI 32. SGIs and PPIs  are not included because they are private.


Indanz:
I see where you're coming from, the GIC datasheet seems to imply that IROUTERn starts for `n = 32` and that 0-31 are not mapped. The GIC datatsheet says:

```
GICD_IROUTER<n>, Interrupt Routing Registers, n = 32 - 1019

The maximum value of n is given by (32*(GICD_TYPER.ITLinesNumber+1) - 1).
GICD_IROUTER<n> registers where n=0 to 31 are reserved.
```

but it also says:

```
GICD_IROUTER<n> can be accessed through its memory-mapped interface:
                               Component         Frame       Offset           Instance
                               GIC Distributor   Dist_base   0x6000 + (8 * n) GICD_IROUTER<n>
```

The 32 offset is what makes the base change from 0x6000 to 0x6100 already. The 0x6100 offset mentioned elsewhere is for convenience, but is actually misleading.

I think the current code is correct as it is.

Indanz:
It would be good to add a debug assertion that incoming interrupts are arriving at the expected core, that would have caught this bug.

Indanz:
`setIRQTarget` uses the value in `mpidr_map[target]` directly after masking it, while `mpidr_to_gic_affinity` translates back and forth. For consistency they should be the same. `MPIDR_MT` should be removed, it's unused and defined twice. Otherwise it seems correct.

JorgeMVP:
> I see where you're coming from, the GIC datasheet seems to imply that IROUTERn starts for `n = 32` and that 0-31 are not mapped. The GIC datatsheet says:
> 
> ```
> GICD_IROUTER<n>, Interrupt Routing Registers, n = 32 - 1019
> 
> The maximum value of n is given by (32*(GICD_TYPER.ITLinesNumber+1) - 1).
> GICD_IROUTER<n> registers where n=0 to 31 are reserved.
> ```
> 
> but it also says:
> 
> ```
> GICD_IROUTER<n> can be accessed through its memory-mapped interface:
>                                Component         Frame       Offset           Instance
>                                GIC Distributor   Dist_base   0x6000 + (8 * n) GICD_IROUTER<n>
> ```
> 
> The 32 offset is what makes the base change from 0x6000 to 0x6100 already. The 0x6100 offset mentioned elsewhere is for convenience, but is actually misleading.
> 
> I think the current code is correct as it is.
> 
> EDIT: The code would be correct if iroutern would start at offset 0x6000 in `struct gic_dist_map`, but it starts at 0x6100. I think the correct fix is to change the structure and leave the code as it is, otherwise it's confusing.

Well, it would be misleading if we add the **reserved part** of struct gic_dist_map as part of IROUTERn. 

It would be also confusing if you change the base from 0x6100 to 0x6000 since 0x6000 doesn't show up in the data structure of the datasheet.

Also, by introducing GICv3, ARM says SGIs and PPIs are no longer responsibility of the GIC Distributor. Why would you give the chance to any new sw to write there?

If the code was correct then it would work, but it doesn't ;)

Guys, let me know what is the preferred fix, I can also change the offset. This one I think it is the most indicated fix but you can chose.

JorgeMVP:
> It would be good to add a debug assertion that incoming interrupts are arriving at the expected core, that would have caught this bug.
> 
> Probably best to make a debug array that remembers the intent of the `setIRQTarget` calls and check in `ackInterrupt` that `CURRENT_CPU_INDEX()` matches.

Could be,  I'd say first fix the bug, then improving the code is a separate PR. This is kind of urgent and clear one.

Indanz:
This bug wouldn't be there if the base started at 0x6000, and `n` refers to the interrupt number, it's even defined starting at 0x6000. If you keep the base at 0x6100, at least rename it to `iroutern_32_1019` or something.

> Could be, I'd say first fix the bug, then improving the code is a separate PR. This is kind of urgent and clear one.

Agree with fixing the bug before improving the code, but disagree with not adding a debug check. It's clearly needed, and it's the only way to prove that the code does what it's supposed to do (not just your change, but all code together). The check can also be in `getActiveIRQ`, where `active_irq` is set. I think such kernel check is needed, and also a sel4test test that uses interrupts in user space on different cores to trigger the code path.

JorgeMVP:
> gree with fixing the bug before improving the code, but disagree with not adding a debug check. It's clearly needed, and it's the only way to prove that the code does what it's supposed to do (not just your change, but all code together). The check can also be in `getActiveIRQ`, where `active_irq` is set. I think such kernel check is needed, and also a sel4test test that uses interrupts in user space on different cores to trigger the code path.

I'm not against the kernel check ;)  you want to know when the IRQ is generated, it is generated in the target core? So we need to track the targets in sw for each SPI?

Indanz:
> you want to know when the IRQ is generated, it is generated in the target core? So we need to track the targets in sw for each SPI?

Yup, that's right. But for greater assurance, I propose keeping a shadow bookkeeping that registers the intention.

Currently the target core can't be changed on-the-fly, so I don't think there are obscure corner cases that can cause false positives for this check.

kent-mcleod:
> Guys, let me know what is the preferred fix, I can also change the offset. This one I think it is the most indicated fix but you can chose.

I think what you have is ok for a fix, would you be able to add a comment to the struct definition of `iroutern[960]` referencing the `n-32` requirement. (The docs also give the equation as `0x6000 + 8n`, but I think you made a good point about the 0x6000 not being listed in the overall register map means we shouldn't declare it in the struct.)

> Yup, that's right. But for greater assurance, I propose keeping a shadow bookkeeping that registers the intention.

I think we would need an issue for this.  It makes sense that the kernel tracks a shadow (or it could just look up the register value) also so that we can support faster invocations for reassigning a target or level setting without having to revoke and recreate the IRQ.

Indanz:
> > Yup, that's right. But for greater assurance, I propose keeping a shadow bookkeeping that registers the intention.
> 
> I think we would need an issue for this.

I think it should be part of the fix, as it's the only way to prove this fix actually fixes things. It seemed okay before, so that it seems okay now doesn't mean much, maybe we're being extra stupid.

> It makes sense that the kernel tracks a shadow (or it could just look up the register value) also so that we can support faster invocations for reassigning a target or level setting without having to revoke and recreate the IRQ.

You can't look up the register value because that will be consistent with how it was written, so the same mistake will be made twice, rendering the check useless.

Being able to change the affinity on the fly without recreating the IRQ object is a different issue, but changing affinity on the fly doesn't need a shadow bookkeeping. It's only this debug check that needs it.

kent-mcleod:
> I think it should be part of the fix, as it's the only way to prove this fix actually fixes things. It seemed okay before, so that it seems okay now doesn't mean much, maybe we're being extra stupid.

Sure.  Would the following be sufficient?: 
- shadow copy exists only in debug configurations,
- It's stored as part of the archNodeState so that checks on irq injection are cache-local,
- A bitmap where if the n'th bit is set then the IRQ is targeting the current core.
- When the IRQ handler is created, the bit in the target core's nodeState is set.
- When IRQs are delivered the core asserts that the bit in its bitmap is set.
- When IRQs are revoked the bit is cleared for all targets

Indanz:
I think you're making it more complicated than necessary, I would just go for a simple array of target core per irq. Costs at most one cache miss more per IRQ, but it's debug code only. As the data is practically read-only it doesn't need to be NodeState. That said, if seL4 already has good bitmap code your solution is fine too.

Indanz:
This is what I had in mind (untested): https://github.com/seL4/seL4/compare/master...Indanz:seL4:irq_affinity_check

JorgeMVP:
> This is what I had in mind (untested): [master...Indanz:seL4:irq_affinity_check](https://github.com/seL4/seL4/compare/master...Indanz:seL4:irq_affinity_check)

@Indanz and @kent-mcleod all good stuff and discussions.

I'm trying my best to do the bug fixes, I have allocation time for that right now and I have a list of kernel bug fixes on to do list.

Please, don't ask me to do test cases or debug features that should be added when the feature is introduced ;) I will be very happy to fix the bugs I found.

JorgeMVP:
Added a comment to avoid confusion as suggested by @kent-mcleod.
Let me know if anything is missing.

Thanks for the review. Looking forward for the next one.

kent-mcleod:
> Added a comment to avoid confusion as suggested by @kent-mcleod.
> Let me know if anything is missing.

Thanks!

> Please, don't ask me to do test cases or debug features that should be added when the feature is introduced ;) I will be very happy to fix the bugs I found.

I think Indan was mainly looking for evidence that the fix was doing the right thing.  We'll probably want to add a generic check that can detect this for all interrupt controllers. I've validated that this fix resolves the interrupt target delivery selection using a modified version of Indan's check. So I'm happy to merge this now.

kent-mcleod (approved):

PR #1013, opened 2023-04-13
boot: add missing line break in error messages

axel-h:
Follow up from https://github.com/seL4/seL4/pull/1006 to add the line breaks in the error messages. There is also another message that does not have the line break, so fixing this also.

Found this because one of my experimental patches indeed failed in one of these new checks.


kent-mcleod (approved):

lsf37 (approved):
Thanks for those.

PR #1021, opened 2023-04-21
gcc.cmake: add more known ARM cross compilers

maaaat:
These are the prefixes used by the official Arm GNU Toolchain [1] It would be convenient if seL4 would recognise it out of the box.

[1] https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads


Indanz (approved):
Change seems fine to me, it saves setting `CROSS_COMPILER_PREFIX` for the most common ARM cross-compilers.

kent-mcleod (approved):

PR #1026, opened 2023-04-23
aarch64: Add option for user cache maintenance 

kent-mcleod:
Add a config option, KernelAArch64UserCacheEnable, that enables user
level access to DC CVAU, DC CIVAC, DC CVAC, and IC IVAU which are cache
maintenance operations for the data caches and instruction caches
underlying Normal memory and also access to the read-only cache-type
register CTR_EL0 that provides cache type information. The ArmV8-A
architecture allows access from EL0 as fast cache maintenance operations
improves DMA performance in user-level device drivers.

These instructions are a subset of the available cache maintenance
instructions as they can only address lines by virtual address (VA).
They also require that the VA provided referrs to a valid mapping
with at least read permissions. This corresponds to lines that the
EL0 could already affect via regular operation and so it's not expected
to break any cache-partitioning scheme.

The config option allows this policy to be selected for a particular
kernel configuration, but it is default enabled as this has been the
existing behavior for current aarch64,hyp configurations and have not
been explicitly disabled in non-hyp configurations.

### DIFF
--- a/src/arch/arm/64/head.S
+++ b/src/arch/arm/64/head.S
@@ -23,6 +23,14 @@
 #define CR_ALIGN_CLEAR   0
 #endif
 
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_AARCH64_USER_CACHE_ENABLE)
+#define CR_USER_CACHE_OPS_SET (BIT(CONTROL_UCT) | BIT(CONTROL_UCI))
+#define CR_USER_CACHE_OPS_CLEAR 0
+#else
+#define CR_USER_CACHE_OPS_SET 0
+#define CR_USER_CACHE_OPS_CLEAR (BIT(CONTROL_UCT) | BIT(CONTROL_UCI))
+#endif
+
 #ifndef CONFIG_DEBUG_DISABLE_L1_ICACHE
     #define CR_L1_ICACHE_SET   BIT(CONTROL_I)
     #define CR_L1_ICACHE_CLEAR 0
@@ -42,11 +50,13 @@
 #define CR_BITS_SET    (CR_ALIGN_SET | \
                         CR_L1_ICACHE_SET | \
                         CR_L1_DCACHE_SET | \
+                        CR_USER_CACHE_OPS_SET | \
                         BIT(CONTROL_M))
 
 #define CR_BITS_CLEAR  (CR_ALIGN_CLEAR | \
                         CR_L1_ICACHE_CLEAR | \
                         CR_L1_DCACHE_CLEAR | \
+                        CR_USER_CACHE_OPS_CLEAR | \
                         BIT(CONTROL_SA0) | \
                         BIT(CONTROL_EE) | \
                         BIT(CONTROL_E0E))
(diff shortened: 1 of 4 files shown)


kent-mcleod:
This is an updated version of https://github.com/seL4/seL4/pull/215

Indanz (approved):
Looks good to me, thanks for implementing this.

Security note:

So the only "dangerous" cache maintenance operation is invalidate without clean, as that could give tasks access to data that they're not supposed to see, when the cache content differs from RAM. (For DMA this is exactly what you want, but not for memory from other sources. For DMA clean-and-invalidate works as well as invalidate-only, as there should be only one writer for a cache line.)

On aarch64 data invalidate instructions are silently changed to clean and invalidate when done in EL0, so no problem there.

For instruction cache, `IC IVAU` is allowed and that doesn't has the same behaviour. But in practice, because ARM has no coherency between data and instruction caches, whoever loaded the instructions needs to unify the caches any way. So no problem here either.

Other cache operations are safe because the same effect can be achieved with normal memory operations.

kent-mcleod:
> On aarch64 data invalidate instructions are silently changed to clean and invalidate when done in EL0, so no problem there.
> 

Which instructions are you referring to here? AFAIK none of the invalidate-only instructions are able to be accessed by EL0 anyway.

yyshen (approved):

lsf37 (approved):
Happy with this from the verification and security perspective.

Indanz:
> > On aarch64 data invalidate instructions are silently changed to clean and invalidate when done in EL0, so no problem there.
> 
> Which instructions are you referring to here? AFAIK none of the invalidate-only instructions are able to be accessed by EL0 anyway.

`DC IVAC`, but I actually misremembered. It always acts as clean and invalidate, even in EL1 and higher, at least on a [Cortex-A35](https://developer.arm.com/documentation/100236/0002/functional-description/cache-behavior-and-cache-protection/invalidating-or-cleaning-a-cache):

_DCIMVAC operations in AArch32 and DC IVAC instructions in AArch64 perform an invalidate of the target address. If the data is dirty within the cluster then a clean is performed before the invalidate._

src/arch/arm/64/head.S
```diff
@@ -23,6 +23,14 @@
 #define CR_ALIGN_CLEAR   0
 #endif
 
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_AARCH64_USER_CACHE_ENABLE)
+#define CR_USER_CACHE_OPS_SET (BIT(CONTROL_UCT) | BIT(CONTROL_UCI))
+#define CR_USER_CACHE_OPS_CLEAR 0
+#else
+#define CR_USER_CACHE_OPS_SET 0
+#define CR_USER_CACHE_OPS_CLEAR (BIT(CONTROL_UCT) | BIT(CONTROL_UCI))
```
Ivan-Velickovic, line 31: Is there are reason why `CONFIG_ARM_HYPERVISOR_SUPPORT` needs to be enabled in order to do the operations? From my understanding of the patch, the config option is on by default and so, regardless of whether the kernel is built as a hypervisor or not, user-level programs can execute the cache instructions? But it looks like the bits are not set unless the kernel is built as a hypervisor.

Ivan-Velickovic: (spotted by @lucypa)

kent-mcleod: It seems like it should be `#if !defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_AARCH64_USER_CACHE_ENABLE)`. The bits are only present in the SCTLR_EL1 register and head.S programs SCTLR_EL2 when CONFIG_ARM_HYPERVISOR_SUPPORT is set.

PR #1028, opened 2023-04-23
gic_v3: use CNTPCT instead of CNTFRQ

JorgeMVP:
- CNTFRQ is a constant value, and does not work for this case, while CNTPCT is the one that should be used as 64-bit physical counter.

### DIFF
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -81,20 +81,20 @@ static uint32_t gicv3_do_wait_for_rwp(volatile uint32_t *ctlr_addr)
     bool_t waiting = true;
     uint32_t ret = 0;
 
-    uint32_t gpt_cnt_tval = 0;
+    uint64_t gpt_cnt_tval = 0;
     uint32_t deadline_ms =  GIC_DEADLINE_MS;
-    uint32_t gpt_cnt_ciel;
+    uint64_t gpt_cnt_ciel;
 
     /* Check the value before reading the generic timer */
     val = *ctlr_addr;
     if (!(val & GICD_CTLR_RWP)) {
         return 0;
     }
-    SYSTEM_READ_WORD(CNTFRQ, gpt_cnt_tval);
+    SYSTEM_READ_64(CNT_CT, gpt_cnt_tval);
     gpt_cnt_ciel = gpt_cnt_tval + (deadline_ms * TICKS_PER_MS);
 
     while (waiting) {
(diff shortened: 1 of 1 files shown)


src/arch/arm/machine/gic_v3.c
```diff
@@ -81,20 +81,20 @@ static uint32_t gicv3_do_wait_for_rwp(volatile uint32_t *ctlr_addr)
...
-    uint32_t gpt_cnt_ciel;
+    uint64_t gpt_cnt_ciel;
 
     /* Check the value before reading the generic timer */
     val = *ctlr_addr;
     if (!(val & GICD_CTLR_RWP)) {
         return 0;
     }
-    SYSTEM_READ_WORD(CNTFRQ, gpt_cnt_tval);
+    SYSTEM_READ_WORD(CNTPCT, gpt_cnt_tval);
```
kent-mcleod, line 93: What do you think about using `getCurrentTime()` which the kernel uses elsewhere for reading the 64bit counter. It selects a different count value based on whether the kernel is running in EL1 or EL2 as I recall that when running under some hypervisors, the physical counter trapped and emulated or the guest is simply killed (https://events19.linuxfoundation.org/wp-content/uploads/2017/12/Christopher-Dall_Arm-Timers-and-Fire.pdf).

Indanz: I agree that using `getCurrentTime()` would be nicer, but the above function is called before `initTimer()` by the boot code, so it's not as simple as that.

Always using the virtual timer instead of the physical one here would avoid trapping to any hypervisor. It's just a protection to avoid looping forever, which in practise will never trigger and timing details don't matter too much.

I would be fine with removing the check altogether, because if this fails the system is broken any way. It was a good debug check, but it probably served its purpose now.

kent-mcleod: In this case we can just use `CNT_CT` which is defined to whatever time source the kernel uses for getCurrentTime(). (And maybe use `SYSTEM_READ_64` instead of  `SYSTEM_READ_WORD` in case there's an issue on aarch32?

JorgeMVP: Thanks, added. Let me know if there is anything missing.

kent-mcleod (approved):

src/arch/arm/machine/gic_v3.c
kent-mcleod: Great, I'll just check all the configurations build, and then should be good to merge.

Issue #902, opened 2022-08-21
Append instead of enqueue tasks woken up by notifications


Indanz:
Notifications should influence scheduling as little as possible. Tasks woken up by notifications should not be put at the head of the release queue.

Enqueue is used to optimise the case where a high priority server replies to a lower priority client. Basically higher priority tasks are allowed to influence the release queue order of lower priority tasks. This makes sense in general.

I don't this makes sense for notifications though. If I understood correctly, this means that if a high priority task notifies multiple tasks, they will be woken up in reverse order. For network packets this means that the last received packets will be handled first instead of the oldest.

For high priority servers it's easy to wake-up multiple tasks by sending multiple notifications. If the newly woken-up tasks keeps getting priority, it could lead to starvation of other equal priority tasks.

For IPC calls the lower priority task being woken up was likely the caller. Although it's possible to handle multiple calls at the same time, that's uncommon, as long blocking operations are generally better implemented with signals.

See [here](https://github.com/seL4/seL4/pull/793#discussion_r913060048) for some context.

pingerino:
"Notifications should influence scheduling as little as possible." -- this is not obviously true to me. It's not obvious what the right thing is here without thinking deeply about it. 
 Perhaps @kevinelp or @gernotheiser  can remember the reasoning for the current logic?

The main use case I can think is to skip threads stuck on a spin lock that might be waiting for a thread blocked on a notification to kick the lock.  Deeper thinking needed here to validate that's a sensible scenario though .

gernotheiser:
I cannot be sure why the implementation ended up the way it is, but I suspect it goes back to originally notifications sharing most of the code path with IPC (that was back in the days when they were misleadingly called "asynchronous IPC", which caused no end of confusion, see https://trustworthy.systems/publications/nictaabstracts/Heiser_Elphinstone_16.abstract).

Conceptually I agree with @Indanz: Notifications should not change scheduling more than necessary, and in particular should not re-order round-robin queues (that's sort-of implied in "round robin"). In general, system calls should avoid side effects that are not *required*.

Having said that, I'm not sure I exactly understand the situation here. Is the scenario you're describing the following?

1. thread A, B, both of prio P, are blocked on notification N
2. thread C with prio >P signals N, A is moved from N's queue to the ready queue of prio P
3. thread C signals N again, B is moved from N's queue to the ready queue of prio P
4. C blocks, and the highest runnable prio is now P

[...]

PR #1029, opened 2023-04-27
SMP: only enter scheduler if node holds BKL

andybui01:
In the IPI remote call paths where target nodes enter the kernel under the protection of the calling core's BKL, the scheduler may be invoked after handling the interrupt. The MCS kernel protects against this by doing a check, however, the commit that added this check did not do the same for the non-MCS kernel, so we remove the ifdef and check for both MCS and non-MCS.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -55,14 +55,10 @@ exception_t handleInterruptEntry(void)
         handleSpuriousIRQ();
     }
 
-#ifdef CONFIG_KERNEL_MCS
     if (SMP_TERNARY(clh_is_self_in_queue(), 1)) {
-#endif
         schedule();
         activateThread();
-#ifdef CONFIG_KERNEL_MCS
     }
-#endif
 
     return EXCEPTION_NONE;
 }


lsf37:
Nice find. Do you have a test case that fails before and passes after?

lsf37 (commented):
I'm not sure the change achieves what the commit message says, though. It looks to me like this this now ends up calling the scheduler for any interrupt, not just for IPIs or on SMP configs. E.g. the 

```c
if (SMP_TERNARY(clh_is_self_in_queue(), 1))
```

translates to just 

```c
if (1) {
```

in the verified unicode configs.

This might already be unintended on MCS, I'm not clear on that.

lsf37:
Ignore my ramblings. We do want the scheduler to always be called on unicore, so this is fine -- it already is being called, I just misread the diff.

andybui01:
Agree with @jearc. This might not currently break anything, however, IMO it is assumed that the scheduler is called under the protection of the BKL. We should make this explicit with either an assertion/check, and rely less on everything playing nicely (such as  the scheduler action only being `SchedulerAction_ResumeCurrentThread` )

lsf37:
> Agree with @jearc. This might not currently break anything, however, IMO it is assumed that the scheduler is called under the protection of the BKL. We should make this explicit with either an assertion/check, and rely less on everything playing nicely (such as the scheduler action only being `SchedulerAction_ResumeCurrentThread` )

I'm happy with that line of reasoning. If we ever get to verify the SMP config, this would make it easier. @kent-mcleod any concerns from your side?

Indanz (approved):
The commit that added this if-check is 798fb76ded. I guess that MCS and SMP were added around the same time and that SMP issues found were fixed for MCS-only sometimes, like here.

lsf37:
The proof test fails, which means this will need a (hopefully trivial) proof update before it can get merged.

src/api/syscall.c
```diff
@@ -55,14 +55,10 @@ exception_t handleInterruptEntry(void)
         handleSpuriousIRQ();
     }
 
-#ifdef CONFIG_KERNEL_MCS
     if (SMP_TERNARY(clh_is_self_in_queue(), 1)) {
-#endif
         schedule();
         activateThread();
```
kent-mcleod, line 60: `IpiRemoteCall_VMCheckBoundNotification` on x86 looks like it is relying on the call to schedule on the way out of the ipi.  So there's potentially impact there if this were to be changed (and maybe it's broken on MCS as it is and could explain some MCS + SMP odd behavior reported in the past: https://github.com/seL4/seL4/pull/118#issuecomment-1473072760).

andybui01: I suppose the performance penalty for a `IpiRemoteCall_VMCheckBoundNotification` + `irq_reschedule_ipi` would be too big?

The MCS + SMP case here looks like it's incorrect and presents a race condition, as the target core's scheduling queues become a critical section.

chrisguikema: @andybui01 can you expand upon this a bit? I'm actually investigating a problem where a VM is getting starved of execution time, and as far as I can tell, the userspace VCPU execution loops aren't the culprit.

andybui01: I'm not really familiar with the x86 virtualization support but here's what I recall: currently the `VMCheckBoundNotification()` function that runs on the destination core may ask for a switch to a new thread. This new thread gets scheduled and activated in `handleInterruptEntry@syscall.c`. This can work because remote call IPIs operate with the acquired BKL of the calling core.

If we follow this PR so that `schedule()` checks that the _current_ core holds the BKL, this IPI will break.

Indanz: > @andybui01 can you expand upon this a bit? I'm actually investigating a problem where a VM is getting starved of execution time, and as far as I can tell, the userspace VCPU execution loops aren't the culprit.

If you are running multiple tasks on the same core that are woken up by notifications, perhaps #902 is related. To quickly test if this is the cause, enable signal fast paths and change the `SCHED_ENQUEUE` with `SCHED_ENQUEUE` at https://github.com/seL4/seL4/blob/0260fba168b74dd2504c9447f2fa8da1161ead87/src/fastpath/fastpath.c#L714

But try removing the whole if-check and calling schedule unconditionally first.

> I'm not really familiar with the x86 virtualization support but here's what I recall: currently the `VMCheckBoundNotification()` function that runs on the destination core may ask for a switch to a new thread. This new thread gets scheduled and activated in `handleInterruptEntry@syscall.c`. This can work because remote call IPIs operate with the acquired BKL of the calling core.
> 
> If we follow this PR so that `schedule()` checks that the _current_ core holds the BKL, this IPI will break.

Everyone takes the lock if IPI is not a remote call, but only x86 wants a schedule if the remote call is `IpiRemoteCall_VMCheckBoundNotification`.

I don't understand why x86 has a `IpiRemoteCall_VMCheckBoundNotification` IPI instead of just using the reschedule IPI.

chrisguikema: This is an x86 system, so the Notification fastpath isn't an option I can work with.

Indanz: Try removing the if-check and see if that solves your problem on x86.

PR #1030, opened 2023-04-27
cmake: define KERNEL_ROOT_DIR

axel-h:
Define the (relative) location of the kernel project in one place only.


lsf37 (approved):
Looks good from my side, removes duplication.

PR #1033, opened 2023-05-01
gic_v3: AP1R for group 1 IRQs

JorgeMVP:
- In non-secure world, group 1 IRQs are used instead of group 0.
- APR for GICv2 is 32-bit while for GICv3 is 64-bit wide.

### DIFF
--- a/include/arch/arm/arch/machine/gic_v3.h
+++ b/include/arch/arm/arch/machine/gic_v3.h
@@ -386,13 +386,13 @@ static inline void set_gic_vcpu_ctrl_vmcr(uint32_t vmcr)
 static inline uint32_t get_gic_vcpu_ctrl_apr(void)
 {
     uint32_t reg;
-    MRS(ICH_AP0R0_EL2, reg);
+    MRS(ICH_AP1R0_EL2, reg);
     return reg;
 }
 
 static inline void set_gic_vcpu_ctrl_apr(uint32_t apr)
 {
-    MSR(ICH_AP0R0_EL2, apr);
+    MSR(ICH_AP1R0_EL2, apr);
 }
 
 static inline uint32_t get_gic_vcpu_ctrl_vtr(void)


Indanz (approved):
Agreed that group 1 makes more sense. The upper 32-bits hold the `NMI` flag.

kent-mcleod:
> APR for GICv2 is 32-bit while for GICv3 is 64-bit wide.

Earlier revisions of GICv3 had the register at 32-bits wide.  It seems that it became 64-bits with the introduction of FEAT_GICv3_NMI.  Do you know of any platforms that use this feature? I'm not opposed to extending the register size now but it may be worthwhile to know what versions of the GIC we are committing to support:
From https://developer.arm.com/documentation/198123/0302/What-is-a-Generic-Interrupt-Controller-:
GICv3 and GICv4 have also received minor updates since they were released:

  -  GICv3.1 added support additional wired interrupts, Secure virtualization and Memory System Resource Partitioning and Monitoring (MPAM)
  -  GICv3.2 added support for Armv8-R AArch64.
  -  GICv3.3 added support for non-maskable interrupts.
  -  GICv4.1 extended virtualization support to cover direct-injection of virtual Software Generated Interrupts (SGIs)
  -  GICv4.2 added support for directly-injecting non-maskable virtual interrupts.

JorgeMVP:
> > APR for GICv2 is 32-bit while for GICv3 is 64-bit wide.
> 
> Earlier revisions of GICv3 had the register at 32-bits wide. It seems that it became 64-bits with the introduction of FEAT_GICv3_NMI. Do you know of any platforms that use this feature? I'm not opposed to extending the register size now but it may be worthwhile to know what versions of the GIC we are committing to support: From https://developer.arm.com/documentation/198123/0302/What-is-a-Generic-Interrupt-Controller-: GICv3 and GICv4 have also received minor updates since they were released:
> 
> * GICv3.1 added support additional wired interrupts, Secure virtualization and Memory System Resource Partitioning and Monitoring (MPAM)
> * GICv3.2 added support for Armv8-R AArch64.
> * GICv3.3 added support for non-maskable interrupts.
> * GICv4.1 extended virtualization support to cover direct-injection of virtual Software Generated Interrupts (SGIs)
> * GICv4.2 added support for directly-injecting non-maskable virtual interrupts.

right, I just had a new  GIC revision, and yep it used to be 32-bit wide register. I think for now supporting only GIC3.0 should be the target, then when this is mature enough, we can think about extending it.

Now, we can think of two things:
- I can either remove the changes of the variable type.
- Or I can keep it there for future use.

Let me know what is preffered.

Thanks

kent-mcleod:
> Now, we can think of two things:
> 
>     * I can either remove the changes of the variable type.
> 
>     * Or I can keep it there for future use.
> 
> 
> Let me know what is preffered.

I think we keep as 32-bit but make the change to group 1.

Thanks

JorgeMVP:
> I think we keep as 32-bit but make the change to group 1.

Done ;)

kent-mcleod (approved):

PR #1038, opened 2023-05-11
generic_timer: force timer to de-assert irq

JorgeMVP:
Generic Timer IRQs are level-sensitive, when the CNT_TVAL is updated the trigger condition is de-asserted and the change is propagated to the GIC in a finite time to clear the pending state.

However, we have to make sure the timer deasserts before EOIR/DIR, otherwise the interrupt happens again. Therefore, we need an isb() to cause the timer to de-assert before EOIR/DIR. 

There is also a chance of spurious IRQ. A spurious IRQ can be generated, in the case we have a level-sensitive IRQ, and its pending state is cleared at device-level but not yet propagated to the GIC. In between the IRQ deactivation and IRQ ack of the new interrupt if the requested change from the timer gets propagated then it causes a spurious IRQ.

https://community.arm.com/support-forums/f/architectures-and-processors-forum/54580/generic-timer-irq-handling

### DIFF
--- a/include/drivers/timer/arm_generic.h
+++ b/include/drivers/timer/arm_generic.h
@@ -32,6 +32,11 @@ static inline void ackDeadlineIRQ(void)
 {
     ticks_t deadline = UINT64_MAX;
     setDeadline(deadline);
+    /* Ensure that the timer deasserts the IRQ before GIC EOIR/DIR.
+     * This is sufficient to remove the pending state from the GICR
+     * and avoid the interrupt happening twice because of the level
+     * sensitive configuration. */
+    isb();
 }
 #else /* CONFIG_KERNEL_MCS */
 #include <arch/machine/timer.h>
@@ -39,6 +44,11 @@ static inline void resetTimer(void)
 {
     SYSTEM_WRITE_WORD(CNT_TVAL, TIMER_RELOAD);
     SYSTEM_WRITE_WORD(CNT_CTL, BIT(0));
+    /* Ensure that the timer deasserts the IRQ before GIC EOIR/DIR.
+     * This is sufficient to remove the pending state from the GICR
+     * and avoid the interrupt happening twice because of the level
+     * sensitive configuration. */
+    isb();
 }
 #endif /* !CONFIG_KERNEL_MCS */
 
--- a/include/drivers/timer/arm_generic.h
+++ b/include/drivers/timer/arm_generic.h
@@ -43,7 +43,6 @@ static inline void ackDeadlineIRQ(void)
 static inline void resetTimer(void)
 {
     SYSTEM_WRITE_WORD(CNT_TVAL, TIMER_RELOAD);
-    SYSTEM_WRITE_WORD(CNT_CTL, BIT(0));
     /* Ensure that the timer deasserts the IRQ before GIC EOIR/DIR.
      * This is sufficient to remove the pending state from the GICR
      * and avoid the interrupt happening twice because of the level
--- a/src/drivers/timer/generic_timer.c
+++ b/src/drivers/timer/generic_timer.c
@@ -26,10 +26,10 @@ BOOT_CODE void initGenericTimer(void)
 #ifdef CONFIG_KERNEL_MCS
     /* this sets the irq to UINT64_MAX */
     ackDeadlineIRQ();
-    SYSTEM_WRITE_WORD(CNT_CTL, BIT(0));
 #else /* CONFIG_KERNEL_MCS */
     resetTimer();
 #endif /* !CONFIG_KERNEL_MCS */
+    SYSTEM_WRITE_WORD(CNT_CTL, BIT(0));
 }
 
 /*


kent-mcleod:
> However, we have to make sure the timer deasserts before EOIR/DIR, otherwise the interrupt happens again. Therefore, we need an isb() to cause the timer to de-assert before EOIR/DIR.

Can you also add a comment above the isb() explaining this, and I'm happy to merge it

Indanz:
The same seems to apply to `ackDeadlineIRQ` for MCS.

It's probably good to mention explicitly that the `isb` should be between the timer register write and the `ackInterrup()` call.

However, I'm not totally convinced adding an `isb` is an improvement, flushing the pipeline seems like a very heavy handed way of enforcing ordering between the timer update and the interrupt ack, just to avoid a possible, but harmless race condition. Overall performance may be better without the `isb`, assuming reordering happens at all.

Do you see spurious interrupts that disappear with this change on your platform?

Indanz:
For non-MCS a spurious timer interrupt would mess up time accounting because `timerTick` would be called too soon, so it is not harmless there.

JorgeMVP:
> For non-MCS a spurious timer interrupt would mess up time accounting because `timerTick` would be called too soon, so it is not harmless there.

I think we should be defensive in the seL4. There is nothing in the contract that tells us the timer will deassert before the ICC_DIR/EOIR. If we get an extra valid or spurious interrupt then the overhead will be much higher than a simple isb.

How often this can happen? I dont know, but that is possible to happen everytime we get a tick, in this case will be two ticks and two kernel exits and several scheduling invocations.

Unpredictable behaviour brings a lot of errors sooner or later and very hard to debug.

@Indanz if you are okay I will add to the MCS version as well?

Indanz:
I agree with you, please add it to MCS too.

Linux doesn't use isb, but checks the `ISTATUS` flag to double check that the timer interrupt actually happened. The conservative option would be to do the same in seL4, but that would impact verification. We could add a debug check that tells us when spurious timer interrupts happen, but if they happen too often then it will break the system.

JorgeMVP:
@Indanz @kent-mcleod please double check if you like it.

Also moved SYSTEM_WRITE_WORD(CNT_CTL, BIT(0)); to the initialization, since the timer is always enabled dont need to re-enable it every system tick.

Let me know if there is anything missing.

Thanks.

include/drivers/timer/arm_generic.h
```diff
@@ -32,13 +32,22 @@ static inline void ackDeadlineIRQ(void)
...
+     * sensitive configuration. */
+    isb();
 }
 #else /* CONFIG_KERNEL_MCS */
 #include <arch/machine/timer.h>
 static inline void resetTimer(void)
 {
     SYSTEM_WRITE_WORD(CNT_TVAL, TIMER_RELOAD);
-    SYSTEM_WRITE_WORD(CNT_CTL, BIT(0));
+    /* Ensure that the timer deasserts the IRQ before GIC EOIR/DIR.
```
axel-h, line 46: I wonder if the comment is a bit misleading here (while technically still correct). How many assumptions are we making here who is calling us in which context? Shouldn't this function just fulfill the calling contract, which seems to give the implicit guarantee that the hw timer is really reset before this returns? That alone should justify the barrier

include/drivers/timer/arm_generic.h
```diff
@@ -32,13 +32,22 @@ static inline void ackDeadlineIRQ(void)
 {
     ticks_t deadline = UINT64_MAX;
     setDeadline(deadline);
+    /* Ensure that the timer deasserts the IRQ before GIC EOIR/DIR.
```
axel-h, line 35: It's maybe a bit over the scope of this PR, but I wonder, why are we not simply disable the timer (`CNT_CTL.ENABLE`) here or mask the interrupt (`CNT_CTL.IMASK`)? That should also de-assert the interrupt (after an `isb()`). Is it a performance optimization that quasi-infinite value is used, or was it just a practical choice/hack? The git history is not clear, but interrupt masking usually seems the common way. 
 
The docs says the underlying counter keeps counting if `CNT_CTL.ENABLE` is cleared. Disabling it is explicitly mentioned to be a potential power saving option option. Also, we would not abuse `setDeadline()' then, it would only be used when somebody wants to actually set a deadline. It should have an `isb()` then also, we the high level calling contract seems to imply that once the function returns the deadline is active.

Indanz: The signal line might be floating if you disable the timer, you may also need to disable the timer interrupt in the GIC then. Masking would work, but may make verification harder because an additional hardware property needs to be modelled.

`setDeadline()` does not need an `isb()`, just writing the new value is enough to set a new deadline. Only here when the interrupt is supposed to be cleared is an `isb` necessary. The other callers don't have any race with the GIC.

You keep mentioning the calling contract, but with out-of-order execution and write buffering you're saying you want hard barriers at the end of every "contract" function.

Indanz (approved):
Please rename your second commit to: "generic_timer: Enable once for MCS".
That fixes a typo and makes it clear that only MCS changed.

Otherwise I'm happy with the changes.

JorgeMVP:
> That fixes a typo and makes it clear that only MCS changed.

you mean for non-MCS? because at the end I only changed for non-MCS version, the MCS was correct :)

maybe I'm missing something. Just confirm, and I will change accordingly. Thanks

kent-mcleod:
I saw that there was an update to the support forum thread:
"The ISB ensures that the effects of the write to TVAL is visible to the ICC_EOIR/ICC_DIR.  So yes, without you increase the chance of the timer interrupt signal still being asserted to the GIC when the Deactivate is sent if you remove the ISB.", which seems to clarify that the isb is needed if you need to guarantee ordering.

kent-mcleod (approved):

lsf37 (approved):
Fine from the verification side insofar as this is below the machine interface for us, i.e. we can't say wether it's better or worse, but it won't break the proofs.

Issue #214, opened 2020-06-30
seL4_ARM_Page_Invalidate_Data doesn't check if the page is writeable (ARMv8)


oscarwcl:
Calling `seL4_ARM_Page_Invalidate_Data` on a page cap that is mapped without write permissions will cause a kernel fault:
```
KERNEL DATA ABORT!
Faulting instruction: 0xffffff8001011d08
FAR: 0x400000 ESR (DFSR): 0x9600014f
halting...
Kernel entry via Syscall, number: 1, Call
Cap type: 1, Invocation tag: 49
```
The faulting instruction is the `dc ivac` instruction here:
https://github.com/seL4/seL4/blob/1adb7e9b9162902b3f9868232827b90456d04d53/include/arch/arm/arch/64/mode/machine.h#L291-L295
which requires write permission to the vaddr. I traced the function calls from here back to `decodeARMFrameInvocation` and as far as I can tell there isn't anything that checks the permissions on the page.

yyshen:
@oscarwcl good catch!

yyshen:
@kent-mcleod @lsf37 What I have in mind to fix this is to use "address translation" instruction to check the write permission of the address before performing an invalidation, unless the the access right is stored somewhere (other than in the page table entries) for a quick check.

lsf37:
The ultimate right (including masks etc) is only in the page table entries, so I guess we do have to walk the table, yes. Is the plan to put the instruction directly inside `invalidateByVA`? (That would probably be convenient for the proofs).

I need to double-check the `Page_Invalidate_Data` invocation if it can only be called on the current page table, or on any frame cap in any table. If the latter, we need to walk the right table..

Is this requirement for write permission new in aarch64, or have we missed this in Armv7 and before as well?

lsf37:
Ah, for flushing we're setting the address space to be flushed to be current for the flush anyway, so that part is all good.

yyshen:
> The ultimate right (including masks etc) is only in the page table entries, so I guess we do have to walk the table, yes. Is the plan to put the instruction directly inside `invalidateByVA`? (That would probably be convenient for the proofs).

It depends if an error would be returned to the caller. If we could simply ignore invalidate without write permission, putting the check in invalidateByVA is OK. But this is somewhat inefficient, since the function is executed for each cache line, but the permission is for a page.

> 
[...]

PR #1041, opened 2023-05-23
aarch64: check if page is writeable for DC IVAC

kent-mcleod:
When performing data cache invalidate by virtual address, the mapping must have write permissions otherwise a fault will be generated. When the kernel performs this operation during cache maintenance invocations it must first check that the mapping has write rights in the decode phase of the operation.

This is a suggested starting point of a fix for https://github.com/seL4/seL4/issues/214


lsf37:
I'm going to close this one, because it's built on the old page table API -- the one for the new version is at  #1265.

PR #1043, opened 2023-05-24
Remove VMKernelReadOnly from AArch64 VM rights

Ivan-Velickovic:
From investigating https://github.com/seL4/seL4/issues/214 it looks like the VM right `VMKernelReadOnly` is not used at all. It was introduced in the initial AArch64 port of seL4, I do not know why it exists, perhaps there is a reason and I am missing something.

### DIFF
--- a/include/arch/arm/arch/64/mode/object/structures.h
+++ b/include/arch/arm/arch/64/mode/object/structures.h
@@ -26,7 +26,6 @@ typedef struct arch_tcb {
 enum vm_rights {
     VMKernelOnly = 0,
     VMReadWrite = 1,
-    VMKernelReadOnly = 2,
     VMReadOnly = 3
 };
 typedef word_t vm_rights_t;
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -149,14 +149,6 @@ static word_t CONST APFromVMRights(vm_rights_t vm_rights)
             return 1;
         }
 
-    case VMKernelReadOnly:
-        if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
-            /* no corresponding AP for S2AP, return None */
-            return 0;
-        } else {
-            return 2;
-        }
-
     case VMReadOnly:
         if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
             return 1;


lsf37:
I think you're right, no other architecture uses it and the verification doesn't mention it either. Let me check that we're using the enum values as enum only before we remove it, maybe there is some bit pattern thing going on.

include/arch/arm/arch/64/mode/object/structures.h
```diff
@@ -26,8 +26,7 @@ typedef struct arch_tcb {
 enum vm_rights {
     VMKernelOnly = 0,
     VMReadWrite = 1,
-    VMKernelReadOnly = 2,
-    VMReadOnly = 3
+    VMReadOnly = 2,
```
lsf37, line 29: I don't think we should change the value of `VMReadOnly` -- these can be used as bit patterns, i.e. `vm_rights_t` below is just `word` and is probably used in PTEs etc.

Which might explain why there is a value here at all for 2, it might just be a pattern we don't use, but where it would look slightly strange to have a jump in the enum. I'd still be happy with that jump and removing the name, because it does suggest to the user that it does something, which is not the case.

Ivan-Velickovic: I thought that the value of a `vm_rights_t` was not directly used anywhere, for example for the PTE encoding `APFromVMRights` is used.

lsf37: You're right, I didn't trace that to the very end, we do appear to only use it symbolically. As long as we don't have C verification I'm still reluctant -- if we've overlooked anything this will break stuff, whereas if we keep the value we can be certain that we are not changing behaviour.

Also from the user perspective, changing this value breaks the ABI. Not that we usually care that much about this aspect, but if it's easy, why not keep it compatible. We have a lot of breaking changes already in this release, better not to introduce more subtle things that might go wrong for people.

Ivan-Velickovic: I don't see how it breaks the ABI as I think this is an enum internal to the kernel and cannot be used by user-space.

Regardless, I understand the value in not changing the value and will drop that change.

lsf37: Oh, you are totally right, I should really not stop tracing things after each point. The API and ABI is via `seL4_CapRights_t`, so this is not a breaking change either way.

So the only concern that would remain with changing the value is being sure that we are right that it's used as pure enum only.

lsf37 (approved):

PR #1047, opened 2023-05-30
elfloader: Replace calls to reset_cntvoff

kent-mcleod:
These calls can now be implemented via binding the /timer driver in the elfloader's device tree configuration.

Test with: seL4/seL4_tools#168

### DIFF
--- a/src/plat/tx1/overlay-tx1.dts
+++ b/src/plat/tx1/overlay-tx1.dts
@@ -8,7 +8,8 @@
 	chosen {
 		seL4,elfloader-devices =
 		    "serial0",
-		    &{/psci};
+		    &{/psci},
+		    &{/timer};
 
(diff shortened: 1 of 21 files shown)


src/plat/tx1/overlay-tx1.dts
```diff
@@ -8,6 +8,7 @@
 	chosen {
 		seL4,elfloader-devices =
 		    "serial0",
+		    &{/timer},
 		    &{/psci};
```
Indanz, line 12: For consistency's sake, timer should be after psci.

Indanz (approved):
I haven't looked at the Elfloader changes yet, but I assume this is needed there.

kent-mcleod:
Need approval on https://github.com/seL4/seL4_tools/pull/168 before this can be merged.

PR #1049, opened 2023-05-31
macros: avoid breaking C subset

lsf37:
`__builtin_offsetof` is not part of the verification C subset -- avoid accidental use by not declaring a macro for it and filter out the single use by explicitly marking it as invisible to verification.

(I've checked that the macro is not used outside the kernel repo either)


lsf37:
This unbreaks  the build that I broke in #1049

lsf37:
just saw that I removed and empty line, will fix that

kent-mcleod (approved):

PR #1050, opened 2023-06-01
Fix setting of CMake variable in BCM2711 config

Ivan-Velickovic:
Apologies, I had made an error in my previous commit which causes a compile error when trying to build the kernel without passing `-DRPI4_MEMORY`.


lsf37 (approved):
👍

PR #1051, opened 2023-06-02
gic_v3: mark gicv3_do_wait_for_rwp dont-translate

lsf37:
The verification C parser is failing to translate this function, but it does not actually need to since this is behind the machine interface anyway. Mark the function as dont-translate to avoid the problem.

### DIFF
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -75,6 +75,7 @@ static inline uint64_t mpidr_to_gic_affinity(void)
 }
 
 /* Wait for completion of a distributor change */
+/** DONT_TRANSLATE */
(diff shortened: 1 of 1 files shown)


axel-h:
what does the parser complain about?

lsf37:
> what does the parser complain about?

It aborts with an internal type error in translation (after parsing, i.e. this only happens when you actually run the proofs). This usually happens when an inline assembly block is called with different types and it looks like `SYSTEM_READ_64` might be the one. Probably called correctly here and might be called with a uint32 somewhere else, but I haven't tracked it down.

The function also contains a `printf`, which is not in the verification subset, but I don't think that is what it was choking on.

corlewis (approved):

lsf37:
Manually merged in 1b7da3627e6f9735fb641eb2fe833eda528b9905

PR #1052, opened 2023-06-02
additional verification configs

lsf37:
- make AArch32 FPU possible for verification builds (cherry-pick 453ad135e2bea30f3)
- switch FPU off explicitly in current verification builds
- cherry-pick imx8mm FPU config for verification from branch imx8-fpu-ver as `ARM_imx8mm_verified.cmake`
- cherry-pick exynos5 hyp config for verification from branch exynos5-ver as `ARM_HYP_exynos5_verified.cmake`

These are designed to work together with https://github.com/seL4/l4v/pull/639 so that multiple verification platforms can be present for a particilar L4V_ARCH settings.

### DIFF
--- a/configs/ARM_HYP_exynos5_verified.cmake
+++ b/configs/ARM_HYP_exynos5_verified.cmake
@@ -1,0 +1,27 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+# If this file is executed then build the kernel.elf and kernel_all_pp.c file
+include(${CMAKE_CURRENT_LIST_DIR}/../tools/helpers.cmake)
+cmake_script_build_kernel()
+
+set(KernelPlatform "exynos5" CACHE STRING "")
+set(KernelARMPlatform "exynos5422" CACHE STRING "")
+set(KernelSel4Arch "arm_hyp" CACHE STRING "")
+set(KernelAArch32FPUEnableContextSwitch OFF CACHE BOOL "")
+set(KernelVerificationBuild ON CACHE BOOL "")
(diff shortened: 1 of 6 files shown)


lsf37:
(style is failing as usual on the exec bits for the cmake files)

lsf37:
I'm planning to add these configs to the preprocess test as well when the corresponding [CI actions PR](https://github.com/seL4/ci-actions/pull/275) is merged.

corlewis (approved):
This all looks good to me, it will be very nice to have better support for multiple verified platforms!

It's minor, but the first commit message could be changed slightly. As you said in the PR description, it's now more about making FPU possible for AArch32 verification builds, instead of it being explicitly for the "ARM" target.

lsf37:
> It's minor, but the first commit message could be changed slightly. As you said in the PR description, it's now more about making FPU possible for AArch32 verification builds, instead of it being explicitly for the "ARM" target.

Indeed. Have adjusted the commit message and force-pushed.

Having FPU on by default for AArch32 might also influence the standard AArch32 verification builds for sel4test in CI. I'll merge here as is, but depending on what we want there, we should potentially start explicitly putting the verification configs into sel4test CI to make sure we cover the precise config, not just something that is close.

Indanz:
> (style is failing as usual on the exec bits for the cmake files)

You can remove execution permission with `git update-index –-chmod=-x <file>`, if doing it the normal way of chmod -x doesn't work for weird reasons. Or is there a reason they are executable?

lsf37:
> > (style is failing as usual on the exec bits for the cmake files)
> 
> You can remove execution permission with `git update-index –-chmod=-x <file>`, if doing it the normal way of chmod -x doesn't work for weird reasons. Or is there a reason they are executable?

We do want them to be executable in this case, they're specifically made executable so that you can just run them to build a kernel image with that specific config.

PR #1056, opened 2023-06-08
x86: missing x2apic.h header inclusion

maaaat:
This looks like a trivial omission. The header is required to build with `KernelLAPICMode=X2APIC`.

### DIFF
--- a/include/arch/x86/arch/machine/timer.h
+++ b/include/arch/x86/arch/machine/timer.h
@@ -12,7 +12,7 @@
 
 #ifdef CONFIG_KERNEL_MCS
 #include <mode/util.h>
-#include <arch/kernel/xapic.h>
+#include <arch/kernel/apic.h>
 
 static inline CONST time_t getKernelWcetUs(void)
 {


maaaat:
> Just wondering, but wouldn't it make more sense to include `include/arch/x86/arch/kernel/apic.h` instead of both xapic.h and x2apic.h?

It certainly does look like it would indeed.

Indanz (approved):
So my only concern is that x2apic apparently isn't tested or used at all, otherwise this issue would have been found earlier.

maaaat:
> So my only concern is that x2apic apparently isn't tested or used at all, otherwise this issue would have been found earlier.

Indeed, unless this code path was once working and accidentally got broken since.

From a quick check, x2apic support was added back in 2016 by commit d746cc9cf61b6f50aee6aacf89e91b1f555b1a88 and not touch much ever since. The line in question however (from timer.h) was changed more recently in 2020 by commit 96456c6ae7eb2dbaca5ab062620963322c72f24c. In general x86 isn't used much compared to ARM so it could be that nobody has tried that since 2020?

kent-mcleod (approved):

lsf37:
> So my only concern is that x2apic apparently isn't tested or used at all, otherwise this issue would have been found earlier.

I admit I haven't looked at any details, but I strongly agree with the principle. What do we need to include it in the tests we currently have? Does this need hardware we currently don't have in the machine queue or do we just need to configure something differently?

(Happy to merge this independently of that, but we should at least open an issue for the test)

Indanz:
>  Indeed, unless this code path was once working and accidentally got broken since.

I'm sure it has worked and that it was tested at the time, it probably was that commit you found that broke it.

seL4 combines all C files into one huge blob, so getting the includes (and static variables) correct is very hard, as the compiler doesn't check it. If an unrelated files includes the needed header and gets put above the broken code then it compiles just fine. All in all seemingly independent innocent include changes can break things. In this case it might be just an untested config option though.

> > So my only concern is that x2apic apparently isn't tested or used at all, otherwise this issue would have been found earlier.
> 
> I admit I haven't looked at any details, but I strongly agree with the principle. What do we need to include it in the tests we currently have? Does this need hardware we currently don't have in the machine queue or do we just need to configure something differently?

Other than enabling the config option in seL4, all modern x86 cpus support x2apic, you can check in /proc/cpuinfo. It can also be enabled in qemu with the +x2apic, but then you also need to specify the cpu model.

> (Happy to merge this independently of that, but we should at least open an issue for the test)

[Done](https://github.com/seL4/sel4test/issues/96).

Issue #942, opened 2022-10-28
RISC-V: Use CLINT.mtime instead of rdtime


Indanz:
At least for the platforms that emulate `rdtime` in SBI. See also [this comment](https://github.com/seL4/seL4/pull/863#issuecomment-1150427868) by @axel-h.

E.g. on HIFIVE `rdtime` can take more than one microsecond, which is a hit taken on every context switch.

I don't have RISC-V hardware, just putting this issue up here so someone who cares about RISC-V performance can pick this up.

Ivan-Velickovic:
So in the case of the HiFive Unleashed which has a U54-MC, using the `mtime` register would involve mapping the CLINT into the kernel's address space (having it as a kernel device?) and then just changing `riscv_read_time` to use `mtime` without going to SBI?

This makes me wonder why isn't the CLINT already a kernel device as is the PLIC? I doubt user-space is supposed to access anything in the CLINT?

Ivan-Velickovic:
The Polarfire also has a U54-MC, the QEMU RISC-V virt platform seems to implement the same CLINT. Not sure about the other supported RISC-V platforms.

Ivan-Velickovic:
By "seems to implement the same CLINT" I mean that they all have the `"riscv,cpu-intc"` compatible string, and I've just checked all the other RISC-V platforms and they also have the same compatible string. Whether having the same compatible string results in the same memory map is not something I know.

Indanz:
Seems like the right approach. Only thing I'm not sure about is whether seL4 needs to configure anything. Presumably that's already done by the SBI or bootloader.

According to [this](https://www.rtos.org/FreeRTOS_Support_Forum_Archive/August_2019/freertos_CLINT_vs_mtime_in_FreeRTOS_RISC-V_port_2e8d6a0c4fj.html) FreeRTOS post, `mtime` and `mtimecmp` registers are standard RISC-V registers, it's just their location that differs per platform. So you can count on all RISC-V ports having it.

However, in the future I would expect native hardware support for the `rdtime` instruction, and if it's available it's always better to use that. So retain the `rdtime` implementation, but make it a platform config option.

User space has to use `rdtime`, you can't give it access without giving it the power to mess up timer interrupts for the kernel. As it seems normal device periphery memory, you could give user space access to it.

The CLINT specifications [here](https://github.com/riscv/riscv-aclint/blob/main/riscv-aclint.adoc) might be useful too.

Indanz:
[...]

PR #1057, opened 2023-06-09
Mark CLINT as reserved device on RISC-V platforms

Ivan-Velickovic:
Without this patch, user-level programs have the ability to map in the core-local interrupt controller on RISC-V platforms which contains the memory-mapped registers for the core-local timer the kernel uses. This is a level of privilege that user-level programs should not have. Writing to the `mtime` register is possible which can then affect the timer interrupts that are delivered to the kernel.

An example of the CLINT memory map can be found in Chapter 9.5 of the SiFive U54-MC manual (which is what the HiFive Unleashed uses). While seL4 does not directly use `mtime`, I believe the emulated `rdtime` instruction uses this register.

I have a patch to sel4test to reproduce my claims here: https://github.com/Ivan-Velickovic/sel4test/commit/289201d6f2cc75aad4702dc9d62a6b71135919ff. Running this test shows that the CLINT and hence the `mtime` register is accessible. Producing an MCS kernel shows an assertion fire immediately after messing with the `mtime` register. This kernel patch at least fixes the issue on the HiFive Unleashed.

Another thing to note is that I failed to write to any other registers in the CLINT such as `msip` or `mtimecmp`, this results in a write-fault. It is not clear to me why this occurs. All testing was performed on a HiFive Unleashed.

This is marked as a draft as I cannot see an agreement across platforms for the size of the CLINT.

On the HiFive Unleashed it is `0x10000`. On the Spike it is `0x10000`. This [page for the Ariane](https://www.cl.cam.ac.uk/~jrrk2/docs/memory-map/) has the size as `0xc0000`.

### DIFF
--- a/tools/hardware.yml
+++ b/tools/hardware.yml
@@ -208,6 +208,17 @@ devices:
         kernel: PLIC_PPTR
         kernel_size: 0x04000000
 
+  # SiFive CLINT (HiFive, Polarfire, Ariane, QEMU RISC-V virt, Spike)
+  # Note that not all CLINTs with this compatible string are of the same size.
+  # However, omitting the kernel_size field works as each kernel device frame
+  # is of size 0x200000, which is currently larger than the CLINT's of all
+  # supported platforms.
+  - compatible:
+      - riscv,cpu-intc
+    regions:
+      - index: 0
+        kernel: CLINT_PPTR
+
   # elfloader rules
   - compatible:
       - arm,psci-0.2
--- a/CHANGES
+++ b/CHANGES
@@ -49,6 +49,9 @@ Upcoming release: BREAKING
 * Changed how `gen_config.h` files are generated. Previously, they were generated at CMake configure time. Now, they
   are generated at build time as a dependency of the `${prefix}_Gen` target. To manually build the kernel
   `gen_config.h` file after running `cmake`, run `ninja gen_config/kernel/gen_config.h`.
+* Remove the ability for user-space on RISC-V platforms to access the core-local interrupt controller (CLINT). The
+  CLINT contains memory-mapped registers that the kernel depends on for timer interrupts and hence should not be
+  accessible by user-space.
 
 ## Upgrade Notes
 
--- a/src/plat/ariane/overlay-ariane.dts
+++ b/src/plat/ariane/overlay-ariane.dts
@@ -7,6 +7,19 @@
 / {
 	chosen {
 		seL4,kernel-devices =
+            &{/soc/clint@2000000},
 		    &{/soc/interrupt-controller@c000000};
 	};
+
+    /*
+     * The size and address of the CLINT is from the memory map listed in the
+     * CVA6 documentation. It can be found here:
+     * https://docs.openhwgroup.org/projects/cva6-user-manual/05_cva6_apu/cva6_apu.html#memory-map
+     */
+    soc {
+        clint@2000000 {
+            compatible = "riscv,cpu-intc";
+            reg = <0x00000000 0x2000000 0x00000000 0x0000c0000>;
+        };
+    };
 };
--- a/src/plat/polarfire/overlay-polarfire.dts
+++ b/src/plat/polarfire/overlay-polarfire.dts
@@ -7,6 +7,18 @@
 / {
 	chosen {
 		seL4,kernel-devices =
+            &{/soc/clint@2000000},
 		    &{/soc/interrupt-controller@c000000};
 	};
+
+    /*
+     * According to the "PolarFire SoC MSS Technical Reference Manual"
+     * (revision H), the CLINT is mapped from 0x0200_0000 to 0x0200_FFFF.
+     */
+    soc {
+        clint@2000000 {
+            compatible = "riscv,cpu-intc";
+            reg = <0x00000000 0x2000000 0x00000000 0x000010000>;
+        };
+    };
 };
--- a/src/plat/qemu-riscv-virt/overlay-qemu-riscv-virt.dts
+++ b/src/plat/qemu-riscv-virt/overlay-qemu-riscv-virt.dts
@@ -15,6 +15,15 @@
          *
          */
         seL4,kernel-devices =
+            &{/soc/clint@2000000},
             &{/soc/plic@c000000};
+    };
+
+    /* The size and address of the CLINT is derived from QEMU source code. */
+    soc {
+        clint@2000000 {
+            compatible = "riscv,cpu-intc";
+            reg = <0x00000000 0x2000000 0x00000000 0x000010000>;
+        };
     };
 };
--- a/src/plat/rocketchip/config.cmake
+++ b/src/plat/rocketchip/config.cmake
@@ -50,6 +50,7 @@ if(KernelPlatformRocketchip)
             INTERRUPT_CONTROLLER drivers/irq/riscv_plic0.h
         )
     else()
+        list(APPEND KernelDTSList "src/plat/rocketchip/overlay-rocketchip-base.dts")
         config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "generic")
         # This is an experimental platform that supports accessing peripherals, but
         # the status of support for external interrupts via a PLIC is unclear and
--- a/src/plat/rocketchip/overlay-rocketchip-base.dts
+++ b/src/plat/rocketchip/overlay-rocketchip-base.dts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2023, UNSW
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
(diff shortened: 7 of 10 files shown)


Ivan-Velickovic:
I cannot find any memory map for the Rocketchip, so I am not sure what the size of the CLINT there. The HiFive Unleashed, Polafire, and QEMU RISC-V virt all have `0x10000` as the size.

For the Spike, in QEMU source code it has the size as `0x10000` but in the Spike simulator source code it's `0xc0000`. In order to have both simulators working maybe it is best to have the size as `0xc0000`?

And the [Ariane/CVA6 documentation](https://docs.openhwgroup.org/projects/cva6-user-manual/05_cva6_apu/cva6_apu.html) has the size as `0xc0000`. I'm not that familiar with the Ariane so not sure if I am looking in the right place.

tools/hardware.yml
```diff
@@ -208,6 +208,14 @@ devices:
         kernel: PLIC_PPTR
         kernel_size: 0x04000000
 
+  # SiFive CLINT
+  - compatible:
+      - riscv,cpu-intc
```
Ivan-Velickovic, line 213: This is somewhat problematic as in fact all the currently supported RISC-V boards have `riscv,cpu-intc` despite having differently sized CLINTs. If there a current way in this hardware YAML to not specify the `kernel_size` as fixed?

Ivan-Velickovic: Right, so omitting it will work assuming the CLINT is always less than `1 << 21` in size. This seems like a reasonable assumption, but a very subtle one. In this case since all platforms have a CLINT that is less than `1 << 21`, specifying `kernel_size` doesn't actually do anything.

lsf37: Nice! Please do leave a comment for that if you're using it, though.

Ivan-Velickovic: Done

Ivan-Velickovic:
Okay, the only platform left to deal with now is the Rocketchip. If someone could provide me with documentation on it I can update the PR, and then it should be complete and ready for review.

Indanz:
After digging a bit, Rocketchip is a softcore using [this](https://github.com/li3tuo4/rocket-chip-zcu/blob/9ec337de15a5fba21962e3f21b2788f57c071e72/src/main/scala/devices/tilelink/CLINT.scala#L76) CLINT implementation (hurray for open source hardware), and according to `CLINTConsts` the size is 0x10000.

The reason the CLINT is variable in size is because the real size depends on the number of harts configured, but 0x10000 seems to be the maximum size for 4096 harts.

Indanz:
This seems like a prerequisite for seL4 using mtime/mtimecmp directly (#942) too.

Ivan-Velickovic:
> The reason the CLINT is variable in size is because the real size depends on the number of harts configured, but 0x10000 seems to be the maximum size for 4096 harts.

Ah yes of course.

Ivan-Velickovic:
> After digging a bit, Rocketchip is a softcore using [this](https://github.com/li3tuo4/rocket-chip-zcu/blob/9ec337de15a5fba21962e3f21b2788f57c071e72/src/main/scala/devices/tilelink/CLINT.scala#L76) CLINT implementation (hurray for open source hardware), and according to `CLINTConsts` the size is 0x10000.

Thanks!

lsf37 (approved):

lsf37:
I think this is ready to merge, but should probably get a Changelog entry that explains the bug and logs that it is fixed.

Ivan-Velickovic:
> I think this is ready to merge, but should probably get a Changelog entry that explains the bug and logs that it is fixed.

Yes good point, done.

src/plat/hifive/overlay-hifive.dts
```diff
@@ -5,8 +5,20 @@
...
+            &{/soc/clint@2000000},
+            &{/soc/interrupt-controller@c000000};
+    };
+
+    /*
+     * According to the SiFive "FU540-C000 Manual" (version v1p4)
+     * the CLINT is mapped from 0x0200_0000 to 0x0200_FFFF.
+     */
+    soc {
+        clint@2000000 {
```
kent-mcleod, line 19: Isn't this the same device that's already defined as a child node inside each cpu node? Are there separate copies of the registers for each CPU?

kent-mcleod:
I'm not strongly against this change because I haven't needed to optimize "gettimeofday" performance on riscv yet, but one thing that's typically important is access to a timestamp counter register without having to trap to a different privilege level. On platforms where rdtime is implemented as a trap to mmode, a way to have a direct mmio mapping to get the same value is going to be desired for performance reasons.

src/plat/hifive/overlay-hifive.dts
Ivan-Velickovic: > Isn't this the same device that's already defined as a child node inside each cpu node? 

Yes, I thought it would be better to have this change in the overlay rather than the original DTS.

> Are there separate copies of the registers for each CPU?

I don't think the registers are banked if that's what you're asking, the memory map [here](https://sifive.cdn.prismic.io/sifive/da6d8344-efb9-4732-875c-4926eb5eff77_u54mc_core_complex_manual_21G2.pdf#%5B%7B%22num%22%3A412%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C0%2C678%2C0%5D) shows separate registers for each HART. However, it looks like there's only one `mtime` register for the whole system.

Ivan-Velickovic:
> I'm not strongly against this change because I haven't needed to optimize "gettimeofday" performance on riscv yet, but one thing that's typically important is access to a timestamp counter register without having to trap to a different privilege level. On platforms where rdtime is implemented as a trap to mmode, a way to have a direct mmio mapping to get the same value is going to be desired for performance reasons.

Do you mean if someone wanted to have a way to access `mtime` from user-space? I guess in the future we could have a similar configuration option that exists on ARM that exports the timer?

kent-mcleod:
A separate comment I have is that some device untyped being able to be used to subvert the kernel operation shouldn't be a sufficient condition to restrict access to said device untyped from user level. There are many device untyped regions that give access to hardware mechanisms that can subvert the kernel:
- clock and reset controllers can reset the clocks that the kernel uses for timer interrupts or even potentially the core clock,
- DDR controllers can affect how the kernel access its memory
- memory controllers can dynamically create alias mappings of physical memory regions that the kernel won't know to restrict access to,
- disk controllers can be used to overwrite the kernel boot image while watchdog timers can trigger a system reset.

It's typically impractical to precisely restrict access to all of these mechanisms while still allowing user level drivers for these components as it'd require platform specific kernel drivers and invocations for each hardware component.  Privileged userlevel components shouldn't hand out device untyped to untrusted components if it's not independently confident that the untyped cannot be used to subvert the system. In the case of the mtime mapping, it should be sufficient to just not give that region out to untrusted components, this is what's done in almost every other case already.

lsf37:
> A separate comment I have is that some device untyped being able to be used to subvert the kernel operation shouldn't be a sufficient condition to restrict access to said device untyped from user level. There are many device untyped regions that give access to hardware mechanisms that can subvert the kernel:

There is a difference here: this is a device the kernel directly interacts with and makes assumptions about that are formally modelled. Those devices can't be accessible to user mode unless we explicitly model that in the proofs.

Ivan-Velickovic:
True, but then I don't quite understand why you would have any reserved kernel devices? In the case of RISC-V, why is the PLIC reserved then?

kent-mcleod:
> True, but then I don't quite understand why you would have any reserved kernel devices? In the case of RISC-V, why is the PLIC reserved then?

So far we typically reserve the regions of registers that the kernel's actually accessing, but wee don't require reserving all possible interfaces to the hardware, eg
- we don't reserve all of the memory mapped CPU interface registers for the GICv3 because the kernel uses the system registers,
- we don't reserve the GenericTimer memory mapped registers either which (if they're not independently restricted to EL3) can be used to change the clock value and the clock frequency.

Ivan-Velickovic:
> So far we typically reserve the regions of registers that the kernel's actually accessing

Right, well the kernel is accessing the CLINT registers (although currently it is doing it indirectly through SBI). But if you reserve registers that the kernel is actually accessing, shouldn't you reserve the generic timer on ARM as well?

Ivan-Velickovic:
> > So far we typically reserve the regions of registers that the kernel's actually accessing
> 
> Right, well the kernel is accessing the CLINT registers (although currently it is doing it indirectly through SBI, which maybe makes a difference). But if you reserve registers that the kernel is actually accessing, shouldn't you reserve the generic timer on ARM as well?

Assuming the ARM memory-mapped registers are actually accessible above (below?) EL3.

lsf37:
> But if you reserve registers that the kernel is actually accessing, shouldn't you reserve the generic timer on ARM as well?

Yes we absolutely should and if we're not doing that we should definitely fix it. These all definitely used to be reserved, I don't know when this happened and I absolutely do not remember agreeing to make these mappable from the verification side. If we're doing that the proofs are unsound.

kent-mcleod (approved):

tools/hardware.yml
```diff
@@ -208,6 +208,17 @@ devices:
         kernel: PLIC_PPTR
         kernel_size: 0x04000000
 
+  # SiFive CLINT (HiFive, Polarfire, Ariane, QEMU RISC-V virt, Spike)
+  # Note that not all CLINTs with this compatible string are of the same size.
+  # However, omitting the kernel_size field works as each kernel device frame
+  # is of size 0x200000, which is currently larger than the CLINT's of all
```
axel-h, line 214: I wonder, can't we simply say `kernel_size: 0x200000` here explicitly then? Then there is less implicit behavior.

PR #1058, opened 2023-06-13
cmake: detect ia32 GCC cross-compilers

Ivan-Velickovic:
From my understanding something like `gcc-multilib` that lets an x86_64 host compile for 32-bit x86 doesn't exist on macOS or any non-x86 host. This patch adds detection for the specific ia32 cross-compilers, e.g https://packages.debian.org/unstable/gcc-i686-linux-gnu.


lsf37 (approved):
I've hit that as well in the past (in aarch64 docker mainly), thanks for looking into it.

PR #1059, opened 2023-06-16
Add missing ifdefs around printing functions

Ivan-Velickovic:
These are only called when CONFIG_DEBUG_BUILD is on, which *usually* means that CONIFG_PRINTING is also enabled, but, it's not necessarily the case.

### DIFF
--- a/src/arch/arm/32/machine/capdl.c
+++ b/src/arch/arm/32/machine/capdl.c
@@ -506,7 +506,9 @@ void debug_capDL(void)
 {
     printf("arch aarch32\n");
     printf("objects {\n");
+#ifdef CONFIG_PRINTING
     print_objects();
+#endif
     printf("}\n");
 
     printf("caps {\n");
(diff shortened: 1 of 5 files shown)


axel-h:
I wonder, wouldn't it be a bit more intuitive when we rename the function `debug_capDL()` to `print_capDL()` and then ifdf-out the whole function if printing is disabled? I should have done this in b75ded65 already, when I created this function. Seems this function can be used in debug and release builds.

axel-h:
What do you think about doing this change: https://github.com/axel-h/seL4/pull/104/commits/1a35bc7bf8d3c21511c0d80f3852c5bda390c62d

Ivan-Velickovic:
Yea sure, I don't really mind as long as the kernel compiles. I just did this patch as it was consistent with what was already there. I can discard the capDL printing changes if you open a PR with those changes.

kent-mcleod (approved):

PR #1060, opened 2023-06-17
boot: remove redundant initialization

axel-h:
The function `init_core_state()` does the initialization already.

I assume the actual goals is using most recent timestamp, even after getting the BKL and debug printing?


Indanz:
> I assume the actual goals is using most recent timestamp, even after getting the BKL and debug printing?

Yes, and as booting takes time, the extra call is not redundant. You don't want to charge the first task for booting up the kernel, do you?

Indanz:
Could you please do these kind of experimental work in your own branch instead of a draft pull request? The current commits have diverged a lot from the initial pull request, so I'm closing this for now if you don't mind. Feel free to re-open when it's a bit more settled and the descriptions are up-to-date.

PR #1062, opened 2023-06-19
Fix enabling of AArch64 user-level cache maintenance operations

Ivan-Velickovic:
These operations should be possible on seL4 with and without hypervisor mode assuming the `KernelAArch64UserCacheEnable` option is set.

Fix according to https://github.com/seL4/seL4/pull/1026#discussion_r1224125835.

I plan on adding a test in sel4test for the user-cache ops since this code seems somewhat fragile. Might be a good idea to hold off on merging till I've done that.


kent-mcleod (approved):

src/arch/arm/64/head.S
```diff
@@ -23,7 +23,7 @@
 #define CR_ALIGN_CLEAR   0
 #endif
 
-#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_AARCH64_USER_CACHE_ENABLE)
+#if !defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_AARCH64_USER_CACHE_ENABLE)
```
axel-h, line 26: I think it would help if there is a comment here in the code with what is mentioned in https://github.com/seL4/seL4/pull/1026#discussion_r1224125835. We can't explain all details, but give some hints at least.

Ivan-Velickovic: Yes, it should have a comment explaining.

Indanz (approved):
Just to be sure I checked that for HYP the config option also works. Then `SCTLR_NATIVE_USER_CACHE_OPS` is set in vcpu.h, so all is right.

PR #1063, opened 2023-06-19
trivial: fix typos

axel-h:


Indanz (approved):
"modeled" is correct in American English. However, seL4 is mostly written by Australians and Europeans, who use UK English, so I'd say this is a fix indeed.

kent-mcleod (approved):

PR #1064, opened 2023-06-19
remove obsolete assert()

axel-h:
Commit f4c41f39 removed a check that dereferenced `tcbSchedContext`. It should have removed this assert() also then.

It's unclear to me what this is supposed to check here. If this is to be kept, then the assertion check should move to generic kernel function `migrateTCB()` and get some explanation also.


Indanz:
You also remove this in #1065, let's close this one and keep the discussion there?

I noticed the removal there, but didn't think worth mentioning. No matter what the reason for the assert, there's nothing x86 specific about it and the other arches don't have it, so it's fine to just remove. Whatever purpose it had, it fulfilled that and it's not needed any more.

So I'm going to close this one, and if for whatever reason your other pull request doesn't get merged, I'll re-open this one again.

Issue #901, opened 2022-08-21
SMP: Avoid unnecessary IPIs because of lazy FPU loading


Indanz:
- Merge `Arch_migrateTCB` with `migrateTCB` and only migrate if newcore != oldcore.

Currently `migrateTCB` is called unconditionally and always migrates a TCB, even if the old core is equal to the new one. This seems like a bug, as it means that if a signal is send to a cross-node task with a bound notification, FPU switching will be done all the time even if there are no other FPU users on that core. See [this](https://github.com/seL4/seL4/pull/793#discussion_r927468009) discussion.

- Disable lazy FPU loading for passive threads.

Lazy FPU loading seems to work badly in combination with SMP and passive threads that use the FPU, as TCB migration should be fast. 

It is to be expected that on SMP systems passive servers will be called from multiple cores, and on SMP cross-node calls are the common case, statistically speaking. However, if a passive server uses the FPU, the next call from a different node will cause a blocking IPI to safe the FPU state, and the server using the FPU will cause a trap to restore and enable the FPU.

An alternative, simpler solution is to get rid of lazy FPU loading altogether. This would also stop lazy FPU based timing side-channels.

Somewhat relevant: [Pre-RFC: FPU as an object](https://sel4.discourse.group/t/pre-rfc-fpu-as-an-object/475).

Indanz:
Now PR #1325 is merged, this problem is reduces as by default all tasks use the FPU (except the idle task). And when configuring people are more likely to disable the FPU for the passive task than for the callers. But keeping this issue open because it is still relevant for the new FPU, as it is semi-lazy.

PR #1065 merged `Arch_migrateTCB` with`migrateTCB`, but a check for newcore != oldcore is still missing.

midnightveil:
Is this fixed now with the merge of 1571?

Indanz:
Yes, thanks. The main issues have been fixed now.

It's not worth keeping it open for the one remaining corner case of a passive task using FPU and being called from multiple cores: Just don't do that, that will be extremely slow. The fix for that, saving FPU state when returning the SC, would slow down passive tasks that don't cross cores, so is not clearly an improvement.

PR #1065, opened 2023-06-20
remove Arch_migrateTCB()

axel-h:
Remove `Arch_migrateTCB()`, because nowadays It does the same on all architectures, so the contents can be moved into the generic code. Architectures that support `CONFIG_HAVE_FPU` must provide proper implementations for `nativeThreadUsingFPU()` and `switchFpuOwner()` anyway, as this is also used in other places in the generic kernel code.

I also have doubts that keeping `Arch_migrateTCB()` as empty function with the current signature has a value, as it lacks the destination core parameter. If there is more to do, introducing more specific function might be the better way. That is why it's removed completely.

### DIFF
--- a/src/model/smp.c
+++ b/src/model/smp.c
@@ -15,7 +15,15 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
 #ifdef CONFIG_DEBUG_BUILD
     tcbDebugRemove(tcb);
 #endif
-    Arch_migrateTCB(tcb);
+#ifdef CONFIG_HAVE_FPU
+    /* If the thread owns the FPU of the core it is currently running on (which
+     * is not necessarily the core, that we are now running on), then release
+     * this cores's FPU.
+     */
+    if (nativeThreadUsingFPU(thread)) {
+        switchFpuOwner(NULL, thread->tcbAffinity);
+    }
+#endif /* CONFIG_HAVE_FPU */
     tcb->tcbAffinity = new_core;
 #ifdef CONFIG_DEBUG_BUILD
     tcbDebugAppend(tcb);
--- a/include/object/tcb.h
+++ b/include/object/tcb.h
@@ -223,10 +223,6 @@ word_t CONST Arch_decodeTransfer(word_t flags);
 exception_t CONST Arch_performTransfer(word_t arch, tcb_t *tcb_src,
                                        tcb_t *tcb_dest);
 
-#ifdef ENABLE_SMP_SUPPORT
-void Arch_migrateTCB(tcb_t *thread);
-#endif /* ENABLE_SMP_SUPPORT */
-
 #ifdef CONFIG_DEBUG_BUILD
 void setThreadName(tcb_t *thread, const char *name);
 #endif /* CONFIG_DEBUG_BUILD */
--- a/src/arch/arm/object/tcb.c
+++ b/src/arch/arm/object/tcb.c
@@ -21,15 +21,3 @@ exception_t CONST Arch_performTransfer(word_t arch, tcb_t *tcb_src, tcb_t *tcb_d
 {
     return EXCEPTION_NONE;
 }
-
-#ifdef ENABLE_SMP_SUPPORT
-void Arch_migrateTCB(tcb_t *thread)
-{
-#ifdef CONFIG_HAVE_FPU
-    /* check if thread owns its current core FPU */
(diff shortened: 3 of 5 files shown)


src/model/smp.c
```diff
@@ -15,7 +15,12 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
 #ifdef CONFIG_DEBUG_BUILD
     tcbDebugRemove(tcb);
 #endif
-    Arch_migrateTCB(tcb);
+#ifdef CONFIG_HAVE_FPU
+    /* check if thread owns its current core's FPU */
```
Indanz, line 19: I'd change the comment to the less awkward: "Check if thread owns the current core's FPU"

Indanz (approved):
This has bugged me before too, it just adds a lot of extra work checking that all implementations are doing the same for no gain.

src/model/smp.c
axel-h: I've made this a bit more verbose now. However, having loooked deeper into this, I wonder if we should add another check in this function, if the core really changes (userland may do all koing of strange things) . If the core does not change, then don't do any FPU handling (and maybe don't even call this function). It would be a slight optimization

Indanz: Actually, I opened an issue for that already: #901

src/model/smp.c
```diff
@@ -15,7 +15,15 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
 #ifdef CONFIG_DEBUG_BUILD
     tcbDebugRemove(tcb);
 #endif
-    Arch_migrateTCB(tcb);
+#ifdef CONFIG_HAVE_FPU
+    /* If the thread owns the FPU of the core it is currently running on (which
+     * is not necessarily the core, that we are now running on), then release
+     * this cores's FPU.
```
Indanz, line 21: Good point, the migrated TCB doesn't need to be running on the current CPU core. So fix the comment all the way:

"then release that core's FPU."

Indanz (approved):

src/model/smp.c
axel-h: Cool, that's the next step then after this is merged.

kent-mcleod (approved):

src/model/smp.c
axel-h: @kent-mcleod was too fast, so I'll fix my typos in another PR then.

Indanz: Well, "this" is ambiguous enough and can refer to the correct core, I thought it would be slightly clearer if "that" was used, but no big deal.

kent-mcleod: oops sorry. I reviewed that the initial feedback had been action-ed but missed the follow up comment

lsf37:
This PR seems to have broken compilation for SMP in some configs, e.g. SABRE_SMP_32 and TX2_SMP_64: https://github.com/seL4/sel4bench/actions/runs/5328166727/jobs/9652486125

kent-mcleod:
> This PR seems to have broken compilation for SMP in some configs, e.g. SABRE_SMP_32 and TX2_SMP_64: https://github.com/seL4/sel4bench/actions/runs/5328166727/jobs/9652486125

It doesn't compile on any smp configuration it seems.....

axel-h:
> > This PR seems to have broken compilation for SMP in some configs, e.g. SABRE_SMP_32 and TX2_SMP_64: https://github.com/seL4/sel4bench/actions/runs/5328166727/jobs/9652486125
> 
> It doesn't compile on any smp configuration it seems.....

Apologies for this merge issue from my branches. Why didn't CI catch this?

lsf37:
> > > This PR seems to have broken compilation for SMP in some configs, e.g. SABRE_SMP_32 and TX2_SMP_64: https://github.com/seL4/sel4bench/actions/runs/5328166727/jobs/9652486125
> > 
> > 
> > It doesn't compile on any smp configuration it seems.....
> 
> Apologies for this merge issue from my branches. Why didn't CI catch this?

CI did catch it, that's why we know of it :-) But the PR checks only include SMP builds for `hw-build`. Maybe we should add at least one SMP config there, but the problem is that there are too many configs, that's why we have the `hw-build` label.

Maybe we should pick one board of each architecture and build all combinations for that on PR. That will still not catch everything, but it would be way better coverage than what we have now.

axel-h:
It should not use the board farm, the QEMUs can handle this also nicely. And that will also allow catching it on forks, which scales much better.

PR #1066, opened 2023-06-20
manual: improve thread affinity description

axel-h:
The manual does not say much about the core affinity of a thread. Add some description.

- replace "master" by "non-MCS"
- improve thread affinity description

### DIFF
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -53,7 +53,7 @@ In non-MCS configurations of the kernel, this will result in the thread immediat
 the scheduler. On the MCS kernel, the thread will only begin running if it has a
 scheduling context object.
 
-In a SMP configuration of the kernel, the thread will resume on the core
+In a SMP configuration of the kernel, the thread will resume on the node
 corresponding to the affinity of the thread. For the non-MCS configurations of the kernel, this is set using
 \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}, while on the MCS kernel the affinity is
 derived from the scheduling context object.
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -54,9 +54,11 @@ the scheduler. On the MCS kernel, the thread will only begin running if it has a
 scheduling context object.
 
 In a SMP configuration of the kernel, the thread will resume on the node
-corresponding to the affinity of the thread. For the non-MCS configurations of the kernel, this is set using
-\apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}, while on the MCS kernel the affinity is
-derived from the scheduling context object.
+corresponding to the affinity of the thread. For non-MCS configurations, the
+default thread affinity is the node the thread's \obj{TCB} object was created
+on, and \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity} can be used to
+explicitly set the affinity. On MCS configurations, the affinity is derived
+from the scheduling context object (see \autoref{sec:sc_creation}).
 
 \subsection{Thread Deactivation}
 \label{sec:thread_deactivation}
@@ -71,6 +73,14 @@ reused or left suspended indefinitely if not needed. Threads will be
 automatically suspended when the last capability to their \obj{TCB} is
 deleted.
 % an example of which is demonstrated in \nameref{ex:second_thread}.
+
+\subsection{Affinity}
+\label{sec:thread_affinity}
+
+It is architecture and platform specific, how an affinity value maps to a
+specific node (core, hart) on a specific platform. There is no guarantee that
+affinity values are compatible across different platforms.
+
 
 \subsection{Scheduling}
 \label{sec:sched}
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -49,12 +49,12 @@ pointer. The thread can then be activated either by setting the
 or by separately calling the \apifunc{seL4\_TCB\_Resume}{tcb_resume} method. Both of these methods
 place the thread in a runnable state.
 
-On the master kernel, this will result in the thread immediately being added to
+In non-MCS configurations of the kernel, this will result in the thread immediately being added to
 the scheduler. On the MCS kernel, the thread will only begin running if it has a
 scheduling context object.
 
 In a SMP configuration of the kernel, the thread will resume on the core
-corresponding to the affinity of the thread. For master, this is set using
+corresponding to the affinity of the thread. For the non-MCS configurations of the kernel, this is set using
 \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}, while on the MCS kernel the affinity is
 derived from the scheduling context object.
 


manual/parts/threads.tex
```diff
@@ -49,14 +49,16 @@ \subsection{Thread Creation}
 or by separately calling the \apifunc{seL4\_TCB\_Resume}{tcb_resume} method. Both of these methods
 place the thread in a runnable state.
 
-On the master kernel, this will result in the thread immediately being added to
+On the standard (non-MCS) kernel, this will result in the thread immediately being added to
```
lsf37, line 52: We should keep the terminology consistent. I checked -- the mentions in this paragraphs are the only ones that refer to "master" kernel, so if we get it correct here, we should be consistent. The origin of calling it "master kernel" is that MCS used to be a separate branch, which is no longer the case.

I'd propose `In non-MCS configurations of the kernel, this will in ..`

manual/parts/threads.tex
```diff
@@ -49,14 +49,16 @@ \subsection{Thread Creation}
...
 the scheduler. On the MCS kernel, the thread will only begin running if it has a
 scheduling context object.
```
lsf37, line 54: And here then: `In the MCS configuration of the kernel, ...`

manual/parts/threads.tex
```diff
@@ -72,6 +74,21 @@ \subsection{Thread Deactivation}
 deleted.
 % an example of which is demonstrated in \nameref{ex:second_thread}.
 
+\subsection{Core Affinity}
+\label{sec:thread_affinity}
+
+On the standard (non-MCS) kernel in a SMP configuration, the core affinity of
+a thread can be changed anytime using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}.
+A thread can change it own affinity, if it has access to it's \obj{TCB} object.
```
lsf37, line 82:
```suggestion
A thread can change its own affinity, if it has access to its \obj{TCB} object.
```

manual/parts/threads.tex
```diff
@@ -72,6 +74,21 @@ \subsection{Thread Deactivation}
...
+Affinity changes triggers a rescheduling of the target thread. If the new
```
lsf37, line 83:
```suggestion
Affinity changes trigger a rescheduling of the target thread. If the new
```

manual/parts/threads.tex
```diff
@@ -72,6 +74,21 @@ \subsection{Thread Deactivation}
 deleted.
 % an example of which is demonstrated in \nameref{ex:second_thread}.
 
+\subsection{Core Affinity}
+\label{sec:thread_affinity}
+
+On the standard (non-MCS) kernel in a SMP configuration, the core affinity of
+a thread can be changed anytime using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}.
```
lsf37, line 81:
```suggestion
In non-MCS SMP configurations of the kernel, the core affinity of
a thread can be changed any time using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}.
```

manual/parts/threads.tex
```diff
@@ -54,9 +54,11 @@ \subsection{Thread Creation}
...
 
 In a SMP configuration of the kernel, the thread will resume on the core
-corresponding to the affinity of the thread. For the standard (non-MCS) kernel, this is set using
-\apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}, while on the MCS kernel the affinity is
-derived from the scheduling context object.
+corresponding to its affinity. For the standard (non-MCS) kernel, this is set
+using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}. By default, a thread's
+affinity is the core where the thread has been running on, that created the
+\obj{TCB} object. On the MCS kernel, the affinity is derived from the
+scheduling context object.
```
Indanz, line 61: Agreed that this change is an improvement.

Perhaps clearer: "By default, a thread's affinity is the core where its \obj{TCB} object was created."

Indanz:
The convention seems to be to call CPU cores "nodes" in the rest of the manual.

manual/parts/threads.tex
```diff
@@ -72,6 +74,21 @@ \subsection{Thread Deactivation}
...
 
+\subsection{Core Affinity}
+\label{sec:thread_affinity}
+
+On the standard (non-MCS) kernel in a SMP configuration, the core affinity of
+a thread can be changed anytime using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}.
+A thread can change it own affinity, if it has access to it's \obj{TCB} object.
+Affinity changes triggers a rescheduling of the target thread. If the new
+affinity is the same than the old affinity, the behaviour is effectively similar
+to what would happen if the target thread did a \apifunc{seL4\_Yield}{sel4_yield}.
```
lsf37, line 85:
```suggestion
affinity is the same as the old affinity, the behaviour is effectively similar
to what would happen if the target thread did a \apifunc{seL4\_Yield}{sel4_yield}.
```

"effectively similar" is not very useful, because it doesn't give people any certainty on what actually happens. We. should either commit to that it really is effectively a  `yield` operation, or we should leave out that part and leave unspecified what happens.

axel-h: Thanks for the feedback. I made it this blurry because I was not sure if there is a defined behaviour at all. That is just what the implementation roughly does. Is there a formal spec here? Otherwise, what is a wording to state what to expect in the worst case. Saying "undefined" is also not helpful to set some informal limits. Can it be worse than 'yield'?

lsf37: For SMP, the manual is the only spec, that's why it matter a bit more what we say here -- I would in this case probably just not talk about this case, because the operation will do what is said before (set the affinity), and whether this comes out to a yield or not in total if the affinity hasn't changed is not really important. If we commit here to it being the same we'd have to make sure that the implementation covers the same corner cases etc and people might start relying on it.

Indanz: Exactly what @lsf37 said: You don't want people to count on behavioural details like these. There is no practical use for these kind of details, please don't document them, that only lowers the signal to noise ration in the manual.

Please drop the whole new section, it's mostly redundant: non-MCS is clear as it is, especially after your improvement. If MCS needs further explanation it should be done in the Scheduling Context Creation chapter and maybe add a reference.

manual/parts/threads.tex
axel-h: I'll change this, if this is the official wording now

manual/parts/threads.tex
```diff
@@ -49,14 +49,16 @@ \subsection{Thread Creation}
...
 
 In a SMP configuration of the kernel, the thread will resume on the core
-corresponding to the affinity of the thread. For master, this is set using
-\apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}, while on the MCS kernel the affinity is
-derived from the scheduling context object.
+corresponding to its affinity. For the standard (non-MCS) kernel, this is set
+using \apifunc{seL4\_TCB\_SetAffinity}{tcb_setaffinity}. By default, a thread's
+affinity is the core where the thread has been running on, that created the
+\obj{TCB} object. On the MCS kernel, the affinity is derived from the
+scheduling context object.
```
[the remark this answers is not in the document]
axel-h: done

manual/parts/threads.tex
axel-h: done

manual/parts/threads.tex
```diff
@@ -72,6 +74,21 @@ \subsection{Thread Deactivation}
...
+Affinity changes triggers a rescheduling of the target thread. If the new
+affinity is the same than the old affinity, the behaviour is effectively similar
+to what would happen if the target thread did a \apifunc{seL4\_Yield}{sel4_yield}.
+
+It is architecture and platform specific, how an affinity value maps to a
+specific core in the current platform. There is no guarantee that affinity
+values are compatible across different platform.
+
+
 \subsection{Scheduling}
```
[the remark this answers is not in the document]
axel-h: will add a link to `sec:sc_creation` then

manual/parts/threads.tex
axel-h: done

manual/parts/threads.tex
axel-h: Extended commit message to state that merge happened for version 11.0.0.

axel-h:
> The convention seems to be to call CPU cores "nodes" in the rest of the manual.

Added a commit that changes this.

manual/parts/threads.tex
axel-h: I think it worth keeping the note that affinity values are platform specific and there is a mapping. Once does not pass the actual physical ID here.

Indanz: Affinity values are a platform specific mapping by definition, between some random number and a physical core. Pointing out the obvious doesn't help any one, so I'm still in favour of dropping the section as it is.

If you mean that on some RISCV platforms the core IDs are not `0...N` then be specific and just state that instead (and personally I would consider that a bug, but oh well). If you mean that the startup task's affinity is platform specific, then that could be mentioned somewhere too.

lsf37 (approved):
Happy with the current version.

PR #1067, opened 2023-06-20
move definition to config file

axel-h:
Since we have a `config.h`, this should be place where the flags are set instead of scattering this across the code.


include/kernel/boot.h
```diff
@@ -136,13 +136,7 @@ static inline BOOT_CODE pptr_t it_alloc_paging(void)
...
-#if defined(CONFIG_DEBUG_BUILD) && defined(ENABLE_SMP_SUPPORT) && defined(CONFIG_KERNEL_MCS)
-/* Test whether clocks are synchronised across nodes */
-#define ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
-#endif
-
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
+/* Test whether clocks are synchronised across nodes */
 BOOT_CODE void clock_sync_test(void);
-#else
-#define clock_sync_test()
```
Indanz, line 147: Removing this breaks non-MCS SMP builds.

Indanz (commented):
I disagree with this change because the clock_sync_test is not a configuration option, so doesn't belong in config.h. It is an internal debug check of the boot code and currently it's contained to boot code. If you move the defines you scatter it around more.

(Could you please capitalise your commit sentences and put the description you put in the pull request also in the commit message, separated by an empty line, so people looking at the commit history can see the reason for why a change was made? And if it's a fix for an issue, note the issue number. This applies to all your pull requests. Thanks.)

Indanz:
Closing this as it's not a clear improvement compared to the existing code. Changes need to be substantially better compared to existing code before it's worth the effort of merging it.

In this case, it also introduces a compile error. That's why dubious changes like these are not worth the effort: The chance that something gets broken by accident is too high compared to any small gain achieved.

Indanz:
As a rule of thumb, if you can't be bothered to compile the code for all configurations affected by your changes, it's probably not worth pushing it upstream. And bigger changes should be tested on real hardware for at least one relevant configuration before you should consider doing a pull request.

axel-h:
> I disagree with this change because the clock_sync_test is not a configuration option, so doesn't belong in config.h. It is an internal debug check of the boot code and currently it's contained to boot code. If you move the defines you scatter it around more.

I do not agree with this. It is a configuration option after all, even if this just affects the boot code. The intention of a `config.h` should be collecting all this things in on place.

Indanz:
> I do not agree with this. It is a configuration option after all, even if this just affects the boot code.

No, it's not. It's used to enable a test and the test only applies to SMP and MCS. It is not a config option, it should always be enabled and there is no upside to disabling it. At least that was my intention when I wrote the test.

PR #1072, opened 2023-07-03
Remove obsolete guard value and only use guard size for CNodes

andybui01:
CNode guards are currently configured with 2 values:
1. guard values
2. guard sizes

In practice, guard values are not used and only the guard size is actually used. This commit removes the guard value from the kernel in an attempt to simplify the API. I'd like to simplify the process to create CSpaces as well as simplify the manual so that beginners to seL4 have an easier time grasping CNodes and how to configure CSpaces.

Questions:
- What is the proof impact?
- How much of the API should we deprecate. For example, `api_make_guard_skip_word()` in `seL4_libs/libsel4utils` is a helper function that is often used to generate an `seL4_CNode_CapData` type with an empty guard value. Should this function be kept for backwards compatibility or completely deprecated? (the CapData type will be removed in favor of an seL4_Word type)

If people are happy with this change, the manual changes will follow. Also see: https://sel4.discourse.group/t/why-do-cnodes-have-guard-bits/49/3

### DIFF
--- a/include/fastpath/fastpath.h
+++ b/include/fastpath/fastpath.h
@@ -89,7 +89,7 @@ static inline cap_t FORCE_INLINE lookup_fp(cap_t cap, cptr_t cptr)
     word_t cptr2;
     cte_t *slot;
     word_t guardBits, radixBits, bits;
-    word_t radix, capGuard;
+    word_t radix;
 
     bits = 0;
 
@@ -101,16 +101,6 @@ static inline cap_t FORCE_INLINE lookup_fp(cap_t cap, cptr_t cptr)
         guardBits = cap_cnode_cap_get_capCNodeGuardSize(cap);
         radixBits = cap_cnode_cap_get_capCNodeRadix(cap);
         cptr2 = cptr << bits;
-
-        capGuard = cap_cnode_cap_get_capCNodeGuard(cap);
-
-        /* Check the guard. Depth mismatch check is deferred.
-           The 32MinusGuardSize encoding contains an exception
-           when the guard is 0, when 32MinusGuardSize will be
-           reported as 0 also. In this case we skip the check */
-        if (likely(guardBits) && unlikely(cptr2 >> (wordBits - guardBits) != capGuard)) {
-            return cap_null_cap_new();
-        }
 
         radix = cptr2 << guardBits >> (wordBits - radixBits);
         slot = CTE_PTR(cap_cnode_cap_get_capCNodePtr(cap)) + radix;
--- a/include/object/structures_32.bf
+++ b/include/object/structures_32.bf
@@ -74,10 +74,9 @@ block reply_cap(capReplyCanGrant, capReplyMaster, capTCBPtr, capType) {
 -- The user-visible format of the data word is defined by cnode_capdata, below.
 block cnode_cap(capCNodeRadix, capCNodeGuardSize, capCNodeGuard,
                 capCNodePtr, capType) {
-    padding 4
+    padding 22
     field capCNodeGuardSize 5
     field capCNodeRadix 5
-    field capCNodeGuard 18
 
     field_high capCNodePtr 27
     padding 1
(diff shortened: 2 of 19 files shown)


lsf37:
I vaguely remember a discussion a very long time ago about this and had actually (mistakenly) thought we had already done this, but we have not and must have decided against it. I couldn't find any issue tracking this, so this might have been even before we started doing that.

@kevinelp do you remember anything about removing CNode guards?

From the verification side, I don't think the effort would be very high, so this is mainly a question of what we want the API to be. If the only benefit is easier explanation then I agree with Indan that this is better fixed by improving the explanation. I thought there were other benefits, though, at least in that discussion years ago.

This will need an RFC, but we can start a discussion here.

If the feature is actually used in deployed systems, that is a fairly strong argument against removal compared to the benefits listed so far.

andybui01:
@Indanz thanks for the example, makes sense.

@lsf37 yes I agree, there is a reduced argument for this if guard values are actually being used.

I'd like to add that there is, of course, a performance benefit out of this being cptr resolution becomes faster without the guard value check, but this wasn't as important to me as the API change.

Indanz:
> I thought there were other benefits, though, at least in that discussion years ago.

If that discussion is the one @andybui01 linked to in his post, it was mostly about removing guard size. The performance cost for guard values is one extra if-statement on cache-hot data, so probably not really noticeable.

What's confusing in the current documentation is that it behaves as if the guard is part of a CNode, while it is part of a CNode cap, so effectively a property of a slot, not the CNode it points to. It also doesn't say how to set the guard values and sizes, you need to dig into the source or find the seL4lib helper function to figure that out.

Indanz:
The CNode operations can be simplified by getting rid of the source and destination depth arguments: So instead of making it CSpace operations, make it strict CNode operations. That is, `index` becomes a true index, not a relative CPtr.

The depth will depend on the CNode size with no further traversal of sub-CNodes. This would also give more fine grained control over what rights a CNode cap gives in the case of multilevel CSpaces. The destination and source root caps must already be at a depth equivalent to the wordsize.

For a 1-level CSpace this change would be backward compatible. For multilevel CSpaces this change would mean that to build a CSpace you need to use the individual caps of each sub-CNode instead of just one root. But tightening the access that a CNode caps gives is only a good thing if you want to build more complex multilevel CSpace systems.

lsf37:
> > I thought there were other benefits, though, at least in that discussion years ago.
> 
> If that discussion is the one @andybui01 linked to in his post, it was mostly about removing guard size. 

No, I meant even earlier than that, probably before 2009 when the API wasn't even public yet and this all was just an internal research project. In the discussion Andy linked I was already misremembering that one :-)

> The performance cost for guard values is one extra if-statement on cache-hot data, so probably not really noticeable.

Agreed. I don't really remember what we thought back then. Maybe it was just simplification, and after all we did not do the change, so maybe we reached the same conclusion.
 
> What's confusing in the current documentation is that it behaves as if the guard is part of a CNode, while it is part of a CNode cap, so effectively a property of a slot, not the CNode it points to. It also doesn't say how to set the guard values and sizes, you need to dig into the source or find the seL4lib helper function to figure that out.

That is one of the main problems of the current docs (and in places also the tutorials): they pun caps and objects way too much and that ends up being super confusing. It is a bit awkward to always say "cap to the object" instead of just talking about the object, but it's necessary.

kent-mcleod:
Marking this draft.

lsf37:
@andybui01 it looks like we shouldn't do this one. Any concerns about me closing this one?

If there is more need for discussion I'd encourage to put this in the [RFC channel on discourse](https://sel4.discourse.group/c/rfc-discussion/9) as a pre-RFC discussion. Since it is an API change, it would need an RFC first in any case.

andybui01:
> @andybui01 it looks like we shouldn't do this one. Any concerns about me closing this one?
> 
> If there is more need for discussion I'd encourage to put this in the [RFC channel on discourse](https://sel4.discourse.group/c/rfc-discussion/9) as a pre-RFC discussion. Since it is an API change, it would need an RFC first in any case.

I'm fine with that :)

PR #1076, opened 2023-07-24
Read timestamp directly from CLINT for RISC-V platforms

Ivan-Velickovic:
Currently getting the timestamp for RISC-V platforms uses the `rdtime` pseudo-instruction which ends up invoking OpenSBI and then accessing the CLINT memory-mapped `mtime` register. This patch eliminates the use of `rdtime` in favour of directly getting the timestamp for performance.

See https://github.com/Ivan-Velickovic/sel4_riscv_clint_mtime_benchmarks for benchmarking results.

Closes https://github.com/seL4/seL4/issues/942.


Ivan-Velickovic:
Missing the patch to 32-bit RISC-V otherwise I think this is good to go. Unfortunately I don't have access to any 32-bit RISC-V hardware so cannot benchmark/test other than on simulator.

Indanz (approved):
Code looks good to me. If in the future a RISC platform is released where rdtime is faster or better to use, we can add a config option for that then.

axel-h:
> Code looks good to me. If in the future a RISC platform is released where rdtime is faster or better to use, we can add a config option for that then.

Maybe we should add this option now, so we can leave reading `rdtime` as the default and just enable this optimization for the known platforms - even this these are all at the moment. Is setting `CLINT_MTIME_OFFSET` mandatory now, or is this set to `0x00` if nothing is set?

heshamelmatary:
> > Code looks good to me. If in the future a RISC platform is released where rdtime is faster or better to use, we can add a config option for that then.
> 
> Maybe we should add this option now, so we can leave reading `rdtime` as the default and just enable this optimization for the known platforms - even this these are all at the moment. Is setting `CLINT_MTIME_OFFSET` mandatory now, or is this set to `0x00` if nothing is set?

I'd vote for that. The aim was that high-performance hardware shouldn't trap on `rdtime`, the fact that no current hardware supports that doesn't mean it should be the default.

Indanz:
> I'd vote for that. The aim was that high-performance hardware shouldn't trap on rdtime, the fact that no current hardware supports that doesn't mean it should be the default.

I agree it would be better, but disagree that too much effort should be put into this. Clearly performance isn't important to RISCV currently, otherwise this patch wouldn't be necessary to begin with.

Ivan-Velickovic:
> Code looks good to me. If in the future a RISC platform is released where rdtime is faster or better to use, we can add a config option for that then.

How do you know that in the future `rdtime` will be faster in the future?  (genuinely asking) Given that the RISC-V specification says that the instruction is a pseudo-instruction, is it possible for hardware to implement `rdtime` as a proper instruction while still following the ISA (assuming the ISA isn't changed to not make it a pseudo-instruction anymore)?

If we don't know for sure that `rdtime` will ever be optimised, I don't see why we would generalise now?

src/plat/spike/config.cmake
```diff
@@ -30,6 +30,7 @@ if(KernelPlatformSpike)
     declare_default_headers(
         TIMER_FREQUENCY 10000000 PLIC_MAX_NUM_INT 0
         INTERRUPT_CONTROLLER drivers/irq/riscv_plic_dummy.h
+        CLINT_MTIME_OFFSET 0xbff8
```
kent-mcleod, line 33: This offset seems to be the same for every CLINT supported, wouldn't it also work as a #define nearer to `riscv_read_time`?

kent-mcleod:
> Missing the patch to 32-bit RISC-V otherwise I think this is good to go. Unfortunately I don't have access to any 32-bit RISC-V hardware so cannot benchmark/test other than on simulator.

Are you able to add the rv32 changes too? It's likely needing to do something special to turn 32-bit register reads into a 64-bit value?

src/plat/spike/config.cmake
Ivan-Velickovic: Yes I can do that instead, but is it simpler? That is what I had initially but I thought it would be better to have it in `config.cmake` given it is a platform-specific value rather than the kernel source code.

Ivan-Velickovic:
> Are you able to add the rv32 changes too? It's likely needing to do something special to turn 32-bit register reads into a 64-bit value?

Yep, I'll do that today.

src/plat/spike/config.cmake
axel-h: > This offset seems to be the same for every CLINT supported, wouldn't it also work as a #define nearer to `riscv_read_time`?

I think it's because all the current platforms just copy the same SiFive CLINT. If we make platform state this in their DTB, there could be less redundancy,

Ivan-Velickovic: I was told that the memory layout of the SiFive CLINT is not guaranteed to be the same.

Ivan-Velickovic: Oh, do you mean have the mtime offset as a field in the CLINT DTS node or something?

kent-mcleod: Another platform-specific way would be to define a CLINT driver that can bind to all of the platforms with the matching compatibility string and it implements `riscv_read_time` and knows the register offset. But I think if every platform we support has the same CLINT with the same offset then it's more pragmatic to just have less abstraction and move the definitions closer to the code that's using them.

Indanz:
> How do you know that `rdtime` will be faster in the future? (genuinely asking) Given that the RISC-V specification says that the instruction is a pseudo-instruction, is it possible for hardware to implement `rdtime` as a proper instruction while still following the ISA (assuming the ISA isn't changed to not make it a pseudo-instruction anymore)?
> 
> If we don't know for sure that `rdtime` will ever be optimised, I don't see why we would generalise now? I made the patch this way because I assumed that accessing the CLINT would always be faster than `rdtime`, but maybe that's not true.

"Pseudo-instruction" doesn't mean trapped and emulated, it means that although it's described as a separate instruction and the assembler recognises it, it's just a special encoding of another instruction.`rdtime` is a pseudo-instruction because it is a specific invocation of the `CSR` instruction. 

I don't know whether `rdtime` will get faster, but it's bound to happen if people start using RISCV seriously and want higher performance. Trapping has too much overhead and side-effects for any accurate timing. And when `rdtime` is implemented in hardware, using the instruction has way less overhead than a normal memory read which needs to do address translation and check access permissions. That can cause multiple TLB misses and real SDRAM accesses.

Ivan-Velickovic:
> > How do you know that `rdtime` will be faster in the future? (genuinely asking) Given that the RISC-V specification says that the instruction is a pseudo-instruction, is it possible for hardware to implement `rdtime` as a proper instruction while still following the ISA (assuming the ISA isn't changed to not make it a pseudo-instruction anymore)?
> > If we don't know for sure that `rdtime` will ever be optimised, I don't see why we would generalise now? I made the patch this way because I assumed that accessing the CLINT would always be faster than `rdtime`, but maybe that's not true.
> 
> "Pseudo-instruction" doesn't mean trapped and emulated, it means that although it's described as a separate instruction and the assembler recognises it, it's just a special encoding of another instruction.`rdtime` is a pseudo-instruction because it is a specific invocation of the `CSR` instruction.
> 
> I don't know whether `rdtime` will get faster, but it's bound to happen if people start using RISCV seriously and want higher performance. Trapping has too much overhead and side-effects for any accurate timing. And when `rdtime` is implemented in hardware, using the instruction has way less overhead than a normal memory read which needs to do address translation and check access permissions. That can cause multiple TLB misses and real SDRAM accesses.

Yea my bad, I misunderstood/forgot that all `rdtime` is is just a CSR access. The trap into M-mode has nothing really to do with `rdtime` itself but instead the CSR that is accessed when executing `rdtime`.

So what is the conclusion, am I adding a config option to use `rdtime` or not?

axel-h:
> So what is the conclusion, am I adding a config option to use `rdtime` or not?

Can't we just jump on the timer driver concept that exists from ARMv7? Seems on RISC-V we will have these timer drivers then
- SBI API + rdtime (generic default)
- SBI API + SiFive-CLINT-mtime (where we can hard-code the offset then)

heshamelmatary:
> > Code looks good to me. If in the future a RISC platform is released where rdtime is faster or better to use, we can add a config option for that then.
> 
> How do you know that `rdtime` will be faster in the future? (genuinely asking) Given that the RISC-V specification says that the instruction is a pseudo-instruction, is it possible for hardware to implement `rdtime` as a proper instruction while still following the ISA (assuming the ISA isn't changed to not make it a pseudo-instruction anymore)?
> 
> If we don't know for sure that `rdtime` will ever be optimised, I don't see why we would generalise now? I made the patch this way because I assumed that accessing the CLINT would always be faster than `rdtime`, but maybe that's not true.

Some current implementations actually already don't trap (if `mcounteren.tm` == 1) on reading `time` and behave either the same as `cycle` as per the riscv-priv spec [1], or linked directly to the CLINT's `mtime` without trapping [2]. This is likely the case for most single-core (embedded) systems (e.g., Piccolo, Flute, Toooba) without DVFS. In that case, reading from the CLINT's MMIO in S-mode may actually be worse as it would involve the TLB.

[1] "Simple fixed-frequency systems can use a single clock for both cycle counting and wall-clock time."
[2] https://github.com/bluespec/Flute/blob/da86cb05519d2ba0502c41a20c44654baea239bd/src_Core/RegFiles/CSR_RegFile_MSU.bsv#L1319

kent-mcleod:
> Some results for sel4bench for HiFive Unleashed for the MCS kernel, I am not very familiar with the project, please let me know if I should publish more results:

Are any of these benchmarks using `rdtime`? I thought they'd all be reading cycle counters

Ivan-Velickovic:
> > Some results for sel4bench for HiFive Unleashed for the MCS kernel, I am not very familiar with the project, please let me know if I should publish more results:
> 
> Are any of these benchmarks using `rdtime`? I thought they'd all be reading cycle counters

The benchmarks themselves, I'm not sure. The point of the results is to show the same benchmarks with and without this path to the kernel to at least have some idea of how much of an improvement the patch makes.

kent-mcleod:
> The benchmarks themselves, I'm not sure. The point of the results is to show the same benchmarks with and without this patch to the kernel to at least have some idea of how much of an improvement the patch makes.

Oh, then is it possible to show the results for non-mcs too? (Where we'd expect no change)

kent-mcleod:
For MCS we'd expect that there'd be a big drop in all the benchmarks that reprogram the timer when exiting the kernel (which would be the ones that don't use the fastpath).  So my question is for the lines in the table that are showing > 10% reduction, are they all not hitting the fastpath (I can't tell what the last couple seL4_Call and seL4_ReplyRecv) ones are? Maybe they're doing larger message lengths beyond the fastpath limit?

kent-mcleod:
> Can't we just jump on the timer driver concept that exists from ARMv7? Seems on RISC-V we will have these timer drivers then

I don't think you can tell from the device tree whether the CPU will trap the rdtime though and so mechanism of selecting the driver based on the compatibility string won't be sufficient and still require a per-platform compile config.  So maybe we should just define this config now and leave the default to call rdtime directly?

Ivan-Velickovic:
> Oh, then is it possible to show the results for non-mcs too? (Where we'd expect no change)

Here you go https://github.com/Ivan-Velickovic/sel4_riscv_clint_mtime_benchmarks.

Ivan-Velickovic:
> For MCS we'd expect that there'd be a big drop in all the benchmarks that reprogram the timer when exiting the kernel (which would be the ones that don't use the fastpath). So my question is for the lines in the table that are showing > 10% reduction, are they all not hitting the fastpath (I can't tell what the last couple seL4_Call and seL4_ReplyRecv ones are?) Maybe they're doing larger message lengths beyond the fastpath limit?

Yea my bad I should have made the table clearer, hopefully it's clearer in the repo's README.

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -9,19 +9,23 @@
 #include <util.h>
 #include <arch/model/smp.h>
 #include <stdint.h>
+#include <plat/machine/devices_gen.h>
+#include <plat/platform_gen.h>
+
+#define CLINT_MTIME_LO_OFFSET 0xbff8
+#define CLINT_MTIME_HI_OFFSET 0xbffc
```
Indanz, line 16: Shouldn't this use `CLINT_MTIME_OFFSET` for the low offset and `CLINT_MTIME_OFFSET + 4` for the high offset? (For little-endian riscv platforms, no idea if there are any big-endian ones.)

gernotheiser:
Comment on benchmarking results: Presently we're running our benchmarks on a SoC that doesn't have ASIDs, which makes AS crossings quite expensive on RISC-V. Once we have supported platform with ASIDS, numbers will drop and should be more in line with the other ISAs. This will result in increasing the relative impact of the timer cost, i.e. making the change more worthwhile

Ivan-Velickovic:
I have added the `KernelRiscvUseClintMtime` configuration option. Hopefully this is the right level of generalisation everyone is happy with.

include/arch/riscv/arch/32/mode/machine.h
Ivan-Velickovic: Yes that's right but when I had it that way initially, the `FPU0001` test hung. Which is weird, maybe I did something else wrong. Reading the timestamp into a `uint64_t` is simpler though so I've just done.

Indanz (approved):
Looks good to me.

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -9,9 +9,20 @@
 #include <util.h>
 #include <arch/model/smp.h>
 #include <stdint.h>
+#include <plat/machine/devices_gen.h>
+
+/*
+ * Currently all RISC-V 32-bit platforms supported have the mtime register
+ * mapped at the same offset of the base address of the CLINT.
+ */
+#define CLINT_MTIME_OFFSET 0xbff8
```
axel-h, line 18: Could you move this into `#ifdef CONFIG_RISCV_USE_CLINT_MTIME` also?

Ivan-Velickovic: Resolved

Ivan-Velickovic:
Rebased on master, let me know if there's anything else I need to do ....

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -9,9 +9,22 @@
...
+#endif
 
 static inline uint64_t riscv_read_time(void)
 {
+#ifdef CONFIG_RISCV_USE_CLINT_MTIME
+    uint64_t n = *(volatile uint64_t *)(CLINT_PPTR + CLINT_MTIME_OFFSET);
+    return n;
+#else
```
kent-mcleod, line 27: It seems like this read can produce invalid values because the compiler breaks it into two 32-bit reads (each `lw` instruction is reading half of the value from the register):
```
{
#ifdef CONFIG_RISCV_USE_CLINT_MTIME
    uint64_t n = *(volatile uint64_t *)(CLINT_PPTR + CLINT_MTIME_OFFSET);
ff800104:       ffc0c7b7                lui     a5,0xffc0c
ff800108:       ff87a703                lw      a4,-8(a5) # ffc0bff8 <ki_end+0x3f0ff8>
/**
   DONT_TRANSLATE
 */
BOOT_CODE void initTimer(void)
{
    sbi_set_timer(riscv_read_time() + RESET_CYCLES);
ff80010c:       6515                    lui     a0,0x5
ff80010e:       ffc7a783                lw      a5,-4(a5)
```

This probably needs to do a similar thing as the register reads by explicitly requesting a sequence of 32-bit reads and detecting overflow.

Ivan-Velickovic: Good point, I will change that then.

Ivan-Velickovic: I've pushed a commit but there's an issue, see the commit message. I will have to investigate properly when I have more time.

include/arch/riscv/arch/32/mode/machine.h
```diff
@@ -16,16 +16,24 @@
  * Currently all RISC-V 32-bit platforms supported have the mtime register
  * mapped at the same offset of the base address of the CLINT.
  */
-#define CLINT_MTIME_OFFSET 0xbff8
+#define CLINT_MTIME_OFFSET_LO 0xbffc
+#define CLINT_MTIME_OFFSET_HI 0xbff8
```
kent-mcleod, line 20: Are these swapped around? IE LO should be `0xbff8` and HI should be `0xbffc`

include/arch/riscv/arch/32/mode/machine.h
kent-mcleod: https://github.com/seL4/seL4/pull/1076/commits/5cf253ac219e8252d06d7f3ba38dac6ab7e5b33a#r1290876449

include/arch/riscv/arch/32/mode/machine.h
Ivan-Velickovic: Turns out I have had a misconception about endianness, my bad.

kent-mcleod: So it works now?

kent-mcleod (approved):

include/arch/riscv/arch/32/mode/machine.h
Ivan-Velickovic: Yes

Ivan-Velickovic: The CI for Spike 32-bit will confirm this.

Indanz:
It would be nice to be able to restart only the ODROID HW run, instead of everything.

lsf37:
> It would be nice to be able to restart only the ODROID HW run, instead of everything.

If it's just re-running the job on the same commit, that should be possible (under "Rerun Jobs" there should be a "Rerun failed jobs only")

kent-mcleod:
> It would be nice to be able to restart only the ODROID HW run, instead of everything.

I'm going to assume the ODROID failure is spurious as it's not a riscv platform and this change set only modifies behavior for riscv platforms.

PR #1084, opened 2023-08-02
riscv: Use jumps instead of branches jumping to C

heshamelmatary:
Using branch instructions only allows -+4 KiB offseting range from the current PC. If the C code is placed (depending on linkers and address layout) far more than -+4 KiB, linking will fail (e.g., R_RISCV_BRANCH out of range: 4735 is not in [-2048, 2047]; references c_handle_fastpath_call.

This commit accounts for this case and uses jumps when jumping to C code that should give enough offset range (-+ 1 MiB) and make the assembly code more portable.


kent-mcleod:
> This commit accounts for this case and uses jumps when jumping to C code that should give enough offset range (-+ 1 MiB) and make the assembly code more portable.

Because this is fastpath entry code we should be assuming that the code is fairly tightly integrated so it's not expected that there's > 4Kib offset between these two points in the code.

heshamelmatary:
> > This commit accounts for this case and uses jumps when jumping to C code that should give enough offset range (-+ 1 MiB) and make the assembly code more portable.
> 
> Because this is fastpath entry code we should be assuming that the code is fairly tightly integrated so it's not expected that there's > 4Kib offset between these two points in the code.

Unfortunately that assumption doesn't always hold for different toolchains/linkers (or with different flags e.g., w/o compressed instruction or linker relaxation), especially that both traps and fast path code are in .text, and shared with the remaining kernel code; no custom handling to ensure that traps and the fastpath code fit within 4KiB memory (e.g., placed in a specific section?). So you rely on whatever the linker does.

For instance, the following is the output when building with llvm/lld:

```
➜  build llvm-readelf -s kernel/kernel.elf | grep  "trap_entry"            
 19593: ffffffff84004000     0 NOTYPE  GLOBAL DEFAULT     5 trap_entry
➜  build llvm-readelf -s kernel/kernel.elf | grep  "c_handle_fastpath_call"
 19594: ffffffff840065c0    58 FUNC    GLOBAL DEFAULT     5 c_handle_fastpath_call
```

src/arch/riscv/traps.S
```diff
@@ -115,14 +115,17 @@ handle_syscall:
...
+1:
   li t3, SYSCALL_REPLY_RECV
 #ifdef CONFIG_KERNEL_MCS
   /* move reply to 3rd argument */
   mv a2, a6
 #endif
-  beq a7, t3, c_handle_fastpath_reply_recv
+  bne a7, a3, 1f
+  j c_handle_fastpath_reply_recv
+1:
```
Indanz, line 128: I would give this label a different name, it weird to have two labels with the same name in the same function.

kent-mcleod:
> If you really want to place traps and fastpath code in adjacent memory, I'd follow what Arm is doing, but also still add this extra instruction for portability.

Well my preference would be to add the section annotations like Arm has to control the locations more explicitly to be within 4KiB.  Can you create an issue about making this potential optimization in the future?

src/arch/riscv/traps.S
```diff
@@ -115,14 +115,19 @@ handle_syscall:
 
 #ifdef CONFIG_FASTPATH
   li t3, SYSCALL_CALL
-  beq a7, t3, c_handle_fastpath_call
+  bne a7, a3, .Ltest_replyrecv
```
kent-mcleod, line 118: Should these changes be `t3` instead of `a3`?

heshamelmatary: Opps, fixed

heshamelmatary:
> > If you really want to place traps and fastpath code in adjacent memory, I'd follow what Arm is doing, but also still add this extra instruction for portability.
> 
> Well my preference would be to add the section annotations like Arm has to control the locations more explicitly to be within 4KiB. Can you create an issue about making this potential optimization in the future?

Sure I can create an issue. Is there any other optimisation benefit from having this code within 4KiB other than spatial cache locality?

kent-mcleod:
> Sure I can create an issue. Is there any other optimisation benefit from having this code within 4KiB other than spatial cache locality?

That and the ability to encode the jumps with less instructions which marginally reduces instructions executed in the fastpath and also marginally reduces the code size of the kernel... 1 instruction at a time.

kent-mcleod (approved):

PR #1088, opened 2023-08-10
libsel4: Make bootinfo consistent

Furao:
Some slot positions in the rootnode would depend on configuration. However that makes it difficult to add new root caps, especially if multiple caps only exist based on configuration. Make all caps always there, but null if not configured.

### DIFF
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -23,14 +23,10 @@ enum {
     seL4_CapBootInfoFrame       =  9, /* bootinfo frame cap */
     seL4_CapInitThreadIPCBuffer = 10, /* initial thread's IPC buffer frame cap */
     seL4_CapDomain              = 11, /* global domain controller cap */
-    seL4_CapSMMUSIDControl      = 12,  /*global SMMU SID controller cap, null cap if not supported*/
-    seL4_CapSMMUCBControl       = 13,  /*global SMMU CB controller cap, null cap if not supported*/
-#ifdef CONFIG_KERNEL_MCS
-    seL4_CapInitThreadSC        = 14, /* initial thread's scheduling context cap */
+    seL4_CapSMMUSIDControl      = 12, /* global SMMU SID controller cap, null cap if not supported */
+    seL4_CapSMMUCBControl       = 13, /* global SMMU CB controller cap, null cap if not supported */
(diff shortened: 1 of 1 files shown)


Furao:
This change was done to address the concern in this thread: https://github.com/seL4/seL4/pull/701/files#r1073743999

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -70,9 +66,7 @@ typedef struct seL4_BootInfo {
     seL4_SlotRegion   extraBIPages;    /* caps for any pages used to back the additional bootinfo information */
     seL4_Word         initThreadCNodeSizeBits; /* initial thread's root CNode size (2^n slots) */
     seL4_Domain       initThreadDomain; /* Initial thread's domain ID */
-#ifdef CONFIG_KERNEL_MCS
     seL4_SlotRegion   schedcontrol; /* Caps to sched_control for each node */
```
lsf37, line 69: If we keep this slot unconditionally, we should probably add a comment what we expect to see here in the non-MCS case.

Furao: I wasn't really sure if it should be left in our not. Should we just leave this as conditional?

kent-mcleod:
The way that the initial caps enum is implemented has always bugged me.  Skipping slots in the CNode to make it easier to handle the various configurations of possible optional initial caps is kinda annoying.  I'm happy with your proposed change because it's consistent with the other optional cap types that aren't #ifdef guarded.

I reckon that we should remove the Enum though and instead add a lot of uint8_t named fields to the BootInfo_t struct which can be set to the correct seL4_Cptr index if the cap exists, but left at 0 otherwise.  And the kernel boot code just inserts the initial caps into the CNode in initialization order.

libsel4/include/sel4/bootinfo_types.h
Furao: I just put the ifdef back in.

Furao:
> I reckon that we should remove the Enum though and instead add a lot of uint8_t named fields to the BootInfo_t struct which can be set to the correct seL4_Cptr index if the cap exists, but left at 0 otherwise. And the kernel boot code just inserts the initial caps into the CNode in initialization order.

Probably worth creating an issue or an RFC to track this.

kent-mcleod (approved):

PR #1077, opened 2023-07-27
Minor changes to enable linking with lld

heshamelmatary:

PR #1092, opened 2023-08-14
riscv: place traps and fastpath code adjacently

### LINKED ISSUE #1091: riscv: place traps and fastpath code adjacently
Follow Arm's code where it tries to place traps and vector
code adjacently in a 4KiB page to optimise performance
(through spatial cache locality).

This will also allow reducing the number of instructions
in the fastpath code by one by using conditional branches instead
of a check followed by a jump instruction. Though, if the future traps
or fastpath code got bigger than 4 KiB, this may fail linking again.

heshamelmatary:
riscv: place traps and fastpath code adjacently
    
    Follow Arm's code where it tries to place traps and vector
    code adjacently in a 4KiB page to optimise performance
    (through spatial cache locality).
    
    Closes #1091

### DIFF
--- a/include/arch/riscv/arch/kernel/traps.h
+++ b/include/arch/riscv/arch/kernel/traps.h
@@ -25,22 +25,22 @@ void c_handle_fastpath_reply_recv(word_t cptr, word_t msgInfo, word_t reply)
 #else
 void c_handle_fastpath_reply_recv(word_t cptr, word_t msgInfo)
 #endif
-VISIBLE NORETURN;
+VISIBLE NORETURN SECTION(".text.fastpath");
 
 void c_handle_fastpath_call(word_t cptr, word_t msgInfo)
-VISIBLE NORETURN;
+VISIBLE NORETURN SECTION(".text.fastpath");
 
 void c_handle_syscall(word_t cptr, word_t msgInfo, syscall_t syscall)
-VISIBLE NORETURN;
+VISIBLE NORETURN SECTION(".text.traps");
 
 void c_handle_interrupt(void)
-VISIBLE NORETURN;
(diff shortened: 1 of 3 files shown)


heshamelmatary:
Depends on https://github.com/seL4/seL4/pull/1077 , hence a draft. I'll rebase once #1077 is merged.

I haven't benchmarked that change, so if it's possible to do that (using sel4bench) on your HiFive board and paste results here, that'll be great.

Indanz:
Looks good to me, but haven't compiled it though, just checked that you didn't miss any.

lsf37:
@heshamelmatary if you rebase now, I think the sel4bench trigger should work

lsf37:
So the raw sel4bench results are now available in the summary tab of the [sel4bench run here](https://github.com/seL4/seL4/actions/runs/6718245240) and for comparison the latest raw sel4bench results from the master branch are [here](https://github.com/seL4/sel4bench/actions/runs/6722000180).

The change seems to be mostly in the noise (i.e. mean shifted by values smaller than the standard deviation). Some have gotten slightly better it seems, e.g. `seL4_ReplyRecv (server -> client)` in the same address space has a new mean of `334.8125` with std dev `1.69` vs the old `340` with std dev `2.19`. Some means have gone up, but within the std dev of at least one of the measurements (the std dev is not necessarily stable).

It'd be handy to get a more visual comparison of the json results automatically from CI, but that's a project for a different days. (If anyone is keen, let me know, though)

lsf37:
Adding a brief proof check to make sure the C parser isn't getting confused by the new attribute (I don't think it should).

lsf37 (approved):
Ok, I'm happy with this. The MCS proof failure is unrelated (the recent kernel build config  change)

PR #1095, opened 2023-08-17
AArch64: updates for verification compatibility

Xaphiosis:
These are the changes I needed to introduce to get the C kernel parsed and imported into Isabelle:
- aarch64: avoid void-return statements
- aarch64: DONT_TRANSLATE 32-bit __asm__
- aarch64: don't take address of local vars
- arm: pde_C type is only available on aarch32

I was a bit surprised by the `&cap` where `cap` is local. Well-known limitation of the verification infrastructure.
The 32-bit `__asm__` is a bit disappointing on the verification side (the limitation is irritating, but tracking down the error is terrible), maybe we should think of what we can do about it next time binary verification comes up.

### DIFF
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -487,61 +487,82 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
 {
     switch (reg_index) {
     case seL4_VCPUReg_SCTLR:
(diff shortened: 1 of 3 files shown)


lsf37 (approved):
Looks good.

The `&cap` is probably old and wasn't spotted because of that, but it's a bit sad that the C parser check doesn't flag it. Not surprising, because it's a later analysis phase in the parser, but still sad. I wonder if we could let the standalone parser to run a few of those later phases as well.

PR #1096, opened 2023-08-25
clarify names to reflect the exact content

axel-h:
Follow-up from https://github.com/seL4/seL4/pull/922. In the code we are sometimes dealing with us/ms and sometimes with ticks. Give the constants a clear name to indicate when they hold ticks, so there is less chance to confuse this.


axel-h:
Can we merge this?

Indanz:
To me replacing one unitless name with another unitless name doesn't seem like a clear improvement, and when in doubt I don't merge. So I'm waiting for feedback from others before doing anything.

Indanz:
Closing because it's not a clear improvement and because of lack of interest.

PR #1097, opened 2023-08-28
manual: adjust AArch64 VSpace section for RFC-10

Ivan-Velickovic:


manual/parts/vspace.tex
```diff
@@ -109,22 +109,19 @@ \subsubsection{AArch32}
...
-9 or 10 bits. The macro \texttt{seL4\_VSpaceIndexBits} makes this value available under a generic
-name. The table below shows the four-level configuration. On three-level configurations, the
-invocations for \texttt{seL4\_ARM\_VSpaceObject} capabilities are instead available on
-\obj{PageUpperDirectory} caps.
+Regardless of the configuration, the VSpace object is implemented by the \obj{PageTable} object.
+To help with writing code that is generic, \texttt{libsel4} provides the macro \texttt{seL4\_ARM\_VSpaceObject}
+that expands to the correct corresponding object across architectures. All intermediate paging
+structures are indexed by 9 bits of the virtual address. Depending on configuration, the
+top-level object is indexed by either 9 or 10 bits. The macro \texttt{seL4\_VSpaceIndexBits}
+makes this value available under a generic name. The table below shows the four-level configuration.
```
Ivan-Velickovic, line 117: Should we have another table for the three-level configuration for completeness?

manual/parts/vspace.tex
```diff
@@ -109,22 +109,19 @@ \subsubsection{AArch32}
...
-generic in this distinction, \texttt{libsel4} provides the macro \texttt{seL4\_ARM\_VSpaceObject}
-that expands to the correct corresponding object. All intermediate paging structures are indexed by
-9 bits of the virtual address. Depending on configuration, the top-level object is indexed by either
-9 or 10 bits. The macro \texttt{seL4\_VSpaceIndexBits} makes this value available under a generic
-name. The table below shows the four-level configuration. On three-level configurations, the
-invocations for \texttt{seL4\_ARM\_VSpaceObject} capabilities are instead available on
-\obj{PageUpperDirectory} caps.
+Regardless of the configuration, the VSpace object is implemented by the \obj{PageTable} object.
+To help with writing code that is generic, \texttt{libsel4} provides the macro \texttt{seL4\_ARM\_VSpaceObject}
+that expands to the correct corresponding object across architectures. All intermediate paging
```
Ivan-Velickovic, line 114: It might be better to just omit the mention of `seL4\_ARM\_VSpaceObject`, RISC-V doesn't mention it either.

Indanz: How else are you supposed to find out about it?

manual/parts/vspace.tex
Indanz: Isn't the three-level exactly the same, but with one level less? AFAIK the only difference used to be at which level the page upper directory pointed at, but now it's all unified it's just all the same type of object.

Code can be very level agnostic now, it can just create and map PageTable objcts on demand when a mapping fails.

Indanz (approved):
If you fix the alignment in the table I'm happy with these changes.

Ivan-Velickovic:
> If you fix the alignment in the table I'm happy with these changes.

Yes, sorry that was a bit sloppy of me.

manual/parts/vspace.tex
Ivan-Velickovic: I just thought given there are only two possible configurations, having each one outlined in the table would not take up too much space will also being complete by showing every possible paging setup.

But it's just a minor point.

manual/parts/vspace.tex
Ivan-Velickovic: I was trying to say that there is no need for `seL4_ARM_VSpaceObject` now that the top-level page table object is always the same on AArch64.

Ivan-Velickovic: and if there is no need for it, why should the manual mention it?

lsf37: Sorry for the drive-by comment, I haven't had time to read the full change yet, but  `Regardless of the configuration, the VSpace object is implemented by the \obj{PageTable} object` is not true for AArch64 -- depending on config these can be different objects and object sizes.

Ivan-Velickovic: What other objects is a VSpace represented by?

Ivan-Velickovic: Also what I said here with: "expands to the correct corresponding object across architectures" is wrong obviously since `seL4_ARM_VSpaceObject` is specific to ARM.

lsf37: > What other objects is a VSpace represented by?

VSpaceObject is its own type in AArch64, mostly the same but not quite the same as a PageTableObject. There are other operations available on it and its size can be different from a PageTableObject.

Ivan-Velickovic: Okay yep, thanks for clearing up my confusion. The thing I don't understand is why the current manual says the following: `The VSpace object is therefore implemented either by the \obj{PageGlobalDirectory} object (4 level
configs) or the \obj{PageUpperDirectory} object (3 level configs)`. I was just trying to adjust this sentence with this PR to remove PUD/PGD mentions.

Should we just remove this sentence to avoid confusing an actual VSpace object with what it is "implemented by"?

Indanz: I think it's good to mention that the root page table is special and a different object, namely `VSpaceObject` and not a `PageTableObject`. And describing the page table layout options for Aarch64 is useful too. What about:

```
Depending on configuration, Arm AArch64 processors have page-table structures with 3 or 4 levels.
The VSpace object is \texttt{seL4\_ARM\_VSpaceObject}, which is a distinct object type used for the 
top level page table. All intermediate paging structures are indexed by 9 bits of the virtual address 
and  are \obj{PageTable} objects. Depending on configuration, the top-level object is indexed by either 
9 or 10 bits. The macro \texttt{seL4\_VSpaceIndexBits} makes this value available under a generic name.
The table below shows the four-level configuration.
```

Ivan-Velickovic: Thanks, I've adjusted the wording to this.

manual/parts/vspace.tex
```diff
@@ -109,23 +109,19 @@ \subsubsection{AArch32}
...
+The VSpace object is \texttt{seL4\_ARM\_VSpaceObject}, which is a distinct object type used for the
+top level page table. All intermediate paging structures are indexed by 9 bits of the virtual
+address and are \obj{PageTable} objects. Depending on configuration, the top-level object is
+indexed by either 9 or 10 bits. The macro \texttt{seL4\_VSpaceIndexBits} makes this value available
+under a generic name. The table below shows the four-level configuration.
 
 \begin{tabularx}{\textwidth}{Xlll} \toprule
 \emph{Object}                    & \emph{Address Bits} & \emph{Level} & \emph{Methods} \\ \midrule
-    \texttt{PageGlobalDirectory}/\texttt{seL4\_ARM\_VSpaceObject}
+    \texttt{PageTable}/\texttt{seL4\_ARM\_VSpaceObject}
```
Indanz, line 120: Shouldn't this be just `seL4_ARM_VSpaceObject`? With the old code you could argue that `PageGlobalDirectory` is the same object, but that doesn't hold for `PageTable`.

Ivan-Velickovic: Yes, my bad.

alwin-joshy (approved):

PR #1098, opened 2023-08-30
VT-d: add remappable MSI support

abrandnewusername:
This pull request does:
- add interrupt remapping table features for VT-d, and support remappable MSIs.
- minor fixes

and it also exposes some potential future works including:
- converts ioapic into the remappable format
- converts serial and pit interrupts from compatibility format to remappable format
- implements the queued invalidation feature and replaces the legacy invalidation methods that are used by iotlb and context tables


src/plat/pc99/machine/intel-vtd.c
```diff
@@ -249,7 +249,7 @@ static void vtd_process_faults(drhu_id_t i)
             address[0] = vtd_read32(i, fr_reg);
             reason = vtd_read32(i, fr_reg + 12) & FR_MASK;
 
-            printf("IOMMU: DMA %s page fault ", fault_type ? "read" : "write");
+            printf("IOMMU 0x%x: DMA %s page fault ", i, fault_type ? "read" : "write");
```
axel-h, line 252: It might be better to not put this change into the commit that does the renaming.

Indanz: Agreed it's better, but life's too short for that. The amount of hassle far outweighs the gain in this case, so I'd say just leave it this time. (I checked that `i` is always 32-bits, so it should be fine.)

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +278,33 @@ void vtd_handle_fault(void)
...
+    printf("IOMMU: adding interrupt remapping table entry %lu for vector %lu\n", interrupt_index, vector);
+
+    /* Remapping hardware units are configured to share the interrupt-remapping table,
+     * so we just read the table address from the first unit
+     */
+    if (x86KSnumDrhu == 0) {
+        return;
+    }
+    vtd_irte_t *irt = (vtd_irte_t *) paddr_to_pptr(vtd_read64(0, IRTA_REG) & ~0xfff);
+    vtd_irte_t *vtd_irte_slot = irt + interrupt_index;
```
axel-h, line 292: instead of explicit pointer arithmetic, can't we use `irt[interrupt_index] = vtd_irte_new(...)' ?

Indanz:
I assume we have at least one IOMMU configuration in CI. I don't think it's easy to add tests to sel4test for this, but this should at least check that nothing broke for non-VM code.

src/plat/pc99/machine/intel-vtd.c
axel-h: Sorry, but non-clean commits will cost you much more lifetime in the long run when you try to rebase or cherry-pick and run into conflict. Also, the amount of work doing this is rather trivial.

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +278,33 @@ void vtd_handle_fault(void)
     }
 }
 
+void vtd_create_irte(word_t interrupt_index, word_t vector, word_t pci_bus, word_t pci_dev, word_t pci_func)
```
Indanz, line 281: These two new functions should be behind an `#ifdef CONFIG_IOMMU`, otherwise the build fails with:

>  /github/workspace/kernel/src/plat/pc99/machine/intel-vtd.c:291:5: error: unknown type name 'vtd_irte_t'; did you mean 'vtd_rte_t'?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -88,6 +104,10 @@
 
 #define N_VTD_CONTEXTS 256
 
+/* The amount of IRTEs that a 4K IRT can fit */
+#define VTD_IRT_SIZE    0x7
```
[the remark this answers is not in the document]
wom-bat: I think it just needs a variable rename.

wom-bat:
Can you pleae check that CVE-2013-3495 and CVE-2011-1898 are addressed by this patch (those CVEs are for a parallel situation with Xen)

wom-bat (changes requested):
There are a few minor changes; plus a chack against the CVEs for IRQ remapping to check against.

src/arch/x86/object/interrupt.c
```diff
@@ -175,6 +174,9 @@ exception_t Arch_decodeIRQControlInvocation(word_t invLabel, word_t length, cte_
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+#ifdef CONFIG_IOMMU
```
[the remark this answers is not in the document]
abrandnewusername: I am following the convention of the existing VT-d code. I suggest we can do some refactoring for other VT-d functions?

abrandnewusername:
> Can you pleae check that [CVE-2013-3495](https://github.com/advisories/GHSA-6hmm-m7qc-9hw4) and [CVE-2011-1898](https://github.com/advisories/GHSA-567h-gr95-57mc) are addressed by this patch (those CVEs are for a parallel situation with Xen)

For [CVE-2013-3495](https://github.com/advisories/GHSA-6hmm-m7qc-9hw4), in our implementation, a malformed MSI does cause an Interrupt Remapping Fault but does not panic the kernel. I can also disable the fault if that's desirable.

For [CVE-2011-1898](https://github.com/advisories/GHSA-567h-gr95-57mc), we try to prevent it by adding the remappable MSI, however, since serial and pit are using compatibility format ioapic, I am not able to disable compatibility format interrupts, which means that the users are still able to craft compatibility format MSIs that bypass the remapping table.

src/arch/x86/object/interrupt.c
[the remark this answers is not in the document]
wom-bat: Yes, let's refactor *after* the stuff gets merged.

src/plat/pc99/machine/intel-vtd.c
wom-bat: Did you miss this comment, @abrandnewusername ?

wom-bat (changes requested):
Indan's comment on the need for #ifdeffing a couple of functions needs to be addressed.

src/plat/pc99/machine/intel-vtd.c
abrandnewusername: I forgot to add the IRTE structure for 32 bit. Fixed now.

wom-bat (commented):
This looks good to me.  But I'm not a great expert in IOMMU/VT-D ... I'd like someone who is to review.

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -451,8 +529,44 @@ BOOT_CODE static bool_t vtd_enable(cpu_id_t cpu_id)
...
 
-        printf(" enabled\n");
+        status = vtd_read32(i, GSTS_REG);
+        status |= BIT(IRE);
+        /* Enable interrupt-remapping */
+        vtd_write32(i, GCMD_REG, status);
+        /* Wait for interrupt-remapping operation to complete by polling
+         * IRES bit from GSTS_REG
+         */
+        while (!((vtd_read32(i, GSTS_REG) >> IRES) & 1));
```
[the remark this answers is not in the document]
Indanz: I thought it's well established style to do it with the semicolon on the same line as the while, your alternative looks horrible and IMHO confusing and cluttered, especially if there are multiple while loops close to each other (which happens sometimes when doing hardware stuff). I would have left a blank line after the while though.

Indanz:
> For [CVE-2013-3495](https://github.com/advisories/GHSA-6hmm-m7qc-9hw4), in our implementation, a malformed MSI does cause an Interrupt Remapping Fault but does not panic the kernel (see `vtd_handle_fault`). I can also disable the fault if that's desirable.

The problem seems to be that if System Error Reporting (SERR) is enabled, such malformed transactions raise a PCI System Error NMI instead of a normal vtd fault. SERR may not make it to seL4, but might halt the whole system, depending on host configuration. See https://www.openwall.com/lists/oss-security/2013/08/20/8

Indanz:
> For [CVE-2011-1898](https://github.com/advisories/GHSA-567h-gr95-57mc), we try to prevent it by adding the remappable MSI, however, since serial and pit are using compatibility format ioapic, I am not able to disable compatibility format interrupts, which means that the users are still able to craft compatibility format MSIs that bypass the remapping table.

If it's an all or nothing setting then a work-around is needed for serial and timers, because hardware pass-through without remapping isn't secure. What is Xen's solution in this regard? Looking around it seems possible to remap IOAPIC interrupts, why can't that be done for serial and PIT?

If compatibility format can be enabled only for PIT and serial, that should be relatively safe: They are not DMA capable bus masters and are not PCI devices, so they can't be used for the attacks mentioned in the paper. That said, compatibility format is only allowed if x2APIC is disabled. So switching to x2APIC should make it possible to remap serial and timer interrupts too.

wom-bat:
> > For [CVE-2013-3495](https://github.com/advisories/GHSA-6hmm-m7qc-9hw4), in our implementation, a malformed MSI does cause an Interrupt Remapping Fault but does not panic the kernel (see `vtd_handle_fault`). I can also disable the fault if that's desirable.
> 
> The problem seems to be that if System Error Reporting (SERR) is enabled, such malformed transactions raise a PCI System Error NMI instead of a normal vtd fault. SERR may not make it to seL4, but might halt the whole system, depending on host configuration. See https://www.openwall.com/lists/oss-security/2013/08/20/8

I think that fixing that is out of scope for this patch.  To do the job _properly_ would mean teaching seL4 about SERR NMIs --- and given that most of the time we're targetting embedded systems (even for X86) I reckon just making sure that this is noted somewhere obvious would be enough.  Malformed MSIs are only going to be generated by buggy code (which will be picked up during debugging), and malicious code (which *is* an issue, but one that's not made worse by this.)

src/arch/x86/object/interrupt.c
```diff
@@ -175,6 +175,9 @@ exception_t Arch_decodeIRQControlInvocation(word_t invLabel, word_t length, cte_
...
+        vtd_create_irte(handle, vector, pci_bus, pci_dev, pci_func);
```
Indanz, line 179: Does CONFIG_IOMMU imply VT-d support? What if the CPU has IOMMU support, but no VT-d support?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +293,38 @@ void vtd_handle_fault(void)
...
+    }
+
+    printf("IOMMU: adding IRTE %lu for vector %lu.\n", interrupt_index, vector);
+
+    /* Remapping hardware units are configured to share the interrupt-remapping table,
+     * so we just read the table address from the first unit
+     */
+    if (x86KSnumDrhu == 0) {
+        return;
+    }
```
Indanz, line 311: This check should be done first, otherwise we may get irrelevant and possibly misleading printfs.

Indanz (commented):
Assuming this doesn't break x86 on CPUs not supporting VT-d, and everyone is happy with the code, I think it's ready to be merged?

src/arch/x86/object/interrupt.c
abrandnewusername: We don't yet support AMD-V, so yes, CONFIG_IOMMU implies VT-d

abrandnewusername:
Updates: 
- refactor my MSI code to produce meaningful error messages
- add remappable IOAPIC support. Ideally I should make a new pull request for remappable IOAPIC, but I think it makes more sense to put the commit here since the implementation of remappable IOAPIC is similar to remappable MSI

include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
```diff
@@ -45,4 +45,23 @@ block vtd_pte {
     field       read                1
 }
 
+-- Intel VT-d Interrupt Remapping Table Entry for Remapped Interrupts
```
Indanz, line 48: The "for Remapped Interrupts" part is not necessary, that's obvious because it's the interrupt remapping table.

include/plat/pc99/plat/64/plat_mode/machine/hardware.bf
```diff
@@ -42,4 +42,23 @@ block vtd_pte {
...
+    padding                         7
+    field       delivery_mode       3
+    field       trigger_mode        1
+    field       redir_hint          1
+    field       dest_mode           1
+    field       fault_disable       1
+    field       present             1
+}
+
 #endif
```
Indanz, line 64: This seems the same as the 32-bit one, perhaps put one version in include/arch/x86/arch/object/structures.bf instead?

include/plat/pc99/plat/machine/ioapic.h
```diff
@@ -10,6 +10,8 @@
 #include <plat/machine.h>
 #include <arch/types.h>
 
+#define MAX_IOAPIC  (IOAPIC_IRQ_LINES * MAX(1, CONFIG_MAX_NUM_IOAPIC))
```
Indanz, line 13: One IOAPIC is allowed if there is no IOAPIC?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +294,103 @@ void vtd_handle_fault(void)
...
+    /* Remapping hardware units are configured to share the interrupt-remapping table,
+     * so we just read the table address from the first unit
+     */
+    vtd_irte_t *irt_pptr = (vtd_irte_t *) paddr_to_pptr(vtd_read64(0, IRTA_REG) & IRTA_REG_IRTA_MASK);
+    /* Sanity checking whether the Interrupt Remapping Table is initialized */
+    assert(irt_pptr != paddr_to_pptr(0));
+
+    word_t entry;
+    for (entry = 0; entry <= MAX_IOAPIC && vtd_irte_ptr_get_present(&irt_pptr[entry]); entry++);
```
Indanz, line 310: Shouldn't this be `entry < MAX_IOAPIC`?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +294,103 @@ void vtd_handle_fault(void)
...
+                          0, /* Source-id Qualifier */
+                          0, /* Source Identifier */
+                          delivery_cpu, /* Destination ID */
+                          vector, /* Vector */
+                          0, /* IRTE Mode, value 0 indicates remapped IRQ */
+                          0, /*  Delivery Mode */
+                          level, /* Trigger Mode */
+                          0, /* Redirection Hint */
+                          0, /* Destination Mode */
+                          false, /*  Fault Processing Disable */
```
Indanz, line 323: What does it mean that fault processing is disabled?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +294,103 @@ void vtd_handle_fault(void)
...
+    /* Remapping hardware units are configured to share the interrupt-remapping table,
+     * so we just read the table address from the first unit
+     */
+    vtd_irte_t *irt_pptr = (vtd_irte_t *) paddr_to_pptr(vtd_read64(0, IRTA_REG) & IRTA_REG_IRTA_MASK);
+    /* Sanity checking whether the Interrupt Remapping Table is initialized */
+    assert(irt_pptr != paddr_to_pptr(0));
+
+    if (vtd_irte_ptr_get_present(&irt_pptr[entry]) == 1) {
+        userError("IOMMU: Invalid handle %lu for MSI vector %lu, the handle is occupied by vector %llu.",
+                  entry, vector, vtd_irte_ptr_get_vector(&irt_pptr[entry]));
```
Indanz, line 355: The `%llu` gives a compile error for 32-bit x86. Not sure why 32-bit x86 is supported, but oh well.

include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
abrandnewusername: The Interrupt Remapping Table also supports posted interrupts, and we are planning to add support to posted interrupts.

src/plat/pc99/machine/intel-vtd.c
abrandnewusername: It means that Fault Processing Disable (FPD) is disabled, which indicates fault recording/reporting is enabled.

include/plat/pc99/plat/64/plat_mode/machine/hardware.bf
abrandnewusername: I am not really familiar with the kernel. Does `structures.bf` and `hardware.bf` work the same way? The [docs](https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/tools/bitfield_gen.md?plain=1#L788C39-L788C47) doesn't say much about it.

include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
Indanz: Posted interrupts are a way to inject IRQs into a VM running on another CPU core, and they're partially configured with this table. Whether interrupts are remapped or posted is configured independently, but practically you also want posted interrupts to be remapped.

But they used previously reserved bits to configure posting, so the entry format will be the same for either type. Then the "for Remapped Interrupts" will become "for remapped and posted interrupts", which also seems unnecessary verbose to me.

include/plat/pc99/plat/64/plat_mode/machine/hardware.bf
Indanz: I'm not very familiar with bitfield generation either, but I think it should just work?

include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
abrandnewusername: The IRTE structures for remapped interrupts and posting interrupts are slightly different according to the spec, the one for posted interrupts has extra fields for Posted Descriptor Address and other stuff.

include/plat/pc99/plat/machine/ioapic.h
wom-bat: I don't think seL4 runs on machines  with no IOAPIC.

src/arch/x86/object/interrupt.c
```diff
@@ -175,6 +173,13 @@ exception_t Arch_decodeIRQControlInvocation(word_t invLabel, word_t length, cte_
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+#ifdef CONFIG_IOMMU
+        status = vtd_inspect_irte_msi(handle, vector);
+        if (status != EXCEPTION_NONE) {
+            return status;
+        }
+        vtd_create_irte_msi(handle, vector, pci_bus, pci_dev, pci_func);
```
lsf37, line 181: This does not look like code that is legal to do in a decode function. At least `vtd_create_irte_msi` sounds like it changes state, and if `vtd_inspect_irte_msi` is just decoding and not changing state, it should also be called `decode..` or `check..`.

All state-changing code must go into `Arch_invokeIRQControl`.

lsf37:
So, there is a higher-level discussion we need to have on this one. @seL4/tsc -- strictly speaking this fix/feature needs an RFC, because it changes the API. It is also a breaking change, because it reserves IRQ handler 0, which user-level systems so far will assume to be available (see sel4test test failing for x86).

It is relatively obvious that we do want this feature and it is security relevant, so we may want it for this release (13.0). The details on how to handle it are less obvious to me. Any opinions on whether this should be on hold for the release and go through the RFC process instead?

src/arch/x86/object/interrupt.c
abrandnewusername: Thanks. Is this convention documented? Where can I find the documentation?

lsf37: https://docs.sel4.systems/processes/style-guide.html in the Kernel code/Error handling section has that requirement.

lsf37:
> So, there is a higher-level discussion we need to have on this one. @seL4/tsc -- strictly speaking this fix/feature needs an RFC, because it changes the API. It is also a breaking change, because it reserves IRQ handler 0, which user-level systems so far will assume to be available (see sel4test test failing for x86).
> 
> It is relatively obvious that we do want this feature and it is security relevant, so we may want it for this release (13.0). The details on how to handle it are less obvious to me. Any opinions on whether this should be on hold for the release and go through the RFC process instead?

Not exactly a quick discussion so far, but I think we can already say that we at least need a user-level guide description: what do the changes mean for the user, which new invocations are added, what do they do. You mentioned that you made a design decision with using IRQ 0. Can you also describe with other options you considered and what the implications would be?

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +294,103 @@ void vtd_handle_fault(void)
...
+    }
+    /* Remapping hardware units are configured to share the interrupt-remapping table,
+     * so we just read the table address from the first unit
+     */
+    vtd_irte_t *irt_pptr = (vtd_irte_t *) paddr_to_pptr(vtd_read64(0, IRTA_REG) & IRTA_REG_IRTA_MASK);
+    /* Sanity checking whether the Interrupt Remapping Table is initialized */
+    assert(irt_pptr != paddr_to_pptr(0));
+
+    word_t entry;
+    for (entry = 0; entry < MAX_IOAPIC && vtd_irte_ptr_get_present(&irt_pptr[entry]); entry++);
```
Indanz, line 310: Is this guaranteed to find an empty slot? If so, maybe add a comment why.

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -265,6 +294,103 @@ void vtd_handle_fault(void)
...
+    return true;
+}
+
+exception_t vtd_inspect_irte_msi(word_t entry, word_t vector)
+{
+    if (x86KSnumDrhu == 0) {
+        return EXCEPTION_NONE;
+    }
+
+    if (entry < MAX_IOAPIC || entry >= N_VTD_IRTES_4K) {
```
Indanz, line 337: Now that there is remappable IOAPIC support, is the check `entry < MAX_IOAPIC` still relevant? If so, what needs to be done to allow user space to get IRQ handlers for IOAPIC IRQs? sel4test fails because it seems to try to get IRQ handlers for IOAPIC IRQs.

Indanz (commented):
The remaining issue stopping this from being merged, as far as I can tell, are:

- That all state-changing code must go into `Arch_invokeIRQControl`.
- Fix IOAPIC IRQs if that's accidentally broken now.
- Fix seL4test if necessary if it does something it's not supposed to.

Indanz:
> strictly speaking this fix/feature needs an RFC, because it changes the API. It is also a breaking change, because it reserves IRQ handler 0, which user-level systems so far will assume to be available (see sel4test test failing for x86).

I don't see the API change, the only change I see is an extra error code. That's not API breakage. Maybe I'm missing something or misunderstanding things here, I only looked at the code changes.

I think the RFC procedure is to decide whether something is wanted and to decide on the API for new features. The first is already decided, the second doesn't apply as far as I can see.

>Not exactly a quick discussion so far,

I thought I commented on this already, but apparently not. :-/

> but I think we can already say that we at least need a user-level guide description: what do the changes mean for the user, which new invocations are added, what do they do. You mentioned that you made a design decision with using IRQ 0. Can you also describe with other options you considered and what the implications would be?

Where is that IRQ 0 thing mentioned? I don't see it. I also don't see new invocations.

src/plat/pc99/machine/ioapic.c
```diff
@@ -151,16 +159,31 @@ void ioapic_map_pin_to_vector(word_t ioapic, word_t pin, word_t level,
...
     index = ioapic * IOAPIC_IRQ_LINES + pin;
     ioapic_write(ioapic, IOAPIC_REGSEL, IOREDTBL_HIGH(pin));
-    ioredtbl_high = ioapic_read(ioapic, IOAPIC_WINDOW) & MASK(IOREDTBL_HIGH_RESERVED_BITS);
+    ioredtbl_high = ioapic_read(ioapic, IOAPIC_WINDOW) & MASK(IOREDTBL_HIGH_DESTINATION_BITS);
+#ifdef CONFIG_IOMMU
+    word_t entry;
+    if (!vtd_create_irte_ioapic(vector, level, ioapic_target_cpu, &entry)) {
+        printf("IOAPIC: failed to create IRTE for vector %lu.", vector);
+        return;
+    }
```
lsf37, line 168: The `&entry` takes the address of a local variable. This is [not allowed in the kernel](https://docs.sel4.systems/processes/style-guide.html#verification-requirements).

lsf37:
> > strictly speaking this fix/feature needs an RFC, because it changes the API. It is also a breaking change, because it reserves IRQ handler 0, which user-level systems so far will assume to be available (see sel4test test failing for x86).
> 
> I don't see the API change, the only change I see is an extra error code. That's not API breakage. Maybe I'm missing something or misunderstanding things here, I only looked at the code changes.

It's a bit of a grey area, that's why I was calling for discussion. A new error condition that is not just something that is missing from before, but adds new behaviour, is strictly speaking API breaking, and `vtd_create_irte_msi ` adds new behaviour, i.e. we're now supporting remapped interrupts, but there is no mention of them in the manual or the API, so it's not clear to me what it actually is that is now supported.

> I think the RFC procedure is to decide whether something is wanted and to decide on the API for new features. The first is already decided, the second doesn't apply as far as I can see.

That's why it's not a clear-cut case, it's a minor API change that on the high level is clear we want (hence could skip RFC), but the user-level explanation is missing.

> > Not exactly a quick discussion so far,
> 
> I thought I commented on this already, but apparently not. :-/

That's alright, there are also 12 more people who I had hoped might comment :-)

> > but I think we can already say that we at least need a user-level guide description: what do the changes mean for the user, which new invocations are added, what do they do. You mentioned that you made a design decision with using IRQ 0. Can you also describe with other options you considered and what the implications would be?
> 
> Where is that IRQ 0 thing mentioned? I don't see it. I also don't see new invocations.

All PC99 tests are failing (e.g. [here](https://github.com/seL4/seL4/actions/runs/6556674333/job/17985319239?pr=1098#step:4:483)) because IRQ handler 0 is now reserved, which is also a breaking change. I did not actually see that in the code anywhere either, though. @abrandnewusername can you point us to where that happens?

lsf37 (changes requested):
The verification code requirements and the high-level API question have to be addressed before we can merge.

abrandnewusername:
Update:
* the code no longer breaks the kernel C coding conventions
* more comments added

@lsf37 @Indanz Currently the kernel IOMMU driver supports two kinds of interrupts, MSI and IOAPIC. We don't have an MSI driver in the kernel, so the handle (which will be the entry of this MSI in the interrupt remapping table (IRT)) of an MSI is passed as a parameter of this MSI cap to the kernel by the user. We do have a kernel IOAPIC driver, so we don't rely on the user to provide the handle for IOAPICs.

It leads to two issues:
- The IOMMU driver creates IRT entries for MSIs and IOAPICs during the cap invocation, and it wouldn't know which one comes first (at least that's what I think happens). To avoid entry number clashes, I decided to reserve entry 0 to `#MAX_IOAPIC - 1` (the maximum number of IOAPICs this board can have)  for IOAPICs, which means that MSI can only use entry `#MAX_IOAPIC` to `#N_VTD_IRTES_4K - 1` (the number of entries the IRT has). This is the reason why seL4test dislikes me.
- The IRT is fix-sized. The number of IRTEs in a 4K interrupt remapping table is 256. The reason for doing that is because I use `it_alloc_paging()` to allocate memory for the IRT, which gives me exactly one page, and I couldn't find other tools in the kernel that allocate N contiguous pages for me. This means that if `#MAX_IOAPIC` is somehow too big, we leave no space for MSIs. I could make the size of the IRT configurable, but I guess it requires some more changes in the kernel.

Indanz:
If user space is managing the entries for MSI, but the kernel does it for IOAPIC, and it's the same range, it makes much more sense to reserve the last `MAX_IOAPIC` entries for IOAPIC instead of the beginning. Any reason that can't be done? 

Then the only user space visible change is a lower max handle for MSI IRQs, right? That should cause much less breakage. If there is a define that user space can use to check the max handle, that should be updated too.

Indanz:
A limit of 256 seems quite low. You could use a static array instead of dynamically allocating the page, just align it properly. Or use `alloc_rootserver_obj()` instead. Not sure if there is a clean solution for this, but it seems weird to use root server memory for IRT, but I guess that's the only possible way to allocated memory dynamically in kernel code.

lsf37:
> A limit of 256 seems quite low. You could use a static array instead of dynamically allocating the page, just align it properly. Or use `alloc_rootserver_obj()` instead. Not sure if there is a clean solution for this, but it seems weird to use root server memory for IRT, but I guess that's the only possible way to allocated memory dynamically in kernel code.

If a static array is possible, then that would be preferable. I don't think there is precedent for the kernel to root server resources to allocate kernel-level tables (unless it is a proper object with a retype API). We should not start to introduce some kind of hybrid -- that would definitely need an RFC discussion and TSC decision.

Indanz:
> If a static array is possible, then that would be preferable. I don't think there is precedent for the kernel to use root server resources to allocate kernel-level tables (unless it is a proper object with a retype API). We should not start to introduce some kind of hybrid -- that would definitely need an RFC discussion and TSC decision.

The current vtd code is using `it_alloc_paging()`, which allocates from the root server memory.

lsf37:
> The current vtd code is using `it_alloc_paging()`, which allocates from the root server memory.

You're right it does, I hadn't seen that. It'd be good if someone from the systems side with experience of the x86 port could comment on this. To me, this is against the design and we should ideally fix it, not do more of it.

For instance, if we had accepted one of the schemes that reclaim root server memory, this would immediately have lead to crashes. Using special ways to give memory to the kernel sounds extremely dangerous to me.

Indanz:
> It'd be good if someone from the systems side with experience of the x86 port could comment on this. To me, this is against the design and we should ideally fix it, not do more of it.

The amount of memory needed is hardware dependent and can be queried at startup, so I think the intention was to allocate as much as necessary instead of the theoretical maximum.

kent-mcleod:
> For instance, if we had accepted one of the schemes that reclaim root server memory, this would immediately have lead to crashes. Using special ways to give memory to the kernel sounds extremely dangerous to me.

Some allocation decisions need to be computed at boot time because the number of objects the kernel init code needs to create is not known at compile time. 

But all allocations still need to be declared up front with calculate_rootserver_size().  So it_alloc_paging() should be called exactly as many times as what `arch_get_n_paging(it_v_reg)` returns. So all allocations the kernel does can be kept in one block that is allocated once during boot. But most of the objects created inside the rootserver allocation block correspond to caps in the root thread's cnode.  Recovering root memory involves reusing the resources described by caps in the root's CNode, I can't see how other ways would work.

lsf37:
> > For instance, if we had accepted one of the schemes that reclaim root server memory, this would immediately have lead to crashes. Using special ways to give memory to the kernel sounds extremely dangerous to me.
> 
> Some allocation decisions need to be computed at boot time because the number of objects the kernel init code needs to create is not known at compile time.
> 
> But all allocations still need to be declared up front with calculate_rootserver_size(). So it_alloc_paging() should be called exactly as many times as what `arch_get_n_paging(it_v_reg)` returns. So all allocations the kernel does can be kept in one block that is allocated once during boot. But most of the objects created inside the rootserver allocation block correspond to caps in the root thread's cnode. Recovering root memory involves reusing the resources described by caps in the root's CNode, I can't see how other ways would work.

Ok, thanks for explaining it. So this is an old approach and while I would still call it flawed, we don't need to fix it now for this PR. The main problem I have with the approach is that user-level functions and kernel-level functions are mixed -- `it_*` functions are supposed to operate on the initial thread, not set up the kernel. 

For the table needed specifically here, the approach seems to be limiting, because it can use only one page. Would it make sense to have a compile-time configured fixed-size table that can be as large as needed, and then only use as much of it as the system dynamically supports at boot time? (with and enforced max at the size of the table). This probably won't work for everything, but it sounds like it might work here, and it gives the user some control over the memory trade-off.

Indanz:
> For the table needed specifically here, the approach seems to be limiting, because it can use only one page. Would it make sense to have a compile-time configured fixed-size table that can be as large as needed, and then only use as much of it as the system dynamically supports at boot time?

Agreed that for a fixed-size table like in this patch using `it_alloc_paging()` makes no sense, but solving that can go either way: Either do make it dynamic in size so it's always big enough for the hardware it runs on, following what the other IOMMU code does, or make it a static array. If doing the latter, I wouldn't bother with a compile-time option though.

Long-term, as x86's memory range is figured out dynamically anyway, memory like this should be reserved before it is given to the rest of the seL4 kernel to use, instead of stealing it from the root server.

PR #1102, opened 2023-09-07
zynqmp/u96: Update device tree to Petalinux 2021.1

chrisguikema:
This commit adds devices, but also updates the majority of the PS devices to use the "axi" bus, instead of the "amba" bus.


Ivan-Velickovic:
Should this kind of information be stored in the kernel? My understanding is that this has nothing to do with seL4 and the device trees are for passing up to the Linux guest?

chrisguikema:
There are two things: 

1. The kernel device tree can be used for generation of guest device trees
2. The 2021.1 device tree actually has more devices than the 2018.3 device tree

I know there is technically a workaround to #1 (https://github.com/seL4/camkes-vm/commit/d7d2ae3bfdfb44046ae9fdb3c1eb17f4d2d253c3). But the second is somewhat concerning if Xilinx is adding/removing devices from the device tree based on a Petalinux version, especially since the bootloader firmware (ATF, FSBL, u-boot) varies based off Petalinux version

Ivan-Velickovic:
My understanding of the goal of device trees in the seL4 kernel is so the kernel understands the layout of the hardware.

I don't see how any virtual machine could affect the layout of the hardware, and hence don't see why you would ever have to change the device trees that seL4 uses.

The friction comes from the CAmkES VM project using the seL4 device trees to describe the (virtualised) hardware to the virtual machine, which seems like the wrong approach to me and will continue to cause friction.

Indanz:
> The friction comes from the CAmkES VM project using the seL4 device trees to describe the (virtualised) hardware to the virtual machine, which seems like the wrong approach to me and will continue to cause friction.

Agreed.

For MEP I used the modified Linux kernel device tree from the guest exactly for this reason: seL4 is mostly interested in memory regions and those usually don't change much, considering the hardware is fixed. (In the case of the zynq hardware is also configurable, but the FPGA part has a fixed address range on the bus.)

Indanz:
To be clear, if you want to update the device tree because there are device regions missing that's fine. Adding device trees and config options for different boards or SoCs is also fine. But doing that for different Linux versions is a step too far I think.

PR #1105, opened 2023-09-23
Generate configuration headers at configure-time

nspin:
https://github.com/seL4/seL4/pull/975, which added JSON configuration output, caused the configuration headers to be generated at build-time instead of configure-time.  This broke CAmkES build systems which depend on configuration headers in CAmkES files at configure-time.

This PR makes the configuration headers available at configure-time.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -46,9 +46,6 @@ Upcoming release: BREAKING
 * aarch32 VM fault messages now deliver original (untranslated) faulting IP in a hypervisor context, matching
   aarch64 behaviour.
 * Correct the minimum size of a scheduling context. This changes the value of `seL4_MinSchedContextBits`.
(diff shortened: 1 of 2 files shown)


kent-mcleod (approved):

kent-mcleod:
@nspin thanks, are you able to address the style failure?

lsf37:
It'd be good to include this in the release, because it avoids a breaking change.

lsf37:
> Can you add a comment in the CMake file also, that CAmkES requires the files to be created during configuration? I wonder if CAmkEs can't be changes to only use the files at build time.

CAmkES can be changed (I think it has already been changed for that), but the whole point of this PR is to avoid a breaking change for the user and keep the old behaviour which is more convenient.

PR #1106, opened 2023-09-23
cmake: Install object API files under better names

nspin:
These files cannot be installed under the same names as those in the source tree because the two named `sel4arch.xml` conflict. These cannot be renamed in the source tree because of other projects which expect them under their current names.

https://github.com/seL4/seL4/pull/974 enabled them to be installed them under non-conflicting names, but those names (`sel4.xml`, `sel4-arch.xml`, `sel4-sel4arch.xml`) are confusing and too close to the original names.

This commit changes the names they are installed under to be more clear and descriptive (`object-api.xml`, `object-api-arch.xml`, `object-api-sel4-arch.xml`). I opted not to include a `sel4-*` prefix in these names to match `libsel4/include/api/syscall.xml`.


kent-mcleod:
> These files cannot be installed under the same names as those in the source tree because the two named `sel4arch.xml` conflict. These cannot be renamed in the source tree because of other projects which expect them under their current names.

I think it is likely that renaming the source files will cause [downstream projects](https://gitlab.com/robigalia/sel4-sys/-/blob/master/build.rs?ref_type=heads#L32) to break, but I think renaming the files to not have the same name would be a justifiable breaking change.

lsf37:
We should resolve this before the release, so that other projects can interface properly.

If we do the PR as is we're only changing output that has been added since the last release (in #974).

If we're renaming the source files, it'd be a proper breaking change, but would make things clearer overall and easier to interface with.

I don't have a strong opinion on this. Maybe a slight preference for not breaking downstream projects more than we have to.

Is there anyone with a strong opinion on which way this should go?

ratmice:
I thought my [downstream](https://github.com/ratmice/rust_generated/blob/main/CMakeLists.txt) would be impacted but is actually referring to the source tree rather than in those installed in the build location.  I should probably change that. I do not have strong preference, but find Nick's clearer (and was going to say the same thing before I realized I'm probably not impacted by the change).

kent-mcleod (approved):

kent-mcleod:
> Is there anyone with a strong opinion on which way this should go?

I'm happy with leaving the source names the same and renaming the installation names.  In my use case, I'm sometimes trying to build rust apps in the same build unit that builds the kernel, but I can manually copy and rename the files in this case.

lsf37 (approved):
Ok, as I said above I'm happy with both options, and this one has less chance of breaking downstream projects.

nspin:
Until the root of the problem is addressed, the `sel4-sys` crate supports the build system specifying the location of each object interface XML file individually with environment variables, overriding the default include path search that will not work due to the clobbering. I will add this to the project and crate documentation.

https://github.com/seL4/rust-sel4/blob/9acd9177c256266f0e2eb2d60ef80b2e7c81f01a/crates/sel4/sys/build/main.rs#L54-L62

nspin:
Ideally, the filenames we choose in this PR are the same filenames that we would like to use for a potential future change that actually renames them in the source tree. That way, if that future change is ever made, it won't change the behavior of `cmake install`, and thus won't break projects using it.

The names currently implemented in this PR are:
- `include/interfaces/object-api.xml`
- `include/interfaces/object-api-arch.xml`
- `include/interfaces/object-api-sel4-arch.xml`

kent-mcleod:
Could we even name them more specifically? EG for aarch32, the sel4-arch specific top level tag is `ObjectApiAarch32`. So could it be named object-api-aarch32.xml?

nspin:
> Could we even name them more specifically? EG for aarch32, the sel4-arch specific top level tag is ObjectApiAarch32. So could it be named object-api-aarch32.xml?

I see it like this: the advantages of keeping the filenames configuration-independent are the same as the advantages of keeping the filenames of headers which are used in configuration-independent code architecture-independent (e.g. `<sel4/arch/types.h>` and `<sel4/sel4_arch/types.h>`). Configuration-independent code which references these files doesn't need to compute their names according to the configuration to do so.

lsf37:
@kent-mcleod are you happy for this to be merged as it is now?

kent-mcleod:
> @kent-mcleod are you happy for this to be merged as it is now?

Yea

lsf37:
> This change may break current users depending on the currently installed API files, right? In that case it should probably be documented in `CHANGES`.

The overall support for the `install` target was added after the last release, so there is no change as such to the last release. But we should probably mention in the CHANGES file that the whole `cmake install` target now exists. Will add a separate issue for that.

Issue #1023, opened 2023-04-22
Improving IRQ Handling


JorgeMVP:
Sharing some thoughts about what would be nice to see implemented in seL4, to increase the efficiency of virtualization use-cases for IRQs specifically. @axel-h @kent-mcleod @lsf37

**Kernel IRQ-level:**
- Separation of Priority-drop from Deactivation (Feature supported by GICv2/3/4).
- Dynamically enable/disable Maintenance IRQ for the underflow condition based on the number of pending LRs.

**Kernel vCPU-level:**
- Adding a bitmap for each vCPU object with 2-bits per pending vIRQ, max 1024 vIRQ as spec mentions.
- Add pending vIRQ (on demand - syscall), and schedule pending vIRQ helper functions (on vcpu entry/kernel exit).
- Direct injection of the vtimer IRQ since when triggered it belongs to the current vCPU.
We can add a new IRQ State - IRQVCPUstate.
- New vCPU syscall to bind an IRQHandler cap to a vCPU cap. (Instead of Notification Object, we can use a vCPU Object)

(i) I'm not aware of the sel4 verification story, I'd like to know if this brings any problems to it?

(ii) Also, I'd like to know your thoughs about improving the IRQ Handling with some of the above ideas (if not all).

kent-mcleod:
Wow, you must be psychic because I'd been thinking about making this issue earlier today also.

> Separation of Priority-drop from Deactivation (Feature supported by GICv2/3/4).

Yes, I've been trying to find time to set up some benchmarks to measure the impact of this to show the impact of the changes. (And then figure out all the edge cases to migrate the current design).

> Dynamically enable/disable Maintenance IRQ for the underflow condition based on the number of pending LRs.

Can you elaborate what you mean here?

> Adding a bitmap for each vCPU object with 2-bits per pending vIRQ, max 1024 vIRQ as spec mentions.

Is this for the state that is currently tracked at user level so that the kernel can directly refill LRs without involving userlevel? 

> New vCPU syscall to bind an IRQHandler cap to a vCPU cap. (Instead of Notification Object, we can use a vCPU Object)

[...]

Issue #952, opened 2022-11-15
Arm, GIC: Consider using EOImode = 1 insteade of 0


kent-mcleod:
When running on platforms using the GICv2 or GICv3, the kernel likely takes longer to handle interrupts that are delegated to user level using `EOImode = 0` than if `EOImode = 1`. 

When `EOImode = 1`, the interrupt acknowledgement is split into priority-drop and deactivation. Priority-drop allows other interrupts to be delivered, while deactivation advances the state-machine for the specific interrupt number to be advanced allowing new deliveries of the same interrupt. When using `EOImode = 0`, both of these operations are performed together.

Currently the kernel essentially emulates the behavior of `EOImode = 1` using `EOImode = 0` which requires additionally disabling the interrupt until the user level handler capability is acknowledged at which point the IRQ is reenabled. The disable and re-enable operation requires modifying registers at the distributor instead of just at the CPU local interface and so likely has a much higher latency to perform for each interrupt delivery.

Using `EOImode = 1` will also make it easier to use some of the GIC virtualization optimizations that allow the hardware to automatically deactivate a physical interrupt that's being "passed-through" to a guest without needing an additional fault to be generated.

JorgeMVP:
About the idea of using different EOIMode settings for different configurations (different config builds with static EOIMode) is possible but I think we don’t gain anything there. 
For changing dynamically EOIMode, I doubt that is possible, but even if it does not bring any benefit.

I guess, EOIMode always set to 1 is the way to go.

kent-mcleod:
> For changing dynamically EOIMode, I doubt that is possible, but even if it is, it does not bring any benefit.

Yea, for some reason when I wrote that for some reason I was thinking that the EOI mode was configured per IRQ rather than per cpu interface. So it doesn't make sence.

Indanz:
Implemented by #1183.

kent-mcleod:
This is still open for gicv2

Indanz:
Well, I thought we didn't plan on supporting this for GICv2: https://github.com/seL4/seL4/pull/1183#issuecomment-1922355618. From your post, it doesn't seem worth the hassle.

midnightveil:
@kent-mcleod A question, if you please:

<details><summary>a question I realised I was wrong about</summary>

I realised I mixed up the ordering of EOI and DIR in my head.

seL4 itself does IAR, then does EOI within the same kernel operation. DIR then happens later when userspace does an ACK. So this is actually fine.

---

[...]

PR #1107, opened 2023-09-25
GICv2 Distributor Config Cache

danshea00:
Every time an IRQ occurs the kernel reads from a GICv2 distributor configuration register to determine whether the interrupt is edge-triggered or level-triggered. If it is edge-triggered, the pending status of the interrupt must be cleared. The read from the distributor was observed to be expensive and so this pull request implements a cache for the interrupt configuration registers.

The following seL4Bench benchmarks demonstrate the performance improvement of this change using unicore MCS as measured from userspace (100 runs of IRQUser "With context switch" benchmark):

| Platform | cycles before: mean (stddev) | cycles after: mean (stddev) | diff: absolute (%) |
|---|---|---|---|
| OdroidC2 | 1132 (7) | 904 (7) | 228 (20%) |
| OdroidC4 | 1111 (7) | 889 (6) | 222 (20%) |
| TX1 | 1019 (7) | 813 (3) | 206 (20%) |

### DIFF
--- a/include/arch/arm/arch/machine/gic_v2.h
+++ b/include/arch/arm/arch/machine/gic_v2.h
@@ -22,6 +22,7 @@
 
 #define IRQ_MASK MASK(10u)
 #define GIC_VCPU_MAX_NUM_LR 64
+#define GIC_DIST_CONFIG_SIZE 64
 
 /* Helpers for VGIC */
 #define VGIC_HCR_EOI_INVALID_COUNT(hcr) (((hcr) >> 27) & 0x1f)
@@ -124,13 +125,14 @@ struct gic_cpu_iface_map {
 
 extern volatile struct gic_dist_map *const gic_dist;
 extern volatile struct gic_cpu_iface_map *const gic_cpuiface;
+extern uint32_t gic_dist_config_cache[GIC_DIST_CONFIG_SIZE];
 
 /* Helpers */
 static inline int is_irq_edge_triggered(word_t irq)
 {
     int word = irq >> 4;
     int bit = ((irq & 0xf) * 2);
-    return !!(gic_dist->config[word] & BIT(bit + 1));
+    return !!(gic_dist_config_cache[word] & BIT(bit + 1));
 }
 
 static inline void dist_pending_clr(word_t irq)
--- a/src/arch/arm/machine/gic_v2.c
+++ b/src/arch/arm/machine/gic_v2.c
@@ -25,6 +25,7 @@
 #else  /* GIC_DISTRIBUTOR_PPTR */
 volatile struct gic_dist_map *const gic_dist =
     (volatile struct gic_dist_map *)(GIC_V2_DISTRIBUTOR_PPTR);
+uint32_t gic_dist_config_cache[GIC_DIST_CONFIG_SIZE];
 #endif /* GIC_DISTRIBUTOR_PPTR */
 
 #ifndef GIC_V2_CONTROLLER_PPTR
@@ -95,6 +96,10 @@ BOOT_CODE static void dist_init(void)
         gic_dist->config[i >> 5] = 0x55555555;
     }
 
+    for (i = 0; i < GIC_DIST_CONFIG_SIZE; i++) {
+        gic_dist_config_cache[i] = gic_dist->config[i];
+    }
+
     /* group 0 for secure; group 1 for non-secure */
     for (i = 0; i < nirqs; i += 32) {
         if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT) && !config_set(CONFIG_PLAT_QEMU_ARM_VIRT)) {
@@ -157,8 +162,10 @@ void setIRQTrigger(irq_t irq, bool_t trigger)
     if (trigger) {
         /* set the bit */
         gic_dist->config[index] |= BIT(offset + 1);
+        gic_dist_config_cache[index] |= BIT(offset + 1);
     } else {
         gic_dist->config[index] &= ~BIT(offset + 1);
+        gic_dist_config_cache[index] &= ~BIT(offset + 1);
     }
 }
 


Ivan-Velickovic:
It should be noted that GICv3 seems to not suffer from this issue.

include/arch/arm/arch/machine/gic_v2.h
```diff
@@ -124,13 +125,14 @@ struct gic_cpu_iface_map {
 
 extern volatile struct gic_dist_map *const gic_dist;
 extern volatile struct gic_cpu_iface_map *const gic_cpuiface;
+extern uint32_t gic_dist_config_cache[GIC_DIST_CONFIG_SIZE];
```
Ivan-Velickovic, line 128: There should be a comment here (as well as in the commit message) as to why the change is necessary (in my opinon).

Indanz (commented):
Code looks good. However, I fear you are unnecessarily optimising bad code.

I don't think the `dist_pending_clr()` call is needed for edge-triggered interrupts. Signalled interrupts are masked before acking them. See also @kent-mcleod reply at https://github.com/kent-mcleod/seL4/pull/14.

At best another edge-triggered interrupt which came before the ack will get cleared with the current code and save one unnecessary interrupt by coalescing the two edge-triggered irqs (assuming that's wanted behaviour). However, if the interrupt comes after the ack, it will still be delivered as a new interrupt. The time window in the kernel is very small, while the time window between the signal and user space calling `seL4_IRQHandler_Ack()` is much bigger, so the current code isn't very effective.

Clearing pending IRQs for edge-triggered ones would make much more sense when enabling an edge-triggered IRQ in `maskInterrupt()`, if that's what's wanted (and then your caching code makes sense too). But that won't be necessary with EOIMode = 1, as that removes the need for masking.

With the current code I think removing `is_irq_edge_triggered()` and the special code for level-triggered IRQs is the better solution. Longer-term, EOIMode = 1 should be implemented. See also #952, #1023.

danshea00:
Thanks for the comments @Indanz! I have a few questions.

> At best another edge-triggered interrupt which came before the ack will get cleared with the current code and save one unnecessary interrupt by coalescing the two edge-triggered irqs (assuming that's wanted behaviour). However, if the interrupt comes after the ack, it will still be delivered as a new interrupt. The time window in the kernel is very small, while the time window between the signal and user space calling seL4_IRQHandler_Ack() is much bigger, so the current code isn't very effective.

I don’t see why the time window between `dist_pending_clr()` being called and the interrupt being reenabled by `seL4_IRQHandler_Ack()` is an issue. The interrupt is disabled prior to the `dist_pending_clr()` by `maskInterrupt()`. Can an interrupt become pending even if it is disabled?

> Clearing pending IRQs for edge-triggered ones would make much more sense when enabling an edge-triggered IRQ in maskInterrupt(), if that's what's wanted (and then your caching code makes sense too). 

Are you suggesting to move `dist_pending_clr()` within `seL4_IRQHandler_Ack()` (i.e. when `maskInterrupt()` reenables the interrupt)? This doesn’t make much sense to me, given that a lower priority pending interrupt would not be handled until the userspace calls `seL4_IRQHandler_Ack()` for the higher priority pending interrupt.

https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/arch/arm/machine/gic_v2.c#L93-L96

Is this the “special code for level-triggered IRQs'' you are referring to?

It seems to me like for the current code, simply removing `is_irq_edge_triggered()` would make sense. The GICv2 docs indicate that this shouldn’t affect the pending status for level-triggered interrupts as the kernel never writes to `GICD_ISPENDR`.

Indanz:
> Can an interrupt become pending even if it is disabled?

Yes. Disabling an IRQ has no effect on its state, it only stops interrupting the CPU.

> Are you suggesting to move dist_pending_clr() within seL4_IRQHandler_Ack() (i.e. when maskInterrupt() reenables the interrupt)?

_If_ special handling of edge-triggered interrupts is wanted, then yes, I propose it should be done when doing the ACK.

> This doesn’t make much sense to me, given that a lower priority pending interrupt would not be handled until the userspace calls seL4_IRQHandler_Ack() for the higher priority pending interrupt.

That's not how it works, priorities have nothing to do with it: The `dist_pending_clr()` clears the pending state of the currently active interrupt. It only has an effect if the IRQ state is active-and-pending, which happens for edge-triggered interrupts if another edge happens on the IRQ line for the currently active interrupt. See ["Figure 6 Interrupt life cycle - edge-triggered interrupts"](https://documentation-service.arm.com/static/5f1068720daa596235e7f6ef) in the GICv3 docs, which explains it better, but it also applies to GICv2.

> Is this the “special code for level-triggered IRQs'' you are referring to?

No, I meant the if-check code that uses `is_irq_edge_triggered()`.

> The GICv2 docs indicate that this shouldn’t affect the pending status for level-triggered interrupts as the kernel never writes to GICD_ISPENDR.

The main way interrupts become pending is from HW, not SW, so that's not relevant.

Level-triggered interrupts are always active and pending because the hardware still asserts the IRQ line when the interrupt becomes active. But because seL4 disables the IRQ till user-space ACKs the interrupt, it doesn't matter. For level-triggered IRQs user space will do something with the device to make the IRQ become idle again, so when it's done and it calls `seL4_IRQHandler_Ack()`, the pending will have disappeared (calling the ACK before handling the device would be a bug).

Edge-triggered interrupts behave a lot more like seL4 notifications. For those you need to call `seL4_IRQHandler_Ack()` first before handling the device, otherwise you may miss events and that would be a bug. I would even go as far as saying that you wouldn't need to call `seL4_IRQHandler_Ack()` for edge-triggered interrupts.

Common state transition for edge-triggered interrupts:

1. Edge -> IRQ pending
2. seL4 gets the `active_irq` by reading `int_ack` -> IRQ active
3. ...
4. seL4 calls `dist_pending_clr()` -> No change in IRQ state
5. seL4 Acknowledges the interrupt by writing EIO and signals user space -> IRQ idle
6. ...
7. User space calls `seL4_IRQHandler_Ack()` -> No change in IRQ state

Rare state transition for edge-triggered interrupts when another edge happens during kernel IRQ handling:

1. Edge -> IRQ pending
2. seL4 gets the `active_irq` by reading `int_ack` -> IRQ active
3. Edge -> IRQ active & pending
4. seL4 calls `dist_pending_clr()` -> IRQ active
5. seL4 Acknowledges the interrupt by writing EIO and signals user space -> IRQ idle
6. ...
7. User space calls `seL4_IRQHandler_Ack()` -> No change in IRQ state

So the only change clearing pending has is that edges happening while seL4 is in the interrupt handler are ignored. Any edged at 6 will still cause a new interrupt, even though user space was just notified an IRQ happed:

Less rare state transition for edge-triggered interrupts when another edge happens during user space IRQ handling:

1. Edge -> IRQ pending
2. seL4 gets the `active_irq` by reading `int_ack` -> IRQ active
3. (Nothing happens)
4. seL4 calls `dist_pending_clr()` -> No change in IRQ state
5. seL4 Acknowledges the interrupt by writing EIO and signals user space -> IRQ idle
6. Edge -> IRQ pending
7. User space calls `seL4_IRQHandler_Ack()` -> No change in IRQ state, but IRQ unmasked
8. seL4 gets the interrupt by reading `int_ack` -> IRQ active

Covering the race at 3, but not the bigger window of 6 doesn't make any sense. Either cover windows 6 and do something special during `seL4_IRQHandler_Ack()`, or remove all edge-triggered special casing. I propose the latter, because edge-triggered interrupts are not expected to happen often. Making all interrupt handling slightly faster and the code simpler is the better trade-off.

I'm going to open a new pull request to remove the edge-triggered special code, with the same explanation.

danshea00:
Thanks so much for the detailed explanation, that clears up my understanding. Feel free to close this PR when you open yours.

PR #1109, opened 2023-10-05
VM fault fastpath: missing rename of vspace cap ASID function

danshea00:
This change was missing from commit dc808b3d76fc09741e91ccc53e2738d052c08f0a. Renames `cap_vtable_root_get_mappedASID` to `cap_vspace_cap_get_capMappedASID`, allowing builds with the exception fastpath enabled to build.

### DIFF
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -779,7 +779,7 @@ void NORETURN fastpath_vm_fault(vm_fault_type_t type)
 
 #ifdef CONFIG_ARCH_AARCH64
     /* Need to test that the ASID is still valid */
-    asid_t asid = cap_vtable_root_get_mappedASID(newVTable);
+    asid_t asid = cap_vspace_cap_get_capMappedASID(newVTable);
     asid_map_t asid_map = findMapForASID(asid);
     if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
                  VSPACE_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {


lsf37:
I really should be adding those fast paths to CI somewhere. The question is where (without completely exploding the number of tests)

lsf37 (approved):

alwin-joshy (approved):

alwin-joshy:
> I really should be adding those fast paths to CI somewhere. The question is where (without completely exploding the number of tests)

Would an extra compile and seL4test sim test be reasonable (enabling both of the new fastpaths)? Though I'm not sure if sel4test sim infrastructure currently lets you set and unset specific config options.

Ivan-Velickovic:
> Would an extra compile and seL4test sim test be reasonable (enabling both of the new fastpaths)? Though I'm not sure if sel4test sim infrastructure currently lets you set and unset specific config options.

The current infrastructure doesn't let you (at least easily) do it. We've run into cases before where it's needed (e.g https://github.com/seL4/ci-actions/pull/277/files) so it might be worthwhile changing the infrastructure.

lsf37:
> > Would an extra compile and seL4test sim test be reasonable (enabling both of the new fastpaths)? Though I'm not sure if sel4test sim infrastructure currently lets you set and unset specific config options.
> 
> The current infrastructure doesn't let you (at least easily) do it. We've run into cases before where it's needed (e.g https://github.com/seL4/ci-actions/pull/277/files) so it might be worthwhile changing the infrastructure.

Might need a slight tweak, but generally the infrastructure as such should be sufficient for adding specific additional runs (exceptions), as long as you do that in the test yaml spec for `sel4test` in the `ci-actions` repo.  You would not want to do this in the workflow files (i.e. adding a new actual github check), because you'd then have to replicate those exceptional runs over many repos (anything that wants to do `sel4test` runs). I hope to have a bit more time next week to look at this.

Indanz:
The first sel4test run failed with the following two errors:
```
<testcase classname="sel4test" name="MULTICORE0004">
  Running test MULTICORE0004 (Test core stalling is behaving properly (flaky))
  seL4 failed assertion 'isRunnable(NODE_STATE(ksCurThread))' at /github/workspace/kernel/include/kernel/thread.h:291 in function checkBudgetRestart
  halting...
  Kernel entry via Syscall, number: 11, Yield

ODROID_C4_debug_SMP_MCS_clang_64 FAILED  
```
```
   	<testcase classname="sel4test" name="FPU0001">
  
  [[Timeout]]

ODROID_XU4_release_clang_32 FAILED
```

MULTICORE0004 is known to fail sometimes, FPU0001 failing is new to me, only mention I know of that is [this](https://sel4.discourse.group/t/sel4test-get-stuck-in-fpu0001-test-on-qemu-arm-virt-with-givc3-enable/772) post.

Neither have anything to do with this pull request, so merging the fix upstream.

lsf37:
Hm, MULTICORE0004 is known to fail, but not like this. The failure mode is that it's hanging, not that it's triggering an assertion. This does seem to be a new form. The FPU test hanging is known.

I do agree that neither is connected to this PR, though.

Issue #1110, opened 2023-10-16
AArch64: 4k user page PTEs appear to be created incorrectly


Xaphiosis:
While working on proof setup for C refinement for AArch64, I had to write a spec for `makeUserPage` (which is also interestingly not named makeUserPTE like on other platforms). The function contains this interesting block (`src/arch/arm/64/kernel/vspace.c`):

```
    /* If The target page size is 4k, then really pte_pte_4k_page_new() should be used
       but instead we just update the type manually after using pte_pte_page_new(). */
    if (page_size == ARMSmallPage) {
        ret.words[0] |= 0x3;
    }
```
("just", haha)

From a type safety perspective, this kind of re-enactment of `reinterpret_cast` on the bitfield generator is not good, and from the verification side we strongly suggest avoiding it. I tried to prove that taking a `pte_page` and ORing 3 will result in a `pte_4k_page` with identical parameters and ended up with this unprovable situation:
```
pte_get_tag pte = SCAST(32 signed → 64) pte_pte_page 
⟹ pte_get_tag (pte_C.words_C_update (λv. Arrays.update v 0 (v.[0] || 3)) pte) =
      SCAST(32 signed → 64) pte_pte_4k_page
```
This indicates that there is no way one can take a word that is tagged as a pte_page, OR it with 3 and have it be interpreted as a pte_4k_page.

The tags are explained in `structures.bf` with:
```
-- There are two page type fields because the 4k page size
-- uses a different hardware encoding. We use bit 58
-- which is reserved for software use to encode this
-- difference in the tag for these types.
tagged_union pte pte_type(pte_hw_type, pte_sw_type) {
    tag pte_table               (3, 0)
    tag pte_page                (1, 0)
    tag pte_4k_page             (3, 1)
    tag pte_invalid             (0, 0)
}
```

So for a pte_page we need bit 0 set, and for a 4k page we need bits 58,1,0 set. Note that the integer 3 does not contain bit 58.

Unless there is some other workaround that magically fixes this issue at some other location, it means that no user pages on AArch64 are being placed in the page table as 4k pages. Not sure what this means when running on actual hardware.

I have assigned myself to this issue as I want to rewrite and rename this function to avoid its current issues of being annoying to find, understand and reason about, additionally avoiding bypassing the bitfield generator with that tag cast that also happens to be wrong. Will aim for a PR tomorrow.

kent-mcleod:
> which is also interestingly not named makeUserPTE like on other platforms

[...]

PR #1111, opened 2023-10-17
aarch64: makeUserPage rework and some verification MODIFIES annotations

Xaphiosis:
This overhauls `makeUserPage` to take care of the bit 58 and bitfield gen issues outlined in #1110 
I threw in a verification MODIFIES annotation commit to do less PR littering, it doesn't have any impact on the code.

### DIFF
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -196,6 +196,7 @@ static inline void writeAMAIR(word_t reg)
     MSR(REG_AMAIR_EL1, reg);
 }
 
+/** MODIFIES: */
 /** DONT_TRANSLATE */
 static inline word_t readCIDR(void)
 {
@@ -204,6 +205,7 @@ static inline word_t readCIDR(void)
     return (word_t)reg;
 }
 
+/** MODIFIES: phantom_machine_state */
 /** DONT_TRANSLATE */
 static inline void writeCIDR(word_t reg)
 {
@@ -222,6 +224,7 @@ static inline void writeACTLR(word_t reg)
     MSR(REG_ACTLR_EL1, reg);
 }
 
+/** MODIFIES: */
 /** DONT_TRANSLATE */
 static inline word_t readAFSR0(void)
 {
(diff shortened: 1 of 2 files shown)


lsf37 (approved):
This looks good from my side. The new formulation of `makeUserPage` should be a lot easier on verification and addresses #1110.

I'd still be up for renaming `makeUserPage` into `makeUserPagePTE`.

Indanz (approved):
Looks good to me, nice clean-up. I would also use a variable for `APFromVMRights(vm_rights)` and I agree with the name change proposed.

I checked in the ARM documentation that reading or writing those system registers has no side effects.

Xaphiosis:
Thanks @lsf37 and @Indanz! I did the rename due to popular demand, and tweaked the commit line to fit in the character limit. Not sure pulling out `APFromVMRights(vm_rights)`, it's the obvious accessor function already. Might save a line of space, but I'm inclined to leave it.

Xaphiosis:
Corey spotted a typo in a commit messages. Also, rebased.

Indanz:
ODROID_XU4_release_clang_32 FAILED with:
```
  	<testcase classname="sel4test" name="PREEMPT_REVOKE">
  
  [[Timeout]]
  None
```
which I haven't seen before.

TX1_debug_SMP_hyp_MCS_gcc_64 FAILED with:

```
  Warning:  gpt_cntfrq 19200000, expected 12000000
  Bootstrapping kernel
  available phys memory regions: 2
    [80000000..ff000000]
    [100000000..180000000]
    [8080000000..8080260000]
    [8080260000..808026d4d3]
    [808026e000..8080684000]
  WWWaaarrrnnniiinnnggg:::      gggpppttt___cccnnntttfffrrrqqq   111999222000000000000000,,,   eeexxxpppeeecccttteeeddd   111222000000000000000000
  
  clock_sync_test[2]: t0 = 267991177, t = 267991190, td = 13
  seL4 failed assertion 't0 <= margin + t && t <= t0 + margin' at /github/workspace/kernel/src/kernel/boot.c:587 in function clock_sync_test
  halting...
  
  [[Timeout]]
```
It seems that `TIMER_FREQUENCY` is wrong for the TX1 board. If set correctly the td would fall within the margin of the `clock_sync_test`.

Both failures are unrelated to your change.

Xaphiosis:
Thanks for checking on those, they did not look related. The `PREEMPT_REVOKE` looks a little concerning (XU4 32-bit non-SMP which is not far from a verified platform), but it only happened on clang, so I dunno.

lsf37:
> It would be nice if we knew where the code is hanging, in the kernel, in the test infrastructure code, or in the test itself. The test has no loops that can hang forever, so it must be something else. The hang may be unrelated to the test being run and could be a random board issue too.

Peter did mention that the ODROID_XU4 is reaching its end of life, it's a pretty old board and is no longer being sold (you can get some 2nd hand ones, but that's not really useful). So that one really might be board related. 

Also agree on the timer frequency for TX1, that would be worth looking at.

PR #1114, opened 2023-10-20
ARM: No special handling for edge-triggered IRQs

Indanz:
Clearing the pending state only has an effect if the IRQ state is active-and-pending, which happens for edge-triggered interrupts if another edge happens on the IRQ line for the currently active interrupt. This window is small enough to ignore, at worst user space will get another notification, which is harmless.

If unnecessary notifications are unwanted, the pending state should be cleared during seL4_IRQHandler_Ack(), as that covers a much bigger window. However, edge-triggered interrupts are not expected to happen often. Making all interrupt handling slightly faster and the code simpler is the better trade-off.

Reading the GIC config word is very slow for GICv2, see pull #1107.

For a more detailed explanation, see [this](https://github.com/seL4/seL4/pull/1107#issuecomment-1745278447) comment.

### DIFF
--- a/include/arch/arm/arch/machine/gic_v2.h
+++ b/include/arch/arm/arch/machine/gic_v2.h
@@ -126,21 +126,6 @@ extern volatile struct gic_dist_map *const gic_dist;
 extern volatile struct gic_cpu_iface_map *const gic_cpuiface;
 
 /* Helpers */
-static inline int is_irq_edge_triggered(word_t irq)
-{
-    int word = irq >> 4;
-    int bit = ((irq & 0xf) * 2);
-    return !!(gic_dist->config[word] & BIT(bit + 1));
-}
-
-static inline void dist_pending_clr(word_t irq)
-{
-    int word = IRQ_REG(irq);
-    int bit = IRQ_BIT(irq);
-    /* Using |= here is detrimental to your health */
-    gic_dist->pending_clr[word] = BIT(bit);
-}
-
 static inline void dist_enable_clr(word_t irq)
 {
     int word = IRQ_REG(irq);
@@ -198,9 +183,6 @@ static inline void ackInterrupt(irq_t irq)
 {
     assert(IS_IRQ_VALID(active_irq[CURRENT_CPU_INDEX()])
            && (active_irq[CURRENT_CPU_INDEX()] & IRQ_MASK) == IRQT_TO_IRQ(irq));
-    if (is_irq_edge_triggered(IRQT_TO_IRQ(irq))) {
-        dist_pending_clr(IRQT_TO_IRQ(irq));
-    }
     gic_cpuiface->eoi = active_irq[CURRENT_CPU_INDEX()];
     active_irq[CURRENT_CPU_INDEX()] = IRQ_NONE;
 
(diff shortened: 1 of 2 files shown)


Indanz:
The preprocessor failures are expected. ODROID_XU4 failing because it's rebooting every few minutes is not, but that has happened before. It rebooted 30 times. Getting hot in the server room?

kent-mcleod:
I couldn't find why the special handling was added originally.  I had a look back through the history to when this code was added, but it was added before the code was open sourced. In the original release commit, there was support for cortex-a9 (imx6 sabrelite) and a15, (I think exynos/Odroid) platforms. Looking at the code for this commit, the interrupt controller initialization code sets all IRQs to level triggered, and there's no code for changing any to edge triggered....
Although looking in the device tree for the imx6 platform, the MPCore timer specifies an edge triggered interrupt, and the GIC2.0 architecture claims it's implementation defined whether PPI IRQs can have their CFG programmed. So I'd conclude that at the time, this code was not ever impacting SPIs and only potentially impacting PPIs where only the core timer was being used.  Maybe long running kernel operations were causing timer interrupts to become active+pending and this was added to try and work around?

In any case I agree with your analysis @Indanz, and agree this can get fixed finally.

kent-mcleod (approved):

Indanz:
> Maybe long running kernel operations were causing timer interrupts to become active+pending and this was added to try and work around?

But for timer interrupts you do want to get the pending one immediately, so you can handle the timeout event as quickly as possible. With edge-triggered interrupts you want to ack them before handling them, otherwise you may miss events.

lsf37:
The commit message (by Alex) that introduced it is not very helpful:

```
Date:   Wed Oct 2 13:59:58 2013 +1000

   GIC, ARMv7 Cortex-A9/A15: Created some inline helper functions for probing irq status at the distributer
```

It adds a bunch of helpers, including

```
static inline int
__is_edge_triggered(irq_t irq)
{
    int word = irq/16;
    int bit = ((irq & 0xf) * 2);
    return !!(gic_dist->config[word] & BIT(bit + 1));
}
```

and adds the check:
```diff
ackInterrupt(irq_t irq)
{
    assert(IS_IRQ_VALID(active_irq) && (active_irq & IRQ_MASK) == irq);
+    if(__is_edge_triggered(irq)){
+        __pending_clr(irq);
+    }
    gic_cpuiface->eoi = active_irq;
    active_irq = IRQ_NONE;
}
```

lsf37 (approved):
I'm also happy with this, currently running the proof check to see where (and if) it beaks. Almost all of this is behind the machine interface, but I think `ackInterrupt` itself will probably break.

Please hold off merging this until verification is done. It looks like a small update, which I can do on the side.

kent-mcleod:
> But for timer interrupts you do want to get the pending one immediately, so you can handle the timeout event as quickly as possible. With edge-triggered interrupts you want to ack them before handling them, otherwise you may miss events.

I agree with this.  I'm just trying to understand why the branch and clearPending call was added in the first place.  At the time I don't believe there were edge triggered interrupts being delivered to userlevel as there was no code for programming IRQs to be anything other than level triggered. The only other case would be it was for handling the kernel timer interrupt. Why would it be needed for the kernel timer interrupt though? 

Another possibility is that it was added based on an incorrect understanding of how edge triggered interrupts are processed.

lsf37:
The proofs are actually still passing, they are ending at the `ackInterrupt` interface, so they don't see the change. This also means you need to be extra sure this is right before you merge, but you do have the green light from verification :-)

Indanz:
Someone could ask Alex, assuming he can be found?

I am sure it's not needed.

lsf37:
> Someone could ask Alex, assuming he can be found?

I don't have any current contact details for Alex, it's been quite a while (10 years).

I'm happy with the change, there is nothing to indicate in the original commit that this was done for very deep reasons.

kent-mcleod:
Ok. lets merge it

PR #1120, opened 2023-10-25
aarch64: (verification) performPageInvocationUnmap, performASIDPoolInvocation, checkVPAlignment

Xaphiosis:
This is a bunch of tweaks that make it easier for verification to deal with cap updates, and also to make some things more like what's on the other platforms.

TLDR: a single update using a `_ptr_set_` generated function is fine, but if multiple updates are needed, best fetch the thing, do updates to a local var, and then post back.

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1435,6 +1435,11 @@ static exception_t decodeARMPageTableInvocation(word_t invLabel, unsigned int le
     return performPageTableInvocationMap(cap, cte, pte, ptSlot.ptSlot);
 }
 
+static inline bool_t CONST checkVPAlignment(vm_page_size_t sz, word_t w)
+{
+    return (w & MASK(pageBitsForSize(sz))) == 0;
+}
+
 static exception_t decodeARMFrameInvocation(word_t invLabel, unsigned int length,
                                             cte_t *cte, cap_t cap, bool_t call, word_t *buffer)
 {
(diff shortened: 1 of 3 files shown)


src/arch/arm/64/kernel/vspace.c
```diff
@@ -1172,8 +1172,12 @@ static exception_t performPageInvocationUnmap(cap_t cap, cte_t *ctSlot)
                   cap_frame_cap_get_capFBasePtr(cap));
     }
 
-    cap_frame_cap_ptr_set_capFMappedASID(&ctSlot->cap, asidInvalid);
-    cap_frame_cap_ptr_set_capFMappedAddress(&ctSlot->cap, 0);
+    cap_t slotCap = ctSlot->cap;
```
axel-h, line 1175: Can't the variable name be simply `cap` to be consistent with the naming?

Xaphiosis: Indeed. I copied it from RISCV so that the proof would be identical, did not notice that page inv. unmap and asid pool invocations use different names for the var. Now the question is whether to be consistent with RISCV or with the other function, as we don't have time to update RISCV right this second. Thoughts welcome.

axel-h: I don't know enough about the proof side and the impact there. Are parts of the proofs shared? I consider the name "slotCap" a bit unfortunate because we have "cap slots" also. So using just "cap" makes the C code a bit easier to understand in term of what is done.

kent-mcleod (approved):

src/arch/arm/64/kernel/vspace.c
kent-mcleod: We should do whatever makes the verification work easier.

Xaphiosis: It's only a small impact on verification (the C refinement proofs are not shared between arches, maybe someday), so yeah, might be nicer as "cap" then in line with the other functions. I'll do that, and we can leave the RISCV64 tweak to some another day.

lsf37 (approved):
Happy with this change. I don't really mind whether it's `cap` or `slotCap`, both are in use in other functions, but probably best to stay consistent within the same architecture file.

src/arch/arm/64/kernel/vspace.c
lsf37: Right, in that case `slotCap` it is. The name is used a lot in decode functions, it just means "the cap in `slot`".

PR #1121, opened 2023-10-25
aarch64: use VS prefix for vspace_cap fields

Xaphiosis:
The vspace (top level PT) cap had non-standard names which were a bit confusing to reason about. The name conflict with capPTBasePtr also spits out fully-qualified names in verification.
This commit updates all the field names of vspace_cap to start with "capVS" and updates all the call sites.


lsf37 (approved):
Looks good. The empty/PT prefix on vspace caps was probably to try and unify the names with page table caps, but it's nicer on the verification side to have the field names globally unique.

Indanz (approved):

Issue #530, opened 2021-08-31
manual: Missing documentation for RISC-V seL4_IRQControl_GetTrigger


canarysnort01:
This is defined in the IDL but doesn't show up in the manual. I think it is because the interface name and method name are identical to one that exists on arm.

https://github.com/seL4/seL4/blob/12.1.0/libsel4/arch_include/riscv/interfaces/sel4arch.xml#L169

canarysnort01:
I've done some further analysis on this.

I can see that two definitions of `seL4_IRQControl_GetTrigger` are written into `manual/stage/object_invocations.h` by `manual/tools/gen_invocations.py`, one for ARM and another for RISC-V.

When doxygen processes this file, it appears to ignore the duplicate RISC-V definition, as `manual/doxygen-output/xml/group__riscv__seL4__IRQControl.xml` does not contain the definition.

kent-mcleod:
It makes sense to me that the issue is caused by Doxygen not allowing multiple functions with the same name within a single invocation. When I've run into this problem elsewhere the fix has been to have separate doxygen output xml folders and then selectively import functions into the document.

lsf37:
LaTeX is not the only thing generated, though, there is at least also markdown output, which goes onto the docs site, eg. <https://docs.sel4.systems/projects/sel4/api-doc.html#sel4_tcb>.

It may be sufficient to produce markdown directly, and convert that to LaTeX.

lsf37:
We could branch off just LaTeX and do a separate thing there, but that would make the output diverge and still leave the same issue for other formats. I'm not saying we need to stick with doxygen, but if we drop it, we should drop it for all of the paths.

pingerino:
I am somewhat responsible for the evolution of this tool chain of docs, so here's my 2c:

- The assumption was that some people would like to view the documentation as doxygen.
- It also allows us to generate docs for things not already in XML but comments in the C code. 

Ultimately my goal was to express the entire libsel4 api in a machine readable format and present consistent docs and code in whatever format was convenient for the consumer (latex vs in the code vs the doc site vs doxygen) that would always be up to date. 

The doxygen step was just a half way point that let us get the comments out of the code and out of the xml using an existing tool, but the intention was always to remove this step.

lsf37:
Generating libsel4 + possibly other language bindings is definitely on the cards. We're trying to line up some funding for that, and we should probably get the set of people together who are interested in the topic (specifically also for Rust bindings).

[...]

Issue #558, opened 2021-09-06
seL4_TCB_SetSpace is not documented for MCS


bennoleslie:
The `SetSpace` call for MCS has different parameters than non-MCS kernels, however the documentation does not mention this.

bennoleslie:
My understanding is that `fault_ep`  on non-MCS is an `seL4_Word` and is a cap address in the context of the `cspace_root`, while on MCS `fault_ep` is an `seL4_CPtr` and is a cap address in the caller's CSpace. Maybe for most cases this doesn't matter as libsel4 hides the difference. It definitely matter if you are making your own system call bindings.

pingerino:
The documentation does have it (you can see it in the source) but it's probably not being generated properly: https://github.com/seL4/seL4/blob/master/libsel4/include/interfaces/sel4.xml#L259

bennoleslie:
I have no idea how the documentation generator works, but could it be because the `manual_name` attribute is the same as for the non-MCS `manul_name` ? The other methods with variations for MCS have ` (MCS)` suffix. For example: https://github.com/seL4/seL4/blob/master/libsel4/include/interfaces/sel4.xml#L197

pingerino:
Possibly, unfortunately I don't have time to get my hands dirty with code at the moment to check. Just emptying the contents of my head in the hope that it is useful :).

bennoleslie:
Very helpful thanks! Someone with more knowledge of how the manual is built (perhaps @canarysnort01 ) might be able to take a look with that point.

canarysnort01:
> I have no idea how the documentation generator works, but could it be because the `manual_name` attribute is the same as for the non-MCS `manul_name` ?

I haven't looked at the root cause yet but I think this is very likely. This is similar to #530.

lsf37:
We figured out in #530 what to do for this, so hopefully the fix there will also resolve this issue.

PR #1123, opened 2023-11-04
manual: handle name duplication between groups 

lsf37:
Different API groups may contain the same function name, for instance `IRQ_Control` `GetTrigger` for RISC-V vs the same for ARM. Duplicate function names with identical parameter lists confuse `doxygen`, leading it to generate a single merged xml entry for both, which means one of the entires will be missing and the other will be potentially wrong.

As @Ivan-Velickovic  and I found out in #530, when the functions are placed in different files and different groups at the same time, doxygen no longer is confused in all cases.

Therefore:

- generate a separate file for each API group
- generate a separate file group_defs.h that contains group definitions
  and declares group nesting

Unfortunately, this does not seem to always work (e.g. the toplevel MCS/non-MCS syscalls), so manual inspection is still necessary when adding new calls and separate doxygen runs for duplicate function names may be necessary. Generating separate files as above enables this option, should it become necessary in the future.

This fixes https://github.com/seL4/seL4/issues/530 and could be the basis for addressing #558

Also in this PR:
- parse_doxygen_xml: avoid XMLParsedAsHTMLWarning
- trivial: spelling (sel4 -> seL4)

### DIFF
--- a/manual/tools/gen_invocations.py
+++ b/manual/tools/gen_invocations.py
@@ -13,9 +13,9 @@ doxygen comments from a given interface defined in an xml file.
 import os
 import sys
 import argparse
-import operator
 import logging
 import itertools
+from pathlib import Path
 from libsel4_tools import syscall_stub_gen
 from lxml import etree
 
@@ -53,7 +53,7 @@ def generate_prototype(interface_name, method_name, method_id, inputs, outputs, 
     return "%s\n%s %s %s(%s);" % (comment, prefix, return_type, name, param_list)
 
 
-def gen_invocations(input_files, output_file, args):
+def gen_invocations(input_files, output_dir, args):
     """
     Given a collection of input xml files describing seL4 interfaces,
     generates a c header file containing doxygen-commented function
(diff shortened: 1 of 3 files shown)


lsf37:
I checked the generated manual, and the only difference is that GetTrigger now appears for RISCV as well.

manual/tools/gen_invocations.py
```diff
@@ -79,24 +88,37 @@ def gen_invocations(input_files, output_file, args):
...
+                else:
+                    prototypes[group_id].append(prototype)
+
+        with open(group_file_name, "a") as groups_file:
+            groups_file.write("/**\n * @defgroup %s %s\n * @{\n */\n\n" % (api.name, api.name))
+            for group_id, group_name in sorted(groups.items()):
+                groups_file.write("/**\n * @defgroup %s %s\n */\n\n" % (group_id, group_name))
+            groups_file.write("/** @} */\n\n")
+
+        for group_id, prototypes in prototypes.items():
```
Ivan-Velickovic, line 115: I don't like having a shadowed variable here as it is often the source of errors/confusion in Python code, but this is just a minor point.

Ivan-Velickovic (approved):
I do not see any issue with the logic. Glad we can finally resolve #530.

manual/tools/gen_invocations.py
lsf37: You're right, that is an unfortunate name choice. I'll tweak it.

lsf37:
Rebased and addressed name tweak.

PR #1125, opened 2023-11-05
manual: group invocations by MCS/non-MCS

lsf37:
(this PR is on top of #1123)

Put MCS-only invocations into their own groups and files. This solves the problem that doxygen gets confused by duplicate function names with the same parameters.

MCS/non-MCS is distinguished by evaluating the `<condition>` field in the API XML definition. If the condition evaluates to true when `CONFIG_KERNEL_MCS` is set, it is an MCS-only method, otherwise it is assumed to be non-MCS or present in both configs.

The longest part of all of this is parsing the condition field, but it's still relatively short and I don't think it warrants use of a full-blown parser generator or library.

Fixes https://github.com/seL4/seL4/issues/558

Also:
- distinguish SetSpace (MCS) by name in the manual to avoid confusion for the reader
- remove obsolete doxygen settings to avoid warnings in more recent versions

### DIFF
--- a/libsel4/include/interfaces/sel4.xml
+++ b/libsel4/include/interfaces/sel4.xml
@@ -487,7 +487,7 @@
             </error>
         </method>
 
-        <method id="TCBSetSpace" name="SetSpace" manual_name="Set Space" manual_label="tcb_setspace_mcs">
+        <method id="TCBSetSpace" name="SetSpace" manual_name="Set Space (MCS)" manual_label="tcb_setspace_mcs">
             <condition><config var="CONFIG_KERNEL_MCS"/></condition>
             <brief>
                 Set the fault endpoint, CSpace and VSpace of a thread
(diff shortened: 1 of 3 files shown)


Ivan-Velickovic:
Have we checked there are no other cases like `SetSpace`? Or is it just an edge-case?

Ivan-Velickovic (approved):
Looks good. I have not checked the parsing logic thoroughly given that this is not super critical code.

lsf37:
> Have we checked there are no other cases like `SetSpace`? Or is it just an edge-case?

I have not actually checked and we probably should. Should be easy enough to get a list of MCS calls from the xml files and check them against the PDF.

lsf37:
(rebased)

lsf37:
Ok, looked through them all, and it was only SetSpace that was missing.

PR #1127, opened 2023-11-06
MCS: preserve sc_sporadic invariant

lsf37:
#498 stipulates that `sc_sporadic` should imply `sc_active`, but the code showed assertion failures for a corresponding check.

It turns out that the failing assertion was wrong in the sense that it checked a different condition. It also turns out that the invariant as stipulated there is true if `scSporadic` is reset in `finaliseCap`. Thanks to @michaelmcinerney for proving that.

This PR:
- fixes `finaliseCap` so that the expected invariant holds
- simplifies `sc_sporadic` again so that it does not have to check `sc_active`
- adds the correct corresponding assertion
- slightly tweaks code ordering in `invokeSchedControl_ConfigureFlags` to make verification easier

### DIFF
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -160,7 +160,10 @@ static inline bool_t sc_released(sched_context_t *sc)
  */
 static inline bool_t sc_sporadic(sched_context_t *sc)
 {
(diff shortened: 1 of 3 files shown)


lsf37:
(updated commit msg)

michaelmcinerney (approved):
Looks good!

PR #1132, opened 2023-11-14
debug: use c99 standard instead of gcc extensions

axel-h:


axel-h:
> There's another one in libsel4/include/sel4/assert.h, please change that one too.

Done

Indanz (approved):

PR #1133, opened 2023-11-14
exynos4: remove explicit setting of TIMER_PRECISION

axel-h:
The default value is zero anyway.

### DIFF
--- a/src/plat/exynos4/config.cmake
+++ b/src/plat/exynos4/config.cmake
@@ -25,7 +25,6 @@ if(KernelPlatformExynos4)
         KERNEL_WCET 10u
         CLK_MAGIC 2863311531llu
         CLK_SHIFT 36u
-        TIMER_PRECISION 0u
     )
 endif()
 


andybui01:
Is there any documentation for how TIMER_PRECISION and TIMER_OVERHEAD_TICKS are obtained? This might help with future ARMv8 MCS+SMP ports, where the `clock_sync_test()` might fail, for example.

axel-h:
I assume these values are set whenever it turns out to be necessary.

Indanz:
`clock_sync_test` has nothing to do with those config option, that tests whether clocks used by different cpu cores are synchronised.

commit 4a7d08def09b2b8b30cb76dee15144a9cfaeb5c7
Author: Indan Zupancic <Indan.Zupancic@mep-info.com>
Date:   Thu May 19 15:22:18 2022 +0200

    MCS, ARM: Introduce TIMER_OVERHEAD_TICKS
    
    For ARM currently TIMER_PRECISION exists, but that is in microseconds
    and not fine-grained enough.
    
    This is needed to make periodic tasks synchronous with the system clock.
    If this value is zero every period will be extended with the overhead of
    taking an interrupt and reading the system clock. To avoid this drift,
    the configured value should be set to at least the average overhead.
    
    See also issue #844.

I agree that a comment would have helped, but there wasn't one for `TIMER_PRECISION` either... Bad excuse, I know.

Indanz (approved):

Indanz:
Tests for `TIMER_OVERHEAD` are added by https://github.com/seL4/sel4test/pull/76, but not merged yet. The test should probably only run if `TIMER_OVERHEAD_TICKS` is set, but not sure if that's exported or not.

It's only relevant for platforms that care about precise timing.

axel-h:
After merging this `ODROID_C4_release_SMP_hyp_clang_64` failed in `MULTICORE0004` with a timeout, see https://github.com/seL4/seL4/actions/runs/6881147660/job/18718428807#step:4:30957
This seems unrelated to this change, as it uses a different platform.

Indanz:
Yes, that's a known failure case and unrelated to your commit.

PR #1135, opened 2023-11-17
NVIDIA Jetson Orin support

andybui01:
**Summary:** initial support for the NVIDIA Jetson Orin

I'm currently tidying up the elfloader and libplatsupport changes, but the kernel port is mostly ready to be reviewed.

**sel4test results**
no-MCS no-SMP Test suite passed. 124 tests passed. 47 tests disabled.
MCS no-SMP Test suite passed. 150 tests passed. 40 tests disabled.
MCS SMP Test suite passed. 153 tests passed. 37 tests disabled.

A significant amount of tests are disabled as I couldn't get the ARM generic ltimer implementation to play nice, any test that required the timer to sleep would not sleep at all, causing a lot of instability, for example in `MULTICORE0001` this check fails:

https://github.com/seL4/sel4test/blob/ab9189188ff3eff46d703e36081350f0f637b647/apps/sel4test-tests/src/tests/multicore.c#L33

Where both counter and old_counter remain at 0, even though the counter should've incremented. I get similar failures on `PREEMPT_REVOKE` and `SCHED0008`, as well as a few others I have forgotten.

**Caveats:**
The Orin has a couple 8250 serial devices that _could_ use the `tegra_omap3_dwapb.c` driver in the kernel, but as I have only tested with the `tegra194-tcu` device, I'll leave that out for now.


include/arch/arm/arch/model/smp.h
```diff
@@ -13,7 +13,11 @@
 #ifdef ENABLE_SMP_SUPPORT
 static inline cpu_id_t cpuIndexToID(word_t index)
 {
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    return index;
+#else
     return BIT(index);
+#endif
 }
```
Indanz, line 21: I'd prefer to remove this function altogether and do any `cpuIndexToID` operation required within `ipi_send_target` itself.

src/arch/arm/config.cmake
```diff
@@ -32,6 +32,11 @@ elseif(KernelArmCortexA72)
     # (https://developer.arm.com/documentation/100095/0001/memory-management-unit/about-the-mmu)
     set(KernelArmPASizeBits44 ON)
     math(EXPR KernelPaddrUserTop "(1 << 44)")
+elseif(KernelArmCortexA78AE)
+    # Even though CortexA78AE supports 48-bits
+    # 44 bits PA is used
```
axel-h, line 37: This seems Oring specific then? So the comment should clearly say this, or we could even have a platform-specific setting then for Orin besides the A78AE's default

src/drivers/serial/tegra194-tcu.c
```diff
@@ -0,0 +1,43 @@
...
+    /* We are writing one byte */
+    reg_val = (uint32_t)(1UL << NUM_BYTES_FIELD_BIT);
+    reg_val |= BIT(INTR_TRIGGER_BIT);
+    reg_val |= c;
+
+    if (c == '\r' || c == '\n') {
+        reg_val |= BIT(FLUSH_BIT);
+    }
+
+    while (*UART_REG(UART_PPTR, 0) & BIT(INTR_TRIGGER_BIT));
```
axel-h, line 32: Please define a name for the register at offset 0 also.

src/arch/arm/machine/gic_v3.c
```diff
@@ -345,34 +340,19 @@ BOOT_CODE void cpu_initLocalIRQController(void)
 }
 
 #ifdef ENABLE_SMP_SUPPORT
-#define MPIDR_MT(x)   (x & BIT(24))
-
-void ipi_send_target(irq_t irq, word_t cpuTargetList)
+/* This function is called with a single node. */
+void ipi_send_target(irq_t irq, word_t cpuID)
```
axel-h, line 344: The prototype for this function is in `include/smp/ipi.h`, but the semantics change here. We have align this everywhere then to have a consistent behavior, otherwise this will get quite confusing across architectures.

src/arch/arm/config.cmake
andybui01: Ah, that should be more clear, while the A78AE supports 48 bits, I haven't figured out a way to enabke this without causing a chain reaction of assertion failures within the kernel.

My understanding is that the ARM64 kernel only has _47_ useable bits of physical memory unless hypervisor mode is enabled, if anyone knows the context to this please share :).

src/arch/arm/machine/gic_v3.c
andybui01: The semantics change here because the RISC-V (and I think x86) kernel has a similar example. Really, ipi_send_target is only used to send to one target at a time, and this could be reworked to be explicit instead.

However, I'd like to avoid touching any code that might require additional verification efforts for this port.

axel-h: Well, if we have a function with a feature set nobody uses, but everybody uses this in a simpler form, then it seems worth cleaning this up instead of making more wierd implementations. At least we could make dedicated PR for that, so the verification effort becomes better assessible. There are clear use cases why this change seems useful - or somebody can put up use cases we are not seeing.

Indanz:
So how much does Cortex-A78AE differ from Cortex-A78? That is, are the AE differences visible to seL4? If not, perhaps just call it Cortex-A78? Or leave this to be figured out by whoever needs Cortex-A78 support?

src/arch/arm/machine/gic_v3.c
```diff
@@ -14,8 +14,10 @@
 #define RDIST_BANK_SZ 0x00010000
 /* One GICR region and one GICR_SGI region */
 #define GICR_PER_CORE_SIZE  (0x20000)
-/* Assume 8 cores */
-#define GICR_SIZE           (0x100000)
+/* Assume 12 cores
+ * NOTE: this is hardcoded to the same value for the GICR region in hardware.yml
+ */
+#define GICR_SIZE           (0x180000)
```
Indanz, line 20: I wonder, could we use `CONFIG_MAX_NUM_NODES * GICR_PER_CORE_SIZE` instead?

include/arch/arm/arch/model/smp.h
andybui01: Same, would it be OK to open a separate issue for this?

andybui01:
> So how much does Cortex-A78AE differ from Cortex-A78? That is, are the AE differences visible to seL4? If not, perhaps just call it Cortex-A78? Or leave this to be figured out by whoever needs Cortex-A78 support?

They're pretty similar, the seL4-visible difference would be that the A78 is only 40-bit PA while the A78AE is 48-bit PA: https://developer.arm.com/documentation/102826/latest/

include/arch/arm/arch/model/smp.h
Indanz: I'm fine with a new PR that reworks `ipi_send_target` to only take a single target instead of a list, together with moving all `cpuIndexToID` into architecture specific code.

The only reason to do it the way you did is to hide any changes from verification. All it does is move the `cpuIndexToID` from the caller to where it's actually needed, I hope that's also a trivial change on the verification side.

src/arch/arm/config.cmake
Indanz: Considering the Jetson Orinmodules don't support more than 64GB, let alone 1TB, why not treat is as a Cortex-A78?

Or are you also adding lockstep support to seL4 later on?

src/drivers/serial/tegra194-tcu.c
```diff
@@ -0,0 +1,44 @@
...
+#define TX_INTR_TRIGGER_BIT     (31)
+#define UART_REG(mmio, x)       ((volatile uint32_t *)(mmio + (x)))
+
+#ifdef CONFIG_PRINTING
+void uart_drv_putchar(unsigned char c)
+{
+    uint32_t reg_val;
+
+    /* We are writing one byte */
+    reg_val = (uint32_t)(1UL << TX_NUM_BYTES_FIELD_BIT);
```
Indanz, line 25: Any reason you're not using the BIT macro here?

Also, it would be nicer to move all the BIT() into the define itself instead of at the user's sites.

tools/hardware.yml
```diff
@@ -46,8 +46,10 @@ devices:
       # GICRedistributor
       - index: 1
         kernel: GICR_PPTR
-        # Assume 8 cores max
-        kernel_size: 0x100000
+        # Assume 12 cores max
+        # Changing this requires changing the GICR_SIZE constant
+        # in gic_v3.c as well.
+        kernel_size: 0x180000
```
Indanz, line 52: Why 12 and not 16? I wonder if it's not time to make GICR_SIZE a platform define, with the default being 8.

Indanz (approved):
All in all code seems fine.

src/drivers/serial/tegra194-tcu.c
axel-h: Using `BIT()` seems semantically wrong here, because this is setting a shifted value.

Indanz: In that case, the name of `TX_NUM_BYTES_FIELD_BIT` is misleading, it should be called `TX_NUM_BYTES_FIELD_SHIFT`.

andybui01: Ah yes, I'll change that.

src/drivers/serial/tegra194-tcu.c
```diff
@@ -0,0 +1,44 @@
...
+#include <config.h>
+#include <stdint.h>
+#include <util.h>
+#include <machine/io.h>
+#include <plat/machine/devices_gen.h>
+
+#define TCU_TX_REG              (0)
+#define TX_NUM_BYTES_FIELD_BIT  (24)
+#define TX_FLUSH_BIT            (26)
+#define TX_INTR_TRIGGER_BIT     (31)
```
Indanz, line 16:
```suggestion
#define TX_NUM_BYTES(n) ((n) << 24)
#define TX_FLUSH            BIT(26)
#define TX_INTR_TRIGGER     BIT(31)
```

tools/hardware.yml
andybui01: Just don't want to be excessive, considering the existing boards only use 8 cores.

include/arch/arm/arch/model/smp.h
andybui01: I've opened an issue [here](https://github.com/seL4/seL4/issues/1138) so that this does not get lost.

tools/hardware.yml
Indanz: But both untyped regions as hardware peripheral regions are usually power-of-two, so I don't see any downside to upping it to 16 instead of 12..

Indanz (approved):

tools/hardware.yml
andybui01: I double checked `devices_gen.h`. Seems like the kernel devices are always mapped in as 4K pages, and with the GICR region specifically, it is not `userAvailable` so wouldn't be created as untyped right?

lsf37:
> A significant amount of tests are disabled as I couldn't get the ARM generic ltimer implementation to play nice, any test that required the timer to sleep would not sleep at all, causing a lot of instability, [..]

So, what's the plan for this? It does not seem like a good idea to merge a port that needs a lot of tests disabled..

andybui01:
https://github.com/seL4/seL4/actions/runs/7823562148/job/21344707601?pr=1135

Is the preprocess failure suggesting changes to me? Not sure I understand what it's trying to say.

lsf37:
> Is the preprocess failure suggesting changes to me? Not sure I understand what it's trying to say.

It's showing the code changes that are visible to verification and that might break the proofs. Will have a look.

lsf37:
> > Is the preprocess failure suggesting changes to me? Not sure I understand what it's trying to say.
> 
> It's showing the code changes that are visible to verification and that might break the proofs. Will have a look.

This will indeed break verification, because it changes non-platform code. I'll see if I can comment on the code.

src/arch/arm/config.cmake
lsf37: There is some discussion on 48 bit physical address space in https://github.com/seL4/seL4/issues/1175 and https://github.com/seL4/seL4/pull/1157, but the result is basically that the full 48 bit are currently not supported and would require further changes.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -665,15 +664,12 @@ static pte_t makeUserPagePTE(paddr_t paddr, vm_rights_t vm_rights, vm_attributes
...
-
     if (page_size == ARMSmallPage) {
         return pte_pte_4k_page_new(nonexecutable, paddr, nG, 1 /* access flag */,
-                                   shareable, APFromVMRights(vm_rights), attridx);
+                                   INNER_SHAREABLE, APFromVMRights(vm_rights), attridx);
     } else {
         return pte_pte_page_new(nonexecutable, paddr, nG, 1 /* access flag */,
-                                shareable, APFromVMRights(vm_rights), attridx);
+                                INNER_SHAREABLE, APFromVMRights(vm_rights), attridx);
     }
```
lsf37, line 673: This is the part that breaks verification currently. It would be relatively easy to update, but I'd like to understand better what we are really doing here. Why are we usually not setting `inner shareable`, and could setting it unconditionally have any adverse effect on older platforms?

Is there any potential interaction with device memory?

The explanation in the commit message was useful. I think it would be good to have more of it as a comment in the code (maybe where INNER_SHARABLE is defined).

andybui01: > Is there any potential interaction with device memory?

None, ARM states in RPYFVQ (DDI 0487J.a) that the value doesn't matter for device memory, which it will just treat as Outer Shareable.

> Why are we usually not setting inner shareable, and could setting it unconditionally have any adverse effect on older platforms?

I think the current reasoning is that it's not really _needed_ for UP systems. Having a more permissive shareability such as `inner shareable` might mean wasted cycles trying to be coherent with the other PEs when it's not really needed. However, I believe that even while running seL4 in UP mode, another agent in the system may still need to be coherent with the PE/node. I have not had a deeper look as to which part of the Orin SoC might be requiring this.

andybui01: Example: We've had some existing problems with the Orin's System Cache. It is not an architectural cache and is shared by the CPU/GPU, making it effectively a CPU L4 cache. It may be the case that this is _not_ within the non-shareable domain.

I use this example because we added a cache line invalidate by VA on a range of memory, which seemed to half-fix some of the instabilities we were encountering. Setting the shareability to IS fully fixed this issue and made the fix redundant.

lsf37: Thank you, that makes sense. It would be good to see if there is any measurable performance impact then, even if we're not expecting any. I guess we could just kick off a sel4bench run and see what happens on the existing platforms.

andybui01:
Cursory comparison of the TX1 benchmark run with the 2024-02-09 results here https://sel4.systems/About/Performance/home.pml:

Figures in cycles (std dev)

Mainline (Non-shareable mappings):
Call: 403 (5)
Reply: 415 (0)
MCS Call: 433 (9)
MCS Reply: 453 (12)

This PR (Inner-Shareable mappings):
Call: 406 (12)
Reply: 416 (1)
MCS Call: 432 (8)
MCS Reply: 444 (14)

src/arch/arm/64/kernel/vspace.c
Indanz: > However, I believe that even while running seL4 in UP mode, another agent in the system may still need to be coherent with the PE/node.

Being cache coherent with other cores should be the default behaviour for seL4, to avoid very hard to debug surprises. So I am strongly in favour of this change.

> It can also mean that the system has to fall back to uncached memory if it doesn't provide hardware cache-coherency. So increasing the shareability from none, to inner to outer can degrade performance if it's not needed.

Those pages would be marked `cacheable` and this change doesn't change anything for them on SMP.

These SoCs are made for coherency between cores, I don't think downgrading non-SMP performance to the same level as with SMP for some obscure corner case is a problem.

And the proper fix for that would be to extend the API and provide more fine grained cacheability and shareability attribute values to user space, like you propose.

> Wouldn't this be an issue with inner vs outer cacheability attributes rather than shareability?

It does sound like that, so I'm also a bit surprised this change fixed that. Maybe the GPU is in the same cluster?

src/arch/arm/machine/gic_v3.c
Indanz, line 20: Perhaps a good idea to add a `compile_assert(CONFIG_MAX_NUM_NODES * GICR_PER_CORE_SIZE <= GICR_SIZE);` to reduce future frustration.

Indanz:
Did you manage to fix the timer problems in sel4test?

andybui01:
> Did you manage to fix the timer problems in sel4test?

Unfortunately no. We don't use libplatsupport so I'll need to find the bandwidth to take a deeper look.

Indanz:
> > Did you manage to fix the timer problems in sel4test?
> 
> Unfortunately no. We don't use libplatsupport so I'll need to find the bandwidth to take a deeper look.

How do you know the problem is in libplatsupport? Do other interrupts do work in your port, or is it just the sel4test timer not getting any? Or is the problem that the interrupt triggers immediately? My concern is that there is a deeper underlying problem that causes this issue, so I really want to get to the bottom of this before merging.

src/arch/arm/64/kernel/vspace.c
kent-mcleod: Could this commit be split out into a separate PR and discussed separately. It's scope is more than the Jetson Orin support.

PR #1137, opened 2023-11-17
CMake config property cleanup

axel-h:
- style: support property `SMMU`. This was missing and gives some annoying style complaints
- use `MAX_IRQ` instead of `PLIC_MAX_NUM_INT`. Was also missing in the style file and thus there are annoying complaints. But it turns out there is no reason to have this, we can simply share `MAX_IRQ` from ARM.


Indanz (approved):

kent-mcleod:
> use `MAX_IRQ` instead of `PLIC_MAX_NUM_INT`. Was also missing in the style file and thus there are annoying complaints. But it turns out there is no reason to have this, we can simply share `MAX_IRQ` from ARM.

This is wrong. On arm `MAX_IRQ` is assigned to `maxIRQ` and represents the maximum logical irq number that the kernel uses to size the storage for interrupt handlers. But on riscv  `PLIC_MAX_NUM_INT` is used to set `PLIC_MAX_IRQ` and `maxIRQ` is set to a different value. This was why the parameters were explicitly given different names.

Indanz:
> But on riscv PLIC_MAX_NUM_INT is used to set PLIC_MAX_IRQ and maxIRQ is set to a different value. This was why the parameters were explicitly given different names.

So the code is still correct, but it's more confusing now because `maxIRQ` has a different value than `MAX_IRQ`, while having almost the same name?

kent-mcleod:
> So the code is still correct, but it's more confusing now because `maxIRQ` has a different value than `MAX_IRQ`, while having almost the same name?

The proofs use `maxIRQ` too and the values are usually manually copied across for each verification platform. If you see the definition for RISCV64 (linked below) it's (correct) definition of maxIRQ for hifive is 54, while the sel4.git sources now also have `MAX_IRQ 53`.


- https://github.com/seL4/l4v/blob/master/spec/machine/RISCV64/Platform.thy#L166-L168
- https://github.com/seL4/l4v/blob/master/spec/machine/ARM/Platform.thy#L72-L73

axel-h:
I see your point here, the naming clash is a bit unfortunate. So, should this have a generic name like `NUM_PLAT_INTR` then? The way I have seen this config option so far was, that it is about the number interrupts on the platform (or the max interrupt) - so what one find in a SoC TRM. Details still depend on the architecture and platform specific details. This is not necessarily the same number, as the number of interrupts the kernel internally manages. But this is an seL4 internal detail then and may change with the internal needs.

PR #1141, opened 2023-11-24
riscv: drop unused PLIC_IRQ_OFFSET

axel-h:
I can't find a good reason why `PLIC_IRQ_OFFSET` is needed today, so let's remove this to avoid confusion. It was added in with 9a552d84b6f4f77ccf2b5e0e73d14b91ee7ed3d2 but I'm not sure why it was needed there, and then 375a98c8b36cea2dc3a9c8e94545a582f5c23548 moved it to this file.

### DIFF
--- a/src/arch/riscv/platform_gen.h.in
+++ b/src/arch/riscv/platform_gen.h.in
@@ -27,8 +27,7 @@
  * field name.
  */
 enum IRQConstants {
-    PLIC_IRQ_OFFSET = 0,
-    PLIC_MAX_IRQ = PLIC_IRQ_OFFSET + (@CONFIGURE_MAX_IRQ@),
+    PLIC_MAX_IRQ = @CONFIGURE_MAX_IRQ@,
 #ifdef ENABLE_SMP_SUPPORT
     INTERRUPT_IPI_0,
     INTERRUPT_IPI_1,


Indanz (approved):

kent-mcleod:
> I can't find a good reason why `PLIC_IRQ_OFFSET` is needed today

It records the offset between PLIC interrupt numbers to seL4 logical interrupt numbers as it's not a 1-1 mapping. x86 has `IRQ_INT_OFFSET` for the same reason (but it's set to 0x20, not 0x0).

axel-h:
But why should be keep this, if it is practically not used anywhere. ARM also does not have this. if x86 needs this, that should not affect others, unless this is used in the generic kernel code also.

kent-mcleod:
> But why should be keep this, if it is practically not used anywhere. ARM also does not have this. if x86 needs this, that should not affect others, unless this is used in the generic kernel code also.

Because on RISC-V and x86 there's multiple domains of interrupt numbers and they need to get flattened into a single logical domain for the seL4 IRQ CNode. Arm IRQ numbers are assigned within a single domain and so directly map to IRQ CNode index. The RISC-V IRQ CNode has to have both PLIC irqs and CLINT irqs and so uses the same code structure as x86 which defines these enum values.

axel-h:
I understand you point. But I am a bit unhappy with keeping a zero here. Since ARM has 16 PPI and SGI each, would it make sense to reserve the first 32 IDs on RISC-V then also? The use case seems similar.

lsf37:
> I understand you point. But I am a bit unhappy with keeping a zero here. Since ARM has 16 PPI and SGI each, would it make sense to reserve the first 32 IDs on RISC-V then also? The use case seems similar.

It allows you to write code without baking in the assumption that this value will always be zero. If we want to change things later, the only thing we have to do now is change this one number. If we leave it out, the rest of the code will assume that there never will be an offset.

I'm not sure what the problem is with keeping such place holders and interfaces in general, the compiler will easily remove them and they communicate intent.

PR #1142, opened 2023-11-27
CMake: show DTS name in error message

axel-h:
Printing the input DTS file name makes more sense than printing the output DTB name on error. The output likely does not exists and thus this does not help.

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -164,7 +164,7 @@ if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
             RESULT_VARIABLE error
         )
         if(error)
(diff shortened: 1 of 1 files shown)


Indanz (approved):

Indanz:
Did you actually ran into this error message?

axel-h:
> Did you actually ran into this error message?

Yes. I had a typo in the input filenames. This is an error one can run into when adding a new platform or board variant.

Indanz:
In the case of fixing missing input files, having an `EXISTS` check on them would help, if the CMake `READ` silently fails when the file doesn't exist.

`KernelDTSIntermediate` is always set to "${CMAKE_CURRENT_BINARY_DIR}/kernel.dts", though that's the file you want to check if this command fails for syntax reasons I suppose.

Issue #566, opened 2021-09-09
Confusing parameter name `index` in CNode and Untyped object invocations


kent-mcleod:
Context: Invocations that need to address CNodes use a different addressing method than just a seL4_CPtr.  They need:
- a seL4_CPtr to a CNode cap that selects which CNode object to start the lookup from, 
- a seL4_Word of address/path information to use in the lookup.
- a seL4_Word of the number of bits to resolve in the lookup, the "depth".

The API documentation names the second parameter (with the address/path information) `index` or `node_index`.  I find this name makes it easy to confuse the parameter as being a CNode object index instead of an address/path of multiple CNode indexes.  

Another thing that contributes to the confusion is that the `index` is almost a seL4_CPtr and is treated the same if the depth argument is `WORD_BITS`, but for every bit less you want to resolve, the index argument needs to get right shifted by that amount.  However, elsewhere you can use a seL4_CPtr to perform an invocation on a non-CNode object located at a depth less than `WORD_BITS` and you aren't supposed to right shift by the difference.

A better name would be something like `address` or `slot_address`.

kent-mcleod:
A followup:
The following two operations do the same thing if the target slot doesn't contain a CNode.  Assume it contains an Endpoint cap.  Assume platform word width of 32. Assume a single CNode with radix size 20.  And the endpoint is in the CNode's second slot ([1] using array syntax).  The target slot is reached after resolving 20 bits.  When using seL4_Recv on the endpoint, the seL4_CPtr would be `0x00001000`.  

Calling revoke on the endpoint would be using a depth argument of 20.  The Index argument needs to be shifted `>>` by 12: 
```
seL4_CNode_Revoke(
      root_node_cap,
      0x00000001,
      20
    )
```

But the kernel supports exiting a translation early if it finds a cap that isn't a CNode.  So the following will also do the same thing:
```
seL4_CNode_Revoke(
      root_node_cap,
      0x00001000,
      32
    )
```

My opinion is that the following is more intuitive (If the kernel didn't require the index to be down shifted when depth was less than the word width):
```
seL4_CNode_Revoke(
      g_root_node_cap,
      0x00001000,
      20
    )
```
This way, the index slot the same as the seL4_CPtr and the depth is used to control how far the kernel processes from the top.

kent-mcleod:
Another reason why it's non-intuitive is because the way page-table addressing is handled in the API is the invocation takes an absolute virtual-address and doesn't require the unused bits to be `>>` shifted.

[...]

PR #1143, opened 2023-12-01
Add glossary to manual

bbrcknl:
- add a glossary to the manual
- consistent spelling of CSpace and IRQ in the API docs
- AU spelling and remove trailing whitespace in `manual/parts/io.tex`

### DIFF
--- a/manual/Makefile
+++ b/manual/Makefile
@@ -228,6 +228,7 @@ clean:
 		*.out *.ps *-diff.tex *.mps .log *.pdf *.tgz *~ *.lof *.lot env.tex
 	rm -rf ${DoxygenOutput} ${GeneratedLatexDir} ${GeneratedMarkdownDir}
 	rm -rf ${Stage}
+	rm -rf *.acn *.glo *.gls *.ilg *.ist
 
 tar:	clean
 	( p=`pwd` && d=`basename "$$p"` && cd .. && \
@@ -291,6 +292,8 @@ references.bib: $(addsuffix .tex, $(Targets))
 	${Q}if ! test -e $*.bbl || test $(Bib) -nt $*.bbl; then rm -f $*.bbl; fi
 	@echo "====> LaTeX first pass: $(<)"
 	${Q}$(LaTeX) $< >.log || if egrep -q $(Error) $*.log ; then cat .log; rm $@; false ; fi
+	@echo "====> Glossary: $(<)"
+	${Q}makeindex -s manual.ist -o manual.gls manual.glo
 	${Q}if egrep -q $(Rerun_Bib) $*.log ; then echo "====> BibTex" && $(BibTeX) $* > /dev/null && echo "====> LaTeX BibTeX pass" && $(LaTeX) >.log $< ; fi
 	${Q}if egrep -q $(Rerun) $*.log ; then echo "====> LaTeX rerun" && $(LaTeX) >.log $<; fi
 	${Q}if egrep -q $(Rerun) $*.log ; then echo "====> LaTeX rerun" && $(LaTeX) >.log $<; fi
--- a/manual/parts/glossary.tex
+++ b/manual/parts/glossary.tex
@@ -1,0 +1,116 @@
+%
+% Copyright 2014, General Dynamics C4 Systems
+%
+% SPDX-License-Identifier: GPL-2.0-only
+%
+
+\newglossaryentry{asid}{
+    name=ASID,
+    description={Address Space Identifier. Depending on architecture, the kernel provides software ASIDs, which are associated with VSpace root objects, and define the virtual address space of a thread. They are mapped to hardware ASIDs on demand when the architecture supports these. Multiple threads may be in the same address space}
+}
+
+\newglossaryentry{badge}{
+    name=Badge,
+    description={A badge is a piece of extra information stored in a capability, mostly used for endpoint and notification capabilities. It can be used by applications to identify caps previously handed out to clients}
+}
+
+\newglossaryentry{cap}{
+    name=Capability,
+    description={The main access control concept in seL4. A capability conceptually is a reference to a kernel object together with a set of access rights. Most seL4 capabilities store additional bits of information. Some of this additional information may be exposed to the user, but the bulk of it is kernel-internal book-keeping information. Capabilities are stored in CNodes and TCBs}
+}
+
+\newglossaryentry{cdt}{
+    name=CDT,
+    description={Capability Derivation Tree. A kernel-internal data structure that tracks the child/parent relationship between capabilities. Capabilities to new objects are children of the Untyped capability the object was created from. Capabilities can also be copied and result in either child or sibling capabilities, depending on the operation that was used and the depth of the existing derivation tree. The revoke operation will delete all children of the invoked capability}
+}
+
+\newglossaryentry{cnode}{
+    name=CNode,
+    description={Capability Node. Kernel-controlled storage that holds capabilities. Capability nodes can be created in different sizes and be shared between CSpaces}
+}
+
+\newglossaryentry{cptr}{
+    name=CPtr,
+    description={Capability Pointer. A user-level reference to a capability, relative to a specified root CNode or the thread’s CSpace root}
+}
+
+\newglossaryentry{cspace}{
+    name=CSpace,
+    description={A directed graph of CNodes. The CSpace of a thread defines the set of capabilities it has access to. The root of the graph is the CNode capability in the CSpace slot of the thread. The edges of the graph are the CNode capabilities residing in the CNodes spanned by this root}
+}
+
+\newglossaryentry{endpoint}{
+    name=Endpoint,
+    description={IPC is facilitated by small kernel objects known as endpoints, which act as general communication ports. Invocations on endpoint objects are used to send and receive IPC messages}
+}
+
+\newglossaryentry{guard}{
+    name=Guard,
+    description={Guard of a CNode capability. From the user’s perspetive the CSpace of a thread is organised as a guarded page table. The kernel will resolve user capability pointers into internal capability slot pointers. The guard of one link/edge in the CSpace graph defines a sequence of bits that will be stripped from the user-level capability pointer before resolving resumes at the next CNode}
+}
+
+\newglossaryentry{iommu}{
+    name=IOMMU,
+    description={Input–Output Memory Management Unit. Applies virtual address translation and memory protection to DMA capable I/O devices}
+}
+
+\newglossaryentry{iopagetable}{
+    name=IOPageTable,
+    description={This object represents a node in the multilevel page-table structure used by IOMMU hardware to translate hardware memory accesses}
+}
+
+\newglossaryentry{iospace}{
+    name=IOSpace,
+    description={This object represents the address space associated with a hardware device. It represents the right to modify a device’s address space. See \autoref{ch:io}}
+}
+
+\newglossaryentry{ipc}{
+    name=IPC,
+    description={Inter Process Communication is facilitated by endpoints, which act as general communication ports. Invocations on endpoint objects are used to send and receive messages}
+}
+
+\newglossaryentry{irqcontrol}{
+    name=IRQControl,
+    description={A single capability from which IRQHandler capabilities to all IRQ numbers in the system can be derived. This capability can be moved between CSpaces and CSlots but cannot be duplicated. Revoking this capability removes all IRQHandlers}
+}
+
+\newglossaryentry{irqhandler}{
+    name=IRQHandler,
+    description={Capabilities that represent the ability of a thread to handle a certain interrupt. See \autoref{ch:io}}
+}
+
+\newglossaryentry{notificationobject}{
+    name=Notification Object,
+    description={A word-size array of flags that provides a non-blocking signalling mechanism similar to a binary semaphore. Operations are signalling a subset of flags in a single operation, polling to check any flags, and blocking until any are signalled. Notification capabilities can be signal-only or wait-only}
+}
+
+\newglossaryentry{replyobject}{
+    name=Reply Object,
+    description={(MCS only) A reply object is a vessel for tracking reply messages, used to send a reply message and
+    wake up the caller}
+}
+
+\newglossaryentry{schedulingcontext}{
+    name=Scheduling Context,
+    description={(MCS only) An abstraction of CPU execution time}
+}
+
+\newglossaryentry{tcb}{
+    name=TCB,
+    description={Thread Control Block. The kernel object that stores management data for threads, such as the thread's CSpace, VSpace, thread state, or user registers}
+}
+
+\newglossaryentry{untypedmemory}{
+    name=Untyped Memory,
+    description={Memory that can be used to create kernel objects via the \apifunc{seL4\_Untyped\_Retype}{untyped_retype} invocation. It is the foundation of memory allocation in the seL4 kernel. See \autoref{sec:kernmemalloc}}
+}
+
+\newglossaryentry{vm}{
+    name=VM,
+    description={Virtual Memory. The concept of translating virtual memory addresses to physical frames. See \autoref{ch:vspace}}
+}
+
+\newglossaryentry{vspace}{
+    name=VSpace,
+    description={Virtual Address Space. In analogy to CSpace, the virtual memory space of a thread. See \autoref{ch:vspace}}
+}
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,11 @@
 *.blg
 *.toc
 *.mpx
+*.acn
+*.glo
+*.gls
+*.ilg
+*.ist
 
 tools/parsetab.py
 manual/env.tex
--- a/include/machine/interrupt.h
+++ b/include/machine/interrupt.h
@@ -80,7 +80,7 @@ static inline bool_t isIRQPending(void);
  * When an IRQ is disabled, it should not raise an interrupt on the processor.
  *
  * @param[in]  disable  True to disable IRQ, False to enable IRQ
- * @param[in]  irq      The irq to modify
+ * @param[in]  irq      The IRQ to modify
  */
 static inline void maskInterrupt(bool_t disable, irq_t irq);
 
@@ -93,7 +93,7 @@ static inline void maskInterrupt(bool_t disable, irq_t irq);
  * but before user level has handled the cause and does not imply that the cause
  * of the interrupt has been handled.
  *
- * @param[in]  irq   irq to ack
+ * @param[in]  irq   IRQ to ack
  */
 static inline void ackInterrupt(irq_t irq);
 
--- a/libsel4/include/sel4/syscalls.h
+++ b/libsel4/include/sel4/syscalls.h
@@ -87,14 +87,14 @@ seL4_DebugSnapshot(void);
 
 /**
  * @xmlonly <manual name="Cap Identify" label="sel4_debugcapidentify"/> @endxmlonly
- * @brief Identify the type of a capability in the current cspace.
+ * @brief Identify the type of a capability in the current CSpace.
  *
  * This debugging system call returns the type of capability in a capability
- * slot in the current cspace. The type returned is not a libsel4 type, but
+ * slot in the current CSpace. The type returned is not a libsel4 type, but
  * refers to an internal seL4 type. This can be looked up in a built kernel by
  * looking for the (generated) `enum cap_tag`, type `cap_tag_t`.
  *
- * @param cap A capability slot in the current cspace.
+ * @param cap A capability slot in the current CSpace.
  * @return The type of capability passed in.
  *
  */
--- a/manual/manual.tex
+++ b/manual/manual.tex
@@ -13,6 +13,11 @@
 
 \usepackage{url}
 
+% Glossary
+\usepackage[acronym,toc]{glossaries}
+\input{parts/glossary}
+\makeglossaries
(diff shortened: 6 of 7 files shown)


bbrcknl:
@Indanz Can you pls check over the first draft of the glossary for the seL4 manual?

libsel4/arch_include/riscv/interfaces/sel4arch.xml
```diff
@@ -100,12 +100,12 @@
             </description>
             <error name="seL4_IllegalOperation">
                 <description>
-                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                    The <texttt text="_service"/> is a CPTR to a capability of the wrong type.
```
lsf37, line 103: I think these should stay CPtr -- with CPTR I'd expect an acronym. (I might have missed a discussion on this while I was away, apologies if I'm late to to the party)

bbrcknl: Fair point. I made the (wrong!) choice while trying to fix:

"3: The manual uses CPTR in some places and CPtr in other. We should be consistent. (Assuming the outcome of this issue is updating docs)."
from https://github.com/seL4/seL4/issues/566

I'll change the CPTR to CPtr :-)

manual/Makefile
```diff
@@ -309,3 +309,11 @@ fixmes:
...
 		echo; \
 	done
+
+##############################################################################
+# Instructions to manually generate glossary
+	@echo "FIXME: Add script to generate Glossary to Makefile"
+	@echo "====> To generate glossary, run:"
+	@echo "              makeindex -s manual.ist -o manual.gls manual.glo"
+	@echo "              make
+	@echo "              touch latex.tex"
```
Indanz, line 319: This should be part of the Makefile and shouldn't need to be done manually. That way it can't be forgotten either.

Is there any new dependency introduced by glossaries? If so, they should be added to the README.md. I noticed when trying to build the manual that the dependency on `roboto.sty` is missing, haven't tried your branch yet.

I'll do the Makefile change if you want, but may need your help. It seems make already needs to run twice to get references rights, so all we need to do is adding the `makeindex -s manual.ist -o manual.gls manual.glo` somewhere.

Indanz (commented):
I'll wait for the update before reviewing the rest.

What are the criteria for adding an entry and where are the descriptions coming from?

E.g. do we want to limit it to seL4 specific things, or should it be broader, like hardware specific things? But then where do we draw the line? It's a seL4 manual, not a general computer architecture manual, so what baseline knowledge should we expect?

manual/Makefile
bbrcknl: @Indanz I mentioned making consistent capitalisation in the description of this PR, but should have explicitly pointed out that it was for CPtr. Should I do a separate PR for consistent naming of CPtr for clarity?

I should have also pointed out that the FIXME in the Makefile is a temporary instruction to generate manual.pdf. The aim was to include the glossary generation as part of the Makefile once I figured out how to do it, and then remove the temporary instructions. I'll fix it now.

manual/Makefile
```diff
@@ -271,6 +271,8 @@ references.bib: $(addsuffix .tex, $(Targets))
 		$(LaTeX) $$i.tex >>.log; \
 		cat $$i.aux | grep -e "\(citation\|bibdata\|bibstyle\)" | sed 's/bibdata{references}/bibdata{$(BibFiles)}/g' >> all_refs.aux; \
 	done
+	# make glossary
```
bbrcknl, line 274: @Indanz manual.ist is generating a pdf build error: https://github.com/seL4/seL4/actions/runs/7079820384/job/19266878821?pr=1143

It runs locally but not in the PR checks.

Can you pls help fix this?

lsf37:
> What are the criteria for adding an entry and where are the descriptions coming from?
> 
> E.g. do we want to limit it to seL4 specific things, or should it be broader, like hardware specific things? But then where do we draw the line? It's a seL4 manual, not a general computer architecture manual, so what baseline knowledge should we expect?

Good questions. I'd say, terms that have special meaning for/under seL4 should definitely be there. E.g. "thread" is a general OS term, but we should define what we mean by it in seL4 because there are so many variants of meaning out there (and the current "= CPU abstraction" is probably not enough for that purpose). ASID is also a general term, but seL4 ASIDs are not exactly the hardware ASIDs of the underlying platforms, so that should be defined.

Other than that, I'd only include terms if they are obscure and wouldn't expected to be known to a general OS audience. We should be able to assume general OS knowledge, maybe someone with a CS degree, but not specific microkernel knowledge or very detailed computer architecture knowledge. I.e. I don't think we should define things like "scheduler" or similar, but even if ASID didn't have a specific seL4 meaning, we should probably include it, because it's not something people would interact with unless they are implementing a kernel themselves. Of course, things that people would not interact with on the seL4 API level don't need to be in there.

The current set of terms and definitions is from a long time ago when we first published the abstract specification, and the target audience was a bit different there -- it was for people who know formal methods, but not so much the OS side, so it's possible that a few of these could be dropped.

manual/Makefile
lsf37: Looking at the Makefile, I'd put this command somewhere after "LaTeX first pass" in the `%.pdf` target, either directly before or after the `BibTex` run.

bbrcknl: Thanks @lsf37! That works, at least locally. But it needs to runs of `make` for the glossary to be included in the pdf.

manual/Makefile
Indanz: > I mentioned making consistent capitalisation in the description of this PR

Hmm, maybe something went wrong, I see no description at all for this PR, hence my surprise seeing other changes than the ones mentioned in the title.

Separate PR would be better in this case because both commits change many lines and are quite independent, and it would make it possible to merge one before the other, but it's not crucial, both are manual improvements and in that sense related.

manual/Makefile
Indanz: The others check the log to see if a rerun is necessary, if you can check `makeindex`'s ouput to see if a rerun is necessary, that would be ideal. Otherwise add a manual rerun to the same section.

Indanz: For me it seems to work with:

```diff
diff --git a/manual/Makefile b/manual/Makefile
index bc17ce8e1..0bce8e7bf 100644
--- a/manual/Makefile
+++ b/manual/Makefile
@@ -228,6 +228,7 @@ clean:
                *.out *.ps *-diff.tex *.mps .log *.pdf *.tgz *~ *.lof *.lot env.tex
        rm -rf ${DoxygenOutput} ${GeneratedLatexDir} ${GeneratedMarkdownDir}
        rm -rf ${Stage}
+       rm -rf *.acn *.glo *.gls *.ilg *.ist
 
 tar:   clean
        ( p=`pwd` && d=`basename "$$p"` && cd .. && \
@@ -271,8 +272,6 @@ references.bib: $(addsuffix .tex, $(Targets))
                $(LaTeX) $$i.tex >>.log; \
                cat $$i.aux | grep -e "\(citation\|bibdata\|bibstyle\)" | sed 's/bibdata{references}/bibdata{$(BibFiles)}/g' >> all_refs.aux; \
        done
-       # make glossary
-       makeindex -s manual.ist -o manual.gls manual.glo
        @echo "====> Removing duplicate bib entries";
        ${Q}cat all_refs.aux | uniq > tmp.aux;
        ${Q}diff references.aux tmp.aux > references.diff 2> /dev/null; \
@@ -293,6 +292,7 @@ references.bib: $(addsuffix .tex, $(Targets))
        ${Q}if ! test -e $*.bbl || test $(Bib) -nt $*.bbl; then rm -f $*.bbl; fi
        @echo "====> LaTeX first pass: $(<)"
        ${Q}$(LaTeX) $< >.log || if egrep -q $(Error) $*.log ; then cat .log; rm $@; false ; fi
+       makeindex -s manual.ist -o manual.gls manual.glo # glossary
        ${Q}if egrep -q $(Rerun_Bib) $*.log ; then echo "====> BibTex" && $(BibTeX) $* > /dev/null && echo "====> LaTeX BibTeX pass" && $(LaTeX) >.log $< ; fi
        ${Q}if egrep -q $(Rerun) $*.log ; then echo "====> LaTeX rerun" && $(LaTeX) >.log $<; fi
        ${Q}if egrep -q $(Rerun) $*.log ; then echo "====> LaTeX rerun" && $(LaTeX) >.log $<; fi
```
That said, for me the generated pdf looks awful (unrelated to the addition of the glossary, probably wrong `roboto.sty`) and I don't see any references to the glossary definitions, but I don't know if there were supposed to be any.

manual/Makefile
bbrcknl: @Indanz 

> I mentioned making consistent capitalisation in the description of this PR
> 
> Hmm, maybe something went wrong, I see no description at all for this PR, hence my surprise seeing other changes than the ones mentioned in the title.

Yes something went wrong - I put "consistent capitalisation" in the commit message, not the description. Sorry :-/

> Separate PR would be better in this case because both commits change many lines and are quite independent, and it would make it possible to merge one before the other, but it's not crucial, both are manual improvements and in that sense related.

Have created a separate PR: https://github.com/seL4/seL4/pull/1146

ratmice:
Probably not worth considering at this point, given that perhaps the makeindex things seem like they might be sorted in the Makefile now?  But perhaps worth mentioning that the latex acro package doesn't require these separate build steps.
It doesn't have a built in glossary being more suited to acronyms but, it doesn't appear difficult to define one. 
There appears to be a glossary template in this bug report even https://github.com/cgnieder/acro/issues/238

manual/Makefile
bbrcknl: When I remove
`makeindex -s manual.ist -o manual.gls manual.glo`
after
`done`
(~line 274), the Glossary disappears from the TOC.

Also, there are no references to glossary items within the text. I can create them, but it will take some time.

kent-mcleod:
> The purpose of ASIDs is avoid flushing caches on every context switch, nothing else. I don't think seL4 supports using more ASIDs than the underlaying hardware supports. That is, the current task limit is equal to the hardware's ASID limit, seL4 doesn't do any full cache flushes and dynamic hw ASID assignment.

For architecture configurations that don't have a large number of hw ASIDs, seL4 does do dynamic hw ASID assignment. This is at least the case on aarch32 and aarch64 with hyp enabled.

Indanz:
> > The purpose of ASIDs is avoid flushing caches on every context switch, nothing else. I don't think seL4 supports using more ASIDs than the underlaying hardware supports. That is, the current task limit is equal to the hardware's ASID limit, seL4 doesn't do any full cache flushes and dynamic hw ASID assignment.
> 
> For architecture configurations that don't have a large number of hw ASIDs, seL4 does do dynamic hw ASID assignment. This is at least the case on aarch32 and aarch64 with hyp enabled.

You are right, I missed that code (I looked at fast path code and misinterpreted the ASID code in there). I understand the need for it on 32-bit ARM, but aarch64 seems unnecessary? ASIDs are still 16-bits there, but for some reason seL4 choses to use VMIDs as ASIDs there?

kent-mcleod:
> You are right, I missed that code (I looked at fast path code and misinterpreted the ASID code in there). I understand the need for it on 32-bit ARM, but aarch64 seems unnecessary? ASIDs are still 16-bits there, but for some reason seL4 choses to use VMIDs as ASIDs there?

(I think we discussed this in another thread already, but just briefly replying again here too.)

When aarch64 is running in EL2 as a hypervisor, it only provides stage 2 address space translation from IPA to PA and the TLB tags for these entries use VMIDs which on the currently supported platforms only have 8bits.

Indanz:
> (I think we discussed this in another thread already, but just briefly replying again here too.)

Yes, we did [here](https://github.com/seL4/seL4/pull/1156#issuecomment-1882376212).

> When aarch64 is running in EL2 as a hypervisor, it only provides stage 2 address space translation from IPA to PA and the TLB tags for these entries use VMIDs which on the currently supported platforms only have 8bits.

I suspect this is the reason I saw a huge performance drop when going above 256 tasks. Even if tasks are mostly idle, the occasional wakeups flush the L2 cache, degrading system performance badly. I didn't look in detail though, it could have been general cache thrashing too.

To use ASIDs for EL0 tasks too, two-stage translation need to be used. That seems to have about 3% [overhead](https://genode.org/documentation/articles/arm_virtualization), which could be much better than now. However, the seL4 changes for that are quite big and no one else runs that many tasks.

Anyway, describing the purpose of the seL4 ASID objects shouldn't be the glossary's goal, however, it should give a short summary that ASID can refer to either HW Adress-Space Identifiers or seL4's object to help managing the hardware ASIDs.

lsf37:
There still seem to be some capitalisation changes in the same commit that adds the glossary and an accidentally committed file `q` in there. @bbrcknl are you happy for me to rebase and rearrange the commits a bit?

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
+%
+% Copyright 2014, General Dynamics C4 Systems
+%
+% SPDX-License-Identifier: GPL-2.0-only
+%
+
+\newglossaryentry{asid}{
+    name=ASID,
+    description={Address Space Identifier. ASIDs are associated with page directories (PDs), and define the virtual address space of a thread. Multiple threads may be in the same address space}
```
lsf37, line 9:
```suggestion
    description={Address Space Identifier. Depending on architecture, the kernel provides software ASIDs, which are associated with VSpace root objects, and define the virtual address space of a thread. They are mapped to hardware ASIDs on demand when the architecture supports these. Multiple threads may be in the same address space}
```

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
...
+}
+
+\newglossaryentry{badge}{
+    name=Badge,
+    description={A badge is a piece of extra information stored in a capability. It can be used by applications to identify caps previously handed out to clients}
```
lsf37, line 14:
```suggestion
    description={A badge is a piece of extra information stored in a capability, mostly used in endpoint and notification capabilities. It can be used by applications to identify caps previously handed out to clients}
```

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
...
+}
+
+\newglossaryentry{cap}{
+    name=Capability,
+    description={The main access control concept in seL4. A capability conceptually is a reference to a kernel object together with a set of access rights. Most seL4 capabilities store additional bits of information. Some of this additional information may be exposed to the user, but the bulk of it is kernel-internal book-keeping information. Capabilities are stored in CNodes and TCBs}
+}
+
+\newglossaryentry{cdt}{
+    name=CDT,
+    description={Capability Derivation Tree. A kernel-internal data structure that tracks the child/parent relationship between capabilities. Capabilities to new objects are children of the Untyped capability the object was created from. Capabilities can also be copied; in this case the user may specify if the operation should produce children or siblings of the source capability. The revoke operation will delete all children of the invoked capability}
```
lsf37, line 24:
```suggestion
    description={Capability Derivation Tree. A kernel-internal data structure that tracks the child/parent relationship between capabilities. Capabilities to new objects are children of the Untyped capability the object was created from. Capabilities can also be copied and result in either child or sibling capabilities, depending on the operation that was used and the depth of the existing derivation tree. The revoke operation will delete all children of the invoked capability}
```

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
...
+}
+
+\newglossaryentry{cptr}{
+    name=CPtr,
+    description={Capability Pointer. A user-level reference to a capability, relative to a specified root CNode or the thread’s CSpace root}
+}
+
+\newglossaryentry{cspace}{
+    name=CSpace,
+    description={A directed graph of CNodes. The CSpace of a thread defines the set of capabilities it has access to. The root of the graph is the CNode capability in the CSpace slot of the thread. The edges of the graph are the CNode capabilities residing in the CNodes .spanned by this root}
```
lsf37, line 39:
```suggestion
    description={A directed graph of CNodes. The CSpace of a thread defines the set of capabilities it has access to. The root of the graph is the CNode capability in the CSpace slot of the thread. The edges of the graph are the CNode capabilities residing in the CNodes spanned by this root}
```

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
...
+
+\newglossaryentry{notificationobject}{
+    name=Notification Object,
+    description={A word-size array of flags that provides a non-blocking signalling mechanism similar to a binary semaphore. Operations are signalling a subset of flags in a single operation, polling to check any flags, and blocking until any are signalled. Notification capabilities can be signal-only or wait-only}
+}
+
+\newglossaryentry{replyobject}{
+    name=Reply Object,
+    description={A reply object is a vessel for tracking reply messages, used to send a reply message and
+    wake up the caller}
```
lsf37, line 90:
```suggestion
    name=Reply Object,
    description={(MCS only) A reply object is a vessel for tracking reply messages, used to send a reply message and
    wake up the caller}
```

manual/parts/glossary.tex
```diff
@@ -0,0 +1,116 @@
...
+}
+
+\newglossaryentry{tcb}{
+    name=TCB,
+    description={Thread Control Block. The kernel object that stores management data for threads, such as the thread's CSpace, VSpace, thread state, or user registers}
+}
+
+\newglossaryentry{untypedmemory}{
+    name=Untyped Memory,
+    description={The foundation of memory allocation in the seL4 kernel. See \autoref{sec:kernmemalloc}}
```
lsf37, line 105:
```suggestion
    name=Untyped Memory,
    description={Memory that can be used to create kernel objects or user-accessible frames. The foundation of memory allocation in the seL4 kernel. See \autoref{sec:kernmemalloc}}
```

manual/parts/glossary.tex
Indanz: "with" or "for" seems more appropriate than "in". Otherwise a good improvement. Any other use cases for badges are truly obscure and I can't think of anything useful, other than perhaps limiting further copying or revocation.

lsf37: Yup, happy with "for"

manual/parts/glossary.tex
Indanz: Strictly speaking it can only be used to create kernel objects, except if Page objects are not kernel objects, in which case that should be mentioned explicitly (instead of frame). I would go for:

"The foundation of kernel object creation and memory allocation in the seL4 kernel."

Indanz:
> the glossary doesn't have a chapter number, but the bibliography coming after the glossary does, which looks weird in the TOC. If someone knows how to give it a chapter number, we should do that. Otherwise I'd propose to swap the order so that the glossary without number comes last.

I would remove the chapter number from the bibliography instead.

Unrelated, but somewhat related (and I don't know where else to mention it): The page numbering restarts at chapter 1 for me. So it goes from List of Figures on page 10, to Introduction on page 1.

manual/parts/glossary.tex
lsf37: It's correct that frames are also kernel objects, the difference I was trying to point out is that these are directly writeable by user space, and the rest are not. Happy to tweak that. What I'm not happy with in the "foundation" sentence before was that it doesn't say what you can do with these objects. That untyped memory is the foundation of something doesn't help the reader to understand where to go with that. Maybe we should just point at the retype operation. (I'm happy mentioning the foundation bit, it should just not be the only thing).

lsf37:
> > the glossary doesn't have a chapter number, but the bibliography coming after the glossary does, which looks weird in the TOC. If someone knows how to give it a chapter number, we should do that. Otherwise I'd propose to swap the order so that the glossary without number comes last.
> 
> I would remove the chapter number from the bibliography instead. That's achieved by removing `numbib` from line 53.

I'd be fine with that as well, no specific preference.

> Unrelated, but somewhat related (and I don't know where else to mention it): The page numbering restarts at chapter 1 for me. So it goes from List of Figures on page 10, to Introduction on page 1.

That's intended for that style. It basically has a front matter with its own page numbering and then the main text. We can think about if that's really what we want, but it's at least not a bug or oversight :-). Should do that in a separate issue or PR, though. (There is something to be said for page numbers matching up with PDF pages, but there are also arguments for the traditional typesetting way of doing things).

manual/parts/glossary.tex
lsf37: You're right, the whole user-writeable thing is leading in the wrong direction. Happy to remove that.

bbrcknl: @lsf37 @Indanz I'm getting a bit bogged down in details :-)
What wording would you like to use for the Description of Untyped Memory?

lsf37: I'd go for

```suggestion
    name=Untyped Memory,
    description={Memory that can be used to create kernel objects via the \apifunc{seL4\_Untyped\_Retype}{untyped_retype} invocation. It is the foundation of memory allocation in the seL4 kernel. See \autoref{sec:kernmemalloc}}
```

manual/parts/glossary.tex
bbrcknl: I made the change using:
description={A badge is a piece of extra information stored in a capability, mostly used for endpoint and notification capabilities. It can be used by applications to identify caps previously handed out to clients}

mbrcknl:
Some [hardware tests failed](https://github.com/seL4/seL4/actions/runs/8353597602) after merging, but all the failures appear to be due to a temporary loss of network connectivity. There were plenty of hardware tests that passed, so I think this is fine.

PR #1145, opened 2023-12-01
Arm 64-bit: do not use unsigned int for arg length

Xaphiosis:
Arch_decodeInvocation takes a word_t length and then passes it to functions that take an unsigned int length. This was OK on 32-bit where these types are the same, but on 64-bit this is a downcast without a range check. It isn't clear why this doesn't trip a compiler warning.

### DIFF
--- a/include/arch/arm/arch/machine/gic_v2.h
+++ b/include/arch/arm/arch/machine/gic_v2.h
@@ -216,7 +216,7 @@ struct gich_vcpu_ctrl_map {
 };
 
 extern volatile struct gich_vcpu_ctrl_map *gic_vcpu_ctrl;
-extern unsigned int gic_vcpu_num_list_regs;
+extern word_t gic_vcpu_num_list_regs;
 
 static inline uint32_t get_gic_vcpu_ctrl_hcr(void)
 {
--- a/include/arch/arm/arch/machine/gic_v3.h
+++ b/include/arch/arm/arch/machine/gic_v3.h
@@ -320,7 +320,7 @@ static inline void ackInterrupt(irq_t irq)
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 
-extern unsigned int gic_vcpu_num_list_regs;
+extern word_t gic_vcpu_num_list_regs;
 
 static inline uint32_t get_gic_vcpu_ctrl_hcr(void)
 {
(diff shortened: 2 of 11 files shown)


Xaphiosis:
> > I'd really like to know why the compiler doesn't complain on an implicit downcast like this.
> 
> Automatic casting between int types is legal in C, even when downcasting. If you want warnings, you need the `-Wconversion` compiler option (works for both gcc and clang).

Thanks! I guess I had `-Wall` mentally internalised. You're right, and it's somewhat terrifying.

The proof failure is legit, I think there might be some accidental name shuffling that I'll need to look at next week.

Xaphiosis:
I had to add more word_t stuff to keep AArch64 proofs happy (we have to prove that those implicit downcasts are safe, and the easiest way is to not do them).
I have an ARM_HYP l4v version that's fixed, and will put up a PR for that shortly.
@Indanz : could you check again that I didn't do anything that would worry you? apologies for the noise

Xaphiosis:
> > > Automatic casting between int types is legal in C, even when downcasting. If you want warnings, you need the `-Wconversion` compiler option (works for both gcc and clang).
> > 
> > 
> > Thanks! I guess I had `-Wall` mentally internalised. You're right, and it's somewhat terrifying.
> 
> What's even more terrifying is the lack of enum checking, as enums are not proper types in C, they're just integers in most contexts, so no nice strong type checking by the compiler for them. There's recently `-Wenum-conversion` and `-Wenum-compare`, but even then the checks are fairly relaxed.

Did they address this as part of C23's typed enums? I really wish we had access to that feature, but unlikely we'll be able to use it for a long time. Can't believe it took this long.

Xaphiosis:
Corresponding l4v PR is up : https://github.com/seL4/l4v/pull/698

Indanz:
> Indanz : could you check again that I didn't do anything that would worry you? apologies for the noise

No problem. Looks good to me.

> Did they address this as part of C23's typed enums?

At first glance it does solve this problem by making enums proper types, but because of the wording I am not 100% sure. It does solve the problem that enums had undefined sizes which made them unsuitable for binary interoperability.

Just let me know when the proofs are updated and this can be merged.

Xaphiosis:
Proofs are good. I'll merge these now and bump the manifest.

PR #1147, opened 2023-12-08
github: upload kernel.elf build artifact

lsf37:
The kernel.elf file is occasionally more useful for debugging than the final board image.

This will need seL4/ci-actions#292 to be merged first.

### DIFF
--- a/.github/workflows/sel4test-hw.yml
+++ b/.github/workflows/sel4test-hw.yml
@@ -50,6 +50,11 @@ jobs:
       with:
         name: images-${{ matrix.march }}-${{ matrix.compiler }}
         path: '*-images.tar.gz'
+    - name: Upload kernel.elf files
+      uses: actions/upload-artifact@v3
+      with:
+        name: kernel.elf-${{ matrix.march }}-${{ matrix.compiler }}
+        path: '*-kernel.elf'
 
   the_matrix:
     name: Matrix


Indanz (approved):

Indanz:
So I'm trying to download the ELF file for the failed run of https://github.com/seL4/seL4/actions/runs/8921506649/attempts/1, but I can't find it anywhere. I can download images-armv8a-clang.zip, but it doesn't contain the kernel.elf file. I see a "cp kernel/kernel.elf ../ZYNQMP106_release_SMP_MCS_clang_64-kernel.elf" in the HW build logs, but no idea where I can actually down the file. I see a "Upload images", but not the expected "Upload kernel.elf files" that was added with this commit.

I noticed it before, but I thought it had to trickle through before this was actually used or something, but it's still not there.

Ivan-Velickovic:
> So I'm trying to download the ELF file for the failed run of https://github.com/seL4/seL4/actions/runs/8921506649/attempts/1, but I can't find it anywhere. I can download images-armv8a-clang.zip, but it doesn't contain the kernel.elf file. I see a "cp kernel/kernel.elf ../ZYNQMP106_release_SMP_MCS_clang_64-kernel.elf" in the HW build logs, but no idea where I can actually down the file. I see a "Upload images", but not the expected "Upload kernel.elf files" that was added with this commit.
> 
> I noticed it before, but I thought it had to trickle through before this was actually used or something, but it's still not there.

It appears here https://github.com/seL4/seL4/actions/runs/8919015329. I think you're looking at the 'seL4Test' set of actions instead of 'seL4Test-HW' which is what this PR changed. It is quite confusing though, I'm not sure what the difference between seL4Test and seL4Test-HW is since they both seem to produce hardware images?

Indanz:
Thanks @Ivan-Velickovic, that was it. I guess the only difference is that one also runs the HW tests, and that one didn't got updated to upload the elf file.

PR #1150, opened 2023-12-15
Replace deprecated pkg_resources usage

hcindyl:
Use importlib.metadata to check the jinja2 version


tools/invocation_header_gen.py
```diff
@@ -184,6 +177,15 @@ def generate(args, invocations):
 
 
 if __name__ == "__main__":
+    # We require jinja2 to be at least version 2.10,
+    # In the past we used the 'namespace' feature from that version.
+    # other versions of jinja, particularly `minijinja`, don't support
+    # namespaces. However in case `namespace` is needed in the future require a
+    # version which supports it.
+    jinja2_version = version("jinja2")
+    if jinja2_version < "2.10":
```
axel-h, line 186: Can't we do this check at the place where we did it before, so this will always fail regardless how this file is used? Otherwise, instead of putting this here, it might be better to put it at the place where we need this specific feature then.

hcindyl: Done. Move it to the function that actually use jinja2 methods

tools/invocation_header_gen.py
```diff
@@ -163,6 +156,14 @@ def parse_xml(xml_file):
...
 
 def generate(args, invocations):
+    # We require jinja2 to be at least version 2.10,
+    # In the past we used the 'namespace' feature from that version.
+    # other versions of jinja, particularly `minijinja`, don't support
+    # namespaces. However in case `namespace` is needed in the future require a
+    # version which supports it.
+    jinja2_version = version("jinja2")
+    if jinja2_version < "2.10":
+        sys.exit("Jinja2 should be >= 2.10")
```
axel-h, line 166: Let's play nice and throw just an exception instead of forcing a hard exit. That will allow callers to handle failures in case they care.

hcindyl: Done

lsf37:
> Use importlib.metadata to check the jinja2 version

Thanks for the contribution. Can you explain what is wrong with the previous version? I.e. why is the proposed one better?

hcindyl:
> > Use importlib.metadata to check the jinja2 version
> 
> Thanks for the contribution. Can you explain what is wrong with the previous version? I.e. why is the proposed one better?

Please see https://setuptools.pypa.io/en/latest/pkg_resources.html. At ToT you will see the DeprecationWarning when generating `syscall.h` and various `invocation.h` files.

Indanz:
It looks like importlib requires at least Python 3.8 and is part of the standard Python package, so doesn't introduce a new dependency.

axel-h (approved):
Looks ok to me. It avoids using deprecates methods, so this make it worth merging it.

However, the jinja python package v2.10 was released on 2017-Nov-08, and this version got picked up in Debian "buster" and Ubuntu 18.04. So, I wonder if we should just remove this check completely by now, as it seem obsolete. Also, the whole namespace thing was removed in commit 190e0046 and d7424240. The logic got moved into the python code, which seems a cleaner approach. So I don't see much gain in keeping the comment either. We can make removing all this a follow-up commit.

lsf37 (approved):
> Please see https://setuptools.pypa.io/en/latest/pkg_resources.html. At ToT you will see the DeprecationWarning when generating syscall.h and various invocation.h files.

Thanks, happy to avoid deprecated features.

I'm also fine with @axel-h's idea for follow-up.

Issue #622, opened 2021-10-20
Separate documentation for common error codes in kernel object API


canarysnort01:
_Background: This issue was identified during a review of #528_

Currently, all error codes that can be returned from kernel object invocations when using the libsel4-provided invocation stubs are individually documented on each kernel object method in the libsel4 XML IDL files.

However, there are some error conditions that exist for all kernel object methods. These are:

Error Code | Possible Cause
--- | ---
seL4_IllegalOperation | The _service is a CPtr to a capability of the wrong type.
seL4_InvalidCapability | The _service is a CPtr to a capability of the wrong type.

This results in duplication of these error messages in the XML IDL files and makes identifying method-specific error conditions more difficult to see in the rendered manual.

This issue would investigate removing or shortening these generic error codes in the IDL file and separating the generic error codes from the method-specific error codes in the table in the rendered manual.

Note that the generic error codes would still be displayed for each kernel object method in the manual, just in its own section. This is to support the scenario where a user has encountered a specific error code and is looking up the kernel object method in the manual to see what could have caused it.

Also note that any changes to the rendering will need to be done for both the LaTeX output to the manual and the markdown output for the documentation website.

Indanz:
I'd argue that `seL4_TruncatedMessage` should be treated as a common error code too.

Indanz:
> How would you suggest separating the generic and specific error codes in the manual and the markdown?

To be honest, I do not know what @canarysnort01 has in mind exactly.

The only way I can think of is using a `href` pointing to the generic description instead of copy & pasting the text. The generic description would need an `id` on its element to be able to refer it. That way you can have one unique text that is re-used in many places within the document, with the content copied to all places where it's referred in the generated manual. But it depends on the tools used whether they support XML with hrefs or not.

Alternatively, a user clickable link to the global description can be given for common errors, while still listing them per invocation. That way it's clear it's a common error, while it's still being documented per invocation, without too much clutter (assuming the ref is short enough). Could also make the error itself the link and give no description at all perhaps.

bbrcknl:
@Indanz I've attempted a fix, pls see https://github.com/seL4/seL4/pull/1151
But it's made the error codes long and clunky.
Any hints/suggestions would be helpful.

PR #1151, opened 2023-12-17
combine common api error codes

bbrcknl:
The seL4 API has three common error codes
 - `seL4_IllegalOperation`
 - `seL4_InvalidCapability`
 - `seL4_TruncatedMessage`
as described in https://github.com/seL4/seL4/issues/622

I've replaced the text in the descriptions in the xml files, E.g.
```
    <description>
        The <texttt text='_service'/> is a CPtr to a capability of the wrong type
    </description>
</error>
 ```
is now
```
    <description>
        &illegal_operation;
    </description>
</error>
 ```

There are a few problems with this approach.

**1. Global variable declaration is done at the top of each api xml file.** I'm too new to xml and python to place the global variable declaration into a single file (e.g. py code) such that it can be called or added to each api xml file.

**2.  The text strings that describe error messages are too long.** I aimed to include a useful error message + a link to a table with common error codes, e.g.

illegal_operation = "The `_service` is a CPtr to a capability of the wrong type (see `error:illegal_operation` in common error codes)" 
where `error:illegal_operation` links to a common error codes table

The long strings are particularly unreadable when there are multiple error codes 
E.g.
10.3.3 seL4_IRQControl
10.3.3.1 Get
`seL4_IllegalOperation`  On x86, an IOAPIC is being used. Or, the _service is a CPtr to a capability of the wrong type (see seL4_IllegalOperation in common error codes)

This could be made readable by listing the error codes
E.g. 
10.3.3 seL4_IRQControl
10.3.3.1 Get
`seL4_IllegalOperation`  
- On x86, an IOAPIC is being used. 
- Or, the _service is a CPtr to a capability of the wrong type (see seL4_IllegalOperation in common error codes)

But I'm struggling with the xml and python, can can't figure out how to include lists in the descriptions of errors.

@Indanz @lsf37 Any suggestions/hints?


bbrcknl:
@Indan Zupancic Thanks for the xinclude code
I've tried to use it to include a test fail, but have unfortunately failed.
Can you please take a look at the files I changed in https://github.com/seL4/seL4/pull/1151/commits/85c6504abcf3052d3d0cec6dd691f44264f2dcac to see what might be going wrong?

libsel4/arch_include/arm/interfaces/sel4arch.xml
```diff
@@ -39,7 +46,7 @@
             </error>
             <error name="seL4_IllegalOperation">
                 <description>
-                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                    <xi:include href="common_error_illegal_operation.xml" xpointer="xpointer(//commonerrorref)" />
                 </description>
             </error>
```
Indanz, line 51: If you put the whole error into the common error code, you can replace the whole error block instead of just the description line. This would save a lot of duplication in many places. If you keep the error codes in separate files, it would also avoid the need to use xpointer altogether for this common case.

Using xpointer you can still extract the description if necessary, when there are also non-common reasons for the same error.

If you prefer always using xpointer, I would combine all common errors into one file.

libsel4/include/interfaces/sel4.xml
```diff
@@ -372,20 +384,21 @@
                 description="Capability to the scheduling context that the TCB should run on. If the scheduling context is already bound to a notification or TCB that is not this TCB this operation will fail. Similarly, if this TCB is already bound to a scheduling context that is not this scheduling context, this will also fail."/>
             <param dir="in" name="fault_ep" type="seL4_CPtr"
                 description="CPtr to the endpoint which receives IPCs when this thread faults."/>
-            <error name="seL4_IllegalOperation">
+            <!-- <error name="seL4_IllegalOperation">
```
Indanz, line 387: This seems some stray test change.

libsel4/arch_include/arm/interfaces/sel4arch.xml
bbrcknl: I replaced the code blocks where possible. In cases where the common error code was in addition to other errors, I just put the line description, e.g.
```
   <description>
      The <texttt text="_service"/> is not mapped in a VSpace.
      Or, if hypervisor support is configured, the requested range overlaps the kernel physical address range.
      <xi:include href="common_error_illegal_operation_append.xml" />
   </description>
</error>
```

libsel4/arch_include/riscv/interfaces/sel4arch.xml
Indanz, line 1: Not fully updated yet, still has xincludes for whole errors in description and xpointer references.

libsel4/arch_include/x86/interfaces/sel4arch.xml
Indanz, line 1: Not fully updated yet, still has xincludes for whole errors in description and xpointer references.

libsel4/include/interfaces/sel4.xml
Indanz, line 1: Not fully updated yet, still has xincludes for whole errors in description and xpointer references.

libsel4/sel4_arch_include/x86_64/interfaces/sel4arch.xml
Indanz, line 1: Not fully updated yet, still has xincludes for whole errors in description and xpointer references.

manual/manual.tex
```diff
@@ -88,6 +88,13 @@
 % Read information about the repository.
 \input{env}
 
+\makeatletter
+\newcommand{\setword}[2]{%
+  \phantomsection
+  #1\def\@currentlabel{\unexpanded{#1}}\label{#2}%
+}
+\makeatother
+
```
Indanz, line 97: Perhaps add a comment what this does and why it's needed? I have no idea.

If possible, maybe move it slightly higher, above the `\input{env}` with the other `\newcommand`.

Indanz (commented):
I think we're almost there. You forgot to update a few files (missing git add?), but otherwise looks a lot cleaner now.

libsel4/arch_include/arm/interfaces/sel4arch.xml
Indanz: Having the common errors twice makes it possible to have different texts, so I'm fine with how you did it.

manual/manual.tex
bbrcknl: This block of instructions provides hyperrefs within the pdf. E.g. An error might say ```See seL4_IllegalOperation in common error codes.``` **seL4_IllegalOperation** links to **10.1.11 Common Error Codes**. Without the enveloping ```\makeatletter``` and ```\makeatother``` the line would read ```See 10.1.11 in common error codes.```

Indanz: So similar to `\apifunc`, but more generic.

manual/tools/parse_doxygen_xml.py
```diff
@@ -30,13 +30,16 @@ def get_parse_table(self):
...
             'obj': lambda p, r: "%s" % p['name'],
             'errorenumdesc': lambda p, r: "",
             'orderedlist': self.parse_ordered_list,
             'listitem': lambda p, r: self.parse_para(p.para, r),
             'itemizedlist': self.parse_itemized_list,
             'autoref': lambda p, r: "%s" % p['label'],
-            'docref': self.parse_recurse
+            'docref': self.parse_recurse,
+            'commonerrorref': self.parse_recurse,
+            'apiref': self.parse_recurse
```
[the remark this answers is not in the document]
Indanz: I get [manual.pdf](https://github.com/seL4/seL4/files/15126628/manual.pdf). I'm using a Debian 11 Bullseye based distro, I guess there is some version difference in some tool that explains this difference. I have Python 3.9.2. @lsf37 does it work for you? Any ideas?

PR #1152, opened 2023-12-25
cmake: propagate `CROSS_COMPILER_PREFIX`

nspin:
Mark `CROSS_COMPILER_PREFIX` as meaningful to `gcc.cmake`, so that it is propagated for all cases where `gcc.cmake` used.

`gcc.cmake` may be evaluated multiple times by CMake during configuration. Without this patch, `gcc.cmake` may be evaluated without `CROSS_COMPILER_PREFIX` set in some cases, even if `-DCROSS_COMPILER_PREFIX=<...>` was specified in the CMake command line.

[Documentation for CMAKE_TRY_COMPILE_PLATFORM_VARIABLES](https://cmake.org/cmake/help/latest/variable/CMAKE_TRY_COMPILE_PLATFORM_VARIABLES.html).

### DIFF
--- a/gcc.cmake
+++ b/gcc.cmake
@@ -145,6 +145,8 @@ set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
 
 set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
 
+set(CMAKE_TRY_COMPILE_PLATFORM_VARIABLES CROSS_COMPILER_PREFIX)
+
 mark_as_advanced(FORCE CMAKE_TOOLCHAIN_FILE)
 
 # Invoke compiler via ccache. This has no effect if ccache cannot be found.


nspin:
After a closer look, I realized that the bug I was running into only applies when `-DCMAKE_TOOLCHAIN_FILE=gcc.cmake` is explicitly set. Code in `seL4Config.cmake` makes explicitly setting `CMAKE_TOOLCHAIN_FILE` to one of the provided ones unnecessary:

https://github.com/seL4/seL4/blob/7bad3610f1adad0062f1babb4090dd1788ef2a03/configs/seL4Config.cmake#L235-L262

axel-h:
I'm curious, what error/issue did you run into that makes adding this necessary?

kent-mcleod:
> This code also bakes `CMAKE_TOOLCHAIN_FILE` into the toolchain file.

This approach was taken because the toolchain file was used in contexts where the variable context may not be present, so baking the config into the toolchain file allows the file to not depend on any variables in the cache. (https://github.com/seL4/seL4/pull/1152/files#diff-cb0e9a6aa431817ec09570ffd67f1141b9b76a5660cd19cdb3ab8a922458c360L18-L21)

That said, I wasn't aware of `CMAKE_TRY_COMPILE_PLATFORM_VARIABLES` when the original approach was chosen so maybe it would have been a better approach? I don't think it deals with the other contexts that gcc.cmake might be used though?

nspin:
> I'm curious, what error/issue did you run into that makes adding this necessary?

I was building on an aarch64 machine, targeting aarch64, without an x86_64-targeting toolchain installed.

This is the line that failed:

https://github.com/seL4/seL4/blob/7bad3610f1adad0062f1babb4090dd1788ef2a03/gcc.cmake#L129

nspin:
> This approach was taken because the toolchain file was used in contexts where the variable context may not be present, so baking the config into the toolchain file allows the file to not depend on any variables in the cache. (https://github.com/seL4/seL4/pull/1152/files#diff-cb0e9a6aa431817ec09570ffd67f1141b9b76a5660cd19cdb3ab8a922458c360L18-L21)

It seems like that basically solves the same problem, and is known to work in most cases. However, as far as I can tell, it is prone to failure in some development environments (such as mine) when `-DCMAKE_TOOLCHAIN_FILE=gcc.cmake` is explicitly provided. That said, from looking at the code, I think that providing `CMAKE_TOOLCHAIN_FILE` explicitly is never actually necessary. 

Some seL4 ecosystem documentation demonstrates explicitly providing `CMAKE_TOOLCHAIN_FILE`. Maybe the best way to solve the problem that is the subject of this PR would be removing `-DCMAKE_TOOLCHAIN_FILE` CMake flags from the documentation.

kent-mcleod:
> Some seL4 ecosystem documentation demonstrates explicitly providing `CMAKE_TOOLCHAIN_FILE`. Maybe the best way to solve the problem that is the subject of this PR would be removing `-DCMAKE_TOOLCHAIN_FILE` CMake flags from the documentation.

Yea, I think if you could make it work with passing in the -DCMAKE_TOOLCHAIN_FILE option then we could take another look at whether preprocessing the file ends up being necessary at all which would be a positive simplification in my opinion.

kent-mcleod (approved):

axel-h:
> I think that providing CMAKE_TOOLCHAIN_FILE explicitly is never actually necessary.

Isn't this the way to select between GCC or clang?

nspin:
> > I think that providing CMAKE_TOOLCHAIN_FILE explicitly is never actually necessary.
> 
> Isn't this the way to select between GCC or clang?

I’m not sure how robust this is considered, but the presence of ‘-DTRIPLE’ is used here [1] to determine the toolchain flavor.

[1] https://github.com/seL4/seL4/blob/73553606f1bc18547027b0ceaaa3dd90af96952d/configs/seL4Config.cmake#L235C1-L235C1

axel-h:
> > > I think that providing CMAKE_TOOLCHAIN_FILE explicitly is never actually necessary.
> > Isn't this the way to select between GCC or clang?
> I’m not sure how robust this is considered, but the presence of ‘-DTRIPLE’ is used here [1] to determine the toolchain flavor.
> [1] https://github.com/seL4/seL4/blob/73553606f1bc18547027b0ceaaa3dd90af96952d/configs/seL4Config.cmake#L235C1-L235C1

Interesting. But it has a specific use case according to the commit that put it in, see https://github.com/seL4/seL4/commit/7d9297326bc6527645700c4d8318b0c7a12ddb32

lsf37 (approved):

PR #1154, opened 2024-01-04
CI: build and test kernel for RISCV64 with LLVM

axel-h:
Changes
- CI: build standalone kernel for RISCV64 with LLVM
- CI: run sim and HW build with LLVM for RISCV

To be me merged one these docker containers have been updated, so the workflows use them
- https://hub.docker.com/r/sel4/sel4test-hw
- https://hub.docker.com/r/sel4/sel4test-sim

related CI-Actions PR
- https://github.com/seL4/ci-actions/pull/295
- https://github.com/seL4/ci-actions/pull/299

### DIFF
--- a/.github/workflows/sel4test-deploy.yml
+++ b/.github/workflows/sel4test-deploy.yml
@@ -58,11 +58,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        march: [armv7a, armv8a, nehalem]
+        # There is no "rv32imac" hardware yet.
+        march: [armv7a, armv8a, nehalem, rv64imac]
         compiler: [gcc, clang]
-        include:
-          - march: rv64imac
-            compiler: gcc
     steps:
     - name: Build
       uses: seL4/ci-actions/sel4test-hw@master
--- a/.github/workflows/sel4test-hw.yml
+++ b/.github/workflows/sel4test-hw.yml
@@ -33,11 +33,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        march: [armv7a, armv8a, nehalem]
+        # There is no "rv32imac" hardware yet.
+        march: [armv7a, armv8a, nehalem, rv64imac]
         compiler: [gcc, clang]
-        include:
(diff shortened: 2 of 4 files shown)


axel-h:
Seems to work, see test run at https://github.com/axel-h/seL4/actions/runs/7472296390?pr=137. The test run on this PR will not pick up the changes in the HW build, because CI is configured to use the workflow from the base branch due to security settings.

.github/workflows/sel4test-deploy.yml
```diff
@@ -58,11 +58,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        march: [armv7a, armv8a, nehalem]
+        march: [armv7a, armv8a, nehalem, rv32imac, rv64imac]
```
lsf37, line 61: We don't have any rv32imac hardware to run and build for, so we should probably leave that one out of the matrix to save on GitHub runners a bit. This reminds me that we should mark the `RISCVVIRT32` platform as `no_hw_build` in the ci-actions repo, because it's a simulation platform only (it is being built and tested in the simulation job).

.github/workflows/sel4test-hw.yml
```diff
@@ -33,11 +33,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        march: [armv7a, armv8a, nehalem]
+        march: [armv7a, armv8a, nehalem, rv32imac, rv64imac]
```
lsf37, line 36: Would also leave out `rv32imac` here, since we currently have no hardware for it.

lsf37 (approved):
This is looking good and makes things simpler -- if you remove the 32 bit runs for RISCV for now, it's ready to go.

.github/workflows/sel4test-deploy.yml
axel-h: `RISCVVIRT32` has `no_hw_build` set. I just removed this for a test in my branch to check RV32 works as expected.

axel-h: fixed

Issue #1155, opened 2024-01-08
is MAX_IRQ correct for tqma8xqp1gb?


lsf37:
The current value of `MAX_IRQ` for the `tqma8xqp1gb` [is 512](https://github.com/seL4/seL4/blob/ed613d7d490be91f7f164ff98e09d9098de64d6e/src/plat/tqma8xqp1gb/config.cmake#L23). Should it be 511?

The boot code expects to be able to set the irq state of `maxIRQ`, which in this case would mean that the IRQ CNode has to have 2^10 slots with 511 unused slots instead of 2^9.

We're also currently claiming that the `irq` type has 9 bits on aarch64. 512 doesn't fit that, i.e. we'd need to do something about that on the verification side. (All other current platforms fit in 9 bit)

lsf37:
@Indanz do you know more about the `tqma8xqp1gb` setup?

Indanz:
> The current value of `MAX_IRQ` for the `tqma8xqp1gb` [is 512](https://github.com/seL4/seL4/blob/ed613d7d490be91f7f164ff98e09d9098de64d6e/src/plat/tqma8xqp1gb/config.cmake#L23). Should it be 511?

Yes, it should. The highest defined interrupt in the datasheet is `543 - 32 = 511`. The highest non-reserved IRQ is 470.

PR #1156, opened 2024-01-08
aarch64 verification tweaks

lsf37:
- refactor/remove `getMapRefForASID` for verification convenience
- add retype annotations for AArch64
- constrain `irq_w` field width in `VPPIEvent` to maxIRQ bit width

The last of these (`irq_w` field width) might be a problem on `tqma8xqp1gb` where `maxIRQ` is 512, which means the irq bit width is 10 instead of 9 like on the other platforms. This looks like a mistake to me, though, it'd be strange if the platform had exactly 513 IRQs instead of 512 (in which case maxIRQ should be set to 511). See also #1155.

### DIFF
--- a/include/arch/arm/arch/64/mode/object/structures.bf
+++ b/include/arch/arm/arch/64/mode/object/structures.bf
@@ -203,7 +203,8 @@ block VCPUFault {
 }
 
 block VPPIEvent {
-    field irq_w     64
+    padding         55
+    field irq_w      9
     padding         32
     padding         28
     field seL4_FaultType  4
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1244,10 +1244,13 @@ static exception_t performPageGetAddress(pptr_t base_ptr, bool_t call)
 static exception_t performASIDControlInvocation(void *frame, cte_t *slot,
                                                 cte_t *parent, asid_t asid_base)
 {
+    /** AUXUPD: "(True, typ_region_bytes (ptr_val \<acute>frame) 12)" */
+    /** GHOSTUPD: "(True, gs_clear_region (ptr_val \<acute>frame) 12)" */
     cap_untyped_cap_ptr_set_capFreeIndex(&(parent->cap),
                                          MAX_FREE_INDEX(cap_untyped_cap_get_capBlockSize(parent->cap)));
 
     memzero(frame, BIT(seL4_ASIDPoolBits));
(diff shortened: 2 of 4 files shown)


Indanz (approved):
Code seems good to me. Just add the tqma IRQ fix and it's ready to go.

I don't understand why aarch64 is using VMID as ASID instead of VMId+ASID, doesn't that unnecessarily limit the ASID space, with all downsides of that?

lsf37:
> Code seems good to me. Just add the tqma IRQ fix and it's ready to go.

👍 now added.
 
> I don't understand why aarch64 is using VMID as ASID instead of VMId+ASID, doesn't that unnecessarily limit the ASID space, with all downsides of that?

I'm not sure, I was actually not aware that they could be combined. @kent-mcleod do you know more on that?

Non-hyp mode has a large enough hardware ASID space that we don't need any seL4 software ASID translation, which is also a bit easier on verification. But now that we've almost finished the proofs for all that, I'm not that keen to change it, at least not before this set of proofs is finished (which should be by end of March).

kent-mcleod:
> I don't understand why aarch64 is using VMID as ASID instead of VMId+ASID, doesn't that unnecessarily limit the ASID space, with all downsides of that?

Can you be more specific? When seL4 is running in EL2, it only manages stage 2 mmu translation tables, the TLB entries for these mappings are tagged with the VMID.

Guest VCPUs are able to program stage 1 mmu and can then manage their own ASID allocations.

However native seL4 applications aren't able to program stage 1 mmu and so the logical seL4 VSpace is implemented only by stage 2 mmu translations. So only VMIDs are used and ASIDs are not used. Is this what you are referring to?

Indanz:
> > I don't understand why aarch64 is using VMID as ASID instead of VMId+ASID, doesn't that unnecessarily limit the ASID space, with all downsides of that?
> 
> Can you be more specific? When seL4 is running in EL2, it only manages stage 2 mmu translation tables, the TLB entries for these mappings are tagged with the VMID.
> 
> Guest VCPUs are able to program stage 1 mmu and can then manage their own ASID allocations.
> 
> However native seL4 applications aren't able to program stage 1 mmu and so the logical seL4 VSpace is implemented only by stage 2 mmu translations. So only VMIDs are used and ASIDs are not used. Is this what you are referring to?

Yes, I assumed that ASIDs could be used for stage 2 translations too, and that the TLB would always check VMID+ASID. But it seems that only VMID is checked at stage 2 and that the ASID check is only done for stage 1 translations.

lsf37:
Ok, so the VMID part is fine as is for now. We could maybe do more for v8.1, but that's a longer story.

I have added one more commit with more verification annotations (no code change). From my side, this is ready to merge once the seL4/l4v#703 is ready.

PR #1162, opened 2024-01-16
Hardware Debug API: Single stepping on ARMv7

alwin-joshy:
The implementation of single-stepping on ARMv7 relies on instruction breakpoint mismatches. In the implementation, it sets one of the breakpoint value registers (BVR) to the virtual address NULL and configures the corresponding breakpoint control register to generate a debug exception whenever an instruction that does not match this address is executed. I believe this is incorrect because it means that when single-stepping is enabled, no progress will ever be made and every instruction will mismatch with NULL and result in a debug exception being generated.

I believe the correct behavior is to instead set the debug value register to be the current LR. Then, what will happen instead is that once the thread is permitted to run again, the instruction that is returned to will be executed, and the PC will then advance, resulting in a debug exception being generated before another instruction can be executed.

This PR changes the logic in the initial configuring of single stepping to set the BVR to the current LR. It also changes the handling of a single-stepping debug exception to update the BVR register to the new LR so that it works when single-stepping is configured to step over several instructions before sending a debug fault to the appropriate fault handler thread.

Test with: https://github.com/seL4/sel4test/pull/111

### DIFF
--- a/src/arch/arm/machine/debug.c
+++ b/src/arch/arm/machine/debug.c
@@ -671,15 +671,13 @@ bool_t configureSingleStepping(tcb_t *t,
         bp_num = convertBpNumToArch(bp_num);
     }
 
-    /* On ARM single-stepping is emulated using breakpoint mismatches. So you
-     * would basically set the breakpoint to mismatch everything, and this will
-     * cause an exception to be triggered on every instruction.
-     *
-     * We use NULL as the mismatch address since no code should be trying to
-     * execute NULL, so it's a perfect address to use as the mismatch
-     * criterion. An alternative might be to use an address in the kernel's
-     * high vaddrspace, since that's an address that it's impossible for
-     * userspace to be executing at.
+    /* On ARM single-stepping is emulated using breakpoint mismatches. The aim
+     * of single stepping is to execute a single instruction. By setting an
+     * instruction mismatch breakpoint to the current LR of the target thread,
+     * the thread will be able to execute this instruction, but attempting to
+     * execute any other instruction will result in the generation of a debug
+     * exception that will be delivered to the kernel, allowing us to simulate
+     * single stepping.
      */
     dbg_bcr_t bcr;
 
@@ -701,7 +699,7 @@ bool_t configureSingleStepping(tcb_t *t,
     bcr = dbg_bcr_set_byteAddressSelect(bcr, convertSizeToArch(1));
     bcr = Arch_setupBcr(bcr, false);
 
-    writeBvrContext(t, bp_num, 0);
+    writeBvrContext(t, bp_num, t->tcbArch.tcbContext.registers[FaultIP]);
     writeBcrContext(t, bp_num, bcr.words[0]);
 
     t->tcbArch.tcbContext.breakpointState.n_instructions = n_instr;
@@ -1072,7 +1070,7 @@ seL4_Fault_t handleUserLevelDebugException(word_t fault_vaddr)
 #endif
 
     word_t method_of_entry = getMethodOfEntry();
-    int i, active_bp;
+    int active_bp;
     seL4_Fault_t ret;
     word_t bp_reason, bp_vaddr;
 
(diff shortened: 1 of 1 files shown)


Indanz (commented):
A few comments:
- Comment lines are too long, please limit to 80 characters.
- Please use `curr_thread` everywhere instead of sometimes `NODE_STATE(ksCurThread)`.
- Please add a local variable for the `curr_thread->tcbArch.tcbContext` pointer and use that.
- Remove the confusing `n_instr == 0` part of the comment, it's not relevant here.

Hard to believe it never worked and was never tested, it was introduced in 2016 by commit bebfcf6d2755d956d7c4dad499dde1527afd89c5.

Indanz:
Do we get two traps if single-stepping over a breakpoint? If not, just checking whether single-stepping is enabled could be sufficient.

alwin-joshy:
> A few comments:
> 
>     * Comment lines are too long, please limit to 80 characters.
> 
>     * Please use `curr_thread` everywhere instead of sometimes `NODE_STATE(ksCurThread)`.
> 
>     * Please add a local variable for the `curr_thread->tcbArch.tcbContext` pointer and use that.
> 
>     * Remove the confusing `n_instr == 0` part of the comment, it's not relevant here.
> 
> 
> Hard to believe it never worked and was never tested, it was introduced in 2016 by commit [bebfcf6](https://github.com/seL4/seL4/commit/bebfcf6d2755d956d7c4dad499dde1527afd89c5).

Agreed, it would be quite strange if it didn't work, but I tried it as is with the test I added in https://github.com/seL4/sel4test/pull/111 and it didn't work. It's possible that I'm using the API incorrectly but I don't really see how else it would be used and there were no existing single-stepping tests for ARM.

Indanz:
> With this being the case, I don't understand how setting DBGBVR to NULL would let you simulate single stepping, as it should result in a breakpoint exception whenever the code being executed is not at vaddr NULL.

The ARM manuals says:

> Breakpoint debug events are synchronous. That is, the debug event acts like an exception that cancels the breakpointed instruction.

So I assume the code was tested to see if it triggers breakpoint events, but never actually used by anyone. I guess no one made a remote gdb debugger for seL4 yet.

alwin-joshy:
> So I assume the code was tested to see if it triggers breakpoint events, but never actually used by anyone. I guess no one made a remote gdb debugger for seL4 yet.

That's what I'm working on at the moment, which is the reason for this sudden noise about the hardware debug api :p

Indanz:
> > I guess no one made a remote gdb debugger for seL4 yet.
> 
> That's what I'm working on at the moment, which is the reason for this sudden noise about the hardware debug api :p

Cool. Could you make the IO part modular and write it as a library so it's easily integrated into existing systems? Then UART, IP, USB or even GPIO could be used for debugging. (Assuming you're releasing the code of course, if not, carry on.)

alwin-joshy:
> > > I guess no one made a remote gdb debugger for seL4 yet.
> > 
> > 
> > That's what I'm working on at the moment, which is the reason for this sudden noise about the hardware debug api :p
> 
> Cool. Could you make the IO part modular and write it as a library so it's easily integrated into existing systems? Then UART, IP, USB or even GPIO could be used for debugging. (Assuming you're releasing the code of course, if not, carry on.)

The intention is to have the stub as a microkit component that can be hooked up with the sDDF for whatever interfaces you need.

Indanz (approved):

alwin-joshy:
> @alwin-joshy Can you please update the commit messages to give more context what is going on? (as in the PR description) E.g. in `fix single stepping ..` explain what is being fixed, what was wrong before and it is the difference now, in `Improve ..` describe what is being improved and why, in `Update comment`, describe why (what was wrong with it).

They all just build on the first commit so I think I'll just squash them. Was just waiting till Indan was happy with the state of the PR.

Indanz:
It looks good and passes the test, so I'm happy with it.

PR #1163, opened 2024-01-16
Zero capPtr for device untyped capabilities

heshamelmatary:
If the untyped cap is a device memory, the kernel has no mapping for it and hence pptr does not make sense and is not used. Zeroing pptr here makes it clear that this untyped is not used/mapped, and avoids an assertion error in the generated cap_untyped_cap_new where it checks if the pptr is canonical virtual address regardless if it is device untyped or not, which may fail on systems that have 48-bit physical address space or larger.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -692,6 +692,19 @@ BOOT_CODE static bool_t provide_untyped_cap(
             .isDevice = device_memory,
             .padding  = {0}
         };
+
+        /* If the untyped cap is a device memory, the kernel has no mapping for it
+         * and hence pptr does not make sense and is not used. Zeroing pptr here
+         * makes it clear that this untyped is not used/mapped, and avoids an
+         * assertion error in the generated cap_untyped_cap_new where it checks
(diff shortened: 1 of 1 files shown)


src/kernel/boot.c
```diff
@@ -692,6 +692,19 @@ BOOT_CODE static bool_t provide_untyped_cap(
             .isDevice = device_memory,
             .padding  = {0}
         };
+
+        /* If the untyped cap is a device memory, the kernel has no mapping for it
```
axel-h, line 696:
```suggestion
        /* If the untyped cap is for device memory, the kernel has no mapping for it
```

lsf37:
The capPtrs of device untypeds are used.

heshamelmatary:
> The capPtrs of device untypeds are used in a number of places (e.g. sameRegionAs, untyped reset, probably a few more), they are just not dereferenced. They can definitely not be just zero.

Ah thanks for pointing that out. I thought I got away with it since sel4tests and sel4bench were happy with it. It was a bit counterintuitive having a virtual address that isn't (and won't be) mapped.

PR #1164, opened 2024-01-17
Added maaxboard support for microkit

josh-felm:


axel-h:
This targets the branch `microkit`, which is a lot behind `master` where this board already exists. Not sure how to handle this now, ideally `microkit` would be on top of master?

lsf37:
@josh-felm can you please add a git [sign-off line](https://docs.sel4.systems/processes/contributing.html) to the commits so we can accept them?

lsf37:
@Ivan-Velickovic it does look like we should at least rebase the `micorkit` branch if it is still necessary (I forgot where we are with that).

Ivan-Velickovic:
Sorry, perhaps I should have been clearer in https://github.com/seL4/microkit/pull/38. We don't want to back-port stuff to this branch and instead I'll update the `microkit` branch so that the maaxboard support is there.

Once I've done that we can merge https://github.com/seL4/microkit/pull/38.

PR #1168, opened 2024-01-24
aarch64/vspace: test for page type directly

lsf37:
Test the PTE directly for being of page type and avoid ptr access to ease verification in unmapPage.

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1003,6 +1003,7 @@ void unmapPage(vm_page_size_t page_size, asid_t asid, vptr_t vptr, pptr_t pptr)
 {
     findVSpaceForASID_ret_t find_ret;
     lookupPTSlot_ret_t  lu_ret;
+    pte_t pte;
 
     find_ret = findVSpaceForASID(asid);
     if (find_ret.status != EXCEPTION_NONE) {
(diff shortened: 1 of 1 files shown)


Xaphiosis (approved):
looks good to me

Indanz (approved):
Both the original `pte_ptr_get_valid()` and `pte_pte_table_ptr_get_present()` look at `pte` type. The original exception for a PT being present, but with less bits left than PAGE_BITS doesn't make any sense to me.

kent-mcleod (approved):

Indanz:
> > Both the original `pte_ptr_get_valid()` and `pte_pte_table_ptr_get_present()` look at `pte` type. The original exception for a PT being present, but with PAGE_BITS or less bits left doesn't make any sense to me.
> 
> I agree it's weird. I wondered why it ended up like this, and it appears to be because that's what was established for RISC-V: https://github.com/seL4/seL4/blob/master/src/arch/riscv/kernel/vspace.c#L567-L570

The RISC-V check doesn't have the same weird bits left check. It looks like it may be introduced by mistake by commit 5b00dff65652822727eec48f7ac59382755e097c.

lsf37:
Yes, the PAGE_BITS check is new. It was harmless in the sense that it will alway be true if it is a page table PTE. What is from RISCV is the double negation, i.e. checking for invalid or page table instead of checking for a page.

Because AArch64 has two different types of page encodings (4k_page and page), we initially didn't have a function that just checked for that directly, but it turns out it's more convenient in verification to prove once that this function corresponds to a page pte than separately going through the two other cases each time. It's not a huge difference, but it's nicer and I think clearer in the code as well.

PR #1171, opened 2024-01-28
aarch64/vspace: type of counter in ASIDPoolAssign

lsf37:
As in the other architectures, avoid unnecessary casts for the loop counter by making it a `word_t`.

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1726,7 +1726,7 @@ exception_t decodeARMMMUInvocation(word_t invLabel, word_t length, cptr_t cptr,
         cap_t vspaceCap;
         cte_t *vspaceCapSlot;
         asid_pool_t *pool;
-        unsigned int i;
+        word_t i;
         asid_t asid;
 
         if (unlikely(invLabel != ARMASIDPoolAssign)) {


Indanz (changes requested):
It would make a lot more sense to make it a `asid_t` instead of `word_t`, now it relies on `asid_t` really being a `word_t`.

lsf37:
> It would make a lot more sense to make it a `asid_t` instead of `word_t`, now it relies on `asid_t` really being a `word_t`.

I disagree. `i` is not an `asid_t`, it is only the lower bits of an `asid_t` -- in the specs we do use different types for that (`asid_low` for pools, `asid_high` for the ASID table, and `asid` for the entire ASID), in C we can't do that, but I would find it misleading to type `i` as `asid_t`. It would mean that it could be usefully stored wherever an `asid_t` goes, but it can't. It's only an index into the ASID pool. It needs to be a type that is compatible with `+` or `|` on `asid_t`, but it's not an `asid_t` itself.

Indanz:
> > It would make a lot more sense to make it a `asid_t` instead of `word_t`, now it relies on `asid_t` really being a `word_t`.
> 
> I disagree. `i` is not an `asid_t`, it is only the lower bits of an `asid_t` -- in the specs we do use different types for that (`asid_low` for pools, `asid_high` for the ASID table, and `asid` for the entire ASID), in C we can't do that, but I would find it misleading to type `i` as `asid_t`. It would mean that it could be usefully stored wherever an `asid_t` goes, but it can't. It's only an index into the ASID pool. It needs to be a type that is compatible with `+` or `|` on `asid_t`, but it's not an `asid_t` itself.

I came to this conclusion too after thinking about it a bit more yesterday after posting my comment.

What is the problem with `unsigned int`? It should be closer to the true size of `asid_low` than `word_t`, how does changing it to that make things easier for verification?

Indanz (approved):

lsf37:
> I came to this conclusion too after thinking about it a bit more yesterday after posting my comment.

👍 
 
> What is the problem with `unsigned int`? It should be closer to the true size of `asid_low` than `word_t`, how does changing it to that make things easier for verification?

`unsigned int` is not directly `+` or `||` compatible with `word_t` -- the compiler (and C parser) automatically inserts casts for each operation, which we all have to prove don't change the value. This is not impossible or even hard to do, but it inserts a lot of noise into the proof and obscures the bits that are actually interesting.

It is true that `unsigned int` is closer to the actual `asid_low` value, but being closer doesn't really help. If we could declare a 9-bit word type in C, we'd definitely be using that here, but using 32 instead of 64 just means we have to convert between 9, 32, and 64 (9 from the spec) instead of just 9 and 64.

Indanz:
> It is true that `unsigned int` is closer to the actual `asid_low` value, but being closer doesn't really help. If we could declare a 9-bit word type in C, we'd definitely be using that here, but using 32 instead of 64 just means we have to convert between 9, 32, and 64 (9 from the spec) instead of just 9 and 64.

But that actually argues for making it `asid_t`, as it wants to have the same size and compatibility, and would make this intention clearer. Now if `asid_t` changes in size in the future this will cause problems with verification again.

lsf37:
> But that actually argues for making it `asid_t`, as it wants to have the same size and compatibility, and would make this intention clearer. 

But it'd also mislead. I guess the proper way of doing this would be to have new types `asid_low_t` and `asid_high_t` which are all defined to be `word_t` at the point where `asid_t` is defined as well. Then it's clear at the application site what is intended, and they can all be changed together when something changes. There'd be no checking from the compiler, of course, but at least intentions would be clear.

> Now if `asid_t` changes in size in the future this will cause problems with verification again.

If the size of `asid_t` changes, all kinds of things will break in the verification, so while the connection has to be there, it's probably one of the lesser worries when that type of change occurs.

PR #1178, opened 2024-01-30
CI: cancel older concurrent PR runs

axel-h:
This also removes the space in the workflow name to ensure there are no side effects when using it as an identifier.

This applies what has been done in https://github.com/seL4/camkes-vm-examples/pull/69

### DIFF
--- a/.github/workflows/sel4bench-pr.yml
+++ b/.github/workflows/sel4bench-pr.yml
@@ -4,7 +4,7 @@
 
 # Build and run sel4bench on pull requests, on label request
 
-name: seL4Bench PR
+name: seL4Bench-HW
 
 on:
   pull_request_target:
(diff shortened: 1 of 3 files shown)


axel-h:
For a test see https://github.com/axel-h/seL4/actions/runs/7711976407?pr=140 (workflow run 1532) that was cancelled by the newer https://github.com/axel-h/seL4/actions/runs/7711990454/job/21018499143?pr=140 (workflow run 1533)

axel-h:
I wonder, should we have this feature on `proof.yml` also to improve resource usage?

lsf37 (approved):
Thanks. That should speed up things a bit.

lsf37:
> I wonder, should we have this feature on `proof.yml` also to improve resource usage?

Yes, that would make sense and reduce the number of AWS instances that are need.

PR #1179, opened 2024-01-31
Fix benchmark log buffer for aarch64

Kswin01:
Fixing deprecated code left over from RFC-10.


src/arch/arm/64/kernel/vspace.c
```diff
@@ -1918,7 +1918,7 @@ exception_t benchmark_arch_map_logBuffer(word_t frame_cptr)
 
     ksUserLogBuffer = pptr_to_paddr((void *) frame_pptr);
 
-    *armKSGlobalLogPDE = pde_pde_large_new(
+    *armKSGlobalLogPDE = pte_pte_page_new(
```
lsf37, line 1921: Would suggest renaming this one to PTE as well (one more occurrence below)

```suggestion
    *armKSGlobalLogPTE = pte_pte_page_new(
```

src/arch/arm/64/model/statedata.c
```diff
@@ -92,9 +92,9 @@ pte_t armKSGlobalKernelPT[BIT(PT_INDEX_BITS)] ALIGN_BSS(BIT(seL4_PageTableBits))
 #ifdef CONFIG_KERNEL_LOG_BUFFER
 pte_t *armKSGlobalLogPDE = &armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][BIT(PT_INDEX_BITS) - 2];
```
lsf37, line 93:
```suggestion
pte_t *armKSGlobalLogPTE = &armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][BIT(PT_INDEX_BITS) - 2];
```

lsf37 (approved):
There might be more occurrences of `armKSGlobalLogPDE` to fix up, but happy with it otherwise.

Why did this compile so far? Shouldn't this be caught by `sel4bench` for armv8 platforms?

lsf37:
Please change the commit message to a short title (50 characters) and the explanation you have now to wrap inside the body at 72 characters.

Kswin01:
I suspect that the reason why sel4bench compiles is because the irq benchmark (which seems like the only test that users the kernel log buffer) is turned off by default.

lsf37:
> I suspect that the reason why sel4bench compiles is because the irq and vcpu benchmarks (which seems like the only benchmarks that use the kernel log buffer) are turned off by default.

Right, that would explain it. Any harm in turning those on by default? Is there enough config magic in place to skip them when the config doesn't support it?

src/arch/arm/64/model/statedata.c
```diff
@@ -92,9 +92,9 @@ pte_t armKSGlobalKernelPT[BIT(PT_INDEX_BITS)] ALIGN_BSS(BIT(seL4_PageTableBits))
...
 compile_assert(log_pude_is_correct_preallocated_pude,
-               GET_PUD_INDEX(KS_LOG_PPTR) == BIT(PT_INDEX_BITS) - 1);
+               GET_KPT_INDEX(KS_LOG_PPTR, 1) == BIT(PT_INDEX_BITS) - 1);
```
danshea00, line 95: Just wondering if this should be:
```suggestion
               GET_KPT_INDEX(KS_LOG_PPTR, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
```

As per: https://github.com/seL4/seL4/blob/91ec17c5bf29a4ef8b98b8f9769580b2442eb810/src/arch/arm/64/kernel/vspace.c#L237

Kswin01:
> > I suspect that the reason why sel4bench compiles is because the irq and vcpu benchmarks (which seems like the only benchmarks that use the kernel log buffer) are turned off by default.
> 
> Right, that would explain it. Any harm in turning those on by default? Is there enough config magic in place to skip them when the config doesn't support it?

I'm not too sure as I haven't dealt with sel4 bench much myself. In the CMakeLists for the irq app, there was a commit 5 years ago stating that it didn't build. I'm not sure if that's been fixed/revisited?

Ivan-Velickovic:
> Any harm in turning those on by default? Is there enough config magic in place to skip them when the config doesn't support it?

I don't think the IRQ benchmark needs anything extra, the VCPU benchmarks look like they're only meant for the TX2 platform. If I build the VCPU benchmark for a non-TX2 platform then I get a compile error. If I build for the TX2 platform, I also get a compile error, but this time a different one.

Ivan-Velickovic:
This file https://github.com/seL4/sel4bench/blob/master/apps/sel4bench/src/plat/tx2/vcpu.c is responsible for doing the processing of VCPU benchmark results but looking at it I don't see anything that is obviously platform specific. To fix this, maybe it is enough to just fixup the benchmark compile errors and make the processing generic for ARM64.

Ivan-Velickovic:
I think all the feedback has been addressed, can this be merged now?

lsf37 (approved):

lsf37:
Good to be merged from my side.

Indanz (approved):

include/arch/arm/arch/64/mode/model/statedata.h
```diff
@@ -39,7 +39,7 @@ extern hw_asid_t armKSNextASID VISIBLE;
 #endif
 
 #ifdef CONFIG_KERNEL_LOG_BUFFER
-pde_t *armKSGlobalLogPDE;
+pde_t *armKSGlobalLogPTE;
```
axel-h, line 42: Isn't this missing a `extern` keyword?

Indanz: Yes, and the other one is declared as:

https://github.com/seL4/seL4/blob/46f2b4d255bbf50e593103a859d3cf702a5232b6/src/arch/arm/64/model/statedata.c#L93

There's a `typedef pte_t pde_t;` in include/arch/arm/arch/64/mode/object/structures.h, but still.

PR #1180, opened 2024-02-01
AArch64: mark isFpuEnable as DONT_TRANSLATE

Xaphiosis:
Current verification model does not include lazy FPU switching, i.e. it acts as if this function always returns true, so no FPU faults could be produced. In order to guard against deriving a contradiction, we don't allow the C parser to translate it.


lsf37 (approved):

Indanz (approved):
Can this be merged or does it need to wait for something else first?

lsf37:
It's ready now, will merge.

Xaphiosis:
I was waiting to see whether this was a sufficient change for the proofs... as of a few minutes ago, I found out that it was, so merging it turned out to be OK.

Issue #1015, opened 2023-04-14
Allow VMPIDR_EL2 register to be set for VMs when CONFIG_MAX_NUM_NODES=1 


kent-mcleod:
Currently the seL4_VCPUReg enum only defines the seL4_VCPUReg_VMPIDR_EL2 register when the kernel is configured for SMP. This register should be available for unicore systems too as otherwise a user level VMM is not able to control the value of MPIDR_EL1 that the guest VM reads.

```
#if CONFIG_MAX_NUM_NODES > 1
    /* Virtualisation Multiprocessor ID Register */
    seL4_VCPUReg_VMPIDR_EL2,
#endif /* CONFIG_MAX_NUM_NODES > 1 */
```

axel-h:
Make sense to me, seem for ARMv7 and `seL4_VCPUReg_VMPIDR` it's a similar thing then. Could you make a PR?

lsf37:
No blockers from the verification side, should be fine to include.

PR #1181, opened 2024-02-01
ARM: Allow VMPIDR_EL2 reg to be set on unicore

kent-mcleod:
Currently the seL4_VCPUReg enum only defines the seL4_VCPUReg_VMPIDR_EL2 register when the kernel is configured for SMP. This register should be available for unicore systems too as otherwise a user level VMM is not able to control the value of MPIDR_EL1 that the guest VM reads.

Resolves https://github.com/seL4/seL4/issues/1015

### DIFF
--- a/include/arch/arm/arch/object/vcpu.h
+++ b/include/arch/arm/arch/object/vcpu.h
@@ -78,11 +78,13 @@ struct vcpu {
     word_t regs[seL4_VCPUReg_Num];
     bool_t vppi_masked[n_VPPIEventIRQ];
 #ifdef CONFIG_VTIMER_UPDATE_VOFFSET
-    /* vTimer is 8-bytes wide and has same alignment requirement.
-     * The struct will remain packed on 32-bit platforms when n_VPPIEventIRQ
-     * is odd, but were it to become even, an extra word of padding will be
-     * necessary.
-     * */
+    word_t vcpu_padding;
+    /* vTimer is 8-bytes wide and has the same 8-byte alignment requirement.
+     * If the sum of n_VPPIEventIRQ and seL4_VCPUReg_Num is even, we do not need
+     * extra padding. If the sum is odd we do. It currently is odd, so the extra
+     * padding above is necessary for the struct to remain packed on 32 bit
+     * platforms.
+     */
     struct vTimer virtTimer;
 #endif
 };
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -123,6 +123,7 @@
 #define REG_HCR_EL2         "hcr_el2"
 #define REG_VTCR_EL2        "vtcr_el2"
 #define REG_VMPIDR_EL2      "vmpidr_el2"
+#define REG_MPIDR_EL1       "mpidr_el1"
 #define REG_ID_AA64MMFR0_EL1 "id_aa64mmfr0_el1"
 
 /* for EL1 SCTLR */
@@ -481,10 +482,8 @@ static word_t vcpu_hw_read_reg(word_t reg_index)
         return readCNTVOFF_EL2();
     case seL4_VCPUReg_CNTKCTL_EL1:
         return readCNTKCTL_EL1();
-#ifdef ENABLE_SMP_SUPPORT
     case seL4_VCPUReg_VMPIDR_EL2:
         return readVMPIDR_EL2();
-#endif /* ENABLE_SMP_SUPPORT */
     default:
         fail("ARM/HYP: Invalid register index");
     }
@@ -564,11 +563,9 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
     case seL4_VCPUReg_CNTKCTL_EL1:
         writeCNTKCTL_EL1(reg);
         break;
-#ifdef ENABLE_SMP_SUPPORT
     case seL4_VCPUReg_VMPIDR_EL2:
         writeVMPIDR_EL2(reg);
         break;
-#endif /* ENABLE_SMP_SUPPORT */
     default:
         fail("ARM/HYP: Invalid register index");
     }
--- a/CHANGES
+++ b/CHANGES
@@ -83,6 +83,12 @@ Upcoming release: BREAKING
   optimization.
 * Added SMC Capability (smc_cap) and SMC forwarding for aarch64 platforms. See
   [RFC-9](https://sel4.atlassian.net/browse/RFC-9).
+* Arm+Hyp: Enabled access to seL4_VCPUReg_VMPIDR and seL4_VCPUReg_VMPIDR_EL2 for all hyp configurations.
+  Previously this register was only accessible for SMP kernel configurations. Non-SMP configurations can still require
+  access when wanting to control the value of MPIDR that the guest reads.
+  Note that the initial value for new seL4_ARM_VCPUs for this register is 0 which isn't a legal value for MPIDR_EL1 on
+  AArch64. It may be necessary for the register to be explicitly initialized by user level before launching a thread
+  associated with the new seL4_ARM_VCPU.
 
 ## Upgrade Notes
 
--- a/include/arch/arm/arch/32/mode/machine.h
+++ b/include/arch/arm/arch/32/mode/machine.h
@@ -59,6 +59,7 @@
 #define ID_PFR1    " p15, 0,  %0,  c0,  c1, 1" /* 32-bit RO CPU feature register */
 #define CPACR      " p15, 0,  %0,  c1,  c0, 2" /* 32-bit Architectural Feature Access Control Register */
 #define VMPIDR     " p15, 4,  %0,  c0,  c0, 5" /* 32-bit RW Virtualization Multiprocessor ID Register */
+#define MPIDR      " p15, 0,  %0, c0,  c0, 5"
 
 /* Use Hypervisor Physical timer */
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
--- a/include/arch/arm/armv/armv7-a/armv/vcpu.h
+++ b/include/arch/arm/armv/armv7-a/armv/vcpu.h
@@ -296,7 +296,6 @@ static inline void set_cntv_ctl(word_t val)
     MCR(CNTV_CTL, val);
 }
 
-#ifdef ENABLE_SMP_SUPPORT
 static inline word_t get_vmpidr(void)
 {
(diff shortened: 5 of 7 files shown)


lsf37 (commented):
Good from my side. It looks like this has verification impact on ARM_HYP and AARCH64, so we'll have to look  into that first. Should be relatively easy.

Indanz (commented):
I agree with giving access to unicore configuration too. However, I'm not convinced by:

> Each VCPU is initialized with the value of MPIDR of the core it is created on which should provide a legal value to the guest.

The chance that the VCPU object is created on the core it will run on seems low to me for multicore systems. Setting the MPIDR to `1 << 30` to indicate it's a unicore system with ID 0 seems like a better default. For a multicore guest the register needs to be correctly initialised by the VMM anyway.

Indanz:
> What would you also propose as the "obvious" initial value for aarch32? I'd prefer all 0s but that's not a legal value for aarch64. There's also saving the mpidr value of the boot core and using that. That seems most "policy-free" way which is what I was aiming for here.

I don't see a difference between aarch32 and aarch64 for register definitions. Both use bit 30 to indicate a unicore system, and for both 0 seems a valid value? I can only find a description of MPIDR in cpu core TRMs and not in general architecture documents though.

I don't know, but as far as I can tell, MPIDR is only used to find the location of GICv3 mapped registers, so it's something between the VMM and the guest, using 0 should be fine too in practise. If 0 is what guests currently get if the VMM doesn't change it, I would be inclined to keep it that way, for backward compatibility.

kent-mcleod:
> . If 0 is what guests currently get if the VMM doesn't change it, I would be inclined to keep it that way, for backward compatibility.

I think its reset value is architecturally undefined - which makes sense if software is supposed to configure it before enabling virtualization extensions.  Currently without this change the kernel doesn't modify the register, so it'd only be all 0 if that's what it was set to before the kernel was loaded.  Looking at the initial value on some of our platforms it's the same as the value of MPIDR but with an architecturally undefined reset value I wasn't able to add that setting it to the value of MPIDR is strictly backwards compatible. 

> Both use bit 30 to indicate a unicore system, and for both 0 seems a valid value?

Bit(31) is used to indicate the presence of multicore extensions and must be 1 for ARMv8. It can be 0 for ARMv7, but I believe all our ARMv7 platforms with -ve also have the multiprocessing extensions.

So initializing the register to BIT(31)  in all cases would probably be fine by me.

Indanz:
> > . If 0 is what guests currently get if the VMM doesn't change it, I would be inclined to keep it that way, for backward compatibility.
> 
> I think its reset value is architecturally undefined - which makes sense if software is supposed to configure it before enabling virtualization extensions. Currently without this change the kernel doesn't modify the register, so it'd only be all 0 if that's what it was set to before the kernel was loaded. Looking at the initial value on some of our platforms it's the same as the value of MPIDR but with an architecturally undefined reset value I wasn't able to add that setting it to the value of MPIDR is strictly backwards compatible.

What about just not touching it and keeping the existing behaviour? If there is no reason to change it, don't touch it, right?

> Bit(31) is used to indicate the presence of multicore extensions and must be 1 for ARMv8.

You're right, though it seems to be `RES1` now. No idea what the added value was compared to bit 30.

> It can be 0 for ARMv7, but I believe all our ARMv7 platforms with -ve also have the multiprocessing extensions.
> 
> So initializing the register to BIT(31) in all cases would probably be fine by me.

But does it make sense when the VM only has one core though? I can't imagine it hurts, but still.

kent-mcleod:
> What about just not touching it and keeping the existing behaviour? If there is no reason to change it, don't touch it, right?

The existing behavior involves not saving or restoring it so that all VCPUs scheduled see the same value. (It's read only so there's no channel). Preserving the original behavior for VCPUs that don't care probably requires reading it's initial value and saving that to use as the init value for new VCPU objects. I'd be ok with that too.  

As for motivating this change, the baseline only works if the kernel is loaded on the 0th core. If loading it on a different core, the guest VCPUs also start seeing a different boot core and it would be preferable to be able to set it back to 0.

Indanz:
> Preserving the original behavior for VCPUs that don't care probably requires reading it's initial value and saving that to use as the init value for new VCPU objects. I'd be ok with that too.

You change the behaviour for current SMP systems too. Why not keep the old behaviour for SMP and let unicore do the same? That is, don't init it at all (which probably means it's 0 for the guest, as untyped is zeroed).

> As for motivating this change, the baseline only works if the kernel is loaded on the 0th core. If loading it on a different core, the guest VCPUs also start seeing a different boot core and it would be preferable to be able to set it back to 0.

But the old default was 0, so it should work. If the VMM launches a multicore kernel, it needs to set the correct MPIDR anyway.

kent-mcleod:
> But the old default was 0, so it should work.

This wasn't the case on single core configurations. The register was ignored by the kernel and it's initial value appears to be the same as the real MPIDR register it's virtualizing.

My preference is to leave it 0 when the object is initialized which is inline with the multicore behavior. but I think this would be a breaking change for non-multicore.  So I guess we can either introduce a breaking change for single core, and make the behavior consistent, or do something different for single core and have inconsistent behavior across configurations.

Indanz:
> My preference is to leave it 0 when the object is initialized which is inline with the multicore behavior. but I think this would be a breaking change for non-multicore.

This would be my preference too. I agree it would change current behaviour for unicore, but I don't think 0 will break anything.

The `#if CONFIG_MAX_NUM_NODES > 1` should be removed from:

- [vcpu_start](https://github.com/seL4/seL4_projects_libs/blob/4662171c6026c789974acdc60efc9f7ee96de64e/libsel4vm/src/arch/arm/vm.c#L182C22-L182C39). It currently sets it to `BIT(24) | BIT(31)`, but only for SMP.
- [Microkit's libvmm](https://github.com/au-ts/libvmm/blob/a6e3070cacc5afc060e6b5325a29932197dc5c0f/src/arch/aarch64/vcpu.c#L51)'s needs to set it correctly for both SMP and unicore, now it sets it to 0 for SMP only.

And anyone having a custom VMM may need to set `seL4_VCPUReg_VMPIDR_EL2`, in the rare chance when 0 doesn't work, but at least that bit of code will be SMP ready then.

kent-mcleod:
@lsf37 is this still awaiting verification?

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -123,6 +123,7 @@
 #define REG_HCR_EL2         "hcr_el2"
 #define REG_VTCR_EL2        "vtcr_el2"
 #define REG_VMPIDR_EL2      "vmpidr_el2"
+#define REG_MPIDR_EL1      "mpidr_el1"
```
Indanz, line 126: Alignment seems off here.

Indanz (approved):

lsf37:
> @lsf37 is this still awaiting verification?

Yes -- I wasn't sure if the discussion was settled or not. If this is the version we want, I can have a look at it soon.

Indanz:
> If this is the version we want, I can have a look at it soon.

It's the version that Kent and I prefer, but it does change behaviour for unicore. I don't think it will cause problems and the user space change to fix it is small anyway (removing ifdefs).

Ivan-Velickovic:
> * [Microkit's libvmm](https://github.com/au-ts/libvmm/blob/a6e3070cacc5afc060e6b5325a29932197dc5c0f/src/arch/aarch64/vcpu.c#L51)'s needs to set it correctly for both SMP and unicore, now it sets it to 0 for SMP only.

Thanks for pointing that out, I will fix that.

lsf37:
It turns out, we need to add padding the VCPU struct now so that it remains packed. I'll push a commit for that once the proofs have passed.

include/arch/arm/arch/object/vcpu.h
```diff
@@ -78,11 +78,11 @@ struct vcpu {
     word_t regs[seL4_VCPUReg_Num];
     bool_t vppi_masked[n_VPPIEventIRQ];
 #ifdef CONFIG_VTIMER_UPDATE_VOFFSET
-    /* vTimer is 8-bytes wide and has same alignment requirement.
-     * The struct will remain packed on 32-bit platforms when n_VPPIEventIRQ
-     * is odd, but were it to become even, an extra word of padding will be
```
Xaphiosis, line 83: I'd suggest changing it to "but *since it is currently* even, and extra word of padding *is* necessary", so if another word is added we can go back easily. I have difficulty reading the new proposed version.

lsf37: Happy to tweak, but it's not currently even. The previous comment was incomplete in the sense that the padding isn't determined by `n_VPPIEventIRQ` alone -- it's dependent on whether the sum of the length of the two arrays is even or odd. It's currently odd, which means we need padding. If the sum is even we don't.

So maybe the following?
```
    /* vTimer is 8-bytes wide and has same alignment requirement.
     * If the sum of  n_VPPIEventIRQ and seL4_VCPUReg_Num is even,
     * we do not need extra padding. If the sum is odd we do. It currently
     * is odd, so the extra padding above is necessary for the struct to
     * remain packed on 32 bit platforms.
```

lsf37: (comment now updated as above)

lsf37:
(Rebased in preparation for merge when the proofs on https://github.com/seL4/l4v/pull/723 have passed)

lsf37 (approved):
Ready to merge when the proofs have passed.

lsf37:
The proofs are fine, but this PR broke the CAmkES VM, see the runs in <https://github.com/seL4/camkes-vm-examples/actions/runs/8280413677/job/22659343429>

kent-mcleod:
> The proofs are fine, but this PR broke the CAmkES VM, see the runs in https://github.com/seL4/camkes-vm-examples/actions/runs/8280413677/job/22659343429

This was needing to be merged at the same time: https://github.com/seL4/seL4_projects_libs/pull/125

lsf37:
Still seems to be breaking after merging seL4/seL4_projects_libs#125 with the same error:

```
install_vm_devices@main.c:704 module name: plat
  Loading Kernel: 'linux'
  Loading Generated DTB
  seL4 called fail at /github/workspace/kernel/include/arch/arm/armv/armv7ve/armv/vcpu.h:628 in function vcpu_hw_write_reg, saying "ARM/HYP: Invalid register index"
  halting...
  Kernel entry via Syscall, number: 5, Recv
```

Indanz:
We forgot to remove the `#ifdef ENABLE_SMP_SUPPORT` from `vcpu_hw_write_reg`.

PR #1182, opened 2024-02-01
Fix QEMU/arm CPU selection

axel-h:
- CMake, QEMU/arm: fix parser usage, so the "ve" extension is enabled
- QEMU/arm: select CPU from architecture, so aarch32 defaults to Cortex-A15 and aarch64 to Cortex-A53

Fixed the issue from https://github.com/seL4/seL4/pull/824#issuecomment-1921221807

Since we are not using QEMU for ARMv7 virtualzation, these issues have not shown up. See also https://github.com/seL4/camkes-vm-examples/issues/73

### DIFF
--- a/src/plat/qemu-arm-virt/config.cmake
+++ b/src/plat/qemu-arm-virt/config.cmake
@@ -12,7 +12,10 @@ declare_platform(qemu-arm-virt KernelPlatformQEMUArmVirt PLAT_QEMU_ARM_VIRT Kern
 set(qemu_user_top 0xa0000000)
 
 macro(setup_qemu_armv7)
-    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "")
+    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "" ${ARGN})
+    if(ARMV7_OPTIONS_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR "Unknown arguments: ${ARMV7_OPTIONS_UNPARSED_ARGUMENTS}")
+    endif()
     set(QEMU_ARCH "arm")
     set(KernelArchArmV7a ON)
     if(ARMV7_OPTIONS_ve)
--- a/src/plat/qemu-arm-virt/config.cmake
+++ b/src/plat/qemu-arm-virt/config.cmake
@@ -38,8 +38,23 @@ endmacro()
 if(KernelPlatformQEMUArmVirt)
 
     if(NOT ARM_CPU)
-        message(STATUS "ARM_CPU not set, defaulting to cortex-a53")
-        set(ARM_CPU "cortex-a53")
+        # Our default QEMU configuration is AARCH64/Cortex-a53. If
+        # KernelSel4Arch is set to aarch32/arm_hyp, the default is Cortex-a15.
+        # If both ARM_CPU and KernelSel4Arch are set, conflicting values will be
+        # detected eventually. Note that the KernelSel4Archxxx variables are not
+        # set up here, because declare_seL4_arch() has not been called yet.
+        set(
+            arch_cpu_mapping # element format: "KernelSel4Arch:ARM_CPU"
+            ":cortex-a53" # used if KernelSel4Arch is empty or not set
+            "aarch64:cortex-a53"
+            "arm_hyp:cortex-a15"
+            "aarch32:cortex-a15"
+        )
+        if(NOT ";${arch_cpu_mapping};" MATCHES ";${KernelSel4Arch}:([^;]*);")
+            message(FATAL_ERROR "unsupported KernelSel4Arch: '${KernelSel4Arch}'")
+        endif()
+        set(ARM_CPU "${CMAKE_MATCH_1}")
+        message(STATUS "ARM_CPU not set, defaulting to ${ARM_CPU}")
     endif()
 
     if("${ARM_CPU}" STREQUAL "cortex-a7")


src/plat/qemu-arm-virt/config.cmake
```diff
@@ -34,9 +37,32 @@ endmacro()
...
+        if(NOT KernelSel4Arch)
+            set(KernelSel4Arch "aarch64")
+        endif()
+        set(
+            arch_cpu_mapping # element format: "KernelSel4Arch:ARM_CPU"
+            "aarch64:cortex-a53"
+            "arm_hyp:cortex-a15"
+            "aarch32:cortex-a15"
+        )
+        if(NOT ";${arch_cpu_mapping};" MATCHES ";${KernelSel4Arch}(:([^;]*))?;")
```
lsf37, line 61: The `(:..)?` part looks strange to me. Why do we allow that to be optional? Don't we want to force a match?

i.e.
```suggestion
        if(NOT ";${arch_cpu_mapping};" MATCHES ";${KernelSel4Arch}:([^;]*);")
```

(This might change the match number, not sure, i.e. potentially `CMAKE_MATCH_1` below)

This will only really be relevant if we mess up the format in the table above, but I'd still prefer the version that forces the second part to exist.

lsf37 (approved):
This looks fine from my side (modulo that regexp nitpick) and should solve the problem. I'm not exactly a cmake expert though, so would appreciate a second review.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -34,9 +37,32 @@ endmacro()
 
 if(KernelPlatformQEMUArmVirt)
 
+    # Our default QEMU configuration is AARCH64/Cortex-a53. If KernelSel4Arch is
+    # set to aarch32, the default is Cortex-a15. Setting ARM_CPU allows picking
+    # a specific CPU and forces KernelSel4Arch to the respective architecture.
+    # Note that the QEMU virt platform has some limitations:
+    # - only specific ARMv7 and ARMv8 cores can be used.
+    # - ARMv8 cores can't be started in AARCH32 mode
+    # - Cortex-A32 that implements AARCH32 only is not suppoted
```
Indanz, line 46: I think I would remove this section, as it doesn't really help and can get out-of-date easily.

If you do keep it, I would say "qemu-arm-virt platform" instead of "QEMU virt platform" if the limitation is in seL4 and not QEMU, otherwise it's ambiguous.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -34,9 +37,32 @@ endmacro()
...
+
     if(NOT ARM_CPU)
-        message(STATUS "ARM_CPU not set, defaulting to cortex-a53")
-        set(ARM_CPU "cortex-a53")
+        # We don't have any KernelSel4Archxxx varable set up here, because
```
Indanz, line 49:
```suggestion
        # We don't have any KernelSel4Archxxx variable set up here, because
```

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -34,9 +37,32 @@ endmacro()
...
+        # declare_seL4_arch() has no been called yet. So we must check
```
Indanz, line 50: The "yet" implies it will be called in the future, but that isn't the case, is it? Because if it was, we shouldn't guess something here, but wait for that future call instead.

This comment doesn't explain why `setup_qemu_armv7` should set anything. I think the intention is to have sane defaults in case nothing was specified, but that intention isn't clear from the comments.

Indanz (approved):
I guess it's okay, but I don't know CMake. I would have used if-elseif instead of obscure CMake syntax myself.

So in the end all this does is set default `KernelSel4Arch` and `ARM_CPU` if it wasn't set by the project.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -34,9 +37,32 @@ endmacro()
 
 if(KernelPlatformQEMUArmVirt)
 
+    # Our default QEMU configuration is AARCH64/Cortex-a53. If KernelSel4Arch is
+    # set to aarch32, the default is Cortex-a15. Setting ARM_CPU allows picking
+    # a specific CPU and forces KernelSel4Arch to the respective architectue.
```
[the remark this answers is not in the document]
axel-h: fixed

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -12,7 +12,10 @@ declare_platform(qemu-arm-virt KernelPlatformQEMUArmVirt PLAT_QEMU_ARM_VIRT Kern
 set(qemu_user_top 0xa0000000)
 
 macro(setup_qemu_armv7)
-    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "")
+    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "" ${ARGN})
+    if(ARMV7_OPTIONS_UNPARSED_ARGUMENTS)
+        message(FATAL_ERROR "Unknown arguments: ${ARMV7_OPTIONS_UNPARSED_ARGUMENTS}")
+    endif()
```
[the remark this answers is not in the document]
axel-h: Clarified the commit message that it's indeed `${ARGN}` was missing, and it adds an additional check.

axel-h:
> I guess it's okay, but I don't know CMake. I would have used if-elseif instead of obscure CMake syntax myself.

Fair point. It's a bit less redundant than the if-else-cascade. This is the closest CMake has currently to support maps. We are using this in a few other places also, so it could move to a helper function one day to hide the obscure details

> So in the end all this does is set default `KernelSel4Arch` and `ARM_CPU` if it wasn't set by the project.

Yes. And it takes `KernelSel4Arch` into account for selecting `ARM_CPU`. Before we were always using `Cortex-A53`, and that conflicts when the project sets `arm_hyp`.

src/plat/qemu-arm-virt/config.cmake
axel-h: The QEMU platform is "virt". I've added a line with a link to the full QEMU documentation for this platform now. I prefer to keep this comment, as one can be mislead easily to assume the virt platform can emulate everything.

Indanz: If these are limitations of QEMU and not seL4, I would definitely remove this comment, as it is very confusing and doesn't help the user.

src/plat/qemu-arm-virt/config.cmake
axel-h: CMake actually iterates over the configuration multiple times, until it becomes stable (or abort after 100 iterations if not) . Also, CMake cache some configuration. So what one will see that the variable can be set in some cases when we pass here. Only when you do a clean build, one will see that it is not set n the first iteration). 

`setup_qemu_armv7()`/`setup_qemu_armv7()` will call `declare_seL4_arch()` based on the selected CPU in the code below. We have to solve a check-egg problem here. In the seL4  build system `declare_seL4_arch()` sets the default and understands KernelSel4Arch`. But it does not know about QEMU specific `ARM_CPU` that one can specify. So we have to do some guessing here first t ensure `declare_seL4_arch()` picks the right default then.

src/plat/qemu-arm-virt/config.cmake
axel-h: I can remove this again, but this is really intended for people who try to make more special QEMU configurations work. I have spent some time on this topic, just to see at some point, that the `virt` platform does not allows some things. Background it, that it not meant to be a generic ARM emulator, but implements the ARM variant of the "virt" machine that the QEMU developer came up with. It could be much more generic, but there is some resistance in the QEMU community to allow that.

Indanz: Yes, but leave it to QEMU to give errors on launch. A mostly hidden comment in some CMake file is not going to help anyone. And if it is ever added to QEMU, the comment becomes out of date.

src/plat/qemu-arm-virt/config.cmake
Indanz: My worry is that we're working around CMake file bugs in projects here instead of fixing it there. Guessing here will make it much harder to fix problems in the future, even though we may have less problems initially.

axel-h: I think what we have now is the proper way to handle this. There is not much so guessing involved actually, the rulsed are quite simple:

| Config | Result | Notes
|--------|--------|--------|
| set `ARM_CPU` | use it (or error) and overwrite `KernelSel4Arch` | raise error if both `ARM_CPU` and `KernelSel4Arch` are set to conflicting or unsupported combinations?  |
| set `KernelSel4Arch` | cortex-A15 or cortex-A53 (or error) |
| nothing | cortex-A53 | this case was broken as described in https://github.com/seL4/seL4/pull/824#issuecomment-1921221807. Seems ARMv7 QEMU is rarely used nowadays, thus we did not see this earlier |
| conflicting | raises an error | fix project, e.g.  https://github.com/seL4/sel4webserver/pull/18 |

Now, if a projects change the variables after the seL4 build system CMake files have run, things can get tricky. Since CMake iterates over the config, it may end up in a stable state. But we may also have broken some older project that do this. I'd leave it to somebody raising the issues then, so we can check what the project does. If it's not somewhere in CI or somebody actively using it, things might also be obsolete eventually.

Indanz: Silently overwriting `KernelSel4Arch` is what seems wrong to me. But like you said, it may be necessary in case of reconfiguration, if config values are cached.

axel-h: I will add some check that this fail the build then if both are set and conflicting.

src/plat/qemu-arm-virt/config.cmake
axel-h: removed it

src/plat/qemu-arm-virt/config.cmake
axel-h: Turns out that this case is handles already, as `KernelSel4Arch` is not overwritten anywhere. It is set in case the user has not set it. And if the user has set it, then the conflict will be notices already. So things should be fine here also.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -12,7 +12,10 @@ declare_platform(qemu-arm-virt KernelPlatformQEMUArmVirt PLAT_QEMU_ARM_VIRT Kern
 set(qemu_user_top 0xa0000000)
 
 macro(setup_qemu_armv7)
-    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "")
+    cmake_parse_arguments(ARMV7_OPTIONS "ve" "" "" ${ARGN})
```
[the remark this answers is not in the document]
axel-h: The general idea of this mechanism was having a way to be able to specify ARM architecture extensions per core. Practically it is a bit overkill for what we currently do. Especially since both Cortex-A7 and A15 support "ve" actually, and these are the only ARMv7 cores supported by QEMU virt. 
Can we we postpone changing this until after this PR is merged, so the original issue gets fixed?

lsf37 (approved):

src/plat/qemu-arm-virt/config.cmake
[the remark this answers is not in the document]
lsf37: Happy to leave that for a separate PR.

PR #1183, opened 2024-02-01
gicv3: Use split EOI mode

kent-mcleod:
Separate priority drop from deactivation so that when the kernel
delegates interrupts to userlevel, they can be left in the active state
until user level performs the ack invocation which deactivates them.
This is more efficient than doing a disable and enable operation
for each handled interrupt as it is core local and doesn't require
sending operations to the GIC distributor.

### DIFF
--- a/include/arch/arm/arch/machine/gic_v3.h
+++ b/include/arch/arm/arch/machine/gic_v3.h
@@ -306,14 +306,35 @@ static inline void maskInterrupt(bool_t disable, irq_t irq)
     }
 }
 
+
+static inline void deactivateInterrupt(irq_t irq)
+{
+    word_t hw_irq = IRQT_TO_IRQ(irq);
+    /* Perform deactivation of hw_irq */
+    SYSTEM_WRITE_WORD(ICC_DIR_EL1, hw_irq);
+}
+
 static inline void ackInterrupt(irq_t irq)
 {
     assert(IS_IRQ_VALID(active_irq[CURRENT_CPU_INDEX()])
            && (active_irq[CURRENT_CPU_INDEX()] & IRQ_MASK) == IRQT_TO_IRQ(irq));
+    active_irq[CURRENT_CPU_INDEX()] = IRQ_NONE;
+
+    word_t hw_irq = IRQT_TO_IRQ(irq);
 
     /* Set End of Interrupt for active IRQ: ICC_EOIR1_EL1 */
-    SYSTEM_WRITE_WORD(ICC_EOIR1_EL1, active_irq[CURRENT_CPU_INDEX()]);
-    active_irq[CURRENT_CPU_INDEX()] = IRQ_NONE;
+    /* Perform priority drop for current IRQ */
+    SYSTEM_WRITE_WORD(ICC_EOIR1_EL1, hw_irq);
+
+    // If the IRQ is not going to user level then we need to deactivate it too.
+    if (unlikely(hw_irq > maxIRQ) ||
+        intStateIRQTable[IRQT_TO_IDX(irq)] != IRQSignal) {
+        /* There needs to be an isb() to ensure completion of the system
+         * register write in ackInterrupt
+         */
+        isb();
+        deactivateInterrupt(irq);
+    }
 
 }
 
--- a/include/machine/interrupt.h
+++ b/include/machine/interrupt.h
@@ -98,6 +98,20 @@ static inline void maskInterrupt(bool_t disable, irq_t irq);
 static inline void ackInterrupt(irq_t irq);
 
 /**
+ * Deactivates the interrupt
+ *
+ * When the interrupt controller supports delegating the interrupt to a lower
+ * privilege level, this function can be called to signal the completion of
+ * interrupt processing so that the interrupt state machine can be moved out of
+ * the active state.
+ *
+ * Currently only supported by gicv3 driver.
+ *
+ * @param[in]  irq   The interrupt request
+ */
+static inline void deactivateInterrupt(irq_t irq);
+
+/**
  * Called when getActiveIRQ returns irqInvalid while the kernel is handling an
  * interrupt entry. An implementation is not required to do anything here, but
  * can report the spurious IRQ or try prevent it from reoccuring.
@@ -116,3 +130,7 @@ static inline void handleSpuriousIRQ(void);
  */
 static inline void handleReservedIRQ(irq_t irq);
 
+#ifndef CONFIG_ARM_GIC_V3_SUPPORT
+
+static inline void deactivateInterrupt(irq_t irq) {}
+#endif
--- a/src/object/interrupt.c
+++ b/src/object/interrupt.c
@@ -145,7 +145,14 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
 
 #if defined ENABLE_SMP_SUPPORT && defined CONFIG_ARCH_ARM
     if (IRQ_IS_PPI(irq) && IRQT_TO_CORE(irq) != getCurrentCPUIndex()) {
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+        /* According to the GICv3 spec, SPIs can be deactivated from any PE,
+         * but SGIs and PPIs must be deactivated from their target PE.
+         */
+        doRemoteDeactivatePrivateInterrupt(IRQT_TO_CORE(irq), IRQT_TO_IDX(irq));
+#else /* CONFIG_ARM_GIC_V3_SUPPORT */
         doRemoteMaskPrivateInterrupt(IRQT_TO_CORE(irq), false, IRQT_TO_IDX(irq));
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
         return;
     }
 #endif
--- a/src/object/interrupt.c
+++ b/src/object/interrupt.c
@@ -149,8 +149,12 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
         return;
     }
 #endif
-    maskInterrupt(false, irq);
-#endif
+    if (config_set(CONFIG_ARM_GIC_V3_SUPPORT)) {
+        deactivateInterrupt(irq);
+    } else {
+        maskInterrupt(false, irq);
+    }
+#endif /* CONFIG_ARCH_RISCV */
 }
 
 void invokeIRQHandler_SetIRQHandler(irq_t irq, cap_t cap, cte_t *slot)
@@ -218,8 +222,11 @@ void handleInterrupt(irq_t irq)
 #endif
         }
 #ifndef CONFIG_ARCH_RISCV
-        maskInterrupt(true, irq);
+        if (!config_set(CONFIG_ARM_GIC_V3_SUPPORT)) {
+            maskInterrupt(true, irq);
+        }
 #endif
+
         break;
     }
 
--- a/include/arch/arm/arch/smp/ipi.h
+++ b/include/arch/arm/arch/smp/ipi.h
@@ -16,6 +16,9 @@ typedef enum {
     IpiRemoteCall_InvalidateTranslationAll,
     IpiRemoteCall_switchFpuOwner,
     IpiRemoteCall_MaskPrivateInterrupt,
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    IpiRemoteCall_DeactivatePrivateInterrupt,
+#endif
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
     IpiRemoteCall_VCPUInjectInterrupt,
 #endif
--- a/include/arch/arm/arch/smp/ipi_inline.h
+++ b/include/arch/arm/arch/smp/ipi_inline.h
@@ -41,5 +41,12 @@ static inline void doRemoteMaskPrivateInterrupt(word_t cpu, word_t disable, word
 {
     doRemoteOp2Arg(IpiRemoteCall_MaskPrivateInterrupt, disable, irq, cpu);
 }
+
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+static inline void doRemoteDeactivatePrivateInterrupt(word_t cpu, word_t irq)
+{
+    doRemoteOp1Arg(IpiRemoteCall_DeactivatePrivateInterrupt, irq, cpu);
+}
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
 #endif /* ENABLE_SMP_SUPPORT */
 
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -276,9 +276,9 @@ BOOT_CODE static void cpu_iface_init(void)
     /* Set priority mask register: ICC_PMR_EL1 */
     SYSTEM_WRITE_WORD(ICC_PMR_EL1, DEFAULT_PMR_VALUE);
 
-    /* EOI drops priority and deactivates the interrupt: ICC_CTLR_EL1 */
+    /* EOI drops priority of the interrupt, deactivation happens separately: ICC_CTLR_EL1 */
     SYSTEM_READ_WORD(ICC_CTLR_EL1, icc_ctlr);
-    icc_ctlr &= ~GICC_CTLR_EL1_EOImode_drop;
+    icc_ctlr |= GICC_CTLR_EL1_EOImode_drop;
     SYSTEM_WRITE_WORD(ICC_CTLR_EL1, icc_ctlr);
 
     /* Enable Group1 interrupts: ICC_IGRPEN1_EL1 */
--- a/src/arch/arm/smp/ipi.c
+++ b/src/arch/arm/smp/ipi.c
@@ -43,6 +43,12 @@ void handleRemoteCall(IpiRemoteCall_t call, word_t arg0, word_t arg1, word_t arg
             maskInterrupt(arg0, IDX_TO_IRQT(arg1));
             break;
 
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+        case IpiRemoteCall_DeactivatePrivateInterrupt:
+            deactivateInterrupt(IDX_TO_IRQT(arg1));
+            break;
+#endif
+
 #if defined CONFIG_ARM_HYPERVISOR_SUPPORT && defined ENABLE_SMP_SUPPORT
         case IpiRemoteCall_VCPUInjectInterrupt: {
             virq_t virq;


kent-mcleod:
I don't plan to add this support to the other GIC versions in this PR as there would be more substantial changes involved:

    GICv2 with SGIs stores the source CPUID in the active IRQ state. This is needed for both priority drop and final deactivation. This feature was removed in the GICv3. The kernel doesn't track this additional state when delegating IRQs to user level and so wouldn't be able to correctly end an SGI interrupt without adding more book-keeping for each IRQ number.
    Some platforms that use the GICv2 code implementation don't actually support split EOI mode (I guess they don't implement gic 2.0 but are more 1.0). The imx6/sabre is an example of this.
    On platforms that do support split-eoi mode, the second register for ending the interrupt (GICC_DIR) is located at address 0x1000, but the kernel has only mapped in a single 4k frame for the device and so can't currently access it.
    None of these are show-stoppers, but it'd be a fair amount of code changes and I don't think it's justified at this stage unless there was a cortex-a15 platform that really wanted the change.

Indanz (changes requested):
Other than the `doRemoteDeactivatePrivateInterrupt` typo and the ifdef maze, it looks good.

Would be good to test this cross-core too though, either manually or with sel4test. But sel4test doesn't have the infrastructure to trigger IRQs on demand yet.

Indanz:
I'm not convinced that RISC-V's `invokeIRQHandler_AckIRQ()` implementation is correct in case the ack is done on a different core than where the IRQ originated. It should use your new `doRemoteDeactivatePrivateInterrupt()` too I think, or somehow remember the originating core and use that in `plic_complete_claim()` instead of `plic_get_current_hart_id()`. Or ACK it immediately like is done for QEMU.

include/arch/arm/arch/machine/gic_v3.h
```diff
@@ -307,14 +307,32 @@ static inline void maskInterrupt(bool_t disable, irq_t irq)
...
     assert(IS_IRQ_VALID(active_irq[CURRENT_CPU_INDEX()])
            && (active_irq[CURRENT_CPU_INDEX()] & IRQ_MASK) == IRQT_TO_IRQ(irq));
 
     /* Set End of Interrupt for active IRQ: ICC_EOIR1_EL1 */
-    SYSTEM_WRITE_WORD(ICC_EOIR1_EL1, active_irq[CURRENT_CPU_INDEX()]);
+    prioDropInterrupt(irq);
     active_irq[CURRENT_CPU_INDEX()] = IRQ_NONE;
+    if (unlikely(IRQT_TO_IRQ(irq) > maxIRQ) || intStateIRQTable[IRQT_TO_IDX(irq)] != IRQSignal) {
+        deactivateInterrupt(irq);
+    }
```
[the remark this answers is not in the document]
kent-mcleod: It's at least necessary for kernel IRQs (IRQTimer, IRQIPI, IRQReserved). In the last PR I had this line at the bottom of `handleInterrupt()` but it was suggested I put it here. I think I'll move it as it seems kind of wrong for ackInterrupt() to have different behavior depending on kernel interrupt type.

kent-mcleod:
One potential issue is with what happens if user-level keeps calling seL4_IRQHandler_Ack for the same interrupt. 
Previously, unmasking the interrupt was idempotent, but `deactivateInterrupt` isn't.   The GICv3 specification states:
> When ICC_CTLR_EL1.EOImode or ICC_CTLR_EL3.EOIMode_EL3 == 1 but the interrupt is not active in the
Distributor, writes to ICC_DIR_EL1 must be ignored. If supported, an implementation might generate a system
error.

include/arch/arm/arch/machine/gic_v3.h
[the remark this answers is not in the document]
Indanz: Ah, of course. For some reason I thought that path was taken for unwanted interrupts.

Code wise it seems cleaner to have as much GICv3 code in GICv3 files and keeping it out of common.

Indanz:
The same document also says:

>  SPIs can be deactivated by a different PE.

So perhaps `IpiRemoteCall_DeactivatePrivateInterrupt()` isn't necessary for SPIs.

src/object/interrupt.c
```diff
@@ -145,12 +145,20 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
...
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
         return;
     }
 #endif
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    deactivateInterrupt(irq);
+#else /* CONFIG_ARM_GIC_V3_SUPPORT */
     maskInterrupt(false, irq);
-#endif
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
```
Indanz, line 160: If the `endif` is this close to the `if` I would drop the comment as it only makes things messy and it's clear enough without it.

Indanz (approved):

src/object/interrupt.c
```diff
@@ -145,12 +145,20 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
 
 #if defined ENABLE_SMP_SUPPORT && defined CONFIG_ARCH_ARM
     if (IRQ_IS_PPI(irq) && IRQT_TO_CORE(irq) != getCurrentCPUIndex()) {
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+        doRemoteDeactivatePrivateInterrupt(IRQT_TO_CORE(irq), IRQT_TO_IDX(irq));
+#else /* CONFIG_ARM_GIC_V3_SUPPORT */
         doRemoteMaskPrivateInterrupt(IRQT_TO_CORE(irq), false, IRQT_TO_IDX(irq));
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
         return;
     }
```
lsf37, line 154: This kind of pattern in generic code will make it impossible for us to switch between GICv2 and GICv3 in verification without making it an entire new architecture. Can this be put behind a generic interface that both GIC versions implement?

src/object/interrupt.c
```diff
@@ -145,12 +145,20 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
...
 #endif
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    deactivateInterrupt(irq);
+#else /* CONFIG_ARM_GIC_V3_SUPPORT */
```
lsf37, line 158: Same here, having ARM config dependent ifdefs in generic code will make generic proofs impossible -- it's fine for those that we're not planning to ever verify, but GICv3 does sound like something we'd want pretty soon.

src/object/interrupt.c
```diff
@@ -217,9 +225,10 @@ void handleInterrupt(irq_t irq)
             printf("Undelivered IRQ: %d\n", (int)IRQT_TO_IRQ(irq));
 #endif
         }
-#ifndef CONFIG_ARCH_RISCV
+#if !defined(CONFIG_ARCH_RISCV) && !defined(CONFIG_ARM_GIC_V3_SUPPORT)
         maskInterrupt(true, irq);
 #endif
```
lsf37, line 230: We don't do this much in the kernel usually, but would it work to us an actual `if` instead of condition compilation here? Then we can follow both paths in the verification and it'll work either way. IIRC, the compile should optimise out any `if (0)` cases.

src/object/interrupt.c
lsf37: I think that would help, yes

src/object/interrupt.c
lsf37: e.g.
```c
if (!config_set(CONFIG_ARCH_RISCV) && !config_set(CONFIG_ARM_GIC_V3_SUPPORT)) {
  maskInterrupt(true, irq);
}
```

(for that the function has to exist with this signature in those configs, which I haven't checked)

Indanz: That would work, but what do the proofs assume? I'd have a look at the IRQ abstract model, because all we have now is `maskInterrupt` AFAIK, but reality is more complex than that. Does the proof know about individual IRQ states currently?

lsf37: It does, yes. It at least keeps track of what is masked on the IRQ controller side and what not. I would need to check to see if there is more.

Indanz: Then you have no choice other than extending the model to cope with how HW actual work for verified platforms.

lsf37: Yes we'll have to do that when we get to gicv3, but we would still have more trouble with conditional compilation (`#ifdef`) vs control flow (normal `if`). If gicv3 was forever outside verification it wouldn't matter, then `#ifdef` is not bad, because we never see it. But if we want one proof that can do both GIC versions, we need to be able to see both control flow paths.

It's possible that this rabbit hole is too deep, we can look at all this more properly when we actually get to the verification of GICv3. It'll actually be easier in the short term if it's an `#ifdef`, just wanted to get some feedback on if there would be problems from the kernel side with using `if` and `config_set`. I've searched through the code a bit more, and we do seem to use it a lot more than I initially thought, so it seems generally fine.

Indanz: If changing the model you would have an unconditional call for both here. One would do nothing for GICv3, and perhaps another function would do nothing for GICv2. I suspect you want to replace maskInterrupt and ackInterrupt with more fine grained versions that make it clearer what the intention is, that would hopefully also get rid of the RISCV ifdefs.

src/object/interrupt.c
lsf37: We do now support GICv3 in the verification, and having an `ifdef` here will break the proofs.

I do need the version where this is an actual `if` and where one branch does nothing. It means we do have to declare prototypes for `doRemoteDeactivatePrivateInterrupt` for platforms that don't support it, which is annoying, but there is no other way around it that I can see.  It's Ok if those platform panic and call `halt()` -- we will proof that the function is not called there (and the compiler should be able to see the same).

lsf37: Actually, this specific one is fine, because it is also guarded by SMP, which we never see -- but I will go through and point out the other ones.

src/object/interrupt.c
lsf37: This one is also still under the SMP guard, so it's fine for the proof.

src/object/interrupt.c
lsf37: In this one we do need the actual `if`. The proofs are fine dealing with the two models (conditionally on config), but they can't deal with conditional compilation.

(There will be more such changes from `ifdef` to `if (defined..)`  in the future as we are drawing in more config options into the proof.)

lsf37: Indan's version of having one function that does nothing in one version would also work if we drive it down far enough to be under the machine interface for the proofs -- then we don't see the conditional compilation, instead we make a model that has a condition on the config for its behaviour.

In this case, we could do that for `maskInterrupt` and pull it in there. That would also be fine here.

src/object/interrupt.c
```diff
@@ -263,6 +272,14 @@ void handleInterrupt(irq_t irq)
...
      * kernel code does report here that it is done with handling it. */
     ackInterrupt(irq);
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    /* GICv3 uses split priority drop from deactivation, so need to explicitly
+     * deactivate non-forwarded interrupts after they're acknowledged.
+     */
+    if (intStateIRQTable[IRQT_TO_IDX(irq)] != IRQSignal) {
+        deactivateInterrupt(irq);
+    }
+#endif /* CONFIG_ARM_GIC_V3_SUPPORT */
```
lsf37, line 282: The `#ifdef` where will need to become an `if` for the proofs.

lsf37 (approved):
Ok, so this is generally fine from my side if we change the one `#ifdef` at the end into and `if` instead and either do the same for `maskInterrupt` or move the `ifdef` inside `maskInterrupt`. The latter only really works if we *never* want to do anything for GICv3 if `maskInterrupt` is called -- I haven't looked if that is actually the case.

When we have that change, I'll need a bit of time for re-verification, but it looks like it should only be a few days.

src/object/interrupt.c
lsf37: I agree that would be nicer. I'd want to do that in a separate PR, though, so we can verify it separately.

src/object/interrupt.c
```diff
@@ -263,6 +268,14 @@ void handleInterrupt(irq_t irq)
      * that for any interrupt reported by the platform specific code the generic
      * kernel code does report here that it is done with handling it. */
     ackInterrupt(irq);
+#ifdef CONFIG_ARM_GIC_V3_SUPPORT
+    /* GICv3 uses split priority drop from deactivation, so need to explicitly
+     * deactivate non-forwarded interrupts after they're acknowledged.
+     */
+    if (intStateIRQTable[IRQT_TO_IDX(irq)] != IRQSignal) {
+        deactivateInterrupt(irq);
```
kent-mcleod, line 280: There probably needs to be an isb() before `ICC_DIR_EL1` to synchronize with `ICC_EOIR0_EL1`

kent-mcleod:
A quick comparison on the imx8mq-evk with irquser benchmark (tweaked to measure full round trip IRQ processing with this diff:
```
--- a/apps/irquser/src/main.c
+++ b/apps/irquser/src/main.c
@@ -64,9 +64,9 @@ void ticker_fn(ccnt_t *results, volatile ccnt_t *current_time)
     for (int i = 0; i < N_RUNS; i++) {
         /* wait for irq */
         seL4_Wait(timer_signal, &badge);
+        sel4platsupport_irq_handle(irq_ops, timer_ntfn_id, badge);
         /* record result */
         SEL4BENCH_READ_CCNT(end);
-        sel4platsupport_irq_handle(irq_ops, timer_ntfn_id, badge);
         end_low = (seL4_Word) end;
         start = (seL4_Word) * current_time;
         results[i] = end_low - start;
```
)
IRQ path cycle count (measured from user level)
Without context switch
mean: 3266.3699999999999, "Stddev": 19.200957467708939 
goes to:
Mean": 1915.8, "Stddev": 24.935877361704758

sel4bench manifest: [sel4bench-manifest#4a75125e724a0f6ed7e2d4aa972cf24ed1fed654](https://github.com/seL4/sel4bench-manifest/commit/4a75125e724a0f6ed7e2d4aa972cf24ed1fed654)
```
{"Benchmark": "IRQ path cycle count (measured from user level)", "Results": [{"Type": "Measurement overhead", "Min": 1, "Max": 1, "Mean": 1.0, "Stddev": 0.0, "Variance": 0.0, "Mode": 1.0, "Median": 1.0, "1st quantile": 1.0, "3rd quantile": 1.0, "Samples": 100, "Raw results": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}, {"Type": "Without context switch", "Min": 3227, "Max": 3310, "Mean": 3266.3699999999999, "Stddev": 19.200957467708939, "Variance": 364.99000000000001, "Mode": 3269.0, "Median": 3269.0, "1st quantile": 3267.0, "3rd quantile": 3270.0, "Samples": 100, "Raw results": [3309, 3267, 3308, 3307, 3306, 3310, 3269, 3306, 3310, 3309, 3306, 3267, 3230, 3269, 3270, 3270, 3267, 3270, 3267, 3270, 3266, 3270, 3270, 3267, 3270, 3270, 3269, 3267, 3269, 3227, 3227, 3229, 3227, 3227, 3267, 3268, 3269, 3267, 3230, 3270, 3267, 3227, 3270, 3266, 3230, 3270, 3269, 3227, 3269, 3228, 3267, 3268, 3268, 3268, 3230, 3269, 3270, 3269, 3270, 3269, 3228, 3268, 3227, 3269, 3269, 3270, 3269, 3270, 3269, 3266, 3268, 3267, 3270, 3267, 3268, 3267, 3268, 3269, 3268, 3267, 3269, 3268, 3268, 3269, 3270, 3270, 3269, 3270, 3269, 3270, 3266, 3268, 3266, 3268, 3269, 3266, 3269, 3270, 3267, 3268]}, {"Type": "Without context switch (early processing)", "Min": 0, "Max": 0, "Mean": 1224.0, "Stddev": 11.415920126000897, "Variance": 129.02000000000001, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}, {"Type": "With context switch", "Min": 3226, "Max": 3309, "Mean": 3261.21, "Stddev": 19.596175673077159, "Variance": 380.17000000000002, "Mode": 3267.0, "Median": 3267.0, "1st quantile": 3266.0, "3rd quantile": 3269.0, "Samples": 100, "Raw results": [3306, 3309, 3309, 3267, 3306, 3308, 3267, 3270, 3270, 3230, 3270, 3270, 3268, 3267, 3267, 3266, 3268, 3267, 3270, 3270, 3270, 3229, 3268, 3229, 3228, 3269, 3268, 3228, 3269, 3268, 3270, 3230, 3268, 3270, 3267, 3270, 3267, 3228, 3265, 3267, 3228, 3268, 3266, 3226, 3269, 3267, 3269, 3269, 3270, 3230, 3269, 3266, 3268, 3266, 3226, 3266, 3268, 3229, 3226, 3266, 3265, 3267, 3267, 3266, 3230, 3267, 3267, 3230, 3266, 3267, 3228, 3227, 3270, 3268, 3226, 3269, 3230, 3267, 3267, 3269, 3267, 3230, 3230, 3267, 3268, 3226, 3270, 3266, 3270, 3266, 3270, 3268, 3267, 3267, 3266, 3270, 3269, 3269, 3269, 3268]}, 

```

With kernel: abe726c43edae9183236714e91d7e0535ffa11f3
```
"Benchmark": "IRQ path cycle count (measured from user level)", "Results": [{"Type": "Measurement overhead", "Min": 1, "Max": 1, "Mean": 1.0, "Stddev": 0.0, "Variance": 0.0, "Mode": 1.0, "Median": 1.0, "1st quantile": 1.0, "3rd quantile": 1.0, "Samples": 100, "Raw results": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}, {"Type": "Without context switch", "Min": 1891, "Max": 1991, "Mean": 1915.8, "Stddev": 24.935877361704758, "Variance": 615.58000000000004, "Mode": 1917.0, "Median": 1911.0, "1st quantile": 1901.0, "3rd quantile": 1917.0, "Samples": 100, "Raw results": [1982, 1988, 1978, 1967, 1989, 1991, 1985, 1987, 1991, 1968, 1969, 1907, 1904, 1911, 1917, 1901, 1907, 1897, 1921, 1911, 1897, 1918, 1891, 1917, 1921, 1895, 1914, 1921, 1891, 1917, 1921, 1891, 1915, 1903, 1909, 1914, 1901, 1893, 1915, 1904, 1911, 1898, 1901, 1911, 1914, 1903, 1911, 1897, 1920, 1892, 1917, 1903, 1911, 1898, 1902, 1909, 1917, 1902, 1893, 1915, 1904, 1895, 1917, 1917, 1893, 1913, 1920, 1892, 1917, 1917, 1911, 1916, 1917, 1909, 1916, 1919, 1910, 1898, 1901, 1909, 1917, 1903, 1911, 1916, 1917, 1895, 1901, 1920, 1892, 1897, 1918, 1908, 1898, 1901, 1911, 1898, 1901, 1907, 1917, 1917]}, {"Type": "Without context switch (early processing)", "Min": 0, "Max": 0, "Mean": 604.0, "Stddev": 10.13544635066817, "Variance": 101.7, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}, {"Type": "With context switch", "Min": 1891, "Max": 1993, "Mean": 1915.1600000000001, "Stddev": 19.200694431886227, "Variance": 364.98000000000002, "Mode": 1911.0, "Median": 1913.0, "1st quantile": 1904.0, "3rd quantile": 1919.0, "Samples": 100, "Raw results": [1974, 1993, 1981, 1971, 1978, 1967, 1973, 1900, 1921, 1909, 1916, 1902, 1911, 1913, 1928, 1907, 1911, 1919, 1918, 1897, 1918, 1911, 1917, 1928, 1911, 1926, 1920, 1904, 1917, 1921, 1920, 1897, 1919, 1908, 1913, 1921, 1892, 1914, 1901, 1911, 1916, 1903, 1911, 1915, 1920, 1891, 1900, 1919, 1911, 1913, 1921, 1907, 1913, 1902, 1907, 1899, 1921, 1910, 1917, 1902, 1909, 1917, 1917, 1907, 1899, 1921, 1908, 1917, 1901, 1911, 1914, 1904, 1891, 1897, 1921, 1907, 1900, 1919, 1907, 1914, 1918, 1904, 1928, 1918, 1911, 1917, 1915, 1907, 1913, 1918, 1894, 1916, 1919, 1910, 1900, 1904, 1893, 1901, 1902, 1891]}, {"Type": "With context switch (early processing)", "Min": 0, "Max": 0, "Mean": 603.0, "Stddev": 7.7875840057404613, "Variance": 60.039999999999999, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}]}, 
```

src/object/interrupt.c
kent-mcleod: This is what the binary looks like for the relevant part of handleInterrupt:
```
ffffff804001544c:       d518cc33        msr     icc_eoir1_el1, x19
ffffff8040015450:       91124294        add     x20, x20, #0x490
ffffff8040015454:       f0000000        adrp    x0, ffffff8040018000 <ksDomSchedule+0x478>
ffffff8040015458:       d2807fe1        mov     x1, #0x3ff                      // #1023
ffffff804001545c:       f9050801        str     x1, [x0, #2576]
ffffff8040015460:       f8737a80        ldr     x0, [x20, x19, lsl #3]
ffffff8040015464:       f100041f        cmp     x0, #0x1
ffffff8040015468:       54000040        b.eq    ffffff8040015470 <handleInterrupt+0xa0>  // b.none
ffffff804001546c:       d518cb33        msr     icc_dir_el1, x19
```

Indanz: Agreed, but only for the call in handleInterrupt, so I'd add an ISB there with a comment explaining why it's necessary.

Indanz:
What are the numbers without patch? If you want to measure seL4_IRQHandler_Ack, it's probably better to measure it directly instead of including noisy, random timer code from sel4platsupport_irq_handle. I don't understand where the huge difference is coming from.

kent-mcleod:
Without the patch: on imx8mq-evk
"Mean": 1224.6600000000001, "Stddev": 11.649771316089073,  
goes to
"Mean": 605.09000000000003, "Stddev": 9.7706011311815644,
```
 {"Benchmark": "IRQ path cycle count (measured from user level)", "Results": [{"Type": "Measurement overhead", "Min": 1, "Max": 1, "Mean": 1.0, "Stddev": 0.0, "Variance": 0.0, "Mode": 1.0, "Median": 1.0, "1st quantile": 1.0, "3rd quantile": 1.0, "Samples": 100, "Raw results": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}, {"Type": "Without context switch", "Min": 1178, "Max": 1255, "Mean": 1224.6600000000001, "Stddev": 11.649771316089073, "Variance": 134.36000000000001, "Mode": 1225.0, "Median": 1225.0, "1st quantile": 1218.0, "3rd quantile": 1226.0, "Samples": 100, "Raw results": [1253, 1254, 1253, 1255, 1255, 1255, 1253, 1253, 1255, 1253, 1227, 1215, 1216, 1218, 1219, 1215, 1218, 1230, 1215, 1227, 1225, 1225, 1226, 1225, 1226, 1225, 1225, 1228, 1226, 1224, 1228, 1225, 1215, 1178, 1215, 1218, 1219, 1219, 1219, 1215, 1226, 1225, 1228, 1226, 1225, 1225, 1224, 1224, 1227, 1225, 1226, 1228, 1224, 1218, 1219, 1215, 1215, 1218, 1219, 1215, 1217, 1215, 1215, 1215, 1216, 1228, 1225, 1226, 1226, 1226, 1225, 1226, 1225, 1225, 1225, 1225, 1226, 1226, 1215, 1217, 1215, 1217, 1215, 1218, 1219, 1219, 1215, 1218, 1215, 1219, 1224, 1224, 1228, 1226, 1226, 1225, 1226, 1225, 1225, 1226]}, {"Type": "Without context switch (early processing)", "Min": 0, "Max": 0, "Mean": 1224.0, "Stddev": 11.873491398059713, "Variance": 139.56999999999999, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}, {"Type": "With context switch", "Min": 1214, "Max": 1264, "Mean": 1223.1500000000001, "Stddev": 10.407359478423492, "Variance": 107.23, "Mode": 1219.0, "Median": 1219.0, "1st quantile": 1217.75, "3rd quantile": 1225.0, "Samples": 100, "Raw results": [1251, 1262, 1264, 1264, 1261, 1262, 1223, 1219, 1216, 1219, 1218, 1219, 1219, 1217, 1217, 1218, 1218, 1218, 1218, 1219, 1219, 1217, 1214, 1215, 1218, 1218, 1215, 1217, 1218, 1227, 1224, 1223, 1224, 1224, 1224, 1225, 1227, 1228, 1226, 1225, 1225, 1224, 1219, 1219, 1215, 1218, 1217, 1215, 1216, 1215, 1217, 1219, 1219, 1216, 1227, 1225, 1224, 1225, 1225, 1225, 1226, 1226, 1226, 1225, 1225, 1226, 1225, 1219, 1216, 1218, 1216, 1216, 1216, 1217, 1219, 1219, 1218, 1219, 1217, 1224, 1227, 1226, 1228, 1225, 1224, 1225, 1227, 1225, 1224, 1226, 1227, 1225, 1219, 1217, 1218, 1215, 1216, 1219, 1215, 1219]}, {"Type": "With context switch (early processing)", "Min": 0, "Max": 0, "Mean": 1223.0, "Stddev": 9.186617407617419, "Variance": 83.549999999999997, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}]},
```

```
 {"Benchmark": "IRQ path cycle count (measured from user level)", "Results": [{"Type": "Measurement overhead", "Min": 1, "Max": 1, "Mean": 1.0, "Stddev": 0.0, "Variance": 0.0, "Mode": 1.0, "Median": 1.0, "1st quantile": 1.0, "3rd quantile": 1.0, "Samples": 100, "Raw results": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}, {"Type": "Without context switch", "Min": 595, "Max": 633, "Mean": 605.09000000000003, "Stddev": 9.7706011311815644, "Variance": 94.510000000000005, "Mode": 608.0, "Median": 605.0, "1st quantile": 598.0, "3rd quantile": 607.0, "Samples": 100, "Raw results": [632, 630, 630, 631, 629, 629, 629, 632, 632, 633, 607, 595, 599, 597, 597, 599, 598, 596, 605, 604, 608, 605, 604, 608, 605, 607, 606, 607, 606, 606, 607, 596, 596, 599, 596, 598, 595, 598, 604, 608, 604, 607, 605, 604, 608, 607, 606, 606, 608, 607, 605, 597, 596, 598, 595, 599, 595, 597, 595, 598, 599, 598, 595, 606, 604, 604, 606, 607, 605, 608, 607, 605, 608, 608, 605, 608, 596, 595, 597, 598, 595, 599, 595, 598, 597, 596, 596, 595, 607, 606, 605, 604, 608, 608, 607, 605, 608, 608, 605, 608]}, {"Type": "Without context switch (early processing)", "Min": 0, "Max": 0, "Mean": 605.0, "Stddev": 10.161324975244383, "Variance": 102.22, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}, {"Type": "With context switch", "Min": 595, "Max": 642, "Mean": 603.65999999999997, "Stddev": 9.2594478095280461, "Variance": 84.879999999999995, "Mode": 608.0, "Median": 604.0, "1st quantile": 597.0, "3rd quantile": 607.0, "Samples": 100, "Raw results": [632, 630, 632, 632, 642, 638, 616, 605, 599, 599, 599, 597, 596, 598, 595, 595, 595, 598, 599, 608, 608, 608, 607, 608, 605, 608, 604, 604, 605, 606, 605, 608, 604, 604, 605, 608, 607, 596, 595, 597, 599, 595, 597, 595, 598, 597, 596, 599, 596, 604, 604, 608, 607, 608, 605, 607, 608, 608, 604, 606, 604, 604, 598, 596, 598, 597, 599, 599, 598, 595, 597, 598, 596, 599, 607, 606, 605, 605, 606, 604, 606, 605, 607, 608, 608, 608, 605, 598, 595, 596, 595, 597, 595, 597, 597, 597, 599, 597, 599, 608]}, {"Type": "With context switch (early processing)", "Min": 0, "Max": 0, "Mean": 603.0, "Stddev": 9.2529001732517155, "Variance": 84.760000000000005, "Mode": 0.0, "Median": 0.0, "1st quantile": 0.0, "3rd quantile": 0.0, "Samples": 100, "Raw results": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}]}, 

```

kent-mcleod:
> Edit: I think that "writes to ICC_DIR_EL1 must be ignored" means that the hardware has to ignore writes to ICC_DIR_EL1 without an active corresponding interrupt in the distributor. So it's just a question of whether our platforms have implementations that generate system errors in this case.
> 
>     * [ ]  add sel4test that multiple acks are allowed by the platform.

So I think that the kernel has to track whether the controller is expecting a ICC_DIR_EL1 write for each IRQ to avoid generation of serrors. This would require a tracking bit for each IRQ to check this. And multiple calls to seL4_IrqHandler_Ack would only produce a single write to ICC_DIR_EL1

lsf37:
> So I think that the kernel has to track whether the controller is expecting a ICC_DIR_EL1 write for each IRQ to avoid generation of serrors. This would require a tracking bit for each IRQ to check this. And multiple calls to seL4_IrqHandler_Ack would only produce a single write to ICC_DIR_EL1

I think that would work, yes.

kent-mcleod:
I added a commit adding the bitmap. 
It changes the injection cycles to: "Mean": 610.16999999999996, "Stddev": 10.215357818184117

One issue with this is that smp interrupt injection will contend on a cache line for SPIs that are in the same cache line.
(But IRQ injection is behind the single kernel lock at the moment anyway)

kent-mcleod:
> I added a commit adding the bitmap.

I'm actually leaning towards taking it back out. I haven't seen any example of multiple deactivate calls generating any serror and can't find anything in gic-500/gic-600 reference manuals either.

Indanz:
> I'm actually leaning towards taking it back out. I haven't seen any example of multiple deactivate calls generating any serror and can't find anything in gic-500/gic-600 reference manuals either.

Agreed, the [Arm website](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/ICC-DIR-EL1--Interrupt-Controller-Deactivate-Interrupt-Register?lang=en) says:

```
Accessing ICC_DIR_EL1

There are two cases when writing to [ICC_DIR_EL1](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/ICC-DIR-EL1--Interrupt-Controller-Deactivate-Interrupt-Register?lang=en) that were UNPREDICTABLE for a corresponding GICv2 write to [GICC_DIR](https://developer.arm.com/documentation/ddi0601/2024-12/External-Registers/GICC-DIR--CPU-Interface-Deactivate-Interrupt-Register?lang=en):

    When EOImode == 0. GICv3 implementations must ignore such writes. In systems supporting system error generation, an implementation might generate an SEI.
    When EOImode == 1 but no EOI has been issued. The interrupt will be deactivated by the Distributor, however the active priority in the CPU interface for the interrupt will remain set (because no EOI was issued).
```

Indanz:
> Without the patch: on imx8mq-evk "Mean": 1224.6600000000001, "Stddev": 11.649771316089073, goes to "Mean": 605.09000000000003, "Stddev": 9.7706011311815644,

Okay, that means we get double interrupts all the time on that board, so we really need to do something, either #1399, or this PR or something else.

kent-mcleod:
> Edit: To clarify, we don't get the spurious interrupt warning because the GIC hasn't cleared it yet by the time we handle the new IRQ. That's my theory anyway.

I think you're right. Because the current code completely handles the interrupt and then disables it before returning to user level, for a level triggered interrupt, like a timer timeout, the controller will just deliver the interrupt again and the kernel will process it and deliver it to user level. (So maybe the bit for tracking whether an interrupt is in progress would be useful for reporting this too...) 
The rarer case is when the interrupt is signalled to the CPU and disabled at the same time so the CPU reads irqInvalid starting the irq handling. 

I should measure how long the RWP polling takes on the imx8mq I guess

Indanz:
>  (So maybe the bit for tracking whether an interrupt is in progress would be useful for reporting this too...) 

https://github.com/seL4/seL4/pull/1399#issuecomment-2671226137

> The rarer case is when the interrupt is signalled to the CPU and disabled at the same time so the CPU reads irqInvalid starting the irq handling.

Seems to depend on the hardware, looking at the max numbers for the IRQ user test for @midnightveil, they didn't get any double interrupts, only sporadic ones once.

> I should measure how long the RWP polling takes on the imx8mq I guess

I strongly suspect the tqma board has the same double interrupt bug. The IRQ performance was horrible, but in line of expectations when compared to another imx platform. That might have been an imx8mq...

kent-mcleod:
@midnightveil thanks so much for running all three configurations together! It's as I expected that the ratios are 1x: 2x: 0.5x. The 1x:0.5x could be explained by doubling up the interrupt delivery, but it's a super coincidence if the 1x:2x caused by RWP polling is close to double

kent-mcleod:
@lsf37 this is ready to merge now. So balls in your court re verification

lsf37:
> @lsf37 this is ready to merge now. So balls in your court re verification

👍 Have added it to my verification queue.

lsf37:
Rebased and verification PR is up at https://github.com/seL4/l4v/pull/862

corlewis (approved):
~~Looks good to me!~~

~~How do you feel about the new `if` technique now that you've put it into practice? I guess it only makes sense for relatively simple config options like this but is it a solution to part of our config matrix problem?~~

Whoops,  this was meant to be a review of the connected l4v PR.

src/object/interrupt.c
```diff
@@ -218,8 +229,11 @@ void handleInterrupt(irq_t irq)
 #endif
         }
 #ifndef CONFIG_ARCH_RISCV
```
axel-h, line 231: Looking at the code, this seems a bit confusing. Why do we keep the `#ifndef CONFIG_ARCH_RISCV` here when the code block contains something that is clearly only for ARM GICv3?

src/object/interrupt.c
```diff
@@ -145,12 +145,23 @@ void invokeIRQHandler_AckIRQ(irq_t irq)
...
     }
 #endif
-    maskInterrupt(false, irq);
-#endif
+    if (config_set(CONFIG_ARM_GIC_V3_SUPPORT)) {
+        deactivateInterrupt(irq);
+    } else {
+        maskInterrupt(false, irq);
+    }
+#endif /* CONFIG_ARCH_RISCV */
```
axel-h, line 164: This comment it also a bit confusing, as the block above is actually the `!CONFIG_ARCH_RISCV` part. Actually, it seems `invokeIRQHandler_AckIRQ` has become a very architectuere specific function now.

include/machine/interrupt.h
```diff
@@ -97,6 +97,20 @@ static inline void maskInterrupt(bool_t disable, irq_t irq);
  */
 static inline void ackInterrupt(irq_t irq);
 
+/**
+ * Deactivates the interrupt
+ *
+ * When the interrupt controller supports delegating the interrupt to a lower
+ * privilege level, this function can be called to signal the completion of
+ * interrupt processing so that the interrupt state machine can be moved out of
+ * the active state.
```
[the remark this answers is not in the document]
Indanz: (This can be done in a future PR together with other cleanups though.)

include/arch/arm/arch/machine/gic_v3.h
```diff
@@ -306,14 +306,35 @@ static inline void maskInterrupt(bool_t disable, irq_t irq)
...
+
+    // If the IRQ is not going to user level then we need to deactivate it too.
+    if (unlikely(hw_irq > maxIRQ) ||
+        intStateIRQTable[IRQT_TO_IDX(irq)] != IRQSignal) {
+        /* There needs to be an isb() to ensure completion of the system
+         * register write in ackInterrupt
+         */
+        isb();
+        deactivateInterrupt(irq);
+    }
```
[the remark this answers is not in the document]
kent-mcleod: > I don't think the maxIRQ check is necessary: `handleInterrupt` will mask the interrupt before acking it because it doesn't want to receive it again, so there is no need to call `deactiveInterrupt` here, as that is contrary to the intention.

Well the check is needed regardless to bounds check the memory access to intStateIRQTable. And then the question is whether to deactivate or not.  I think deactivating seems like the better choice because it completes the irq state machine for the interrupt. Masking just disables the IRQ at the distributor but it doesn't reset any existing IRQ state.
 
Deactivating here keeps the behavior the same as before the change.  Before this change, both interrupts delivered to user level and interrupts > maxIRQ would call `maskInterrupt`, and then `ackInterrupt` and ackInterupt would write to the EOI register which would priority drop and deactivate the interrupt.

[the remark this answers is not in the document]
kent-mcleod: > `VPPIEvent()` also still does `maskInterrupt(true, irq)`, it probably shouldn't with GICv3.

It has to deactivate and mask because the interrupts are virtualized. When the kernel context switches a vcpu thread, the vtimer hardware may move into a level triggered pending state and if the hw_irq is still active, it may never be delivered.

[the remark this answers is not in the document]
kent-mcleod: > "We model interrupt deactivation as unmasking for the purposes of the interrupt oracle". So the proofs will think the IRQ is unmasked after the deactiveInterrupt, while in reality it isn't.

Well deactivate doesn't produce the same outcome as calling mask(true) and then mask(false).

src/object/interrupt.c
kent-mcleod: I mainly try to make my kernel changes that are introducing new functionality have as little change to the AST of the code so that the proof changes for the new feature is minimized.

The irq layers of the kernel are starting to get a lot of stuff like this. It seems like it would be nicer to start cleaning it up, but maybe it'd be better to wait until more changes have come through for improving IRQ virtualization on aarch64 virtualization systems.

include/arch/arm/arch/machine/gic_v3.h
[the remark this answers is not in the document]
Indanz: Thanks for the clarification, some good points there.

To me it looks like the GIC maintenance interrupt and VPPI are similar, however, neither call maskInterrupt under all circumstances, so it seems that the code can keep getting the interrupt till the vcpu gets scheduled again in some circumstances? That sounds like a bug, I think the kernel should do the right thing at vcpu switch/enable/disable to prevent this (and perhaps it already does and that's why we don't see the bug happening).

Seems like a small change to call `deactiveInterrupt` when switching to a different vcpu, assuming it's harmless to call if it's not currently active. (Would need to set the timer to the far future, call deactivate and only then restore the new timer registers to avoid spurious timer interrupts.)

My main concern with doing the mask is that we might be maintaining the IRQ bug causing spurious or double interrupts and it also increases differences in the IRQ handling. At least the timer VPPI could be considered performance sensitive. But all in all it doesn't look that straightforward, so perhaps something for a future PR.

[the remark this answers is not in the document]
kent-mcleod: > To me it looks like the GIC maintenance interrupt and VPPI are similar, however, neither call maskInterrupt under all circumstances, so it seems that the code can keep getting the interrupt till the vcpu gets scheduled again in some circumstances? That sounds like a bug, I think the kernel should do the right thing at vcpu switch/enable/disable to prevent this (and perhaps it already does and that's why we don't see the bug happening).

Yea it relies on the GIC maintenance and vtimer interrupts being individually managed properly when seL4 switches the VCPUs it also switches away the interrupt condition: 
- For the vtimer it does mask the interrupts in vcpu_disable and when the vppi is delivered to user level, 
- and for gic maintenance the interrupt trigger is processed when the interrupt is received, and when the VCPU is context switched the list registers are unloaded. A few years ago I remember I worked out that the spurious gic maintenance interrupt that can occur when a VCPU is disabled before the interrupt is taken by the kernel can be ignored because the source is always level triggered and the interrupt state is encoded in the state of the list registers that are saved/restored.

I'm still not convinced it would be better to not call deactivate here. Are you concerned with the cycle cost of doing this here?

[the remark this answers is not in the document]
kent-mcleod: > My main concern with doing the mask is that we might be maintaining the IRQ bug causing spurious or double interrupts and it also increases differences in the IRQ handling. At least the timer VPPI could be considered performance sensitive. But all in all it doesn't look that straightforward, so perhaps something for a future PR.

Regarding VPPI, because it's a per-processor interrupt, the mask operation is performed by the redistributor which is assigned to the core. I'm not sure what the relative cost of masking is compared to masking at the distributor, but it is probably more expensive than not masking if we can avoid it. We don't need to set the timer to the far future because we can just disable the interrupt enable in the control register. If we want to avoid masking and unmasking I think we'd instead need to manually save and restore the interrupt state in the redistributor when context switching the vcpu which could be reasonable.

Thinking about this more, in the cases where we currently mask due to the interrupt not being expected, not-deactivating may be a better response than masking. Masking may not be something that the kernel has authority to do for example if there are multiple OS kernels running and sharing the interrupts of the system. 
My only concern with skipping deactivating is a worry that it could trigger weird behavior where the interrupt controller stops delivering new interrupts if too many are active and not deactivated. EG This can happen with a virtual gicv3 if the virtual driver keeps list registers allocated to a virq until it's been deactivated.

So in summary I think I've changed my mind - maybe it actually would be better to not mask and not deactivate in the cases where we want to block future interrupts. But it'd probably need a bunch of testing. And shouldn't block this PR. Currently I've intended to keep the behavior the same as before except for IrqSignal interrupts that are delivered to userlevel.

lsf37:
Aright, I

include/arch/arm/arch/machine/gic_v3.h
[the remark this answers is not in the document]
Indanz: > for gic maintenance the interrupt trigger is processed when the interrupt is received, and when the VCPU is context switched the list registers are unloaded. 

That's true for races where the vcpu isn't active any more, but there is also an MCS `thread_state_get_tcbQueued` check where this isn't the case (same for VPPIevent).

> We don't need to set the timer to the far future because we can just disable the interrupt enable in the control register. 

That would work too.

> My only concern with skipping deactivating is a worry that it could trigger weird behavior where the interrupt controller stops delivering new interrupts if too many are active and not deactivated. EG This can happen with a virtual gicv3 if the virtual driver keeps list registers allocated to a virq until it's been deactivated.

I think the virtual IRQs are fully independent from physical interrupt delivery. If you run out of LRs, you need to keep track of things in software. The hardware isn't going to stop delivering interrupts as far as I know.

It seems one reason to have split EOI mode is to make this more efficient for virtual machine, so it would be weird to not use it that way in seL4. As seL4's normal use case also looks more like VM handling, where IRQ handling is delegated, split EOI mode makes a lot of sense for both.

From the [Linux patch](https://patchwork.kernel.org/project/linux-arm-kernel/patch/1439454526-1185-2-git-send-email-marc.zyngier@arm.com/) it seems like the guest can write to `ICC_EOIR1_EL1` directly themselves, but not sure if that's true with seL4.

include/machine/interrupt.h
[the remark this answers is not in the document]
Indanz: I must say the Arm documentation is very sloppy in its wording here, not making clear distinctions between PE priority and individual IRQ priorities. It does say "All interrupts, when acknowledged, modify the Active Priorities Registers." but I can't find the details of how exactly.

So a write to `ICC_EOIR0_EL1` does a priority drop, which means drop the priority of the core back to what it was before the ` ICC_IAR1_EL1` read, but more importantly, it also marks the IRQ as handled so the IRQ won't cause another interrupt (basic ACK behaviour).

All-in-all the second part is more important than the priority drop itself, so calling it priority drop without mentioning the ACK state is quite misleading overall. The priority drop of the core is just an Arm interrupt handling detail which isn't very interesting as we don't use priorities, it's the ack action that acts on the IRQ itself that is important as it changes per-IRQ state in the interrupt controller.

PR #1184, opened 2024-02-01
cmake: Support multiple overlay files

kent-mcleod:
Allow multiple overlay files to be specified. This supports custom tooling support to add additional memory reserve regions to a platform.

EG:
```
cmake -DKernelCustomDTSOverlay=../src/plat/imx6/overlay-nitrogen6sx.dts .
-- Found GCC with prefix arm-linux-gnueabi-
-- Appending /tmp/tmp.5Kv410Y6vE/kernel/src/plat/imx6/overlay-nitrogen6sx.dts overlay
-- /tmp/tmp.5Kv410Y6vE/kernel/build/gen_headers/plat/machine/devices_gen.h is out of date. Regenerating from DTB...


cmake -DKernelCustomDTSOverlay="../src/plat/imx6/overlay-nitrogen6sx.dts;../src/plat/imx6/overlay-wandq.dts" .
-- Found GCC with prefix arm-linux-gnueabi-
-- Appending /tmp/tmp.5Kv410Y6vE/kernel/src/plat/imx6/overlay-nitrogen6sx.dts overlay
-- Appending /tmp/tmp.5Kv410Y6vE/kernel/src/plat/imx6/overlay-wandq.dts overlay
-- /tmp/tmp.5Kv410Y6vE/kernel/build/gen_headers/plat/machine/devices_gen.h is out of date. Regenerating from DTB...


```

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -124,15 +124,19 @@ if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
     set(config_schema "${CMAKE_CURRENT_SOURCE_DIR}/tools/hardware_schema.yml")
     set(
         KernelCustomDTSOverlay ""
-        CACHE FILEPATH "Provide an additional overlay to append to the selected KernelPlatform's \
+        CACHE
+            STRING
+            "Provide an additional list of overlays to append to the selected KernelPlatform's \
         device tree during build time"
     )
     if(NOT "${KernelCustomDTSOverlay}" STREQUAL "")
-        if(NOT EXISTS ${KernelCustomDTSOverlay})
-            message(FATAL_ERROR "Can't open external overlay file '${KernelCustomDTSOverlay}'!")
-        endif()
-        list(APPEND KernelDTSList "${KernelCustomDTSOverlay}")
-        message(STATUS "Using ${KernelCustomDTSOverlay} overlay")
+        foreach(dts_entry IN ITEMS ${KernelCustomDTSOverlay})
+            if(NOT EXISTS ${dts_entry})
+                message(FATAL_ERROR "Can't open external overlay file '${dts_entry}'!")
+            endif()
+            list(APPEND KernelDTSList "${dts_entry}")
+            message(STATUS "Appending ${dts_entry} overlay")
+        endforeach()
     endif()
 
     find_program(DTC_TOOL dtc)


Indanz (approved):
Good idea.

lsf37 (approved):
Looks good. Still to fix up the style failure, though.

Indanz:
The style failure is nonsensical though, it suggests:
```diff
diff --git a/config.cmake b/config.cmake
index e33b8721..3fc6ee6c 100644
--- a/config.cmake
+++ b/config.cmake
@@ -124,7 +124,9 @@ if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
     set(config_schema "${CMAKE_CURRENT_SOURCE_DIR}/tools/hardware_schema.yml")
     set(
         KernelCustomDTSOverlay ""
-        CACHE STRING "Provide an additional list of overlays to append to the selected KernelPlatform's \
+        CACHE
+            STRING
+            "Provide an additional list of overlays to append to the selected KernelPlatform's \
         device tree during build time"
     )
     if(NOT "${KernelCustomDTSOverlay}" STREQUAL "")
```

axel-h:
> The style failure is nonsensical though

Unfortunately, we have to accept this for now. Otherwise this issue will pop up again for every future change of this file, that is also annoying. It's not the only case where the tool's rules seem to lack something. Maybe this gets fixed in future versions (see also https://github.com/seL4/ci-actions/issues/329)

config.cmake
```diff
@@ -124,15 +124,17 @@ if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
...
+        CACHE STRING "Provide an additional list of overlays to append to the selected KernelPlatform's \
         device tree during build time"
     )
     if(NOT "${KernelCustomDTSOverlay}" STREQUAL "")
-        if(NOT EXISTS ${KernelCustomDTSOverlay})
-            message(FATAL_ERROR "Can't open external overlay file '${KernelCustomDTSOverlay}'!")
-        endif()
-        list(APPEND KernelDTSList "${KernelCustomDTSOverlay}")
-        message(STATUS "Using ${KernelCustomDTSOverlay} overlay")
+        foreach(dts_entry IN ITEMS ${KernelCustomDTSOverlay})
```
axel-h, line 131: In this case, can't we drop the `if(NOT "${KernelCustomDTSOverlay}" STREQUAL "")` and just iterate? Iterating over an empty variable would just do nothing then.

lsf37:
The cmake style checks are made by cmake-format and are generally sensible. It has a few situations (like this one) where it produces output that is not ideal, but the TSC decision on that is that it is much preferred to have that than constant discussion about what is better.

config.cmake
kent-mcleod: The if deals with the case that the variable can be left unassigned. The inner foreach block relies on the variable having an assigned value to dereference.

PR #1188, opened 2024-02-07
libsel4: Eliminate unnamed enums

nspin:
Unnamed enums don't play nicely with [bindgen](https://github.com/rust-lang/rust-bindgen).

For example, they necessitate this [unstable hack](https://github.com/seL4/rust-sel4/blob/8b0e968cdf69309d599bb40f060962bc83254036/crates/sel4/sys/src/c.rs#L11-L15) in the `sel4-sys` crate.


kent-mcleod (approved):

Indanz (approved):

lsf37 (approved):
Happy with it, and has no impact on verification. (The two proof failures are unrelated and a bug in the corresponding GitHub action)

```Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

PR #1193, opened 2024-02-09
CI: apply concurrency on the whole workflow

axel-h:
Follow-up on https://github.com/seL4/seL4/pull/1178#issuecomment-1918081647
It simplifies things if we apply the concurrency setting for the whole workflow, instead of doing it on a fine grained job/matrix level. This aligns thing with what we do in the other workflows.

### DIFF
--- a/.github/workflows/proof.yml
+++ b/.github/workflows/proof.yml
@@ -8,6 +8,19 @@ on:
   pull_request_target:
     types: [labeled, synchronize]
 
+# To reduce the load (especiually on the AWS instances) we cancel any older runs
+# of this workflow for the current PR. Such runs exist, if there were new pushes
+# to the PR's branch without waiting for the workflow to finish. As a side
+# effect, pushing new commits now becomes a convenient way to cancel all the
+# older runs, e.g. if they are stuck and would only be stopped by the timeout
+# eventually.
+# Note that we could do the concurrency handling at a finer level and wrap just
+# the actual runs on the AWS instances. But there is no gain in this level of
+# detail. This might change if additional jobs or steps are added here.
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.number }}
+  cancel-in-progress: true
+
 jobs:
   cproof:
     name: C Proofs
@@ -33,8 +46,6 @@ jobs:
           - arch: RISCV64
             features: MCS
             session: CRefine
-    # test only most recent push to PR:
-    concurrency: seL4-PR-C-proofs-pr-${{ github.event.number }}-idx-${{ strategy.job-index }}
     steps:
     - name: Proofs
       uses: seL4/ci-actions/aws-proofs@master


.github/workflows/proof.yml
```diff
@@ -8,6 +8,19 @@ on:
   pull_request_target:
     types: [labeled, synchronize]
```
axel-h, line 9: Is `opened`, `reopened` not used on purpose?

.github/workflows/proof.yml
```diff
@@ -8,6 +8,19 @@ on:
...
+# to the PR's branch without waiting for the workflow to finish. As a side
+# effect, pushing new commits now becomes a convenient way to cancel all the
+# older runs, e.g. if they are stuck and would only be stopped by the timeout
+# eventually.
+# Note that we could do the concurrency handling at a finer level and wrap just
+# the actual runs on the AWS instances. But there is no gain in this level of
+# detail. This might change if additional jobs or steps are added here.
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.number }}
+  cancel-in-progress: true
```
axel-h, line 22: Turn out there is a pitfall here. The workflow is always triggered and the concurrency kicks in. However, the actual check if this does anything is done in the job. So even if it turns out that the job does not nothing because the conditions are not set, the concurrency-killer has acted already. This has the side effect that any label change can kill a run - even if if this label does something different. What I've seen during my tests is that this problem affects `sel4bench-pr.yml` and `sel4test-hw.yml` also. Github is missing some features here to better control this, so there are two options:
- go back to a more fine grained concurrency group setup.
- use callable workflows to only have one mater workflow be triggered on pull_request_target, have it do the checks and then call all worker woorflows `proof.yml`, `sel4bench-pr.yml` and `sel4test-hw.yml` that have very reduces condition checks.

.github/workflows/proof.yml
lsf37: I think we might actually want to remove at least `opened` from all of them, because when it's opened it's not going to have any of the labels on it, and I don't think we want to automatically re-trigger if it's only reopened without any other action. I just forgot about this again after I did the proof test workflow.

Removing `opened` also removes it from the list of checks initially, i.e. currently the hw tests always show up, but as cancelled. If they are not triggered on `open`, I don't think they show up in the list until the label is activated the first time.

.github/workflows/proof.yml
lsf37: Good catch, that will happen when the group is on the workflow level, yes.

I don't think callable workflows change anything here -- in fact, I think they make it slightly worse, because if we put it in the caller it's just one concurrency group for all of them and it still gets triggered for any event specified in `on` before any of the other checks run and will cancel everything on any event, in particular `labeled`. So that's not going to fix it. I think we do have to put them on the job level.

That would mean for `proofs`, we don't change anything, and in the factored-out workflows in `ci-actions`, we move the concurrency groups inside the job level.

lsf37: Have now updated seL4/ci-actions#332 to move it to the job level there for the workflows that are triggered by labeling actions.

PR #1194, opened 2024-02-09
CI: cancel older concurrent PR runs

axel-h:
It turned out to be quite convenient on my forks, that pushes to PRs kill any older runs, so this PR enables this feature for the other workflows also. Even if they are not running very long at their core, many of them do pull docker images or do other setup stepts. This can cause them ending up in a queue and effectively block runners unnecessarily.
In my tests the concurrency check worked fine to just kill runs from PRs, but not the pushes to master for deployment. Picking a group name with a really unique ID is a bit of a hack, but unfortunately there seems no explicit concurrency property for conditional usage.
Note that I renamed the workflow "C Parser" to "C-Parser" to avoid spaces in the concurrency group name, as it uses `github.workflow`. I'm not sure if this is really necessary, but this has been done in the other workflows also to play safe. An alternative way could be using `github.workflow_sha` instead everywhere, so the workflow names can be anything.

### DIFF
--- a/.github/workflows/compilation-checks.yml
+++ b/.github/workflows/compilation-checks.yml
@@ -12,6 +12,13 @@ on:
       - master
   pull_request:
 
+# To reduce the load we cancel any older runs of this workflow for the current
+# PR. For deployment to the master branch, the workflow will run on each push,
+# and the `run_id` parameter will prevent cancellation.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.number) || format('run-{0}', github.run_id) }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   standalone_kernel:
     name: kernel


lsf37:
I don't think we should do this, at least not for this many of the workflows. These are cheap and quick runs and it makes the workflows unnecessarily complicated when I'm trying to remove as much complexity as possible from the workflow files in each of the repos.

We could think about adding it to the compilation workflow, because that is a whole matrix of jobs and cancelling that might be useful when fewer runners are available. For `manual.yml`, we could think about a better trigger, i.e. work with more `path`/`ignore_path` (noting that it does run doxygen on at least some part of the code) so that it doesn't get started in the first place when it's not necessary.

Definitely not for pr/push/xmllint, they are very quick compared to everything else and some of these might actually be useful to see on previous commits in a PR to see when something got fixed. C Parser is usually done before I can get there to cancel it manually in my fork, so I'd probably also leave it.

We should also not replicate that huge comment everywhere and the explanation of the ternary hack. Maybe we can put a reference explanation somewhere in `ci-actions` (new `docs` directory?) that explains these things and can be linked to. 
(Including the explanation of who to avoid cancel on push, which is kinda neat).

.github/workflows/compilation-checks.yml
```diff
@@ -12,6 +12,17 @@ on:
       - master
   pull_request:
 
+# To reduce the load we cancel any older runs of this workflow for the current
+# PR. Such runs exist, if there were new pushes to the PR's branch without
+# waiting for the workflow to finish. As a side effect, pushing new commits now
+# becomes a convenient way to cancel all the older runs, e.g. if they are stuck
+# and would only be stopped by the timeout eventually.
+# For deployment to the master branch, the workflow will run on each push, no
+# skipping or cancellation will practically happen.
```
lsf37, line 21:
```suggestion
# To reduce the load we cancel any older runs of this workflow for the current
# PR. For deployment to the master branch, the workflow will run on each push, and
# the `run_id` parameter will prevent cancellation.
```

lsf37 (commented):
With the adjusted comment I think this would be fine to merge now.

.github/workflows/compilation-checks.yml
axel-h: done

PR #1196, opened 2024-02-10
github: factor out common workflows

lsf37:
We can now reuse workflow definitions, which means we can factor them out into the ci-actions repo and re-use over multiple other repositories.

Needs seL4/ci-actions#332 merged first.

The plan is to use this on the seL4 repo first for a while and then deploy to the rest when we are sure that everything works as we want it to.


lsf37:
(This needs https://github.com/seL4/ci-actions/pull/332 merged before the required actions can trigger, because they point to `master` where they don't exist yet)

.github/workflows/sel4test-hw.yml
```diff
@@ -1,113 +1,26 @@
...
-        with:
-          platform: ${{ matrix.platform }}
-          compiler: ${{ matrix.compiler }}
-          mode: ${{ matrix.mode }}
-          index: $${{ strategy.job-index }}
-        env:
-          HW_SSH: ${{ secrets.HW_SSH }}
+  sel4test:
+    name: seL4Test
+    uses: seL4/ci-actions/.github/workflows/sel4test-hw.yml@master
```
axel-h, line 20: I wonder if we should keep the label filtering here, as this give a better control when the called workflow runs at all. Also, we could pass it parameters that allow controlling the behavior:
```
    if: ${{ ((github.event.action == 'labeled') &&
             (github.event.label.name == 'hw-bench')) ||
            ((github.event.action != 'labeled') &&
             contains(github.event.pull_request.labels.*.name, 'hw-bench')) }}
    uses: axel-h/ci-actions/.github/workflows/sel4bench-hw.yml@patch-axel-7
    with:
      context: pr-${{ github.event.number }}
      labels: join(github.event.pull_request.*.name, ', ')
```
Then the called workflow could simply use globla concurrency

lsf37: No, the whole point is to get all of this out of this repo and use one central workflow in as many other repos as possible. If we have to make the central workflow a bit more complex, that is completely fine. Updating 20 repos each time we want to make a change is not fine, that's why those workflows are currently missing in most repos.

axel-h (approved):
Looks good, nice.

.github/workflows/sel4test-hw.yml
axel-h, line 1: Maybe rename the file also to clarify this is a generic hw test: `sel4test-hw.yml' -> 'test-hw.yml'

.github/workflows/sel4test-deploy.yml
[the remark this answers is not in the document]
axel-h: ah no, ignore that.

.github/workflows/sel4test-hw.yml
```diff
@@ -1,113 +1,26 @@
...
-# effect, pushing new commits now becomes a convenient way to cancel all the
-# older runs, e.g. if they are stuck and would only be stopped by the timeout
-# eventually.
-# Note that we could do the concurrency handling at a finer level, and only wrap
-# the actual run on the hardware. But there seems not much gain in letting the
-# older builds run, as these are usually obsolete with new pushes also.
-concurrency:
-  group: ${{ github.workflow }}-pr-${{ github.event.number }}
-  cancel-in-progress: true
+    types: [synchronize, labeled]
```
axel-h, line 15: Was it intentional no drop `opened` and `reopened` here?
```suggestion
    types: [opened, reopened, synchronize, labeled]
```

lsf37: Yes, the idea was that for `opened`, the test can't trigger because there is no label yet, and for `reopened` that it first should be updated (merge from master or otherwise triggering `synchronize`) before it makes sense to run hardware tests again.

I might be wrong about the trigger at `opened`, because it's possible that there will be only one event if the label is added while opening the PR. If that is the case and is annoying, I'd be happy to add that one back in.

lsf37:
I'll be merging this today to fix the label triggers. Unfortunately this will mean that we will have to manually re-run checks on PRs because they will change their names and, so the "new" required checks won't have run yet.

PR #1200, opened 2024-02-16
vcpu: guard against NULL ptr dereference

### LINKED ISSUE #1199: Potential bug/vulnerability
Is it possible for user-space (maliciously or not) to trigger a CPU fault due to a NULL pointer access in the kernel?

It seems that we only need a vCPU cap to inject a vIRQ. However if the thread that injects the vIRQ is not running on the same pCPU of the target vCPU then an IPI is sent to the pCPU owner of the vCPU thread.

`vcpu->vcpuTCB` this pointer might be `NULL` if the user space injected a vIRQ to a vCPU that has no TCB associated. This is not the normal behavior of correct code but user-space should **not be allowed** to provoke this or am I missing something?

```
exception_t invokeVCPUInjectIRQ(vcpu_t *vcpu, unsigned long index, virq_t virq)
{
    if (likely(ARCH_NODE_STATE(armHSCurVCPU) == vcpu)) {
        set_gic_vcpu_ctrl_lr(index, virq);
#ifdef ENABLE_SMP_SUPPORT
    } else if (vcpu->vcpuTCB->tcbAffinity != getCurrentCPUIndex()) {
        doRemoteOp3Arg(IpiRemoteCall_VCPUInjectInterrupt, (word_t)vcpu, index, virq.words[0],      vcpu->vcpuTCB->tcbAffinity);
#endif /* CONFIG_ENABLE_SMP */
    } else {
        vcpu->vgic.lr[index] = virq;
    }

    return EXCEPTION_NONE;
}
```

lsf37:
The vcpu is not guaranteed to be associated with a TCB at this point, so access to vcpuTCB must be guarded.

Fixes #1199

Test with: https://github.com/seL4/sel4test/pull/114

- [x]  update CHANGES

### DIFF
--- a/src/arch/arm/object/vcpu.c
+++ b/src/arch/arm/object/vcpu.c
@@ -378,8 +378,10 @@ exception_t invokeVCPUInjectIRQ(vcpu_t *vcpu, unsigned long index, virq_t virq)
     if (likely(ARCH_NODE_STATE(armHSCurVCPU) == vcpu)) {
         set_gic_vcpu_ctrl_lr(index, virq);
 #ifdef ENABLE_SMP_SUPPORT
-    } else if (vcpu->vcpuTCB->tcbAffinity != getCurrentCPUIndex()) {
-        doRemoteOp3Arg(IpiRemoteCall_VCPUInjectInterrupt, (word_t)vcpu, index, virq.words[0],      vcpu->vcpuTCB->tcbAffinity);
+    } else if (vcpu->vcpuTCB != NULL && vcpu->vcpuTCB->tcbAffinity != getCurrentCPUIndex()) {
+        doRemoteOp3Arg(IpiRemoteCall_VCPUInjectInterrupt,
+                       (word_t)vcpu, index, virq.words[0],
+                       vcpu->vcpuTCB->tcbAffinity);
 #endif /* CONFIG_ENABLE_SMP */
     } else {
         vcpu->vgic.lr[index] = virq;
--- a/CHANGES
+++ b/CHANGES
@@ -26,6 +26,24 @@ Upcoming release: BREAKING
 
 ## Changes
 
+### Security-relevant Changes
+
+* Fixed a kernel-crashing NULL pointer dereference when injecting an IRQ for a non-associated VCPU on SMP
+  configurations. This can be triggered from user-level by any thread that has access to or can create non-associated
+  VCPU objects. While HYP+SMP is not a verified configuration and is not thoroughly tested, it is generally assumed to
+  be working. If you are using this configuration, it is strongly recommended to upgrade.
+
+  * Affected configurations: only unverified HYP+SMP configurations on Arm platforms are affected.
+  * Affected versions: seL4 versions 12.0.0 and 12.1.0.
+  * Exploitability: Any thread that can create or that has access to an unassociated VCPU can cause the crash. In static
+    systems, only the system initialiser thread can create VCPUs and the standard capDL system initialiser will not
+    trigger the issue. VMMs could have the authority to dissociate an existing VCPU from a TCB if they have both
+    capabilities. That is, a malicious VMM could cause a crash, but generally VMMs are trusted, albeit not verified
+    code. Guest VMs generally do not have sufficient authority to exploit this vulnerability.
+  * Severity: Critical. This crashes the entire system.
+
+### Other Changes
+
 * Added support for the ARM Cortex A55
 * Added support for the ODroid C4
 * Added support for the Avnet MaaXBoard


lsf37:
Credits to @JorgeMVP for reporting this bug!

Indanz (approved):

lsf37:
We should probably let people know about this, at least in the CHANGES file.

lsf37:
(removing the hw-test label, it has confirmed that the fix passes the new test and that the test fails before)

JorgeMVP:
Maybe a better fix would be to check if the vCPU is associated with TCB and if not return illegal operation to the user space. 

Injecting vIRQs in a vCPU that has no TCB is a weird operation and could be an indication that something is not going well in the user space.

@lsf37 what do you think?

Indanz:
> Injecting vIRQs in a vCPU that has no TCB is a weird operation and could be an indication that something is not going well in the user space.

Won't the IRQ be delivered when the VCPU is associated with a TCB later on?

If not, I agree giving an error would make more sense. But if it is delivered later, artificially disallowing it seems wrong.

lsf37:
The current fix makes the behaviour consistent with the unicore case.

I'd have to track down if/when the IRQ is injected in this case (I'd hope it would be), but changing that behaviour should at least be a separate discussion to this fix.

src/arch/arm/object/vcpu.c
```diff
@@ -378,8 +378,10 @@ exception_t invokeVCPUInjectIRQ(vcpu_t *vcpu, unsigned long index, virq_t virq)
     if (likely(ARCH_NODE_STATE(armHSCurVCPU) == vcpu)) {
         set_gic_vcpu_ctrl_lr(index, virq);
 #ifdef ENABLE_SMP_SUPPORT
-    } else if (vcpu->vcpuTCB->tcbAffinity != getCurrentCPUIndex()) {
-        doRemoteOp3Arg(IpiRemoteCall_VCPUInjectInterrupt, (word_t)vcpu, index, virq.words[0],      vcpu->vcpuTCB->tcbAffinity);
+    } else if (vcpu->vcpuTCB != NULL && vcpu->vcpuTCB->tcbAffinity != getCurrentCPUIndex()) {
```
kent-mcleod, line 381: Doesn't this branch only need to be taken if the vcpu is active on another core? If the vcpu is inactive on the other core then the final `else` branch should be sufficient?

kent-mcleod: If you have a multicore guest, then this is the path you need to implement IPIs for the guest.

lsf37: If you're guaranteed that the active VCPU on the other core doesn't change while you're doing that check (I guess that is true since we're inside the lock), then yes, we could add another conjunct to test if it's active on that other core. That would avoid an IPI in those cases.

JorgeMVP: This would delay the vIRQ delivery since we cannot access the physical LR of the host pCPU. The vIRQ would only be injected on the next vCPU entry. The time depends on the number of threads running in the target pCPU and their time slice.

And would also make the removal of big kernel lock even more challenging. Maybe you guys don't care about this.

So far I have been following this protocol, only the host pCPU can change LRs of a vCPU.

I think IPI is important here.

kent-mcleod: > Edit: I think I got your point now, only if it is a active vCPU in the target we send the IPI, otherwise we write in the shadow registers.

Yea, I mean in the case where the ipi handler doesn't access the physical LRs and just writes to memory.

Indanz: Actually, I don't think the code makes much sense for ARM where the VMM task calling `seL4_ARM_VCPU_InjectIRQ()` is a different TCB than the VM itself. I don't think `ARCH_NODE_STATE(armHSCurVCPU) == vcpu` can ever be true except if the VM is injecting IRQs into itself, which is very unlikely.

it seems the whole function can be reduced to just always doing `vcpu->vgic.lr[index] = virq;`? What am I missing?

> This would delay the vIRQ delivery since we cannot access the physical LR of the host pCPU. The vIRQ would only be injected on the next vCPU entry.

I suspect this always happens now, except for cross-core IRQ injection. If the VMM is running, the VM isn't, and it is up to the VMM to schedule the VM so it can receive the IRQ.

kent-mcleod: > it seems the whole function can be reduced to just always doing `vcpu->vgic.lr[index] = virq;`? What am I missing?

When a thread with an associated VCPU switches to a thread without an associated VCPU, the VCPU is disabled, but it's context is left in system registers. So armHSCurVCPU would typically always refer to the last VCPU that was run (unless the object was deleted).

lsf37: `ARCH_NODE_STATE(armHSCurVCPU) == vcpu` can be true when the VMM itself doesn't have a VCPU associated, because VCPU switching is done lazily. In that case `armHSVCPUActive` would be false.

Indanz: I think I mixed up TCB with tcbVCPU when reading `vcpu_switch()`, you're right of course.

PR #1202, opened 2024-02-23
Fix Odroid-C4 MAX_IRQ setting

Ivan-Velickovic:
The S905X3 manual (Revision 02) specifies the highest IRQ to be 255 in section 7.10.2 of the manual.

This issue was encountered when trying to use the PCIe device on the platform

### DIFF
--- a/src/plat/odroidc4/config.cmake
+++ b/src/plat/odroidc4/config.cmake
@@ -15,9 +15,10 @@ if(KernelPlatformOdroidc4)
     config_set(KernelARMPlatform ARM_PLAT odroidc4)
     set(KernelArmMachFeatureModifiers "+crc" CACHE INTERNAL "")
     list(APPEND KernelDTSList "tools/dts/odroidc4.dts" "src/plat/odroidc4/overlay-odroidc4.dts")
+    # MAX_IRQ is based on the section 7.10.2 of the S905X3 SoC manual
     declare_default_headers(
         TIMER_FREQUENCY 24000000
-        MAX_IRQ 250
+        MAX_IRQ 255
         NUM_PPI 32
         TIMER drivers/timer/arm_generic.h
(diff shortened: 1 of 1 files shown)


Indanz (approved):
The manual says "There are 224 interrupt sources in the chip. `224 + 32 = 256`.

Indanz:
The style check is ridiculous and it seems sel4.systems is down. I don't have the power to merge it anyway.

Indanz:
Fix the style issue by removing the comment, it's documented in the commit message already.

src/plat/odroidc4/config.cmake
```diff
@@ -17,7 +17,8 @@ if(KernelPlatformOdroidc4)
     list(APPEND KernelDTSList "tools/dts/odroidc4.dts" "src/plat/odroidc4/overlay-odroidc4.dts")
     declare_default_headers(
         TIMER_FREQUENCY 24000000
-        MAX_IRQ 250
+        # MAX_IRQ is based on the section 7.10.2 of the S905X3 SoC manual
+        MAX_IRQ 255
```
axel-h, line 21: For what it's worth, this will pass the style check.
```suggestion
+        MAX_IRQ
+            255 # based on section 7.10.2 of the S905X3 SoC manual
```

axel-h:
> The S905X3 manual (Revision 02) specifies the highest IRQ to be 255 in section 7.10.2 of the manual.
> This issue was encountered when trying to use the PCIe device on the platform

Now, what was the issue actually?

Indanz:
> Now, what was the issue actually?

`MAX_IRQ` was 250, which means some PCIE interrupts weren't useable because they're higher than that.

src/plat/odroidc4/config.cmake
Indanz: That looks pretty horrible too.

Indanz (approved):

PR #1205, opened 2024-02-26
Python: fix errors found by type checker

axel-h:
Fix errors found by type checker while working on https://github.com/seL4/seL4/pull/797

### DIFF
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -27,12 +27,12 @@ def get_memory_regions(tree: FdtParser):
 
 def merge_memory_regions(regions: Set[Region]) -> Set[Region]:
     ''' Check all region and merge adjacent ones '''
-    all_regions = [dict(idx=idx, region=region, right_adj=None, left_adj=None)
-                   for (idx, region) in enumerate(regions)]
+    all_regions = [dict(region=region, right_adj=None, left_adj=None)
+                   for region in regions]
 
     # Find all right contiguous regions
-    for dreg in all_regions:
-        for dnreg in all_regions[dreg['idx']+1:]:
+    for idx, dreg in enumerate(all_regions):
+        for dnreg in all_regions[idx+1:]:
             if dreg['region'].owner == dnreg['region'].owner:
                 if dnreg['region'].base == dreg['region'].base + dreg['region'].size:
                     dreg['right_adj'] = dnreg


Indanz (commented):
The commit comments can be improved with explanations for the emtpy ones and fixed typos.

tools/hardware/memory.py
```diff
@@ -30,14 +30,22 @@ def __str__(self):
             self.size)
 
     def __eq__(self, other):
-        return self.base == other.base and self.size == other.size
+        if isinstance(other, type(self)):
+            return self.base == other.base and self.size == other.size
+        else:
+            # duck typing is not supported
+            return NotImplemented
```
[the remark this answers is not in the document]
axel-h: I am following the python rules here. Looks  odd, but this is indeed what this function shall return to comply.

tools/hardware/outputs/c_header.py
```diff
@@ -189,7 +189,7 @@ def create_c_header_file(config, kernel_irqs: List, kernel_macros: Dict,
                          kernel_regions: List, physBase: int, physical_memory,
                          outputStream):
 
-    jinja_env = jinja2.Environment(loader=jinja2.BaseLoader, trim_blocks=True,
+    jinja_env = jinja2.Environment(loader=jinja2.BaseLoader(), trim_blocks=True,
```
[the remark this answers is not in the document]
axel-h: Good point, I just fixed the warning here. Explicitly setting this parameter might come from history following recommendations from older versions. Will have a look into this.

tools/hardware/fdt.py
```diff
@@ -91,7 +91,10 @@ def get_devices_list(self, prop) -> List[WrappedNode]:
         for path in paths:
             if path[0] != '/':
                 path = self.lookup_alias(path)
-            ret.append(self.get_path(path))
+            node = self.get_path(path)
+            if node is not None:
+                ret.append(node)
+
```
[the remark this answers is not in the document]
axel-h: The static type annotation checker reported this based on the signatures. In this PR I left out addressing the semantics also.

tools/hardware/fdt.py
```diff
@@ -92,8 +92,9 @@ def get_devices_list(self, prop) -> List[WrappedNode]:
             if path[0] != '/':
                 path = self.lookup_alias(path)
             node = self.get_path(path)
-            if node is not None:
-                ret.append(node)
+            if node is None:
+                raise KeyError(f"could not resolve path in choosen node: '{path}'")
```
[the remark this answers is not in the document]
axel-h: fixed

tools/hardware/memory.py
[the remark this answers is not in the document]
axel-h: To clarify this a bit more, returning `NotImplemented` here will raises an error if somebody tries to compare things that can't really be compared. This usually indicates something is semantically broken in the script and should better be fixed. Returning `False` would hide this.

tools/hardware/utils/cpu.py
```diff
@@ -16,6 +16,8 @@
 def get_cpus(tree: FdtParser) -> List[WrappedNode]:
     ' Return a list of all the CPUs described in this device tree. '
     cpus_node = tree.get_path('/cpus')
+    if cpus_node is None:
+        return []
```
lsf37, line 20: How about this idiom?

```suggestion
    cpus_node = tree.get_path('/cpus') or []
```

tools/hardware/memory.py
```diff
@@ -17,10 +17,6 @@ def __init__(self, base: int, size: int, owner: 'WrappedNode' = None):
         self.size = size
         self.owner = owner
 
-    @staticmethod
-    def clone(other):
-        return Region(other.base, other.size)
-
```
lsf37, line 23: Why is it safe to remove `clone()`?

tools/hardware/fdt.py
```diff
@@ -82,27 +82,31 @@ def get_elfloader_devices(self) -> List[WrappedNode]:
...
+                raise KeyError(f"could not resolve path in chosen node: '{path}'")
+            ret.append(node)
+
         return ret
 
     def get_boot_cpu(self) -> int:
         ''' Returns phandle of the cpu node specified by the seL4,boot-cpu property '''
         chosen = self.get_path('/chosen')
-        if not chosen.has_prop('seL4,boot-cpu'):
+        if (chosen is None) or (not chosen.has_prop('seL4,boot-cpu')):
```
lsf37, line 104:
```suggestion
        if (not chosen) or (not chosen.has_prop('seL4,boot-cpu')):
```

tools/hardware/fdt.py
```diff
@@ -82,27 +82,31 @@ def get_elfloader_devices(self) -> List[WrappedNode]:
     def get_devices_list(self, prop) -> List[WrappedNode]:
         ''' Returns a list of devices that are used by the kernel '''
         chosen = self.get_path('/chosen')
-        if not chosen.has_prop(prop):
+        if (chosen is None) or (not chosen.has_prop(prop)):
```
lsf37, line 85:
```suggestion
        if (not chosen) or (not chosen.has_prop(prop)):
```

tools/hardware/device.py
```diff
@@ -135,12 +135,9 @@ def get_interrupt_affinities(self) -> List[int]:
...
     def visit(self, visitor: Any):
         ''' Visit this node and all its children '''
-        ret = [visitor(self)]
-        if ret[0] is None:
-            ret = []
+        visitor(self)
         for child in self.children.values():
-            ret += child.visit(visitor)
-        return ret
+            child.visit(visitor)
```
lsf37, line 140: This changes the semantics of `visit`, we're not returning the list of visited nodes any more. Why is this safe? Do we know all callers of this function?

tools/hardware/utils/memory.py
```diff
@@ -27,12 +27,12 @@ def visitor(node: WrappedNode):
...
-    all_regions = [dict(idx=idx, region=region, right_adj=None, left_adj=None)
-                   for (idx, region) in enumerate(regions)]
+    all_regions = [dict(region=region, right_adj=None, left_adj=None)
+                   for region in regions]
 
     # Find all right contiguous regions
-    for dreg in all_regions:
-        for dnreg in all_regions[dreg['idx']+1:]:
+    for idx, dreg in enumerate(all_regions):
+        for dnreg in all_regions[idx+1:]:
```
lsf37, line 35: Are we guaranteed that this preserves the order in which the regions are visited? In the first line, we're making a new dict, and it doesn't seem to me like that will necessarily have the same order of elements in `enumerate` as the old `regions` dict.

It's not clear to me if that is semantically important, but if we're changing behaviour based on linter suggestions, I'd like to understand properly what the effects are first.

What was the linter complaining about in the original code?

lsf37 (commented):
Modulo the comments above, I'm fine with merging this independently of the type annotation/CI linter discussion.

tools/hardware/utils/memory.py
lsf37: Just saw the commit message and I get it now, it's a list of `dicts`, not a `dict`, so the order is the same. Happy with that.

tools/hardware/device.py
lsf37: I don't understand the commit message on this one: the previous implementation was perfectly sane in constructing the list of return values of the visited elements, leaving out `None` values. I'm not sure why that was done, but there is nothing specifically wrong with it.

What was the linter complaint for it?

tools/hardware/memory.py
lsf37: Ok, found the commit message, sorry. How sure are we that this is only used in this repo? (I assume that is the scope used to find out that it wasn't called).

I agree that the clone functionality looks a bit strange -- `clone` potentially was added initially and not updated when `owner` was added, which would also need to support `clone` properly.

tools/hardware/fdt.py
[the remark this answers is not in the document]
lsf37: With the explicit error that it ends up with eventually, I think this is fine.

tools/hardware/utils/cpu.py
```diff
@@ -17,6 +17,9 @@ def get_cpus(tree: FdtParser) -> List[WrappedNode]:
     ' Return a list of all the CPUs described in this device tree. '
     cpus_node = tree.get_path('/cpus')
 
+    if cpus_node is None:
+        return []
+
```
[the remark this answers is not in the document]
lsf37: Agreed. If we don't want to change `get_path` directly, using the idiom `cpus_node = tree.get_path('/cpus') or []` would make it clearer what we want there.

tools/hardware/device.py
axel-h: I think I got this pattern wrong in the end, when trying to define the return type properly. The return type of the custom `visitor()` function is simply an `object` and `visit()` returns a list of objects. The linter should accept this - will check again if there was some issues here.

PR #1206, opened 2024-02-26
AArch64: Hardware Debug API port

alwin-joshy:
This pull request relates to the porting of the hardware debug API to aarch64. The API exposes hardware debugging features to userspace and enables the development of debugging tools on top of seL4.

The first four commits are to do with splitting arm/machine/debug.c, which contains the current implementation of the hardware debug API for aarch32, into mode-dependent and mode-independent files while retaining their git history.

[aarch32: change dbg register fields](https://github.com/seL4/seL4/commit/d53b0591a1eef63fad616ec22bf79c73530f4b7c) renames some fields in the aarch32 implementation of the hardware debug API to enable cleaner reuse of code and makes fields better correspond with their names in the ARMv7 and ARMv8 manuals.

[hw debug api: aarch32 - modify single stepping checks](https://github.com/seL4/seL4/commit/d5851b0a84a1df0134b7cc73379894ac584f8e73) improves a check on aarch32 related to the configuration of single stepping (more details can be found in the commit message).

[hw debug api: aarch64 port](https://github.com/seL4/seL4/commit/f94b53065012f235216f1e1bb43393dc3a2c7677) finally ports the hardware debug API to aarch64.

The code has been tested to work still pass seL4test on ARMv7 boards (sabre, odroidxu4) as well as on AARCH64 (odroidc4/odroidc2/qemu-arm-virt). I have not been able to confirm that it works with AARCH32 (armv8 32-bit), because I have been unable to run a 32-bit seL4test image (even unmodified) on an ARMv8 board.


lsf37 (commented):
There seem to be a number of changes to code that is visible to verification. See the diffs in the failed preprocess tests -- they show what changes the verified configurations see. Whitespace and comment changes there can be ignored, but there should be no new functions declarations or new instructions visible, which they are.

If this PR only concerns the hardware debug API, there should ideally be no such changes visible. If they are, this will need funding for re-verification.

include/api/syscall.h
```diff
@@ -34,6 +34,7 @@ exception_t handleInterruptEntry(void);
 exception_t handleUnknownSyscall(word_t w);
 exception_t handleUserLevelFault(word_t w_a, word_t w_b);
 exception_t handleVMFaultEvent(vm_fault_type_t vm_faultType);
+exception_t handleDebugFaultEvent(word_t esr);
```
Indanz, line 37: Leftover that should be removed too.

include/arch/arm/arch/64/mode/machine/debug.h
```diff
@@ -6,4 +6,141 @@
...
+ * This could be handling a single-stepping exception, or a breakpoint or
+ * watchpoint.
+ */
+seL4_Fault_t handleUserLevelDebugException(word_t esr, word_t fault_vaddr);
+bool_t isDebugFault(word_t esr);
+
+#define MAKE_DBGBVR(num) "DBGBVR" #num "_EL1"
+#define MAKE_DBGBCR(num) "DBGBCR" #num "_EL1"
+#define MAKE_DBGWVR(num) "DBGWVR" #num "_EL1"
+#define MAKE_DBGWCR(num) "DBGWCR" #num "_EL1"
```
Indanz, line 24: Shouldn't this be `_EL2` when HYP is enabled?

include/arch/arm/arch/64/mode/machine/debug.h
```diff
@@ -6,4 +6,141 @@
...
+
+/** Generates read functions for the CP14 control and value registers.
+ */
+#define DEBUG_GENERATE_READ_FN(_name, _reg)                                    \
```
Indanz, line 28: I hate these macros and don't understand why they're needed. I honestly think writing all the used cases out would be less code and clearer than this indirect macro stuff.

That 32-bit debug code has this is no excuse to copy it, but if you do, at least share the macros: The only difference is MRS versus MRC, easily defined away...

include/arch/arm/arch/64/mode/machine/registerset.h
```diff
@@ -232,6 +233,33 @@ extern const register_t msgRegisters[];
 extern const register_t frameRegisters[];
 extern const register_t gpRegisters[];
 
+#ifdef ARM_BASE_CP14_SAVE_AND_RESTORE
+typedef struct debug_register_pair {
+    word_t cr, vr;
+} debug_register_pair_t;
+
+// @alwin: Copied from armv7, necessary?
```
Indanz, line 241: This was introduced by commit f99f6790abe9c0d8b27a0aee0b6b29a0dc73c211 for 32-bit code. I would remove it, unless the variation of the value of `seL4_NumExclusiveWatchpoints` would change the TCB size. That is, the `seL4_TCBBits` define should always be correct, no matter which CPU variant is used.

include/arch/arm/arch/machine/debug.h
```diff
@@ -206,12 +282,15 @@ static inline syscall_error_t Arch_decodeSetBreakpoint(tcb_t *t,
         }
     }
 
+#ifdef CONFIG_ARCH_AARCH32
     if (size == 8 && !byte8WatchpointsSupported()) {
```
Indanz, line 286: Cleaner to define a `byte8WatchpointsSupported()` macro for aarch64 that returns true.

include/arch/arm/arch/machine/debug.h
```diff
@@ -248,21 +327,21 @@ static inline syscall_error_t Arch_decodeUnsetBreakpoint(tcb_t *t, uint16_t bp_n
         return ret;
     }
 
+#ifdef CONFIG_ARCH_AARCH32
```
Indanz, line 330: This ifdef can go now that you have the aarch64 `Arch_breakpointIsSingleStepping()` returning false.

include/arch/arm/arch/machine/debug.h
```diff
@@ -147,6 +221,7 @@ static inline syscall_error_t Arch_decodeConfigureSingleStepping(tcb_t *t,
             return ret;
         }
 
+#ifdef CONFIG_ARCH_AARCH32
```
Indanz, line 224: At least add a comment why the below doesn't apply to aarch64.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
```diff
@@ -174,7 +174,11 @@ typedef enum {
 #define seL4_LargePageBits 21
 #define seL4_HugePageBits 30
 #define seL4_SlotBits 5
+#if defined(CONFIG_HARDWARE_DEBUG_API) || (defined(CONFIG_ARM_HYPERVISOR_SUPPORT) && defined(CONFIG_ARM_HYP_ENABLE_VCPU_CP14_SAVE_AND_RESTORE))
```
Indanz, line 177: Checking for CONFIG_HARDWARE_DEBUG_API and CONFIG_ARM_HYP_ENABLE_VCPU_CP14_SAVE_AND_RESTORE seems enough?

src/arch/arm/c_traps.c
```diff
@@ -222,7 +231,16 @@ VISIBLE NORETURN void c_handle_vcpu_fault(word_t hsr)
...
+
+#ifdef CONFIG_HARDWARE_DEBUG_API
+    word_t esr = getESR();
+    if (isDebugFault(esr)) {
+        handleDebugFaultEvent(esr);
+    } else
+#endif
+    {
+        handleVCPUFault(hsr);
+    }
```
Indanz, line 243: Currently `armv_handleVCPUFault()` already checks for FPU faults and user level faults, it would be more consistent to have the debug check there too.

src/arch/arm/machine/debug.c
```diff
@@ -941,8 +431,10 @@ static int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
              * range, which means it's not guaranteed to match the aligned value
              * that was programmed into the address register.
              */
-            align_mask = convertArchToSize(dbg_bcr_get_byteAddressSelect(bcr));
+#ifdef CONFIG_ARCH_AARCH32
```
Indanz, line 434: If you put the ifdef around the comment above, it would be clearer that it only applies to 32-bit. I assume aarch64 has support for unaligned values in the address register?

src/arch/arm/machine/debug.c
```diff
@@ -1290,6 +608,12 @@ void restore_user_debug_context(tcb_t *target_thread)
      *
      * So we don't need to execute ISB here because we're about to RFE.
      */
+
+#ifdef CONFIG_ARCH_AARCH64
+    aarch64_restore_user_debug_context(target_thread);
+#endif /* CONFIG_ARCH_ARCH64 */
+
+
```
Indanz, line 616: Stray empty lines. Also, perhaps move this chunk to above the ISB comment.

Indanz:
The DCO check failing is bogus: I just redid @alwin-joshy's commits in a way that preserves history, so the sign-off should be in his name.

lsf37:
> The DCO check failing is bogus: I just redid @alwin-joshy's commits in a way that preserves history, so the sign-off should be in his name.

Happy to set this to "pass". It'll have to be re-set when force-pushed, but that's fine.

include/arch/arm/arch/64/mode/machine/debug.h
alwin-joshy: I'm pretty sure those registers don't exist.

include/arch/arm/arch/64/mode/machine/debug.h
alwin-joshy: Yeah I don't really like the macros either, but I thought that duplicating them would at least make it slightly less horrible than adding another level of pre-processing in there with extra ifdefs and more macros.

include/arch/arm/arch/64/mode/machine/registerset.h
alwin-joshy: Ah, I think I understand now. Since we bump the TCB size up as a result of ARM_BASE_CP14_SAVE_AND_RESTORE, we need to make sure the actual size of the TCB matches what we say. 

I think the ARMv8 manual says that all CPUs that implement the spec need to have a minimum of 4 hardware breakpoint and watchpoint registers, so I just need to check if with this many, the TCB size needs to be bumped up, and if so, I can get rid of the padding.

alwin-joshy:
> > I have not been able to confirm that it works with AARCH32 (armv8 32-bit), because I have been unable to run a 32-bit seL4test image (even unmodified) on an ARMv8 board.
> 
> The file [platforms.yml](https://github.com/seL4/ci-actions/blob/63d8c710be07bcb1fe230bccc779a73941efb922/seL4-platforms/platforms.yml) in the [ci-actions](https://github.com/seL4/ci-actions/) repo has a list which armv8 platforms support 32bit seL4 images, e.g. the [`imx8mm-evk`](https://github.com/seL4/ci-actions/blob/63d8c710be07bcb1fe230bccc779a73941efb922/seL4-platforms/platforms.yml#L144-L150):
> 
> ```yaml
>   IMX8MM_EVK:
>     arch: arm
>     modes: [32, 64]
>     smp: [64]
>     platform: imx8mm-evk
>     req: [imx8mm]
>     march: armv8a
> ```

```
../init-build.sh -DPLATFORM=imx8mm-evk -DAARCH32=ON -DHardwareDebugAPI=ON
```

passes seL4test.

include/arch/arm/arch/64/mode/machine/debug.h
Indanz: I should have checked. About EL2, the ARM manual says:

[To enable OS debugging, set MDCR_EL2.TDE to 1, which enables the exceptions from EL0 and EL1 to be handled at EL2. This means that the debugger is installed at EL2.](https://developer.arm.com/documentation/102120/0101/Enabling-various-self-hosted-debug-models).

alwin-joshy: Yep, that's done in the `setHDCRTrapDebugExceptionState` function and I've tested that the breakpoint and single-stepping tests still pass when running the kernel with ARM_HYP enabled.

src/arch/arm/64/machine/debug.c
```diff
@@ -34,6 +34,11 @@
 #define OSDLR_LOCK (BIT(0))
 #define OSLAR_LOCK (BIT(0))
 
+bool_t byte8WatchpointsSupported(void)
```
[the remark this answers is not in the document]
alwin-joshy: I think it can't be inlined because the 32-bit version relies on `dbg`, which is a static struct.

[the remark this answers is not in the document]
Indanz: Considering all code gets concatenated anyway, making it non-static and declaring it as usual will fix that with no downsides. But perhaps that's more something for a future clean-up.

lsf37:
@alwin-joshy What's the status of this PR? It's marked draft, has conflicts, and is failing a bunch of tests. Is work proceeding on this or should it be closed?

Supporting the HW debug API in AArch64 would be nice, and since it's just the existing API (as far as I can see), it does not need an RFC. When this is cleaned up and ready, I don't think there would be any barriers to merging it.

alwin-joshy:
Hey @lsf37,

This PR is complete in terms of functionality and I'd like to get it merged, just moved on to some other things. You had said before that

> There seem to be a number of changes to code that is visible to verification. See the diffs in the failed preprocess tests -- they show what changes the verified configurations see. Whitespace and comment changes there can be ignored, but there should be no new functions declarations or new instructions visible, which they are.

I was able to reduce the diff but IIRC there were still a few small visible changes. I can get the branch back up to date and let the tests run again so we can assess what changes are necessary.

lsf37:
It looks like something broke the preprocess test in this change (for all architectures). The log shows

```
REF1=seL4@6b6bb125018cdbb5d0f7050a024da007431582fe
REF2=seL4@112734cae3e4789f9e9b69097690d4896b9698e5
L4V_ARCH=ARM
WORKING...
kernel.mk:118: *** multiple target patterns.  Stop.
```

`kernel.mk` is a Makefile include in l4v that invokes the kernel build system.

Any idea what could have lead to this?

alwin-joshy:
Not sure, I just rebased it on seL4/master. There were a bunch of conflicts so maybe something went wrong, I'll have a look

alwin-joshy:
Yep, that seems to have fixed it

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -624,11 +624,23 @@ static inline void armv_vcpu_boot_init(void)
...
 {
     vcpu_save_reg_range(vcpu, seL4_VCPUReg_TTBR0, seL4_VCPUReg_SPSR_EL1);
+
+#ifdef ARM_HYP_CP14_SAVE_AND_RESTORE_VCPU_THREADS
+    /* This is done when we are asked to save and restore the CP14 debug context
+     * of VCPU threads; the register context is saved into the underlying TCB.
+     */
+    saveAllBreakpointState(vcpu->vcpuTCB);
+#endif
+    isb();
```
lsf37, line 643: This `isb()` is visible outside the `#ifdef` -- was that intentional?

lsf37 (commented):
Apart from the `isb()` this now does not look like it has any verification breaking changes.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
alwin-joshy: It might not be necessary on ARMv8, as the original code didn't have it. This was probably a mistake from when I was adapting the equivalent ARMv7 code, which does have an isb() outside of the #ifdef in this function.

alwin-joshy: Although I'm not sure why it is needed on armv7. Do you have any idea?

lsf37: No, it doesn't really make sense to me. It does make sense (maybe) after breakpoint stuff, but not after just saving registers.

alwin-joshy: Yeah, I can kind of see why you might want to do something like this after enabling/disabling breakpoints, but that's not what saveAllBreakpointState does anyway. I will remove the isb() from the ARMv8 version but leave the ARMv7 version as it is for now.

lsf37: I've been looking into the git history a bit, and the `isb()` has been there after or inside `vcpu_save` for armv7 since at least the open source release. There is a `dsb()` somewhere at the beginning of the `vcpu_save` chain, which makes sense to me, but the `isb()` still does not (especially after doing the save).

lsf37:
Running a proof test to make sure I'm right about the remaining preprocess changes not having any impact. If they pass, the only thing remaining is squashing the PR feedback commit.

alwin-joshy:
> Running a proof test to make sure I'm right about the remaining preprocess changes not having any impact. If they pass, the only thing remaining is squashing the PR feedback commit.

Sounds good, I'll squash after you've given the thumbs up.

lsf37:
Ok, the proofs have passed (the RISCV-MCS failure can be ignored). After squashing, we can merge.

lsf37 (approved):

PR #1208, opened 2024-02-28
Fix DDR region for Odroid-C4

Ivan-Velickovic:
The Odroid-C4 is supposed to have 4GB of DDR memory.

According to the SoC manual (S905X3 Revision 02) the DDR region goes from 0x0 to 0xF57FFFFF in Table 7-1.

### DIFF
--- a/tools/dts/odroidc4.dts
+++ b/tools/dts/odroidc4.dts
@@ -2945,7 +2945,7 @@
 
 	memory@0 {
 		device_type = "memory";
-		reg = <0x00 0x00 0x00 0x40000000>;
+		reg = <0x00 0x00 0x00 0xf5800000>;
 	};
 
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
I think when the Odroid-C4 support was added what happened was the default Linux Device Tree for the Odroid-C4 was decompiled and used for seL4.

wom-bat (commented):
Second field is size, not last address.
So perhaps should be 0xf5800000

Ivan-Velickovic:
> Second field is size, not last address. So perhaps should be 0xf5800000

Whoops, that was a silly mistake.

lsf37 (approved):
Looks good from my side.

wom-bat (approved):
Looks good now.

Issue #1209, opened 2024-02-29
USER_TOP lower than expected on AArch64 platforms?


Ivan-Velickovic:
I am looking at USER_TOP in the generated configs in Microkit and I notice that all of them have the same value for USER_TOP which is 0xa0000000. This surprised me since this is quite low for all the platforms currently supported by Microkit.

I notice that if I build sel4test, e.g:
`../init-build -DPLATFORM=odroidc4 && ninja`

and look at `kernel/gen_config/kernel/gen_config.h` I see:
```c
#define CONFIG_USER_TOP  0xa0000000
```

To my understanding, USER_TOP is not something configurable by the user of seL4 and so I don't think it's something going wrong with how I am building sel4test/Microkit?

Ivan-Velickovic:
Okay if I look at `kernel_all.i` I see in `decodeARMPageTableInvocation` the value of USER_TOP after pre-processing is 0x00007fffffffffff, which makes more sense.

This means there must be mismatch between USER_TOP used in the kernel code and what is exposed to user-space.

Ivan-Velickovic:
Wonderful, it looks like that https://github.com/seL4/seL4/blob/3740c2902c3084514dd1beaaf122162e1ac36dd6/src/plat/qemu-arm-virt/config.cmake#L12 leaks out into *every* platform because of https://github.com/seL4/seL4/blob/3740c2902c3084514dd1beaaf122162e1ac36dd6/src/plat/qemu-arm-virt/config.cmake#L274.

kent-mcleod:
Looks like `UNDEF_DISABLED` needs to be added to that config declaration.

Ivan-Velickovic:
> Looks like `UNDEF_DISABLED` needs to be added to that config declaration.

Won't that remove the config from being exported?

Ivan-Velickovic:
(I would like to use USER_TOP in code/tools)

kent-mcleod:
The option you linked to is only declared for the qemu platform, so on other platforms it shouldn't be being declared as USER_TOP that the kernel uses isn't known by the build system apart from the qemu platform on aarch32: https://github.com/seL4/seL4/blob/master/libsel4/sel4_plat_include/qemu-arm-virt/sel4/plat/api/constants.h#L28

`seL4_UserTop` is supposed to be declared in libsel4.h for all platforms. So it's in a similar situation to the object size constants that are also defined in header files only.

Ivan-Velickovic:
[...]

PR #1210, opened 2024-02-29
Fix USER_TOP config option being unconditional

Ivan-Velickovic:
Closes https://github.com/seL4/seL4/issues/1209.


kent-mcleod (approved):
Thanks for fixing

PR #1211, opened 2024-02-29
style: fix cmake formatting spec

axel-h:
Looking at the changes of https://github.com/seL4/seL4/pull/1210, it seem the cmake style file needs some fixes.

### DIFF
--- a/.cmake-format.yaml
+++ b/.cmake-format.yaml
@@ -12,8 +12,6 @@ additional_commands:
       DEP: '*'
       PREFIX: '*'
   config_option:
-    flags:
-    - UNQUOTE
     kwargs:
       DEFAULT: '*'
       DEFAULT_DISABLED: '*'
@@ -21,10 +19,11 @@ additional_commands:
   config_string:
     flags:
     - UNQUOTE
+    - UNDEF_DISABLED
     kwargs:
       DEFAULT: '*'
       DEPENDS: '*'
-  cpp_file:
+  cppfile:
     kwargs:
       EXACT_NAME: '*'
       EXTRA_DEPS: '*'


Ivan-Velickovic:
If we do change the style we should also update all other uses of `UNDEF_DISABLED` then

lsf37:
Actual style changes would need TSC decisions. Happy to discuss those if there is a good reason to change something.

The description seems to indicate that it's fixing something that was previously already intended to work a certain way.

If that's the case, that would be fine to do in a PR, but it's not clear to me what the intention was previously, what is not going as expected, and what the fix now provides. @axel-h can you clarify that a bit?

axel-h:
We had some complaint about the style checker recently, as it is sometimes rather dump and enforces formatting that is not aligned with the semantics. This addresses some of the issues.

* the name "cpp_file" in the style file seem wrong, as the function is called "cppfile". Fixing the name stops the style checker from demanding a rather odd formatting in some cases.
* `config_option` has no parameter `UNQUOTE`. It might have had this parameter in the past. Removing this from the style file also now.
* `config_string` has a flag `UNDEF_DISABLED`, which was missing in the style file. That is why the style checker insists on putting this parameter into a long line rather then accepting that it on a separate line also, which is often much more readable.

axel-h:
> If we do change the style we should also update all other uses of `UNDEF_DISABLED` then

It has not been allows by the style checker before. Now it is allowed, but not enforced. But you are right, might be nice to update this then everywhere to improve readability a bit more.

lsf37:
> We had some complaint about the style checker recently, as it is sometimes rather dumb and enforces formatting that is not aligned with the semantics. This addresses some of the issues.
> 
> * the name "cpp_file" in the style file seem wrong, as the function is called "cppfile". Fixing the name stops the style checker from demanding a rather odd formatting in some cases.
> * `config_option` has no parameter `UNQUOTE`. It might have had this parameter in the past. Removing this from the style file also now.

Thanks, that makes sense and it looks like these two are just straight-up fixes of things that were always intended.

> * `config_string` has a flag `UNDEF_DISABLED`, which was missing in the style file. That is why the style checker insists on putting this parameter into a long line rather then accepting that it on a separate line also, which is often much more readable.

That also seems in line with current behaviour and would be fine from my side. @kent-mcleod any opinions on this?

> > If we do change the style we should also update all other uses of `UNDEF_DISABLED` then
> 
> It has not been allows by the style checker before. Now it is allowed, but not enforced. But you are right, might be nice to update this then everywhere to improve readability a bit more.

That would be contrary to what the usual policy is: we don't update for style globally. We (don't) do that to reduce merge conflicts and code churn. We only update when those files are touched anyway.

It looks to me like the old style would still be accepted, so there is no real need to change existing files apart from aesthetics. I don't mind making a file consistent when we touch it anyway (same as for actually required style changes), but a global change does not seem justified.

kent-mcleod (approved):

axel-h:
Removed the commit that changes the cmake files to use the improved style option. Can we merge the style file changes then?

lsf37 (approved):
Happy with that.

PR #1217, opened 2024-03-11
RISC-V: make trap_entry a function 

axel-h:
Making this a function makes debugging a bit nicer, as we have a nice function symbol available then


Indanz (approved):

axel-h:
Seems the proofs are fine with this.

kent-mcleod:
Does this produce an identical binary?

axel-h:
> Does this produce an identical binary?

The binaries are identical and the ELF dumps also show no functional differences.

PR #1218, opened 2024-03-12
riscv: add wrapper read_sscratch()

axel-h:
hide assembly code behind a wrapper function


axel-h:
Seems the proofs are fine with this.

lsf37 (approved):
No problem from the proof side, happy to remove a bit of duplication.

lsf37:
This PR broke the sel4test hardware build -- see <https://github.com/seL4/seL4/actions/runs/8263995752/job/22606633755>.

Seems to be something trivial, can you please have a look?

axel-h:
> I created #1221, hopefully that fixes it.

Thanks.

Indanz:
Okay, that created a circular dependency, so it's not a fix. I propose to revert the smp.h change instead.

PR #1219, opened 2024-03-12
Manual: Xinclude support

Indanz:
This way common error codes can be in their own file.

To use, add the `xmlns:xi="http://www.w3.org/2001/XInclude` attribute to the top level node and use:

`<xi:include href="file-to-include.xml"/>`

The content of the file will be included verbatim.

Include files must be complete XML documents, but without `<?xml version="1.0" ?>` at the top. Practically this means all nodes within the file need to be contained in one root element.

Caveat: There is no proper dependency for Xincludes files in the Makefile, so you need to do make clean after changing an included XML file!


Indanz:
The [limitation](https://stackoverflow.com/questions/8827782/python-lxml-using-xiinclude-with-multiple-xml-fragments) of having to have one root element in Xincludes practically means that common error codes need to have their own section and can't be mixed in with other error codes.

See commit 0b81757982a33db87ddc7fb1a7df70c0c99b1747 for a (bad) example usage, but it shows the syntax at least.

Indanz:
Alternative is to create our own tag, scan for that and include the content of the file verbatim into the XML ourselves. That would give maximum flexibility, but it would also be non-standard.

lsf37:
That doesn't look too bad. Does this need any additional dependencies for python that we should specify?

Indanz:
> That doesn't look too bad.

It's the hacky simple version. The full conversion from minidoc to lxml isn't too bad either, but new things keep popping up and it's not complete. Worse, it's spreading to other files and gets to the point I don't now what the code is trying to achieve, so it's not easy to rewrite/fix it for lxml. Anyway, you can see how far I got in commit 5eedde78f14e6d2efb92ce528add7a5c4dff4e6c.

> Does this need any additional dependencies for python that we should specify?

No, lxml was already used to check the DTD and BeautifulSoup uses lxml for parsing too.

Indanz:
It seems we can perhaps use `<xi:include href="file-to-include.xml" xpointer="xpointer(//error)"/>` to skip the root element, but haven't tried it out yet. If that doesn't work, the worst case is that each common error needs to be xi:included separately instead of all of them at once. Downside is that they each have to be in their own file then. It will still reduce the amount of duplicate code compared to using external entities though.

lsf37 (approved):
Happy to merge this. If we later find that it is still too annoying to use, we can still refactor again.

PR #1220, opened 2024-03-12
trivial CMake cleanup

axel-h:
- remove obsolete variable. Seems it's a merge artifact that should never have been in commit ad4ea6cd, and commit c642a398 moved handling to a python script anyway.
- add missing bracket in comment.

### DIFF
--- a/tools/helpers.cmake
+++ b/tools/helpers.cmake
@@ -297,7 +297,6 @@ endfunction(config_option)
 # configuration headers.
 macro(config_set optionname configname value)
     set(${optionname} "${value}" CACHE INTERNAL "" FORCE)
-    set(c_define "CONFIG_${configname}")
     if("${value}" STREQUAL "OFF")
         cfg_str_add_disabled(configure_string ${configname})
     else()


tools/helpers.cmake
```diff
@@ -297,7 +297,6 @@ endfunction(config_option)
 # configuration headers.
 macro(config_set optionname configname value)
     set(${optionname} "${value}" CACHE INTERNAL "" FORCE)
-    set(c_define "CONFIG_${configname}")
```
kent-mcleod, line 300: As `config_set` is a macro, a caller can use `c_define` to get name of the config in C as a return value.

kent-mcleod: (That being said, that probably wasn't the intention of the original variable assignment), but some downstream caller could be using it for this purpose by now.

axel-h: That is true, but I have doubts this is really the case anywhere. It was added by accident, as it was just necessary in some intermediate development state as helper variable. So I'd prefer to drop this to avoid building legacy. It was never documented either.
Nevertheless, if there is a use case, it might be interesting to learn about that. We publish things as JSON and YAML now that handle all this, so these should be used.

lsf37: It is true that this has only ever been available in the current devel version, it was added after the most recent release. According the GitHub search function, it is not used anywhere in the seL4 org. If there is downstream code that uses it, it at least never had any guarantee of stability. So I'd be fine with removing it.

lsf37 (approved):

axel-h:
Could this get merged?

PR #1221, opened 2024-03-13
RISC-V SMP: Fix Compile Error

Indanz:
Partially revert commit https://github.com/seL4/seL4/commit/70d1a5f792fbe1ae59a79dd5dcc413cf69447aa0.

Can't include arch/machine.h because that will cause a circular include dependency.

### DIFF
--- a/include/arch/riscv/arch/model/smp.h
+++ b/include/arch/riscv/arch/model/smp.h
@@ -51,7 +51,8 @@ static inline bool_t try_arch_atomic_exchange_rlx(void *ptr, void *new_val, void
 
 static inline CONST cpu_id_t getCurrentCPUIndex(void)
 {
-    word_t sp = read_sscratch();
+    word_t sp;
+    asm volatile("csrr %0, sscratch" : "=r"(sp));
     sp -= (word_t)kernel_stack_alloc;
     sp -= 8;
     return (sp >> CONFIG_KERNEL_STACK_BITS);


include/arch/riscv/arch/model/smp.h
```diff
@@ -51,7 +51,8 @@ static inline bool_t try_arch_atomic_exchange_rlx(void *ptr, void *new_val, void
 
 static inline CONST cpu_id_t getCurrentCPUIndex(void)
```
axel-h, line 52: What about just pulling in the prototype here:

```suggestion
/* We can't include arch/machine.h due because there is a circular include
 * dependency. Until that can be resolved, we need to define the function
 * prototype here.
 */
 
static inline word_t read_sscratch(void);
static inline CONST cpu_id_t getCurrentCPUIndex(void)
```

axel-h: We could do this also, but I don't see a real gain here compared to just using the prototype. The compiler will warn if the prototype does no longer match and we have the comment that explains things. Te long term solution would be refactoring things then to resolve the include quirk, but seem this is not trivial.

lsf37 (approved):
I'll merge this as is for now as a partial revert so the build works again, and I can bump the verification manifest to point to a consistent version.

Happy to discuss a nicer solution in a separate PR.

PR #1224, opened 2024-03-15
Arm32: Fix vcpu_hw_write_reg

Indanz:
Forgotten in commit 5bce5c0c1ed642a1690e7950e27e50561ef77811.


kent-mcleod (approved):

lsf37 (approved):
Maybe we should move the machine interface one level lower here (verification would have found this if it wasn't behind the machine interface). Something to look at on the proof side.

This is fine to merge once the proof tests pass. I don't anticipate any breakage there (because behind machine interface), but want to make sure.

PR #1226, opened 2024-03-18
Wrap docref references in <docref> tag

bbrcknl:
The seL4 API Reference points to a number of references that can only be resolved in the manual, and not in the online reference. E.g. search for "autoref" on https://docs.sel4.systems/projects/sel4/api-doc.html

The updates in this PR wrap the references in <docref> tags so that they only appear in the manual, not online. This is standard practise for the seL4 manual and API references.

### DIFF
--- a/libsel4/arch_include/x86/interfaces/sel4arch.xml
+++ b/libsel4/arch_include/x86/interfaces/sel4arch.xml
@@ -432,12 +432,12 @@ contain the mapping"/>
             <param dir="in" name="vaddr" type="seL4_Word" description="Virtual address at which to map page."/>
             <param dir="in" name="rights" type="seL4_CapRights_t">
                 <description>
-                    Rights for the mapping. Possible values for this type are given in <autoref label='sec:cap_rights'/>.
+                    Rights for the mapping. <docref>Possible values for this type are given in <autoref label='sec:cap_rights'/>.</docref>
                 </description>
             </param>
             <param dir="in" name="attr" type="seL4_X86_VMAttributes">
                 <description>
-                    VM attributes for the mapping. Possible values for this type are given in <autoref label='ch:vspace'/>.
+                    VM attributes for the mapping.<docref> Possible values for this type are given in <autoref label='ch:vspace'/>.</docref>
                 </description>
             </param>
             <error name="seL4_AlignmentError">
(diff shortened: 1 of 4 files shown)


libsel4/arch_include/x86/interfaces/sel4arch.xml
```diff
@@ -432,12 +432,12 @@ contain the mapping"/>
...
             <param dir="in" name="rights" type="seL4_CapRights_t">
                 <description>
-                    Rights for the mapping. Possible values for this type are given in <autoref label='sec:cap_rights'/>.
+                    Rights for the mapping. <docref>Possible values for this type are given in <autoref label='sec:cap_rights'/>.</docref>
                 </description>
             </param>
             <param dir="in" name="attr" type="seL4_X86_VMAttributes">
                 <description>
-                    VM attributes for the mapping. Possible values for this type are given in <autoref label='ch:vspace'/>.
+                    VM attributes for the mapping.<docref> Possible values for this type are given in <autoref label='ch:vspace'/>.</docref>
```
lsf37, line 440: (or if the whitespace creates strange output, we should change the other one a few lines above to match up with this one)
```suggestion
                    VM attributes for the mapping. <docref>Possible values for this type are given in <autoref label='ch:vspace'/>.</docref>
```

libsel4/include/interfaces/sel4.xml
```diff
@@ -1452,7 +1452,7 @@
                 If the object is a notification and it is bound to the scheduling context, unbind it.
             </brief>
             <description>
-                See <autoref label="sec:passive"/>
+                <docref>See <autoref label="sec:passive"/></docref>
             </description>
```
lsf37, line 1456: These will then be empty `<description>` tags. Have you had a look at the API docs output if that works out Ok?

lsf37 (approved):

Indanz (approved):
This will be forgotten when updating the manual in the future, except if we throw errors during build in manual/tools/parse_doxygen_xml.py. But that doesn't seems worth the hassle.

libsel4/include/interfaces/sel4.xml
bbrcknl: @lsf37 
The description gets removed in parse_doxygen.xml:
```
def generate_api_doc(self, level, member, params, ret, details, errors):
        manual_node = member.manual

        # Descriptions that just contain a document reference are removed.
        # Found by the 'DOCREF' symbol
        match_details = re.match(r'^DOCREF$', details, re.M | re.I)
        if match_details:
            details_string = ""
    ```

bbrcknl:
@Indanz @lsf37 Pls merge :-)

PR #1228, opened 2024-03-22
avoid pointer arithmetic

axel-h:


Indanz:
What is the motivation for this change, other than making the code slightly better at the cost of extra effort updating the verification?

lsf37:
The `&intStateIRQNode[idx]` construction would make the term in the proof much more annoying. This should not be merged.

axel-h:
Out of curiosity, why does this make the proof harder actually?

lsf37:
The array index operator wants the object to be an array, and the address-of operator then has to back-simulate the original pointer arithmetic just to get to what the original source said. The original is just clear and direct in the proof.

I do actually also find it much clearer in C, but that's not that important.

Issue #1083, opened 2023-08-01
Completing the manual's ARM virtualisation section


Ivan-Velickovic:
I'd be keen to have this section finished/fleshed out, but I'm only familiar with virtualising ARM guests on seL4 from a user-space perspective, which is probably not enough to do the manual section properly. This is a list of things I would put in the manual section, but am looking for things that are missing from the list:
* Explanation of each of the ARM-specific invocations (and how they relate to the faults delivered by seL4).
* VCPU starting state (so that it can be reset properly if you wanted to restart a guest/VCPU).
* GIC versions supported and that GIC is not fully virtualised by the hardware and hence needs some software virtualisation.
* Current limitations of virtualising on seL4. So far I only know of one, which is based on this comment from @kent-mcleod https://github.com/seL4/seL4_projects_libs/pull/71#issuecomment-1185062965 which says that suspending and resuming a guest requires more information from the kernel than it currently provides.
* Any differences between AArch32 and AArch64 virtualisation
* Virtualising 32-bit guests on 64-bit host and vice versa, I know it's possible on RISC-V but not sure about ARM.

Ivan-Velickovic:
Now that SMC forwarding is in the kernel, that should also be mentioned.

Indanz:
> Now that SMC forwarding is in the kernel, that should also be mentioned.

I keep saying this, but: SMC forwarding has nothing to do with virtualisation as such and forwarding is a bad name to being with, because it just makes it possible to make SMC calls from user space. So the "forwarding" bit is forwarding SMC calls upwards from the kernel to the system monitor. This should be documented separately for ARM and not in the virtualisation section, because that will be just confusing.

When a guest VM calls SMC, it traps into the VMM, so SMC calls from the guest are a way to communicate with the host. This you could document in the virtualisation section.

The VMM can implement e.g. ARM's standard PSCI calls any way it wants, a reboot could restart the VM for instance. That it wants to implement an SMC call by doing the real SMC call is more or less coincidence (and the chance of that happening increases with obscure platform specific SMC calls needed to make specific hardware work). It's fine to have a reference to the SMC forwarding section.

Ivan-Velickovic:
> When a guest VM calls SMC, it traps into the VMM, so SMC calls from the guest are a way to communicate with the host. This you could document in the virtualisation section.

That's what I meant yes.

Ivan-Velickovic:
[...]

PR #1232, opened 2024-03-26
Manual: Document ARM Virtualisation

Indanz:
This resolves issue #1083.


lsf37 (commented):
Modulo the compile error, this looks good from my side.

manual/parts/threads.tex
```diff
@@ -658,11 +658,28 @@ \section{Virtualisation}
...
 
-When a \obj{TCB} has a bound \obj{VCPU} it is allowed to have the mode portion of the
-\texttt{cpsr} register set to values other than \texttt{user}. Specifically it may have any value other than
-\texttt{hypervisor}.
+When a \obj{TCB} has a bound \obj{VCPU} it will have access to (virtualised) system registers,
+cache and TLB maintenance instructions and be able to handle some exceptions itself.
+The virtual GIC will be enabled, allowing virtual interrupt delivery.
 
-TODO: this section needs more detail
+The vitualised system registers can be modified with \apifunc{seL4\_ARM\_VCPU\_WriteRegs}{arm_vcpu_write_registers}.
```
Ivan-Velickovic, line 665:
```suggestion
The virtualised system registers can be modified with \apifunc{seL4\_ARM\_VCPU\_WriteRegs}{arm_vcpu_write_registers}.
```

Ivan-Velickovic:
Cool, thank you for doing this.

Looking back at the issue I made, some of the things mentioned probably don't belong in the manual.

What about things like vGIC version supported? Right now VMMs have emulated GICv2 and GICv3 but not GICv4. Is virtualising the ITS for GICv4 just in the VMM, or is some support from seL4 needed? Either way, I think it is not harmful to be clear about what vGIC versions are supported.

Indanz:
> Looking back at the issue I made, some of the things mentioned probably don't belong in the manual.

I must admit I wrote the documentation before re-reading your wish list. There are some items there that would be good to add.

Finding the right balance of what makes sense in the manual and what not is hard.

> What about things like vGIC version supported? Right now VMMs have emulated GICv2 and GICv3 but not GICv4. Is virtualising the ITS for GICv4 just in the VMM, or is some support from seL4 needed? Either way, I think it is not harmful to be clear about what vGIC versions are supported.

I would say virtual GIC support follows the platform GIC support. We don't specify what platforms are supported in the manual either. Furthermore, documenting which vGIC versions seL4 supports doesn't help the user. Also, this documentation can get out of date easily when the kernel gets updated.

The difference in the kernel's implementation of virtual GICv2 and GICv3 is small enough that it's not worth mentioning. AFAIK the only user visible change is the amount of list registers. When we add vGICv4 support things may change enough that it needs pointing out, but if it's all hidden away in registers and ITS programming in user space, the kernel API stays the same.

The main goal of the manual is to document what functionality seL4 provides to write a VMM and virtual GIC driver in user space. Its goal is not to document how to write a VMM or virtual GIC driver in general. Basically, it's assumed that the reader will or has read the ARM documentation and needs to know how to apply that to an seL4 system. That said, perhaps it's good to say this explicitly and give a broader overview how ARM virtualisation works and to define the scope of seL4's part. Would that be helpful?

manual/parts/threads.tex
```diff
@@ -658,11 +658,28 @@ \section{Virtualisation}
...
+with \apifunc{seL4\_ARM\_VCPU\_InjectIRQ}{arm_vcpu_inject_irq}, which provides a way to manage
+Virtual GIC List Registers, a queue of pending IRQs to be delivered to the guest. To help
+with managing the list, the Virtual GIC will send GIC maintenance interrupts as PPIs.
+
+Shared Peripheral Interrupts (SPIs) can be handled like any normal IRQs and forwarded as required.
+
+However, Virtual Private Peripheral Interrupts (PPI) are trapped and delivered as faults instead and
+need to be acknowledged with \apifunc{seL4\_ARM\_VCPU\_AckVPPI}{arm_vcpu_acknowledge_virtual_ppi_irq}.
+
+Stage 2 translation is enabled when the kernel supports virtualisation, but only for \obj{VCPUs}.
```
kent-mcleod, line 679: This sentence is a bit confusing to me. Stage 2 translation is enabled for all threads when the kernel supports virtualisation. Stage 1 translation can only be used by threads with an associated VCPU. 

But then the next sentences in this paragraph make sense.

Indanz: I keep mixing up stage 1 and 2 terminology, I'll fix it.

Ivan-Velickovic:
> Its goal is not to document how to write a VMM or virtual GIC driver in general. Basically, it's assumed that the reader will or has read the ARM documentation and needs to know how to apply that to an seL4 system

Sure, makes sense.

> That said, perhaps it's good to say this explicitly and give a broader overview how ARM virtualisation works and to define the scope of seL4's part. Would that be helpful?

I think so. I guess I was mainly looking for somewhere that describes seL4's capabilities as a hypervisor on ARM, in way that could be potentially compared to other hypervisors. This is why I thought it would be good to document things like current limitations.

Whether this belongs in the manual, I do not know.

lsf37 (approved):
This is now ready to merge from my side.

PR #1236, opened 2024-04-05
boot: Fix root cnode size compile-time assert

cmcl:
The check was not enforcing the minimum radix intended by 5fac9e8 and still allowed a radix of 4 to be specified.


Indanz (approved):

Indanz:
Gitlint complains with:

> 1: T3 Title has trailing punctuation (.): "boot: Fix root cnode size compile-time assert."

Could you change the title and force push the same commit again?

axel-h (approved):
That fix looks right, indeed.

PR #1237, opened 2024-04-08
make SMP conditional check more general

axel-h:
Check for `CONFIG_ENABLE_SMP_SUPPORT` everywhere instead of checking `CONFIG_MAX_NUM_NODES > 1`. This are the remaining places in the code where `CONFIG_MAX_NUM_NODES` was used in the check. This change now allows enabling SMP support manually with just one node also, which comes handy for running some test configurations.

### DIFF
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -304,7 +304,7 @@ static inline void invalidateByVA_I(vptr_t vaddr, paddr_t paddr)
 
 static inline void invalidate_I_PoU(void)
 {
-#if CONFIG_MAX_NUM_NODES > 1
+#ifdef CONFIG_ENABLE_SMP_SUPPORT
     asm volatile("ic ialluis");
 #else
     asm volatile("ic iallu");
--- a/libsel4/include/sel4/syscalls.h
+++ b/libsel4/include/sel4/syscalls.h
@@ -118,7 +118,7 @@ seL4_DebugCapIdentify(seL4_CPtr cap);
  */
 LIBSEL4_INLINE_FUNC void
 seL4_DebugNameThread(seL4_CPtr tcb, const char *name);
-#if CONFIG_MAX_NUM_NODES > 1 && defined CONFIG_ARCH_ARM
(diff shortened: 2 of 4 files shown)


Indanz (approved):

lsf37 (approved):
I think it makes sense to standardise on `CONFIG_ENABLE_SMP_SUPPORT` and use that consistently. @kent-mcleod, any concerns?

axel-h:
See, the hw tests failure comes from https://github.com/seL4/seL4/actions/runs/8621690507/job/23635445744?pr=1237 where some spurious interrupt that confuses the log parser.
```
 Test TRIVIAL0000 passed
  	</testcase>
  	<testcase classname="sel4test" name="TRIVIAL0<<seL4(CPU 0) [handleInterruptEntry/53 T0xffffff80bffe9400 "sel4test-driver" @442f4c]: Spurious interrupt!>>
  001">
```
Seem the other hardware job fail because they are running too long. And then there are some that got cancelled, but I don't see what causes this actually - maybe github does this because the other jobs running into a timeout, despite we have set `fail-fast: false`?

lsf37:
> Seems the other hardware job fail because they are running too long. And then there are some that got cancelled, but I don't see what causes this actually - maybe github does this because the other jobs running into a timeout, despite we have set `fail-fast: false`?

The maximum running time of any GitHub job is 6h, and they got cancelled at 5:59h, so it was probably that. The question is why the job was running that long -- that should only be possible if it was waiting for the machine lock the entire time (everything else has timeouts), but that didn't seem the case. Need to investigate a bit more.

kent-mcleod:
> This change now allows enabling SMP support manually with just one node also, which comes handy for running some test configurations.

I don't expect this to be supported by the current implementation. `CONFIG_ENABLE_SMP_SUPPORT` was added to mean `CONFIG_MAX_NUM_NODES > 1`.  So allowing `CONFIG_ENABLE_SMP_SUPPORT` to be set when `CONFIG_MAX_NUM_NODES=1` is going to break a bunch of assumptions that the SMP code makes.  The main one I expect is the assumption that there is at least 2 cores running so any IPI broadcast operation is always going to get a response.

PR #1239, opened 2024-04-08
libsel4: remove obsolete --cfile and --word-size

axel-h:
Practically, it seems the word size can be derived from `--arch` easily and that is all we need. The rest is removed to simplify things.
Seem creating the `config` file to pass the word size is a left over from the Kconfig build system and it looks quite obsolete nowadays. Removing the parameters `--cfile` and `--word-size` might break 3rd party workflows, but this could be fixed when a properly defined custom architecture passed via `--arch`.

### DIFF
--- a/libsel4/CMakeLists.txt
+++ b/libsel4/CMakeLists.txt
@@ -146,19 +146,14 @@ set(
     "${CMAKE_CURRENT_SOURCE_DIR}/arch_include/${KernelArch}/interfaces/sel4arch.xml"
     "${CMAKE_CURRENT_SOURCE_DIR}/include/interfaces/sel4.xml"
 )
+
 add_custom_command(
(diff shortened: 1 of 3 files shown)


axel-h:
The generated files are exactly the same.

Indanz:
> The generated files are exactly the same.

How can it break 3rd party workflows then?

axel-h:
When they call the pyhon tool manually and use `--cfile`/`-c` or `--word-size`/`-w`, the invocation fails as these parameters don't exist any longer. Buit since `--arch` is mandatory anyway, just deleting might be enough.

Indanz:
Quick search on Github shows that Genode, CantripOS and Kata-OS may be affected. They should be competent enough to fix any problems.

PR #1244, opened 2024-04-20
fix: invisible chars due to ANSI escape codes

### LINKED ISSUE #1243: A specific  specific colorscheme is assumed
# Problem

I just spent some time trying to debug my solution for the tutorial, when I found this: a spurious print of seL4 in green.

image

Playing with the terminal, I finally found the following, crucial information:

image

# Additional Information

For reference, my colorscheme is as following (generated using `terminal-colors -o`):

image

# Cause

Now, investigating the [output](https://github.com/seL4/microkit/files/15047830/out.txt) yields an interesting observation: the prints are styled as following: 

1. Reset style
2. Set foreground (text) to whatever the current colorscheme considers black
3. Set the font to bold
4. (put text)
5. Reset style

Obviously, ~~most~~ many people run a dark colorscheme. And in many cases, for these the background color will be equal to what the colorscheme considers black. So generally speaking, I would advice the following:

- __Never change the foreground color to either black (`30m`) or white (`37m`)__
  - Reason:  these are quite likely to collide with the background color of a given terminal
- It is of course ok to change the foreground color __if__ the background color is changed as well to a different color not equal to the default (`49m`)

Now my personal preference would be to leave the background intact, thus forcing the use of a foregroundcolor  (how about yellow, to indicate a non fatal yet significant problem?) except for black/white for the seL4/microkit generated output, or the use of __bold__ alone (which blends in well with any terminal colorscheme, thus my preference).

As is, the UX is unnecessarily poor for fellow dark-colorscheme users tinkering with the (otherwise excellent!) tutorial, or microkit in general.

wucke13:
The prior behavior would emit ANSI escape codes which set the terminal
foreground color to black bold (`[30;1m`), leaving the default
background color. On many terminals configured with a dark color scheme,
this creates black on black characters, i.e. the output is invisible.

This change moves to a better adapted output, where the output is only
marked as bold (`[1m]`), which works well on all color schemes.

Fixes #1243

### DIFF
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -103,11 +103,11 @@ static inline word_t CONST wordFromMessageInfo(seL4_MessageInfo_t mi)
 #ifdef CONFIG_COLOUR_PRINTING
 #define ANSI_RESET "\033[0m"
 #define ANSI_GREEN ANSI_RESET "\033[32m"
-#define ANSI_DARK  ANSI_RESET "\033[30;1m"
+#define ANSI_BOLD  ANSI_RESET "\033[1m"
 #else
 #define ANSI_RESET ""
 #define ANSI_GREEN ANSI_RESET ""
-#define ANSI_DARK  ANSI_RESET ""
+#define ANSI_BOLD  ANSI_RESET ""
 #endif
 
 /*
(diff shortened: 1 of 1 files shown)


lsf37:
Thanks, I have long found this black on black output annoying -- you can guess my terminal background colour ;-). @kent-mcleod do you remember any reasons why we would not want this change? (bold only, instead of setting black)

Indanz (approved):

include/api/types.h
```diff
@@ -135,8 +135,8 @@ extern struct debug_syscall_error current_debug_error;
  */
 #define userError(M, ...) \
     do {                                                                       \
-        out_error(ANSI_DARK "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
-                ANSI_DARK " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n", \
+        out_error(ANSI_BOLD "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
```
andybui01, line 138: Nice PR, completely unrelated question for whoever to answer: in addition to printing the CPU would it be a good idea to print the current domain? `NODE_STATE(ksCurThread)->tcbDomain` this can be helpful to quickly determine which domain the error is coming from.

kent-mcleod (approved):

lsf37 (approved):

PR #1247, opened 2024-05-02
arm tlb: guard TLB lockdown count

lsf37:
- `lockTLBEntry` uses the global `tlbLockCount` as input without checking bounds. This is fine, because the function is called at most 2 times on unicore platforms, but this is only apparent when checking the entire possible calling context.

  Make this bound obvious locally by doing nothing if the function is called with values of `tlbLockCount` of 2 or greater. This is safe, because TLB lockdown is a performance change only. Also add an assert for debug mode, because we want to know if calling context ever changes.

- remove TLB locking for Cortex A9. The code shared the A8 locking sequence with A9, but not all of these instructions are supported on A9. As far as we have been able to tell, they are silently ignored on the A9 platforms that seL4 supports, but it is of course better to remove them in case future platforms behave differently, which they would be allowed to do.

To properly add TLB locking for A9 in the future, the code would need to use the corresponding A9 instructions and take SMP into account for e.g. the `sabre` platforms, where `lockTLBEntry` is called 2 times *per core*, instead of 2 times overall.

These two potential issues were reported by The UK's National Cyber Security Centre (NCSC).

### DIFF
--- a/src/arch/arm/armv/armv7-a/machine_asm.S
+++ b/src/arch/arm/armv/armv7-a/machine_asm.S
@@ -7,18 +7,15 @@
 #include <config.h>
 #include <machine/assembler.h>
 
-#if defined(CONFIG_ARM_CORTEX_A15) || defined(CONFIG_ARM_CORTEX_A7)
-  /* The hardware does not support tlb locking */
-
-#else
+#if defined(CONFIG_ARM_CORTEX_A8)
 
 .code 32
 .section .text, "ax"
 .global lockTLBEntryCritical
 
-.balign (16*4)    
+.balign (16*4)
 BEGIN_FUNC(lockTLBEntryCritical)
- 
+
  /* lockTLBEntryCritical should lie within a single page so that spurious TLB walks do
  * not interfere. Aligning to a 64-byte instruction boundary will suffice, as
  * the critical section (i.e. this function) fits within 64 bytes.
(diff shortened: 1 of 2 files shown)


Indanz (approved):

lsf37:
Well, it looks like the number of times this function is called is not 4 on the sabre platform..

lsf37:
It's 8, for which the bit operations are also still safe. There is something sub-optimal going on, though: it's 8, because it locks 2 TLB entries and does that each time a core is initialised (4 cores on sabre).

Sabre is the only A9 multicore platform, and for A9 it turns out that the lockdown code is wrong anyway (doesn't do anything  as far as we can see), so I'm not going set the bound up, but will instead add another commit for removing A9 lockdown instead. I was going to do that in a separate PR, but with this, it's better if the A9 lockdown removal comes first.

Indanz (approved):
Only suggestion I have is to remove it altogether and see if it makes any difference in the benchmark results.

Indanz:
Both failing runs failed with:

>  No log to check boot failure on.

But otherwise all is well.

lsf37:
> Only suggestion I have is to remove it altogether and see if it makes any difference in the benchmark results.

It'd be fine to do this in a separate experiment, but I'd like to keep this one to the issue it is trying to fix. I don't have any data either way if TLB locking is good or bad for performance (and under which work loads), but we do know that removing it for A9 doesn't change anything, because it was never really there.

It may well be the case that it never makes sense (I doubt it, I'm sure it was added for a reason), but it may also be the case that it does make sense on larger TLBs on more recent CPUs, so I don't think the setup as such should be removed.

PR #1249, opened 2024-05-15
Add new imx8mp-evk platform

nomadeel:
This new platform is NXP Semiconductor's Evaluation Kit for the i.MX 8M Plus Applications Processor. It's from the i.MX 8M family of processors and is largely similar to the existing i.MX 8M Quad and i.MX 8M Mini platforms.

This port is not tested with the EVK, but it's tested against [Compulab's IOT-GATE-IMX8PLUS](https://www.compulab.com/products/iot-gateways/iot-gate-imx8plus-industrial-arm-iot-gateway/#overview) which is based on the same SoC but has a different memory layout. The device tree included in the port is for the EVK.

### DIFF
--- a/tools/dts/imx8mp-evk.dts
+++ b/tools/dts/imx8mp-evk.dts
@@ -1,0 +1,2668 @@
+/*
+ * Copyright Linux Kernel Team
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *
+ * This file is derived from an intermediate build stage of the
+ * Linux kernel. The licenses of all input files to this process
+ * are compatible with GPL-2.0-only.
+ */
+
+/dts-v1/;
+
+/ {
+	interrupt-parent = <0x01>;
+	#address-cells = <0x02>;
+	#size-cells = <0x02>;
+	model = "NXP i.MX8MPlus EVK board";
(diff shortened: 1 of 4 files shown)


tools/dts/imx8mp-evk.dts
```diff
@@ -0,0 +1,2668 @@
...
+			reg = <0x00 0x942f8000 0x00 0x8000>;
+			no-map;
+		};
+
+		vdev0buffer@94300000 {
+			compatible = "shared-dma-pool";
+			reg = <0x00 0x94300000 0x00 0x100000>;
+			no-map;
+		};
+	};
```
Indanz, line 302: You probably want to remove these reserved-memory regions in your overlay, as it's normal memory used in specific ways by Linux.

Indanz (approved):

tools/dts/imx8mp-evk.dts
nomadeel: Done in the latest commit.

nomadeel:
These two PRs are related to the port as well:
- https://github.com/seL4/seL4_tools/pull/201
- https://github.com/seL4/util_libs/pull/182

PR #1250, opened 2024-05-20
bitfield_gen: properly escape backslash

lsf37:
Replace `"\<"` in strings with `"\\<"`. Until recently python did not complain about this illegal escape sequence, but now it warns.

I've checked that we're not replacing anything in raw strings (`r"..."` and `r'...'`).


Indanz (approved):
Why do we need the '\' in the output though? That is, why not just remove the '\' instead?

It's probably clearer if all strings that can be made raw are, instead of all this escaping.

lsf37:
The `"` in the output is part of the Isabelle syntax, i.e. what we want to produce are strings like

```isabelle
definition x :: "some_type \<Righarrow> some_other_type" where
```

Using raw strings would be nicer in many places, but that needs actual thinking about where that is appropriate. A lot of these are templates with substitution etc.

PR #1251, opened 2024-05-21
vcpu: only trap WFx instructions from vCPUs

andybui01:
When KernelArmDisableWFIWFETraps is disabled (trapping of WFI/WFE is enabled), the kernel traps WFx instructions from both native and vCPU threads. This change makes the code in line with the config description.

In any case, there don't exist many use cases for trapping WFx from native threads vs vCPU threads. Example: kvm uses this to yield physical CPU time to other vCPUs instead of busy waiting.

### DIFF
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -14,18 +14,18 @@
 #include <drivers/timer/arm_generic.h>
 
 /* Note that the HCR_DC for ARMv8 disables S1 translation if enabled */
-#ifdef CONFIG_DISABLE_WFI_WFE_TRAPS
 /* Trap SMC and override CPSR.AIF */
-#define HCR_COMMON ( HCR_VM | HCR_RW | HCR_AMO | HCR_IMO | HCR_FMO )
-#else
-/* Trap WFI/WFE/SMC and override CPSR.AIF */
-#define HCR_COMMON ( HCR_TWI | HCR_TWE | HCR_VM | HCR_RW | HCR_AMO | HCR_IMO | HCR_FMO )
-#endif
+#define HCR_COMMON ( HCR_VM | HCR_RW | HCR_AMO | HCR_IMO | HCR_FMO | HCR_TSC)
 
 /* Allow native tasks to run at EL0, but restrict access */
 #define HCR_NATIVE ( HCR_COMMON | HCR_TGE | HCR_TVM | HCR_TTLB | HCR_DC \
                    | HCR_TAC | HCR_SWIO |  HCR_TSC )
-#define HCR_VCPU   ( HCR_COMMON | HCR_TSC)
+
+#ifdef CONFIG_DISABLE_WFI_WFE_TRAPS
+#define HCR_VCPU   ( HCR_COMMON)
+#else
+#define HCR_VCPU   ( HCR_COMMON | HCR_TWE | HCR_TWI)
+#endif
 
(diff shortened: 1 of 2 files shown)


include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -14,18 +14,18 @@
...
-#endif
 
 /* Allow native tasks to run at EL0, but restrict access */
 #define HCR_NATIVE ( HCR_COMMON | HCR_TGE | HCR_TVM | HCR_TTLB | HCR_DC \
                    | HCR_TAC | HCR_SWIO |  HCR_TSC )
+
+#ifdef CONFIG_DISABLE_WFI_WFE_TRAPS
 #define HCR_VCPU   ( HCR_COMMON | HCR_TSC)
+#else
+#define HCR_VCPU   ( HCR_COMMON | HCR_TSC | HCR_TWFX)
```
Indanz, line 27: I would add `HCR_TSC` to `HCR_COMMON` and drop the `HCR_TWFX` define. That's also what ARMv7 is doing already.

Indanz (approved):
Agree with the changes.

Before this change, WFI/WFE are normally allowed for native tasks, but with virtualisation enabled, it depended on `KernelArmDisableWFIWFETraps`. After this change, `KernelArmDisableWFIWFETraps` only affects VCPUs.

Edit: Hmm, `SCTLR_EL1_NTWI` and `SCTLR_EL1_NTWE` are only set if HYP is on. Otherwise, the only initialisation of `SCTLR_EL1` happens at src/arch/arm/64/head.S, which leaves NTWI/NTWE unchanged.

Indanz:
And according to the ARM manual, for both nTWE and nTWI (and a lot of other bits too):

_On a Warm reset, in a system where the PE resets into EL1, this field resets to an architecturally UNKNOWN value._

Perhaps seL4 should initialise the bits as it want them to be, instead of using unknown reset values.

lsf37 (commented):
I can't comment much from the verification side on this -- it's all behind the machine interface for us, i.e. we don't see this change. Which is a bit worrying in terms of complexity, but that's where the cut-off currently is.

The intention of the change seems right to me, I don't know enough about the hardware details to judge all implementation aspects.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
andybui01: > drop the HCR_TWFX define

Not sure what you meant here, I pushed some changes, do you mind checking?

Indanz: The define is there to simplify `HCR_TWE | HCR_TWI`, but overall the extra indirection makes things less clear I think.

Otherwise the changes you made are what I meant.

Indanz:
> The intention of the change seems right to me, I don't know enough about the hardware details to judge all implementation aspects.

I double checked with the ARM documentation and the changes do what's intended.

We still have a discrepancy in behaviour between HYP and non-HYP, but that's an independent issue already present. The issue of uninitialised register bits only applies to non-HYP builds, where WFE/WFI trapping is controlled by `SCTLR_EL1`. In HYP builds, `SCTLR_EL1` is set by `vcpu_enable`, `vcpu_disable` and `armv_vcpu_boot_init`.

It could very well be that seL4's HYP init values match the CPU implementation specific reset values for all supported ARM cores though.

Indanz (approved):

Indanz (approved):
Inconsistent spacing around `(` and `)`, but that's a common problem in that file.

Otherwise looks good to me.

kent-mcleod (approved):

lsf37:
It looks like this does need a verification update, because we have some of these values replicated in the spec. I'll have a look at that.

PR #1253, opened 2024-05-23
risc-v: remove special handling for restoring tp

axel-h:
I can't find a reason why it is done in such a complicated way, when the RISC-V port was added in 83ba0847.

### DIFF
--- a/include/arch/riscv/arch/fastpath/fastpath.h
+++ b/include/arch/riscv/arch/fastpath/fastpath.h
@@ -121,7 +121,8 @@ static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t m
         LOAD_S "  ra, (0*%[REGSIZE])(t0)  \n"
         LOAD_S "  sp, (1*%[REGSIZE])(t0)  \n"
         LOAD_S "  gp, (2*%[REGSIZE])(t0)  \n"
-        /* skip tp/x4, t0/x5, t1/x6, they are restored later */
+        LOAD_S "  tp, (3*%[REGSIZE])(t0)  \n"
+        /* skip t0/x5, t1/x6, they are restored later */
         LOAD_S "  t2, (6*%[REGSIZE])(t0)  \n"
         LOAD_S "  s0, (7*%[REGSIZE])(t0)  \n"
         LOAD_S "  s1, (8*%[REGSIZE])(t0)  \n"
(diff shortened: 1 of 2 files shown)


Indanz (approved):
No idea.

lsf37 (changes requested):
Let's not merge this before we have found out. Not knowing is not a good enough reason to change something.

Indanz:
> Let's not merge this before we have found out. Not knowing is not a good enough reason to change something.

Whatever it is, it warrants a comment. Only reason I can think of is that it's to work around a performance crippling quirk on a specific implementation. Another more far fetched reason is because Rust inline assembly didn't allow direct use of `tp` and that influenced @heshamelmatary (but the same is true for `gp` though).

FreeBSD](https://cgit.freebsd.org/src/tree/sys/riscv/riscv/exception.S#n143) and [OpenBSD](https://github.com/openbsd/src/blob/c7101648ccc694093c4a805ebde0f53056d1a24c/sys/arch/riscv64/riscv64/exception.S#L146) just load it too. Linux stores `tp` in `sscratch`.

kent-mcleod:
It should be safe to make this change. 

It was most likely due to IPC buffer handling from before 11.0.0 (https://docs.sel4.systems/releases/sel4/11.0.0.html). 
Before this version, a thread-local register was reserved by the kernel for storing the pointer to the thread's IPC buffer which was the `tp` register on riscv. Architectures such as aarch64 would set the register inside `Arch_switchToThread()` and then not touch the register in `restore_user_context()`. 

When the RISC-V port was up-streamed, it didn't restore the register with the other registers, but also didn't restore it in `Arch_switchToThread()` and so ended up restoring it right at the end of the restore process.

lsf37:
> It should be safe to make this change.
> [..]
> When the RISC-V port was up-streamed, it didn't restore the register with the other registers, but also didn't restore it in `Arch_switchToThread()` and so ended up restoring it right at the end of the restore process.

Thanks, that makes perfect sense. With that I'm happy with the change.

lsf37 (approved):

lsf37:
It'd be good to record that reasoning in the commit message (i.e. that `tp` was used for thread-local storage before and since that is gone, the special handling is no longer necessary). Since it's uniform with the rest, we probably don't need a comment in the source for it, if the change reasoning is documented.

axel-h:
> It'd be good to record that reasoning in the commit message (i.e. that `tp` was used for thread-local storage before and since that is gone, the special handling is no longer necessary). Since it's uniform with the rest, we probably don't need a comment in the source for it, if the change reasoning is documented.

done

heshamelmatary:
I don't remember the reason, but @kent-mcleod remembers better than I do. It wasn't for Rust though as I haven't accounted for that. Should be fine to merge.

Issue #1258, opened 2024-06-10
Emulation with `KernelRiscvUseClintMtime=ON` fails for QEMU >= 8.1.0


nspin:
I've observed that emulating seL4 on RISCV{32,64} platforms (at least `virt`, `spike`, and `hifive`) using QEMU fails with QEMU >= 8.1.0. In particular, PMP rejects accesses to the CLINT MTIME register. OpenSBI configures the MTIMER regions without any `SBI_DOMAIN_MEMREGION_SU_*` permissions [1]. seL4 was able to access the MTIME register on older versions of QEMU, even with the same version of OpenSBI, up until this commit, which appears to fix a bug in PMP policy enforcement:

https://github.com/qemu/qemu/commit/a574b27af4fe340684ca9f20560a1b01905e4364

Is anyone who knows more about all of this able to shed light on how `KernelRiscvUseClintMtime=ON` was able to work on platforms which use OpenSBI `PLATFORM=generic` in the first place, given that PMP configuration code I'm seeing?

Is `KernelRiscvUseClintMtime=ON` suitable for all of the platforms it is currently set for?

Tagging @Ivan-Velickovic, who drove the PR adding `KernelRiscvUseClintMtime` (https://github.com/seL4/seL4/pull/1076).

[1] https://github.com/riscv-software-src/opensbi/blob/d962db280725b03a0340e05a07e4c85c93f35bc5/lib/utils/timer/aclint_mtimer.c#L225-L260

axel-h:
I think it is a platform specific choice if access is allowed there - or if things are implemented this way at all behind the scenes of  `rdtime` and the CSRs `time`/`mtime`. `KernelRiscvUseClintMtime` came in a a faster way to do things based on how many platforms technically implement things, copying what SiFive implemented. I think it should not be used unless one knows that a platform support this, and SBI allows this, and timer reads (or even timer compare updates) must be as fast as possible. Otherwise stick with the standard RISC-V methods and hope platforms implement it in a efficient way instead of having M-Mode trap and emulate it.

Ivan-Velickovic:
@nspin more details here https://github.com/seL4/seL4/issues/1094.

I tried to resolve the general issue but didn’t come to a conclusion. It seems clear now though that it should probably be an opt-in option.

I will make a PR in the next couple days.

nspin:
Do we know which platforms it works for?

[...]

Issue #1094, opened 2023-08-17
Possible regression with Polarfire SoC


Ivan-Velickovic:
The commit that I made https://github.com/seL4/seL4/commit/69040396083959df9bb7cdc193cdf3f13af5bde8 recently seems to have caused a regression when booting seL4, as assertions such as:
```
Booting all finished, dropped to user space
seL4 failed assertion 'refill_ready(thread->tcbSchedContext)' at /host/seL4/src/kernel/thread.c:447 in function switchToThread
halting...
Kernel entry via Interrupt, irq 0
```

are going off.

The reason I call this a *possible* regression is that I do not know whether it is an issue with QEMU or the commit itself. What is confusing to me is that I believe the Polarfire SoC has the *same* U54-MC CPU that the HiFive Unleashed has, which has no problems with this patch. I unfortunately do not have access to the hardware, so cannot confirm whether this is an issue with QEMU or not.

Ivan-Velickovic:
This manual https://www.microsemi.com/document-portal/doc_view/1245725-polarfire-soc-fpga-mss-technical-reference-manual for the Polarfire (which I believe is the right manual to be looking at) states in Table 6-25 that `mtime` is at `0x0200_BFF8`, which matches with what the kernel ends up accessing to get the value of the register.

Indanz:
Could you add an if-check before the assert with the same condition and print out what the actual values are?

Perhaps it just overflows because it's not initialised correctly.

Indanz:
I think it got broken by [this](https://github.com/riscv-software-src/opensbi/commit/59a08cd7d632bb2394a29956497dfe7829099997) commit, which is part of [Split region permissions into M-mode and SU-mode](http://lists.infradead.org/pipermail/opensbi/2023-January/004127.html). Perhaps platforms are expected to add explicit S/U access to CLINT in their OpenSBI device tree since that patch, but not all platforms have done that?

I opened a new OpenSBI [issue](https://github.com/riscv-software-src/opensbi/issues/330), let's see what they say.

Ivan-Velickovic:
I don't think it was broken by that commit since v1.2 of OpenSBI was released last year and that commit is from January this year.

Ivan-Velickovic:
Thanks for posting that issue though, maybe it will give us some clarity.

axel-h:
[...]

PR #1259, opened 2024-06-11
Disable KernelRiscvUseClintMtime by default

Ivan-Velickovic:
This configuration option requires access to the CLINT in S-mode which is not available on newer versions of OpenSBI. Because of this, it makes more sense to not turn on this option by default and instead let it be 'opt-in' for those who want the performance benefit and also have firmware that does not restrict S-mode access to the CLINT.

Closes https://github.com/seL4/seL4/issues/1094.
Closes https://github.com/seL4/seL4/issues/1258.


lsf37 (approved):
Looks good, thanks for resolving this one!

axel-h (approved):
Looks good

Ivan-Velickovic:
The style check is failing on something unrelated.

Ivan-Velickovic:
But other than that I think it’s okay to merge?

lsf37:
> The style check is failing on something unrelated.

Yes, probably because the style checker settings have been updated a while ago, it's nothing you changed, just happens to be in a file you touched.

Ivan-Velickovic:
I don't have access to merge a PR that has a failing 'required' check so if someone else could that would be great.

PR #1261, opened 2024-06-11
Make write-only mapping message consistent

Ivan-Velickovic:
There is a warning when creating a write-only mapping on AArch32/AArch64.

This patch simply makes the message consistent across all architectures since the behaviour when creating a write-only mapping is also consistent across all architectures.

`x86/kernel/vspace.c` is used for x86 and ia32 right?

### DIFF
--- a/src/arch/riscv/kernel/vspace.c
+++ b/src/arch/riscv/kernel/vspace.c
@@ -641,6 +641,11 @@ vm_rights_t CONST maskVMRights(vm_rights_t vm_rights, seL4_CapRights_t cap_right
             return VMReadWrite;
         }
     }
+    if (vm_rights == VMReadWrite &&
+        !seL4_CapRights_get_capAllowRead(cap_rights_mask) &&
+        seL4_CapRights_get_capAllowWrite(cap_rights_mask)) {
+        userError("Attempted to make unsupported write only mapping");
+    }
     return VMKernelOnly;
 }
 
--- a/src/arch/x86/kernel/vspace.c
+++ b/src/arch/x86/kernel/vspace.c
@@ -682,6 +682,11 @@ vm_rights_t CONST maskVMRights(vm_rights_t vm_rights, seL4_CapRights_t cap_right
             return VMReadWrite;
         }
     }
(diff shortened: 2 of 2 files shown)


axel-h:
I think this should get a verbose comment for each architecture also why this is forbidden or what this the behavior.
- W=1,X=0: reserved for future use
- W=0,X=1: execute-only page
- W=0,X=1: reserved for future use
I remember darkly reading something in the ARM specs, that each page must be readable. Not sure about the current state there.
Also doe x86, what does the spec say?

Ivan-Velickovic:
Is it the job of the manual to explain the reasoning behind behaviour? I thought it was just to specify the behaviour.

Ivan-Velickovic:
Whoops wrong PR my bad

lsf37:
I'd be happy with this without additional comments for now, it just makes things more consistent.

Longer term it would make sense to document our assumptions on what the hardware does, just so that it's clear where the mistake is if they turn out to be wrong. But that doesn't need to be now, and this function is not the right place for it.

> Is it the job of the manual to explain the reasoning behind behaviour? I thought it was just to specify the behaviour.

Definitely not the manual, but I think Axel means the source.

>   I remember darkly reading something in the ARM specs, that each page must be readable. Not sure about the current state there.

I don't actually know, but that is independent of this function and this change. When both read and write are false, the rights are set to kernel-only, which can then be whatever it needs to be for the architecture. 

>   Also doe x86, what does the spec say?

It also doesn't support write-only pages.

lsf37 (approved):
Happy with this, but need to wait for the proofs -- they might need a small tweak for this, depending on how they are written.

lsf37:
> `x86/kernel/vspace.c` is used for x86 and ia32 right?

Yes

Indanz (approved):

lsf37:
It looks like there might be some verification impact on it after all, need to re-run and double check (one of the sessions was killed before it completed).

lsf37:
Fine to merge from the verification side. The MCS proof failure is unrelated.

PR #1269, opened 2024-06-21
mcs: use sc_active in isSchedulable

michaelmcinerney:


lsf37 (approved):
`sc_active` is declare inline, so this should be the same binary as before.

Indanz (approved):

PR #1270, opened 2024-06-23
arm: add missing `seL4_VPPIEvent_Length` constants

nspin:
`*_Length` constants exist for all `*_Msg` enums except for `seL4_VPPIEvent_Msg`.


lsf37 (approved):
Good to merge from my side. Proofs are passing and it makes sense to have these consistent.

PR #1271, opened 2024-06-23
DRAFT: domain,fpu: Reset FPU on domain switch

kent-mcleod:
The FPU is lazily switched to allow lower latency context switching when only some threads in the system are accessing the FPU. However domain switching should cause any FPU state to be written back the thread eagerly to avoid writing the state back from a different domain.


lsf37:
Closing this for now, we can do this if we decide not to go with the FPU RFC at seL4/rfcs#26.

PR #1272, opened 2024-06-24
manual: make list of fault types easier to read

Ivan-Velickovic:
I find it difficult to read the fault types like this, there's too many of them (and there are more architecture dependent ones that are not part of this list yet).

### DIFF
--- a/manual/parts/threads.tex
+++ b/manual/parts/threads.tex
@@ -371,11 +371,19 @@ The register contents are transferred via the IPC buffer.
 A thread's actions may result in a fault. Faults are delivered to the
 thread's exception handler so that it can take the appropriate action.
 The fault type is specified in the message label and is one of:
-\texttt{seL4\_Fault\_CapFault}, \texttt{seL4\_Fault\_VMFault},
-\texttt{seL4\_Fault\_UnknownSyscall}, \texttt{seL4\_Fault\_UserException},
-\texttt{seL4\_Fault\_DebugException},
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
> I think that list should go altogether, it gives no more information that the table of content.

The problem is, from what I have seen, there is no way for me to find what faults are available and what they are actually called in libsel4 without compiling the kernel myself and looking through the auto-generated code.

Indanz:
That's what the Faults section should solve though, it should list all faults, just like the "Error Codes" section does for errors. The error section is clearer because the subsections per error are shorter. 

Or do you also miss an overview table in the error section? 

The main problem of the faults section is that the debug API is described there instead of elsewhere. If it would list just the faults the whole Faults section would be more readable.

Ivan-Velickovic:
Sorry, I should've been clearer. I just talking about the fault *labels*. So let's say if we removed this list and I was trying to write a fault handler that looks at the label to decide what to do, I would not know what values correspond to what kind of fault without looking through the kernel source code/auto generated code.

The manual should probably say that the label corresponds to a value of `seL4_Fault_tag_t` and then list the enum values.

Indanz:
Yes, you're right, the labels aren't listed. The error section solves that by adding the label to the table, but that's more awkward with faults. With that in mind, I approve your changes. Could you add the missing faults to the list now it's not a cumbersome long sentence anymore? I would move the debug and arch specific faults to the end of the list.

lsf37 (approved):

Indanz (approved):

PR #1274, opened 2024-06-24
Fix macro to get irq idx from irqt

MilicatZhang:


Indanz (approved):
The mistake was introduced by commit 2d362cb7c8e96a64b9848df06e226f50e79a1ec9.

lsf37 (approved):
Nice catch.

PR #1275, opened 2024-06-25
Enable SMP on qemu-arm-virt

andybui01:
Tried to get some multicore tests running on qemu-arm-virt and I couldn't, which lead me here. Seems like I'm not the only one to have issues with qemu + smp (was this ever used?) https://lists.sel4.systems/hyperkitty/list/devel@sel4.systems/message/UQD6RAMY2AEIJXHIGFIITEKMH7U35NJ7/

See: https://github.com/seL4/seL4_tools/pull/202


lsf37:
qemu-arm-virt with SMP is entirely unexplored, in fact -virt itself is pretty new. Happy to try it, though.

src/plat/qemu-arm-virt/config.cmake
```diff
@@ -125,9 +125,7 @@ if(KernelPlatformQEMUArmVirt)
             if(NOT DEFINED QEMU_MACHINE)
                 set(QEMU_MACHINE "virt")
                 list(APPEND QEMU_MACHINE "secure=off")
-                if(KernelArmHypervisorSupport OR KernelSel4ArchArmHyp)
-                    list(APPEND QEMU_MACHINE "virtualization=on")
-                endif()
+                list(APPEND QEMU_MACHINE "virtualization=on")
```
lsf37, line 128: This now always switches virtualisation on. Do we want that? In fact, I don't understand how this interacts with SMP at all.

Indanz: As the commit message says:

> We always use the qemu EL2 implementation, as this allows the elfloader
to boot secondary cores using SMC (the only method the elfloader
supports). Note this is fine even if CONFIG_ARM_HYPERVISOR_SUPPORT is
not set as the elfloader will drop from EL2 to EL1 before entering the
seL4 kernel.

However, to me it sounds like Elfloader may need to learn to use `HVC` instead of `SMC` when there is no EL3.

This commit seems to be a work around for Elfloader's limitations. Why can't you just enable `KernelArmHypervisorSupport` when using QEMU manually?

andybui01: Yeah this method isn't the best, pasting Gerwin's comment on the other PR
> We do want to test simulated kernels on platforms that don't support virtualisation

I think it'll be better to just implement PSCI secondary core booting via HVC in the elfloader. I'll try and find some time to do this.

Indanz:
Feel free to re-open if adapting Elfloader doesn't work.

Ivan-Velickovic:
@andybui01 I'm adding SMP support for Microkit now and only use SMC to turn on other CPUs, it works fine on QEMU.

andybui01:
@Ivan-Velickovic  SMP+HYP or just SMP?

Ivan-Velickovic:
SMP and HYP

andybui01:
SMP+HYP works fine AFAIK as there's an EL3 implementation as @Indanz pointed out, loading into SMP non-HYP (so EL1) doesn't seem to work well. My idea was to launch qemu in HYP/EL2 and let the elfloader drop to EL1, but a more permanent solution exists in just implementing the HVC, when I remember to do it :').

Ivan-Velickovic:
> SMP+HYP works fine AFAIK as there's an EL3 implementation as @Indanz pointed out, loading into SMP non-HYP (so EL1) doesn't seem to work well. My idea was to launch qemu in HYP/EL2 and let the elfloader drop to EL1, but a more permanent solution exists in just implementing the HVC, when I remember to do it :').

Ah whoops I missed that.

PR #1277, opened 2024-06-26
update caveats file

lsf37:
- turn into an .md file for GitHub rendering
- add SMP and MCS descriptions
- be more precise about verified platforms
- [x] MCS WCET caveat
- [x] MSI remapping
- [x] Spectre/Meltdown

[Rendered version](https://github.com/seL4/seL4/blob/lsf37/caveats/CAVEATS.md)


lsf37:
Requesting a review from the TSC, because this basically describes what we are committing to maintaining. I don't expect anything controversial in here, but it'd be good to get at least a few people from the TSC to have a look over this new version of the CAVEATS file.

CAVEATS.md
```diff
@@ -0,0 +1,175 @@
...
+they are simpler and closer to the current sequential seL4 proofs.
+
+In a multi-kernel configuration, each CPU core runs a separate instance of seL4,
+with each kernel instance getting access to disjoint subsets of memory of the
+machine. User-level memory can be shared as device-untyped memory, which the
+kernel manages but does not access. These configurations can already be set up
+without kernel changes by providing suitable device tree overlays to each kernel
+instance. Further work is planned to make such configurations easier to use and
+more robust against unsafe use/configurations, e.g. by managing IRQ controller
+access for each instance.
```
Indanz, line 136: Considering this is the caveats file, perhaps make it clearer whether current SMP will be supported in the future, or if it the plan is to obsolete it with multi-kernel. Now it hints at the latter.

CAVEATS.md
```diff
@@ -0,0 +1,175 @@
...
+installed in it should be revoked. The kernel will not do this automatically
+for the user.
+
+If, for instance, page cap `c` is installed in the address space denoted by a
+page directory under ASID `A`, and the page directory is subsequently revoked or
+deleted, and then a new page directory is installed under that same ASID `A`,
+the page cap `c` will still retain some authority in the new page directory,
+even though the user intention might be to run the new page directory under a
+new security context. The authority retained is to perform the unmap operation
+on the page the cap `c` refers to.
```
Indanz, line 150: Surely this only applies to platforms that can't invalidate all TLB entries by ASID? ARM invalidates the ASID when unmapping the page table or destroying a mapped vspace. A quick look at x86's and riscv's `deleteASID()` confirms the same behaviour there. So I'm mystified how the stale mappings can be still active after a delete.

Does it only apply to re-using ASIDs and page tables without a delete and assigning it to a different TCB? Perhaps say that explicitly? But why would the user expect the kernel to unmap existing mappings from a page table automatically for any reason?

Revoking a page table does nothing, as there are no derived objects, so why mention it?

CAVEATS.md
lsf37: I don't actually know -- if we can fix the current issues there is no reason to remove it, but we haven't been very successful at that in the past year.

lsf37:
One bit that is missing in the current text is that the WCET numbers are made up and need to be set based on the use case.

- [ ] add MCS WCET caveat

lsf37:
And one more thing I forgot:

- [ ] Spectre/Meltdown
- [ ] x86 IRQ remapping issue

CAVEATS.md
Indanz: Mostly because apparently only Kent and me do anything SMP related and we were both distracted by other things last year (Kent more than me, but I can't do anything SMP related without him).

The only advantage of current SMP are simpler system initialisation (partly because of the ability to do cross-core calls) and more efficient cross-core notifications directly to the correct task. In all other regards it's worse: Lower performance and less verifiable.

If the plan isn't to obsolete current SMP, then perhaps multi-kernel support shouldn't be mentioned in the caveats file.

CAVEATS.md
Indanz: Maybe it's clearer if you say that page caps store the vspace's ASID when mapped and will keep granting rights to do operations on that mapped virtual address range with that ASID even after that ASID has been re-used for another vspace.

lsf37: I can give that a shot

lsf37: Does the new version work better?

CAVEATS.md
lsf37: Have now clarified that multikernel is an intermediate step. Hopefully that makes it sound less like SMP is on the way to the scrap heap.

lsf37:
@seL4/tsc today is the last chance for comments on this one, the release is planned for tomorrow.

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+The following seL4 architectures have platforms with a C-level functional
+correctness proof. Proof support for further platforms within these
+architectures is on the roadmap and expected in 2025.
+
+- AArch32: Armv7-a with and without hypervisor extensions, no SMMU, with
+  fast path
+  - Platforms (non-hyp): `sabre` (no FPU), `imx8mm-evk` (with FPU)
+  - Platforms (hyp, no FPU): `tk1`, `exynos5`
+- AArch64: Armv8-a with hypervisor extensions only, no SMMU, with fast path
+  - Platforms: `tx2`
```
Xaphiosis, line 20: Too early to say something about the PROVERS platform here?

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+- x64: without VT-x and VT-d, no fast path
+  - Platforms: `pc99`
+
+This proof covers the functional behaviour of the C code of the kernel. It does
+not cover machine code, compiler, linker, boot code, cache and TLB management.
+Compiler and linker can be removed from this list by additionally running the
+binary verification tool chain for seL4 for AArch32 or RISC-V. The proof shows
+that the seL4 C code implements the formal [abstract API specification][ASpec]
+of seL4 and is free from standard C implementation defects such as buffer
+overruns or NULL pointer dereferences.
```
Xaphiosis, line 32: This makes it sound a bit like the binary verification proof shows that ASpec is refined by the C code. Can leave it, as "the proof" refers to "this proof" earlier, and if someone doesn't know binary verification is a proof, it'll probably be ok.

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+## Verified Configurations
+
+For the precise configuration of the verified platforms above, see the
+corresponding files in the seL4 `configs/` directory.
+
+The proofs are generally sensitive to changes in configuration parameters, and
+will break if these are changed. For some parameters, the proofs are explicitly
+set up to be robust, such as the number of domains `NUM_DOMAINS`, and the domain
+schedule. More such parameters are on the roadmap to be added and documented
+here.
```
Xaphiosis, line 65: is your maxIRQ stuff in, or too early?

CAVEATS.md
lsf37: Yes, it's not merged yet, and won't be in time for the release. But we can probably add it next week..

CAVEATS.md
lsf37: Could say "The overall proof shows". Would that be clearer?

CAVEATS.md
lsf37: also not merged yet, but looking forward to adding it to the next release notes :-)

CAVEATS.md
corlewis: Just to add another data point, I had the same thought about first reading the second half of this paragraph as still talking about binary verification.

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+- AArch64: Armv8-a with hypervisor extensions only, no SMMU, with fast path
+  - Platforms: `tx2`
+- RISC-V: 64-bit only, no fast path
+  - Platforms: `hifive`
+- x64: without VT-x and VT-d, no fast path
+  - Platforms: `pc99`
+
+This proof covers the functional behaviour of the C code of the kernel. It does
+not cover machine code, compiler, linker, boot code, cache and TLB management.
+Compiler and linker can be removed from this list by additionally running the
```
corlewis, line 28: Very minor
```suggestion
The compiler and linker can be removed from this list by additionally running the
```

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+
+The MCS configuration of the kernel addresses many of these real-time problems
+and provides principled access control for execution time, but its formal
+verification is currently still in progress. For RISC-V, design-level proofs
+have completed, and C-level proofs are in progress. Similar proofs for AArch64
+are planned.
+
+The MCS configuration is supported by the seL4 foundation and should generally
+be stable, with small API changes to be expected while verification is ongoing
+and the configuration is used in more systems. See open [requests for
```
Xaphiosis, line 88: suggest "deployed" instead of "used"

corlewis (approved):

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+the kernel platform definitions are defaults only and need to be determined
+based on the specific use case -- for instance, static systems can be set up to
+have low latency and WCET, whereas dynamic systems or systems where untrusted
+code has authority to perform longer-running kernel operations may need higher
+values.
+
+## SMP
+
+A symmetric multi-processor (SMP) configuration for seL4 exists and is supported
+by the seL4 foundation, but currently without formal verification. While
```
Xaphiosis, line 101: Foundation (capital)

CAVEATS.md
```diff
@@ -0,0 +1,232 @@
...
+While no configuration in this release of seL4 provides timing or other
+micro-architectural channel guarantees, timing channel exploits on the x86
+architecture are more widespread and relevant, especially for dynamic systems.
+
+The kernel does provide configuration options for mitigating [Meltdown] and
+[Spectre]-style attacks, but there are no specific mitigations for more recent
+such attacks such as [Zenbleed] and [Inception].
+
+Note that the kernel does not depend on secrets, cryptographic or otherwise, so
+most kernel targets for such attacks are not present, but timing channels
```
Xaphiosis, line 207: channel (non-plural)

Xaphiosis (approved):
Minor comments and questions, otherwise fine by me.

CAVEATS.md
Xaphiosis: The "overall" might help clarify here, yes. Probably not worth re-engineering the whole paragraph.

Issue #1186, opened 2024-02-05
`physBase` isn't aligned properly on `ARM_HYP`


nspin:
`physBase` is required to be aligned to `seL4_SuperSectionBits`:

https://github.com/seL4/seL4/blob/c9989664e312d971048f687895d4917758fcfd3b/include/arch/arm/arch/32/mode/hardware.h#L98-L100

On `ARM_HYP`, `seL4_SuperSectionBits = 25`:

https://github.com/seL4/seL4/blob/c9989664e312d971048f687895d4917758fcfd3b/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h#L195

However, in `hardware/config.py` the value of `SUPERSECTION_BITS` is hard-coded to `24`, its value on `ARM`:

https://github.com/seL4/seL4/blob/c9989664e312d971048f687895d4917758fcfd3b/tools/hardware/config.py#L48-L51

While this bug may not have caused issues for currently supported platform, I noticed it when testing https://github.com/seL4/seL4/pull/976 on `ARM_HYP`. In this case, `physBase` is set to `2**24` by `hardware/config.py` and the assertion in `hardware.h` fails.

lsf37:
Nice find, and I'm glad that the assert catches it, because that would have been super hard to find at runtime. The value in `config.py` is just wrong, it needs to depend on whether hyp mode is on or not.

PR #1278, opened 2024-06-26
config.py: fix SUPERSECTION_BITS for arm_hyp

lsf37:
This addresses #1186

It's not entirely clear to me what this does for AArch64 -- is that code used there?

### DIFF
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -54,7 +54,7 @@ class ARMConfig(Config):
         self.SUPERSECTION_BITS = 25 if sel4arch == 'arm_hyp' else 24
 
     def get_kernel_phys_align(self) -> int:
-        ''' on ARM the ELF loader expects to be able to map a supersection page to load the kernel. '''
+        ''' On AArch32 the kernel requires at least super section alignment for physBase. '''
         return self.SUPERSECTION_BITS
 
     def align_memory(self, regions: Set[Region]) -> List[Region]:
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -48,7 +48,10 @@ class Config:
 class ARMConfig(Config):
     ''' Config class for ARM '''
     arch = 'arm'
-    SUPERSECTION_BITS = 24  # 2^24 = 16 MiByte
+
+    def __init__(self, sel4arch, addrspace_max):
+        super().__init__(sel4arch, addrspace_max)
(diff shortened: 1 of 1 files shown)


lsf37:
(In any case, it's no worse than before..)

tools/hardware/config.py
```diff
@@ -48,7 +48,10 @@ def align_memory(self, regions: Set[Region]) -> List[Region]:
...
     ''' Config class for ARM '''
     arch = 'arm'
-    SUPERSECTION_BITS = 24  # 2^24 = 16 MiByte
+
+    def __init__(self, sel4arch, addrspace_max):
+        super().__init__(sel4arch, addrspace_max)
+        self.SUPERSECTION_BITS = 25 if sel4arch == 'arm_hyp' else 24
 
     def get_kernel_phys_align(self) -> int:
         ''' on ARM the ELF loader expects to be able to map a supersection page to load the kernel. '''
```
kent-mcleod, line 57: This comment doesn't seem right to me.
- On arm (non-hyp) the elfloader maps the kernel with 1Mib pages,
- On arm-hyp 2MiB pages are used,
- On aarch64 1GiB pages are used.

There's a static assert in the kernel that talks about supersections:
```

/* For alignment conditions to translate over addrFromPPtr, physBase must be
   aligned to at least the size of a SuperSection. */
compile_assert(physBase_aligned, IS_ALIGNED(PHYS_BASE_RAW, seL4_SuperSectionBits));
```

Is there a requirement that the kernel mapping window needs to at least preserve alignment of the largest page size for the architecture? (Assuming that that's the largest kernel object that has hardware alignment requirements).

lsf37: Yes, for AArch32, there is arithmetic that will fail (in the kernel) if the mapping window is not aligned to the largest page size. I'm not sure any more why that is not the case for AArch64, but it could be that the alignment there is hard coded to be much bigger anyway.

The comment in the file is old, I don't know where it's from, but I guess the alignment of the mapping window doesn't really have to coincide with the page size that is used.

axel-h (approved):
Looks good to me to fix the original issue.

tools/hardware/config.py
lsf37: Ok, so we do know that the comment is wrong, because the ELF loader does not actually do what the comment says. I'll update it to say that it is a kernel requirement for AArch32.

lsf37: Now done.

Issue #1281, opened 2024-07-01
stale frame cap could be used to flush "wrong" memory


lsf37:
If a frame cap whose ASID/vaddr mapping information is out of date is used in one of the page flush operations (`ARMPageClean_Data`, `ARMPageInvalidate_Data`, `ARMPageCleanInvalidate_Data`, `ARMPageUnify_Instruction`), and the kernel runs in EL1 where the user page table is used for flushing, we could end up flushing an unintended memory region.

It's debatable what unintended means here, because the flush is by virtual address, and it will flush the correct virtual address, but the frame may either not be mapped any more or a different frame (of maybe different size) may be mapped at the same virtual address. No matter if the behaviour is actually breaking integrity or not, a stale frame cap should not be used in a new VSpace, so the behaviour should be excluded.

This means, the same checks as for `unmapPage` should be performed for page flush.

Thanks to @kent-mcleod for spotting that during the discussion on #1265

lsf37:
Reopening this, but removing it from the release. The situation is now safe from the kernel perspective, but we should still check for stale mapping info in all configs -- also in hyp mode, and possibly also on AArch32.

Less urgent, but should still be implemented for the next release.

kent-mcleod:
aarch32 has the same flush operation that doesn't check if the frame is still mapped.

lsf37:
This is now addressed for aarch32 as well, but ideally we should also check this on hyp configs, just on the principle that we shouldn't allow operations on stale frame caps.

PR #1282, opened 2024-07-01
vspace: check for stale mapping on all flush types

lsf37:
Frame caps should be checked for stale mapping info before authorising any operation on them. In this particular case, the mapping may have become invalid.

See also #1281

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1664,26 +1664,25 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
 #else
+        /* When in EL1, we are using the user page table for flushing and need to make sure
+           the mapping info in the cap is not stale. */
+        lookupPTSlot_ret_t lu_ret = lookupPTSlot(find_ret.vspace_root, vaddr);
+        pte_t pte = *lu_ret.ptSlot;
+        void *base_ptr = (void *) cap_frame_cap_get_capFBasePtr(cap);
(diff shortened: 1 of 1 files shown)


src/arch/arm/64/kernel/vspace.c
```diff
@@ -1664,26 +1664,25 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
...
+        pte_t pte = *lu_ret.ptSlot;
+        void *base_ptr = (void *) cap_frame_cap_get_capFBasePtr(cap);
+        if (unlikely(lu_ret.ptBitsLeft != pageBitsForSize(cap_frame_cap_get_capFSize(cap)) ||
+                     !pte_is_page_type(pte) ||
+                     pte_get_page_base_address(pte) != pptr_to_paddr(base_ptr))) {
+            userError("Page Flush: Attempting to use cap with stale mapping information.");
+            current_syscall_error.type = seL4_InvalidCapability;
+            current_syscall_error.invalidCapNumber = 0;
+            return EXCEPTION_SYSCALL_ERROR;
+        }
```
kent-mcleod, line 1679: Could this be lifted even higher to line: 1637, right after `findVSpaceForASID`?
In EL2, the kernel still shouldn't flush a cap with a stale ASID.

kent-mcleod (approved):

src/arch/arm/64/kernel/vspace.c
lsf37: In EL2, the operation is at least safe, because we do know the requested virtual address is mapped in the kernel window. I do agree that it would be better long-term to make this an error, because operating on stale mapping info is a user error.

Since moving it into EL2 will need re-verification I would like to merge this variant first, though, so we can release a version that is at least safe.

PR #1288, opened 2024-07-01
x86: add VGA as possible earlyprint device

ericspidledw:
Currently on x86 the only way to see early output is through the COM ports. If a platform is missing these it can be difficult to debug early boot. This change allows for users to output early kernel prints to a VGA display. It leverages the common VGA text buffer address (`0xB8000` ) to do so.


include/plat/pc99/plat/machine/devices.h
```diff
@@ -10,7 +10,13 @@
 
 #define PPTR_APIC KDEV_BASE
 
+#if defined(CONFIG_VGA_PRINTING)
+#define PPTR_VGA    (PPTR_APIC + BIT(PAGE_BITS))
+#define PPTR_IOAPIC_START (PPTR_VGA + BIT(PAGE_BITS))
+#else
```
axel-h, line 16: I wonder, so you want to make this exclusive or should both output options be available at the same time also for convenience?

ericspidledw: Good question. I presumed if the COM port was available it'd be more convenient to use that but I could see there being some value in having both. I can rework it to allow both devices to be used if you'd like

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,96 @@
...
+    		current_xpos = 0;
+    		current_ypos++;
+    	}
+    }
+}
+
+void kernel_putDebugChar(unsigned char c)
+{
+    /* this will take care of CR/LF handling and call uart_drv_putchar() */
+    uart_drv_putchar(c);
```
axel-h, line 89: This looks odd, why not call `vga_putchar(c);` there directly?

ericspidledw: Good point I changed this to just call the function directly

include/plat/pc99/plat/machine/devices.h
ericspidledw: I started working on this but realized if this were the case we'd have some repeat functions like `kernel_putDebugChar` and `serial_init`. There is probably a workaround for this but it could make things more confusing/entangled do we really want too make sure that a user can have both on at the same time?

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,96 @@
...
+static uint64_t vga_addr = 0;
+
+/* Name kept 'serial_init' since other code expects to call this */
+void serial_init(uint16_t port)
+{
+	set_vga_addr(VGA_PHYS);
+    return;
+}
+
+void set_vga_addr(uint64_t addr)
```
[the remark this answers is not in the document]
ericspidledw: Added a check for reads and writes that makes sure nothing over the text buffer's size is allowed

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,107 @@
...
+
+uint64_t get_vga_addr(void)
+{
+    return vga_addr;
+}
+static int verify_vga_offset(uint64_t offset)
+{
+    if(offset > vga_size)
+    {
+        printf("VGA offset of 0x%llx is outside size of 0x%llx", offset, vga_size);
```
axel-h, line 45: I think we should not print anything there, as this is called in the context of printing something? Might end in a bad recursion in the worst case.

ericspidledw: Good catch I didn't think about that. Should we just fail silently then?

axel-h: Could put an assert() or halt() there, as this likely a fatal configuration of implementation error.

ericspidledw: I put an `assert(0)`  here just to ensure it fails. I could just change it to be `assert((offset <= vga_size))` if we think that'd look cleaner

axel-h: Yes, it falls down to this in the end. But assert() also tries to print something, so do nothing seems best?

ericspidledw: oh right I forgot assert prints too and so does Halt. Yeah silent failing may just be the best course of action here. A user realistically shouldn't ever index outside of the allocated space if they just make calls to printf

ericspidledw:
@axel-h  The only failing test now is the x86_64 preprocess. The snapshot doesn't match up since I added an extra config for the VGA printing. How would I go about remedying this or is that something that needs to be done the maintainers end?

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+#define VGA_COLOR 0x07
+#define CLEAR_SCREEN_CHAR 32
+static uint64_t vga_addr = 0;
+static uint64_t vga_size = 0;
+// assume 80x25 console
+#define MAX_X_POS 80
+#define MAX_Y_POS 25
+
+/* Name kept 'serial_init' since other code expects to call this */
+void serial_init(uint16_t port)
```
axel-h, line 25: Since this is used on x86 only and looking at the places in `src/arch/x86/kernel/cmdline.c`where this is called, there seems some benefit if we make this `kernel_putDebugChar_init()`. And sanitize the code there, that deals with the command line arguments and setup. Looking at `src/arch/x86/model/statedata.c` we have `x86KSconsolePort` and `x86KSdebugPort`, which now only seem to have a use when `src/plat/pc99/machine/dev/com/io.c` is used. So I wonder if these variables should move then also to make this fully modular, as this seem the counterpart of `vga_addr` and `vga_size` in the end.

src/arch/x86/kernel/boot_sys.c
```diff
@@ -23,6 +23,8 @@
 #include <plat/machine/pic.h>
 #include <plat/machine/ioapic.h>
 #include <sel4/arch/bootinfo_types.h>
+#include <drivers/uart.h>
```
axel-h, line 26: Do we really need this include here?

src/arch/x86/kernel/vspace.c
```diff
@@ -149,6 +149,17 @@ BOOT_CODE bool_t map_kernel_window_devices(pte_t *pt, uint32_t num_ioapic, paddr
     assert(idx == (PPTR_APIC & MASK(LARGE_PAGE_BITS)) >> PAGE_BITS);
     pt[idx] = pte;
     idx++;
+
+#ifdef CONFIG_VGA_PRINTING
```
axel-h, line 153: If we do this mapping last, then then mapping address of the other device does not change if this is enabled or not used. Might be more convenient for debugging.

lsf37:
The preprocess check sees a changed order of include files. Should be fine for the proofs, but I've started a test to make sure.

include/plat/pc99/plat/machine/devices.h
lsf37: I think it'd be fine to leave that for a future iteration.

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+        current_ypos = MAX_Y_POS - 1;
+    }
+
+    if (c == '\n') {
+        current_xpos = 0;
+        current_ypos++;
+    } else if (c != '\r')  {
+        vga_write(((VGA_COLOR << 8) | (unsigned short) c),
+                  2 * (MAX_X_POS * current_ypos +
+                       current_xpos++));
```
lsf37, line 94: Please don't use side effects in expressions, even in unverified code (I would argue especially in unverified code):

```suggestion
        vga_write(((VGA_COLOR << 8) | (unsigned short) c),
                  2 * (MAX_X_POS * current_ypos + current_xpos));
        current_xpos++;
```

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+        for (prev_column = 1, column = 0; prev_column < MAX_Y_POS; prev_column++, column++) {
+            for (row = 0; row < MAX_X_POS; row++) {
+                vga_write(vga_read(2 * (MAX_X_POS * prev_column + row)),
+                          2 * (MAX_X_POS * column + row));
+            }
+        }
+        for (row = 0; row < MAX_X_POS; row++) { // clear last line so we can write too it
+            vga_write((VGA_COLOR << 8) | CLEAR_SCREEN_CHAR, 2 * (MAX_X_POS * column + row));
+        }
+        current_ypos = MAX_Y_POS - 1;
```
lsf37, line 85: This suggests that the numbers 80/25 are fine, but the name `VGA_NUM_Y_POS` or `VGA_NUM_ROWS` would be correct. If we want to write to the last line, we are setting the position to `VGA_NUM_ROWS - 1`.

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+
+static uint16_t vga_read(uint64_t offset)
+{
+    if (!verify_vga_offset(offset)) {
+        return *(uint16_t *)(vga_addr + offset);
+    }
+    return 0;
+}
+
+static int current_ypos = MAX_Y_POS, current_xpos = 0;
```
lsf37, line 65: This means the first time we print, we scroll and clear the bottom line. I think that is good. Might be worth adding a comment here that this is what we want to achieve.

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+        for (row = 0; row < MAX_X_POS; row++) { // clear last line so we can write too it
+            vga_write((VGA_COLOR << 8) | CLEAR_SCREEN_CHAR, 2 * (MAX_X_POS * column + row));
+        }
+        current_ypos = MAX_Y_POS - 1;
+    }
+
+    if (c == '\n') {
+        current_xpos = 0;
+        current_ypos++;
+    } else if (c != '\r')  {
```
lsf37, line 91: This means we ignore `\r` -- is that what we want? I'd have expected `current_xpos = 0` for the `\r` case instead. (Linux does not do this either, since the code is from there -- @wom-bat do you know why?)

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return 0;
+}
+
+static int current_ypos = MAX_Y_POS, current_xpos = 0;
+
+/***
+ * Slight rework of function 'early_vga_write' from
+ * the linux kernel at arch/x86/kernel/early_printk.c
+ ***/
```
lsf37, line 70:
```suggestion
/**
 * Slight rework of function 'early_vga_write' from
 * the linux kernel at arch/x86/kernel/early_printk.c
 *
 * Invariant: 0 <= current_ypos <= MAX_Y_POS && 0 <= current_xpos < MAX_X_POS
 * current_ypos == MAX_Y_POS means we scroll first and write to MAX_Y_POS-1
 */
```

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+{
+    if (offset > vga_size) {
+        return -1;
+    }
+    return 0;
+}
+
+static void vga_write(uint16_t data, uint64_t offset)
+{
+    if (!verify_vga_offset(offset)) {
```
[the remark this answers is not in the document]
lsf37: Since we are reading and writing `uint16`, we should only be writing to offsets that are divisible by 2, otherwise we'd be writing the content field of one char and the colour field of the next. We could also potentially be reading/writing one byte past the screen area. So we should probably make this something like:

```c
    if (offset < vga_size && (offset & 1 == 0)) {
      *(uint16_t *)(vga_addr + offset) = data;
    }
```

Then again, if this function is only used inside `vga_putchar` we could also not check, because we do always multiply by 2 in there. In that case, adding a comment that we expect offset to be divisible by 2 here would be good.

lsf37 (commented):
Overall this looks good and I think we should merge it. I do have a few more small suggestions to work through before that, but nothing major.

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+#define VGA_PHYS 0xb8000 // Text buffer address
+
+//black background, light grey text, no blinking
+#define VGA_COLOR 0x07
+#define CLEAR_SCREEN_CHAR 32
+static uint64_t vga_addr = 0;
+static uint64_t vga_size = 0;
+// assume 80x25 console
+#define MAX_X_POS 80
+#define MAX_Y_POS 25
```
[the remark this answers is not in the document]
lsf37: (further reading shows it should be renamed, the numbers are fine unless we want to change more of the code)

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+
+/***
+ * Slight rework of function 'early_vga_write' from
+ * the linux kernel at arch/x86/kernel/early_printk.c
+ ***/
+static void vga_putchar(char c)
+{
+    int  row, prev_column, column;
+    if (current_ypos >= MAX_Y_POS) { // do we need to scroll
+        /* scroll 1 line up */
```
axel-h, line 75: Might be simpler to do a `memmove()` here. Also, I think the names for row and columns are mixed up. Maybe just use x/y for the variables to stick to one set of names.

lsf37: I would prefer not to use `memmove()`. The correspondence to the Linux implementation is still pretty clear here and this is not performance critical.

lsf37: And good catch with the row/column names, you are absolutely right, they are swapped. Columns should be < MAX_X_POS, rows should be < MAX_Y_POS.

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+#include <machine/io.h>
+#include <plat/machine/io.h>
+#include <drivers/uart.h>
+
+#define VGA_PHYS 0xb8000 // Text buffer address
+
+//black background, light grey text, no blinking
+#define VGA_COLOR 0x07
+#define CLEAR_SCREEN_CHAR 32
+static uint64_t vga_addr = 0;
```
axel-h, line 18:
```suggestion
static word_t vga_addr = 0;
static word_t vga_size = 0;
```

src/plat/pc99/machine/dev/vga/io.c
```diff
@@ -0,0 +1,105 @@
...
+
+static int verify_vga_offset(uint64_t offset)
+{
+    if (offset > vga_size) {
+        return -1;
+    }
+    return 0;
+}
+
+static void vga_write(uint16_t data, uint64_t offset)
```
axel-h, line 50: It's a bit nitpicking, but when looking at the usage, isn't this the signature to use: 
```suggestion
static void vga_write(uint16_t data, word_t x, word_t y)
```

src/plat/pc99/machine/dev/vga/io.c
axel-h, line 65:
```suggestion
static word_t current_ypos = MAX_Y_POS, current_xpos = 0;
```

lsf37: I'd prefer that as well. If we do this, then the other `int`s should also all become `word_t`

src/plat/pc99/machine/dev/vga/io.c
lsf37: Could then also centralise the (x, y) to offset calculation into one function that is called for both read and write. I was thinking about suggesting all that, but left it because it'd change a lot of code without making that big a difference. It'd overall be slightly nicer to read, but the code is not that complex. On the downside it would prevent the compiler from optimising the arithmetic in the `for` loop where the scrolling happens. 

Happy to leave it up to @ericspidledw to decide, I'm fine with either version.

lsf37:
One more consideration to add: what happens when this runs on multiple cores? For SMP we should be in a big lock, for a multikernel setup that would not be the case, and at boot time that may also not be the case yet when printing happens early.

As far as I can see, running this code concurrently would violate the invariant the putchar function is assuming -- mostly in fairly safe ways (i.e. it can overrun both xpos and ypos, but not by terribly much. If xpos >= MAX_X_POS, the offset calculation just puts things on the next line, which is what we would expect). We could be losing characters that get overwritten when one thread writes before the other updates the position, and we could be attempting to write outside the screen, but we have the offset checks in the read/write functions, so these would be no-ops (but also losing characters).

So: semi-chaotic behaviour in a way you might expect from uncontrolled concurrent printing, but so far I do not see safety violations.

PR #1289, opened 2024-07-03
move cache flush from untyped reset to retype

lsf37:
Do not perform cache flushing in clearMemory. Instead flush the cache only for those object types where it is necessary, and only when the object is retyped, not when the untyped cap is reset.

This reduces overall need for flushing and delays it to the point of use. This should speed up boot time significantly, but may impact WCET, because the largest flush is now the largest page size (e.g. 24) instead of CONFIG_RESET_CHUNK_BITS (8). The user could already request a flush of the largest page size before, though, so it this may not actually impact WCET. Remains to be investigated.

Why this is safe:

- Flushing is only necessary for objects that are seen by other parts of the system, not for kernel-internal object. These objects are non-device frames (including IOMMU pages) and page tables. All other objects are only read/written by the kernel or are never read/written by the kernel (device frames + untypeds). Frames need to be flushed to RAM (as clearMemory did), because they could be seen uncached by devices. Page tables only to PoU for the page table walker.

- Before createNewObject in retype, these objects do not exist and cannot be seen by any part of the system. createNewObject is the point where new objects can become visible to the user.

- Theoretically, we could defer flushing further to the point where frames or page tables are mapped, but it is more complex to track whether a flush has already happened when they are mapped multiple times, whereas at retype the object cannot have been flushed already.

- The original implementation, before clearing memory was moved into reset untyped, also flushed at the same points.

This implements option 2 of the discussion in #481

### DIFF
--- a/include/arch/arm/arch/machine.h
+++ b/include/arch/arm/arch/machine.h
@@ -42,12 +42,10 @@ void cleanInvalidate_L1D(void);
 void cleanCaches_PoU(void);
 void cleanInvalidateL1Caches(void);
 
-/* Cleaning memory before user-level access */
+/* Cleaning memory before user-level access. Does not flush cache. */
 static inline void clearMemory(word_t *ptr, word_t bits)
 {
     memzero(ptr, BIT(bits));
-    cleanCacheRange_RAM((word_t)ptr, (word_t)ptr + BIT(bits) - 1,
-                        addrFromPPtr(ptr));
 }
 
 /* Cleaning memory before page table walker access */
--- a/include/arch/riscv/arch/machine.h
+++ b/include/arch/riscv/arch/machine.h
@@ -146,7 +146,7 @@ static inline void hwASIDFlush(asid_t asid)
 word_t PURE getRestartPC(tcb_t *thread);
 void setNextPC(tcb_t *thread, word_t v);
 
-/* Cleaning memory before user-level access */
+/* Cleaning memory before user-level access. Does not flush cache. */
 static inline void clearMemory(void *ptr, unsigned int bits)
 {
     memzero(ptr, BIT(bits));
--- a/include/arch/x86/arch/machine.h
+++ b/include/arch/x86/arch/machine.h
@@ -339,11 +339,10 @@ static inline void x86_load_fsgs_base(tcb_t *thread, cpu_id_t cpu)
     x86_write_gs_base(gs_base, cpu);
 }
 
-/* Cleaning memory before user-level access */
+/* Cleaning memory before user-level access. Does not flush cache. */
 static inline void clearMemory(void *ptr, unsigned int bits)
 {
     memzero(ptr, BIT(bits));
-    /* no cleaning of caches necessary on IA-32 */
 }
 
 /* Initialises MSRs required to setup sysenter and sysexit */
--- a/src/arch/arm/32/object/objecttype.c
+++ b/src/arch/arm/32/object/objecttype.c
@@ -392,6 +392,9 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
             /** GHOSTUPD: "(True, gs_new_frames vmpage_size.ARMSmallPage
                                                     (ptr_val \<acute>regionBase)
                                                     (unat ARMSmallPageBits))" */
+            cleanCacheRange_RAM((word_t)regionBase,
+                                (word_t)regionBase + MASK(pageBitsForSize(ARMSmallPage)),
+                                addrFromPPtr(regionBase));
         }
         return cap_small_frame_cap_new(
                    ASID_LOW(asidInvalid), VMReadWrite,
@@ -415,6 +418,9 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
             /** GHOSTUPD: "(True, gs_new_frames vmpage_size.ARMLargePage
                                                     (ptr_val \<acute>regionBase)
                                                     (unat ARMLargePageBits))" */
+            cleanCacheRange_RAM((word_t)regionBase,
+                                (word_t)regionBase + MASK(pageBitsForSize(ARMLargePage)),
+                                addrFromPPtr(regionBase));
         }
         return cap_frame_cap_new(
                    ARMLargePage, ASID_LOW(asidInvalid), VMReadWrite,
@@ -444,6 +450,9 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
             /** GHOSTUPD: "(True, gs_new_frames vmpage_size.ARMSection
                                             (ptr_val \<acute>regionBase)
                                             (unat ARMSectionBits))" */
+            cleanCacheRange_RAM((word_t)regionBase,
+                                (word_t)regionBase + MASK(pageBitsForSize(ARMSection)),
+                                addrFromPPtr(regionBase));
         }
         return cap_frame_cap_new(
                    ARMSection, ASID_LOW(asidInvalid), VMReadWrite,
@@ -473,6 +482,9 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
             /** GHOSTUPD: "(True, gs_new_frames vmpage_size.ARMSuperSection
                                                 (ptr_val \<acute>regionBase)
                                                 (unat ARMSuperSectionBits))" */
+            cleanCacheRange_RAM((word_t)regionBase,
+                                (word_t)regionBase + MASK(pageBitsForSize(ARMSuperSection)),
+                                addrFromPPtr(regionBase));
         }
         return cap_frame_cap_new(
                    ARMSuperSection, ASID_LOW(asidInvalid), VMReadWrite,
@@ -487,7 +499,9 @@ cap_t Arch_createObject(object_t t, void *regionBase, word_t userSize, bool_t de
         /** AUXUPD: "(True, ptr_retyps 1
               (Ptr (ptr_val \<acute>regionBase) :: (pte_C[256]) ptr))" */
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
-
+        cleanCacheRange_PoU((word_t)regionBase,
+                            (word_t)regionBase + MASK(seL4_PageTableBits),
+                            addrFromPPtr(regionBase));
         return cap_page_table_cap_new(false, asidInvalid, 0,
                                       (word_t)regionBase);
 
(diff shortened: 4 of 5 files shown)


lsf37:
This will have minor verification impact, and I might need to reorder some of the verification annotations over the flush instructions, but I'd like to test and discuss first, before I do that.

lsf37:
I should add that these are all the instances of `clearMemory`:

-  1 instance in `Arch_finaliseCap` for revoking the log buffer frame — this is clearing memory in the global PT and only needs `cleanCacheRange_PoU`, which is already done manually there
- 3 instances in `boot.c` for the initial IPC buffer and `populate_bi_frame`. These are mapped cached by the kernel and can't be remapped by the user. If these ever switch to being created from a proper untyped cap, they will be safe because the untyped mechanism does the right thing.
- the 2 instances in reset_untyped that this change is mostly about

kent-mcleod:
Are there any options for considering also delaying actually clearing the memory until retype? I think most of the boot time ends up spent on clearing 2-3GiB of RAM when the first large untyped is retyped by userlevel.

lsf37:
> Are there any options for considering also delaying actually clearing the memory until retype? I think most of the boot time ends up spent on clearing 2-3GiB of RAM when the first large untyped is retyped by userlevel.

Not really unfortunately. Clearing memory used to happen at retype time, but was specifically moved to reset time so it can be made preemptable for bounding WCET.

lsf37:
Here some extremely scientific quick-and-dirty numbers. Time measured is end-to-end for

```sh
time mq.sh run -s <system> -c "All is well" -f images/sel4test-driver-image-arm-<system>
```

This is quite noise, includes network time for uploading the image and all the machine queue reboot shenanigans for a sel4test image that doesn't run any tests (just start up, complain there are no tests, shut down).

```none
tx2a 8GB, old:
real	1m12.824s
real	1m7.171s
real	1m7.187s

tx2a 8GB, new:
real	0m48.622s
real	0m48.542s
real	0m44.133s

odroidc4_1 4GB, old:
real	0m34.648s
real	0m33.769s
real	0m34.432s

odroidc4_1 4GB, new:
real	0m23.781s
real	0m24.201s
real	0m23.443s
```

For the tx2 with 8GB, we're about 20 solid seconds faster, for odroidc4 with 4GB a nicely proportional 10 sec. This is the time it no longer spends just sitting there waiting to start.

Looks like it's worth it.

kent-mcleod:
> Here some extremely scientific quick-and-dirty numbers. Time measured is end-to-end for

If you pipe the output through `ts -i %.s` you can get incremental timestamps printed between each line of text output.



> Not really unfortunately. Clearing memory used to happen at retype time, but was specifically moved to reset time so it can be made preemptable for bounding WCET.

`invokeUntyped_Retype` calls `resetUntypedCap` which can return early due to preemption, and then it calls `createNewObjects` which can't currently return early. Considering that the cacheFlush isn't moving into a different kernel invocation (which I incorrectly assumed), then it seems like there's limited benefit to moving clearing memory into createObject.

include/arch/arm/arch/machine.h
```diff
@@ -42,12 +42,10 @@ void cleanInvalidate_L1D(void);
 void cleanCaches_PoU(void);
 void cleanInvalidateL1Caches(void);
 
-/* Cleaning memory before user-level access */
+/* Cleaning memory before user-level access. Does not flush cache. */
 static inline void clearMemory(word_t *ptr, word_t bits)
```
axel-h, line 46: It seems the architecture specific `clearMemory()` is now just a wrapper around `memzero()` everywhere. We could drop it completely then to simplify things?

lsf37: I would prefer not to. `clearMemory()` is a machine interface function in the proofs and things have changed underneath a few times in the past. While it'd be nice if this was the last time, I doubt that this will be the case.

We could check back in a few years and if it has turned out to be stable, we can still move the interface.

lsf37:
Rebased and put cache ops consistently after the verification annotations, because that is where the spec requires extra ops in retype (the annotations "create" the object, and all other operations are batched in a separate function afterwards in the spec).

> `invokeUntyped_Retype` calls `resetUntypedCap` which can return early due to preemption, and then it calls `createNewObjects` which can't currently return early. Considering that the cacheFlush isn't moving into a different kernel invocation (which I incorrectly assumed), then it seems like there's limited benefit to moving clearing memory into createObject.

Yes, it's the same overall syscall. We would still be getting better average performance if we moved the clearing of memory together with the flushing because the reset part of the syscall always works on the entire untyped region, and the retype part only on the requested part, but we'd make WCET (potentially much) worse, because we'd be moving it to the non-preemptable part as you say.

So I think this is the best we can do without much larger internal changes.

> If you pipe the output through ts -i %.s you can get incremental timestamps printed between each line of text output.

TIL about `ts`, nice! Do we really need more accurate measurement or was this more for education? :-) I think the 20 sec for the board with largest RAM in the test are pretty convincing already.

Indanz (approved):

kent-mcleod:
> TIL about `ts`, nice! Do we really need more accurate measurement or was this more for education? :-) I think the 20 sec for the board with largest RAM in the test are pretty convincing already.

Well on some boards the dhcp time and link negotiation in uboot can take multiple seconds longer.

But it'd be worth having the exact line in seL4 test where the largest untypeds are first retyped where you can see the time drop significantly.  Especially if we need to come back to this in the future if anything unexpected happens

lsf37:
Ok, here some timings with `ts` -- it's actually more than 20s on the `tx2`:

```
- tx2 master:

0.000030 seL4 Test
0.000030 =========
0.000029 
25.187899 vspace_reserve_range_at@vspace.h:738 vspace is NULL
0.060415 create_reservations@elf.c:245 Failed to make reservation: 0x400000, 917504
0.000114 elf_reserve_regions_in_vspace@elf.c:438 Failed to create reservations
0.000026 sel4utils_elf_reserve@elf.c:465 Failed to reserve regions
0.000018 Starting test suite sel4test

- tx2 after:

0.000030 seL4 Test
0.000030 =========
0.000029 
1.331402 vspace_reserve_range_at@vspace.h:738 vspace is NULL
0.000184 create_reservations@elf.c:245 Failed to make reservation: 0x400000, 917504
0.000056 elf_reserve_regions_in_vspace@elf.c:438 Failed to create reservations
0.118881 sel4utils_elf_reserve@elf.c:465 Failed to reserve regions
0.000140 Starting test suite sel4test

- odroid c4 master:

0.000030 seL4 Test
0.000030 =========
0.000029 
0.000029 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 2147483648, error 1
5.744922 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 1073741824, error 1
2.862004 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 536870912, error 1
1.414178 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 268435456, error 1
0.437004 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 134217728, error 1
0.406553 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 67108864, error 1
0.000187 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 33554432, error 1
0.037588 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 16777216, error 1
0.000154 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 8388608, error 1
0.000037 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 4194304, error 1
0.000026 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 2097152, error 1
0.000025 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 1048576, error 1
0.000026 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 524288, error 1
0.000023 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 262144, error 1
0.000020 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 131072, error 1
0.138094 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 65536, error 1
0.000112 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 32768, error 1
0.000037 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 16384, error 1
0.000031 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 8192, error 1
0.000029 vspace_reserve_range_at@vspace.h:738 vspace is NULL
0.000029 create_reservations@elf.c:245 Failed to make reservation: 0x400000, 909312
0.000029 elf_reserve_regions_in_vspace@elf.c:438 Failed to create reservations
0.000028 sel4utils_elf_reserve@elf.c:465 Failed to reserve regions
0.000028 Starting test suite sel4test

- odroid c4 after:

0.000028 seL4 Test
0.000029 =========
0.000027 
0.000038 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 2147483648, error 1
0.307162 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 1073741824, error 1
0.306380 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 536870912, error 1
0.306752 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 268435456, error 1
0.000149 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 134217728, error 1
0.000043 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 67108864, error 1
0.000033 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 33554432, error 1
0.000031 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 16777216, error 1
0.000029 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 8388608, error 1
0.000028 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 4194304, error 1
0.000028 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 2097152, error 1
0.000028 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 1048576, error 1
0.000029 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 524288, error 1
0.000027 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 262144, error 1
0.000029 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 131072, error 1
0.000028 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 65536, error 1
0.000028 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 32768, error 1
0.000027 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 16384, error 1
0.000027 vka_alloc_object_at_maybe_dev@object.h:57 Failed to allocate object of size 8192, error 1
0.066836 vspace_reserve_range_at@vspace.h:738 vspace is NULL
0.000100 create_reservations@elf.c:245 Failed to make reservation: 0x400000, 909312
0.000014 elf_reserve_regions_in_vspace@elf.c:438 Failed to create reservations
0.000011 sel4utils_elf_reserve@elf.c:465 Failed to reserve regions
0.000010 Starting test suite sel4test
```

Verification is turning out to be a bit more fiddly than I thought, but AArch64 is done now. Only AArch32 missing.

lsf37:
Verification PR is now up at seL4/l4v#811

kent-mcleod (approved):

PR #1290, opened 2024-07-05
configs: add zynqmp and rpi4 to verified platforms

lsf37:
The AARCH64 config now also works for functional correctness for zcu102/zcu106 and rpi4.

Needs the corresponding l4v PR to be merged first.

### DIFF
--- a/configs/AARCH64_zynqmp_verified.cmake
+++ b/configs/AARCH64_zynqmp_verified.cmake
@@ -1,0 +1,26 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2022, Proofcraft Pty Ltd
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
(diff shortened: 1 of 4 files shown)


lsf37:
The style check failure is expected (these specific cmake files should be marked executable in contrast to our usual policy).

configs/AARCH64_zynqmp_verified.cmake
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2022, Proofcraft Pty Ltd
```
Xaphiosis, line 3: do we care about updating this to 2024 (same in other config)?

Xaphiosis (approved):
For the commit: "now also works for functional correctness *on* zcu102/zcu106 and rpi4"?

configs/AARCH64_zynqmp_verified.cmake
lsf37: There are no significant changes apart from the platform string, so I left it on the old date.

PR #1291, opened 2024-07-05
fix parameter for log buffer cleanup

axel-h:
I ran into this while reviewing https://github.com/seL4/seL4/pull/1289. It looks broken to pass the size here, as the function just expects the 2^n bit-size. This wipes the entire memory practically.
This came in with commit 16c34811.


lsf37 (approved):
Ok, wow. Looks like nobody has ever revoked the cap to the benchmark log buffer yet. Which actually makes sense, you'd probably not do that for a benchmarking setup, but still.

Does anybody know why we are not doing anything here for 64 bit systems? Does the log buffer work differently there?

PR #1292, opened 2024-07-09
bump autopep8 version

lsf37:
See also https://github.com/seL4/ci-actions/issues/329, our old version of autopep8 crashes on more recent python code. Layout should be stable.

- bump version of autopep8
- bump version of sel4-deps
- update maintainer email

### DIFF
--- a/tools/python-deps/setup.py
+++ b/tools/python-deps/setup.py
@@ -46,6 +46,6 @@ setup(
     url="https://sel4.systems",
     licence='BSD2',
     author='TrustworthySystems',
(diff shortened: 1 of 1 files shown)


lsf37:
There is also the question of bumping the `cmake-format` version, because this also crashes on some of our files, but that is less layout-stable, so I'm leaving it for later.

axel-h (approved):
This needs to be updated on https://pypi.org/project/sel4-deps then also.

lsf37:
Indeed. It looks like we never release version 0.4.0, actually. Will do that after merge.

PR #1294, opened 2024-07-11
sel4-deps: tighten deps; update instructions

lsf37:
- require python >= 3
- require pyyaml < 6, because cmake-format breaks with pyyaml >= 6
- update upload instructions
- provide long-form release description of PyPi
- bump version, because the < 6 requirement might technically break compatibility

### DIFF
--- a/tools/python-deps/setup.py
+++ b/tools/python-deps/setup.py
@@ -10,11 +10,16 @@ Setup script for dependency metapackage.
 
 To add a python dependency, add it to the DEPS list below.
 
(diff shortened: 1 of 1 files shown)


lsf37:
During publishing https://github.com/seL4/seL4/pull/1292 I noticed a bunch of things that have already been updated for `camkes-deps`, but not yet for `sel4-deps`. Adding those here.

Also noted that `pyyaml` >=6 breaks `cmake-format==0.4.5`, so added `<6` as additional requirement.

axel-h (approved):
looks good to me.

PR #1295, opened 2024-07-12
sel4-deps: cypthon >=3 breaks pyyaml < 6 at build

lsf37:
Require cython < 3, because >=3 breaks the install for pyyaml at build/install time for more recent python versions. The cython < 3 fix only works on pyyaml < 5.4, so we add < 5.4 there instead of < 6 (which was already required by cmake-format=0.4.5).

See also <https://github.com/yaml/pyyaml/issues/601>

sel4-deps 0.6.0 isn't published yet (yay for pypitest..), so no version bump.

This whole thing is getting somewhat frustrating, but this combination should at least work in a fresh install.

### DIFF
--- a/tools/python-deps/setup.py
+++ b/tools/python-deps/setup.py
@@ -35,7 +35,8 @@ DEPS = [
     'pyelftools',
     'sh',
     'pexpect',
-    'pyyaml>=5.1,<6',
+    'cython<3',  # build dependency for pyyaml 5; see https://github.com/yaml/pyyaml/issues/601
+    'pyyaml>=5.1,<5.4',  # <5.4 for cmake-format==0.4.5
     'jsonschema',
     'pyfdt',
     'cmake-format==0.4.5',


lsf37:
The better fix would be to upgrade `cmake-format` to something more recent, then we could drop a lot of the version requirements again.

Indanz:
> The better fix would be to upgrade `cmake-format` to something more recent, then we could drop a lot of the version requirements again.

How do you want to proceed? Still merge this first?

Indanz (approved):

lsf37:
We should probably hold off on this one since we decided to bump cmake-format and with that we won't need to restrict versions quiet so much.

lsf37:
Putting this on draft for now to keep it here until cmake-format is bumped. This PR should be closed once that is done and is working.

Indanz:
> Putting this on draft for now to keep it here until cmake-format is bumped. This PR should be closed once that is done and is working.

Has that been done?

lsf37:
Closing, because it is superseded by #1479

PR #1296, opened 2024-07-12
mcs: refactor awaken

michaelmcinerney:
This refactors awaken, providing an inline function for the while loop condition, and modifying tcbReleaseDequeue to now perform the entire loop body.

Since tcbReleaseDequeue will perform tcbReleaseRemove on the head of the release queue, the variable ksReprogram will be set to true within tcbReleaseRemove, and therefore, we do not need to set this variable separately within the loop body of awaken.

### DIFF
--- a/include/object/tcb.h
+++ b/include/object/tcb.h
@@ -103,7 +103,6 @@ void tcbDebugRemove(tcb_t *tcb);
 #ifdef CONFIG_KERNEL_MCS
 void tcbReleaseRemove(tcb_t *tcb);
 void tcbReleaseEnqueue(tcb_t *tcb);
-tcb_t *tcbReleaseDequeue(void);
 #endif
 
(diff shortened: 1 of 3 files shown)


michaelmcinerney:
Force pushed to hopefully fix errors

lsf37:
Ok, so we seem to have reached a place again where we almost reliably trigger the MULTICORE0004 hang on SMP+MCS+clang+release configs, even though this is just another equivalence transformation. This time only on the `zynqmp` boards, apparently.

I'm not entirely sure what to do about this for now. We've already declared SMP+clang unreliable, so we could just merge, because it's not really related to this PR, even though we do trigger the bug now more often.

lsf37 (approved):
I think we should merge this as is and sort out MULTICORE0004 separately.

Indanz (commented):
Looks good to me.

PR #1297, opened 2024-07-15
thread state: abandon replyObject NULL invariant

lsf37:
We no longer guarantee the invariant that the replyObject reference is `NULL` when the thread state is not `BlockedOnReceive` or `BlockedOnReply`.

It is likely that this invariant was true in the kernel so far, but proving it would require a new proof that the reference is already `NULL` for any `setThreadState` to a simple state such as `Running`, `Inactive`, `Restart`. This either means reasoning about the state the thread had before `setThreadSate`, or explicitly setting the reference to `NULL` more
often.

There are many of these `setThreadState` instances, and the benefit of maintaining the invariant is low. Not maintaining the invariant removes some state updates from low-level functions (called often) at the cost of adding some if-checks in higher-level functions (called less often).

### DIFF
--- a/include/fastpath/fastpath.h
+++ b/include/fastpath/fastpath.h
@@ -46,6 +46,8 @@ static inline void cancelIPC_fp(tcb_t *dest)
         endpoint_ptr_set_state(ep_ptr, EPState_Idle);
     }
 
+    /* we are in BlockedOnReceive, because fastpath_signal explicitly checks for it */
+    assert(thread_state_get_tsType(dest->tcbState) == ThreadState_BlockedOnReceive);
     reply_t *reply = REPLY_PTR(thread_state_get_replyObject(dest->tcbState));
     if (reply != NULL) {
         reply_unlink(reply, dest);
--- a/include/object/reply.h
+++ b/include/object/reply.h
@@ -13,12 +13,16 @@
 /* Unlink a reply from its tcb */
 static inline void reply_unlink(reply_t *reply, tcb_t *tcb)
 {
+    /* check that the tcb has a thread state with reply */
+    assert(thread_state_get_tsType(tcb->tcbState) == ThreadState_BlockedOnReceive ||
+           thread_state_get_tsType(tcb->tcbState) == ThreadState_BlockedOnReply);
+
     /* check the tcb and reply are linked correctly */
     assert(reply->replyTCB == tcb);
     assert(thread_state_get_replyObject(tcb->tcbState) == REPLY_REF(reply));
 
-    thread_state_ptr_set_replyObject(&tcb->tcbState, REPLY_REF(0));
     reply->replyTCB = NULL;
+    /* This means the value of the thread state reply reference no longer matters. */
     setThreadState(tcb, ThreadState_Inactive);
 }
 
--- a/src/object/endpoint.c
+++ b/src/object/endpoint.c
@@ -334,9 +334,11 @@ void cancelIPC(tcb_t *tptr)
         }
 
 #ifdef CONFIG_KERNEL_MCS
-        reply_t *reply = REPLY_PTR(thread_state_get_replyObject(tptr->tcbState));
-        if (reply != NULL) {
-            reply_unlink(reply, tptr);
+        if (thread_state_ptr_get_tsType(state) == ThreadState_BlockedOnReceive) {
+            reply_t *reply = REPLY_PTR(thread_state_get_replyObject(tptr->tcbState));
+            if (reply != NULL) {
+                reply_unlink(reply, tptr);
+            }
         }
 #endif
         setThreadState(tptr, ThreadState_Inactive);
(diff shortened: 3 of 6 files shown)


lsf37:
Putting this on top of #1296 to see if it it still triggers MULTICORE0004 on `zynqmp`

lsf37:
Took a few iterations, but is now ready for review.

src/object/endpoint.c
```diff
@@ -446,7 +450,7 @@ void cancelBadgedSends(endpoint_t *epptr, word_t badge)
             next = thread->tcbEPNext;
 #ifdef CONFIG_KERNEL_MCS
             /* senders do not have reply objects in their state, and we are only cancelling sends */
-            assert(REPLY_PTR(thread_state_get_replyObject(thread->tcbState)) == NULL);
+            assert(thread_state_get_tsType(thread->tcbState) == ThreadState_BlockedOnSend);
```
michaelmcinerney, line 453: Is the comment  here, or at least its first part, still something we would like to say?

michaelmcinerney (approved):
Thanks Gerwin. I was not aware of the invariant we had before, so this does make much more sense to me, and will be much easier for us to verify.

src/object/endpoint.c
lsf37: I think so, yes. The first part is arguing that we don't need to care about replyObject part of the thread state (because it is BlockedOnSend). This still applies without the invariant. Before it was arguing that the replyObject is NULL, but the same text still makes sense, I think.

Indanz (commented):
I don't like it: Overall the code becomes more complex, not less, and as far as  I can tell the TCB can be left with a stale reply pointer that can point to anything if the reply object was re-used.

I as a human can't easily verify that all the threadstate assertions are correct. What about `ThreadState_BlockedOnReply`?

Can't you just abandon the invariant without changing the code?

lsf37:
> I don't like it: Overall the code becomes more complex, not less, and as far as I can tell the TCB can be left with a stale reply pointer that can point to anything if the reply object was re-used.

You're only allowed to use the reply pointer when the thread state is `BlockedOnReceive` or `BlockedOnReply`, and when these states are entered, the pointer always needs to be set fresh. What the current code is trying to do is rely on an additional invariant that a field that shouldn't be accessed at all is NULL allowing you to infer that you are not in `BlockedOnReceive` or `BlockedOnReply`. It is much better to explicitly test for that than relying on other code having done cleanup properly.

But in any case the code doesn't show that this invariant holds, it may or may not do enough cleanup for this to be true (the asserts so far seem to be true, but that doesn't really mean anything, there are by far not enough of them to cover all cases, and even then they would not cover all possible executions). To actually prove this would be very high effort, because every time you go into `Restart` or `Running` or any of these simple states, you either have to know the pointer is already NULL or you have to set it to NULL -- none of the code does that, because most of the code that deals with simple states is old and didn't have to worry about reply pointers.
 
> I as a human can't easily verify that all the threadstate assertions are correct. What about `ThreadState_BlockedOnReply`?

`BlockedOnReply` can't happen in any of the code paths that needed change (they are either in `case` expressions that check for a small number of states or are about endpoint/notification queues that can't have TCBs with `BlockedOnReply` in them). 

> Can't you just abandon the invariant without changing the code?

No, the code currently (probably wrongly) relies on this invariant, that's why there are a few new `if`s. The change needed to actually properly rely on the invariant would introduce a lot more change (I haven't counted, but there are many times we set thread states to simple values -- they would all need something to change).

include/fastpath/fastpath.h
```diff
@@ -46,9 +46,11 @@ static inline void cancelIPC_fp(tcb_t *dest)
         endpoint_ptr_set_state(ep_ptr, EPState_Idle);
     }
 
-    reply_t *reply = REPLY_PTR(thread_state_get_replyObject(dest->tcbState));
-    if (reply != NULL) {
-        reply_unlink(reply, dest);
+    if (thread_state_get_tsType(dest->tcbState) == ThreadState_BlockedOnReceive) {
```
kent-mcleod, line 49: I think that this if could be an assert. cancelIPC_fp is only called when the dest tcb is  bound to the notification and is blocked on an ep (and not blocked on a notification), so the only possible state would be  ThreadState_BlockedOnReceive. If this code was enabled for x86 virtualization configurations then maybe it could be inside a guest, but there are other parts of the signal fastpath that would break in that case.

lsf37: I wasn't sure enough to turn this into an assert, good point. Will have another closer look.

lsf37: You are right, this can be an assert. In fact the only path leading to this function being called explicitly checks for `BlockedOnReceive` and returns early if that is not the case:

https://github.com/seL4/seL4/blob/c679fe77d6441cd412dd253d953ee833dd25d740/src/fastpath/fastpath.c#L618-L623

lsf37: (added assert and rebase)

include/object/reply.h
```diff
@@ -13,12 +13,16 @@
...
+    assert(thread_state_get_tsType(tcb->tcbState) == ThreadState_BlockedOnReceive ||
+           thread_state_get_tsType(tcb->tcbState) == ThreadState_BlockedOnReply);
+
     /* check the tcb and reply are linked correctly */
     assert(reply->replyTCB == tcb);
     assert(thread_state_get_replyObject(tcb->tcbState) == REPLY_REF(reply));
 
-    thread_state_ptr_set_replyObject(&tcb->tcbState, REPLY_REF(0));
     reply->replyTCB = NULL;
+    /* This means the value of the thread state reply reference no longer matters. */
```
Indanz, line 25: It goes against my sensibilities not to clear the pointer, even if no one is supposed to look at it. :-(

Indanz (approved):

PR #1302, opened 2024-07-22
mcs: refactor cancelAllIPC and cancelBadgedSends

michaelmcinerney:
This factors out code from cancelAllIPC and cancelBadgedSends by introducing the inline function restart_thread_if_no_fault

### DIFF
--- a/src/object/endpoint.c
+++ b/src/object/endpoint.c
@@ -372,6 +372,29 @@ void cancelIPC(tcb_t *tptr)
     }
 }
 
+#ifdef CONFIG_KERNEL_MCS
+static inline void restart_thread_if_no_fault(tcb_t *thread)
+{
+    if (seL4_Fault_get_seL4_FaultType(thread->tcbFault) == seL4_Fault_NullFault) {
+        setThreadState(thread, ThreadState_Restart);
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

Indanz (approved):

Indanz:
Can this be merged or does it need to wait for verification?

lsf37:
I think this still needs to wait for the corresponding l4v PR to be up. @michaelmcinerney is that correct?

lsf37:
(Currently the tests on master are broken from the AArch64 debug API PR, so we should hold off on merging code PRs until that is fixed)

michaelmcinerney:
> I think this still needs to wait for the corresponding l4v PR to be up. @michaelmcinerney is that correct?

I would be able to put together an l4v PR for this quite quickly, but I was waiting for the C to be merged first, maybe for no good reason. I'm quite sure that merging this wouldn't break tests for the `rt` branch on l4v, but I can put up the l4v PR over the next couple of days if we would like that first

lsf37:
Better to do them simultaneously, but with the new `mcs` manifest handling we can actually check, I've started a proof test on the PR now.

PR #1304, opened 2024-08-01
Print virtual address already mapped in user error

Ivan-Velickovic:
Useful for debugging in my experience.

### DIFF
--- a/src/arch/riscv/kernel/vspace.c
+++ b/src/arch/riscv/kernel/vspace.c
@@ -888,7 +888,7 @@ static exception_t decodeRISCVFrameInvocation(word_t label, word_t length,
         } else {
             /* check this vaddr isn't already mapped */
             if (unlikely(pte_ptr_get_valid(lu_ret.ptSlot))) {
-                userError("Virtual address (0x%lx) already mapped", vaddr);
+                userError("Virtual address (0x%"SEL4_PRIx_word") already mapped", vaddr);
                 current_syscall_error.type = seL4_DeleteFirst;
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

lsf37:
I approved a bit too quickly: yes, it is useful and we should add it, but are there other architectures that throw the same error or do they all allow over-mapping? If there are others, we should add it there as well.

Ivan-Velickovic:
> do they all allow over-mapping?

I thought they all allow over-mapping. I did a quick grep and could not see any other similar messages for pages themselves.

Ivan-Velickovic:
Yea from https://github.com/seL4/seL4/pull/967 it looks like only RISC-V prevents over-mapping.

lsf37:
> Yea from #967 it looks like only RISC-V prevents over-mapping.

You're right, good to merge after all.

src/arch/riscv/kernel/vspace.c
```diff
@@ -888,7 +888,7 @@ static exception_t decodeRISCVFrameInvocation(word_t label, word_t length,
         } else {
             /* check this vaddr isn't already mapped */
             if (unlikely(pte_ptr_get_valid(lu_ret.ptSlot))) {
-                userError("Virtual address already mapped");
+                userError("Virtual address (0x%lx) already mapped", vaddr);
```
axel-h, line 891:
```suggestion
                userError("Virtual address (0x%"SEL4_PRIx_word") already mapped", vaddr);
```

PR #1305, opened 2024-08-05
Update AOS to use v13.0.0

ralismark:
This PR simply merges all commits for v13.0.0 into the AOS branch. No additional changes are needed within the kernel (but one in AOS) for it to work with the new kernel version.

Also see:
- https://github.com/SEL4PROJ/AOS/pull/21


lsf37 (approved):

lsf37:
This does seem to fail kernel standalone compile tests for Arm platforms -- are you sure you want to merge it like this? (Fine from my side, since it is the AOS branch)

PR #1308, opened 2024-08-14
microkit: remove const qualifiers from domain variables

JE-Archer:
This PR removes the const qualifiers from domain variables. This will allow the Microkit tool to patch the kernel's domain schedule. See #1301 for more information.

Test with: https://github.com/seL4/sel4test/pull/126

### DIFF
--- a/include/model/statedata.h
+++ b/include/model/statedata.h
@@ -103,8 +103,8 @@ extern word_t ksWorkUnitsCompleted;
 extern irq_state_t intStateIRQTable[];
 extern cte_t intStateIRQNode[];
 
-extern const dschedule_t ksDomSchedule[];
-extern const word_t ksDomScheduleLength;
+extern dschedule_t ksDomSchedule[];
+extern word_t ksDomScheduleLength;
 extern word_t ksDomScheduleIdx;
 extern dom_t ksCurDomain;
 #ifdef CONFIG_KERNEL_MCS
--- a/src/config/default_domain.c
+++ b/src/config/default_domain.c
@@ -9,9 +9,9 @@
 #include <model/statedata.h>
 
 /* Default schedule. The length is in ms */
-const dschedule_t ksDomSchedule[] = {
+dschedule_t ksDomSchedule[] = {
     { .domain = 0, .length = 1 },
 };
 
-const word_t ksDomScheduleLength = ARRAY_SIZE(ksDomSchedule);
+word_t ksDomScheduleLength = ARRAY_SIZE(ksDomSchedule);
 


lsf37:
Since this will make simulation and hardware tests fail, could you make corresponding PR in `sel4test` (which we won't merge) that removes the `const` qualifier from the test schedule there?

With that, we can use `Test with` to make sure that the `microkit` branch keeps passing `sel4test`.

axel-h:
I wonder if `const volatile` is what to use them. These are still runtime constants, right?

lsf37:
> I wonder if `const volatile` is what to use them. These are still runtime constants, right?

They are still runtime constants, yes

Ivan-Velickovic:
I was doing some force-pushing and something strange happened which led to this. Apologies @JE-Archer I will fix in a sec.

midnightveil:
Q: Why does `const` need to be removed?

In the original PR, it says "For ELF patching to work, the variables can't be qualified const, as they currently are."

But what is the issue?

In C, `const` means that [they are not modifiable lvalues](https://en.cppreference.com/w/c/language/const), but (as brought up with `const volatile`) they can still be modified (see, e.g. cast of a `char *` to `const char *` being valid, but **not** the other way around). It just means you can't modify a `const` variable.
Also, as declared here, they have [external linkage](https://en.cppreference.com/w/c/language/storage_duration), so they need to have allocated storage (e.g. if we compiled seL4 in multiple translation units).
They will be placed in the rodata segment, but that's fine, since we only modify them in the ELF at compile-ish-time, not at runtime?

So from a purely theoretical perspective, there shouldn't be an issue with patching the value of the const domain schedule? What then was the issue that caused this to be needed?

Indanz:
> So from a purely theoretical perspective, there shouldn't be an issue with patching the value of the const domain schedule? What then was the issue that caused this to be needed?

Probably the compiler optimising the accesses away altogether.

There is a project in the pipeline to make make domain schedules configurable via a syscall, that would solve this problem better than load-time patching by the Microkit loader (see comments in #1301).

lsf37:
Leaving this open so it can get tested until there is a proper fix. 

We do not intend to merge this PR, instead we're working on an API change to allow the domain schedule to change at runtime (properly authorised by cap). Will set the PR to draft to indicate that we don't want to merge.

lsf37:
It looks like I'm not allowed to set this to draft, because maintainer edits are off. @JE-Archer can you please set "Convert to draft" (where the reviews are requested).

> I've done this (see [seL4/sel4test#126](https://github.com/seL4/sel4test/pull/126)). How would I go about changing which seL4test branch this PR is tested with?

You can add "Test with: seL4/sel4test#126" in the PR description. I've done that now.

PR #1317, opened 2024-09-09
boot: fix init_freemem skipping reserved regions

FennelFoxxo:
In cases where the kernel is passed a very fragmented map of available memory (specifically when booting from GRUB on x86), it is possible for multiple available regions to overlap with a single reserved region.

Currently, if there is an overlap (where `avail_reg[a].start >= reserved[r].start`), the available region is trimmed and the region index is incremented. This causes any other overlapping regions to be handled as if they are non-overlapping, since it's now looking at the next reserved region. This eventually leads to reserved memory being passed to the root task and can cause corruption issues if it is used.

It is fine to skip this index increment since the available region is now either empty or non-overlapping, and both cases are already trivially handled.

[Here's a link to the thread about the issue, which has more information about what triggers the bug](https://sel4.discourse.group/t/kernel-triple-fault-when-retyping-untyped-memory-on-x86-qemu/874)

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -970,8 +970,7 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
                 /* the region overlaps with the start of the available region.
                  * trim start of the available region */
                 avail_reg[a].start = MIN(avail_reg[a].end, reserved[r].end);
-                reserve_region(pptr_to_paddr_reg(reserved[r]));
-                r++;
+                /* do not increment reserved index here - there could be more overlapping regions */
             } else {
                 assert(reserved[r].start < avail_reg[a].end);
                 /* take the first chunk of the available region and move
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -980,8 +980,8 @@ BOOT_CODE bool_t init_freemem(word_t n_available, const p_region_t *available,
                 insert_region(m);
                 if (avail_reg[a].end > reserved[r].end) {
                     avail_reg[a].start = reserved[r].end;
-                    reserve_region(pptr_to_paddr_reg(reserved[r]));
-                    r++;
+                    /* we could increment reserved index here, but it's more consistent with the
+                     * other overlapping case if we don't */
                 } else {
                     a++;
                 }


Indanz (approved):
Good find!

I'd also remove the same on lines 984-985 for consistency's sake. Even though it's probably okay over there, it simplifies the code if we remove it.

Indanz:
Could you please fix the Gitlint errors?

FennelFoxxo:
Not really sure what's going on with the tests - Gitlint status has been "Expected - Waiting for status to be reported" for almost a week now. It's possible it has to do with the DCO test - my "Signed-off-by" signature is over 70 characters and had to be split over two lines, but now it's not recognized as a valid signature. I don't have experience with DCO or Gitlint, but unless a Gitlint exception is made for the signature, I'm not sure how to get both tests to pass.

Indanz:
I had to press a button to approve the run now, not sure why.

DCO is more important than gitlint, so if the signed-off is longer than 70 characters, so be it. We can still merge it (or at least, some of us with enough permissions, depending on which check fails).

Indanz:
I think it's part of the unverified boot code, so it's probably okay, but good to have confirmation from a proof's person.

Indanz:
@FennelFoxxo, can you do the signed-offs with your real name (without any email if you want) instead of the Github email? A bit overkill for removing lines of code, but if you do real changes, then legally it's necessary to know who you are for copyright reasons.

@Xaphiosis Could you please confirm that this doesn't touch verified code? (And explain how this bug got through if it is ;-)

FennelFoxxo:
Ok, updated the commits to reflect my name. I just changed the email to my personal one, which should also satisfy the gitlint line length requirement. Not sure why the DCO test failed again because the sign off is the same format, but I think legally it should be fine.

Indanz:
I found and option to manually make the DCO check pass, so I did that.

The reason it still fails is because your Github identity doesn't match your signed-off, so the check thinks you're different people. Maybe adding your real email to Github in settings fixes it.

FennelFoxxo:
Thank you! I think it was because I (apparently) have email privacy turned on, so even though my real email is on github, maybe gitlint only sees the github email? It's not a big deal I suppose, but something to keep in mind for the future. Let me know if there's any more changes I need to make on my side!

Indanz:
The PC99_verification_SMP_MCS_gcc_32 failed, but without printing any extra info:

```
  <testcase classname="sel4test" name="SCHED0014">
    	<failure type="failure">result == SUCCESS at line 296 of file /github/workspace/projects/sel4test/apps/sel4test-driver/src/testtypes.c</failure>
    </testcase>
```
Which is weird and unexpected, but most likely unrelated to your change. I've restarted to test to see if it fails again. Seems unlikely, as it didn't fail before.

Verification had enough time to confirm whether this causes a problem for them, so I'm going to merge it when the test passes, as the current code is clearly buggy. If it does break something, they hopefully just have to remove part of the proof. The only thing that your change makes harder is proving guaranteed forward progress in the loop, but I don't know if they do that or not.

PR #1318, opened 2024-09-13
aarch64/vcpu: refine guest vcpu context

### LINKED ISSUE #1212: PAR_EL1 usage (at handleVMFault) in the kernel leaks info to the vCPU
AT instructions uses the PAR_EL1 to read the result translation from a VA to PA.

First problem: Kernel at VMFault uses PAR_EL1 and propagates the result to user space. The problem here is that the kernel overwrites the PAR_EL1 of the guest, so it is required to temporarily saved the PAR_EL1 for kernel usage. We don't want the vCPU to know what the kernel is doing.

Second problem: The problem is that between AT and PAR_EL1 the vCPU execution can be preempted and the result might be invalid, this is a functional problem. To solve this one we need to save and restore PAR_EL1 at vCPU_switch().

Let me know if I overlooked something. Thanks.

JorgeMVP:
These commits:
* provides support to save and restore par_el1 between multiple vCPUs on the same pCPU
* deletes isr_el1 from the context since it is read-only. It could be that the user-space can spy the kernel using the register.

Parts that might be missing, but would like feedback from reviewers:
- aarch32 support for par save and restore
- vcpu/kernel switch we might want the kernel not to interfere with execution of the vcpu. (see https://github.com/seL4/seL4/issues/1212)

### DIFF
--- a/include/arch/arm/arch/32/mode/machine.h
+++ b/include/arch/arm/arch/32/mode/machine.h
@@ -193,6 +193,18 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+static inline word_t readPAR(void)
+{
+    word_t val = 0;
+    asm volatile("mrc p15, 0, %0, c7, c4, 0":"=r"(val):);
+    return val;
+}
+
+static inline void writePAR(word_t val)
+{
+    asm volatile("mcr p15, 0, %0, c7, c4, 0":: "r"(val));
+}
+
 static inline void writeTPIDRURW(word_t reg)
 {
     asm volatile("mcr p15, 0, %0, c13, c0, 2" :: "r"(reg));
--- a/include/arch/arm/arch/32/mode/machine.h
+++ b/include/arch/arm/arch/32/mode/machine.h
@@ -193,15 +193,57 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
-static inline word_t readPAR(void)
+/* PAR is always a 64-bit register and needs to be context switched, because it
+ * can be written to directly in EL1.
+ *
+ * Ignored for verification, because inline asm formalisation cannot deal
+ * with 64-bit asm block output on 32-bit architectures, nor with two
+ * outputs of the same asm block. Read op is side-effect free.
+ */
+/** MODIFIES: */
+/** DONT_TRANSLATE */
+static inline uint64_t readPAR_64(void)
 {
+    uint64_t val = 0;
+    asm volatile("mrrc p15, 0, %Q0, %R0, c7":"=r"(val):);
+    return val;
+}
+
+/* Ignored for verification, because inline asm formalisation cannot deal
+ * with 64-bit asm block output on 32-bit architectures, nor with two
+ * outputs of the same asm block.
+ */
+/** MODIFIES: [*] */
+/** DONT_TRANSLATE */
+static inline void writePAR_64(uint64_t val)
+{
+    asm volatile("mcrr p15, 0, %Q0, %R0, c7":: "r"(val));
+}
+
+static inline word_t readPARhigh(void)
+{
+    uint64_t par = readPAR_64();
+    return (word_t)(par >> 32);
+}
+
+static inline word_t readPARlow(void)
+{
+    /* use 32-bit read to get only the low bits */
     word_t val = 0;
     asm volatile("mrc p15, 0, %0, c7, c4, 0":"=r"(val):);
     return val;
 }
 
-static inline void writePAR(word_t val)
+static inline void writePARhigh(word_t val)
 {
+    uint64_t par_high = (uint64_t) val << 32;
+    uint64_t par_low = readPARlow();
+    writePAR_64(par_high | par_low);
+}
+
+static inline void writePARlow(word_t val)
+{
+    /* avoid read-modify-write by using 32-bit write directly */
     asm volatile("mcr p15, 0, %0, c7, c4, 0":: "r"(val));
 }
 
--- a/include/arch/arm/arch/32/mode/machine_pl2.h
+++ b/include/arch/arm/arch/32/mode/machine_pl2.h
@@ -121,9 +121,15 @@ static inline void invalidateHypTLB(void)
 static inline paddr_t PURE addressTranslateS1(vptr_t vaddr)
 {
     uint32_t ipa0, ipa1;
+    uint32_t saved_ipa0, saved_ipa1;
+
+    asm volatile("mrrc p15, 0, %0, %1, c7"   : "=r"(saved_ipa0), "=r"(saved_ipa1));
+
     asm volatile("mcr  p15, 0, %0, c7, c8, 0" :: "r"(vaddr));
     isb();
     asm volatile("mrrc p15, 0, %0, %1, c7"   : "=r"(ipa0), "=r"(ipa1));
+
+    asm volatile("mcrr p15, 0, %0, %1, c7"   : "=r"(saved_ipa0), "=r"(saved_ipa1));
 
     return ipa0;
 }
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -380,5 +380,18 @@ void arch_clean_invalidate_L1_caches(word_t type);
 
 static inline paddr_t addressTranslateS1(vptr_t vaddr)
 {
-    return ats1e1r(vaddr);
+    word_t result;
+    UNUSED word_t saved_par;
+
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MRS("par_el1", saved_par);
+    }
+
+    result = ats1e1r(vaddr);
+
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MSR("par_el1", saved_par);
+    }
+
+    return result;
 }
--- a/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
@@ -75,6 +75,7 @@ typedef enum {
     seL4_VCPUReg_SCTLR = 0,
     seL4_VCPURegSaveRange_start, /* begin vcpu save/restore reg range */
     seL4_VCPUReg_ACTLR = seL4_VCPURegSaveRange_start,
+    seL4_VCPUReg_PAR,
     seL4_VCPUReg_TTBCR,
     seL4_VCPUReg_TTBR0,
     seL4_VCPUReg_TTBR1,
--- a/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
@@ -75,7 +75,6 @@ typedef enum {
     seL4_VCPUReg_SCTLR = 0,
     seL4_VCPURegSaveRange_start, /* begin vcpu save/restore reg range */
     seL4_VCPUReg_ACTLR = seL4_VCPURegSaveRange_start,
-    seL4_VCPUReg_PAR,
     seL4_VCPUReg_TTBCR,
     seL4_VCPUReg_TTBR0,
     seL4_VCPUReg_TTBR1,
@@ -112,7 +111,9 @@ typedef enum {
     seL4_VCPUReg_SPSRund,
     seL4_VCPUReg_SPSRirq,
     seL4_VCPUReg_SPSRfiq,
-    seL4_VCPURegSaveRange_end = seL4_VCPUReg_SPSRfiq, /* end vcpu save/restore reg range */
+    seL4_VCPUReg_PARhigh,
+    seL4_VCPUReg_PARlow,
+    seL4_VCPURegSaveRange_end = seL4_VCPUReg_PARlow, /* end vcpu save/restore reg range */
     seL4_VCPUReg_CNTV_CTL,
     seL4_VCPUReg_CNTV_CVALhigh,
     seL4_VCPUReg_CNTV_CVALlow,
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
@@ -83,6 +83,7 @@ typedef enum {
     seL4_VCPUReg_AMAIR,
     seL4_VCPUReg_CIDR,
     seL4_VCPUReg_ACTLR,
+    seL4_VCPUReg_PAR,
 
     /* exception handling registers EL1 */
     seL4_VCPUReg_AFSR0,
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
@@ -83,13 +83,13 @@ typedef enum {
     seL4_VCPUReg_AMAIR,
     seL4_VCPUReg_CIDR,
     seL4_VCPUReg_ACTLR,
-    seL4_VCPUReg_PAR,
 
     /* exception handling registers EL1 */
     seL4_VCPUReg_AFSR0,
     seL4_VCPUReg_AFSR1,
     seL4_VCPUReg_ESR,
     seL4_VCPUReg_FAR,
+    seL4_VCPUReg_PAR,
     seL4_VCPUReg_VBAR,
 
     /* thread pointer/ID registers EL0/EL1 */
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
@@ -89,7 +89,6 @@ typedef enum {
     seL4_VCPUReg_AFSR1,
     seL4_VCPUReg_ESR,
     seL4_VCPUReg_FAR,
-    seL4_VCPUReg_ISR,
     seL4_VCPUReg_VBAR,
 
     /* thread pointer/ID registers EL0/EL1 */
--- a/include/arch/arm/armv/armv7-a/armv/vcpu.h
+++ b/include/arch/arm/armv/armv7-a/armv/vcpu.h
@@ -416,6 +416,8 @@ static word_t vcpu_hw_read_reg(word_t reg_index)
         return getSCTLR();
     case seL4_VCPUReg_ACTLR:
         return getACTLR();
+    case seL4_VCPUReg_PAR:
+        return readPAR();
     case seL4_VCPUReg_TTBCR:
         return readTTBCR();
     case seL4_VCPUReg_TTBR0:
@@ -513,6 +515,9 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
         break;
     case seL4_VCPUReg_ACTLR:
         setACTLR(reg);
+        break;
+    case seL4_VCPUReg_PAR:
+        writePAR(reg);
         break;
     case seL4_VCPUReg_TTBCR:
         writeTTBCR(reg);
--- a/include/arch/arm/armv/armv7-a/armv/vcpu.h
+++ b/include/arch/arm/armv/armv7-a/armv/vcpu.h
@@ -416,8 +416,10 @@ static word_t vcpu_hw_read_reg(word_t reg_index)
         return getSCTLR();
     case seL4_VCPUReg_ACTLR:
         return getACTLR();
-    case seL4_VCPUReg_PAR:
-        return readPAR();
+    case seL4_VCPUReg_PARhigh:
+        return readPARhigh();
+    case seL4_VCPUReg_PARlow:
+        return readPARlow();
     case seL4_VCPUReg_TTBCR:
         return readTTBCR();
     case seL4_VCPUReg_TTBR0:
@@ -516,8 +518,11 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
     case seL4_VCPUReg_ACTLR:
         setACTLR(reg);
         break;
-    case seL4_VCPUReg_PAR:
-        writePAR(reg);
+    case seL4_VCPUReg_PARhigh:
+        writePARhigh(reg);
+        break;
+    case seL4_VCPUReg_PARlow:
+        writePARlow(reg);
         break;
     case seL4_VCPUReg_TTBCR:
         writeTTBCR(reg);
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -108,7 +108,6 @@
 #define REG_AFSR1_EL1       "afsr1_el1"
 #define REG_ESR_EL1         "esr_el1"
 #define REG_FAR_EL1         "far_el1"
-#define REG_ISR_EL1         "isr_el1"
 #define REG_VBAR_EL1        "vbar_el1"
 #define REG_TPIDR_EL1       "tpidr_el1"
 #define REG_SP_EL1          "sp_el1"
@@ -285,16 +284,6 @@ static inline void writeFAR(word_t reg)
     MSR(REG_FAR_EL1, reg);
 }
 
-/* ISR is read-only */
-/** MODIFIES: */
-/** DONT_TRANSLATE */
-static inline word_t readISR(void)
-{
-    uint32_t reg;
-    MRS(REG_ISR_EL1, reg);
-    return (word_t)reg;
-}
-
 static inline word_t readVBAR(void)
 {
     word_t reg;
@@ -462,8 +451,6 @@ static word_t vcpu_hw_read_reg(word_t reg_index)
         return readESR();
     case seL4_VCPUReg_FAR:
         return readFAR();
-    case seL4_VCPUReg_ISR:
-        return readISR();
     case seL4_VCPUReg_VBAR:
         return readVBAR();
     case seL4_VCPUReg_TPIDR_EL1:
@@ -532,9 +519,6 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
         break;
     case seL4_VCPUReg_FAR:
         writeFAR(reg);
-        break;
-    case seL4_VCPUReg_ISR:
-        /* ISR is read-only */
         break;
     case seL4_VCPUReg_VBAR:
         writeVBAR(reg);
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -114,6 +114,7 @@
 #define REG_ELR_EL1         "elr_el1"
 #define REG_SPSR_EL1        "spsr_el1"
 #define REG_CPACR_EL1       "cpacr_el1"
+#define REG_PAR_EL1         "par_el1"
 #define REG_CNTV_TVAL_EL0   "cntv_tval_el0"
 #define REG_CNTV_CTL_EL0    "cntv_ctl_el0"
 #define REG_CNTV_CVAL_EL0   "cntv_cval_el0"
@@ -344,6 +345,18 @@ static inline void writeCPACR_EL1(word_t reg)
     MSR(REG_CPACR_EL1, reg);
 }
 
+static inline word_t readPAR_EL1(void)
+{
+    word_t reg;
+    MRS(REG_PAR_EL1, reg);
+    return reg;
+}
+
+static inline void writePAR_EL1(word_t reg)
+{
+    MSR(REG_PAR_EL1, reg);
+}
+
 static inline word_t readCNTV_TVAL_EL0(void)
 {
     word_t reg;
@@ -443,6 +456,8 @@ static word_t vcpu_hw_read_reg(word_t reg_index)
         return readACTLR();
     case seL4_VCPUReg_CPACR:
         return readCPACR_EL1();
+    case seL4_VCPUReg_PAR:
+        return readPAR_EL1();
     case seL4_VCPUReg_AFSR0:
         return readAFSR0();
     case seL4_VCPUReg_AFSR1:
@@ -507,6 +522,9 @@ static void vcpu_hw_write_reg(word_t reg_index, word_t reg)
         break;
     case seL4_VCPUReg_CPACR:
         writeCPACR_EL1(reg);
+        break;
+    case seL4_VCPUReg_PAR:
+        writePAR_EL1(reg);
         break;
     case seL4_VCPUReg_AFSR0:
         writeAFSR0(reg);


Indanz:
>     * deletes isr_el1 from the context since it is read-only. It could be that the user-space can spy the kernel using the register.

To elaborate, ISR_EL1 shows the physical interrupt status when executed at EL2 and the virtual interrupt status when read by EL1 (because we set HCR.AMO, IMO and FMO). To get the virtual interrupt status in EL2, `HCR.VA`, `HCR.VI` and `HCR.VF` should be checked.

Removing the register altogether is a breaking change, I think we should either return the HCR bits or 0 instead.

Indanz (approved):

Indanz (approved):
Please add the PAR_EL1 saving and restoring the current PAR_EL1 in the ats* functions in machine.h too, otherwise the solution is incomplete.

Aarch32 support would be nice too of course, but doesn't need to be part of this PR.

Indanz (approved):

lsf37:
This looks fine form my side, but it is a breaking API change and will need a short RFC first.

Indanz:
Requiring an RFC for a clear bugfix like this seems a bit heavy handed, but because the public API is tied to the kernel implementation, it does break the API.

To create an RFC make a PR for https://github.com/seL4/rfcs, see the README on that page.

lsf37:
Yeah, ok, it's borderline, you're right that it's not really a new feature. I'm Ok to do this without RFC unless some from the @seL4/tsc objects. @yyshen do you have an opinion on the technical side of this one?

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -350,28 +350,61 @@ static inline word_t PURE getFAR(void)
 static inline word_t ats1e2r(word_t va)
 {
     word_t par;
+
+    UNUSED word_t saved_par;
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MRS("par_el1", saved_par);
+        isb();
```
Indanz, line 357: Is this `isb` necessary? I understand the one before reading PAR, but surely the `at` isn't overtaking the PAR read?

If it is necessary, isn't an `isb` also necessary before writing the saved value back?

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -350,28 +350,61 @@ static inline word_t PURE getFAR(void)
...
+    }
+
     asm volatile("at s1e1r, %0" :: "r"(va));
     isb();
     MRS("par_el1", par);
+
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MSR("par_el1", saved_par);
+    }
```
Indanz, line 386: Could you add two inline functions that are defined to nothing when `!CONFIG_ARM_HYPERVISOR_SUPPORT` to avoid the code repetition? That way you can just add calls to those to all functions and make them nicer to read, e.g.

```
word_t saved_par = arm_hyp_save_par();
...
arm_hyp_restore_par(saved_par);
```

include/arch/arm/arch/64/mode/machine.h
JorgeMVP: I dont find anything that backs me up here, but I'm afraid the CPU can execute this in a different order, and consequently change what we want to do.

I added it defensively, I tried to look at the spec, but didn't find anything. I'm okay if we drop it, we just might get exposed to something I'm not sure.

@Indanz I'm open to remove it, if you think it is the best.

include/arch/arm/arch/64/mode/machine.h
JorgeMVP: @Indanz this is good suggestion.

Other alternative would be instead of adding it in the at function, we could add it in a higher level function. See (https://github.com/seL4/seL4/pull/1318/commits/a089a055531615d9b02d6427437cc1dfd7d98573)

also we need to make sure that future uses of this AT functions in other places in the kernel always save and restore PAR accordinly for the hypervisor mode.

include/arch/arm/arch/64/mode/machine.h
Indanz: All I could find in official ARM documentation is that you need an ISB after writing to a system register (because it's a context altering operation). However, AT is special and it does need an ISB before the MSR for obvious reasons, but I didn't find clear documentation for that. Apple code doesn't has an ISB, but someone in the OpenWRT community found out they needed an ISB before reading PAR, or else it didn't work on their hardware.

All in all I agree with you that we should err on the side of caution. That means we should do an ISB between all operations that read or alter PAR. So we're missing another ISB before writing back the saved PAR, as we don't want it to be reordered with the MRS before it.

include/arch/arm/arch/64/mode/machine.h
Indanz: That's even better, that keeps those functions thin assembly wrappers.

I don't like how seL4 mixes external interface declarations/defines with internal stuff in header files. But basically no one outside of machine.h is using any of the ats* functions directly.

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -380,5 +380,18 @@ void arch_clean_invalidate_L1_caches(word_t type);
...
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MRS("par_el1", saved_par);
+        isb();
+    }
+
+    result = ats1e1r(vaddr);
+
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MSR("par_el1", saved_par);
+    }
```
Indanz, line 394:
```suggestion
    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
        /* Prevent re-ordering with the read of PAR above */
        isb();
        MSR("par_el1", saved_par);
    }
```

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -380,5 +380,20 @@ void arch_clean_invalidate_L1_caches(word_t type);
 
 static inline paddr_t addressTranslateS1(vptr_t vaddr)
 {
-    return ats1e1r(vaddr);
+    word_t result;
+    UNUSED word_t saved_par;
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        MRS("par_el1", saved_par);
+        isb();
```
Indanz, line 387:
```suggestion
        MRS("par_el1", saved_par);
        /* Prevent re-ordering with the AT below */
        isb();
```

Indanz (approved):

Indanz:
@JorgeMVP I updated the branch with rebase, so it will diverge from your local checkout. Apologies if I caused you any trouble.

JorgeMVP:
> @JorgeMVP I updated the branch with rebase, so it will diverge from your local checkout. Apologies if I caused you any trouble.

All good, no issues. Im looking at aarch32 support now.

include/arch/arm/arch/64/mode/machine.h
JorgeMVP: Added, however, still not clear if there is already a dependecy read then write it is similar to read modify write approach.

Indanz:
> Added, however, still not clear if there is already a dependecy read
> then write it is similar to read modify write approach.

The cpu only recognises such dependencies with normal registers, it 
doesn't know that AT affects PAR, otherwise we wouldn't need any ISBs 
around AT.

Whether it reorders system register writes and reads to the same system 
register is another matter, perhaps it's guaranteed not to do that, then 
we can drop one ISB. However, it does say you need an ISB after a system 
reg write in general, with a few exceptions.

include/arch/arm/arch/64/mode/machine.h
Indanz: > Added, however, still not clear if there is already a dependecy read then write it is similar to read modify write approach.

The cpu only recognises such dependencies with normal registers, it doesn't know that AT affects PAR, otherwise we wouldn't need any ISBs around AT.

Whether it reorders system register writes and reads to the same system register is another matter, perhaps it's guaranteed not to do that, then we can drop one ISB. However, it does say you need an ISB after a system reg write in general, with a few exceptions, which implies there are few guarantees.

Edit: Moved to here and expanded reply.

If you do a read, modify write operation on the same system reg a dependency is introduced via a normal register, so no ISB or anything is needed. E.g. `MRS x1 N`, `MSR N x1`.

On pipelined CPUs, to prevent a pipeline flush and having to wait the full e.g. 20 cycles till the first operation is finished before starting the dependent instruction, CPUs have a bypass network which forwards results quicker to the next, dependent instruction so there only need to be a few cycles delay before executing the dependent instruction. The next instruction gets its input register values from the bypass network instead of the register file. 

With out-of-order execution the CPU keeps track of all memory read/write dependencies needed for consistency in addition to all register dependencies, and that's where memory ordering comes in. But even with in-order execution, instructions could finish out-of-order because some instructions take longer than others, and this could be an issue with a MSR after an MRS.

All that to say that for normal instructions things are well-defined because of natural register dependencies and for memory accesses they take great care to define the model in the documentation, but for system registers things are fuzzy and I haven't found anything except the requirement for ISB after most MSR if you want the effect to be visible.

MRS reads can be re-ordered safely with each other because they don't have side-effects, but how are MSR treated in regard to MRS? Are they just executed independently and if there is an ordering requirement you have to either force it via normal registers or ISB? Or are they serialised and executed in-order in regard to each-other? Conceptually this would be treating all system register as one normal register.

include/arch/arm/arch/32/mode/machine_pl2.h
```diff
@@ -121,10 +121,23 @@ static inline void invalidateHypTLB(void)
 static inline paddr_t PURE addressTranslateS1(vptr_t vaddr)
 {
     uint32_t ipa0, ipa1;
+    UNUSED uint32_t saved_ipa0, saved_ipa1;
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
```
kent-mcleod, line 125: This section of this file is already within an `#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT` and there's another definition of `addressTranslateS1()` for non-hyp configs. So because this function is specifically for the hyp config these if guards shouldn't be added here.

kent-mcleod (approved):
No objections from me. Thanks for fixing this!

include/arch/arm/arch/32/mode/machine_pl2.h
JorgeMVP: Thanks, right I will remove them ;)

JorgeMVP: addressed.

include/arch/arm/arch/64/mode/machine.h
JorgeMVP: addressed

Indanz:
I think you should also mention the API changes in CHANGES.md.

Are you aware of any other system registers that share state between EL2 and El1?

include/arch/arm/arch/64/mode/machine.h
Indanz: Okay, I found some more relevant information. See _"Table D24-1 Synchronization requirements"_ in the _"D24.1.2.2 Synchronization requirements for AArch64 System registers"_ chapter. What we're dealing with here is a direct read followed by an indirect write caused by the `AT` instruction. According to the table, a direct or indirect write following a direct read does not require synchronisation.

However, this is only true if `AT` counts as an (in)direct write, and not as a memory operation. As far as I can tell, it doesn't count as a memory operation, even though it does a page table walk to get the result. If it counted as a memory operation, a DSB would also be required in addition to the ISB before reading `PAR_EL1`.

include/arch/arm/arch/64/mode/machine.h
Indanz: As described by the _"D24.1.2.2 Synchronization requirements for AArch64 System registers"_ chapter, this would count as a direct write following a direct read to the same system register, which does not require synchronisation, so my suggested ISB isn't necessary.

Direct system register writes and reads to the same system register are ordered. Synchronisation is only necessary for indirect read following any write, or any read following an indirect write.

There's some vague info about synchronisation requirements with memory mapped system registers and external agents too in that chapter.

include/arch/arm/arch/64/mode/machine.h
Indanz: More info in _"D8.2.13.3   Address translation instructions, synchronization requirements"_:

> IJTYVP:
> When an address translation instruction is executed, explicit synchronization is required to guarantee the result is visible to subsequent direct reads of PAR_EL1.
>      For more information, see Synchronization requirements for AArch64 System registers.
> IPCPWC:
> Requiring explicit synchronization after executing an address translation instruction is consistent with the general requirement that the effect of a write to a System register requires explicit synchronization before the result is guaranteed to be visible to subsequent instructions.

So I guess `AT` counts as an indirect write of `PAR_EL1` then.

include/arch/arm/arch/64/mode/machine.h
Indanz: So this `isb` isn't needed after all, because a direct read followed by an indirect write does not require synchronisation according to Table D24-1 Synchronization requirements.

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -380,5 +380,21 @@ void arch_clean_invalidate_L1_caches(word_t type);
...
+        /* Prevent re-ordering with the AT below */
+        isb();
+    }
+
+    result = ats1e1r(vaddr);
+
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
+        /* Prevent re-ordering with the read of PAR above */
+        isb();
+        MSR("par_el1", saved_par);
```
Indanz: So I was wrong and this `isb` isn't needed, because a direct write following a direct read doesn't require synchronisation.

include/arch/arm/arch/32/mode/machine_pl2.h
```diff
@@ -121,10 +121,20 @@ static inline void invalidateHypTLB(void)
 static inline paddr_t PURE addressTranslateS1(vptr_t vaddr)
 {
     uint32_t ipa0, ipa1;
+    uint32_t saved_ipa0, saved_ipa1;
+
+    asm volatile("mrrc p15, 0, %0, %1, c7"   : "=r"(saved_ipa0), "=r"(saved_ipa1));
+    /* Prevent re-ordering with the AT below */
+    isb();
```
Indanz: An indirect write following a direct read does not require synchronisation, see "Table G8-2 Synchronization requirements for updates to System registers" in chapter "G8.1 About the AArch32 System registers".

include/arch/arm/arch/32/mode/machine_pl2.h
```diff
@@ -121,10 +121,20 @@ static inline void invalidateHypTLB(void)
...
+
     asm volatile("mcr  p15, 0, %0, c7, c8, 0" :: "r"(vaddr));
     isb();
     asm volatile("mrrc p15, 0, %0, %1, c7"   : "=r"(ipa0), "=r"(ipa1));
 
+    /* Prevent re-ordering with the read of PAR above */
+    isb();
```
Indanz: Direct write following a direct read doesn't require synchronisation.

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,18 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+static inline word_t readPAR(void)
+{
+    word_t val = 0;
+    asm volatile("mrc p15, 0, %0, c7, c4, 0":"=r"(val):);
```
Indanz, line 233: In `addressTranslateS1()` it does `asm volatile("mrrc p15, 0, %0, %1, c7"   : "=r"(ipa0), "=r"(ipa1));`, shouldn't we do the same here?

For the PAR description, Arm says to use `MRC p15,0,<Rd>,c7,c4,0;` to read PAR, so this does seem correct. But what's going on in `addressTranslateS1()?

Why the difference in mrc versus mrrc and c4 versus c7? Is it 32-bit read versus 64-bit read because of long mode?

kent-mcleod: I think this needs to be the 64-bit access version (mrrc)

kent-mcleod: https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-system-control-operations-described-by-function/Performing-address-translation-operations?lang=en

When the [PAR](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-System-control-registers-descriptions--in-register-order/PAR--Physical-Address-Register--VMSA?lang=en) is a 64-bit register, 32-bit accesses to the [PAR](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-System-control-registers-descriptions--in-register-order/PAR--Physical-Address-Register--VMSA?lang=en), using MRC or MCR instructions, access the least-significant word of the [PAR](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-System-control-registers-descriptions--in-register-order/PAR--Physical-Address-Register--VMSA?lang=en).

The [PAR](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-System-control-registers-descriptions--in-register-order/PAR--Physical-Address-Register--VMSA?lang=en) is a read/write register, and software can perform a CP15 register write, using the same encodings, to write to the register. No translation operation requires writing to this register, but the write operation might be required to restore the [PAR](https://developer.arm.com/documentation/ddi0406/cb/System-Level-Architecture/System-Control-Registers-in-a-VMSA-implementation/VMSA-System-control-registers-descriptions--in-register-order/PAR--Physical-Address-Register--VMSA?lang=en) value after a context switch.

lsf37:
I'd like to do the verification updates for this PR this coming week. Ok if I rebase and add Indan's last set of review comments?

include/arch/arm/arch/32/mode/machine.h
lsf37: Ok, so if I understand this correctly, in both readPAR() and writePAR(), this register is 64 bit, even on 32-bit platforms, because we are working in hyp mode under LPAE. This means, not only do we need to change the instructions, we also need to change the function signature to take/return uint64 instead. Is this right?

lsf37: Which of course is not possible, because the code that uses it assumes registers are all `word_t`. Great. Do we do read/write PAR hi/low then?

lsf37: I've done a high/low split now as for CNTV_CVAL in https://github.com/seL4/seL4/commit/07233049fe85ed6f792ae7bfd56de9cbbf75d7fc

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,46 @@ static inline void writeTTBCR(word_t val)
...
+    /* use 32-bit read to get only the low bits */
+    word_t val = 0;
+    asm volatile("mrc p15, 0, %0, c7, c4, 0":"=r"(val):);
+    return val;
+}
+
+static inline void writePARhigh(word_t val)
+{
+    uint64_t par_high = (uint64_t) val << 32;
+    uint64_t par_low = (readPAR_64() << 32) >> 32;
```
midnightveil: This should use readPARHigh and readParlow, surely?

lsf37: yes, good point, I should be using `readPARlow` here. I was going off the `read/write_cntv_cval_high` pattern first, but then decided to leave in the 32-bit accesses for the low part and missed this one.

lsf37: done.

midnightveil: Is there a particular reason why you do `(64_bit << 32) >> 32`?

I normally would do this as a `& 0xffffffff` to just mask-off  the high bits.

lsf37: I mainly copy/pasted from `set_cntv_cval_low()` and I was about to do the same as you suggest (`& MASK(32)`), but then left it because I thought someone with more asm experience then me will have had their reasons to write it this way :-). Is it possible that two shifts are slightly faster than a mask?

lsf37: Ok, it's not. If I change it, I get exactly the same binary. So maybe we should change it all to `& MASK(32)` and `& ~MASK(32)`

midnightveil: Yeah, I suspect the compiler optimises it to the mask anyway. (barrel-shifters are somewhat expensive, and it's two instructions not one).

EDIT: https://godbolt.org/z/TE4qx16Wf oh, it doesn't even need to do the mask it seems.

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,46 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+/* hyp mode uses LPAE, which makes PAR a 64-bit register */
+static inline uint64_t readPAR_64(void)
+{
+    uint64_t val = 0;
+    asm volatile("mrrc p15, 0, %Q0, %R0, c7":"=r"(val):);
```
Indanz, line 208: I've never seen this syntax before, what do Q and R mean? It's documented for other archs, but not for ARM in the GCC documentation. Not sure we should use undocumented features like this, especially if manually combining two 32-bit values generates the same code, which it does for me for:

```c
    uint32_t low, high;
    asm volatile("mrrc p15, 0, %0, %1, c7" : "=r"(low), "=r"(high));
    return ((uint64_t)high << 32) | low;
```

But [Arm trusted firmware](https://elixir.bootlin.com/arm-trusted-firmware/v1.3/source/include/lib/aarch32/arch_helpers.h#L58) is also using it though, so oh well.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
```diff
@@ -83,13 +83,13 @@ typedef enum {
     seL4_VCPUReg_AMAIR,
     seL4_VCPUReg_CIDR,
     seL4_VCPUReg_ACTLR,
+    seL4_VCPUReg_PAR,
```
Indanz: I'd put it in the same place as `seL4_VCPUReg_ISR` was, that minimises the binary changes.

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
```diff
@@ -75,6 +75,8 @@ typedef enum {
     seL4_VCPUReg_SCTLR = 0,
     seL4_VCPURegSaveRange_start, /* begin vcpu save/restore reg range */
     seL4_VCPUReg_ACTLR = seL4_VCPURegSaveRange_start,
+    seL4_VCPUReg_PARhigh,
+    seL4_VCPUReg_PARlow,
```
Indanz: I would put this at the end of the saved registers instead of the start, that minimises the binary changes and seems like the right thing to do in general when adding new things.

Indanz (approved):
Just some last nitpicks, nothing major.

include/arch/arm/arch/32/mode/machine.h
lsf37: I like your version better, I wasn’t too sure about this either.

midnightveil: https://developer.arm.com/documentation/101754/0621/armclang-Reference/armclang-inline-assembler/Inline-assembly-template-modifiers/Template-modifiers-for-AArch32-state?lang=en 

(But yes looks undocumented for GCC: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=30527#c8)

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
lsf37: I was about to do that, but that is kind of under a wrong heading:

```
    /* exception handling registers EL1 */
    seL4_VCPUReg_AFSR0,
    seL4_VCPUReg_AFSR1,
    seL4_VCPUReg_ESR,
    seL4_VCPUReg_FAR,
    // previously seL4_VCPUReg_ISR,
    seL4_VCPUReg_VBAR,
 ```

Then again ISR isn't really exception handling as such either. I'll put it there for now and squash, but happy to move it again or change the comment.

File: include/arch/arm/arch/32/mode/machine.h (hunk 10 above)
Comment (lsf37) [reply include/arch/arm/arch/32/mode/machine.h]: yes, it's way too smart :-)

There is no instance of this pattern left in this PR, because I changed low accesses to directly use the 32 bit instructions. We could change the other instances of the pattern, but that should probably be separate.

File: include/arch/arm/arch/32/mode/machine.h [hunk 14]
```diff
@@ -193,6 +193,46 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+/* hyp mode uses LPAE, which makes PAR a 64-bit register */
```
Indanz: This comment is somewhat misleading. PAR is always a 64-bit register, the kernel is just not interested in the top 32-bits itself (not even on HYP because seL4 doesn't support physical addresses >= 2^32 on 32-bit platforms).

But with HYP we need to context switch the PAR register. If we don't do the top bits, then those can be used to pass information between VCPUs (because the top bits are directly writeable).

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
Indanz: Well, PAR is only used to lookup an address after a VM fault, so it is exception handling related in that regard.

include/arch/arm/arch/32/mode/machine.h
lsf37: You are right of course about why we want to context switch it, maybe I should add that to the comment.

What I meant with the comment is that on 32-bit platforms with short descriptor format, the top 32 bits are  RES0 and would not need to context switched for EL1:

https://arm.jonpalmisc.com/latest_sysreg/AArch32-par says

```
PAR is accessed as a 32-bit value:

- When the PE is not in Hyp mode and is using the Short-descriptor translation table format.
- When the PE is in Hyp mode and executes an ATS12NSOPR, ATS12NSOPW, ATS12NSOUR, or ATS12NSOUW instruction and the value of HCR.VM is 0 and the value of TTBCR.EAE is 0.

In these cases, PAR[63:32] is RES0.
```

(we are not in that mode, we are in the other one where the top bits are not RES0, even if we are not interested in them)

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,46 @@ static inline void writeTTBCR(word_t val)
...
+static inline uint64_t readPAR_64(void)
+{
+    uint32_t low, high;
+    asm volatile("mrrc p15, 0, %0, %1, c7" : "=r"(low), "=r"(high));
```
lsf37: It turns out the binary verification pre-pass (SimplExportAndRefine) is freaking out about this one. I need to have another look at what we do elsewhere for these cases.

lsf37: Specifically, we get `Can not handle multiple asm lval specifiers`.

I think this is the reason for this Q/R syntax that these all use:

https://github.com/seL4/seL4/blob/0fcbb71a2a0fb37dabef0e40c97afacc41f7cc71/include/arch/arm/arch/32/mode/machine.h#L49-L53

so that they can then use the `MRRC` and `MCRR` macros which make the asm block produce a single 64-bit value:

```c
static inline uint64_t get_cntv_off_64(void)
{
    uint64_t ret = 0;
    MRRC(CNTVOFF, ret);
    return ret;
}
```

I didn't like the macros (and still don't think they are necessary, we can just use the asm block), but that's how I got to the Q/R syntax. Is there a better way to achieve that or should we just go with what the rest already does after all?

lsf37: (also looked at what it would take to support multiple outputs for asm blocks, but it seems that would be fairly major and go all the way down to the actual binary toolchain)

include/arch/arm/arch/32/mode/machine.h
lsf37: I have updated the comment to be more explicit.

include/arch/arm/arch/32/mode/machine.h
lsf37: I've reverted it to the Q/R form I had in the beginning for now.

include/arch/arm/arch/32/mode/machine.h
Indanz: > What I meant with the comment is that on 32-bit platforms with short descriptor format, the top 32 bits are RES0 and would not need to context switched for EL1:

And I'm trying to explain why that assumption is wrong. The PAR page describes the content of the PAR register after an AT instruction, and for that it is relevant what mode is used. But EL1 can write any value it wants directly to PAR without AT instruction. See the part about MCRR at the bottom where `PAR = R[t2]:R[t];`.

Indanz: The part "PAR is accessed as a 32-bit value" means that the address translation instruction will access it as a 32-bit value, it doesn't mean PAR magically turning into 32-bits in size instead of 64-bits. I agree it's confusing, it's more talking about the AT instruction result format than the PAR register itself.

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,48 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+/* seL4 hyp mode uses LPAE, which means the top 32-bit of PAR are not RES0 and
+ * need to be context-switched on 32-bit platforms.
```
Indanz: Now this is even more misleading. :-(

The top bits would also need to be context switched if both seL4 hyp mode and the guest used the short descriptor format.

include/arch/arm/arch/32/mode/machine.h
lsf37: Ok, now I get it, and I did indeed have the wrong mental model of it. Will update.

include/arch/arm/arch/32/mode/machine.h
lsf37: I've updated it again, hopefully Ok this time :-)

lsf37: (There is now more there as a reminder to verification people why these have DONT_TRANSLATE annotations and the other ones not)

lsf37:
This is now ready to merge from the verification side. @Indanz are you happy with the updated comment?

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,60 @@ static inline void writeTTBCR(word_t val)
     asm volatile("mcr p15, 0, %0, c2, c0, 2":: "r"(val));
 }
 
+/* PAR always is a 64-bit register and needs to be context switched, because it
```
Indanz:
```suggestion
/* PAR is always a 64-bit register and needs to be context switched, because it
```

include/arch/arm/arch/32/mode/machine.h
```diff
@@ -193,6 +193,60 @@ static inline void writeTTBCR(word_t val)
...
+{
+    uint64_t val = 0;
+    asm volatile("mrrc p15, 0, %Q0, %R0, c7":"=r"(val):);
+    return val;
+}
+
+/* Ignored for verification, because inline asm formalisation cannot deal
+ * with 64-bit asm block output on 32-bit architectures, nor with two
+ * outputs of the same asm block.
+ */
```
Indanz, line 215: I think it would be better to explain why this is okay while the function is not side effect free, than to repeat the same text as above.

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -380,5 +380,17 @@ void arch_clean_invalidate_L1_caches(word_t type);
 
 static inline paddr_t addressTranslateS1(vptr_t vaddr)
 {
-    return ats1e1r(vaddr);
+    word_t result;
+    UNUSED word_t saved_par;
+    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
```
Indanz, line 386:
```suggestion
    word_t result;
    UNUSED word_t saved_par;

    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
```
and personally I'd remove the empty line before the return.

Indanz (approved):

include/arch/arm/arch/32/mode/machine.h
lsf37: The spec of this one doesn't say that it is side-effect free, it says that it can modify any unmodelled machine state (the`MODIFIES: [*]` bit).

Indanz: I missed that difference, `/** MODIFIES: */` is too similar to `/** MODIFIES: [*] */`. Wouldn't not specifying MODIFIES have the same effect? Something like MODIFIES: NOTHING would be nice, and maybe a clearer name for `[*]` too. That looks too much like it could modify anything. For modelled machine state we have `phantom_machine_state`, maybe add an `unmodelled_machine_state`?

Also, now that we save and restore PAR, shouldn't it actually be part of the modelled machine state?

lsf37: > I missed that difference, `/** MODIFIES: */` is too similar to `/** MODIFIES: [*] */`. Wouldn't not specifying MODIFIES have the same effect? 

No, if nothing is specified the parser will try do derive something, and in this case fail, because we told it to not translate it, so it will just be missing that assumption.

> Something like MODIFIES: NOTHING would be nice, and maybe a clearer name for `[*]` too. That looks too much like it could modify anything. For modelled machine state we have `phantom_machine_state`, maybe add an `unmodelled_machine_state`?

I was a bit imprecise. `[*]` does actually mean "potentially everything". We do constrain that later with an additional assumption to what is called `machines_state_rest` in the formalisation. `machine_state_rest` is the part of the machine state that we don't model further (apart from "it exists" and "it might change when you run one of these instructions" and "ASpec will change it in exactly the same way and order as C does").
 

> Also, now that we save and restore PAR, shouldn't it actually be part of the modelled machine state?

Should is a strong word. None of these registers are currently modelled. Only their existence, not what they do. That is a safer assumption to make (when it works) than modelling what they do and getting that wrong.

For this specific case, it's a bit weak, because we do actually use the result of this register on AArch64 for the fault return, and the formalisation of that is currently just "what came out of PAR", not that PAR is the result of a page table walk. If user-level proofs ever get to the point where they do need to know that, then we do indeed need to model it.

Indanz: > I was a bit imprecise. `[*]` does actually mean "potentially everything". We do constrain that later with an additional assumption to what is called `machines_state_rest` in the formalisation. `machine_state_rest` is the part of the machine state that we don't model further (apart from "it exists" and "it might change when you run one of these instructions" and "ASpec will change it in exactly the same way and order as C does").

Sounds like it would make sense to have a `MODIFIES: machine_state_rest` then.

> > Also, now that we save and restore PAR, shouldn't it actually be part of the modelled machine state?
> 
> Should is a strong word. None of these registers are currently modelled. Only their existence, not what they do. That is a safer assumption to make (when it works) than modelling what they do and getting that wrong.

This is what I had in mind above, just its existence.

> For this specific case, it's a bit weak, because we do actually use the result of this register on AArch64 for the fault return, and the formalisation of that is currently just "what came out of PAR", not that PAR is the result of a page table walk. If user-level proofs ever get to the point where they do need to know that, then we do indeed need to model it.

PAR really is just a simple read/write register without special architectural meaning which happens to get the result of an AT instruction. In theory the AT instructions could have had a normal register as output, but this was probably easier to implement. So there isn't anything special to model about PAR, only the AT instructions.

lsf37: > > I was a bit imprecise. `[*]` does actually mean "potentially everything". We do constrain that later with an additional assumption to what is called `machines_state_rest` in the formalisation. `machine_state_rest` is the part of the machine state that we don't model further (apart from "it exists" and "it might change when you run one of these instructions" and "ASpec will change it in exactly the same way and order as C does").
> 
> Sounds like it would make sense to have a `MODIFIES: machine_state_rest` then.

It sounds like that, but we still prefer it this way. The extremely weak "might modify anything" forces us to be explicit about when we use those `machine_state_rest` assumptions. The modifies clauses become theorems that are used automatically in the C-level verification, and these are instructions where we want to force things to be manual, because that helps with tracking down anything that needs validation.


> PAR really is just a simple read/write register without special architectural meaning which happens to get the result of an AT instruction. In theory the AT instructions could have had a normal register as output, but this was probably easier to implement. So there isn't anything special to model about PAR, only the AT instructions.

Yes, fair, it's really the effect of the AT instruction that we would be modelling there, not the register as such.

PR #1324, opened 2024-09-25
aarch32: make sure `irqInvalid` is `irq_t`

lsf37:
The GIC platforms use `irq_t` for `irqInvalid` -- bring remaining AArch32 platforms in line with that.

`irq_t` is an unsigned type (when it is an integer) and enum constants are signed. For the proofs to treat these platforms uniformly, they need to be the same kind. They can't all be enums, because `irq_t` can be a more complex type for SMP platforms.

### DIFF
--- a/include/drivers/irq/am335x.h
+++ b/include/drivers/irq/am335x.h
@@ -14,7 +14,7 @@
 /* No SGIs on this platform. */
 #define NUM_SGIS 0
 
-irq_t irqInvalid = 255;
+static const irq_t irqInvalid = 255;
 
 #define CMPER_REG(base, off) ((volatile uint32_t *)((base) + (off)))
 #define CMPER_TIMER3_CLKCTRL    0x84
--- a/include/drivers/irq/am335x.h
+++ b/include/drivers/irq/am335x.h
@@ -14,9 +14,7 @@
 /* No SGIs on this platform. */
 #define NUM_SGIS 0
 
-enum irqNumbers {
-    irqInvalid = 255
-};
+irq_t irqInvalid = 255;
 
 #define CMPER_REG(base, off) ((volatile uint32_t *)((base) + (off)))
 #define CMPER_TIMER3_CLKCTRL    0x84
--- a/include/arch/arm/arch/machine/gic_common.h
+++ b/include/arch/arm/arch/machine/gic_common.h
@@ -52,11 +52,11 @@
                         CORE_IRQ_TO_IRQT(0, (idx) - (CONFIG_MAX_NUM_NODES-1)*NUM_PPI))
 #define IRQT_TO_CORE(irqt) (irqt.target_core)
 #define IRQT_TO_IRQ(irqt) (irqt.irq)
-irq_t irqInvalid = CORE_IRQ_TO_IRQT(-1, -1);
+static const irq_t irqInvalid = CORE_IRQ_TO_IRQT(-1, -1);
 
 #else
 #define IRQ_IS_PPI(irq) HW_IRQ_IS_PPI(irq)
-irq_t irqInvalid = (uint16_t) -1;
+static const irq_t irqInvalid = (uint16_t) -1;
 #endif
 
 /* Setters/getters helpers for hardware irqs */
--- a/include/drivers/irq/bcm2836-armctrl-ic.h
+++ b/include/drivers/irq/bcm2836-armctrl-ic.h
@@ -122,7 +122,7 @@ volatile struct core_regs {
 #define LOCAL_TIMER_CTRL_EN_BIT 28
 #define LOCAL_TIMER_CTRL_RL_MASK MASK(28)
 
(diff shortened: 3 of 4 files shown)


Indanz:
Why not just:
```
#define irqInvalid 255U
```

And for GIC SMP:
```
#define irqInvalid CORE_IRQ_TO_IRQT(-1, -1)
```
And get rid of those stupid globals. It compiles for me.

(I understand why irq_t was made a structure in commit 2d362cb7c8, but passing structures around and doing structure compares is quite inefficient compared to just stuffing it all in one word.)

include/drivers/irq/am335x.h
```diff
@@ -11,9 +11,7 @@
 #include <machine/interrupt.h>
 #include <armv/machine.h>
 
-enum irqNumbers {
-    irqInvalid = 255
-};
+irq_t irqInvalid = 255;
```
axel-h, line 14: Make it const also everywhere

```suggestion
const irq_t irqInvalid = 255;
```

Xaphiosis:
Hang on, if we make it a global variable instead of an enum, we don't get a constant out the C parser. So if we care about `Kernel_C.irqInvalid` being equal to abstract `irqInvalid`, that would have to come via invariant or a static fetch function like we did with the kernel base offset.
If we go with Indan's suggestion, then we end up with a magic number in the parsed code (issue for proof portability across platforms), but no need for an invariant.
Where does this fit in the plan?

Xaphiosis (approved):
There's a broader discussion about what's going on with the `irq_t` type, but this does make it consistent with other platforms.

kent-mcleod:
Looking now, it even seems weird to me that irqInvalid became a global. But before that commit it was an enum and the irq_t type was used.  Was that a problem then?

lsf37:
You're right, the global does look strange, because we are getting a constant in Isabelle there, which we usually wouldn't. I'll investigate more next week.

The problem is only that it is inconsistent between platforms (unsigned on GIC platforms, signed because of enum on others). If was an enum everywhere, that would work too, but that doesn't work with the struct in SMP configs.

lsf37:
> Why not just:
> 
> ```
> #define irqInvalid 255U
> ```

Because we need the value to have a symbolic name in the verification if we want to keep the proof generic. If it's a `#define`, the proof only sees the number. The usual way of doing this nicely is an enum. I'm not actually sure why a global works here at all (for verification) and will need to investigate that a bit more. Will report back.

Indanz:
> Because we need the value to have a symbolic name in the verification if we want to keep the proof generic. If it's a `#define`, the proof only sees the number. The usual way of doing this nicely is an enum.

But enums are signed, while IRQs are unsigned values. Wouldn't that give problems in the proof?

lsf37:
> > Because we need the value to have a symbolic name in the verification if we want to keep the proof generic. If it's a `#define`, the proof only sees the number. The usual way of doing this nicely is an enum.
> 
> But enums are signed, while IRQs are unsigned values. Wouldn't that give problems in the proof?

Yes, completely right. That's why I was adjusting the enums to globals in this PR, because the global worked in `gic_common. h`. But now I want to figure out why that is before we do anything with this, because it's strange that it does. 

The enums are not inherently problematic, they just introduce casts that I have to prove do nothing. That's entirely possible, but it's bad when some configs need that and others don't (and of course it's even nicer if there are no casts).

lsf37:
From @Xaphiosis:
> Hang on, if we make it a global variable instead of an enum, we don't get a constant out of the C parser. 

Turns out that is not true. If it is a global that is never written to, the C parser does turn it into a constant. This is what's happening with the existing `irqInvalid` values on the other platforms. 

From myself:
> But now I want to figure out why that is before we do anything with this, because it's strange that it does.

Nothing strange after all, we just had both forgotten that the C parser has this feature. Must have been one of the early attempts to deal with constants in C, whereas we now mostly use enums. It's not entirely ideal, because I don't think the compiler will deduce that it is a constant, but in this case that doesn't matter much (actually, with whole program optimisation it might, I haven't checked the assembly), and an enum would be more annoying in this case.

midnightveil:
Does this mean we should re-open https://github.com/seL4/seL4/pull/1349 which uses "static const" (so that the compilers are better at inlining/realising it's a constant since it's not externally visible?)

(Also: one day we will have C23 and it's lovely contexpr :P )

lsf37:
> Does this mean we should re-open #1349 which uses "static const" (so that the compilers are better at inlining/realising it's a constant since it's not externally visible?)

Good point. I'll try out what the C parser does with `static const`, if it handles it, we should use it.

lsf37:
> > Does this mean we should re-open #1349 which uses "static const" (so that the compilers are better at inlining/realising it's a constant since it's not externally visible?)
> 
> Good point. I'll try out what the C parser does with `static const`, if it handles it, we should use it.

It does. I'll add a separate commit to add `static const` in the relevant files.

PR #1325, opened 2024-09-26
FPU Context Switching

Indanz:
Implementation of [RFC](https://sel4.github.io/rfcs/active/0180-fpu-switching.html).

Simplest version of semi-lazy FPU switching.

Still work in progress, but in good enough shape for wider review. For previous development, see [here](https://github.com/Indanz/seL4/pull/1/).

TODO:
- Double check Armv8 virtualisation assumptions.
- Fix x86 XSAVES.
- Test virtualisation.
- Test and benchmark.

Possible Optimisations:
- Try alternative domain switch variants.
- Try non-NULL ksActiveFPUState.
- Try global ksFPUEnabled.


Indanz:
For some reason `if (config_set(CONFIG_HAVE_FPU))` doesn't work to hide things. Easily fixed by creating an empty `fpuRelease`, but I wouldn't expect that  would be necessary. Any suggestions?

Indanz:
@chrisguikema If you have time, could you review and test the x86 VCPU changes on your VmMs? Would be good to know whether there is a performance degradation too. I don't expect any, but if there is, it's probably a bug.

I'm not sure if the guest's requested CR0:TS is being preserved or not (because I clear TS when unconditionally enabling the FPU for the guest). Is the CR0 used by guests the same as the CR0 used by hosts, or does it have its own architectural copy? I guess it has it's own copy because of VMX_GUEST_CR0, but not sure what the interaction is with CR0.

corlewis:
Just a couple of last changes that we'd like to make to this for verification. Would it be ok to move `prepareSetDomain` out of `setDomain` and into `decodeDomainInvocation`? That matches up better with how the specifications have been written and allows us to separate out the proofs about FPU and domain state.

Separately, we need to label `enableFpu` with `MODIFIES: phantom_machine_state`, as it us where we currently have the machine interface for the proofs and we need to know that it only modifies state that is not in the model.

See https://github.com/corlewis/seL4/commit/ef617bc43549206be5e31702098a39a19c8854e7 for the versions of these changes that I've been working with.

Also, when you have a chance another rebase would be helpful, to get the most recent verification-visible changes.

corlewis:
> Just a couple of last changes that we'd like to make to this for verification. Would it be ok to move `prepareSetDomain` out of `setDomain` and into `decodeDomainInvocation`? That matches up better with how the specifications have been written and allows us to separate out the proofs about FPU and domain state.

Oh, I've just realised that if we do make this change then for consistency we should probably wrap them up in a new `invokeDomain` function and call that from `decodeDomainInvocation`, like we do with all of the other decode functions.

Indanz:
Okay, done. Let me know if this is fine.

src/object/tcb.c
```diff
@@ -1604,9 +1642,15 @@ exception_t decodeDomainInvocation(word_t invLabel, word_t length, word_t *buffe
         current_syscall_error.invalidArgumentNumber = 1;
         return EXCEPTION_SYSCALL_ERROR;
     }
+    tcb = TCB_PTR(cap_thread_cap_get_capTCBPtr(tcap));
+    return invokeDomainSetSet(tcb, domain);
+}
 
+exception_t invokeDomainSetSet(tcb_t *tcb, dom_t domain)
+{
     setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
```
corlewis, line 1651: Ah, that took a while to get back up to and to remember. To follow the structure of other decode/invokes and to allow us to reuse existing proofs, this `setThreadState` should still be in `decodeDomainInvocation, before the call of `invokeDomainSetSet`.

src/object/tcb.c
```diff
@@ -1604,9 +1642,15 @@ exception_t decodeDomainInvocation(word_t invLabel, word_t length, word_t *buffe
         current_syscall_error.invalidArgumentNumber = 1;
         return EXCEPTION_SYSCALL_ERROR;
     }
+    tcb = TCB_PTR(cap_thread_cap_get_capTCBPtr(tcap));
+    return invokeDomainSetSet(tcb, domain);
```
corlewis, line 1646: Oh and for completeness, now that the two uses of `tcb` have been pulled into a separate function it's up to you whether you prefer it as a separate variable like this or inline in the return like it was before. Either option now shouldn't impact the proofs.

corlewis:
> Okay, done. Let me know if this is fine.

Sorry, I thought that the new `invokeDomainSetSet` function would be fine but it returning an exception causes a surprising amount of pain in the proofs. Do you have any objections to making it a void function, like in https://github.com/corlewis/seL4/commit/c852075ae67fdc4a77fede00a78034479a85eb9d? With that I've been able to update all of the proofs. There is also a precedent for this pattern in the various `invokeIRQHandler` functions that can be seen at https://github.com/seL4/seL4/blob/master/src/object/interrupt.c#L86.

Indanz:
> Sorry, I thought that the new `invokeDomainSetSet` function would be fine but it returning an exception causes a surprising amount of pain in the proofs. Do you have any objections to making it a void function, like in [corlewis@c852075](https://github.com/corlewis/seL4/commit/c852075ae67fdc4a77fede00a78034479a85eb9d)?

That's surprising, as that's how all other invoke functions are in that file. Actually, looking at all invoke functions, only the IRQ ones are void, all others return exception_t. I have no objection to changing it, but it might be worth looking into why this one is so hard compared to the others? Perhaps you can simplify the proofs a lot if you change the other invoke functions to void too.

corlewis:
> That's surprising, as that's how all other invoke functions are in that file. Actually, looking at all invoke functions, only the IRQ ones are void, all others return exception_t. I have no objection to changing it, but it might be worth looking into why this one is so hard compared to the others? Perhaps you can simplify the proofs a lot if you change the other invoke functions to void too.

Honestly, in this case a fair bit of the pain is probably because I already had working proofs and I'm now trying to minimise the changes I'm making for `invokeDomainSetSet`. If we'd had this from the start I would have just copied the specifications and proofs of the other functions and it would presumably have been fine.

One thing when considering the other invoke functions is that some of them actually can be preempted and return an exception, so those ones can't be changed to void. Now that I've done this though I think that in the cases where they can't be preempted it actually is nicer to be void, but I don't think that it's enough of an improvement to be worth going through and updating them.

lsf37:
> One thing when considering the other invoke functions is that some of them actually can be preempted and return an exception, so those ones can't be changed to void. Now that I've done this though I think that in the cases where they can't be preempted it actually is nicer to be void, but I don't think that it's enough of an improvement to be worth going through and updating them.

When there is a quiet period at some point, we should consider it, because it is one of the rare cases where the C type systems actually does help a little bit on the C side (when it returns `void` we know there is no preemption). Axel had a pull request up a while ago to change some of the existing invocations, but I never got around to doing the update. For the same reasons we don't want to do them now :-), but not because it's a bad idea as such.

src/arch/x86/machine/fpu.c
```diff
@@ -10,6 +10,9 @@
 #include <arch/object/structures.h>
 #include <arch/machine/fpu.h>
 
+/* A valid initial FPU state, copied to every new thread. */
+#define x86KSnullFpuState NODE_STATE(ksIdleThread)->tcbArch.tcbContext.fpuState
```
corlewis, line 14: Do you remember why you changed `x86KSnullFpuState` to be a `#define` for the idle thread's fpu state instead of being a separate global? We could probably handle it like this, but the more complicated structure does make the proofs harder.

Indanz: Because you wanted `ksCurFPUOwner` instead of `ksActiveFPUState`, see commit a2d5552b55e045e2469a36906960a23fb97d9d64.

lsf37: The reason we didn't zero was that the initial FPU state doesn't necessarily have to be all zeroes. It's unclear to me If that was too careful at the time or if there are flags and other content that need to be preserved from boot. I do remember someone saying that the architecture manual does not guarantee that all zeroes is what we want.

Indanz: Yes, I remember now, all zeroes might not match the result of `finit` + `MXCSR_INIT_VALUE`. Linux does the FPU init at task start, but seL4 can't easily do that, hence the construction we have. I'll go and check if we can restore zero fpu state, as that would be by far the most elegant solution.

`ksIdleThread` is already a global and a `tcb_t*`, wasting memory on another tcb just for fpu init seems silly, especially while the idle task won't use the one it has already. Would it help to have a local variable in `Arch_initFpuContext` which (partially) resolves `NODE_STATE(ksIdleThread)->tcbArch.tcbContext.fpuState`, or to use `ksIdleThreadTCB` directly?

I can always restore the original `x86KSnullFpuState` and init it from the idle thread's fpu state, but that would make an already ugly construction worse.

lsf37: Definitely fine to use some memory for this. I'd rather not copy from the idle thread directly if that make any sense, but keep a global for this state instead (that could be initialised from the idle thread). That'll make it easier to show that it's not touched by anything else (the idle thread isn't either, but it's slightly more work to show that).

A global that just has the null(ish) FPU state should be fine (as opposed for the current FPU pointer that we wanted to be a `tcb_t`), because we hopefully don't need to write invariants comparing that to anything else. @corlewis please do correct me if I'm wrong there.

corlewis: Having played around with these proofs a fair bit now, yes, it is a bit of a pain to show that it is safe to use the idle thread. It probably isn't too bad in the end but these are old, ugly proofs and I haven't worked it out yet.

The alternative of restoring `x86KSnullFpuState` is better than I thought, the way the code has ended up means that it doesn't need to be a `tcb_t` like I was worried and can remain a `user_fpu_state_t`. I think it even results in nicer proofs, as we no longer need to take the address of the global which simplifies how the C Parser treats it.

corlewis: Ok, it took a bit more iteration yet again, but with https://github.com/corlewis/seL4/commit/4fd9fcf16668ebe3330889e2844f9137c6942fd6 we've been able to finally finish updating the proofs for all verified configurations. If you're ok with something like that then that would resolve this from my side.

corlewis: I should point out that I have only updated the proofs for this, which doesn't say anything about the change to the boot code, so it would still need to be tested.

Indanz: I've merged your commit, it looks okay, let's see what the tests say. I'll rebase later to clean things up and combine later commits with earlier ones so it all makes more sense, but I'll do that at the very end probably.

Do you want me to bring it up-to-date with master?

corlewis: That sounds like a good plan.

Bringing it up to date never hurts but isn't necessary for the proofs at the moment, the recent commits don't make any verification visible changes.

Indanz:
XSAVES works with VCPU too now, we were missing proper FPU state init for VCPUs on x86.

All code paths in `vcpu_fpu_to_guest()` are now tested (verified with printfs).

For `vcpu_fpu_to_host()`, only the complicated path is tested, as that is the default case when the host has FPU enabled. All in all I'm confident it's working for x86 with virtualisation now. I've tested it with x86_64, but can test 32-bit too to be sure.

And because of the debugging, I know it's trying to load the correct fpu state, instead of mixing host with guest FPU state or something silly like that.

Luckily the result of `../init-build.sh -DCAMKES_VM_APP=minimal_64` worked in qemu for me with:
`$ qemu-system-x86_64 -m 512 -cpu host -serial /dev/stdout --enable-kvm -kernel images/kernel-x86_64-pc99 -initrd images/capdl-loader-image-x86_64-pc99 `

lsf37:
> For `vcpu_fpu_to_host()`, only the complicated path is tested, as that is the default case when the host has FPU enabled. All in all I'm confident it's working for x86 with virtualisation now. I've tested it with x86_64, but can test 32-bit too to be sure.

You mean x86 32-bit virtualisation? I guess we should better double-check, but I wasn't even aware that we support that.

Indanz:
> You mean x86 32-bit virtualisation? I guess we should better double-check, but I wasn't even aware that we support that.

Yes. And we do. In CI it's e.g. PC99_debug_hyp_gcc_32.

include/arch/arm/arch/64/mode/machine/registerset.h
```diff
@@ -274,6 +274,7 @@ struct user_context {
     user_breakpoint_state_t breakpointState;
 #endif /* ARM_BASE_CP14_SAVE_AND_RESTORE */
 #ifdef CONFIG_HAVE_FPU
+    word_t _padding;
```
kent-mcleod, line 274: The comment on line 263 makes no sense to me. Is it trying to say how big this structure is? If so it appears completely wrong by a few multiples. (I looked because this padding is changing the struct size)

src/object/tcb.c
```diff
@@ -792,6 +792,40 @@ static exception_t decodeSetTLSBase(cap_t cap, word_t length, word_t *buffer)
     return invokeSetTLSBase(TCB_PTR(cap_thread_cap_get_capTCBPtr(cap)), tls_base);
 }
 
+static exception_t invokeSetFlags(tcb_t *thread, word_t clear, word_t set)
+{
+    word_t flags = thread->flags;
+
+    flags &= ~clear;
+    flags |= set;
+    thread->flags = flags;
```
kent-mcleod, line 801: Thoughts about returning the new flag state? I think this would allow `TCBSetFlags(cap, 0, 0)` to return the current flag value.

libsel4/include/interfaces/object-api.xml
```diff
@@ -825,6 +825,24 @@
             </error>
          </method>
 
+        <method id="TCBSetFlags" name="SetFlags" manual_name="Set Feature Flags" manual_label="tcb_setflags">
```
kent-mcleod, line 828: Should this generic invocation support setting and clearing multiple flags at a time? I think it should only allow updating one flag for now. I predict that some flags will require additional capabilities to be set such as flags that control access to other system registers such as genericTimer ctl registers or perf registers. I think it's possible to support a generic invocation that takes flag-specific capability arguments, but not when multiple flag changes are supported in a single call.

include/arch/arm/arch/32/mode/machine/fpu.h
```diff
@@ -88,8 +88,9 @@ static void setEnFPEXC(void)
     VMSR(FPEXC, fpexc);
 }
```
kent-mcleod, line 77: The commit message describes: "Use the task flag for the host and always enable FPU for the guest". Seems like there's a policy for whether the guest has access to FPU and on x86, TCB's have a second flag defined for guest FPU access?

libsel4/include/interfaces/object-api.xml
kent-mcleod: An additional thing my comment didn't consider is that if you grant authority for a TCB to access a protected system register (like timer or perf registers), then maybe the TCB needs to hold a copy of that capability in a CSlot in the TCB so that the authority can be revoked.

lsf37: I would make that kind of change when we have a concrete proposal for a sys-call that needs it. If it needs a non-TCB cap, that could still be encoded as optional argument(s) that are checked when the corresponding flag is in the set, or it can be excluded from here and be set/unset in a separate syscall.  I don't see any real benefit in removing the generality at this point if we can support it. You can always use the set API to set single flags, but not the other way around.

src/object/tcb.c
lsf37: That might make sense, actually, yes. @corlewis any opinions on that from the verification side?

corlewis: Would have to try it out of course to make sure, but I can't see it causing any problems.

include/arch/arm/arch/64/mode/machine/registerset.h
Indanz: That's copy & pasted from aarch32. I'll remove it, just like I did for the tcb.

src/object/tcb.c
Indanz: Good idea.

Which made me notice that I forgot to do `setThreadState(thread, ThreadState_Running)`. Oops.

Surprisingly cumbersome to return one value in a seL4 syscall. Seems weird to do `lookupIPCBuffer` while we know it will be less than `n_msgRegisters` and that everyone needs to explicitly zero the badge register.

libsel4/include/interfaces/object-api.xml
Indanz: If capabilities are involved, you may not need any flags.

That said, I don't foresee any problems setting or clearing multiple flags, as we only support 4 extra caps and doing 4 cap type lookups in one calls isn't excessive.

And like Gerwin said, we can add such restrictions later if necessary. I'd prefer it to be a requirement of specific flags instead of a generic requirement.

include/arch/arm/arch/32/mode/machine/fpu.h
Indanz: How I implemented it, with x86 virtualisation guests will always have the FPU enabled, so no extra flag is needed.

Although that is hard-coding policy in the kernel, with the reality of SIMD being used everywhere on x86 by both user space and more and more the Linux kernel itself too, mostly for non-FPU SIMD operations, I didn't see the added value of introducing a specific flag for x86 guest FPU access.

I could add a special flags just for x86 guests, but I don't think anyone will disable SIMD for guests in practice, so it would just make things more complicated for no gain.

I think it would be much better to give x86 virtual guests their own TCB like Arm has, then it will have its own flag automatically.

lsf37: >  but I don't think anyone will disable SIMD for guests in practice, so it would just make things more complicated for no gain.

I agree. It is policy, but it doesn't really make sense to not have FPU on for x86 VCPUs. We can also still add another flag later if we want to. The current implementation has most of what is needed for that, so it'd be a much smaller step than this one with not much verification impact. At least I would vote for merging this while we discuss more.

The verification PRs for FPU are large and are blocking other merges at the moment, so it'd be good to get this one done. If we're all in agreement that the mechanism as such are fine we can still fine tune in additional PRs.

src/object/tcb.c
lsf37: I have some ideas for simplifying that protocol, esp the IPC buffer part (Anna had a PR for it ages ago, but it cut some corners). It will probably require a small RFC, because the main reason we're doing a second lookup for the IPC buffer is the difference between the kernel reading and writing from/to the IPC buffer. I don't really think it makes sense to have a read-only IPC buffer, and with that gone, we can simplify more things in the implementation.

But that's for a different time and PR :-)

Indanz: Yes, took me a minute to figure out the second IPC lookup was needed to check for write access. Agreed that supporting read-only buffers makes little sense. If we also stop supporting non-call syscalls too, then things become much more streamlined.

Personally, I think there is no need for IPC buffers altogether, we can just pass everything via registers (all registers except one if necessary, that gives us a total limit of 7 on 32-bit x86, 14 on aarch32, 15 on x86-64 and 30 on aarch64 and riscv). The only reason user space needs an IPC buffer now is because the register-only limit of 4 is too low. Getting rid of IPC buffer altogether is perhaps too major, but we can at least use it for IPC calls only and not for syscalls (I think only seL4_TCB_Read/WriteRegisters and seL4_CNode_Rotate need to be changed for this).

Indanz:
The rebase madness should be done now. Multiple commits combined with previous ones, mistakes made with previous rebase fixed, and last tweaks after a per-commit review.

src/object/tcb.c
```diff
@@ -792,6 +792,48 @@ static exception_t decodeSetTLSBase(cap_t cap, word_t length, word_t *buffer)
...
+    }
+#endif
+    if (call) {
+        word_t *ipcBuffer = lookupIPCBuffer(true, thread);
+        setRegister(thread, badgeRegister, 0);
+        unsigned int length = setMR(thread, ipcBuffer, 0, flags);
+        setRegister(thread, msgInfoRegister, wordFromMessageInfo(
+                        seL4_MessageInfo_new(0, 0, 0, length)));
+    }
+    setThreadState(thread, ThreadState_Running);
```
Indanz, line 816: Is setting to `ThreadState_Running` needed or not? `handleInvocation` seems to change restart to running in the normal case, but it's behind an `unlikely`?

corlewis: I haven't dug through the code to work out if it's actually needed but from the verification perspective, this pattern with the `setThreadState ThreadState_Running` is one that we've already verified elsewhere. This means that unless there is a specific need to avoid the `setThreadState ` that we would have a strong preference to keep it as is so that we can reuse those proofs.

src/object/tcb.c
```diff
@@ -792,6 +792,48 @@ static exception_t decodeSetTLSBase(cap_t cap, word_t length, word_t *buffer)
...
```
corlewis, line 816: Should this be returning to `ksCurThread` instead of the thread that is having its flags updated?

src/object/tcb.c
Indanz: But I only added it recently and you didn't complain about it missing before. Most calls don't seem to do the `setThreadState ThreadState_Running`, but some do and I don't see a pattern.

corlewis: Our proofs require it as part of the reply from kernel pattern, so I didn't previously complain about it because it wasn't needed until now. I haven't done an exhaustive search, but you should see a `setThreadState(NODE_STATE(ksCurThread), ThreadState_Running)` in all of the verified invocations that are returning something to the user.

corlewis:
Okay, I've finally gotten back to this and updated the proofs for the last round of improvements. This did need one additional change however, to do with making it so that `invokeSetFlags` returns the new flag state to the current thread instead of to the thread that is having its flags modified.

See https://github.com/corlewis/seL4/commit/46cd3739547d0f1485989f827eb6024354da41b6 for the version of this that I've verified. If you agree with it, feel free to squash it into the appropriate commit.

Indanz:
> Okay, I've finally gotten back to this and updated the proofs for the last round of improvements. This did need one additional change however, to do with making it so that `invokeSetFlags` returns the new flag state to the current thread instead of to the thread that is having its flags modified.
> 
> See [corlewis@46cd373](https://github.com/corlewis/seL4/commit/46cd3739547d0f1485989f827eb6024354da41b6) for the version of this that I've verified. If you agree with it, feel free to squash it into the appropriate commit.

Of course  I agree with that change, it fixes a silly mistake. No time to squash it now, I'll do it later.

CHANGES.md
```diff
@@ -28,6 +28,8 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
 * riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
   Except for RISCV32 with LLVM clang enabled will default both to OFF.
+* Change fault based FPU context switching to a TCB flag based approach:
+  New system call `seL4_TCB_SetFlags` and new flag `seL4_TCBFlag_fpuDisabled`.
```
lsf37, line 32: The changelog entry should reference the RFC link (https://github.com/seL4/rfcs/pull/26)

lsf37:
> For some reason `if (config_set(CONFIG_HAVE_FPU))` doesn't work to hide things. Easily fixed by creating an empty `fpuRelease`, but I wouldn't expect that would be necessary. Any suggestions?
> 
> Edit: I fixed it by using ifdef.

For now that is the right fix (for verification as well). At some point in the future I'll want to make both options visible at the same time for verification and then I'll need the `if (config_set(..` version with the empty FPU interface functions. But that'll be a while and doing it now will probably break Corey's current proofs, so let's leave it as is. Just wanted to give a heads up.

Indanz:
It seems that commit 407484d causes things to break on the HiFive P550. Which seems to be the only RISCV with SMP enabled. Anyway, the reason why the commit breaks things is because FPU is disabled by clearing both initial and clear bits (which when both set means dirty), so we lose track whether the FPU state was actually dirty or not. All in all doing the dirty optimisation is more complicated than expected, so dropping that commit now. If anyone cares about RISCV FPU context switch performance, we can always add a fixed version back in.

Indanz:
Also tested the FPU changes on aarch32 with virtual machine and aarch64 with virtual machine. They both boot Linux and seem functional.

All seems fine, but I should add some sel4tests that execute the new syscall and also check that disabling FPU works as expected.

Indanz:
Glad I wrote the sel4test, the wrapper function didn't save MR0, so it got lost.

lsf37:
Tests are looking good now, I'll rebase for the cmake restyling.

lsf37:
Alright, things are finally passing. The link check is failing on the RFC link that will only exist once merged, and the zcu106 is failure is unrelated.

Will go ahead with the merge!

PR #1326, opened 2024-09-28
configs: additional verified platforms

lsf37:
With recent proof improvements the proofs now apply to further platforms in the ARM, ARM_HYP and AARCH64 configurations.

The rest of the config values are kept on whatever the default verified platform used, but most of these are now generic as well and could be removed or changed.

### DIFF
--- a/configs/AARCH64_hikey_verified.cmake
+++ b/configs/AARCH64_hikey_verified.cmake
@@ -7,4 +7,4 @@
 
 include(${CMAKE_CURRENT_LIST_DIR}/include/AARCH64_verified_include.cmake)
 
-set(KernelPlatform "bcm2711" CACHE STRING "")
+set(KernelPlatform "hikey" CACHE STRING "")
--- a/configs/AARCH64_odroidc2_verified.cmake
+++ b/configs/AARCH64_odroidc2_verified.cmake
@@ -7,4 +7,4 @@
 
 include(${CMAKE_CURRENT_LIST_DIR}/include/AARCH64_verified_include.cmake)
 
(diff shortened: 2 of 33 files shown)


kent-mcleod:
One of the original requirements we had when switching from Make to CMake for kernel builds was to have a way of expressing the large number of valid configurations without having a different file for each config. Having a separate *_defconfig file for each configuration caused a lot of time spent in trying to keep the configs all in sync with each other  or debugging test bugs when configs got out of sync.

So something that CMake would support would be still having a single config file for each main verification target (eg ARM_verified.cmake) that could take `KernelPlatform` as an argument and internally the config validates/parses the input arguments to accept a valid supported configuration.

lsf37:
Factoring things out would make sense. I still need a separate file per platform for the verification build system to pick the right combination of things, but if they all read in a standard config for each major architecture, that would work and reduce duplication.

lsf37:
Ok, I've now factored these out into a main include file per major architecture, and single platforms are then only:

```cmake
include(${CMAKE_CURRENT_LIST_DIR}/include/ARM_verified_include.cmake)
set(KernelPlatform "tk1" CACHE STRING "")
```

Still need to double check this works as expected on the verification side.

This also means that the `_MCS` variants now include the main variant and just set `MCS` in addition, so they will stay in sync. And `ARM_HYP` only sets hyp mode on top of the standard ARM config. This may tweak some of the root CNode size bits etc, but the verification is by now no longer sensitive to those, and these settings could probably be dropped completely. I just left them in as demo of what settings could be tweaked.

Xaphiosis (approved):
From my side all good, if systems people are happy with it.

PR #1328, opened 2024-10-03
Adding auto-generation of a config file for invocation numbers

alwin-joshy:
The primary purpose of this PR is to generate `generated/invocations_all.json`, a JSON file containing a mapping between invocation labels and the numbers they correspond to.

It works by modifying `invocation_gen_header.py` to eagerly evaluate whether a particular invocation will exist for the current configuration and emitting an intermediate JSON file for each of the three XML invocation definition files. We then add a new build step that stitches the intermediate files together in the correct order to produce a JSON file of the format

```
{
  "UntypedRetype": 1,
  "TCBReadRegisters": 2,
  "TCBWriteRegisters": 3,
...
}
```

The motivation of this PR is to simplify the implementation of the Microkit and avoid having to manually define this mapping, which gets annoying and error-prone when dealing with the multiple supported configurations, which can shift the mapping around.

A possible improvement could be to avoid the intermediate JSON files entirely and change `invocation_gen_header.py` to take in all three XML files, simplifying the process and avoiding the need for a new script.

edit: Actually, we don't need the intermediate files. I think a better solution might be to leave `invocation_gen_header.py` unchanged and pass all three XML directly to `invocation_json_gen.py`.

### DIFF
--- a/tools/invocation_json_gen.py
+++ b/tools/invocation_json_gen.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+#
+# Copyright 2024, UNSW Sydney
+#
+# SPDX-License-Identifier: BSD-2-Clause or GPL-2.0-only
+#
+
+import argparse
+import json
+import xml.dom.minidom
+from condition import condition_to_bool
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Generate JSON file containing list of seL4 \
+                                    invocations.')
+    parser.add_argument('--gen_config', type=argparse.FileType('r'),
+                        help="Location of gen_config JSON file", required=True)
+    parser.add_argument('--invocations', type=argparse.FileType('r'),
+                        help='Location of XML file with invocation definitions', required=True)
+    parser.add_argument('--arch_invocations', type=argparse.FileType('r'),
+                        help='Location of XML file with arch invocation defintions', required=True)
+    parser.add_argument('--sel4_arch_invocations', type=argparse.FileType('r'),
+                        help='Location of XML file with seL4 arch invocation definitions', required=True)
+    parser.add_argument('--dest', type=argparse.FileType('w+'),
+                        help='Location of JSON file to store invocations', required=True)
+
+    return parser.parse_args()
+
+
+def xml_to_json_invocations(xml, gen_config, counter, invocations_dict, ):
+    for method in xml.getElementsByTagName("method"):
+        label = str(method.getAttribute("id"))
+        exists = condition_to_bool(method.getElementsByTagName("condition"), gen_config)
+        if exists:
+            invocations_dict[label] = counter
+            counter += 1
+
+    return counter
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    try:
+        invocations = xml.dom.minidom.parse(args.invocations)
+        arch_invocations = xml.dom.minidom.parse(args.arch_invocations)
+        sel4_arch_invocations = xml.dom.minidom.parse(args.sel4_arch_invocations)
+    except:
+        print('Error: invalid XML file provided', file=sys.stderr)
+        sys.exit(1)
+
+    try:
+        gen_config = json.load(args.gen_config)
+    except:
+        print('Error: invalid JSON file provided', file=sys.stderr)
+        sys.exit(1)
+
+    invocations_dict = {}
+    counter = 1
+    counter = xml_to_json_invocations(invocations, gen_config, counter, invocations_dict)
+    counter = xml_to_json_invocations(sel4_arch_invocations, gen_config, counter, invocations_dict)
+    counter = xml_to_json_invocations(arch_invocations, gen_config, counter, invocations_dict)
+
+    json.dump(invocations_dict, args.dest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ RequireTool(CONFIG_GEN_PATH config_gen.py)
 RequireTool(BF_GEN_PATH bitfield_gen.py)
 RequireTool(HARDWARE_GEN_PATH hardware_gen.py)
 RequireTool(INVOCATION_ID_GEN_PATH invocation_header_gen.py)
+RequireTool(INVOCATION_JSON_GEN_PATH invocation_json_gen.py)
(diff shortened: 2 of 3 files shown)


alwin-joshy:
The implementation has been simplified.

All three XML files are now directly fed into the new script (`invocation_json_gen.py`), which parses them in the correct order and exports `generated/invocations_all.json`

tools/invocation_json_gen.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+#
+# Copyright 2024, Data61, CSIRO (ABN 41 687 119 230)
```
Indanz, line 3: I don't think that's right. Even if it was based on a file by Data61/CSIRO, then the year should be the original year, not this year. But you probably changed enough that you can ignore the original copyright, this file is barely copyrightable as it is.

Indanz (approved):
Code-wise seems correct and neat.

Only question I have is whether we always want to generate invocations_all.json? I think that's what's happening now, but I'm not sure we want to always do that.

tools/invocation_json_gen.py
alwin-joshy: I don't think I've added a new file to seL4 before, so I wasn't too sure what to do here.

Is 

# Copyright 2024, UNSW Sydney

appropriate?

Indanz: Depends, did you do this work as part of a contract for UNSW? Then yes, that's correct, though your contract may allow you to retain copyright and only require the work to be released as open source. If you're a student then you own the copyright and should use your name. See the [UNSW copyright site](https://subjectguides.library.unsw.edu.au/copyright/copyright_ownership) for more details.

alwin-joshy:
> Only question I have is whether we always want to generate invocations_all.json? I think that's what's happening now, but I'm not sure we want to always do that.

Yes, it's currently being generated unconditionally. I think this is benign, but can put it behind a config option if that's better.

Indanz:
> > Only question I have is whether we always want to generate invocations_all.json? I think that's what's happening now, but I'm not sure we want to always do that.
> 
> Yes, it's currently being generated unconditionally. I think this is benign, but can put it behind a config option if that's better.

With make would add a new target, can't you do something similar with cmake? Requiring people to add `list(APPEND gen_files_list "generated/invocations_all.json")` to their project file seems unfriendly.

I'm okay with always generating it if it doesn't introduce new dependencies and doesn't cost much time, definitely seems the simplest way and increases the chance that people become aware of this JSON file.

alwin-joshy:
> doesn't introduce new dependencies 

All of the modules used are builtins

> and doesn't cost much time

without additional build step:
```
ninja  1.40s user 0.57s system 256% cpu 0.766 total
ninja  1.41s user 0.61s system 271% cpu 0.744 total
ninja  1.40s user 0.60s system 294% cpu 0.679 total
ninja  1.45s user 0.63s system 255% cpu 0.815 total
ninja  1.49s user 0.58s system 278% cpu 0.743 total
```

with additional build step:
```
ninja  1.62s user 0.61s system 233% cpu 0.953 total
ninja  1.52s user 0.57s system 273% cpu 0.764 total
ninja  1.47s user 0.57s system 298% cpu 0.683 total
ninja  1.51s user 0.60s system 279% cpu 0.754 total
ninja  1.50s user 0.56s system 299% cpu 0.689 total
```

Indanz:
Fine with me, we can always revert if it causes problems in any way.

lsf37:
This is the output for `sabre` (the build is broken for all architectures, though):

```
-- platform imx6 supports multiple sub platforms, none was given
--   defaulting to: sabre
-- platform imx6 supports multiple architectures, none was given
--   defaulting to: aarch32
-- The C compiler identification is GNU 9.2.1
-- The ASM compiler identification is GNU
-- Found assembler: /usr/bin/arm-none-eabi-gcc
-- Check for working C compiler: /usr/bin/arm-none-eabi-gcc
-- Check for working C compiler: /usr/bin/arm-none-eabi-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Appending /home/test-runner/ver/l4v/spec/cspec/c/overlays/ARM/overlay.dts overlay
-- /home/test-runner/ver/l4v/spec/cspec/c/build/ARM/gen_headers/plat/machine/devices_gen.h is out of date. Regenerating from DTB...
-- Configuring done
-- Generating done
-- Build files have been written to: /home/test-runner/ver/l4v/spec/cspec/c/build/ARM
-- Used default overlay for ARM
cd build/ARM && ninja kernel_all_pp_wrapper
ninja: error: '/home/test-runner/ver/seL4/gen_config/kernel/gen_config.json', needed by 'generated/invocations_all.json', missing and no known rule to make it
```

alwin-joshy:
It looks like something to do with relative paths. I have pushed a patch that I think will fix it https://github.com/alwin-joshy/seL4/commit/3f1e85384cd932b0c40f1d43c2bed294b168c03d.

Not sure why this works locally but not on CI, maybe a difference in how CMake is being invoked. There are other relative paths in the same file, like `set(header_dest "gen_headers/api/invocation.h")`, so not sure why these work okay and these don't.

Indanz:
The proofs do:

```
 cmake -C /home/test-runner/ver/seL4/configs/X64_verified.cmake \
	-DCROSS_COMPILER_PREFIX= \
	-DCMAKE_TOOLCHAIN_FILE=/home/test-runner/ver/seL4/gcc.cmake \
	-DKernelDomainSchedule=/home/test-runner/ver/l4v/spec/cspec/c/config_sched.c \
	-DUMM_TYPES=/home/test-runner/ver/l4v/spec/cspec/c/build/X64/umm_types.txt -DCSPEC_DIR=/home/test-runner/ver/l4v/spec/cspec \
	  -DKernelNumDomains=1 \
	 \
	-G Ninja /home/test-runner/ver/seL4
```

While CI does:

```
cmake -G Ninja -B build -C configs/X64_verified.cmake -DCMAKE_TOOLCHAIN_FILE=gcc.cmake
```

I don't understand why other relative paths do work, while this new one doesn't. I suspect that the fix will work, but I don't understand why it's necessary.

PR #1329, opened 2024-10-04
mcs: use sc_active in commitTime

michaelmcinerney:


lsf37 (approved):
👍

Indanz (approved):
There's also src/api/syscall.c:485, src/fastpath/fastpath.c:666 and src/object/schedcontrol.c:39 and some asserts.

Xaphiosis:
> There's also src/api/syscall.c:485, src/fastpath/fastpath.c:666 and src/object/schedcontrol.c:39 and some asserts.

Would be good to catch these as well. One day we might hit MCS fastpath.

lsf37:
I'm going to merge this one as is, but I agree it'd be good to do the other instances as well, and more systematically. Will put that on the todo list.

PR #1330, opened 2024-10-04
invocations_all.json: make paths explicit

alwin-joshy:


Indanz:
It now fails with:

```
[7/22] Generating generated/invocations_all.json
FAILED: generated/invocations_all.json 
cd /home/test-runner/ver/l4v/spec/cspec/c/build/AARCH64 && rm -f /home/test-runner/ver/l4v/spec/cspec/c/build/AARCH64/generated/invocations_all.json && python3 /home/test-runner/ver/seL4/tools/invocation_json_gen.py --gen_config /home/test-runner/ver/l4v/spec/cspec/c/build/AARCH64/gen_config/kernel/gen_config.json --invocations /home/test-runner/ver/seL4/libsel4/include/interfaces/object-api.xml --arch_invocations /home/test-runner/ver/seL4/libsel4/arch_include/arm/interfaces/object-api-arch.xml --sel4_arch_invocations /home/test-runner/ver/seL4/libsel4/sel4_arch_include/aarch64/interfaces/object-api-sel4-arch.xml --dest /home/test-runner/ver/l4v/spec/cspec/c/build/AARCH64/generated/invocations_all.json
  File "/home/test-runner/ver/seL4/tools/invocation_json_gen.py", line 61, in <module>
    counter = xml_to_json_invocations(invocations, gen_config, counter, invocations_dict)
  File "/home/test-runner/ver/seL4/tools/invocation_json_gen.py", line 34, in xml_to_json_invocations
    exists = condition_to_bool(method.getElementsByTagName("condition"), gen_config)
  File "/home/test-runner/ver/seL4/tools/condition.py", line 85, in condition_to_bool
    return expr_to_bool(children[0], values)
  File "/home/test-runner/ver/seL4/tools/condition.py", line 58, in expr_to_bool
    return not expr_to_bool(expr.childNodes[0], values)
  File "/home/test-runner/ver/seL4/tools/condition.py", line 64, in expr_to_bool
    return values[cfg_var.removeprefix("CONFIG_")]
AttributeError: 'str' object has no attribute 'removeprefix'
```

alwin-joshy:
Is the preferred solution to switch to a newer Python version or remove the prefix with a more manual implementation?

Indanz:
> Is the preferred solution to switch to a newer Python version or remove the prefix with a more manual implementation?

To me it's unexpected CONFIG_ is being stripped away at all, is it actually necessary?

alwin-joshy:
gen_config.json looks like

```
{
    "ARM_HIKEY_OUTSTANDING_PREFETCHERS": "0",
    "ARM_HIKEY_PREFETCHER_STRIDE": "0",
    "ARM_HIKEY_PREFETCHER_NPFSTRM": "0",
    ...
}
```

while the conditions in the XML file are specified like

```
            <condition><config var="CONFIG_TK1_SMMU"/></condition>
```

Indanz:
All very well, but isn't that something the caller should fix, not some generic function like `expr_to_bool`? And wouldn't it make more sense to add the missing CONFIG_ bit instead of stripping it?

Indanz:
The absolute path is necessary for it to compile with normal `ninja` in the build directory too.

alwin-joshy:
> All very well, but isn't that something the caller should fix, not some generic function like expr_to_bool? And wouldn't it make more sense to add the missing CONFIG_ bit instead of stripping it?

Fair enough, I can fix that.

> The absolute path is necessary for it to compile with normal ninja in the build directory too.

That's weird because that's the way I built it locally, and it seemed to work fine.

alwin-joshy:
> And wouldn't it make more sense to add the missing CONFIG_ bit instead of stripping it?

Actually, on second thought, doesn't it make more sense to convert the thing being looked up into the format of the dictionary? In this case, the dictionary is from the JSON file, which is of the form ARM_HIKEY_OUTSTANDING_PREFETCHERS while the condition being looked up is from the XML file and is of the form CONFIG_ARM_HIKEY_OUTSTANDING_PREFETCHERS, so to me it makes more sense to strip the CONFIG_.  

Otherwise, the caller would have to preprocess the whole dictionary and add CONFIG_ to all the keys before calling expr_to_bool.

Ideally, they would be called the same thing so you wouldn't need to change either.

Indanz:
This is getting all a bit silly, anything that works is good enough really, just do whatever is easiest for you.

Indanz:
Very strange, I really got:

```
ninja: error: 'kernel/gen_config/kernel/gen_config.json', needed by 'kernel/generated/invocations_all.json', missing and no known rule to make it
```
But now I can't reproduce it.

alwin-joshy:
> Very strange, I really got:
> 
> ```
> ninja: error: 'kernel/gen_config/kernel/gen_config.json', needed by 'kernel/generated/invocations_all.json', missing and no known rule to make it
> ```
> 
> But now I can't reproduce it.

I really hope there is no non-determinism in there somewhere...

Indanz:
> I really hope there is no non-determinism in there somewhere...

Yes, that's exactly it: It's a parallel build, everything happens at the same time, so it can try to build invocations_all.json before the input files are generated. You forgot to tell CMake that the input files are a dependency.

alwin-joshy:
> You forgot to tell CMake that the input files are a dependency.

This is what

```    DEPENDS
        "${CMAKE_CURRENT_BINARY_DIR}/gen_config/kernel/gen_config.json"
        "${INVOCATION_JSON_GEN_PATH}"
        "${invocations_absolute}"
        "${arch_invocations_absolute}"
        "${sel4_arch_invocations_absolute}"
```

in the custom command I added was meant to do. Did I misunderstand something?

Indanz:
I don't know then. It's what would be the case with a Makefile, but I don't know CMake well enough to figure this one out.

lsf37 (approved):

lsf37:
> Is the preferred solution to switch to a newer Python version or remove the prefix with a more manual implementation?

Switching Python versions is a bigger operation, definitely the manual implementation.

lsf37:
The current state looks good to me and seems to be building the proofs fine. I'll start a hw-build test as well, just to be sure.

PR #1332, opened 2024-10-08
gic_v3: fix SRE in hyp mode

JorgeMVP:
when hypervisor mode is enabled the kernel shall use ICC_SRE_EL2 or ICC_HSRE. This commit fixes this and supports compatibility between aarch32 and aarch64.

### DIFF
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -125,11 +125,17 @@ static void gicv3_enable_sre(void)
 {
     uint32_t val = 0;
 
-    /* ICC_SRE_EL1 */
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* ICC_SRE_EL2/ICC_HSRE */
+    SYSTEM_READ_WORD(ICC_SRE_EL2, val);
+    val |= GICC_SRE_EL2_SRE | GICC_SRE_EL2_EN;
+    SYSTEM_WRITE_WORD(ICC_SRE_EL2, val);
+#else
+    /* ICC_SRE_EL1/ICC_SRE */
     SYSTEM_READ_WORD(ICC_SRE_EL1, val);
     val |= GICC_SRE_EL1_SRE;
-
     SYSTEM_WRITE_WORD(ICC_SRE_EL1, val);
+#endif
     isb();
 }
 
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -137,12 +137,21 @@ static void gicv3_redist_wait_for_rwp(void)
 
 static void gicv3_enable_sre(void)
 {
-    word_t val = 0;
+    word_t val;
 
-    /* ICC_SRE_EL1 */
-    SYSTEM_READ_WORD(ICC_SRE_EL1, val);
-    val |= GICC_SRE_EL1_SRE;
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* Set SRE to enable register interface for GICv3 in EL2. Disable IRQ/FIQ legacy bypass.
+     * Set ICC_SRE_EL2.Enable to 0, so EL1 accesses to ICC_SRE_EL1 will trap.
+     *
+     * SRE may be RAO/WI and DIB/DFB may be read-only aliases of ICC_SRE_EL3 on cores
+     * with EL3. Writing to them is harmless.
+     */
+    val = GICC_SRE_EL2_SRE | GICC_SRE_EL2_DIB | GICC_SRE_EL2_DFB;
+    SYSTEM_WRITE_WORD(ICC_SRE_EL2, val);
+#endif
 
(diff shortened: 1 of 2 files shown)


JorgeMVP:
The ICC_SRE_EL1 also needs to be saved and restored at vCPU switch - I will add this later here

src/arch/arm/machine/gic_v3.c
```diff
@@ -125,11 +125,17 @@ static void gicv3_enable_sre(void)
 {
     uint32_t val = 0;
 
-    /* ICC_SRE_EL1 */
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* ICC_SRE_EL2/ICC_HSRE */
+    SYSTEM_READ_WORD(ICC_SRE_EL2, val);
+    val |= GICC_SRE_EL2_SRE | GICC_SRE_EL2_EN;
```
Indanz, line 131: GICC_SRE_EL2_EN disables trapping to EL2 on change of ICC_SRE_EL1 by EL1, is that what you want? Currently the only field EL1 can change is SRE, the system register access enable bit.

Shouldn't we disable IRQ and FIQ bypass modes and set DIB and DFB to 1? Current hardware has EL3, so these fields are likely read-only versions of ICC_SRE_EL3, but in case that isn't the case, and the hardware supports bypass mode, things may break when enabling it I think. This is true for ICC_SRE_EL1 too.

Indanz:
> The ICC_SRE_EL1 also needs to be saved and restored at vCPU switch - I will add this later here

Is this really necessary? The only bit that EL1 can change is the SRE bit, but if that's already enabled EL1 has no reason to change it.

If EL1 traps on access of ICC_SRE_EL1 with ICC_SRE_EL2.Enable = 0, the hypervisor can decide what to return.

All in all I don't see the added value of letting guests change ICC_SRE_EL1 at all, hence no need to save and restore it.

That said, if we do ICC_SRE_EL2.Enable = 1 I guess we have no choice other than saving and restoring, otherwise one VM can disable SRE and cause another VM to break.

Indanz:
At https://docs.kernel.org/arch/arm64/booting.html it does say:

- If the kernel is entered at EL1:
       - ICC_SRE_EL2.Enable (bit 3) must be initialised to 0b1
       - ICC_SRE_EL2.SRE (bit 0) must be initialised to 0b1.

But KVM also [disallows](https://elixir.bootlin.com/linux/v7.0/source/arch/arm64/kvm/vgic-sys-reg-v3.c#L339) changing the register to anything other than what it wants, and HW implementations are allowed to behave as RAO/WI for all bits.

No Linux guest will count on being able to change those values, so setting ICC_SRE_EL2.Enable to 0 in seL4 is fine. In the worst case VMMs may need to emulate reads of ICC_SRE_EL1.

That said, we should explicitly configure `ICC_SRE_EL2` to what we want somewhere, and set the full value instead of just some bits. This needs to be done in addition to the current code that writes `ICC_SRE_EL1`.

lsf37:
For people like me who don't know all these registers:

- ICC-SRE-EL2: https://developer.arm.com/documentation/ddi0601/2026-06/AArch64-Registers/ICC-SRE-EL2--Interrupt-Controller-System-Register-Enable-Register--EL2-
- ICC-SRE-EL1: https://developer.arm.com/documentation/ddi0601/2026-06/AArch64-Registers/ICC-SRE-EL1--Interrupt-Controller-System-Register-Enable-Register--EL1-

If I understand the discussion correctly, when ICC_SRE_EL2.Enable = 0, access to ICC-SRE-EL1 will trap and deliver a VCPUFault and we don't need to context switch it. Normal guests won't be reading it anyway, so trapping is not a problem.

I agree that we want to disable IRQ bypass and FIQ bypass. When EL3 is present and controls this, our writes will be ignored, so that should also be fine. Should we write, and then read them and complain when IRQ bypass or FIQ bypass are allowed?

I'll rebase and push a separate commit with Indan's comments addressed.

lsf37 (approved):
I'd be happy with the current state after squashing commit, but would appreciate another review by @Indanz

Indanz:
> For people like me who don't know all these registers:

A good site to easily find registers is https://arm.jonpalmisc.com. It's generated from Arm's XML specification, so contains the same information as the official Arm documentation.

> If I understand the discussion correctly, when ICC_SRE_EL2.Enable = 0, access to ICC-SRE-EL1 will trap and deliver a VCPUFault and we don't need to context switch it. Normal guests won't be reading it anyway, so trapping is not a problem.

Yes. It's unlikely to try to access `ICC_SRE_EL1` because the only interesting bit is SRE, which enables the system register access.

> I agree that we want to disable IRQ bypass and FIQ bypass. When EL3 is present and controls this, our writes will be ignored, so that should also be fine. Should we write, and then read them and complain when IRQ bypass or FIQ bypass are allowed?

I wouldn't bother. I think it's an obscure feature that's not used in practice, at least not any more. If it's enabled the user probably did that deliberately and knows what they're doing. It's to bypass the GIC for some legacy interrupts and to connect them directly to a specific CPU core. Seems harmless to have enabled when there is nothing connected too though.

Indanz (approved):

PR #1343, opened 2024-11-03
manual: update table 9.1

yf13:
This adds `seL4_CapSMC` to table 9.1 and updates section 9.1 as well.

### DIFF
--- a/manual/parts/bootup.tex
+++ b/manual/parts/bootup.tex
@@ -27,7 +27,7 @@ slots), but the guard is always chosen so that the CNode resolves exactly
 the number of bits in the architecture (32 bits or 64 bits). This means, the
 first slot of the CNode has CPtr 0x0, the second slot has CPtr 0x1 etc.
 
-The first 15 slots (or 14 slots if not MCS) contain specific capabilities as listed in
+The first 16 slots contain specific capabilities as listed in
 \autoref{tab:cnode_content}.
 
 \begin{table}[htb]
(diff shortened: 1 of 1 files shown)


Indanz (commented):
The problem is that the SMP cap is ARM specific, so you can't mention it in generic documentation like that.

This is a weakness of the current manual setup, it's hard to document arch-specific information everywhere.

Indanz:
Worse, it's only available when `KernelAllowSMCCalls` is enabled, which is off by default.

manual/parts/bootup.tex
```diff
@@ -27,7 +27,7 @@ \section{Initial Thread's Environment}
 the number of bits in the architecture (32 bits or 64 bits). This means, the
 first slot of the CNode has CPtr 0x0, the second slot has CPtr 0x1 etc.
 
-The first 15 slots (or 14 slots if not MCS) contain specific capabilities as listed in
+The first 16 slots (or 15 slots if not MCS) contain specific capabilities as listed in
```
Indanz, line 30:
```suggestion
The first 16 slots contain specific capabilities as listed in
```

manual/parts/bootup.tex
```diff
@@ -61,6 +61,7 @@ \section{Initial Thread's Environment}
       \texttt{seL4\_CapSMMUSIDControl}      & global Arm SMMU SID controller, null cap if unsupported (see \autoref{sec:smmuv2}) \\
       \texttt{seL4\_CapSMMUCBControl}       & global Arm SMMU CB controller, null cap if unsupported (see \autoref{sec:smmuv2}) \\
       \texttt{seL4\_CapInitThreadSC}        & initial thread's scheduling context (MCS only) \\
+      \texttt{seL4\_CapSMC}                 & global SMC cap, null cap if not supported \\
```
Indanz, line 64:
```suggestion
      \texttt{seL4\_CapSMC}                 & global Arm SMC cap, null cap if not supported \\
```

Indanz (changes requested):
I checked the code and enum seL4_RootCNodeCapSlots is defined equally for all configurations, and `ndks_boot.slot_pos_cur = seL4_NumInitialCaps`, so it's always 16 and the existing comment is misleading.

Indanz (approved):

PR #1347, opened 2024-11-06
x86: initialize variables

nspin:
Multiple versions of GCC were unable to determine that these variables in `decodeX86PortInvocation` are initialized in all cases.

Oddly, this compilation failure was only observed with `-O1`. Compilation with `-O2` was not failing.

### DIFF
--- a/src/arch/x86/object/ioport.c
+++ b/src/arch/x86/object/ioport.c
@@ -222,7 +222,7 @@ exception_t decodeX86PortInvocation(
     word_t *buffer
 )
 {
-    exception_t ret;
+    exception_t ret = EXCEPTION_NONE;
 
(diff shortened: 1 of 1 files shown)


nspin:
I've opened https://github.com/seL4/l4v/pull/825 to update the `X64` C refinement proof.

Indanz:
Is this needed for binary verification?

nspin:
No, the `X64` configuration isn't in scope for binary verification.

kent-mcleod:
> Oddly, this compilation failure was only observed with `-O1`. Compilation with `-O2` was not failing.

I think there are a few cases where the optimization level causes the statements to be cut due to not being used before whatever layer that generates the warning sees them.

kent-mcleod (approved):

lsf37 (approved):

PR #1349, opened 2024-11-06
arm,gic: add static const qualifier to `irqInvalid`

nspin:
Commit [2d362cb](https://github.com/seL4/seL4/commit/2d362cb7c8e96a64b9848df06e226f50e79a1ec9) changed `irqInvalid` in `gic_common.h` from an enum constant to a global variable. The latter does not play nicely with the binary verification tools.  Specifically, the `C -> SydTV-GL` lowering inlines its value into `getActiveIRQ`, whereas GCC does not.

Adding the const qualifier encourages GCC to inline it.


Indanz:
I'd recommend to resolve the `irqInvalid` issues in #1324 and keep it all in one place.

There I propose to get rid of the global altogether and replace it wit a constant directly.

lsf37:
> I'd recommend to resolve the `irqInvalid` issues in #1324 and keep it all in one place.
> 
> There I propose to get rid of the global altogether and replace it with a constant directly.

I need to get back to this, but I don't think a constant will work. Will report back on the issue when I get to it, though.

lsf37:
@nspin this is now part of #1324

PR #1354, opened 2024-11-25
Platform support for Cheshire (Digilent Genesys2)

omeh-a:
This pull request adds support for the [Cheshire](https://github.com/pulp-platform/cheshire) platform - a system built around the CVA6 RISC-V APU design. This implementation specifically targets the Xilinx FPGA implementation of Cheshire for the Digilent Genesys 2 development board.

Cheshire is very similar to the extant Ariane CVA6 design supported by seL4, but is a system which is designed to have a stable interface and behaviour for 64 bit operating systems, as contrasted to a focus on 32 bit applications for Ariane.

Cheshire supports the RISC-V H-extension, but as of now it does not appear to work correctly under seL4 (apparently due to HW issues). As a result, this PR omits any H-extension related code.

### DIFF
--- a/src/plat/cheshire/config.cmake
+++ b/src/plat/cheshire/config.cmake
@@ -1,0 +1,29 @@
+#
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+# Copyright 2021, HENSOLDT Cyber
+# Copyright 2024, UNSW
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+cmake_minimum_required(VERSION 3.7.2)
+
+declare_platform(cheshire KernelPlatformCheshire PLAT_CHESHIRE KernelArchRiscV)
+
+if(KernelPlatformCheshire)
+    declare_seL4_arch(riscv64)
+    config_set(KernelRiscVPlatform RISCV_PLAT "cheshire")
+    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)
+    config_set(KernelOpenSBIPlatform OPENSBI_PLATFORM "fpga/cheshire")
+    set(OPENSBI_PLAT_XLEN "64")
+    set(OPENSBI_PLAT_ISA "rv64imafdc_zicsr_zifencei")
+    list(APPEND KernelDTSList "tools/dts/cheshire.dts")
+    list(APPEND KernelDTSList "src/plat/cheshire/overlay-cheshire.dts")
+    declare_default_headers(
+        TIMER_FREQUENCY 1000000
+        MAX_IRQ 51
+        INTERRUPT_CONTROLLER drivers/irq/riscv_plic0.h
+    )
+else()
+    unset(KernelPlatformFirstHartID CACHE)
+endif()
--- a/include/drivers/irq/riscv_plic0.h
+++ b/include/drivers/irq/riscv_plic0.h
@@ -17,6 +17,7 @@
     !defined(CONFIG_PLAT_QEMU_RISCV_VIRT) && \
     !defined(CONFIG_PLAT_ROCKETCHIP_ZCU102) && \
     !defined(CONFIG_PLAT_STAR64) && \
+    !defined(CONFIG_PLAT_CHESHIRE) && \
     !defined(CONFIG_PLAT_ARIANE)
 #error "Check if this platform supports a PLIC."
 #endif
--- a/libsel4/sel4_plat_include/cheshire/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/cheshire/sel4/plat/api/constants.h
@@ -1,0 +1,9 @@
+/*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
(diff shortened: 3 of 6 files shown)


Indanz:
Have you tried running sel4bench? Just  curious.

omeh-a:
> Have you tried running sel4bench? Just curious.

Yes, you it passes all tests :)
(It needs two disabled due to lack of a peripheral timer and small RAM however)

omeh-a:
> Have you tried running sel4bench? Just curious.

Not yet, I can give it a shot though

omeh-a:
> It's just that I'm also working on a port (i.MX93) and I ran into issues with sel4bench, sel4test just worked.

It seems to fail for this platform too. seL4bench hangs forever once it begins `irquser` benchmarks. It works fine for the Microkit however.

omeh-a:
> Note: I have added seL4test support [here](https://github.com/omeh-a/sel4test/tree/cheshire), but blocked on opening a PR as the version of openSBI in seL4test doesn't support the board. It requires [this commit](https://github.com/pulp-platform/opensbi/tree/cheshire) nominally. It should be possible to use a generic install of another newer version however which I will test soon

Update to this: Cheshire doesn't appear to work with a generic version of opensbi.

Indanz:
> It seems to fail for this platform too. seL4bench hangs forever once it begins `irquser` benchmarks.

I found and [fixed some bugs](https://github.com/seL4/sel4bench/pull/50) in sel4bench which were triggered mostly when RELEASE=OFF. But you need to manually edit the settings file to get it to build a debug build though, so probably doesn't apply to you.

The IRQ benchmark is broken, but IRQUSER should work though. Does the rest pass if you disable that particular test via e.g. `ccmake .`?

omeh-a:
> > It seems to fail for this platform too. seL4bench hangs forever once it begins `irquser` benchmarks.
> 
> I found and [fixed some bugs](https://github.com/seL4/sel4bench/pull/50) in sel4bench which were triggered mostly when RELEASE=OFF. But you need to manually edit the settings file to get it to build a debug build though, so probably doesn't apply to you.
> 
> The IRQ benchmark is broken, but IRQUSER should work though. Does the rest pass if you disable that particular test via e.g. `ccmake .`?

Interestingly, nothing changes in terms of the actual test output with your patch AND with disabling IRQUSER! It still gets to printing `irquser Benchmarks` and then promptly hangs the entire board. SMP is also turned off because this FPGA design only has one core. Not super clear why this fails, although this is also my first time using sel4bench so I am not very familiar.

Indanz:
> Interestingly, nothing changes in terms of the actual test output with your patch AND with disabling IRQUSER! It still gets to printing `irquser Benchmarks` and then promptly hangs the entire board

The irquser test is fairly [simple](https://github.com/seL4/sel4bench/blob/master/apps/irquser/src/main.c), add some printfs to see how far it gets and where it hangs. It could very well be just buggy and you're just the unlucky one to run into it first. (In particular, I don't understand how the test can work if the main thread's priority is lower than `seL4_MaxPrio - 2`.)

Reasons why it breaks could be because you don't get timer interrupts. If you disable the irqusr test, do all other tests pass or does it get stuck somewhere else?

Ivan-Velickovic (approved):
Since it's a new platform we will need a docs page for it as well.

Since Cheshire is based on CVA6 I imagine the docs page will look pretty similar to https://github.com/seL4/docs/blob/master/Hardware/ariane.md.

omeh-a:
> Since it's a new platform we will need a docs page for it as well.
> 
> Since Cheshire is based on CVA6 I imagine the docs page will look pretty similar to https://github.com/seL4/docs/blob/master/Hardware/ariane.md.

https://github.com/seL4/seL4/pull/1354

src/plat/cheshire/config.cmake
```diff
@@ -0,0 +1,34 @@
...
+    set(OPENSBI_PLAT_ISA "rv64imafdc_zicsr_zifencei")
+    list(APPEND KernelDTSList "tools/dts/cheshire.dts")
+    list(APPEND KernelDTSList "src/plat/cheshire/overlay-cheshire.dts")
+    # This is an experimental platform that supports accessing peripherals, but
+    # the status of support for external interrupts via a PLIC is unclear and
+    # may differ depending on the version that is synthesized. Declaring no
+    # interrupts and using the dummy PLIC driver seems the best option for now
+    # to avoid confusion or even crashes.
+    declare_default_headers(
+        TIMER_FREQUENCY 1000000 PLIC_MAX_NUM_INT 30
```
Ivan-Velickovic, line 28:
```suggestion
        TIMER_FREQUENCY 1000000
```

src/plat/cheshire/config.cmake
```diff
@@ -0,0 +1,34 @@
...
+        MAX_IRQ 30
```
Ivan-Velickovic, line 29:
```suggestion
        MAX_IRQ 51
```

should be 51 according to `riscv,ndev = <51>;` in the DTS node for the PLIC.

Ivan-Velickovic (approved):

PR #1355, opened 2024-11-25
arch/riscv64: revise comments

yf13:
This revises comments of riscv64 kernel address mapping for easier reading purposes.

### DIFF
--- a/include/arch/riscv/arch/64/mode/hardware.h
+++ b/include/arch/riscv/arch/64/mode/hardware.h
@@ -33,7 +33,7 @@
  *                |  |         Kernel ELF          |
  *            ----|  +-------------KERNEL_ELF_BASE-+ --+ 2^64 - 2GiB + (KERNEL_ELF_PADDR_BASE % 1GiB)
  *            |   |  |                             |
- *            |   -> +-----------------------------+ --+ 2^64 - 2GiB = (KERNEL_ELF_BASE % 1GiB)
+ *            |   -> +--------------------PPTR_TOP-+ --+ 2^64 - 2GiB
  * Shared 1GiB|      |                             |   |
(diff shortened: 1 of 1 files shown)


Indanz:
First impression is that you're changing too much at once and it's overall not obviously better than what we have now, when comparing these side-by-side:

https://raw.githubusercontent.com/seL4/seL4/refs/heads/master/include/arch/riscv/arch/64/mode/hardware.h
https://raw.githubusercontent.com/seL4/seL4/0d608251f653679de2b7cf93a784f89c97dc6a67/include/arch/riscv/arch/64/mode/hardware.h

Maybe have separate commits for content changes/fixes and indentation/layout changes.

I like the addition of PPTR_TOP, removal of the nonsensical `(KERNEL_ELF_BASE % 1GiB)`. I don't like the squashing things to the left so much.

Frankly, I think the whole thing would be clearer if the addresses were printed too, like Arm does. There they keep it narrow by putting the next level translations underneath instead of next to the previous block.

yf13:
@Indanz, I don't understand why changes in comments led to differences in ASTs. would you mind take a look at this?

yf13:
@Indanz, okay I will split contents change from layout changes.

Indanz:
On 2024-11-25 20:26, yf13 wrote:
> @Indanz [1], I don't understand why changes in comments led to
> differences in ASTs. would you mind take a look at this?

It has nothing to do with your changes, just ignore.

It's because I merged a bugfix into master and the proofs
haven't been synced yet. Someone from the verification
side needs to do that, but the person doing it normally
is currently unavailable. Until then, all CI runs will
fail that test with the same error.

kent-mcleod (approved):
I agree that the 2 line change makes the diagram clearer.

PR #1356, opened 2024-11-29
Add i.MX93 SoC support

Indanz:
This generic support for any i.MX93 SoC based platform. It comes with up to two CPU cores and up to 2 GB of memory. Specific board differences can be resolved easily by using `KernelCustomDTSOverlay` to tweak the details.

~~One known issue is that there are some bogus device memory regions created, I will update the PR when I found out what causes that.~~

~~The board has an undocumented gap of 32 MB in its memory space at 0x96000000, I will ask NXP where it's coming from.~~

### DIFF
--- a/src/plat/imx93/config.cmake
+++ b/src/plat/imx93/config.cmake
@@ -1,0 +1,33 @@
+#
+# Copyright 2024, Indan Zupancic
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+cmake_minimum_required(VERSION 3.7.2)
+
+declare_platform(imx93 KernelPlatformIMX93 PLAT_IMX93 KernelArchARM)
+
+if(KernelPlatformIMX93)
+    declare_seL4_arch(aarch64)
+    set(KernelArmCortexA55 ON)
+    set(KernelArchArmV8a ON)
+    set(KernelArmGicV3 ON)
+    config_set(KernelARMPlatform ARM_PLAT ${KernelPlatform})
+    list(APPEND KernelDTSList "tools/dts/${KernelPlatform}.dts")
+    list(APPEND KernelDTSList "src/plat/imx93/overlay-${KernelPlatform}.dts")
+    declare_default_headers(
+        TIMER_FREQUENCY 24000000
+        TIMER drivers/timer/arm_generic.h
+        TIMER_OVERHEAD_TICKS 1
+        NUM_PPI 32
+        MAX_IRQ 254
+        INTERRUPT_CONTROLLER arch/machine/gic_v3.h
+        KERNEL_WCET 10u
+    )
+endif()
+
+add_sources(
+    DEP "KernelPlatformIMX93"
+    CFILES src/arch/arm/machine/gic_v3.c src/arch/arm/machine/l2c_nop.c
+)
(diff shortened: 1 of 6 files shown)


Indanz:
The mysterious 32 MB gap is reserved by ATF for BL32 (ELE) it [seems](https://github.com/nxp-imx/imx-atf/blob/616a4588f333522d50a55bedd2b9a90a51474a75/plat/imx/imx93/platform.mk#L52).

Ivan-Velickovic:
@Indanz can you add a PR to the docs as well?

src/plat/imx93/overlay-imx93.dts
```diff
@@ -0,0 +1,30 @@
...
+			&{/timer};
+	};
+
+	/* Maximum of 2 GB, use KernelCustomDTSOverlay if your board has less */
+	memory@80000000 {
+		device_type = "memory";
+		reg = <0x0 0x80000000 0x0 0x16000000>;
+	};
+
+	/* Undocumented 32MB gap between these two regions at 0x96000000!? */
```
Ivan-Velickovic, line 25: Document that it's because of ATF?

lsf37:
Relevant companion PRs:
- https://github.com/seL4/seL4_tools/pull/211
- https://github.com/seL4/util_libs/pull/189
- https://github.com/seL4/sel4bench/pull/50
- https://github.com/seL4/ci-actions/pull/372

src/plat/imx93/config.cmake
```diff
@@ -0,0 +1,35 @@
...
+declare_platform(imx93 KernelPlatformIMX93 PLAT_IMX93 KernelArchARM)
+
+if(KernelPlatformIMX93)
+    declare_seL4_arch(aarch64)
+    set(KernelArmCortexA55 ON)
+    set(KernelArchArmV8a ON)
+    set(KernelArmGicV3 ON)
+    config_set(KernelARMPlatform ARM_PLAT ${KernelPlatform})
+    set(KernelArmVtimerUpdateVOffset OFF)
+    set(KernelArmDisableWFIWFETraps ON)
```
Ivan-Velickovic, line 18: These are specific to the deployment of a system using the platform rather than something all users of the platform should have right?

lsf37: It looks like these two are from the setup of `tqma8xqp1gb` that also has then set to the same defaults. But I agree, unless they are necessary for the platform, they should be project settings, not platform settings.

Ivan-Velickovic: Yea I thought they had just been copy and pasted.

They look like project settings to me.

Indanz: Exactly what happened. But I think `KernelArmVtimerUpdateVOffset` at least should be default on, otherwise timing will be messed up in the VM. `KernelArmDisableWFIWFETraps` should stay off by default. But agreed that the default shouldn't be determined by platform settings.

src/plat/imx93/overlay-imx93.dts
Indanz: Done locally already, will push later today.

Indanz:
(The mysterious device regions is normal seL4 behaviour: It just passes everything that's not normal memory and not kernel reserved memory as device memory to user space.)

lsf37 (approved):
This is now good to merge from my side.

PR #1357, opened 2024-11-30
riscv: some trivial fixes

yf13:
This contains comment revision and code simplification encountered when reading riscv vspace sources.

### DIFF
--- a/src/arch/riscv/kernel/vspace.c
+++ b/src/arch/riscv/kernel/vspace.c
@@ -123,7 +123,7 @@ BOOT_CODE VISIBLE void map_kernel_window(void)
         pptr += RISCV_GET_LVL_PGSIZE(0);
         paddr += RISCV_GET_LVL_PGSIZE(0);
     }
-    /* now we should be mapping the 1GiB kernel base */
+    /* now map KERNEL_ELF_BASE to KERNEL_ELF_PADDR_BASE */
     assert(pptr == PPTR_TOP);
     pptr = ROUND_DOWN(KERNEL_ELF_BASE, RISCV_GET_LVL_PGSIZE_BITS(0));
     paddr = ROUND_DOWN(KERNEL_ELF_PADDR_BASE, RISCV_GET_LVL_PGSIZE_BITS(0));
@@ -157,7 +157,7 @@ BOOT_CODE VISIBLE void map_kernel_window(void)
         pte_next(kpptr_to_paddr(kernel_image_level2_dev_pt), false);
 #endif
 
-    /* There should be 1GiB free where we put device mapping */
+    /* There should be free space where we put device mapping */
     assert(pptr == UINTPTR_MAX - RISCV_GET_LVL_PGSIZE(0) + 1);
     map_kernel_devices();
 }
--- a/include/arch/riscv/arch/machine/hardware.h
+++ b/include/arch/riscv/arch/machine/hardware.h
@@ -45,7 +45,7 @@
  * configured RISC-V system with CONFIG_PT_LEVEL (which can be 2 on Sv32,
  * 3 on Sv39, or 4 on Sv48)
  */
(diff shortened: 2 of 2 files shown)


src/arch/riscv/kernel/vspace.c
```diff
@@ -105,14 +105,13 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
 
 BOOT_CODE VISIBLE void map_kernel_window(void)
 {
-    /* mapping of KERNEL_ELF_BASE (virtual address) to kernel's
-     * KERNEL_ELF_PHYS_BASE  */
+    /* mapping of physical memory, kernel ELF image and devices. */
```
lsf37, line 108: The kernel ELF image also resides in physical memory, and we can only map physical memory or devices, so I don't think this change is an improvement.

The page tables are a mapping from virtual to physical addresses, so the information that we are mapping KERNEL_ELF_BASE to KERNEL_ELF_PHYS_BASE is actually useful information which should not be removed from the comment.

src/arch/riscv/kernel/vspace.c
```diff
@@ -105,14 +105,13 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
     assert(CONFIG_PT_LEVELS > 1 && CONFIG_PT_LEVELS <= 4);
 
     /* kernel window starts at PPTR_BASE */
     word_t pptr = PPTR_BASE;
 
-    /* first we map in memory from PADDR_BASE */
+    /* first we map physical memory from PADDR_BASE */
```
lsf37, line 114: "map in" is a correct idiomatic expression, I don't see any need to change it. We could add `physical` to make it extra clear that this is not about devices, but `PADDR_BASE` does already imply that.

src/arch/riscv/kernel/vspace.c
```diff
@@ -136,7 +135,7 @@ BOOT_CODE VISIBLE void map_kernel_window(void)
     kernel_root_pageTable[RISCV_GET_PT_INDEX(KS_LOG_PPTR, 0)] =
         pte_next(kpptr_to_paddr(kernel_image_level2_log_buffer_pt), false);
 #endif
-#else
+#else /* __riscv_xlen != 32 */
```
lsf37, line 138: This comment is ambiguous, it could mean that the following block is for the `!=` case, or it could mean that the `!=` is completed, and the `else` case (the `==` case) follows. Having no comment here was better. I'd be fine with just `/* __riscv_xlen */` as comment, then at least there is no ambiguity.

src/arch/riscv/kernel/vspace.c
```diff
@@ -123,7 +122,7 @@ BOOT_CODE VISIBLE void map_kernel_window(void)
         pptr += RISCV_GET_LVL_PGSIZE(0);
         paddr += RISCV_GET_LVL_PGSIZE(0);
     }
-    /* now we should be mapping the 1GiB kernel base */
+    /* now map the 1GiB where kernel ELF image lives */
```
lsf37, line 125: Neither the old nor the new comment are great. Maybe instead: `/* now map KERNEL_ELF_BASE to KERNEL_ELF_PADDR_BASE as a 1 GiB window */`. Is the 1GiB actually correct for all configs? I haven't looked, but the could just below looks like it would depend on what size a level 0 page is.

src/arch/riscv/kernel/vspace.c
lsf37: (of course it's clear which case it is when you scroll up to look at the code, but then you don't need a comment)

yf13:
Thank you @lsf37! I sent an update. Yes the 1GiB thing looks improper for rv32 so I dropped them.

Indanz (approved):

lsf37 (approved):

lsf37:
> Thank you @lsf37! I sent an update. Yes the 1GiB thing looks improper for rv32 so I dropped them.

Thanks!

PR #1361, opened 2024-12-03
machine/statedata: align idle thread

yf13:
The `TCB_PTR_CTE_PTR()` macro assumes that thread block aligns with `bit(seL4_TCBBits)`. However this is not true for the idle thread block, thus using the macro with idle thread leads to invalid results. This fixes the issue by apply same alignment to idle thread block so that the macro applies to all threads.

### DIFF
--- a/src/model/statedata.c
+++ b/src/model/statedata.c
@@ -98,7 +98,7 @@ word_t ksDomScheduleIdx;
 word_t tlbLockCount = 0;
 
 /* Idle thread. */
-SECTION("._idle_thread") char ksIdleThreadTCB[CONFIG_MAX_NUM_NODES][BIT(seL4_TCBBits)] ALIGN(BIT(TCB_SIZE_BITS));
+SECTION("._idle_thread") char ksIdleThreadTCB[CONFIG_MAX_NUM_NODES][BIT(seL4_TCBBits)] ALIGN(BIT(seL4_TCBBits));
 
 #ifdef CONFIG_KERNEL_MCS
 /* Idle thread Schedcontexts */


Indanz:
`TCB_PTR_CTE_PTR` should never be used on the idle thread, how did you ran into this issue?

lsf37:
> `TCB_PTR_CTE_PTR` should never be used on the idle thread, how did you ran into this issue?

I don't think we have explicitly proved this, but I agree, the macro should never be applied to the idle thread. If that happens we have bigger problems.

Having said that, the verification does have the blanket assumption on the initial state that all TCBs are aligned the same way and the idle thread TCB is still a TCB, so I would actually vote for making this change for that reason even though there should be no correctness issue -- if we verify kernel init at some point, we're going to need it to establish that assumption.

lsf37:
I would ask to update the commit message, though, to something along these lines:

```
Change the alignment of the idle thread to be consistent with standard
TCBs. Even though macros like TCB_PTR_CTE_PTR should not be used
on the idle thread, verification assumes consistent alignment for all TCBs.
```

yf13:
The `setThreadName()` uses `TCB_PTR_DEBUG_PTR` which uses `TCB_PTE_CTE_PTR`. Idle thread has name set with debug build.

lsf37:
> The `setThreadName()` uses `TCB_PTR_DEBUG_PTR` which uses `TCB_PTE_CTE_PTR`. Idle thread has name set with debug build.

Well, that's indeed bad. I'm definitely in favour of merging this then. Verification would not pick up on debug mode uses, so nice job spotting this.

I wonder why this has not led to problems so far. It's possible that the linker is picking a larger alignment for the idle thread section anyway and that has saved things. Otherwise setting the thread name would have been corrupting other state at boot time already.

lsf37 (approved):

PR #1362, opened 2024-12-05
document words ordering for bitfield_gen

yf13:
This adds words ordering to bitfield_gen.md.


Indanz:
The paragraph above the one you added tried explaining the word ordering already:

```
The bits in the block are laid out in memory in the order they are mentioned in
the specification. That is, the left-most 32 bits (i.e. bits 63 to 32) of
`VMFault` will correspond to the `address` field, the next 5 bits to `FSR`, the
next 7 bits are inaccessible padding which is set to 0 on creation of the
bitfield, then one bit for `instructionFault`, 15 bits of padding, and finally 4
bits (bits 0 to 3 of the entire type) for `seL4_FaultType`.
```

Your comment isn't much clearer than the above, so I suggest improving what there is instead of adding more fuzzy description.

Personally I would just say that the bits are specified in MSB order, while they will be in Little-Endian order in memory. Meaning that the bits in the block are laid out in memory in reverse order they are mentioned in
the specification.

yf13:
@Indaz, sorry for that.

Let me try to explain by an example:

```
base 64
block sample {
  field a 32
  field b 32
  field b 32
  field c 32
  field d 64
}
```

What I wanted to mean is that field `a` and `b` are in `words[2]`. field `b` and `c` are in `words[1]` and field `d` is in words[0].

I felt the paragraph you quoted is about intra-word bits ordering, what I wanted to add is inter-word ordering.

yf13:
How about we add `words[1]` and `words[1]` to the right of the two machine words in the diagram instead?

Indanz:
> @indaz, sorry for that.

No need to apologise!

> 
> Let me try to explain by an example:
> 
> ```
> base 64
> block sample {
>   field a 32
>   field b 32
>   field c 32
>   field d 32
>   field e 64
> }
> ```
> 
> What I wanted to mean is that field `a` and `b` are in `words[2]`. field `c` and `d` are in `words[1]` and field `d` is in `words[0]`.
> 
> I felt the paragraph you quoted reads like ordering in a 64 bit word (as the diagram doesn't show words boundary), what I failed to find from current document is the inter-word ordering.

Just above the quoted bit it says:

"Assuming that the word size of the underlying architecture is 32 bits, this `VMFault_t` type will have a size of two machine words."

So it's trying to do exactly the same as you are.

That you missed this probably means the example is not clear enough one way or the other and that it could be improved.

yf13:
@lsf37, thanks for the comments.

Here I want see memory usage by scanning CSpace using Python, thus I need know how words are stored in memory for a multiword block. Currently only by reading the generated source, one can know words are stored in reverse order.

Indanz:
> The diagram specifies the bit numbers in the order they end up in memory -- that is the part that is important. 

Except the order in memory is reversed: The first field define ends up last in memory, not firs like described.

> Happy to hear your thoughts on what exactly is unclear, maybe adding more explanation is the right way to go after all.

I think talking about "left-most 32 bits" is very unclear. Upper and lower I understand, but what is left-most supposed to be? It seems to refer to big-endian hexadecimal notation of numbers, but that's not really relevant for bf.

Indanz:
Feel free to re-open if you have a new version, or open a new PR. Closing this now because we won't merge this version and there wasn't activity for a year.

Issue #1053, opened 2023-06-05
Assertion triggerred in SMP+MCS kernel on HiFive Unleashed


Ivan-Velickovic:
The following assertion has been seen to non-deterministically trigger on the HiFive Unleashed platform:
```
seL4 failed assertion 't0 <= margin + t && t <= t0 + margin' at /github/workspace/kernel/src/kernel/boot.c:563 in function clock_sync_test
```

This has been seen on the debug SMP+MCS kernels. An example of the assertion being triggered can be found here https://github.com/seL4/seL4/actions/runs/5141337447/jobs/9276716395.

The full logs with the assertion failing for `HIFIVE_debug_SMP_MCS_gcc_64` can be found here https://gist.github.com/Ivan-Velickovic/c6a08044ed68228edd0b54533b8f9bad.

Ivan-Velickovic:
The following command can be used to build the sel4test image: `../init-build.sh -DRISCV64=TRUE -DPLATFORM=hifive -DBAMBOO=TRUE -DSMP=TRUE -DMCS=TRUE`

Indanz:
The line before the assertion is:

> clock_sync_test[1]: t0 = 23325675, t = 23325678, td = 3

The timer frequency for the HiFive is 1 MHz, so this means that the time difference between two cores is 3 us. When adding this test I already added an arbitrary 1 us margin for the HiFive, but apparently that's not enough.

To rule out a bug in `clock_sync_test()`, does replacing `__atomic_thread_fence(__ATOMIC_ACQUIRE)` with `__atomic_thread_fence(__ATOMIC_ACQ_REL)` help? That might prevent reordering `ksNumCPUs++` with the clock sync test, but I don't expect it to help.

I suspect the real fix is solving #942, which would improve RISCV performance greatly.

Ivan-Velickovic:
Since closing https://github.com/seL4/seL4/issues/942 I have not observed this failure anymore. If the assertion is triggered again we can re-open this issue.

PR #1366, opened 2024-12-18
hifive: reflect lower TIMER_PRECISION in config

lsf37:
Reflect the slow timer read on this platform by increasing the value for TIMER_PRECISION.

MCS+SMP+debug tests have been failing the clock sync boot test (see also issue #1053). The proposed fix, which was to make reading the timer faster, only works for certain OpenSBI versions and is not enabled by default.

### DIFF
--- a/src/plat/hifive/config.cmake
+++ b/src/plat/hifive/config.cmake
@@ -20,6 +20,7 @@ if(KernelPlatformHifive)
         TIMER_FREQUENCY 1000000
         MAX_IRQ 53
         INTERRUPT_CONTROLLER drivers/irq/riscv_plic0.h
+        TIMER_PRECISION 3u
     )
 else()
     unset(KernelPlatformFirstHartID CACHE)


Indanz:
RISC-V's `getTimerPrecision` is hard-coded to 1 us and doesn't use `TIMER_PRECISION`. And that is correct according to the RISC-V standard.

`getTimerPrecision` is also used by `setNextInterrupt`, so you can't change that easily either.

I think the correct solution is to add extra margin to `clock_sync_test` instead. Currently it hard-codes an extra margin of 1 us, we could increase that to 2 us and that will probably fix it for RISC-V.

The problem is that if multiple cores query the time at the same moment, it sometimes takes multiple microseconds before one of the queries returns (probably because of global lock in OpenSBI), not that RISC-V's timer precision isn't 1 us.

lsf37:
You're right, changing `TIMER_PRECISION ` here makes no sense, I missed that RISCV isn't making use of it, so will close the PR.

Need to think a bit more about this. For instance, does the fact that it might take multiple microseconds to read the timer on SMP affect what we should do in `setNextInterrupt`? It's not really the timer precision, but the value might still be off.

Indanz:
> Need to think a bit more about this. For instance, does the fact that it might take multiple microseconds to read the timer on SMP affect what we should do in `setNextInterrupt`? It's not really the timer precision, but the value might still be off.

Assuming I'm right and it's caused by a global lock in OpenSBI, then it's the same problem seL4 already has with it's own global lock on SMP, and the timer interrupt can already be delayed by up to `(NUM_CORES - 1) * WCET`. It doesn't really matter if the delay is caused when getting the time or when taking the IRQ. So I'd ignore it and don't do anything special for RISC-V for now. Perhaps actually use `TIMER_PRECISION` in `getTimerPrecision` for RISC-V in the future.

But getting the time taking long doesn't necessarily mean that the time you get is old, it can still be correct within 1 us, it just took longer than expected to get it. It's just that the time sync test can't see the difference.

lsf37:
Just logging here that we've seen the test fail with `td = 4` and `td = 5` recently, so it does seem to need a fairly big margin on RISCV.

Indanz:
What really annoys me is that RISCV is making all the historical mistakes again instead of learning from the past. It's a new architecture, there is no excuse to be so crappy in so many ways.

PR #1370, opened 2024-12-19
boot: clear freeIndex of memory untypeds

yf13:
Original untyped memory objects appear as fully used right after boot, thus confuses free memory calculation. This uses 0 freeIndex so that free memory calculation based on it is right at the beginning.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -693,7 +693,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
             .isDevice = device_memory,
             .padding  = {0}
         };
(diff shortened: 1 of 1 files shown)


lsf37 (changes requested):
This is not the right change. Device untyped and normal untyped should be set up uniformly at boot, and the overall protocol is to defer initialisation to first use, which is achieved by setting the `MAX_FREE_INDEX`.

I assume there is some debug code that is trying to compute the amount of free memory. It's that debug code that should be fixed. If an untyped cap has no children, then it can be fully reset at next retype, so it should be reported as having all its memory available regardless of the free index.

PR #1372, opened 2024-12-19
aarch64: fix building with clang-18

cl91:
On aarch64, clang interprets the compile option `-mgeneral-regs-only` as disabling all FPU-related registers and instructions, which not only include those generated by the compiler (which is how gcc would interpret this option), but also those in (inline) assemblies written explicitly by the programmer, which gcc does not forbid with this option. This commit enables FPU-related registers and instruction explicitly in the FPU-context switching code, so the kernel will build under clang-18 on aarch64.

LLVM project's issue tracking the inconsistency of this option between clang and gcc: https://github.com/llvm/llvm-project/issues/30140 --- it remains an open issue as of this writing.

Similar commit in other projects: https://github.com/TF-RMM/tf-rmm/commit/ffcc90415364933dcd5670226a0b9c9573f2c939

### DIFF
--- a/include/arch/arm/arch/64/mode/machine/fpu.h
+++ b/include/arch/arm/arch/64/mode/machine/fpu.h
@@ -18,6 +18,7 @@ static inline void saveFpuState(user_fpu_state_t *dest)
 
     asm volatile(
         /* SIMD and floating-point register file */
+        ".arch_extension fp\n"
         "stp     q0, q1, [%1, #16 * 0]      \n"
         "stp     q2, q3, [%1, #16 * 2]      \n"
         "stp     q4, q5, [%1, #16 * 4]      \n"
@@ -40,6 +41,7 @@ static inline void saveFpuState(user_fpu_state_t *dest)
         "str     %w0, [%1, #16 * 32]        \n"
         "mrs     %0, fpcr                   \n"
         "str     %w0, [%1, #16 * 32 + 4]    \n"
+        ".arch_extension nofp\n"
         : "=&r"(temp)
         : "r"(dest)
         : "memory"
@@ -53,6 +55,7 @@ static inline void loadFpuState(user_fpu_state_t *src)
 
     asm volatile(
         /* SIMD and floating-point register file */
+        ".arch_extension fp\n"
         "ldp     q0, q1, [%1, #16 * 0]      \n"
         "ldp     q2, q3, [%1, #16 * 2]      \n"
         "ldp     q4, q5, [%1, #16 * 4]      \n"
@@ -75,6 +78,7 @@ static inline void loadFpuState(user_fpu_state_t *src)
         "msr     fpsr, %0                   \n"
         "ldr     %w0, [%1, #16 * 32 + 4]    \n"
         "msr     fpcr, %0                   \n"
+        ".arch_extension nofp\n"
         : "=&r"(temp)
         : "r"(src)
         : "memory"


cl91:
The commit that changed the behavior between clang-17 and clang-18 is https://github.com/llvm/llvm-project/pull/77817 --- note that Fuchsia is also affected by this change, and the suggested workaround by an LLVM developer is to add the `.arch_extension fp` line, as is done in this commit.

Indanz (approved):
Strictly speaking, `.arch_extensions` is only for optional architecture extensions, and FP is mandatory for Aarch64.

Please mention the clang/llvm version this is needed for in your commit message.

Can you test with Clang 19 or newest git too?

cl91:
> Strictly speaking, `.arch_extensions` is only for optional architecture extensions, and FP is mandatory for Aarch64.

Yes, this is why clang's behavior is incorrect, although they don't plan to fix it.

> 
> Please mention the clang/llvm version this is needed for in your commit message.
> 
I updated the commit message accordingly.

> Can you test with Clang 19 or newest git too?
Sure I'll test the newest git version.

Indanz:
If you add a signed-of-by to the commit (git commit --amend -s), I guess it's ready to go from my end.

I don't like working around Clang's bugs this way, and I don't like the lack of documentation of `.arch_extension`. What is its scope? Does the ".arch_extension nofp" apply to anything outside the inline assembly? If yes, is it at least limited to the file?

I assume that Clang gives compile errors if this isn't done. Please compile with HYP and SMP enabled too to see if there is anything else.

cl91:
> If you add a signed-of-by to the commit (git commit --amend -s), I guess it's ready to go from my end.

Done.
 
> I assume that Clang gives compile errors if this isn't done. Please compile with HYP and SMP enabled too to see if there is anything else.

Yes the compile errors are
```
kernel/include/arch/arm/arch/64/mode/machine/fpu.h:21:9: error: instruction requires: fp-armv8
   21 |         "stp     q0, q1, [%1, #16 * 0]      \n"
      |         ^
<inline asm>:1:2: note: instantiated into assembly here
    1 |         stp     q0, q1, [x11, #16 * 0]      
```
I tested with LLVM git-20.0.0_r521848.a9df1f6cb0dc and the issue persists. After applying this patch the build errors are gone, even with SMP and HYP enabled. Note for LLVM git-20 a new compile error appears: `clang: error: argument unused during compilation: '-c' [-Werror,-Wunused-command-line-argument]`. You'd need to add `-Wno-unused-command-line-argument` to the kernel compile options to build successfully.

> I don't like working around Clang's bugs this way, and I don't like the lack of documentation of `.arch_extension`. What is its scope? Does the ".arch_extension nofp" apply to anything outside the inline assembly? If yes, is it at least limited to the file?

I believe it is limited to the compilation unit which for seL4 means the whole kernel (more or less). This is the desired behavior (it is the whole point of `-mgeneral-regs-only` --- the rest of the kernel is supposed to have FPU instructions disabled).

kent-mcleod (approved):
This fixes the issue for me. And doesn't seem to introduce any noticeable errors for other toolchains based on the CI builds.

lsf37:
Have set the DCO to pass (the tool couldn't figure out that the names actually do match, which is not strictly required anyway, but good when it is at least flagged).

Will wait for the build to pass on all config, after that this is ready to merge.

lsf37 (approved):
Ready to go when build tests have passed.

PR #1373, opened 2025-01-05
configs: add remaining aarch64 plaforms for verification

lsf37:
This adds verified configurations for the remaining AArch64 platforms. The functional correctness proofs now support all AArch64 boards that seL4.

Should be merged after #1337  and seL4/l4v#846.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -309,6 +309,14 @@ else()
     message(FATAL_ERROR "unknown KernelArch '${KernelArch}'")
 endif()
 
+if(
+    (CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    AND (CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL "20.0.0")
+)
+    # To avoid clang: error: argument unused during compilation: '-c'
+    add_compile_options(-Wno-unused-command-line-argument)
+endif()
+
 # Sort the C sources to ensure a stable layout of the final C file
 list(SORT c_sources)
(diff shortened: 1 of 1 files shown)


lsf37:
(style check will fail on the x bit for configs as usual)

Indanz (approved):

kent-mcleod (approved):

configs/AARCH64_quartz64_verified.cmake
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2025, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+include(${CMAKE_CURRENT_LIST_DIR}/include/AARCH64_verified_include.cmake)
+
+set(KernelPlatform "quartz64" CACHE STRING "")
```
kent-mcleod, line 10: Is there any cost to adding these additional platforms as "verified" when the platform may not necessarily meet a quality threshold on the practical systems side? eg this Pine64 RK3566 Quartz64-A Board platform isn't on the docsite and I don't think it has enabled hardware tests.

lsf37: My guideline so far has been to add the things that are built in CI, but if it's not even on the supported platforms list, we should probably not include it, even if the proofs pass.

lsf37: Have removed it now.

Issue #1374, opened 2025-01-21
Riscv, kernel trapping, and "double" faults


midnightveil:
Whilst modifying a RISC-V build of seL4, I was getting messages that looked like this:

```
Caught cap fault in send phase at address 0
while trying to handle:
vm fault on data at address 0 with status ��7
in thread ��ffffffff8401bc00 "idle_thread" at address 0
With stack:
Invalid vspace
```

The kernel would then continue as if nothing had happened.

This was eventually traced down to the modifications corrupting the stack, eventually resulting in a memory access fault within the kernel memory, which got attributed to the idle thread. [^0]

[^0]: It was in particular when switching to the idle thread, so I think there was nothing that would stop an exception inside s-mode from being attributed to an arbitrary thread.

1. I am under the impression that seL4 should not have nested exceptions (in normal operation).

2. It appears that the AArch64 traps.S code calls a `kernelDataAbort()` function if there is a trap taken whilst within the kernel, as per https://github.com/seL4/seL4/blob/1732c05ef6b75fd18ce5161d099f0d73e7fcb6a3/src/arch/arm/64/traps.S#L103-L130

3. In contrast, the RISC-V [traps.S](https://github.com/seL4/seL4/blob/master/src/arch/riscv/traps.S) assumes that any trap is from u-mode. I'm not quite sure how traps within s-mode end up back here (perhaps openSBI forwards them back?), but this is what I was observing. It looks like [the linux kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/riscv/kernel/entry.S?h=v5.7-rc5#n26) stores a 0 in `sscratch` whilst in the kernel, otherwise it stores it stores the tp (~-> sp), of which we could do a very similar thing: [^1]. Unless of course there's an architectural way to determine this (`mcause`? do we have access to that in s-mode?)

4. We should never get "faults" from the idle thread, since it's a pseudo-thread managed by the kernel?

5. (This is more of a query) The fact that there is a function called `handleDoubleFault` that gets invoked by `handleFault()` when `sendFaultIPC` fails: https://github.com/seL4/seL4/blob/1732c05ef6b75fd18ce5161d099f0d73e7fcb6a3/src/kernel/faulthandler.c#L58-L60
[...]

PR #1375, opened 2025-01-24
riscv: traps coming from s-mode should not be treated as a u-mode trap

midnightveil:
These are never supposed to happen.
Note, that this will still run on the stack (for CONFIG_KERNEL_PRINTING) and so if page tables are corrupted, may fault, but it should be easier to flag this when looking at a debugger, provided that the trap handler code is still valid (but if it isn't, we can't do anything at all).

Not all traps will be handled via the s-mode trap handler; it depends on the value of the `medeleg` register. This appears to default to 0xf0b509 (in QEMU) and so we can catch page faults.

Without this, traps in the kernel will be attributed to a trap of the current running thread.

Previously, some kernel modifications that corrupted memory caused:

```
Caught cap fault in send phase at address 0
while trying to handle:
vm fault on data at address 0 with status ��7
in thread ��ffffffff8401bc00 "idle_thread" at address 0
With stack:
Invalid vspace
```

Now, the kernel prints out:

```
KERNEL ABORT (exception within s-mode)!
scause: 0xf, stval: 0x0
halting...
Kernel entry via Syscall, number: 1, Call
Cap type: 2, Invocation tag: 1
```

Part of the debug improvements as described in #1374.

Question:

- [x] I could do the check against the `SPP` bit only in debug mode? As opposed to always with different behaviour? This closer matches ARM but probably affects verification and syscall perf...
- [x] It probably should be possible to print out the $pc at when the fault happened?
- [ ] Sidenote: the traps.S is really inconsistent, and uses `x1` (which is calling convention the `$ra` register seemingly as a temporary) which is really confusing... I could fix this but the code isn't quite clear enough to me the execution flow to know the the rename from `$x1 -> $ra` would be semantically correct. I used `$t1` and `$t2` which I think should be fine.
- [ ] This should probably set `ksKernelEntry`, since it's what halt prints out, but there's not an appropriate entry kind to use?

### DIFF
--- a/src/arch/riscv/c_traps.c
+++ b/src/arch/riscv/c_traps.c
@@ -108,6 +108,15 @@ void VISIBLE NORETURN c_handle_interrupt(void)
 
 void VISIBLE NORETURN c_handle_exception(void)
 {
+#ifdef CONFIG_DEBUG_BUILD
+    if (read_sstatus() & SSTATUS_SPP) {
+        printf("\n\nKERNEL ABORT (exception within s-mode)!\n");
+        printf("scause: 0x%"SEL4_PRIx_word", stval: 0x%"SEL4_PRIx_word"\n",
+               read_scause(), read_stval());
+        halt();
+    }
+#endif
+
     NODE_LOCK_SYS;
 
     c_entry_hook();


src/arch/riscv/traps.S
```diff
@@ -135,6 +135,12 @@ handle_syscall:
 
 /* Not an interrupt or a syscall */
 exception:
+  /* check if SPP (sstatus[8]) was 1, i.e. exception from the kernel */
+  li    t1, 0x100
```
axel-h, line 139: It there a real gain doing this check in the asm code? I could also happen in the C code of `c_handle_exception() then. Rational is, that if we are here, we have pushed all registers on the stack anyway without crashing again, and `c_handle_kernel_exception()` tries to print something before calling `halt()`. So in the worst case all this gets stuck in a loop anyway and we never reach `halt()`. Otherwise, the machine still seems to work somehow, so we can also do this in C code exception hander

lsf37: If it's under an ifdef for debug builds, verification does not see it.

axel-h (approved):
Thanks, that looks quite nice now.

axel-h:
> Just as a note: there is precedent, for arm32 a very similar structure is used (something I just noticed).

Good point. Could move this to C code now also and print a nice message. I wonder what AARCH64 does here, as far as I know it has exception vector table for such fault.

axel-h:
Seems the checks are still unhappy about the git commit message title length, could you fix this?

axel-h:
> If you can suggest something shorter that still explains what this commit is doing, be my guest. I can't think of anything.

Could just say: "riscv: catch s-mode traps and halt", the body message explains the test

axel-h:
So, looking at the ARM code, in the asm part it enters an infinite loop in this case even in release builds. In RISC-V we have still nothing at the moment in the release builds. I wonder if we should add something small then in the asm code (so verification does not see it) and the loop over a `wfi` to allow the core to rest until watchdog reset or whatever comes. It's better than running the code path that tries to handle a user exception.

What do you think @lsf37, @Indanz

Indanz (approved):
I would also print out `sepc` to know where it went wrong. It would be nice if the debug ifdef wasn't necessary, but it seems it would complicate things too much if removed.

To be clear, what happens is that when we return from kernel mode, interrupts get enabled again and we trap immediately and get here. This only happens if M-mode continues the execution or delegates those traps to S-mode, it could as well halt or reboot the machine.

axel-h:
> I would also print out `sepc` to know where it went wrong. It would be nice if the debug ifdef wasn't necessary, but it seems it would complicate things too much if removed.

Well, there is also `CONFIG_VERIFICATION_BUILD` that we could use to hide something from verification....

> To be clear, what happens is that when we return from kernel mode, interrupts get enabled again and we trap immediately and get here. This only happens if M-mode continues the execution or delegates those traps to S-mode, it could as well halt or reboot the machine.

I do not understand that. It's not about interrupts, but about aborts. But you are right that M-Mode could also handle aborts. Seems just OpenSBI does not do this by default.

Indanz:
> Well, there is also `CONFIG_VERIFICATION_BUILD` that we could use to hide something from verification....

That's currently only used for defining `compile_assert` and `unverified_compile_assert`, it doesn't seem like a good idea to use it for anything else, that would only increase the number of permutations of configs.

> > To be clear, what happens is that when we return from kernel mode, interrupts get enabled again and we trap immediately and get here. This only happens if M-mode continues the execution or delegates those traps to S-mode, it could as well halt or reboot the machine.
> 
> I do not understand that. It's not about interrupts, but about aborts. But you are right that M-Mode could also handle aborts. Seems just OpenSBI does not do this by default.

I meant interrupts as in any kind of traps, not specifically IRQs. My main point was to make clear that we don't have nested interrupts, but that the fault in kernel mode will be handled later, after doing undefined behaviour and trying to return to user space.

lsf37:
> Well, there is also CONFIG_VERIFICATION_BUILD that we could use to hide something from verification....

Like Indan already said for other reasons, that's not what it is for. Nothing should be hidden from verification just to hide it from verification. If you need to do that, chances are you are doing something you should not be doing.

In this case, either it's for debugging, then it should be under a debug config or it can actually happen in a release kernel, then we either need to ask ourselves why and remove that, or change everything about how we think the system works.

PR #1376, opened 2025-01-29
vspace: userError for no unallocated ASID pools

midnightveil:
This existed for ARM32, but not for ARM64, RISCV-V, or X86.

I'm not too fussed on the error message or the existence of the comment, feel free to adjust if desired.


lsf37 (approved):
Happy with more consistency in error reporting.

And it looks like proof automation can deal with it for once, so that is nice as well.

Indanz (approved):

PR #1378, opened 2025-01-29
constants: Add seL4_ASIDPoolMaxNewPools define

midnightveil:
Previously sel4test had
```
#define N_ASID_POOLS ((int)BIT(seL4_NumASIDPoolsBits))
```
and then would do `N_ASID_POOLS - 1` for making new ASID pools. The 1 pool that couldn't be made was the initial thread's ASID. This exports a constant so that this becomes an implementation detail of the kernel.

Please see https://github.com/seL4/sel4test/pull/130 for adjustment to sel4test.

I'm not too fussed on the exact name or implementation of this, but had discussed this with @Ivan-Velickovic for the best way to do this. Happy to change if needed.

---

Justification: As part of the [time protection - kernel cloning work](https://github.com/seL4/l4v/blob/experimental-timeprot/proof/infoflow/timeprotection/timeprot-branches.md#initial-sel4test-bringup), an ASIDpool was allocated for the kernel image clones. This broke the sel4tests that depended on creating as many ASIDpools as they could.
We move the calculation of the number of allocate-able ASID pools to a kernel constant so that the kernel cloning changes won't break users.

### DIFF
--- a/libsel4/arch_include/arm/interfaces/object-api-arch.xml
+++ b/libsel4/arch_include/arm/interfaces/object-api-arch.xml
@@ -440,6 +440,9 @@
                 Together with a capability to <texttt text="Untyped Memory"/>, which is passed as an argument,
                 create an <texttt text="ASID Pool"/>. The untyped capability must represent a
                 4K memory object. This will create an ASID pool with enough space for 1024 VSpaces.
+
+                A total of <texttt text="seL4_ASIDPoolMaxNewPools"/> can be created,
+                as some ASID pools are created by the kernel during boot.
             </description>
             <param dir="in" name="untyped" type="seL4_Untyped"
             description="Capability to an untyped memory object that will become the pool. Must be 4K bytes."/>
--- a/libsel4/arch_include/riscv/interfaces/object-api-arch.xml
+++ b/libsel4/arch_include/riscv/interfaces/object-api-arch.xml
@@ -235,6 +235,9 @@
                 Together with a capability to <texttt text="Untyped Memory"/>, which is passed as an argument,
                 create an <texttt text="ASID Pool"/>. The untyped capability must represent a
                 4K memory object. This will create an ASID pool with enough space for 1024 VSpaces.
+
+                A total of <texttt text="seL4_ASIDPoolMaxNewPools"/> can be created,
+                as some ASID pools are created by the kernel during boot.
             </description>
             <param dir="in" name="untyped" type="seL4_Untyped"
             description="Capability to an untyped memory object that will become the pool. Must be 4K bytes."/>
--- a/libsel4/arch_include/x86/interfaces/object-api-arch.xml
+++ b/libsel4/arch_include/x86/interfaces/object-api-arch.xml
@@ -482,6 +482,9 @@ contain the mapping"/>
                 Together with a capability to <texttt text="Untyped Memory"/>, which is passed as an argument,
                 create an <texttt text="ASID Pool"/>. The untyped capability must represent a
                 4K memory object. This will create an ASID pool with enough space for 1024 VSpaces.
+
+                A total of <texttt text="seL4_ASIDPoolMaxNewPools"/> can be created,
+                as some ASID pools are created by the kernel during boot.
             </description>
             <cap_param append_description='The master ASIDControl capability.'/>
             <param dir="in" name="untyped" type="seL4_Untyped"
--- a/libsel4/include/sel4/constants.h
+++ b/libsel4/include/sel4/constants.h
@@ -70,6 +70,11 @@ typedef enum {
 } seL4_LookupFailureType;
 #endif /* !__ASSEMBLER__ */
 
+/* The number of pools that can be created with seL4_ARCH_ASIDControl_MakePool
+   The initial thread is allocated an ASID pool (seL4_CapInitThreadASIDPool)
+   so we cannot make all 2^seL4NumASIDPoolsBits - but instead, one less. */
+#define seL4_ASIDPoolMaxNewPools (LIBSEL4_BIT(seL4_NumASIDPoolsBits)-1)
+
 #ifdef CONFIG_KERNEL_MCS
 /* Minimum size of a scheduling context (2^{n} bytes) */
 #define seL4_MinSchedContextBits 7


lsf37:
Changing the number of allocatable ASID pools is still going to be a breaking change, no matter if the number has a name or not (if your system needed `n-1` pools before and now can only get `n-2` it's still not going to work, so it's neither source not binary compatible, but just breaking). 

Realistic systems are very unlikely to need that many ASID pools, so we're probably mostly doing something for `sel4test` here, but we could still do that if we wanted to. Having a name for that number at least doesn't sound problematic to me.

Are there any other opinions on this?

lsf37:
> > Changing the number of allocatable ASID pools is still going to be a breaking change, no matter if the number has a name or not (if your system needed n-1 pools before and now can only get n-2 it's still not going to work, so it's neither source not binary compatible, but just breaking).
> 
> This is true in the static case I guess, but for e.g. sel4test where it allocates a dynamic number of pools it's probably nice to know the maximum a priori before getting `seL4_DeleteFirst`? I don't know if there's really any other cases where you want to dynamically allocate ASID pools as opposed to ASID, though...

In the truly dynamic case you wouldn't be able to rely on any max number, because it depends on what the rest of the system is using and how you are doing the dynamic partitioning (if any).

But I'm not actually arguing against providing the name, just pointing out that the goal of not breaking users cannot be achieved if this value changes. I can see scenarios where it could be useful.

> Also, if I'm not mistaken, the proofs also seem to do this 1-off-BIT(highAsidBit ~= numAsidPoolsBits): https://github.com/seL4/l4v/blob/911b2147498fec3d0f8c912a2aff6cfc0503d024/spec/haskell/src/SEL4/Kernel/VSpace/RISCV64.hs#L442?

That's just `<= 2^n-1` instead of `< 2^n`. The proofs always try to use the former form if possible, because the former overflows gracefully and the latter does not.

`<= 2^n-1` does mean `2^n` are available, but the kernel doesn't really care about whether an ASID pool is in the initial thread or not, the initial thread is just another user thread. It could in fact give that ASID pool to the rest of the system if it's done and doesn't need to run any more, so in that sense there are `2^n` pools available.

Indanz:
I don't have a good feeling about this. What you are doing is adding a define that summarises part of the kernel-initial task ABI. Of course it doesn't hurt to have, but it seems worrisome that it's useful, as it implies that the initial task isn't aware of what the ABI is.

That you need this is a hint that you are implementing one startup task per time domain or something like that? How is that supposed to work? Isn't that hard-coding everything at compile time, so very inflexible? And why would they need their own ASID pool anyway?

I guess I'm more worried about the time-protection changes are going in the wrong direction than this change itself. Is there a seL4 kernel branch anywhere I can find?

Indanz:
> There's a link to an (admittedly older, but mostly unchanged) version of the kernel cloning work in the PR description.

Ah, I looked at that, but didn't expect the description to be at the bottom. It would really help to have a short summary of what is meant with "time protection" at the top, followed by the implementation details, and then the practical info where to find the branches etc. The first is obvious to anyone involved with this project, but for others it is not so clear.

The seL4 kernel diff can be found [here](https://github.com/seL4/seL4/compare/master...robs-cse:seL4:wip_timeprot-riscv64-minimal).

> Honestly, I'd probably reconsider this for the moment. I'm not entirely sure why there's even a reserved ASID pool for the kernel clone images - it's possible that it's necessary (someone did add it, for a reason presumably), but after making this PR I've been mulling over trying to figure out why (to a suitable degree) and haven't yet.

In my experience you can't assume anything. Only way to find out is by removing it and see if you run into any problems. It might have been added because it seemed necessary at the time, but then it turned out it's not.

Looking at the code, it's used at domain switch to call `setVSpaceRoot`, probably because it wants to switch to an ASID of the new domain instead of continuing using an ASID of the previous domain.

However, you don't need to reserve an ASID for that (because hw ASIDs should already partitioned by colour, so you can use any). But if you do, you only need a reserved hw ASID per domain, not a user visible translated ASID.

> I don't think it's because of one startup task per kernel domain, though. It's something about the "idle kernel image". But again, I'm not so certain this was necessary.

All the `switchToIdleKernelImage` calls are followed by a `setVSpaceRoot` call with ASID 0, so it's not needed for that. The question is why `setKernelImage` can't use ASID 0 too.

> But if it was necessary, then I don't know how else you would do this aside from at compile time? Either we export a constant from the kernel or we hard code the rules in sel4test.

Hard-coding things like this in sel4test is fine.

Indanz:
Looking at the implementation, I'm not sure that it's implementing cache colouring correctly, as it doesn't take into consideration the LLC size? See [here](https://www.lynx.com/en/resources/what-is-cache-coloring) for a decent description of how cache colouring works.

All this seems like a huge effort complexity wise for not much gain. Wouldn't it be a better long-term solution to use CPU extensions for cache partitioning instead of this approach, like they mention in the article?

I also don't understand why it is tied to domains, wouldn't you want to be able to use this without domains too, especially on SMP?

Indanz:
> Empirically, this works. We observe a timing channel using the kernel LLC cache that is hidden by the implementation of kernel cloning. The code does take into account the number of bits overlapped by the page size and the cache set, that's the colour cache bit define, and that determines the number of colours we have. I am also aware of how cache colouring works, so I'm not quite sure why you linked me that article...

I'm sure you do. I know the principle, but not the details, as I've never looked into it in detail before. To me it's mostly unclear how physical addresses are mapped to a set. Hardware is free to do that any way it wants. That link was the best description I found with a quick search, and it seems to deviate from what appears to be implemented, so I gave the link as context.

For a linear 1-1 mapping from physical address to cache set the middle bits are used to select the set, but how many bits are used depends on the cache size and the number of ways. With cache colouring you want to divide the sets between colours. Looking at `inPageOfColour`, it assumes that picking the lowest bits above the first `PAGE_BITS` always works. I don't see how that can work in general. Perhaps practically it's fine because of power-of-two interleaving. I'll think about it more.

> I think I agree with your explanation around domain switching and the extra ASIDpools as why, this matches my understanding.
> 
> Regarding explaining the relevance to kernel cloning more - yea, probably should of, need to see how much is publicly shareable...
> 
> I don't think setKernelImage can use ASID0 always though... unless I misunderstand you, because if we did we wouldn't be able to have separate cache colours ..

It would be only temporary after domain switch, before immediately selecting a new task (which will load a new ASID). And the operations done are always the same, so it's not leaking any relevant info in any way.

I don't think you can fully hide that a domain switch happened anyway, except perhaps if you switch to ASID 0 before doing the domain switch and can contain all code to cache lines that also contain code that will be run when there is no domain switch. But then the domain switch can be detected by some task switching code not being cached.

And yes, the kernel could actually always use ASID 0 for all its kernel operations, but that would require two extra ASID changes for each syscall, which would be too high overhead probably. But it depends on the goal of this cache colouring whether that's sufficient.

If page colouring isn't bound to domains, the kernel would just switch ASIDs / mapping when switching to a task with a different cache colour and the kernel had a mapping for that colour. Depends on whether the kernel mapping is part of the task's page tables or whether the kernel has its own page tables, that depends on the architecture.

> I suspect the best option is to hold off on this for now, and revisit it if it is necessary.

I'll close this PR for now, you can always re-open this PR later if needed (or ask it to be re-opened if that doesn't work, I know it does for issues).

PR #1381, opened 2025-02-03
platform_gen.h.in: remove platform_interrupt_t symbol

midnightveil:
Presumably intended to be something like

    typedef enum IRQConstants { ... } platform_interrupt_t;

this instead declares a symbol that is exported and kept in the binary, but is never used. Remove it.

Spotted as my kernel.elf binary contained a strange symbol:

     ffffffff8401b7f0 g     O .bss	0000000000000004 platform_interrupt_t

cc @wom-bat

### DIFF
--- a/src/arch/arm/platform_gen.h.in
+++ b/src/arch/arm/platform_gen.h.in
@@ -16,7 +16,7 @@
 
 enum IRQConstants {
     maxIRQ                      = @CONFIGURE_MAX_IRQ@
-} platform_interrupt_t;
+};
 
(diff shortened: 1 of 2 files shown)


lsf37:
Trying to see when this was added. So far, before the open source release in 91b7da862.

lsf37:
Yup, really old from 2007 before the kernel even worked, introduced by DaveC with commit message "Started restructuring, finished machine monad". Really does not look like it was intended for anything specific and if it's unused it can go.

Will wait for the proof test to pass, but I expect that it will.

lsf37 (approved):

lsf37:
Proofs passed, am just updating before merging.

PR #1382, opened 2025-02-03
statedata: only define tlbLockCount for cortex-a8

midnightveil:
Again, variable was noticed in the objdump of a RISC-V build, despite only being used for a specific ARM CortexA8 platform.

Hopefully this one doesn't affect the proofs, it's not referenced in l4v so :)

### DIFF
--- a/include/arch/arm/arch/32/mode/model/statedata.h
+++ b/include/arch/arm/arch/32/mode/model/statedata.h
@@ -34,3 +34,6 @@ extern pde_t armUSGlobalPD[BIT(PD_INDEX_BITS)] VISIBLE;
 extern pte_t   armUSGlobalPT[BIT(PT_INDEX_BITS)]   VISIBLE;
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
+#ifdef CONFIG_ARM_HAS_TLB_LOCK
+extern word_t armKSTlbLockCount VISIBLE;
+#endif /* CONFIG_ARM_HAS_TLB_LOCK */
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -255,6 +255,12 @@ elseif(KernelArmCortexA9)
     config_set(KernelArmCacheLineSizeBits L1_CACHE_LINE_SIZE_BITS "5")
 endif()
 
+if(KernelArmCortexA8)
+    config_set(KernelArmHasTlbLock ARM_HAS_TLB_LOCK ON)
+else()
+    config_set(KernelArmHasTlbLock ARM_HAS_TLB_LOCK OFF)
+endif()
+
 add_sources(
     DEP "KernelArchARM"
     PREFIX src/arch/arm
--- a/include/model/statedata.h
+++ b/include/model/statedata.h
@@ -112,7 +112,6 @@ extern ticks_t ksDomainTime;
 #else
 extern word_t ksDomainTime;
 #endif
-extern word_t tlbLockCount VISIBLE;
 
 extern char ksIdleThreadTCB[CONFIG_MAX_NUM_NODES][BIT(seL4_TCBBits)];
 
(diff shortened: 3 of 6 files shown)


include/arch/arm/arch/32/mode/model/statedata.h
```diff
@@ -34,3 +34,6 @@ extern pde_t armUSGlobalPD[BIT(PD_INDEX_BITS)] VISIBLE;
 extern pte_t   armUSGlobalPT[BIT(PT_INDEX_BITS)]   VISIBLE;
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
+#ifdef CONFIG_ARM_CORTEX_A8
+extern word_t tlbLockCount VISIBLE;
+#endif /* CONFIG_ARM_CORTEX_A8 */
```
lsf37, line 39: I'm tempted to define a `ARM_HAS_TLB_LOCK` instead, that currently is defined as `CONFIG_ARM_CORTEX_A8`. It is an accident that this is currently only implemented for Cortex A8, other CPUs do also have the feature, we just haven't used it there. If we ever want to change that, we don't want the `#ifdef`s to go out of sync.

Indanz: To me the whole TLB lock seems to be premature optimisation and should be removed altogether until there is proof that it gives system level improvement in performance, not only in seL4 microbenchmarks.

lsf37: I agree with small changes, i.e. I would not remove anything here. But the patch does introduce logic duplication where there wasn't any before. Unless there's an argument why I'm wrong about it being duplication, I think we should avoid that.

Indanz (approved):
Only further improvement (other than removing all this) would be defining `lockTLBEntry` to a no-op when `CONFIG_ARM_HAS_TLB_LOCK` is disabled, so that the empty assembly and function stubs can be removed. And perhaps `mark_as_advanced` this new config option so it doesn't clutter the config choices.

(Edit: But I'm fine if this version gets merged.)

lsf37:
> Only further improvement (other than removing all this) would be defining `lockTLBEntry` to a no-op when `CONFIG_ARM_HAS_TLB_LOCK` is disabled, so that the empty assembly and function stubs can be removed. 

We could do that. I'm fine with it either way.

> And perhaps `mark_as_advanced` this new config option so it doesn't clutter the config choices.

I wasn't thinking of it as an input config option, just as something computed (currently set only when Cortex A8 is set), and wanted to compute it only once instead of twice. But we could actually also make it a config option. Would be fine to leave that for some other time, though.

lsf37 (approved):
When the build test has passed for all platforms I think this is good to go now.

src/arch/arm/config.cmake
```diff
@@ -130,6 +130,13 @@ config_option(
     DEFAULT_DISABLED OFF
 )
 
+config_option(
+    KernelArmHasTlbLock ARM_HAS_TLB_LOCK "Enable TLB Lock"
```
kent-mcleod, line 134: This should be a config_set unless it's expected to be configured by a user:
```
if (KernelArmCortexA8)
 config_set(KernelArmHasTlbLock ARM_HAS_TLB_LOCK ON) 
else()
 config_set(KernelArmHasTlbLock ARM_HAS_TLB_LOCK OFF) 
endif()
```

kent-mcleod:
> But we could actually also make it a config option. Would be fine to leave that for some other time, though.

If it is a config option it'd belong with: KernelDebugDisableL2Cache, KernelDebugDisableL1ICache, KernelDebugDisableL1DCache, KernelDebugDisableBranchPrediction.

But this config is reporting if the target supports TLB lock which isn't a user choice and should be enabled based on the microarchitecture.

Indanz (approved):

lsf37:
Yes, this can go ahead.

PR #1383, opened 2025-02-03
statedata: clarify ksInterruptState assertions

midnightveil:
The compile_assert is the code is very confusing to read at first glance, so we rephrase it to not confuse people in future.

Add an extra assertion to bound 2^k < sizeof(irq table) < 2^(k+1)

Also just add a comment describing why intStateIRQNode and intStateIRQTable aren't prefixed with ks*.

### DIFF
--- a/include/plat/pc99/plat/machine.h
+++ b/include/plat/pc99/plat/machine.h
@@ -14,7 +14,7 @@
 /* interrupt vectors (corresponds to IDT entries) */
 
 #define IRQ_INT_OFFSET 0x20
-#define IRQ_CNODE_SLOT_BITS 8
+#define IRQ_CNODE_SLOT_BITS 7
 
 typedef enum _interrupt_t {
     int_invalid                 = -1,
@@ -73,4 +73,3 @@ typedef enum _platform_irq_t {
 /* The text mode framebuffer exists part way into the video ram region */
 #define BIOS_PADDR_VIDEO_RAM_TEXT_MODE_START 0x000B8000
 #define BIOS_PADDR_IVDEO_RAM_END 0x000C0000
-


lsf37 (approved):
:+1:

lsf37:
Interesting, the not-excessive assert fails on x64. Can you have a look if x64 is wasting memory or if it actually needs the larger size?

Indanz:
I'm going to be annoying and say that `intStateIRQNode` doesn't need to be power-of-two.

Yes, it's kind of a CNode, and yes, CNodes are assumed to be power-of-two, but no CNode operations are done on `intStateIRQNode` and it is only accessed by index. The correctness of that index is independent of `intStateIRQNode` being a power-of-two, so I don't see how the proofs can possibly rely on that.

lsf37:
> I'm going to be annoying and say that `intStateIRQNode` doesn't need to be power-of-two.

It absolutely does have to be.
 
> Yes, it's kind of a CNode, and yes, CNodes are assumed to be power-of-two, but no CNode operations are done on `intStateIRQNode` and it is only accessed by index. 

That is incorrect. The nodes become part of the CDT when they are populated and CDT operations are not accessed by index. In fact, they critically rely on powers of two and alignment, otherwise the comparison operations in determining object boundaries, overlap and parent/sibling relationships are incomplete. They would also be overflow prone.

> The correctness of that index is independent of `intStateIRQNode` being a power-of-two, so I don't see how the proofs can possibly rely on that.

That doesn't mean they don't. There are global base-level invariants deep in the guts of the proofs that all objects have sizes of powers of and are aligned to their size. They are one of the most basic things we prove and everything relies on it. Removing that would be person decades of work. Not to speak of the fact that every single object comparison would now need either an explicit memory overflow check or would need a proof that it can't overflow. The vast majority of the hundreds (maybe thousands, if you count inline stuff) of proofs about word arithmetic in the kernel critically rely on alignment and would just be plainly not true without it.

lsf37:
> That might affect the proofs (one would hope) but I can't see the equivalent proof code for x86 as there is for arm/riscv.

Yes, it will affect them, hopefully only trivially, but after looking a bit closer it may be messy to change the size of the `irq` type. Will have to defer this to when I have a few hours to look at it in more detail.

kent-mcleod (approved):

PR #1384, opened 2025-02-03
arm: Save VCPU virt timer registers when disabled

kent-mcleod:
The virtual timer registers managed within a VCPU context are not saved if written to, if the VCPU in question is disabled. This results in any updates to the virtual timer registers being lost/written over when that VCPU is next scheduled and context switched in.

### DIFF
--- a/include/arch/arm/armv/armv7-a/armv/vcpu.h
+++ b/include/arch/arm/armv/armv7-a/armv/vcpu.h
@@ -848,6 +848,11 @@ static inline bool_t vcpu_reg_saved_when_disabled(word_t field)
 {
     switch (field) {
     case seL4_VCPUReg_SCTLR:
+    case seL4_VCPUReg_CNTV_CTL:
+    case seL4_VCPUReg_CNTV_CVALhigh:
+    case seL4_VCPUReg_CNTV_CVALlow:
+    case seL4_VCPUReg_CNTVOFFhigh:
+    case seL4_VCPUReg_CNTVOFFlow:
         return true;
     default:
         return false;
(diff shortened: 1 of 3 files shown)


lsf37:
Cool, thank you. @ryybrr just found the same issue during the integrity proofs. I think the problem is the same on AArch32, so we should also adjust `vcpu_reg_saved_when_disabled` there.

Are you happy for me to add that to the commit or would you prefer to do it yourself?

lsf37:
We should probably also add a change log item for it, so people know something has been fixed there.

kent-mcleod:
> Are you happy for me to add that to the commit or would you prefer to do it yourself?

fine by me

Indanz (approved):

Indanz:
Yes, both 32-bit ARMv7 and ARMv8 have the same problem, they're missing:

```
seL4_VCPUReg_CNTV_CVALhigh
seL4_VCPUReg_CNTV_CVALlow
seL4_VCPUReg_CNTVOFFhigh 
seL4_VCPUReg_CNTVOFFlow
```

lsf37:
Ok, commit amended, rebased, and Changelog added.

lsf37:
This will need functional correctness re-verification before we can merge it.

lsf37 (approved):

include/arch/arm/armv/armv7-a/armv/vcpu.h
```diff
@@ -848,6 +848,11 @@ static inline bool_t vcpu_reg_saved_when_disabled(word_t field)
 {
     switch (field) {
     case seL4_VCPUReg_SCTLR:
+    case seL4_VCPUReg_CNTV_CTL:
+    case seL4_VCPUReg_CNTV_CVALhigh:
+    case seL4_VCPUReg_CNTV_CVALlow:
+    case seL4_VCPUReg_CNTVOFFhigh:
+    case seL4_VCPUReg_CNTVOFFlow:
```
kent-mcleod, line 855: I think this list also needs `CNTKCTL`

kent-mcleod: Which is also missing from seL4_VCPUReg

lsf37: We're planning to add this one in a separate PR

PR #1385, opened 2025-02-04
Ensure the memory ordering of IPI arguments and the IPI pending flag for aarch64 using address dependency

yyshen:
The struct ipi_args is introduced to group related IPI remote call data into one struct. The ipi field in struct clh_qnode_p is also changed to void *. When IPI remote call is initiated, the ipi_args is initialised first, and then the address of ipi_args is assigned to the ipi field. A dmb ishst is used to ensure the order of writes. On the reader side, the address of ipi_args is read from the ipi field first, then the fields of the ipi_args are accessed through the pointer, creating address dependencies. Thus, the reads are order without using a memory barrier regarding to the pointer.

RISC-V and x86 are also updated, but the required memory barrier for RISC-V is not added yet.

### DIFF
--- a/src/smp/ipi.c
+++ b/src/smp/ipi.c
@@ -12,6 +12,28 @@
 #include <smp/ipi.h>
 #include <smp/lock.h>
 
+static struct ipi_args ipi_args = { 0 };
+
+static inline void init_ipi_args(IpiRemoteCall_t func, word_t data0,
+                                 word_t data1, word_t data2, word_t mask)
+{
+    ipi_args.remoteCall = (word_t)func;
+    ipi_args.args[0] = data0;
+    ipi_args.args[1] = data1;
+    ipi_args.args[2] = data2;
+
+    /* Get number of cores involved in this IPI */
+    ipi_args.totalCoreBarrier = popcountl(mask);
+
+#ifdef CONFIG_ARCH_AARCH64
+    /* This is to make sure if a write to ebig_kernel_lock.node_owners[target_core].ipi
+     * is observed, the above writes are also observed by readers according the
+     * address dependancy.
+     */
+    asm volatile("dmb ishst" ::: "memory");
+#endif
+}
+
 /* This function switches the core it is called on to the idle thread,
  * in order to avoid IPI storms. If the core is waiting on the lock, the actual
  * switch will not occur until the core attempts to obtain the lock, at which
@@ -22,6 +44,9 @@
 void ipiStallCoreCallback(bool_t irqPath)
 {
     if (clh_is_self_in_queue() && !irqPath) {
+        cpu_id_t core = getCurrentCPUIndex();
+        struct ipi_args *args = (struct ipi_args *)big_kernel_lock.node_owners[core].ipi;
+
         /* The current thread is running as we would replace this thread with an idle thread
          *
          * The instruction should be re-executed if we are in kernel to handle syscalls.
@@ -41,23 +66,23 @@ void ipiStallCoreCallback(bool_t irqPath)
         NODE_STATE(ksSchedulerAction) = SchedulerAction_ResumeCurrentThread;
 
         /* Let the cpu requesting this IPI to continue while we waiting on lock */
-        big_kernel_lock.node_owners[getCurrentCPUIndex()].ipi = 0;
+        big_kernel_lock.node_owners[core].ipi = NULL;
 #ifdef CONFIG_ARCH_RISCV
         ipi_clear_irq(irq_remote_call_ipi);
 #endif
-        ipi_wait(totalCoreBarrier);
-
+        ipi_wait(args->totalCoreBarrier);
         /* Continue waiting on lock */
-        while (big_kernel_lock.node_owners[getCurrentCPUIndex()].next->value != CLHState_Granted) {
-            if (clh_is_ipi_pending(getCurrentCPUIndex())) {
-
+        while (big_kernel_lock.node_owners[core].next->value != CLHState_Granted) {
+            if (clh_is_ipi_pending(core)) {
+                args = (struct ipi_args *)big_kernel_lock.node_owners[core].ipi;
+                IpiRemoteCall_t remoteCall = args->remoteCall;
+                (void)remoteCall;
                 /* Multiple calls for similar reason could result in stack overflow */
                 assert((IpiRemoteCall_t)remoteCall != IpiRemoteCall_Stall);
-                handleIPI(CORE_IRQ_TO_IRQT(getCurrentCPUIndex(), irq_remote_call_ipi), irqPath);
+                handleIPI(CORE_IRQ_TO_IRQT(core, irq_remote_call_ipi), irqPath);
             }
             arch_pause();
         }
-
         /* make sure no resource access passes from this point */
         asm volatile("" ::: "memory");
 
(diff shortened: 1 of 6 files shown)


yyshen:
> Thanks for sharing this, I'm mulling over the changes.
> 
> Does this solves the Clang SMP problems that we've been seeing for you?

I have not used clang to compile the kernel, so I am not sure what the clang SMP issue is. Could you identify which seL4test case?

Indanz:
> > Does this solves the Clang SMP problems that we've been seeing for you?
> 
> I have not used clang to compile the kernel, so I am not sure what the clang SMP issue is. Could you identify which seL4test case?

Mostly the MULTICORE 4 test. I thought you said at the summit that you were using clang for some reason, oh well.

Indanz:
Strange failures in CI for this one.

For [SABRE_debug_SMP_clang_32](https://github.com/seL4/seL4/actions/runs/13178985984/job/36787640507#step:4:3956):
```
  DIDRv: 3, armv 70, coproc baseline only? no.
  CPU is in secure mode. Enabling debugging in secure user mode.
  Bootstrapping kernel
    [10000000..2ff00000]
  reserved virt address space regions: 4
    [e0000000..e0050000]
    [e0050000..e005a2f1]
    [e005b000..e0491000]
    [ff000000..ff100000]
  DDDIIIDDDRRRvvv:::   333,,,   aaarrrmmmvvv   777000,,,   cccoooppprrroooccc   bbbaaassseeellliiinnneee   ooonnnlllyyy???   nnnooo...
  
  
  
  
  CCCPPPUUU   iiisss   iiinnn   ssseeecccuuurrreee   mmmooodddeee...   EEEnnnaaabbbllliiinnnggg   dddeeebbbuuuggggggiiinnnggg   iiinnn   ssseeecccuuurrreee   uuussseeerrr   mmmooodddeee...
  
  
  
  
  
  
  KERNEL PREFETCH ABORT!
  Faulting instruction: 0x0
  IFSR: 0x5
  halting...
```
Considering all SMP configs fail and all others pass, this one seems a real issue.

The TX2_verification_hyp_MCS_gcc_64 seems a fluke. Restarting all tests, let's see what happens.

Indanz:
It failed again the same way. SABRE is the only ARMv7 platform with SMP enabled.

yyshen:
> > > Does this solves the Clang SMP problems that we've been seeing for you?
> > 
> > 
> > I have not used clang to compile the kernel, so I am not sure what the clang SMP issue is. Could you identify which seL4test case?
> 
> Mostly the MULTICORE 4 test. I thought you said at the summit that you were using clang for some reason, oh well.

This probably will help on the multicore4 which is marked with flaky, if the issue is that the test hangs there.

yyshen:
> As far as I can tell there are two problems you try to work around:
> 
> 1. There is no fence or atomic operation with memory ordering enforcement between setting up an IPI and setting `ipi = 1`, so both the compiler and the CPU are free to re-order as they like. `clh_is_ipi_pending` checks `ipi`, but `generic_ipi_send_mask` sets `ipi = 1` before doing `IPI_MEM_BARRIER`. That is too late, another core can observe the `ipi` change before `IPI_MEM_BARRIER` is done.
> 
> You are adding `dmb ishst` and an address dependency to avoid re-ordering, why are both needed?
> 
> 2. The loop checking `value` in `ipiStallCoreCallback` has no fence or atomic operation with memory ordering enforcement.
> 
> Is there anything I'm missing?

You got it right. The address dependency is on the reader side, you could replace the dmb ishst with a store-release on the write to ipi and the address dependency with a load-acquire.

It looks like the compiler optimised the read of value (if my read of assembly is not mistaken). The use of volatile avoids this, and if you replace it to atomic with memory barrier, it also works.

There are different ways of handling the issue. I just present one way to do it ;-)

yyshen:
> Strange failures in CI for this one.
> 
> For [SABRE_debug_SMP_clang_32](https://github.com/seL4/seL4/actions/runs/13178985984/job/36787640507#step:4:3956):
> 
> ```
>   DIDRv: 3, armv 70, coproc baseline only? no.
>   CPU is in secure mode. Enabling debugging in secure user mode.
>   Bootstrapping kernel
>   available phys memory regions: 1
>     [10000000..2ff00000]
>   reserved virt address space regions: 4
>     [e0000000..e0050000]
>     [e0050000..e005a2f1]
>     [e005b000..e0491000]
>     [ff000000..ff100000]
>   DDDIIIDDDRRRvvv:::   333,,,   aaarrrmmmvvv   777000,,,   cccoooppprrroooccc   bbbaaassseeellliiinnneee   ooonnnlllyyy???   nnnooo...
>   
>   
>   
>   
>   CCCPPPUUU   iiisss   iiinnn   ssseeecccuuurrreee   mmmooodddeee...   EEEnnnaaabbbllliiinnnggg   dddeeebbbuuuggggggiiinnnggg   iiinnn   ssseeecccuuurrreee   uuussseeerrr   mmmooodddeee...
>   
>   
>   
>   
>   
>   
>   KERNEL PREFETCH ABORT!
>   Faulting instruction: 0x0
>   IFSR: 0x5
>   halting...
>   Kernel entry via Interrupt, irq 0
> ```
> 
> Considering all SABRE SMP configs fail and all others pass, this one seems a real issue.
> 
> The TX2_verification_hyp_MCS_gcc_64 seems a fluke. Restarting all tests, let's see what happens.

Let me change CONFIG_ARCH_AARCH64 to CONFIG_ARCH_ARM and see if it helps.

src/smp/ipi.c
```diff
@@ -12,6 +12,28 @@
...
+
+    /* Get number of cores involved in this IPI */
+    ipi_args.totalCoreBarrier = popcountl(mask);
+
+#ifdef CONFIG_ARCH_ARM
+    /* This is to make sure if a write to ebig_kernel_lock.node_owners[target_core].ipi
+     * is observed, the above writes are also observed by readers according the
+     * address dependancy.
+     */
+    asm volatile("dmb ishst" ::: "memory");
```
Indanz, line 33: You could use `IPI_MEM_BARRIER` here. That's a dsb instead of a dmb, so perhaps stronger (and slightly slower) than needed.

Indanz:
I added an atomics based solution to #871, could you check if my lock branch passes your internal tests too?

kent-mcleod:
@Indanz, @yyshen, is this PR aiming to address an overlapping set of issues as https://github.com/seL4/seL4/pull/871?  Any objections to rebasing this one onto the result of merging the locking refactoring in #871?

Indanz:
> @Indanz, @yyshen, is this PR aiming to address an overlapping set of issues as #871? Any objections to rebasing this one onto the result of merging the locking refactoring in #871?

Yes, they overlap and address the same issue. I'll close this one as it #871 is more complete, after I nicked @yyshen's IPI fix with commit 99b1d806e81ff03c39e2ee847a1026de05096801. The main difference between this PR and mine is that I'm using explicit atomic operations while yyshen is trying to force the compiler to do the right thing.

kent-mcleod:
> The main difference between this PR and mine is that I'm using explicit atomic operations while yyshen is trying to force the compiler to do the right thing.

I didn't look through this in depth, but I'd assumed the memory order dependency was also for forcing CPU ordering instead of compiler ordering because of the Arm register dependency memory ordering rules. So maybe this change also addresses a separate problem with accessing the IPI RPC data that is out of #871's scope.

Indanz:
Forcing both CPU and compiler ordering is better done by explicitly telling the compiler and CPU what you want, instead of battling both compiler and CPU optimisations. Atomics with barrier properties address both.

Another problem of relying on address dependencies for cross-core serialisation is that it is a subtle and very Arm specific property, which probably doesn't work for RISCV.

PR #1386, opened 2025-02-05
document&rename breakpoint_stack (arm32), remove it (riscv/arm64)

midnightveil:
For ARM:

Essentially, the `sp' register is banked between different PMODEs.

https://developer.arm.com/documentation/den0013/d/ARM-Processor-Modes-and-Registers/Registers

We initialise a stack that we will begin on when handling an exception.

For RISC-V:
Presumably this was copied from the ARM linker script without knowing why it was there, as no RISC-V code references it.

### DIFF
--- a/src/arch/arm/32/head.S
+++ b/src/arch/arm/32/head.S
@@ -153,9 +153,12 @@ BEGIN_FUNC(_start)
 
     /* Hyp kernel always run in Hyp mode. */
 #ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* Initialise ABORT stack pointer */
+    /* Initialise ABORT stack pointer.
+     * For ARM, the `sp' register is banked between PMODEs, so we configure
+     * a stack for when we take a exception (see traps.S, where we then switch
+     * modes and load the real kernel stack). */
     cps #PMODE_ABORT
-    ldr sp, =_breakpoint_stack_top
+    ldr sp, =_abort_stack_top
     cps #PMODE_SUPERVISOR
 #endif
 
--- a/src/arch/arm/common_arm.lds
+++ b/src/arch/arm/common_arm.lds
@@ -64,10 +64,12 @@ SECTIONS
         *(.bss)
         *(COMMON) /* fallback in case '-fno-common' is not used */
 
-        /* 4k breakpoint stack */
-        _breakpoint_stack_bottom = .;
+#if defined(CONFIG_ARCH_AARCH32) && !defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
+        /* 4k abort stack, needed for PMODE_ABORT on ARM32 */
+        _abort_stack_bottom = .;
         . = . + 4K;
-        _breakpoint_stack_top = .;
+        _abort_stack_top = .;
+#endif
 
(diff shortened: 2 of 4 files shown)


src/arch/arm/common_arm.lds
```diff
@@ -64,7 +64,7 @@ SECTIONS
         *(.bss)
         *(COMMON) /* fallback in case '-fno-common' is not used */
 
-        /* 4k breakpoint stack */
+        /* 4k breakpoint stack, needed for PMODE_ABORT on ARM */
```
axel-h, line 67: Seems it's used on AARCH32 only, not on AARCH64. Might be worth documenting at least, not sure if there is a simple way to remove it on AARCH64 also.

src/arch/arm/common_arm.lds
```diff
@@ -64,10 +64,12 @@ SECTIONS
         *(.bss)
         *(COMMON) /* fallback in case '-fno-common' is not used */
 
-        /* 4k breakpoint stack */
+#if defined(CONFIG_ARCH_AARCH32) && !defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
+        /* 4k breakpoint stack, needed for PMODE_ABORT on ARM32 */
         _breakpoint_stack_bottom = .;
```
axel-h, line 69: In order to improve the "documentaton" even more, the names could also change from `_breakpoint_stack_...` to "_abort_mode_stack_..." then, because that is what this really is. Unless the proof use the name, so changing it is not trivial. Breakpoints just happen to trap into abort mode.

axel-h:
Thanks, looks good to me. 

It might be worth adding a line in `CHANGES.md` now, because on AARCH64 and RISC-V this reduces the `.bss` segment by 4 KiByte now. So the kernel footprint gets a bit smaller o the architectures.

axel-h (approved):

Indanz (approved):

src/arch/arm/32/head.S
```diff
@@ -153,9 +153,12 @@ BEGIN_FUNC(_start)
...
     /* Hyp kernel always run in Hyp mode. */
 #ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* Initialise ABORT stack pointer */
     cps #PMODE_ABORT
-    ldr sp, =_breakpoint_stack_top
+    /* Initialise ABORT stack pointer.
+     * For ARM, the `sp' register is banked between PMODEs, so we configure
+     * a stack for when we take a exception (see traps.S). We then load the
+     * real kernel stack. Cross reference to `arm_data_abort_exception'. */
+    ldr sp, =_abort_stack_top
```
Indanz, line 161: I'd prefer the comment to go above the `csp #PMODE_ABORT`. Should be "an exception".

Indanz:
I don't think a mention in CHANGES.md is worth it, the total kernel sizes changes all the time. If it would change the size by one megabyte or something it would be worth mentioning, but that's not happening here.

Indanz:
Gitlint is complaining about the URL pushing the line past the 72 characters, which is unavoidable if the URL is too long. @lsf37 can you force the merge?

kent-mcleod (approved):

PR #1387, opened 2025-02-05
bkl/ipi: Skip handleIPI in sel4_atomic_exchange

yyshen:
When the sel4_atomic_exchange is called, the caller is not linked to the end of CLH lock yet. However, if the IPI is a remote stall call and irqPath is false, the caller will try to acquire the lock in ipiStallCoreCallback. Since the caller's node_owners[core].next does not have the correct value, the while loop in the function ipiStallCoreCallback may never terminate.

### DIFF
--- a/include/smp/lock.h
+++ b/include/smp/lock.h
@@ -71,13 +71,6 @@ static inline void *sel4_atomic_exchange(void *ptr, bool_t
     while (!try_arch_atomic_exchange_rlx(&big_kernel_lock.head,
                                          (void *) big_kernel_lock.node_owners[cpu].node,
                                          (void **) &prev)) {
-        if (clh_is_ipi_pending(cpu)) {
-            /* we only handle irq_remote_call_ipi here as other type of IPIs
-             * are async and could be delayed. 'handleIPI' may not return
-             * based on value of the 'irqPath'. */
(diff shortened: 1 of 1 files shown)


Indanz:
Agreed with your analysis.

This practically reverts commit 3d2ae69f9cb184c34b6b56d1365e57693e832db4 without fixing the underlying problem that inspired that commit, which is unbound CAS delays in pathological cases.

I think my #871 addresses this problem better.

lsf37:
The hardware tests succeeded (the two failures are UNSW networking flakiness, will restart).

Indanz:
On the second run ODROID_C4 fails with the dreaded [MULTICORE0004](https://github.com/seL4/seL4/actions/runs/13177358170/job/36782164532?pr=1387#step:4:20235) hang, so just like my PR didn't fix that one, this one doesn't either. I have more hope that #1385 fixes that, as that one is all about stopping CLang from optimising too much.

Indanz:
> I think we plan to merge #871, which removes sel4_atomic_exchange.

Yes, and that PR is older too. @yyshen only opened these PRs because I asked him to share his locking fixes at the summit. The only thing missing from #871 was the IPI fix in his other PR, the rest was already done one way or the other in #871.

Issue #1197, opened 2024-02-14
Generated platform descriptions (.yaml and .json) do not describe final device regions


Ivan-Velickovic:
Files generated from the kernel build system, such as `platform_gen.json` and `platform_gen.yaml` currently list the device regions used by the kernel, e.g for the ZCU102:

```yaml
devices:
- end: 0xf9010000
  start: 0x80000000
- end: 0xf9020000
  start: 0xf9011000
- end: 0xf9040000
  start: 0xf9021000
- end: 0xfd800000
  start: 0xf9041000
- end: 0x800000000
  start: 0xfd801000
- end: 0x10000000000
  start: 0x880000000
memory:
- end: 0x80000000
  start: 0x0
- end: 0x880000000
  start: 0x800000000
```

The problem that I am finding is that these device regions are not necessarily used by the kernel as some of them depend on the configuration of the kernel. E.g the GIC VCPU control region is listed, but that is only used (and hence reserved) by the kernel in hypervisor mode.

The C header export works by outputting an `#ifdef` macro based on a hardware description file, e.g https://github.com/seL4/seL4/blob/a8eebabf58990c00706c0830fe8e3edca2eaf7e4/tools/hardware.yml#L32

Not sure what the best way around it, would appreciate comments before I make a PR with a particular approach.

This currently blocks https://github.com/seL4/microkit/issues/35 as Microkit needs a way of knowing the *exact* reserved kernel regions as build-time.

Ivan-Velickovic:
The only use case of these files is in https://github.com/seL4/seL4_tools/blob/master/cmake-tool/helpers/shoehorn.py but it only looks at the `memory` field and so that's probably why this has not come up before.

axel-h:
Can't  we just add another field `condition` here that holds an array with the conditions?

kent-mcleod:
I ran into this in an experiment with generating the initial untyped list in the hardware_gen.py list.
My solution was to pass in the config options to the invocation of the script:
```
diff --git a/config.cmake b/config.cmake
index 53deca991..63c9dc6bc 100644
--- a/config.cmake
+++ b/config.cmake
@@ -571,6 +571,11 @@ if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
                 "${config_schema}" --yaml --yaml-out "${platform_yaml}" --sel4arch
                 "${KernelSel4Arch}" --addrspace-max "${KernelPaddrUserTop}" --json --json-out
[...]

PR #1390, opened 2025-02-06
platform_gen: only generate regions that are actually used

Ivan-Velickovic:
Right now the `platform_gen.json` and `platform_gen.yaml` files contain regions that are not actually used by the kernel. This is fine for `platform_gen.h` because they are guarded by #ifdefs with kernel config defines but the same approach does not work for YAML and JSON.

So, this patch changes the Python scripts that generate this files so that they only generate the regions that actually end up being used.

Based on @kent-mcleod's patch from https://github.com/seL4/seL4/issues/1197#issuecomment-1943692023.

Closes https://github.com/seL4/seL4/issues/1197.

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -104,133 +104,6 @@ if(NOT "${KernelCustomDTS}" STREQUAL "")
     message(STATUS "Using custom ${KernelCustomDTS} device tree, ignoring default dts and overlays")
 endif()
 
-if(DEFINED KernelDTSList AND (NOT "${KernelDTSList}" STREQUAL ""))
-    set(KernelDTSIntermediate "${CMAKE_CURRENT_BINARY_DIR}/kernel.dts")
-    set(
-        KernelDTBPath "${CMAKE_CURRENT_BINARY_DIR}/kernel.dtb"
-        CACHE INTERNAL "Location of kernel DTB file"
-    )
-    set(compatibility_outfile "${CMAKE_CURRENT_BINARY_DIR}/kernel_compat.txt")
-    set(device_dest "${CMAKE_CURRENT_BINARY_DIR}/gen_headers/plat/machine/devices_gen.h")
-    set(
-        platform_yaml "${CMAKE_CURRENT_BINARY_DIR}/gen_headers/plat/machine/platform_gen.yaml"
-        CACHE INTERNAL "Location of platform YAML description"
-    )
-    set(
-        platform_json "${CMAKE_CURRENT_BINARY_DIR}/gen_headers/plat/machine/platform_gen.json"
-        CACHE INTERNAL "Location of platform JSON description"
-    )
-    set(config_file "${CMAKE_CURRENT_SOURCE_DIR}/tools/hardware.yml")
-    set(config_schema "${CMAKE_CURRENT_SOURCE_DIR}/tools/hardware_schema.yml")
-    set(
-        KernelCustomDTSOverlay ""
-        CACHE
-            STRING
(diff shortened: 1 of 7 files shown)


lsf37:
Haven't looked through the details yet, but this changes some of the constants in the preprocess output verification sees, and at least to me it doesn't look like those changes are intended. Am I wrong?

Ivan-Velickovic:
> Haven't looked through the details yet, but this changes some of the constants in the preprocess output verification sees, and at least to me it doesn't look like those changes are intended. Am I wrong?

I think that's because the offsets are calculated based on the total number of kernel devices rather than the what is actually used. I can drop the changes to `c_header.py` and then there'll be no diff to the kernel preprocess.

Indanz:
I support this change, but don't think my CMake is good enough to do a proper review of those changes, but I will try. One commit has two almost identical Signed-off-by.

config.cmake
```diff
@@ -599,3 +473,135 @@ config_option(
...
+                --compat-strings-out "${compatibility_outfile}" --c-header --header-out
+                "${device_dest}" --hardware-config "${config_file}" --hardware-schema
+                "${config_schema}" --yaml --yaml-out "${platform_yaml}" --sel4arch
+                "${KernelSel4Arch}" --addrspace-max "${KernelPaddrUserTop}" --json --json-out
+                "${platform_json}"
+                --kernel-config-flags
+                    "CONFIG_PRINTING=${KernelPrinting}"
+                    "CONFIG_ARM_HYPERVISOR_SUPPORT=${KernelArmHypervisorSupport}"
+                    "CONFIG_ARM_SMMU=${KernelArmSMMU}"
+                    "CONFIG_TK1_SMMU=${KernelTk1SMMU}"
```
Indanz, line 589: This is the only change compared to the original code, except for moving this whole block down. That's a relieve,  I thought you rewrote this whole chunk.

Indanz (approved):

config.cmake
Ivan-Velickovic: Yep, I thought it had to be moved down to after the declarations of those config options but it looks like it doesn't need to be so I've left all the CMake stuff where it was originally.

kent-mcleod: I recall there was a dependency for why it needed to be moved down. I'm just taking a look now.

lsf37 (approved):
Looks good from my side. The style check still needs to be fixed, and it looks like @kent-mcleod wants to have a look before we merge.

config.cmake
kent-mcleod: `KernelPrinting` hasn't been defined at this point in the evaluation of CMake as it's defined lower in the config.cmake file. Which would have been why it should be moved down below the rest of the config declarations I recall.

kent-mcleod (approved):

lsf37:
The preprocess diff is back, and it seems to include semantic changes (e.g. `VPPIEventIRQ` padding).

config.cmake
```diff
@@ -599,3 +472,135 @@ config_option(
 )
 
 add_config_library(kernel "${configure_string}")
```
kent-mcleod, line 474: This line has to be at the bottom because there's configs defined in `include(src/drivers/config.cmake)`.

Ivan-Velickovic:
It looks like the CMake changes maybe broke things based on the pre-process check, the the code diff is as if `KernelArmVtimerUpdateVOffset` is no longer set.

kent-mcleod (approved):

config.cmake
Ivan-Velickovic: Thanks.

PR #1391, opened 2025-02-06
state: try to const-ify ksNumCPUs and ksCurDomain

midnightveil:
This means that they do not need to be allocated storage, and the compiler can do constant folding as necessary (instead of the manual constant-folding previously done for ksCurDomain). This removes the exported symbols from the BSS segment, as they are no longer shared.

### DIFF
--- a/include/model/statedata.h
+++ b/include/model/statedata.h
@@ -92,7 +92,11 @@ NODE_STATE_DECLARE(timestamp_t, benchmark_kernel_number_schedules);
 
 NODE_STATE_END(nodeState);
 
+#ifdef ENABLE_SMP_SUPPORT
 extern word_t ksNumCPUs;
+#else
+const word_t ksNumCPUs = 1;
+#endif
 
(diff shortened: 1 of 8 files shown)


lsf37:
This looks painful to reverify, I don't think I have time to do that one for free.

For the change in `statedata.h`, verification would now see different code depending on whether NUM_DOMAINS is 0 or not. This will break proof independence from NUM_DOMAINS.

Same for the conditional appearance of `ksNumCPUs` assignments. While we currently are not doing SMP, we should converge towards fewer code changes based on configs, not more.

@Xaphiosis do you remember if you introduced the `dom` variable in the fast path for verification and NUM_DOMAINS genericity, or was this already there before?

lsf37:
(the proof test will not show if independence from NUM_DOMAINS changes still works, they only test for one value, because it's too expensive to run that many VMs for pull requests)

lsf37:
That looks like it could work, yes.

PR #1399, opened 2025-02-14
arm: guarantee write to GIC ICENABLERn is observed

### LINKED ISSUE #612: Not so spurious IRQs
There is some code in the `handleInterruptEntry` that deals with spurious interrupts:

```
    if (IRQT_TO_IRQ(irq) != IRQT_TO_IRQ(irqInvalid)) {
        handleInterrupt(irq);
        Arch_finaliseInterrupt();
    } else {
#ifdef CONFIG_IRQ_REPORTING
        userError("Spurious interrupt!");
#endif
        handleSpuriousIRQ();
    }
```

I'm not convinced that this is a user error, and that it is in fact possible for the kernel to generate spurious interrupt all by itself.

I don't know if the fact that the kernel is generating spurious should be considered a bug, or just a thing that can happen (on ARM at least).

The scenario occurs when:

1/ While the kernel is operating the current deadline expires (which will assert an interrupt to the GIC, and then to the IRQ line, but since the kernel is running, we aren't taking exception).
2/ The kernel adjusts the current deadline.
3/ The kernel finishes what it is doing and returns to user land.
4/ The kernel traps again.
5/ The kernel gets current interrupt, but the timer interrupt is no longer asserted (since the deadline is now in the future per #2).
6/ Kernel reports spurious interrupt.

I'm not 100% sure what is going on here. But I think some combination of:

A/ Possible the timer logic does not immediately deassert the level interrupt to the GIC; so it is asserted long enough for IRQ to trap, but is deasserted by the time we read the actual current interrupt.

B/ There is some delay in deasserting the IRQ line to the CPU from the GIC.

I think either A or B is feasible. In both cases you end up with a 'spurious' interrupt that is 'expected'  and definitely not a user error.

If either A or B is true I don't think there is anyway of avoiding a spurious error, and it is just something that is going to happen from time to time. Given this it should not be raised as a user error as is currently the case I think.

Of course, maybe I'm missing something, and this truly is an error, but if so I can't work it out.

Note: I've only ever seen this on MCS, which does the deadline adjustment.

midnightveil:
This is necessary, (even if not sufficient) to fix #612. At the moment we also aren't executing a DSB instruction to ensure completion of writes (see §12.1.6 of [1]). DSB instructions to *appear* to completely resolve #612, but more investigation is in progress to see if this is correct.

> Completion of a write to this register does not guarantee that the
> effects of the write are visible throughout the affinity hierarchy. To
> ensure an enable has been cleared, software must write to the register
> with bits set to 1 to clear the required enables. Software must then
> poll `GICD_CTLR.RWP`` until it has the value zero.
> --- §12.9.7, GICD_ICENABLER<n>, Interrupt Clear-Enable Registers. [1]

[1]: ARM Generic Interrupt Controller Architectural Specification (GICv3/v4)
     Issue G. https://www.scs.stanford.edu/~zyedidia/docs/arm/gic_v3.pdf

cc @Ivan-Velickovic

### DIFF
--- a/include/arch/arm/arch/machine/gic_v3.h
+++ b/include/arch/arm/arch/machine/gic_v3.h
@@ -233,6 +233,8 @@ extern volatile struct gic_dist_map *const gic_dist;
 extern volatile struct gic_rdist_map *gic_rdist_map[CONFIG_MAX_NUM_NODES];
 extern volatile struct gic_rdist_sgi_ppi_map *gic_rdist_sgi_ppi_map[CONFIG_MAX_NUM_NODES];
 
+void gicv3_dist_wait_for_rwp(void);
+
 /* Helpers */
 static inline void gic_enable_clr(word_t irq)
 {
@@ -243,6 +245,15 @@ static inline void gic_enable_clr(word_t irq)
         gic_rdist_sgi_ppi_map[CURRENT_CPU_INDEX()]->icenabler0 = BIT(bit);
     } else {
         gic_dist->icenablern[word] = BIT(bit);
+
+        /*
+         * GICD_ICENABLER<n> is special; a completion of a write to this
+         * register does not guarantee that the GIC has observed this write.
+         * The Arm GIC Specification, §12.9.7, GICD_ICENABLER<n> states that
+         * software must poll GICD_CTLR.RWP until it is 0. This is what our
+         * wait_for_rwp() function does.
+         */
+        gicv3_dist_wait_for_rwp();
     }
 
 }
--- a/src/arch/arm/machine/gic_v3.c
+++ b/src/arch/arm/machine/gic_v3.c
@@ -111,7 +111,7 @@ static uint32_t gicv3_do_wait_for_rwp(volatile uint32_t *ctlr_addr)
     return ret;
 }
 
-static void gicv3_dist_wait_for_rwp(void)
+void gicv3_dist_wait_for_rwp(void)
 {
     gicv3_do_wait_for_rwp(&gic_dist->ctlr);
 }


Indanz (approved):
`gicv3_do_wait_for_rwp` has what is basically debug code in it to detect hangs, those should be removed (or at least be put behind a DEBUG ifdef), especially now that it's called for each IRQ ack.

It would be good to see IRQ benchmark numbers before and after this change, just in case. If the overhead is too high, it might be better to have occasional spurious interrupts than the overhead.

I think there are too many inline functions here, but that's another issue.

kent-mcleod:
Anyone against also adding the `dsb`? Is the risk without the `dsb` that the `gicv3_do_wait_for_rwp()` runs early and doesn't see `RWP` bit become set?

Indanz:
> Anyone against also adding the `dsb`? Is the risk without the `dsb` that the `gicv3_do_wait_for_rwp()` runs early and doesn't see `RWP` bit become set?

That shouldn't be necessary: Memory accesses for device memory are not re-ordered in regard to each other and happen in program order, as seL4 maps it with `DEVICE_nGnRnE` ([nR](https://developer.arm.com/documentation/107565/0101/Memory-system/Memory-types-and-attributes/Device-memory) stands for non-reordering).

DSB is only needed if there are cached memory reads or writes that need to be ordered in regard to the device memory access (and further instructions need to wait for all memory writes to complete, if you just want to memory barrier, use DMB). `gicv3_do_wait_for_rwp()` has none of that, it only needs to be ordered in relation to the previous GIC write.

>  I think it's necessary (or it seems to fix things) before or after writing to the EOIR but I haven't quite figured out the precise semantics of when it's needed.

EOIR is a system register on Aarch64, so it doesn't get the automatic program order guarantees like the memory mapped GIC code does.

It depends on the register what kind of barrier is needed, I haven't found any good overview with exceptions.

> ARM trusted firmware does seem to do it before most register writes - dsbisht specifically. And also also polls RWP in different spots.

System register writes can be what Arm calls context altering operations, so the general rule for those is that they need to be followed by an ISB. It depends on the register whether you can get away with anything less, and that's not clearly documented. Some registers don't need synchronisation, as the CPU takes care of it. Some read-only registers also need an ISB. See this [comment](https://github.com/seL4/seL4/pull/1318#discussion_r1778327665) from me with some more info about system registers.

> DSB seem to fix spurious interrupts without this PR at all, though. (so it's hard to tell what's correct experimentally). So DSB was for EOIR, not for this.

DSB halts further instruction execution until the previous memory accesses are finished, according to the flags (unlike DMB, which only prevents re-ordering of load and stores past the barrier).

Importantly, system register accesses via MSR/MRS do not count as memory accesses!

So if DSB works for EOIR, it's by chance because of the extra synchronisation delay added or because of CPU implementation details. To synchronise MRS/MSR you need to use ISB to be sure, at least according to all the official documentation I could find, including https://developer.arm.com/documentation/100941/0101/Barriers.

If ARM trusted firmware is only using DSB they either have better documentation than I can find, or it's buggy and an ISB is missing. DSB may be needed in addition to ISB though. From the quick search I did and the ATF code I checked seemed to do the right thing though.

kent-mcleod:
> That shouldn't be necessary: Memory accesses for device memory are not re-ordered in regard to each other and happen in program order, as seL4 maps it with `DEVICE_nGnRnE` ([nR](https://developer.arm.com/documentation/107565/0101/Memory-system/Memory-types-and-attributes/Device-memory) stands for non-reordering).

Yea you're right. I didn't realize that the non-reordering constraint was all access to the same peripheral, I thought it was just to the same memory address but other addresses of the same peripheral could be reordered.

So yes. It does seem that the DSB isn't necessary in the case where we're trying to read from the same peripheral interface because the read will be serialized after the write in the GICD interface.

kent-mcleod (approved):

Indanz:
> Yea you're right. I didn't realize that the non-reordering constraint was all access to the same peripheral, I thought it was just to the same memory address but other addresses of the same peripheral could be reordered.

To be clear, the non-reordering constraint guarantees that all accesses to any MMIO will not be re-ordered. It _does_ allow re-ordering with normal memory accesses and other instructions, which is why sometimes DSB/DMB are still needed.

> So yes. It does seem that the DSB isn't necessary in the case where we're trying to read from the same peripheral interface because the read will be serialized after the write in the GICD interface.

Not only that, in this case the read check itself will guarantee that the write has completed, making DBS for that reason unnecessary. 

That the polling read is necessary here is because there is extra hardware buffering, so the write completing on a bus level does not mean that the GIC has fully handled it yet.

> There's a few other relevant segments, specifically in regards to execution of DSB, and guarantees of interrupts being deasserted before EOIR; if we haven't guaranteed that the interrupt has been deasserted by the GIC before executing EOIR that can cause spurious interrupts; a DSB can do this.

You and that thread are talking about DSB _before_ doing a system register write. That often makes sense yes, especially for GICv3 where some things happen via system registers and other things via MMIO. But such DSB is required by the system register access, not by the MMIO access itself. Though you could put a DSB after each MMIO access instead of before the system register accesses, but it would require a good comment, because it would be confusing.

> Personally, this is the one part of driver code I'm the most iffy about, and it's also the one thing none of the proofs seem to care about :)

The problem is that the proofs can't help with things like this much: To be able to do that, they would need to know what the requirements are for each system register and MMIO access. But that's exactly the problem we're facing, so moving it to the proofs instead of the code just moves all wrong assumptions from one place to another.

Maybe I should give it another full review with the GIC doc in hand, checking it line by line for these issues specifically.

One thing that surprised me was that Arm gives so little guarantees about system register accesses, it looks like most of them can be re-order in regard to any other instructions. It would be nice if we had the guarantee that system register accesses are executed in program order in regard to each other at least. If you need to do an ISB after each, why use the system register interface instead of the MMIO interface for GICv3? It just complicates things and might even be slower.

kent-mcleod:
Regarding non-reordering: https://developer.arm.com/documentation/107565/0101/Memory-system/Memory-types-and-attributes/Device-memory
The 3rd example here states that separate regions are not guaranteed to not be reordered:
"Region C is marked as nR. It is not guaranteed whether the accesses to region A will occur in program order with respect to accesses to region C." (Region A I'd also marked as nR in the example)

Indanz:
> Regarding non-reordering: https://developer.arm.com/documentation/107565/0101/Memory-system/Memory-types-and-attributes/Device-memory The 3rd example here states that separate regions are not guaranteed to not be reordered: "Region C is marked as nR. It is not guaranteed whether the accesses to region A will occur in program order with respect to accesses to region C." (Region A I'd also marked as nR in the example)

Thanks for the link. Oh wow, I didn't expect it was that bad, though I had my suspicions. The link is for Armv8-M, would need to double check the same holds for Armv8-A.

Sadly the Armv8-A document is pdf only it seems. It says:

```
The non-Reordering attribute does not require any additional ordering, other than that
which applies to Normal memory, between:
[...]
- Accesses with the non-Reordering attribute and accesses to different peripherals of
  IMPLEMENTATION DEFINED size.
```
So it says the same as your link. I guess that the implementation defined size is 64kb on most devices, but I can't find any documentation for this anywhere.

Anyway, in this case `icenablern` is within 512 bytes of `ctrl`, so it should be fine. I don't think we require any ordering between GIC distributor and GIC redistributor modifications.

kent-mcleod:
That overhead is quite significant...

kent-mcleod:
Would #612 also be resolved by the change in https://github.com/seL4/seL4/pull/1183, which would result in no longer enabling/disabling the IRQ via the distributor?

Indanz:
> Would #612 also be resolved by the change in #1183, which would result in no longer enabling/disabling the IRQ via the distributor?

The test proofs that the spurious interrupts are caused by the missing synchronisation, so yes, split EOI mode should resolve them too. The split EOI only uses system register writes, so it doesn't have this performance issue (and it doesn't even need an ISB because it doesn't need to finish before exception return).

Split EOI mode makes a huge amount of sense because that's what seL4 is basically emulating with enabling/disabling interrupts.

Indanz:
I'm tempted to close this PR. Although it's functionally correct, the price is too high. The raw IRQ cost of the old IRQ handler + a spurious IRQ is awfully close (probably within 5%) to the costs of the fixed IRQ handler. So performance wise, it's like getting a spurious interrupt all the time. :-(

I'd rather remove the spurious printf now that we know what causes it, and as a long-term solution implement split EOI mode with a more refined IRQ abstraction for verification.

Does anyone disagree?

Indanz:
We probably should add a comment saying that strictly speaking gicv3_dist_wait_for_rwp is needed and why we don't do it anyway, with occasional spurious interrupts as a result.

lsf37:
Yes, please, to adding a comment. This was very informative and expensive to find out. Link to the discussion here would be fine, and it sounds like we should concentrate on getting split EOI working and merged.

Indanz:
> Personally, I somewhat disagree with just taking the overhead of spurious interrupts as opposed to making it correct.

Oh, I agree it feels wrong. But if the cost of a fixed IRQ handler is 800, and the cost of normal IRQ + spurious one takes 900 (less than the 500 because it does a lot less and everything is cached), while 99% of other IRQs take 500, it's really a no-brainer from a practical point of view.

> For one, they are possible when writing incorrect device drivers; it's not _always_ a kernel bug.

Are you sure? I think you'll get a "Received disabled IRQ" message then. I can't think of a way for user space to trigger this particular corner case.

It would be good if we had some generic counters which can be retrieved when debug and/or benchmarking is enabled, and a simple way of using them in temporary debug code. I think that would be more helpful than specific counters for these kind of spurious interrupts.

That said, people may not find it and then no one will use it, and this is a known issue, so good to keep an eye on it. But if we don't trigger the spurious interrupts in any tests, then it doesn't really fit in sel4bench. I suppose we could  add a test to sel4test, but timing sensitive issues like these are usually not deterministic, so that might not be practical either.

Feel free to re-open this PR to add the comment, that would keep everything in one place. Or keep it closed and open a new PR if you want to retain your patch here.

Indanz:
> It's also possible that moving the polling to just before EOIR instead of after ICENABLER might reduce the cost

It's already just before the EOIR. For signals, handleInterrupt sends the signal first, then it disables the IRQ and just after that, it always does the ack (EOIR write).

You could try moving the `maskInterrupt(true, irq);` to the beginning and keep the polling where it is, but that would add a GICv3 if-check. That way the damage is limited to perhaps one GIC read, but it will still be bad. Maybe if you do the first poll read early, but only check it later.

> I wrote a bad QEMU driver intentionally which takes several hundred cycles before deactivating the GIC to the IRQ following a register write to clear the interrupt (and then an sel4 IRQ ackl, which reliably reproduced this (just made a timer callback that deactivated the IRQ). Afaik there's no requirement on how long device peripherals can take to deassert interrupt lines to the GIC.

That's why drivers for slow hardware like that need a register read to serialise the device level IRQ ack. Even then IRQs could linger on long IRQ wires, that's one of the reasons all modern systems use MSI instead of out-of-band, async IRQ lines. It was a hell to deal with and got worse the more devices were connected to a line.

But I don't understand how your buggy QEMU driver could reproduce this: The spurious interrupt happens just after receiving a real one. At that point the device always still has the IRQ active, as user space had no chance to ACK it yet. This bug is purely a CPU-GIC race, the device should be irrelevant. That said, QEMU GICv3 handling seems somewhat dodgy and not close to real hardware in behaviour, so who knows.

Indanz:
Actually, I think it's a good idea to count how many interrupts we get where `notification_ptr_get_state(ntfnPtr) == NtfnState_Active && (notification_ptr_get_ntfnMsgIdentifier(ntfnPtr) & cap_notification_cap_get_capNtfnBadge(cap))`.

There is a chance that we also get double interrupts. We only get the spurious print if the GIC handles the disable after interrupting the CPU, but before the CPU reads the pending IRQ, and there are no other IRQs pending. Although unlikely, it's theoretically possible the disable is so slow that we receive and handle the same IRQ twice.

Indanz:
(And if without the fix we usually get two interrupts, the performance trade-off is much different.)

Indanz:
Ah, yes, you're right, that's a way to trigger the warning from user space. Chances are high that user space will also get interrupts while there is nothing new at device level, so the print is not crucial in figuring out what's happening, but it would help a bit.

As for my double interrupt concern: It would be measured by the IRQ user benchmark test, as it will measure both. The max value in your before is close to the average, so it's not happening (at least not in 100 runs).

Issue #1216, opened 2024-03-11
aarch32: No easy way for a thread to modify its TPIDRURO register.


kent-mcleod:
TPIDRURO is the default register used by the `-mtp=auto` or `-mtp=cp15` arguments to gcc or clang on aarch32 when targeting armv7 CPUs for both arm-none-eabi and arm-linux-gnueabi targets.  This register is read-only from user privilege and can only be written from supervisor mode. seL4 provides a way for setting this register with seL4_TCB_WriteRegisters, but this invocation cannot be called on itself by the calling thread. On ia32 and some x86_64 configurations seL4 provides a syscall, `seL4_SetTLSBase`, which updates a read-only register used for TLS, however this syscall isn't available on aarch32 configurations for writing the TPIDRURO register.

It would be useful for a thread to be able to set this register without requiring another thread. Currently, the sel4_libs runtimes  use a different register, TPIDRURW, which requires using `-mtp=soft` which causes the compiler to emit a function call to lookup the tp address. Future versions of gcc appear to add an option to select tpidrurw with -mtp=tpidrurw [1] but this doesn't appear to be supported by any stable releases yet.

[1]: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mtp-1

kent-mcleod:
A quick solution would be to add a version of `seL4_SetTLSBase` for aarch32 that can set `TPIDRURO`, but the drawback of this approach is that that syscall doesn't require any authority to be provided where access to the TCB cap should probably be required. 

A harder solution would be to address the drawbacks of seL4_TCB_WriteRegisters() relating to not being able to set individual registers and not being able to be called by the running thread on it's own TCB. 
It would be possible to add a mask variant of seL4_TCB_WriteRegisters and seL4_TCB_ReadRegisters [1] where each bit in the mask selects the associated register in `seL4_UserContext`.

Why can't seL4_TCB_WriteRegisters() be called on the current TCB? The background for this restriction is summarized here [2].  In summary it seems to be for verification convenience. (Resolving this would make it possible to take out `seL4_SetTLSBase` which prevents user level from implementing a policy supported by hardware - preventing access to the tp segment register (on x86).

[1] - Maybe even the current invocations should be replaced with a mask variant, and legacy functions added to libsel4 that just calls the new invocation with all bits of the mask set.
[2] - https://sel4.atlassian.net/browse/SELFOUR-22

kent-mcleod:
[...]

PR #1400, opened 2025-02-16
arm,tls: Add config option KernelArmTLSReg

kent-mcleod:
Make which user thread ID register used by the kernel for its TLS operations configurable. Either tpidru or tpidruro can be selected which map to architecture registers tpidru and tpidruro on aarch32 and tpidr_el0 and tpirrro_el0 on aarch64.

The default value of this option is tpidru to stay compatible with the kernel's current behaviour.

Addresses: #1216

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -37,6 +37,12 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
   `set(KernelArmDisableWFIWFETraps ON)`
   to your project settings to get the same configuration as before if you are using `tqma8xqp1gb`.
 
+#### Arm
+
+* Added config option for selecting which thread ID register is used for Kernel TLS syscalls and invocations.
+  KernelArmTLSReg can be used to select either `tpidru` or `tpidruro` as the TLS register used for `seL4_TCB_SetTLSBase` and `seL4_SetTLSBase` operations.
+  This config option's default value is `tpidru` which is what the register that the kernel currently uses for the TLS register for aarch32 and aarch64 platforms.
+
 ### Upgrade Notes
 
 ---
--- a/include/arch/arm/arch/32/mode/machine/registerset.h
+++ b/include/arch/arm/arch/32/mode/machine/registerset.h
@@ -100,9 +100,13 @@ enum _register {
     /* user readable/writable thread ID register.
      * name comes from the ARM manual */
     TPIDRURW = 18,
-    TLS_BASE = TPIDRURW,
     /* user readonly thread ID register. */
     TPIDRURO = 19,
+#ifdef CONFIG_ARM_TLS_REG_TPIDRU
+    TLS_BASE = TPIDRURW,
+#elif defined(CONFIG_ARM_TLS_REG_TPIDRURO)
+    TLS_BASE = TPIDRURO,
+#endif
     n_contextRegisters = 20,
 };
 
--- a/include/arch/arm/arch/64/mode/machine/registerset.h
+++ b/include/arch/arm/arch/64/mode/machine/registerset.h
@@ -143,9 +143,13 @@ enum _register {
     /* user readable/writable thread ID register.
      * name comes from the ARM manual */
     TPIDR_EL0                   = 35,
-    TLS_BASE                    = TPIDR_EL0,
     /* user readonly thread ID register. */
     TPIDRRO_EL0                 = 36,
+#ifdef CONFIG_ARM_TLS_REG_TPIDRU
+    TLS_BASE = TPIDR_EL0,
+#elif defined(CONFIG_ARM_TLS_REG_TPIDRURO)
+    TLS_BASE = TPIDRRO_EL0,
+#endif
     n_contextRegisters          = 37,
 };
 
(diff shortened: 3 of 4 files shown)


kent-mcleod:
Probably we should discuss this on the linked issue for now to see if we want this approach.

Indanz:
Implementation wise, I don't see the need for a config option. Why not just unconditionally enable KernelSetTLSBaseSelf and use it to modify TPIDRRO? User space can still use TPIDRURW as before and it will work as before.

The current code is weird in the sense that it always has the `seL4_SetTLSBase` syscall, but doesn't define the libsel4 API wrappers for Aarch32. So technically it's an API break, as people could still call `seL4_SetTLSBase` via assembly before and the behaviour has changed. But no one would do that, as it's not necessary to modify TPIDRURW.

kent-mcleod:
> The current code is weird in the sense that it always has the `seL4_SetTLSBase` syscall, but doesn't define the libsel4 API wrappers for Aarch32. So technically it's an API break, as people could still call `seL4_SetTLSBase` via assembly before and the behaviour has changed. But no one would do that, as it's not necessary to modify TPIDRURW.

The current code doesn't implement the seL4_SetTLSBase syscall unless CONFIG_SET_TLS_BASE_SELF is defined. So aarch32 has never had this call available because the config option isn't user-configurable and it currently is only set for ia32 and some configurations of x86_64.

kent-mcleod:
> (Edit: I'm thoroughly confused by `seL4_SetTLSBase` versus `seL4_TCB_SetTLSBase`, we have both??)

`seL4_SetTLSBase` was added as part of this RFC: https://github.com/seL4/rfcs/blob/main/src/implemented/0030-cross-platform-tls.md 

`seL4_TCB_SetTLSBase` was added earlier as a variation on `seL4_TCB_WriteRegisters`. The latter cannot be called on ksCurThread, but `seL4_TCB_SetTLSBase` can

kent-mcleod:
> Implementation wise, I don't see the need for a config option. Why not just unconditionally enable KernelSetTLSBaseSelf and use it to modify TPIDRRO? User space can still use TPIDRURW as before and it will work as before.

Well the kernel defines a virtual register `TLS_BASE` which is what `seL4_SetTLSBase` and  `seL4_TCB_SetTLSBase` are defined to write to. 

Since RFC-0030 landed the kernel doesn't need to define TLS_BASE anymore (but it does need to provide some way of updating user-mode read only registers) . I think at the time we thought that `seL4_SetTLSBase` would be only used for "legacy" systems because it was only needed for ia32 and earlier instances of x86_64 machines that don't have fsgsbase instructions.

If we removed the TLS_BASE concept (which may require a minor RFC), then we could do what you propose.
Separately, if there was a way to set rights onto a TCB itself, then it would allow a TCB to be granted the authority to set its own TLS base register without needing to provide a TCB capability and this would still enable the user level api's that can't take any authority arguments.

Indanz:
Okay, with the mess TLS is, there is no way around a config option, that seems the least bad solution. All alternatives break existing code in some way or other.

The seL4 default should match what all seL4 code we have control over expects, so care needs to be taken to update all repos at once.

kent-mcleod:
> The seL4 default should match what all seL4 code we have control over expects, so care needs to be taken to update all repos at once.

I was going to leave the default as the same behavior as now. Then our projects can switch the config to the alternate TLS register when they support it.

kent-mcleod:
I think this is ready for merge

Indanz:
On 2025-02-21 01:43, Kent McLeod wrote:
> I was going to leave the default as the same behavior as now. Then our
> projects can switch the config to the alternate TLS register when they
> support it.

I thought it would reduce our diff for musl libc if we dropped the 
non-standard support? So I had in mind to only support one register with 
the seL4 libs, and switch them all over at once.

kent-mcleod:
@lsf37 proofs seem to pass

Indanz:
> If we keep the default behavior, I was thinking of adding a test to the musl CMake file that checks the config value and throws an error if it's set to an unsupported value.

The whole point of the config option is to keep source/binary compatibility with existing code. If there is no plan to switch the default TLS register, there is no need for a config option: You can just add the `seL4_SetTLSBase` syscall that modifies tpidruro and keep everything else as it is.

kent-mcleod:
> You can just add the `seL4_SetTLSBase` syscall that modifies tpidruro and keep everything else as it is.

If seL4_SetTLSBase syscall always exists then it weakens the protection provided by the tpidruro register. Some runtime environments may want to use a read only register so that they can store a reference that won't be able to be overwritten.

Indanz:
> If seL4_SetTLSBase syscall always exists then it weakens the protection provided by the tpidruro register. Some runtime environments may want to use a read only register so that they can store a reference that won't be able to be overwritten.

Good point.

Indanz (approved):
If the default TLS register on Aarch64 is tpidr_el0, while the default one for Aarch32 is TPIDRURO, no matter what config you use, you will always get one non-default TLS for one of the archs.

Not a big deal, but something that people using this should be aware of, so perhaps mention what the usual registers for TLS are for Aarch32 and Aarch64 in the help text? Perhaps tricky to do without making things confusing.

lsf37 (approved):

kent-mcleod:
I updated the help message to explain on gnu-elf ABI's the default register  is tpidruro for aarch32 and tpidr for aarch64

Indanz:
> The seL4 default should match what all seL4 code we have control over expects

This didn't happen, so the default TLS register has changed from `tpidruro` to `tpidru` for Aarch32.

PR #1405, opened 2025-02-24
riscv: rename incorrect level2_pt -> level1_pt

midnightveil:
Since cdbae0b3b27c8a69f1b47074523177e428b0b95a, seL4 has used a 0-indexed naming for the levels of the page tables, starting from the root (0) and down to the lowest level (2), for CONFIG_PT_LEVELS=3.

However, the naming of the stored page table was never updated, leading to a variety of confusing code such as:

    kernel_image_level2_dev_pt[RISCV_GET_PT_INDEX(vaddr, 1)] = \
        pte_next(paddr, true);

When CONFIG_PT_LEVELS=4, the entries of the level 1 page table instead point to gigapages (1GiB) instead of megapages (2MiB). (Behaviour left unchanged).

### DIFF
--- a/include/arch/riscv/arch/model/statedata.h
+++ b/include/arch/riscv/arch/model/statedata.h
@@ -27,12 +27,11 @@ extern asid_pool_t *riscvKSASIDTable[BIT(asidHighBits)];
 /* Kernel Page Tables */
 extern pte_t kernel_root_pageTable[BIT(PT_INDEX_BITS)] VISIBLE;
 
-/* We need to introduce a level2 pagetable in order to map OpenSBI to a separate
- * page entry to avoid PMP exception. */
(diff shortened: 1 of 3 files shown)


lsf37 (approved):
This looks good from the verification side, and thanks for removing that confusion.

I've switched on the hw build to get some binaries, because theoretically the change should produce the same binary as before.

Indanz (approved):
Changes look correct and make things consistent.

(Not sure  I agree with the already merged commit cdbae0b, doesn't seem to improve the code much. Calling it level 0, 1 and 2 is extremely confusing when talking about levels in a tree structure, as level 1, 2 and 3 is what I'd expect. Classical start counting at 0 or 1 issue.)

PR #1409, opened 2025-02-26
boot: make clock sync test delta conditional

lsf37:
The clock sync test at boot keeps failing for the hifive board with time deltas as high as 5. We think this is because mmode might be taking a big lock for reading the time.

Add a macro for allowed delta and set it to a higher value for RISC-V.

Supersedes #1366. See also the comments/discussion there.

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -145,6 +145,16 @@ word_t arch_get_n_paging(v_region_t it_veg);
 
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
 BOOT_CODE void clock_sync_test(void);
+
+/* Delta (in us) allowed in the clock sync test in addition to getTimerPrecision().
+   On RISC-V, reading the clock goes via mmode. If mmode uses a big lock, that
+   can take considerable time */
+#ifdef CONFIG_ARCH_RISCV
+#define CLOCK_SYNC_DELTA 5
+#else
+#define CLOCK_SYNC_DELTA 1
+#endif
+
 #else
 #define clock_sync_test()
 #endif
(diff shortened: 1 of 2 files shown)


lsf37:
I'm not actually entirely clear if it's not better to just disable the clock sync test for RISC-V. I guess with this we will know it is at least close-ish and not wildly off.

lsf37:
Looks like I shouldn't declare this in the header, will rearrange slightly.

kent-mcleod:
> I'm not actually entirely clear if it's not better to just disable the clock sync test for RISC-V. I guess with this we will know it is at least close-ish and not wildly off.

Can you just disable for riscv

lsf37:
> > I'm not actually entirely clear if it's not better to just disable the clock sync test for RISC-V. I guess with this we will know it is at least close-ish and not wildly off.
> 
> Can you just disable for riscv

I can also do that.

Ivan-Velickovic:
I don't think we should disable it for all RISC-V platforms, I never see the error on the Star64 or HiFive P550 so it seems to be isolated to the HiFive Unleahsed at least in my experience.

lsf37:
> I don't think we should disable it for all RISC-V platforms, I never see the error on the Star64 or HiFive P550 so it seems to be isolated to the HiFive Unleahsed at least in my experience.

Then we should not merge the fix here either, because it increases the value for all RISC-V platforms.

We don't have either of these boards in CI yet -- are you sure that they really don't have the same problem occasionally? The check only runs for SMP+MCS+debug, so most configs wouldn't show it.

Currently hifive is blocking CI, so I'd rather switch it off now and then add an actual solution when the other boards are tested.

lsf37:
A disable PR is now here: https://github.com/seL4/seL4/pull/1411

Indanz (approved):

lsf37:
Ok, then how about I merge this one here (increased delta for RISC-V), and we make it more configurable (per board) when the other RISC-V boards are in CI and we can see the effects.

@Ivan-Velickovic does that work from your side?

Ivan-Velickovic:
Yes that makes sense.

PR #1410, opened 2025-02-26
Bump minimum CMake version

wom-bat:
Compatibility with versions <3.10 is going away.
As it happens, we're not using any CMake features that have changed between 3.7 and 3.16, so bump the lowest version to 3.16.

Also remove the minimum version statement from the platform config files --- they're all very simple files that are version independent; and the version is checked elsewhere anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
-cmake_minimum_required(VERSION 3.8.2)
+cmake_minimum_required(VERSION 3.16.0)
 include(CheckCCompilerFlag)
 
 include(${CMAKE_CURRENT_LIST_DIR}/configs/seL4Config.cmake)
--- a/config.cmake
+++ b/config.cmake
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
-cmake_minimum_required(VERSION 3.7.2)
+cmake_minimum_required(VERSION 3.16.0)
 
 config_option(
     KernelIsMCS KERNEL_MCS "Use the MCS kernel configuration, which is not verified."
--- a/gcc.cmake
+++ b/gcc.cmake
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
-cmake_minimum_required(VERSION 3.7.2)
+cmake_minimum_required(VERSION 3.16.0)
 
 set(CMAKE_SYSTEM_NAME Generic)
 # For a generic system this is unused, so define it to something that will be
--- a/include/32/mode/config.cmake
+++ b/include/32/mode/config.cmake
@@ -4,6 +4,6 @@
 # SPDX-License-Identifier: GPL-2.0-only
 #
 
-cmake_minimum_required(VERSION 3.7.2)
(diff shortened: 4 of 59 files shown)


ratmice:
Doesn't Ubuntu lts (the oldest supported bootstrap env) ship 3.22 now?
I would personally love to see 3.20+ (preferably 3.22 though which is the minimum required version for [corrosion-rs](https://github.com/corrosion-rs/corrosion) for my projects which build rust stuff within the kernel build tree.

3.20 is the minimum version of cmake required for the cmake builtin JSON parsing.

wom-bat:
> Doesn't Ubuntu lts (the oldest supported bootstrap env) ship 3.22 now? I would personally love to see 3.20+ (preferably 3.22 though which is the minimum required version for [corrosion-rs](https://github.com/corrosion-rs/corrosion) for my projects which build rust stuff within the kernel build tree.
> 
> 3.20 is the minimum version of cmake required for the cmake builtin JSON parsing.

focal is still LTS and ships 3.16

wom-bat:
> > Compatibility with versions <3.10 is going away.
> 
> What does this mean exactly? Is this because CMake 4.0.0 just got released? Does the project not build with CMake 4.0.0?

With CMake 3.31, you get:
```
(cmake_minimum_required):
   Compatibility with CMake < 3.10 will be removed from a future version of
   CMake.

   Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
   to tell CMake that the project requires at least <min> but has been updated
   to work with policies introduced by <max> or earlier.
   ```

File: src/arch/x86/kernel/cmdline.c [hunk 1]
```diff
@@ -136,6 +136,19 @@ void cmdline_parse(const char *cmdline, cmdline_opt_t *cmdline_opt)
...
+    printf("Console Com2 0x2f8\n");
+    serial_init(0x3e8);
+    x86KSconsolePort = 0x3e8;
+    printf("Console Com3 0x3e8\n");
+    serial_init(0x2e8);
+    x86KSconsolePort = 0x2e8;
+    printf("Console Com4 0x2e8\n");
+    serial_init(0x3f8);
+    x86KSconsolePort = 0x3f8;
+
```
kent-mcleod, line 151: There's an accidental change committed here it seems.

kent-mcleod:
> 3.20 is the minimum version of cmake required for the cmake builtin JSON parsing.

Separately, if the builtin JSON parsing is useful enough then we could decide to jump forwards to 3.20 and require newer CMake versions as a build dependency as it's easy enough to install a custom version for your dev environment that doesn't conflict with a system installed version

src/arch/x86/kernel/cmdline.c
wom-bat: Arrgh yes... just pushed a version without it.

kent-mcleod (approved):

wom-bat:
> > 3.20 is the minimum version of cmake required for the cmake builtin JSON parsing.
> 
> Separately, if the builtin JSON parsing is useful enough then we could decide to jump forwards to 3.20 and require newer CMake versions as a build dependency as it's easy enough to install a custom version for your dev environment that doesn't conflict with a system installed version

If particular CMake input files need a later version, then they should have the appropriate minimum version in them.  For now, leaving the main ones at 3.16, it'll all work, at least until CMake changes again.

PR #1411, opened 2025-02-26
riscv: disable clock sync test on boot

lsf37:
Timer reads go via mmode on RISC-V, and the hifive board regularly fails the clock sync test at boot because of that.

We could add a larger allowed delta, but the observed values so far are up to 5us, so it is questionable what we then know after the test if we do that. Disable for RISC-V instead.

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -138,7 +138,7 @@ static inline BOOT_CODE pptr_t it_alloc_paging(void)
 /* return the amount of paging structures required to cover v_reg */
 word_t arch_get_n_paging(v_region_t it_veg);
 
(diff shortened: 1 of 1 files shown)


kent-mcleod (approved):

Indanz (changes requested):
I'm against just disabling this test, it checks the assumption that time is mostly synchronised between the cores. As long as the difference is within about WCET it's functionally okay.

If you disable this test, you also need to disable MCS+SMP for RISCV.

Indanz:
Alternatively, if a global lock in OpenSBI is the problem, we can also serialise the test and run it on one core at a time instead of letting all cores query the time at once. But that would complicate the code and not gain us much.

PR #1414, opened 2025-02-26
Changes frm #871 part 2

kent-mcleod:


kent-mcleod:
All hardware tests passsed

PR #1415, opened 2025-02-28
riscv32,llvm: Disable floating point by default

kent-mcleod:
Change the default llvm configuration for riscv32 to disable floating point due to limited clang compiler support

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,7 +27,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
 * riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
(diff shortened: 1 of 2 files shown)


Indanz (commented):
If it's an issue that only affects old clang, perhaps add a version check?

Also, giving a build warning/error for old clang seems better than silently changing the default if clang is used, that is unexpected behaviour which people won't expect.

lsf37:
> If it's an issue that only affects old clang, perhaps add a version check?
> 
> Also, giving a build warning/error for old clang seems better than silently changing the default if clang is used, that is unexpected behaviour which people won't expect.

Agreed, both make sense to me.

I'd be keen to unbreak the test soon -- currently the failing simulation tests cancel the rest.

lsf37 (approved):
This does make simulation work again.

If it does work on more recent `clang`, it'd make sense to check compiler version and/or emit a warning that the setting is changed like Indan was suggesting.

Indanz (approved):
I'm fine with merging this now and fine tuning it later, especially if users can override the default by changing ExtD and ExtF manually.

PR #1417, opened 2025-03-03
qemu-riscv-virt,dts: Fix cell size definitions

kent-mcleod:
The last device tree overlay was overriding the cell widths to 64-bit, so change the first overlay to be compatible with this.

### DIFF
--- a/src/plat/qemu-riscv-virt/overlay-qemu-riscv-virt.dts
+++ b/src/plat/qemu-riscv-virt/overlay-qemu-riscv-virt.dts
@@ -40,11 +40,11 @@
      * - OpenSBI: 128 KiB (= 0x20000) with PMP protection
      */
     reserved-memory {
-        #address-cells = <0x01>;
-        #size-cells = <0x01>;
+        #address-cells = <0x02>;
+        #size-cells = <0x02>;
         ranges;
         sbi@80000000 {
-            reg = <0x80000000 0x200000>;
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

Ivan-Velickovic:
Does this need to be applied to all the other RISC-V overlays? The Star64 regressed for me which has 2 cells for size/address as well. Looks like a lot of the other platforms do too.

Ivan-Velickovic:
```diff
diff --git a/src/plat/star64/overlay-star64.dts b/src/plat/star64/overlay-star64.dts
index bd68fce4f..2180194e4 100644
--- a/src/plat/star64/overlay-star64.dts
+++ b/src/plat/star64/overlay-star64.dts
@@ -28,11 +28,11 @@
      * - OpenSBI: 128 KiB (= 0x20000) with PMP protection
      */
     reserved-memory {
-        #address-cells = <0x01>;
-        #size-cells = <0x01>;
+        #address-cells = <0x02>;
+        #size-cells = <0x02>;
         ranges;
         sbi@40000000 {
-            reg = <0x40000000 0x200000>;
+            reg = <0x0 0x40000000 0x0 0x200000>;
             no-map;
         };
     };
```

This fixed it for Star64, I don't understand why. I thought the whole point of address and size cells was that it could be overridden. Is there a bug in the DTB parsing or something?

PR #1418, opened 2025-03-04
mcs: rephrase setNextInterrupt to ease verification

michaelmcinerney:


Indanz (approved):
Would it make sense to add a local var for `NODE_STATE(ksReleaseQueue).head` too? It would make the code more readable at least.

lsf37 (approved):

lsf37:
> Would it make sense to add a local var for `NODE_STATE(ksReleaseQueue).head` too? It would make the code more readable at least.

It would not hurt as such, but it wouldn't help much for verification. The issue with `refill_head(NODE_STATE(ksCurThread)->tcbSchedContext` is that it accesses the heap, and if you use the expression twice wanting to know that it produces the same value, you have to prove that any heap changes in between leave it unchanged. That's not impossible to prove, but it is much easier to not have to prove anything.

In contrast `NODE_STATE(ksReleaseQueue).head` only accesses a field of a global variable, and independence of heap writes and other globals is automatic.

lsf37:
@michaelmcinerney you will have to reduce the commit message title by one character to appease the Gitlint check.

kent-mcleod (approved):

Indanz:
> > Would it make sense to add a local var for `NODE_STATE(ksReleaseQueue).head` too? It would make the code more readable at least.
> 
> It would not hurt as such, but it wouldn't help much for verification.

I know that, but if we're changing both the proof and C code, now is the time for such small cleanups. If it's trivial on the verification side it would be nice to change that too.

lsf37:
> > It would not hurt as such, but it wouldn't help much for verification.
> 
> I know that, but if we're changing both the proof and C code, now is the time for such small cleanups. If it's trivial on the verification side it would be nice to change that too.

Fair enough. @michaelmcinerney any objections to including Indan's propoosed change?

michaelmcinerney:
> Fair enough. @michaelmcinerney any objections to including Indan's propoosed change?

No objections from me. I'll make this change.

michaelmcinerney:
Forced pushed to rebase and alter the commit message, which now has a shorter first line and an explanation of what's happening

lsf37 (approved):
Good to go from my side when the verification PR is ready.

PR #1420, opened 2025-03-05
rpi4,elfloader: select firmware node for smp boot

kent-mcleod:
Attach the firmware node so that the elfloader can perform smp boots.

### DIFF
--- a/tools/hardware.yml
+++ b/tools/hardware.yml
@@ -224,8 +224,7 @@ devices:
   - compatible:
       - arm,psci-0.2
       - arm,psci-1.0
-  - compatible:
       - fsl,imx6q-src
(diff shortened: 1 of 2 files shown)


tools/hardware.yml
```diff
@@ -224,8 +224,7 @@ devices:
   - compatible:
       - arm,psci-0.2
       - arm,psci-1.0
-  - compatible:
       - fsl,imx6q-src
       - fsl,imx6sx-src
-  - compatible:
       - xlnx,zynq-reset
```
kent-mcleod, line 229: Not sure why there needed to be separate compatible entries here

lsf37 (approved):

PR #1421, opened 2025-03-05
smp: skip schedule or activateThread when unlocked

kent-mcleod:
For synchronised IPI interrupts, the kernel lock acquire is skipped as the lock is already held by a different node. Once the IPI call returns, it isn't safe for the node to do anything other than restore the context of the current active thread and return to user.

Some asserts are added to try and catch any kernel state inconsistencies at runtime.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -42,6 +42,10 @@ exception_t handleInterruptEntry(void)
     irq_t irq;
 
 #ifdef CONFIG_KERNEL_MCS
+    // If the lock isn't held then checkBudget() can't be called.
+    // The lock won't be held only if it's a remote_call IPI.
+    // Each remote_call IPI handler must ensure that it keeps kernel
+    // state consistent.
     if (SMP_TERNARY(clh_is_self_in_queue(), 1)) {
         updateTimestamp();
         checkBudget();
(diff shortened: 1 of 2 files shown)


src/api/syscall.c
```diff
@@ -58,14 +62,12 @@ exception_t handleInterruptEntry(void)
...
-#ifdef CONFIG_KERNEL_MCS
+    // If the lock isn't held then schedule() and activateThread()
+    // aren't safe to call.
     if (SMP_TERNARY(clh_is_self_in_queue(), 1)) {
-#endif
         schedule();
         activateThread();
-#ifdef CONFIG_KERNEL_MCS
     }
-#endif
```
Indanz, line 68: We talked about this before and I think the conclusion was that this change should be done, but can't remember where. Anyway, I don't see why the behaviour should differ between MCS and non-MCS here.

IIRC, x86 virtualisation might depend on the schedule call here in case of cross-core signals, or something like that? Of course that would be broken for MCS now too, just something to keep in mind.

src/smp/ipi.c
```diff
@@ -123,6 +123,25 @@ void handleIPI(irq_t irq, bool_t irqPath)
...
+                case ThreadState_IdleThreadState:
+            #ifdef CONFIG_VTX
+                case ThreadState_RunningVM:
+            #endif
+            break;
+            default:
+                fail("Current thread is blocked");
+
+            }
+        }
```
Indanz, line 144: Not sure about all these, will need to think about it. I definitely want a comment for each assertion.

It's unclear to me why it depends on `irqPath`. The non-irqPath in ipiStallCoreCallback is much more error prone than when irqPath is true, which is much simpler and can be reviewed easier.

src/api/syscall.c
Indanz: This part is a duplication of #1029.

PR #1425, opened 2025-03-13
aarch32/vcpu: save and restore CNTKCTL

ryybrr:
In aarch32 configurations with hypervisor support, CNTKCTL isn't saved/restored when saving/restoring the virtual timer. We address this by mirroring the aarch64 implementation.

### DIFF
--- a/include/arch/arm/arch/object/vcpu.h
+++ b/include/arch/arm/arch/object/vcpu.h
@@ -78,11 +78,11 @@ struct vcpu {
     word_t regs[seL4_VCPUReg_Num];
     bool_t vppi_masked[n_VPPIEventIRQ];
 #ifdef CONFIG_VTIMER_UPDATE_VOFFSET
-    word_t vcpu_padding;
+    /* word_t vcpu_padding; */
     /* vTimer is 8-bytes wide and has the same 8-byte alignment requirement.
-     * If the sum of n_VPPIEventIRQ and seL4_VCPUReg_Num is even, we do not need
-     * extra padding. If the sum is odd we do. It currently is odd, so the extra
-     * padding above is necessary for the struct to remain packed on 32 bit
+     * If the sum of n_VPPIEventIRQ and seL4_VCPUReg_Num is odd, we do not need
+     * extra padding. If the sum is even we do. It currently is odd, so the extra
+     * padding above is unnecessary for the struct to remain packed on 32 bit
      * platforms.
      */
     struct vTimer virtTimer;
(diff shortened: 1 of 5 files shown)


libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
```diff
@@ -117,6 +117,7 @@ typedef enum {
     seL4_VCPUReg_CNTV_CVALlow,
     seL4_VCPUReg_CNTVOFFhigh,
     seL4_VCPUReg_CNTVOFFlow,
+    seL4_VCPUReg_CNTKCTL,
```
kent-mcleod, line 120: This will probably also need an update in the CHANGES file

kent-mcleod (approved):

lsf37 (approved):
Looks good. Agree that we should add a CHANGES entry.

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
ryybrr: Also discovered that `vcpu_padding` in `struct vcpu` needs to be removed, otherwise the struct is no longer packed

include/arch/arm/arch/object/vcpu.h
```diff
@@ -78,11 +78,11 @@ struct vcpu {
...
     bool_t vppi_masked[n_VPPIEventIRQ];
 #ifdef CONFIG_VTIMER_UPDATE_VOFFSET
-    word_t vcpu_padding;
+    /* word_t vcpu_padding; */
     /* vTimer is 8-bytes wide and has the same 8-byte alignment requirement.
      * If the sum of n_VPPIEventIRQ and seL4_VCPUReg_Num is even, we do not need
-     * extra padding. If the sum is odd we do. It currently is odd, so the extra
-     * padding above is necessary for the struct to remain packed on 32 bit
+     * extra padding. If the sum is odd we do. It currently is even, so the extra
+     * padding above is unnecessary for the struct to remain packed on 32 bit
```
Indanz, line 85: Good to add a compile_assert to check for this.

Also, the padding is only necessary for 32-bit, while it was always added?

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
```diff
@@ -71,6 +71,7 @@ typedef enum {
     SEL4_FORCE_LONG_ENUM(seL4_VCPUFault_Msg),
 } seL4_VCPUFault_Msg;
 
+/* If the parity of seL4_VCPUReg_Num changes, the padding inside the vcpu struct must also change */
```
Indanz, line 74: With the compile_assert, this comment is unnecessary.

Indanz (approved):

libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/constants.h
lsf37: I'd like to leave it in. The assert will fail, but it was still a discussion every single time to find out why the verification failed for this when it was violated (also the assert will not fail in the verified config, because they are compiled out there).

lsf37 (approved):
Good to merge from my side, the proofs have passed.

PR #1426, opened 2025-03-13
tools: fix kernel physBase alignment on AArch64

Ivan-Velickovic:


Ivan-Velickovic:
For context, this was affecting Microkit booting on the Raspberry Pi 4B. This change might not show up in the proofs by chance due to all the AArch64 verified platforms already having a `physBase` alignment of 24-bits. The Raspberry Pi 4B has a reserved region from 0x0000 - 0x1000 which is why the alignment was having an affect on `physBase`.

tools/hardware/config.py
```diff
@@ -37,11 +37,18 @@ class ARMConfig(Config):
 
     def __init__(self, sel4arch, addrspace_max):
         super().__init__(sel4arch, addrspace_max)
-        self.SUPERSECTION_BITS = 25 if sel4arch == 'arm_hyp' else 24
+        # On AArch32 the kernel requires at least super section alignment for physBase.
+        # Page sizes differ on arm_hyp which is why the alignment changes even though
+        # both arm_hyp and arm are 32-bit.
+        if sel4arch == 'arm_hyp':
+            self.KERNEL_PHYS_ALIGN = 25
+        elif sel4arch == 'arm':
```
kent-mcleod, line 45: I think this should be aarch32 because the valid options are: `'aarch32', 'aarch64', 'arm_hyp'`

Ivan-Velickovic: Fixed, thanks.

kent-mcleod (approved):

lsf37 (approved):

Indanz:
This should not have been merged as it is, there should have been an explanation somewhere of what broke when aligning things and why this change was necessary.

Was it because the alignment code aligned it to 0 instead of `(1 << 24)` or something like that? In that case the bug would be in the alignment code, not the alignment itself, and this fix masks the real problem.

Ivan-Velickovic:
> This should not have been merged as it is, there should have been an explanation somewhere of what broke when aligning things and why this change was necessary.

It's in the comments on this PR/another issue but yes I should have put an explanation in the commit message. My bad.

My understanding is that this patch is still valid but masks the true issue which @midnightveil has looked into https://github.com/seL4/seL4/pull/1519.

Indanz:
> > This should not have been merged as it is, there should have been an explanation somewhere of what broke when aligning things and why this change was necessary.
> 
> It's in the comments on this PR/another issue but yes I should have put an explanation in the commit message. My bad.

There is no rationale given anywhere for why this change is correct, nor why it is necessary. In the linked PR the problem is that the JSON file doesn't match the actual kernel devices, and the reason given is:

> I assume the mismatch is due to c_header.py making use of physBase but json.py and yaml.py do not.

So obviously the fix would be to fix those scripts instead of apparently changing the kernel memory layout.

> My understanding is that this patch is still valid but masks the true issue which @midnightveil has looked into #1519.

Yes, it is valid, but for very obscure reasons which aren't explained anywhere. On Aarch64:

1) The kernel defines `PADDR_BASE` to 0 and does not use `physBase()` for that.
2) The "physBase" here that ends up in `physBase()` is only used for defining `KERNEL_ELF_PADDR_BASE`.
3) The kernel maps a fixed, huge area starting from zero `PADDR_BASE`.
4) Then it re-uses that kernel window mapping for the ELF mapping (pre PR #1516).
5)  Resulting in no extra alignment requirements for `KERNEL_ELF_PADDR_BASE`, so it is safe to set it to 0 in this PR.

Maybe it's obvious to all of you that setting `KERNEL_PHYS_ALIGN` to zero is fine because of all the above reasons, but mere mortals want some explicit explanation somewhere.

PR #1429, opened 2025-03-20
mcs: some small changes to ease verification

michaelmcinerney:
Please see the commit messages. I can add some more information to these commit messages for how it is that these changes do ease verification, if people would like that.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -482,7 +482,7 @@ static inline void mcsPreemptionPoint(void)
     if (isSchedulable(NODE_STATE(ksCurThread))) {
         /* if the thread is schedulable, the tcb and scheduling context are still valid */
         checkBudget();
-    } else if (NODE_STATE(ksCurSC)->scRefillMax) {
+    } else if (sc_active(NODE_STATE(ksCurSC))) {
         /* otherwise, if the thread is not schedulable, the SC could be valid - charge it if so */
         chargeBudget(NODE_STATE(ksConsumed), false);
     } else {
@@ -504,7 +504,8 @@ static void handleYield(void)
 #ifdef CONFIG_KERNEL_MCS
     /* Yield the current remaining budget */
     ticks_t consumed = NODE_STATE(ksCurSC)->scConsumed + NODE_STATE(ksConsumed);
-    chargeBudget(refill_head(NODE_STATE(ksCurSC))->rAmount, false);
+    refill_t head = *refill_head(NODE_STATE(ksCurSC));
+    chargeBudget(head.rAmount, false);
     /* Manually updated the scConsumed so that the full timeslice isn't added, just what was consumed */
     NODE_STATE(ksCurSC)->scConsumed = consumed;
 #else
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -607,7 +607,9 @@ void chargeBudget(ticks_t consumed, bool_t canTimeoutFault)
     if (likely(NODE_STATE(ksCurSC) != NODE_STATE(ksIdleSC))) {
         if (isRoundRobin(NODE_STATE(ksCurSC))) {
             assert(refill_size(NODE_STATE(ksCurSC)) == MIN_REFILLS);
-            refill_head(NODE_STATE(ksCurSC))->rAmount += refill_tail(NODE_STATE(ksCurSC))->rAmount;
+            refill_t head = *refill_head(NODE_STATE(ksCurSC));
+            refill_t tail = *refill_tail(NODE_STATE(ksCurSC));
+            refill_head(NODE_STATE(ksCurSC))->rAmount = head.rAmount + tail.rAmount;
             refill_tail(NODE_STATE(ksCurSC))->rAmount = 0;
         } else {
             refill_budget_check(consumed);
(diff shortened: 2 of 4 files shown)


michaelmcinerney:
Gitlint is failing, so I'll have to think of a way of shortening the commit message for the fault and timeout handler cap commit. Suggestions welcome (since I'm not a kernel person and I'm not sure what would be better just yet)

Indanz (approved):
It's okay to combine multiple such small, straightforward changes into one commit too.

Indanz:
> Gitlint is failing, so I'll have to think of a way of shortening the commit message for the fault and timeout handler cap commit. Suggestions welcome (since I'm not a kernel person and I'm not sure what would be better just yet)

The main information that a commit message should give is the what and the why. Usually the what is in the title and the why is in the description (the part after an empty line).

In this case I'd combine all 5 commits into one with a title like "MCS: Small changes to ease verification". As there is no real change, it's sufficient to just say why the changes are done.

src/api/syscall.c
```diff
@@ -504,7 +504,8 @@ static void handleYield(void)
 #ifdef CONFIG_KERNEL_MCS
     /* Yield the current remaining budget */
     ticks_t consumed = NODE_STATE(ksCurSC)->scConsumed + NODE_STATE(ksConsumed);
-    chargeBudget(refill_head(NODE_STATE(ksCurSC))->rAmount, false);
```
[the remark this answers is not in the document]
Indanz: If there was a function call between the copy and the use it would be a semantic difference, as the function could potentially modify the copied value.

In this case there is no semantic difference and I expect the compiler to optimise it away.

Would be good to verify for performance sensitive parts like `chargeBudget` though.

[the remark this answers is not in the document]
Indanz: A quick tests confirms compilers have no trouble optimising this: https://godbolt.org/z/zddjxh97o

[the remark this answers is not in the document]
michaelmcinerney: That’s good to know. Thanks for looking into that, Indan.

michaelmcinerney:
> The main information that a commit message should give is the what and the why. Usually the what is in the title and the why is in the description (the part after an empty line).
> 
> In this case I'd combine all 5 commits into one with a title like "MCS: Small changes to ease verification". As there is no real change, it's sufficient to just say why the changes are done.

While I did want to make these changes to help me with the verification, I thought that maybe a kernel engineer would prefer to see something like “use sc_active” or “better cap accessing” rather than just “ease verification”. But I’m certainly happy to have just one commit with your suggested message, too. If others agree that there’s a good way to go, I’ll make that change.

lsf37:
I'm also happy for all of these to go into one commit. Let me double check that the change in `chargeBudget` leads to the same assembly code in the end. The Godbolt test says it should, but I'd like to make sure there's not some other kind of interference.

lsf37:
Something changes somewhere in the elf file, because addresses are shifting but the assembly code in chargeBudget is the same, at least for AArch32.

src/kernel/thread.c
```diff
@@ -627,7 +629,10 @@ void chargeBudget(ticks_t consumed, bool_t canTimeoutFault)
 
 void endTimeslice(bool_t can_timeout_fault)
 {
-    if (can_timeout_fault && !isRoundRobin(NODE_STATE(ksCurSC)) && validTimeoutHandler(NODE_STATE(ksCurThread))) {
+    bool_t round_robin = isRoundRobin(NODE_STATE(ksCurSC));
+    bool_t valid = validTimeoutHandler(NODE_STATE(ksCurThread));
+
+    if (can_timeout_fault && !round_robin && valid) {
```
lsf37, line 635: This is the bit that does change behaviour slightly, because `validTimeoutHandler` now must be executed no matter if `isRoundRobin` returns true or false. The new version has 3 `ldr` instructions that are executed unconditionally, and then a `cmp` and `beq` later. The old version has one `ldr`, a branch, then the other two `ldr` (I assume `validTimeoutHandler`), and then another comparison.

lsf37: Not sure if that comes out as slower or faster in the end, tbh.

lsf37 (approved):
Happy for this to be merged (squashing into a single commit would be good). I don't think there is a real performance impact from the reordering of the two loads.

kent-mcleod (approved):

Indanz:
> While I did want to make these changes to help me with the verification, I thought that maybe a kernel engineer would prefer to see something like “use sc_active” or “better cap accessing” rather than just “ease verification”.

No, because the change is easily seen in the diff for such small changes.

And "better" is a subjective term that doesn't convey real meaning. In this case it's better for verification and also cleaner code-wise. But it could be better for performance or better for security if you added extra checks. The goal of all commits it to make _something_ better, so saying that doesn't add information. What's important for commit message is to record the reason for the change.

There is a huge amount of code cleanup possible, the main reason it's not done is verification, and I assumed a lot of the awkward code is deliberate to ease verification in one way or the other. But judging by these commits, verification also seems to prefer cleaner code.

src/kernel/thread.c
Indanz: That would be true if `validTimeoutHandler` and `isRoundRobin` would be normal functions that can modify any memory. But in this case, both are inline and the compiler has complete view of all function bodies anyway, because seL4 concatenates all files, so it is free to optimise those calls away in case can_timeout_fault is false.

That is, it could have generated the old assembly if it wanted, but the changes were large enough to lead to different code generation.

lsf37: True, since they are inline and obviously read-only the compiler should be able to rearrange if it wants to.

PR #1430, opened 2025-03-23
utils: wrap config_set for verification

lsf37:
Wrap the `config_set` macro in a static inline function so that verification automation does not simplify away dead code branches based on it, but the compiler still does.

In most parts of the proofs we want to pretend that we don't know the config value yet and consider both options. This makes the proofs more independent on the config value that is selected.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -109,8 +109,16 @@ void __builtin_unreachable(void);
 /* need that for compiling with c99 instead of gnu99 */
 #define asm __asm__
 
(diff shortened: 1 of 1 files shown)


lsf37:
I have checked that the compiler does produce the same binary -- it does, modulo a name change for an otherwise unused internal symbol in the symbol table. The assembly code is the same.

Indanz (approved):

PR #1432, opened 2025-03-24
cmake: remove various flags from compilation

midnightveil:
See the descriptions in the commit messages.

### DIFF
--- a/FindseL4.cmake
+++ b/FindseL4.cmake
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: GPL-2.0-only
 #
+cmake_minimum_required(VERSION 3.16.0)
 
 set(KERNEL_PATH "${CMAKE_CURRENT_LIST_DIR}" CACHE STRING "")
 set(KERNEL_HELPERS_PATH "${CMAKE_CURRENT_LIST_DIR}/tools/helpers.cmake" CACHE STRING "")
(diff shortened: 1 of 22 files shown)


kent-mcleod:
The first commit I'd be happy to approve, but I think the second commit isn't as we do want to use the CMAKE_BUILD_TYPE flags.  I can't see any other discussion about the CMAKE_BUILD_TYPE handling, but as far as I'm aware there's no issue with how it's currently handled?

> We unset these values so that our otherwise set values are the only
ones that appear in the command line options. As our options came last,
we usually overrode the defaults, though flags such as DNDEBUG could
sneak through.

It's expected that the kernel build overrides the optimization level, but that the cmake configuration tools allow the compilation options for the different CMAKE_BUILD_TYPE configs to be changed.  

> Note: I'm not entirely sure if the CMake changes are correct; in particular how this interacts with downstream dependants such as sel4test which pull in our CMakeFiles? I'm not sure if this attribute is "namespaced", so to say.

The `CMAKE_C_FLAGS_RELWITHDEBINFO` and other options are globally defined in the cache, so clearing them out in the kernel's CMakeLists.txt would also clear them for all other projects built from the same CMake project.

There's also other places where the CMAKE_BUILD_TYPE flags are applied such as via `CMAKE_ASM_FLAGS_RELWITHDEBINFO` which would be applied for the assembler steps in the kernel build.

kent-mcleod:
> I find it personally quite strange for the compilation to include a series of flags which are there immediately overridden. It's certainly not obvious to me from reading the code; and it's not obvious that this behaviour which as of now is mostly fine will continue to be fine if it changes. There's no issue right now, but what of when someone reorganises the build system or is trying to modify behaviour?

I think it's one of those scenarios where the principle-of-least-astonishment gives you a different answer based on whether the user has a lot of CMake experience or not.

It does appear that `None` is a valid option to set CMAKE_BUILD_TYPE to according to the generated help string in a new CMakeLists.txt project that causes no additional flags to be appended. 

So I think I'd be happy with setting the CMAKE_BUILD_TYPE to None in the kernel's CMakeLists.txt which seems like a stable way of opting just the kernel build out of the additional flags, while still letting someone edit the CMake file to a different profile when developing.

kent-mcleod:
> So, it causes nothing to be appended by virtue of no-one knowing about it :)

That discourse comment conflicts with the source - It's been in the source as help message documentation for at least 8 years: https://github.com/Kitware/CMake/blob/fd26efc2bd9228c27c30f5b35eda2421b280c548/Modules/CMakeInitializeConfigs.cmake#L23.
But maybe it's meant to imply no value instead of the string "None". It's not possible to unset the local value however, as it's set in the Cache by the parent project, and an unset local value leads the cache'd value to be used.


> And it's true --- if I modify the CMakeLists to not even set these values (or set them to garbage), for the sel4test manifest it gets overridden.

The local value set by the kernel's cmake cannot be overridden by parent projects, but the setting only applies to the targets created within the kernel's cmake directory and subdirectories. 
The sel4test project sets a global default that is stored in the cache, but this is then overridden by the kernel for only its own targets.

This is because of CMake's approach of dereferencing variables in a defined order: https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables

> What I've done here is add a fake CMAKE_BUILD_TYPE called seL4Kernel. This should:
> - make it clear to anyone looking at the CMake configuration cache that this is being set by the seL4Kernel (unless it's overridden, as I mentioned elfloader et al seems to do)

I still think None is better here. (See my previous comment about setting to None not having any effect on the value stored in the cache)

Indanz:
(I can't merge this because of the style problems. The horrible style stuff should be fixed by https://github.com/seL4/seL4_tools/pull/249/https://github.com/seL4/seL4/pull/1612 though.)

midnightveil:
Preprocess failures is because the preprocess sync failed on the last commit to master seL4, not from this PR.

I've fixed the style problems now.

PR #1434, opened 2025-03-26
Add missing userError for VSpace invocation

Ivan-Velickovic:


Ivan-Velickovic:
Should the description of the invocations be updated to say 'page' rather than 'pages'?

lsf37 (approved):
:+1: I haven't checked the other invocations. If they are talking about pages when they it can only be a single page then it would be better to change them, yes.

PR #1438, opened 2025-03-30
x86/vcpu: Allow CR0.PE and CR0.PG to be cleared

losfair:
If "Unrestricted Guest" mode is enabled in secondary control, CR0.PE and CR0.PG may be 0.


src/arch/x86/object/vcpu.c
```diff
@@ -290,7 +290,14 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
...
     entry_control_low = x86_rdmsr_high(entry_ctls);
 
-    cr0_high = x86_rdmsr_low(IA32_VMX_CR0_FIXED0_MSR);
+    // Intel SDM, Vol. 3C, 24.8:
+    //
+    // > Later processors support a VM-execution control called "unrestricted
+    // > guest" (see Section 25.6.2). If this control is 1, CR0.PE and CR0.PG
+    // > may be 0 in VMX non-root operation (even if the capability MSR
+    // > IA32_VMX_CR0_FIXED0 reports otherwise).1 Such processors allow guest
+    // > software to run in unpaged protected mode or in real-address mode.
```
Indanz, line 299: The numbers don't match the combined SDM I just downloaded from https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html. Could you specify which document exactly or double check they didn't change the chapters is the newest version or something?

In my version that text is in chapter 25.8, "Restrictions on VMX Operations". Considering the chapters change, perhaps also give the chapter title so it's easier to find it.

src/arch/x86/object/vcpu.c
```diff
@@ -290,7 +290,14 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
...
+    cr0_high = x86_rdmsr_low(IA32_VMX_CR0_FIXED0_MSR) & ~0x80000001;
```
Indanz, line 300: You should check that unrestricted guest mode is actually allowed before doing this.

Also, if there are no defines for individual CR0 bits yet, then please add them and use them here. Some are defined in include/arch/x86/arch/machine/cpu_registers.h already. Same for the IA32_VMX_TRUE_PROCBASED_CTLS_MSR/IA32_VMX_PROCBASED_CTLS_MSR bits you need to check.

src/arch/x86/object/vcpu.c
```diff
@@ -392,7 +399,7 @@ static bool_t BOOT_CODE check_vtx_fixed_values(bool_t useTrueMsrs)
     uint32_t local_entry_control_high = x86_rdmsr_low(entry_ctls);
     uint32_t local_entry_control_low = x86_rdmsr_high(entry_ctls);
 
-    uint32_t local_cr0_high = x86_rdmsr_low(IA32_VMX_CR0_FIXED0_MSR);
+    uint32_t local_cr0_high = x86_rdmsr_low(IA32_VMX_CR0_FIXED0_MSR) & ~0x80000001;
```
Indanz, line 402: Same here.

Indanz (commented):
So this seems necessary because Intel messed it up again and made the behaviour inconsistent.

Anyway, who controls whether unrestricted guest mode is enabled or not? Is it set by the BIOS? Is this something that user space would want to control?

losfair:
>

losfair:
> Anyway, who controls whether unrestricted guest mode is enabled or not? Is it set by the BIOS? Is this something that user space would want to control?

Unrestricted guest mode is enabled by setting the `UNRESTRICTED_GUEST` bit in `SECONDARY_PROCBASED_EXEC_CONTROLS`. Tested, and works.

> I guess you want to use this for one reason or other, otherwise you wouldn't run into this.

I was trying to boot an image in [PVH direct boot](https://xenbits.xen.org/docs/unstable/misc/pvh.html) mode, and that requires switching from unpaged 32-bit mode to long mode. Non-UG mode requires paging so it's not possible.

Indanz:
Most implemented by #1600 now. Feel free to open a new PR for missing bits.

PR #1440, opened 2025-04-01
mcs: refactor reply functions to ease verification

michaelmcinerney:
The outermost if statement can be removed in reply_pop because reply_pop is called only in reply_remove, which includes an explicit check for this condition.

The new inline function setThreadStateBlockedOnReply is used within reply_push.

### DIFF
--- a/src/object/reply.c
+++ b/src/object/reply.c
@@ -30,8 +30,7 @@ void reply_push(tcb_t *tcb_caller, tcb_t *tcb_callee, reply_t *reply, bool_t can
 
     /* link caller and reply */
     reply->replyTCB = tcb_caller;
-    thread_state_ptr_set_replyObject(&tcb_caller->tcbState, REPLY_REF(reply));
-    setThreadState(tcb_caller, ThreadState_BlockedOnReply);
+    setThreadStateBlockedOnReply(tcb_caller, reply);
 
     if (sc_donated != NULL && tcb_callee->tcbSchedContext == NULL && canDonate) {
         reply_t *old_caller = sc_donated->scReply;
@@ -41,12 +40,12 @@ void reply_push(tcb_t *tcb_caller, tcb_t *tcb_callee, reply_t *reply, bool_t can
                SC_PTR(call_stack_get_callStackPtr(old_caller->replyNext)) == sc_donated);
 
         /* push on to stack */
-        reply->replyPrev = call_stack_new(REPLY_REF(old_caller), false);
         if (old_caller) {
             old_caller->replyNext = call_stack_new(REPLY_REF(reply), false);
         }
-        reply->replyNext = call_stack_new(SC_REF(sc_donated), true);
+        reply->replyPrev = call_stack_new(REPLY_REF(old_caller), false);
         sc_donated->scReply = reply;
+        reply->replyNext = call_stack_new(SC_REF(sc_donated), true);
 
         /* now do the actual donation */
         schedContext_donate(sc_donated, tcb_callee);
@@ -64,27 +63,24 @@ void reply_pop(reply_t *reply, tcb_t *tcb)
     word_t next_ptr = call_stack_get_callStackPtr(reply->replyNext);
     word_t prev_ptr = call_stack_get_callStackPtr(reply->replyPrev);
 
-    if (likely(next_ptr != 0)) {
-        assert(call_stack_get_isHead(reply->replyNext));
+    SC_PTR(next_ptr)->scReply = REPLY_PTR(prev_ptr);
+    if (prev_ptr != 0) {
+        REPLY_PTR(prev_ptr)->replyNext = reply->replyNext;
+        assert(call_stack_get_isHead(REPLY_PTR(prev_ptr)->replyNext));
+    }
 
-        SC_PTR(next_ptr)->scReply = REPLY_PTR(prev_ptr);
-        if (prev_ptr != 0) {
-            REPLY_PTR(prev_ptr)->replyNext = reply->replyNext;
-            assert(call_stack_get_isHead(REPLY_PTR(prev_ptr)->replyNext));
-        }
+    reply->replyNext = call_stack_new(0, false);
 
-        /* give it back */
-        if (tcb->tcbSchedContext == NULL) {
-            /* only give the SC back if our SC is NULL. This prevents
-             * strange behaviour when a thread is bound to an sc while it is
-             * in the BlockedOnReply state. The semantics in this case are that the
-             * SC cannot go back to the caller if the caller has received another one */
-            schedContext_donate(SC_PTR(next_ptr), tcb);
-        }
+    /* give it back */
+    if (tcb->tcbSchedContext == NULL) {
+        /* only give the SC back if our SC is NULL. This prevents
+         * strange behaviour when a thread is bound to an sc while it is
+         * in the BlockedOnReply state. The semantics in this case are that the
+         * SC cannot go back to the caller if the caller has received another one */
+        schedContext_donate(SC_PTR(next_ptr), tcb);
     }
 
     reply->replyPrev = call_stack_new(0, false);
-    reply->replyNext = call_stack_new(0, false);
     reply_unlink(reply, tcb);
 }
 
--- a/include/kernel/thread.h
+++ b/include/kernel/thread.h
@@ -307,5 +307,12 @@ void awaken(void);
 /* Place the thread bound to this scheduling context in the release queue
  * of periodic threads waiting for budget recharge */
 void postpone(sched_context_t *sc);
+
+static inline void setThreadStateBlockedOnReply(tcb_t *tptr, reply_t *reply)
+{
+    thread_state_ptr_set_tsType(&tptr->tcbState, ThreadState_BlockedOnReply);
+    thread_state_ptr_set_replyObject(&tptr->tcbState, REPLY_REF(reply));
+    scheduleTCB(tptr);
+}
 #endif
 


src/object/reply.c
```diff
@@ -64,27 +63,26 @@ void reply_pop(reply_t *reply, tcb_t *tcb)
     word_t next_ptr = call_stack_get_callStackPtr(reply->replyNext);
     word_t prev_ptr = call_stack_get_callStackPtr(reply->replyPrev);
 
-    if (likely(next_ptr != 0)) {
-        assert(call_stack_get_isHead(reply->replyNext));
+    assert(call_stack_get_isHead(reply->replyNext));
```
lsf37, line 66: We should drop this `assert` as well, it is explicitly checked in the `if` just before `reply_pop` is called.

lsf37 (approved):
See comment on the `assert`, otherwise good to go from my side.

Indanz (approved):
Changes seem correct.

Code-wise this change adds code which is less readable, because it deviates from the rest. It would be fine if you added `setThreadStateRestart` and others too, that would actually clean up the code a bit, as 90% of the calls set the state to restart. But just a single function for this corner case is a bit weird.

I assume that setting the reply object before the state is `BlockedOnReply` is the problem. Have you tried swapping the order of the `thread_state_ptr_set_replyObject` and `setThreadState` calls? (And probably doing the `setThreadState` before updating `replyTCB` too.)

As far as reply_op, I would get rid of that altogether and merge it with reply_remove, as most code is duplicated, except the chain gets broken when removing a reply from the middle? Why is that?

michaelmcinerney:
> Code-wise this change adds code which is less readable, because it deviates from the rest. It would be fine if you added `setThreadStateRestart` and others too, that would actually clean up the code a bit, as 90% of the calls set the state to restart. But just a single function for this corner case is a bit weird.
> 
> I assume that setting the reply object before the state is `BlockedOnReply` is the problem. Have you tried swapping the order of the `thread_state_ptr_set_replyObject` and `setThreadState` calls? (And probably doing the `setThreadState` before updating `replyTCB` too.)
> 

I'm sure Gerwin would be able to give you better answers to your questions, but I'll try...

Yes, the problem here is setting the reply object. Here's a very wordy explanation of the issue. Some thread states, like `Running` or `Inactive` do not take any further arguments, but we do have some thread states that do, such as `BlockedOnReply` and `BlockedOnReceive`. In the Haskell, these extra arguments appear as arguments to the constructors of the datatype, so we might have a thread state that is `BlockedOnReply reply`, and therefore, setting the thread state of some TCB to this thread state is an atomic operation. Showing correspondence between an atomic operation in the Haskell and an operation in the formalised C that consists of several lines is rather difficult, so we chose to follow the examples elsewhere, for example, in `sendIPC`, where we set the thread state to `BlockedOnSend` with a bunch of arguments. The order in which the reply object and the thread state type is set is not important, actually. Then we thought it would be nice if this were a small inline function. We did consider writing some other inline functions for the other thread states which take further arguments, but decided to do this as needed.

> As far as reply_op, I would get rid of that altogether and merge it with reply_remove, as most code is duplicated, except the chain gets broken when removing a reply from the middle? Why is that?

I'm not sure either. I was hoping you might have some idea :-) . I would very much like to understand the semantics of the replies better. 

It seems to me like `reply_remove` is rather confused: if we are removing the head of the reply stack, then that's `reply_pop`, which might involve a `schedContext_donate`, passing the scheduling context back down the stack, but if it's not the head, then yes the chain is broken. Breaking the chain seems like a fairly catastrophic operation. The scheduling context can never return to the original caller, so presumably the user would have to do something else to salvage this situation. I mentioned to Gerwin that it would maybe be nice if the kernel cleaned up the rest of the reply stack, unlinking all the replies from the TCBs and from the sched context, but he pointed out that this would be a long-running operation, and would require a preemption point, which would be super difficult to verify. 

I'm also not sure why `reply_remove` might be called in `finalise_cap`, since we may end up performing `schedContext_donate`, which is an odd thing to be doing if we're just trying to finalise objects. Gerwin did say that these things were considered when the semantics were changed (commit `6195ea6651fed` contains this change). And also not sure why we might end up breaking the chain during `do_reply_transfer` when `reply_remove` is called. 

Ideally, to me, we should have `reply_push`, `reply_pop`, and then some operation that cancels the reply, cleaning up the whole reply stack. Maybe this can be done as "future work" if the semantics make sense

Indanz:
> In the Haskell, these extra arguments appear as arguments to the constructors of the datatype, so we might have a thread state that is `BlockedOnReply reply`, and therefore, setting the thread state of some TCB to this thread state is an atomic operation. Showing correspondence between an atomic operation in the Haskell and an operation in the formalised C that consists of several lines is rather difficult

Thanks, that explains it and also why you can't use `setThreadState` in your new `setThreadStateBlockedOnReply`.

> It seems to me like `reply_remove` is rather confused: if we are removing the head of the reply stack, then that's `reply_pop`, which might involve a `schedContext_donate`, passing the scheduling context back down the stack

I think the idea was that the caller would either call reply_remove or reply_pop, depending on what's required, but in time that decision ended up in reply_remove instead, making it appear confused.

> but if it's not the head, then yes the chain is broken. Breaking the chain seems like a fairly catastrophic operation.

Agreed.

> The scheduling context can never return to the original caller, so presumably the user would have to do something else to salvage this situation. I mentioned to Gerwin that it would maybe be nice if the kernel cleaned up the rest of the reply stack, unlinking all the replies from the TCBs and from the sched context, but he pointed out that this would be a long-running operation, and would require a preemption point, which would be super difficult to verify.

But why clean up at all? Why not just remove the intermediate reply and keep the chain intact? Then the SC will find its way to the original caller and the code to remove a reply can be shared and reply_pop can be easily merged with reply_remove: It's only one extra if-check to see whether the SC needs to be moved/donated to someone else.

Maybe the problem is that the reply chain needs to be 1-to-1 with the endpoint call chain? And if an intermediate reply object gets removed, passing back the SC is broken anyway?

> I'm also not sure why `reply_remove` might be called in `finalise_cap`, since we may end up performing `schedContext_donate`, which is an odd thing to be doing if we're just trying to finalise objects.

If the reply object gets revoked, it needs to be removed from any chain. Probably the same if the TCB object gets revoked. In the case of `finale_cap` moving/donating the SC is an unnecessary operation, as that task will probably never be woken up anyway (as the caller can't reply), so it's only there for the common case of a task replying I think.

> Gerwin did say that these things were considered when the semantics were changed (commit `6195ea6651fed` contains this change). And also not sure why we might end up breaking the chain during `do_reply_transfer` when `reply_remove` is called.

That commit message doesn't say why the change was made, just that it was. That's useless. That's why recording the reason for a change is so important, so people in the future know the rationale. @kent-mcleod, can you ask Anna if she remembers why this change was made?

> Ideally, to me, we should have `reply_push`, `reply_pop`, and then some operation that cancels the reply, cleaning up the whole reply stack if we remove a reply from the middle of the chain. Maybe this can be done as "future work" if the semantics make sense

I question the whole existence of reply objects, it could easily have been internal TCB state, except having a reply object makes it possible to gather multiple calls before replying to any of them. But even if you do that, only one caller will donate its SC, so you only need a bit in the caller's TCB to mark its SC as donated and the receiver can return the SC if replying to a TCB with that bit set. So in conclusion, I don't see why you need to maintain a reply chain to begin with. If you're breaking it on any complication, why bother having it at all?

Indanz:
> I'm not sure either. I was hoping you might have some idea :-) . I would very much like to understand the semantics of the replies better.

This applies to all of MCS: We need to know what we want in detail before we spend a lot of time verifying something that might not be it. In large part that should be driven by what seL4 wants to prove, especially for the timing and scheduling properties.

lsf37:
I can answer at least one small bit: we can't remove a reply from the chain and leave the rest intact, because that would mean that someone suddenly gets a reply from somebody they are not expecting (e.g. skipping one server in a server call chain). At best this will just fail catastrophically, at worst this will work by accident and be extremely confusing. Hence the decision to definitely break things at that point.

The problem is that if you have a chain of threads calling the next, and one in the middle disappears, that chain just does not make much sense any more to roll up.

Indanz:
> I question the whole existence of reply objects, it could easily have been internal TCB state, except having a reply object makes it possible to gather multiple calls before replying to any of them. But even if you do that, only one caller will donate its SC, so you only need a bit in the caller's TCB to mark its SC as donated and the receiver can return the SC if replying to a TCB with that bit set. So in conclusion, I don't see why you need to maintain a reply chain to begin with. If you're breaking it on any complication, why bother having it at all?

To elaborate, we already have an implicit call chain with threads blocking on reply and the final thread replying to one caller, so I don't think maintaining the same chain explicitly via reply objects adds anything to that. If you do that, you have to proof that the tcb chain in the reply chain equals the actual caller TCB chain.

A TCB can only have one active SC, so no need to track which one to return when it's donated.

To handle the corner case of a blocked caller losing its non-donated SC because of a syscall/revoke or other exceptional situation instead of donation, you need one bit in the TCB to remember that. That bit should always be cleared when a task gets a SC and only be set on SC donation.

(You could also have a bit to mark a TCB running on a donated SC, but not sure if that helps for any corner cases.)

Indanz:
> I can answer at least one small bit: we can't remove a reply from the chain and leave the rest intact, because that would mean that someone suddenly gets a reply from somebody they are not expecting (e.g. skipping one server in a server call chain). At best this will just fail catastrophically, at worst this will work by accident and be extremely confusing. Hence the decision to definitely break things at that point.

But this is still possible, as anyone can reply, it does not have to be the thread that received the donated SC. Example call chains with passive B, C and active A, D:

`A -> B -> C`

Thread D can reply to B's reply object, if it has access to the cap. A can't receive until it gets a SC and it will never get the one donated to C back. B will probably get the SC and then silently turn into an active server instead of a passive one.

If D replies to C's reply object, C will lose its SC and hang at a random point (without timeout exception I think), but at least A is okay, it's just C that's left broken.

> The problem is that if you have a chain of threads calling the next, and one in the middle disappears, that chain just does not make much sense any more to roll up.

Agreed, when either replying to the middle of the chain or removing an intermediate reply object, things are terminally broken and don't make any sense any more.

I think it would make more sense to fault a task replying to the wrong reply object than to silently break things.

The manual still describes the old behaviour where the SC would have been returned, pre commit 6195ea6651fed. But having an SC and no one able to reply makes no functional difference.

Indanz:
I've looked into the reply code and my conclusion is that we have the reply chain construction for the following reasons:
- To keep track of which SC got donated so it can be returned.
- Keeping track of which SC got donated makes it possible to prevent another SC from being returned.
- To avoid N copies of one SC pointer which would make SC cleanup an O(N) operation, the reply chain is build instead of remembering which SC was donated per TCB.

There is no need to build the reply chain if no SC is being donated.

The reason the above is needed at all is because of exceptional, nonsensical corner cases (aka user inflicted pain):
1) Replying task lost the donated SC and somehow has another one (caused by manual SC modifications through syscalls), and we don't want to return the wrong SC when replying.
2) Replying task is not the called task and is not running on the same donated SC (reply cap passed on or shared CSpace).

In both cases the replying task does not have the expected SC.

Both 1 and 2 are in my opinion design mistakes (bugs) in user space that result in a broken system. I don't think it matters that much how we handle this, as long as we don't break things more. It's okay to break things differently if it simplifies the code and verification.

For 1, if the intention was to turn the task into a non-passive sever, it should have made sure the task is not in the middle of a call when doing this. The SC bind call would have failed, so the user deliberately must have unbound and lost the donated SC first. In a correct implementation the user would also have manually returned the SC to the caller, except it can't know who the caller is if there is more than one client.

When doing 2, either the SC doesn't get returned if the reply is in the middle of the chain, or another task is left in limbo because the donated SC is yanked away. Recovery is difficult because it's unclear where the SC is. Ideally such replies should fault, but that's difficult to implement with less complexity than the call chain.

All in all, if it would be up to me, I would probably get rid of the requirement that only the donated SC can be returned, as that simplifies the code the most. That would mean that task A would get task B's new SC after manual manipulation in case 1 and in case 2 would get task D's SC instead. The implication is that a reply object gives you the right to donate your SC to the caller, even if it's not the original donated SC.

If we want to avoid the above and prevent the wrong SC of being returned, it's relatively easy to prevent for case 1: We just need to keep track of a (TCB) boolean which says whether the SC was donated or not. If the SC was assigned manually via a syscall the bit would be cleared, if the SC is donated the bit is set in both the caller and the receiver. For case 2 that would prevent the wrong SC being returned if the other task is active, but if it's also passive then the wrong SC can still be returned to the wrong task.

lsf37:
Let's continue the discussion on an issue -- I'll merge this one here so Michael can make progress on the verification side.

Indanz:
> Let's continue the discussion on an issue

Just for completeness sake, the current behaviour is necessary to make it possible for timeout fault handlers to do anything useful. Timeout handlers currently only have two options:
- Reply on behalf of the faulting server to recover the client and reset the server.
- Give an emergency SC to the server.

Which correspond to the two cases I mentioned earlier, but with the difference that a timeout fault handler doing it actually makes sense and is not a user space bug or design problem. With my change both options would break, as the client would get the wrong SC and its original one will be lost. With that in mind I don't think my proposal is going to work, not without a better way to handle timeout faults, so I won't open an issue for it.

As for breaking the chain or not when revoking intermediate replies: I still don't know the reason for why it was changed, I suspect it doesn't makes any difference.

PR #1441, opened 2025-04-01
github: deploy mcs verification manifest

lsf37:
Deploy the MCS verification manifest for versions that have no MCS preprocess differences. These may be different from non-MCS preprocess outcomes, and since the MCS verification is on a branch, we can deploy it separately.

C verification on MCS has progressed enough that we want to do this automatically now, as we do for the master branch.

### DIFF
--- a/.github/workflows/preprocess-deploy.yml
+++ b/.github/workflows/preprocess-deploy.yml
@@ -32,11 +32,24 @@ jobs:
     strategy:
       matrix:
         arch: [ARM, ARM_HYP, AARCH64, RISCV64, X64]
-        # no MCS here, auto-updating mcs.xml should be a separate job.
+        # no MCS here, see separate preprocess-mcs.
     steps:
     - uses: seL4/ci-actions/preprocess@master
       with:
         L4V_ARCH: ${{ matrix.arch }}
+
+  preprocess-mcs:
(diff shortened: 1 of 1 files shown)


.github/workflows/preprocess-deploy.yml
```diff
@@ -32,12 +32,25 @@ jobs:
...
       with:
         L4V_ARCH: ${{ matrix.arch }}
 
+  preprocess-mcs:
+    name: Preprocess (MCS)
+    needs: code
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: [RISCV64]
```
Indanz, line 47: Is it intentional to run this only for RSICV64? Don't we have a 32-bit arch that has verified MCS?

Indanz (approved):

.github/workflows/preprocess-deploy.yml
lsf37: Yes, C verification is currently only being done for RISCV64.

When that is finished, the plan is to extend, but we'll need to see how fast that goes. The hope is relatively fast, because MCS mostly doesn't touch architecture specific stuff, but it will still be a lot of work. Until then we don't want to block on other architectures.

PR #1443, opened 2025-04-02
arch/x86: Support Xen x86/HVM direct boot ABI

losfair:
Adds support for booting the kernel with the Xen x86/HVM direct boot ABI (https://xenbits.xen.org/docs/unstable/misc/pvh.html). This allows seL4 to boot natively within virtualized environments like QEMU and Firecracker:

```
qemu-system-x86_64 -m 512M -kernel kernel.elf -initrd app.elf -serial stdio -display none
```


src/arch/x86/kernel/boot_sys.c
```diff
@@ -63,23 +63,25 @@ BOOT_CODE static paddr_t find_load_paddr(paddr_t min_paddr, word_t image_size)
     int i;
 
     for (i = 0; i < boot_state.mem_p_regs.count; i++) {
-        paddr_t start = MAX(min_paddr, boot_state.mem_p_regs.list[i].start);
+        paddr_t start = MAX(boot_state.ki_p_reg.end, MAX(min_paddr, boot_state.mem_p_regs.list[i].start));
```
lsf37, line 66: why is this change needed and correct?

src/arch/x86/kernel/boot_sys.c
```diff
@@ -63,23 +63,25 @@ BOOT_CODE static paddr_t find_load_paddr(paddr_t min_paddr, word_t image_size)
...
         paddr_t end = boot_state.mem_p_regs.list[i].end;
         word_t region_size = end - start;
 
-        if (region_size >= image_size) {
+        if (end > start && region_size >= image_size) {
```
lsf37, line 70: what has changed to necessitate a change here?

src/arch/x86/kernel/boot_sys.c
```diff
@@ -452,9 +475,7 @@ static BOOT_CODE bool_t try_boot_sys(void)
     assert(mods_end_paddr > boot_state.ki_p_reg.end);
 
     printf("ELF-loading userland images from boot modules:\n");
-    load_paddr = mods_end_paddr;
-
-    load_paddr = load_boot_module(boot_state.boot_module_start, load_paddr);
+    load_paddr = load_boot_module(boot_state.boot_module_start, &mods_end_paddr);
```
lsf37, line 478: `mods_end_paddr` is a local variable, you are not allowed to take its address in seL4. See also <https://docs.sel4.systems/processes/style-guide.html#verification-requirements>

src/arch/x86/kernel/boot_sys.c
```diff
@@ -700,6 +721,84 @@ static BOOT_CODE bool_t try_boot_sys_mbi2(
...
+            *mem_lower_out = entry->size;
+        }
+    }
+
+    return true;
+}
+
+static BOOT_CODE bool_t try_boot_sys_hvm(
+    hvm_start_info_t *info
+)
```
lsf37, line 751: Please use the same state as other declarations.

src/arch/x86/kernel/boot_sys.c
```diff
@@ -700,6 +721,84 @@ static BOOT_CODE bool_t try_boot_sys_mbi2(
...
+    hvm_memmap_entry_t *entry;
+    printf("Parsing HVM physical memory map\n");
+
+    for (i = 0; i < num_entries; i++) {
+        entry = &entries[i];
+        printf("\tPhysical Memory Region from %lx size %lx type %d\n", (long)entry->addr, (long)entry->size, entry->type);
+        if (entry->type == XEN_HVM_MEMMAP_TYPE_RAM  && entry->addr >= HIGHMEM_PADDR && entry->size >= BIT(PAGE_BITS)) {
+            if (!add_mem_p_regs((p_region_t) {
+            ROUND_UP(entry->addr, PAGE_BITS), ROUND_DOWN(entry->addr + entry->size, PAGE_BITS),
+            })) {
```
lsf37, line 736: Please use semantic indentation.

Why are all these ROUND_UP, ROUND_DOWN safe?

src/arch/x86/kernel/boot_sys.c
```diff
@@ -63,23 +63,25 @@ BOOT_CODE static paddr_t find_load_paddr(paddr_t min_paddr, word_t image_size)
...
+        if (end > start && region_size >= image_size) {
             return start;
         }
     }
 
     return 0;
 }
 
-BOOT_CODE static paddr_t load_boot_module(word_t boot_module_start, paddr_t load_paddr)
+BOOT_CODE static paddr_t load_boot_module(word_t boot_module_start, paddr_t *load_paddr_p)
```
lsf37, line 78: If there are added side effects like changing the content of `load_paddr_p` because things are moved, these should be documented in a comment.

lsf37 (changes requested):
This needs more correctness arguments and currently does not satisfy the C subset requirements.

src/arch/x86/kernel/boot_sys.c
lsf37: > Please use the same style as other declarations.

src/arch/x86/kernel/boot_sys.c
Indanz: I think that checks for overflows and is just a bugfix. The first two changes should have their own commit(s) with explanations, they're not part of Xen/HVM support directly.

src/arch/x86/kernel/boot_sys.c
Indanz: Probably a bugfix, but it should have much more explanation. It should be well defined what the end of the boot state is, now two addresses are checked, which is weird.

Indanz:
No feedback from original dev for almost a year.

PR #1445, opened 2025-04-06
CMake: Avoid rebuilds by diffing generated headers

FennelFoxxo:
With the current implementation of add_config_library, the config_gen Python script is called on to regenerate the config header and json files every time CMake is configured. Since many targets rely on these files, it practically forces the entire project to be rebuilt whenever CMake is reconfigured, even if the project has already been configured and no settings have changed.

This change makes it so the generated header and json files get saved to temporary locations, and only override the final header and json files if the contents are different, thus avoiding a timestamp update and rebuild if they are the same. The temporary files are deleted after. This is all done at configure time (see #1105 for why).

### DIFF
--- a/tools/config_gen.py
+++ b/tools/config_gen.py
@@ -25,16 +25,20 @@ def main():
         "--write-c",
         metavar="OUT_FILE",
         dest="out_file_c",
+        nargs="?",
         type=argparse.FileType("w"),
-        default=sys.stdout,
+        const=sys.stdout,
+        default=None,
         help="Output C header file.",
     )
     parser.add_argument(
         "--write-json",
         metavar="OUT_FILE",
         dest="out_file_json",
+        nargs="?",
         type=argparse.FileType("w"),
-        default=sys.stdout,
+        const=sys.stdout,
+        default=None,
         help="Output JSON file.",
     )
     args = parser.parse_args()
--- a/tools/config_gen.py
+++ b/tools/config_gen.py
@@ -22,11 +22,18 @@ def main():
         help="Input YAML file.",
     )
     parser.add_argument(
+        "--skip-unchanged",
+        dest="skip_unchanged",
+        action="store_true",
+        help="Only write output file if new content is different"
+    )
+    parser.add_argument(
         "--write-c",
         metavar="OUT_FILE",
         dest="out_file_c",
         nargs="?",
-        type=argparse.FileType("w"),
+        # Use a+ mode to open in read/write mode without truncating existing contents
+        type=argparse.FileType("a+"),
         const=sys.stdout,
         default=None,
         help="Output C header file.",
@@ -36,25 +43,25 @@ def main():
         metavar="OUT_FILE",
         dest="out_file_json",
         nargs="?",
-        type=argparse.FileType("w"),
+        type=argparse.FileType("a+"),
         const=sys.stdout,
         default=None,
         help="Output JSON file.",
     )
     args = parser.parse_args()
     generate(args.in_file_yaml, out_file_c=args.out_file_c,
-             out_file_json=args.out_file_json)
+             out_file_json=args.out_file_json, skip_unchanged=args.skip_unchanged)
 
 
-def generate(in_file_yaml, out_file_c=None, out_file_json=None):
+def generate(in_file_yaml, out_file_c=None, out_file_json=None, skip_unchanged=False):
     config = yaml.safe_load(in_file_yaml)
     if out_file_c is not None:
-        generate_c(config, out_file_c)
+        generate_c(config, out_file_c, skip_unchanged)
     if out_file_json is not None:
-        generate_json(config, out_file_json)
+        generate_json(config, out_file_json, skip_unchanged)
 
 
-def generate_c(config, out_file):
+def generate_c(config, out_file, skip_unchanged):
     header_contents = "#pragma once\n\n"
 
     for key, value in config.items():
@@ -76,11 +83,24 @@ def generate_c(config, out_file):
 
         header_contents += f"{entry}\n"
 
-    out_file.write(header_contents)
+    write_file_lazy(out_file, header_contents, skip_unchanged)
 
 
-def generate_json(config, out_file):
-    json.dump(config, out_file, indent=4)
+def generate_json(config, out_file, skip_unchanged):
+    json_contents = json.dumps(config, indent=4)
+    write_file_lazy(out_file, json_contents, skip_unchanged)
+
+
+def write_file_lazy(out_file, content, skip_unchanged):
+    if out_file == sys.stdout:
+        out_file.write(content)
+        return
+
+    out_file.seek(0)
+    if not skip_unchanged or out_file.read() != content:
+        out_file.seek(0)
+        out_file.truncate()
+        out_file.write(content)
 
 
 if __name__ == "__main__":
--- a/tools/helpers.cmake
+++ b/tools/helpers.cmake
@@ -552,7 +552,7 @@ function(add_config_library prefix configure_template)
 
     execute_process(
         COMMAND
-            "${PYTHON3}" "${CONFIG_GEN_PATH}" "${config_yaml_file}" --write-c
+            "${PYTHON3}" "${CONFIG_GEN_PATH}" "${config_yaml_file}" --skip-unchanged --write-c
             "${config_header_file}" --write-json "${config_json_file}"
         RESULT_VARIABLE error
     )


lsf37 (approved):
This looks good from my side, but @kent-mcleod is the resident CMake expert, so I'll wait for his opinion.

tools/helpers.cmake
```diff
@@ -553,13 +553,22 @@ function(add_config_library prefix configure_template)
     execute_process(
         COMMAND
             "${PYTHON3}" "${CONFIG_GEN_PATH}" "${config_yaml_file}" --write-c
-            "${config_header_file}" --write-json "${config_json_file}"
+            "${config_header_file}.temp" --write-json "${config_json_file}.temp"
```
Indanz, line 556: Now it's generating temp files, it would perhaps make more sense to put them elsewhere than CONFIG_GEN_PATH, then you also don't need to delete them later.

tools/helpers.cmake
```diff
@@ -553,13 +553,22 @@ function(add_config_library prefix configure_template)
...
 
+    execute_process(
+        COMMAND
+            ${CMAKE_COMMAND} -E copy_if_different "${config_header_file}.temp"
+            "${config_header_file}"
+        COMMAND
+            ${CMAKE_COMMAND} -E copy_if_different "${config_json_file}.temp" "${config_json_file}"
+    )
+    file(REMOVE "${config_header_file}.temp" "${config_json_file}.temp")
+
```
Indanz, line 571: I'd add a short comment before the new execute_process to say what the purpose of that code is e.g. "Avoid rebuilding everything if nothing changed after reconfiguration".

Indanz (approved):
I noticed this behaviour recently and wondered where it was coming from.

tools/helpers.cmake
FennelFoxxo: Good point, but where might be the best place to put them? I've noticed that add_config_library is called from many different locations (sel4runtime, libsel4, util_libs, seL4_libs to name a few) so you have to be careful to make sure the path is still unique. I don't see any temp folders in the build path to use as a template, so what about creating a new directory at `${CMAKE_CURRENT_BINARY_DIR}/temp/gen_config/${prefix}/` and placing the temporary gen_config.h and gen_config.json inside there?

Alternatively, the diffing could be done in config_gen.py by accepting the output files in a+ mode and only writing if the contents differ? That would remove the need to create any temporary files.

Indanz: > Alternatively, the diffing could be done in config_gen.py by accepting the output files in a+ mode and only writing if the contents differ? That would remove the need to create any temporary files.

Can you store the output in a variable and compare that to the existing file? That would achieve the same in a hopefully simpler way. Looks like CMake supports reading a file into a variable and comparing variables with `if(var1 STREQUAL var2)`.

FennelFoxxo:
Had to tweak the python script because otherwise, the C header and json content was getting merged into the same stdout stream, so now you can optionally decide to generate only the header or only the json. Also added write_file_if_different as a helper function.

Indanz:
> Had to tweak the python script because otherwise, the C header and json content was getting merged into the same stdout stream, so now you can optionally decide to generate only the header or only the json. Also added write_file_if_different as a helper function.

Hm, what about doing it all in the python script and leaving the cmake file as it is? No need for a new option, it can do it always (though that might break builds that count on rebuilding everything after running the script).

FennelFoxxo:
Doing the diffing in the python script does make sense, although I personally think allowing optional arguments is useful - it's still backwards compatible, more extensible if additional generated formats are added in the future (so you can target only the format you require), and more flexible if you want to avoid creating files at all (to avoid mixing all the generated formats in stdout). I'd be happy to revert it but let me know what you think.

As for builds that rely on the script touching the output files, perhaps a --skip-unchanged option could be added that does the diffing (default would be to always overwrite for backwards compatibility with those builds)

FennelFoxxo:
Hey, just wondering if there's anything I need to do about this? If it's because of the failing DCO check, I was told last time to edit the DCO to use my name. Thanks!

tools/config_gen.py
```diff
@@ -21,36 +21,47 @@ def main():
         default=sys.stdin,
         help="Input YAML file.",
     )
+    parser.add_argument(
+        "--skip-unchanged",
+        dest="skip_unchanged",
+        action="store_true",
+        help="Skip write to output files if their content would not be changed"
```
Indanz, line 28:
```suggestion
        help="Only write output file if new content is different"
```

tools/config_gen.py
```diff
@@ -21,36 +21,47 @@ def main():
...
+    )
     parser.add_argument(
         "--write-c",
         metavar="OUT_FILE",
         dest="out_file_c",
-        type=argparse.FileType("w"),
-        default=sys.stdout,
+        nargs="?",
```
Indanz, line 34: Why this change?

tools/config_gen.py
```diff
@@ -72,11 +83,29 @@ def generate_c(config, out_file):
...
+
+def generate_json(config, out_file, skip_unchanged):
+    json_contents = json.dumps(config, indent=4)
+    write_file_lazy(out_file, json_contents, skip_unchanged)
+
 
+def write_file_lazy(out_file, content, skip_unchanged):
+    if out_file == sys.stdout:
+        out_file.write(content)
+        return
```
Indanz, line 97: Might be better to set `skip_unchanged` to false earlier if this is the case.

tools/config_gen.py
```diff
@@ -72,11 +83,29 @@ def generate_c(config, out_file):
...
 
-def generate_json(config, out_file):
-    json.dump(config, out_file, indent=4)
+    out_file.seek(0)
+    if skip_unchanged:
+        old_content = out_file.read()
+        if old_content != content:
```
Indanz, line 102: What about `if not skip_unchanged or out_file.read() != content:`?

That would avoid the code repetition below, at the cost of one unnecessary seek(0) in case skip_unchanged is false.

Indanz (approved):

Indanz:
> Hey, just wondering if there's anything I need to do about this? If it's because of the failing DCO check, I was told last time to edit the DCO to use my name. Thanks!

Sorry, forgot to review your new code. Other than some unimportant nitpicks, looks fine to me.

I can't merge if DCO fails, so Gerwin needs to do it.

Apologies for the rebase to bring it up-to-date with trunk while you're still busy.

tools/config_gen.py
FennelFoxxo: I asked about the optional arguments and got a thumbs up, so this change enables optional arguments (where --write-c/json is provided with no filename)
If --write-c is not provided, then default (None) is used and nothing is output
If --write-c is provided without a filename, then const (stdout) is used
If --write-c is provided with a filename, then the file handle is used
The `nargs="?"` enables the behavior of passing --write-c with either 0 or 1 arguments

tools/config_gen.py
FennelFoxxo: Well, `skip_unchanged` is global but whether `out_file == sys.stdout` can differ between the c and json output. It's possible that in some future use case, the json is written to stdout but the c is written to a file. In that case, the fact that `out_file == stdout` for the json case shouldn't set `skip_unchanged` to false because that would affect the lazy writing of the c file.

tools/config_gen.py
FennelFoxxo: Yep, I can do that! Good catch

FennelFoxxo:
Got it, I replied to your comments. Would it be better to leave my DCO message as-is so it's easier to merge? Personally I don't care too much about who the changes are attributed to, I just like seeing issues get fixed :3

tools/config_gen.py
Indanz: Thanks for explaining, I didn't realise this change.

Indanz:
> Would it be better to leave my DCO message as-is so it's easier to merge? Personally I don't care too much about who the changes are attributed to, I just like seeing issues get fixed :3

No, because it's not just about attribution, it's also about copyright ownership and keeping track of who made what changes.

The failing style check also prevents me from merging things anyway, so it doesn't matter.

FennelFoxxo:
Got it, thanks! I've never contributed to a large project before so I appreciate your explanation. I made the requested changes, hopefully it should be ok now

Indanz (approved):

lsf37:
Will have a look later. The DCO check on GitHub is a bit stricter than necessary, the main thing we need is a contactable email address from the author of the commit.

I'll have to do temporarily disable the style check since it is crashing. (Updating cmake-format is on my todo list, but it's a bit of an annoying procedure because it'll start complaining about formatting everywhere).

lsf37 (approved):

PR #1454, opened 2025-04-30
cleanup: remove obselete pageType and asidMax

### LINKED ISSUE #1447: types.h: unused asidMax, unused pageType
https://github.com/seL4/seL4/blob/fd14374e9dc5723ed6bdbc71f587980cc1b6286c/include/arch/arm/arch/api/types.h#L13-L19

Same for RISC-V and x86, the `asidMax` and `pageTypes` defines are never used anywhere. But we have `nASIDs` elsewhere which seems similar.

`asid_t` and `asidInvalid` are used elsewhere.

I can't tell why these files/defines exist, they predate the public release of seL4 code.

midnightveil:
Fixes #1447

### DIFF
--- a/include/arch/arm/arch/api/types.h
+++ b/include/arch/arm/arch/api/types.h
@@ -10,14 +10,8 @@
 #include <sel4/sel4_arch/objecttype.h>
 #include <sel4/arch/objecttype.h>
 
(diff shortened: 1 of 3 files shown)


lsf37 (approved):
👍

PR #1460, opened 2025-05-18
aarch32/vspace: use BIT and MASK macros

lsahn-gh:
Follow the style guide[1]

[1]: https://sel4.systems/Contribute/style.html#expressions

### DIFF
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -40,8 +40,8 @@
 
 /* helper stuff to avoid fencepost errors when
  * getting the last byte of a PTE or PDE */
-#define LAST_BYTE_PTE(PTE,LENGTH) ((word_t)&(PTE)[(LENGTH)-1] + (BIT(PTE_SIZE_BITS)-1))
-#define LAST_BYTE_PDE(PDE,LENGTH) ((word_t)&(PDE)[(LENGTH)-1] + (BIT(PDE_SIZE_BITS)-1))
+#define LAST_BYTE_PTE(PTE,LENGTH) ((word_t)&(PTE)[(LENGTH)-1] + MASK(PTE_SIZE_BITS))
+#define LAST_BYTE_PDE(PDE,LENGTH) ((word_t)&(PDE)[(LENGTH)-1] + MASK(PDE_SIZE_BITS))
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 /* Stage 2 */
@@ -277,7 +277,7 @@ BOOT_CODE void map_kernel_window(void)
         );
 
     /* now start initialising the page table */
-    memzero(armKSGlobalPT, 1 << seL4_PageTableBits);
+    memzero(armKSGlobalPT, BIT(seL4_PageTableBits));
 
     /* map vector table */
     map_kernel_frame(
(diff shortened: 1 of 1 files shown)


Indanz (commented):
Personally I'm not sure why using the BIT macro's would be better, `1 << n` is well defined, but `BIT(n)` could do anything, so you have to check the definition to be sure.

Anyway, if you do think using BIT is an improvement, you forgot the ones line 500, 2455, 2650 and 2652.

lsahn-gh:
> Personally I'm not sure why using the BIT macro's would be better, `1 << n` is well defined, but `BIT(n)` could do anything, so you have to check the definition to be sure.
> 
> Anyway, if you do think using BIT is an improvement, you forgot the ones line 500, 2455, 2650 and 2652.

Using the BIT() macro doesn’t offer any performance benefits, but it can help clarify the intention — that we are calculating any sizes by bit shifting.

What do you think about introducing a new macro (or using an existing one, if available), like SIZE_BIT(n), to improve readability or clear our intention for developers?

lsf37:
We should definitely not add yet another macro for the same thing. The [style guide](https://sel4.systems/Contribute/style.html#expressions) says to use `BIT`. Making the change is only worth it if there is no proof impact, but that does seem to be the case.

If you add the missing ones, I think we can merge this.

Indanz:
> it can help clarify the intention — that we are calculating any sizes by bit shifting.

But that was my point, it achieves the opposite: `1 << N` is standard C well understood by everyone, the `BIT()` macro is project specific and you have to at least once check what it does before you can understand the code.

> What do you think about introducing a new macro (or using an existing one, if available), like SIZE_BIT(n), to improve readability or clear our intention for developers?

That won't solve the problem, `SIZE_BIT(n)` is ambiguous too: It could mean the size of `n` in log2 or something. It would also confuse everyone who does know what the current `BIT()` does.

> If you add the missing ones, I think we can merge this.

Agreed.

lsahn-gh:
> > it can help clarify the intention — that we are calculating any sizes by bit shifting.
> 
> But that was my point, it achieves the opposite: `1 << N` is standard C well understood by everyone, the `BIT()` macro is project specific and you have to at least once check what it does before you can understand the code.

I understand your point, and I also agree with your concern.

That said, personally, I don’t mind using a macro like BIT(n) instead of 1 << n and I’m okay with checking its definition once.
In fact, I find code with multiple 1 << N or mixed with any conditions in branches even harder to read and follow.

Thank you and @lsf37 for the feedback.

I will change soon
BR

src/arch/arm/32/kernel/vspace.c
```diff
@@ -497,7 +497,7 @@ BOOT_CODE cap_t create_it_address_space(cap_t root_cnode_cap, v_region_t it_v_re
 
     /* create PD cap */
     copyGlobalMappings(PDE_PTR(rootserver.vspace));
-    cleanCacheRange_PoU(rootserver.vspace, rootserver.vspace + (1 << seL4_PageDirBits) - 1,
+    cleanCacheRange_PoU(rootserver.vspace, rootserver.vspace + BIT(seL4_PageDirBits) - 1,
```
Indanz, line 500:
```suggestion
    cleanCacheRange_PoU(rootserver.vspace, rootserver.vspace + MASK(seL4_PageDirBits),
```
here is also a MASK macro that does this.

Indanz (approved):

lsahn-gh:
I just modified the title of the commit message due to gitlint fail.

PR #1463, opened 2025-05-20
Fix cmake regex to work with qemu version 10

### LINKED ISSUE #1462: Build fails using QEMU v10
I'm running into problems building using qemu version 10:

```
sel4test/build-aarch64 $ ../init-build.sh -DPLATFORM=qemu-arm-virt -DSIMULATION=TRUE  -DCMAKE_TOOLCHAIN_FILE=~/gnu-aarch64.cmake
loading initial cache file /Users/seph/3rdparty/sel4test/projects/sel4test/settings.cmake
-- ARM_CPU not set, defaulting to cortex-a53
-- Extracting device tree from QEMU
-- qemu found at qemu-system-aarch64
CMake Error at /Users/seph/3rdparty/sel4test/kernel/src/plat/qemu-arm-virt/config.cmake:116 (message):
  Error: need at least QEMU version 3.1.0, found '0.0.0'
  /Users/seph/3rdparty/sel4test/kernel/configs/seL4Config.cmake:170 (include)
  /Users/seph/3rdparty/sel4test/kernel/FindseL4.cmake:21 (include)
  settings.cmake:32 (sel4_configure_platform_settings)


-- Configuring incomplete, errors occurred!
```

The problem is that its checking the qemu version, and the regex is failing:

```
 $ qemu-system-aarch64 -version                                                 
QEMU emulator version 10.0.0
```

The problem is [this regex](https://github.com/seL4/seL4/blob/349745c53b6aafaf1d53ace9d91cc93e477e15ca/src/plat/qemu-arm-virt/config.cmake#L107-L113)

This line:

    "[0-9](\\.[0-9])+"

should be:

    "[0-9]+(\\.[0-9]+)+"

Then the problem goes away.

josephg:
The build currently fails for arm / riscv using qemu version 10, because the cmake script incorrectly skips the first digit - so it thinks qemu 10.0.0 is actually version 0.0.0:

```
$ ../init-build.sh -DPLATFORM=qemu-arm-virt -DSIMULATION=TRUE  -DCMAKE_TOOLCHAIN_FILE=~/gnu-aarch64.cmake
loading initial cache file /Users/seph/3rdparty/sel4test/projects/sel4test/settings.cmake
-- ARM_CPU not set, defaulting to cortex-a53
-- Extracting device tree from QEMU
-- qemu found at qemu-system-aarch64
CMake Error at /Users/seph/3rdparty/sel4test/kernel/src/plat/qemu-arm-virt/config.cmake:116 (message):
  Error: need at least QEMU version 3.1.0, found '0.0.0'
```

See #1462 for more details.

This change to the version regex allows qemu version number fields to have multiple digits.

Fixes #1462.

### DIFF
--- a/src/plat/qemu-arm-virt/config.cmake
+++ b/src/plat/qemu-arm-virt/config.cmake
@@ -107,7 +107,7 @@ if(KernelPlatformQEMUArmVirt)
                 string(
                     REGEX
                         MATCH
-                        "[0-9](\\.[0-9])+"
+                        "[0-9]+(\\.[0-9]+)+"
                         QEMU_VERSION
                         "${QEMU_STDOUT_MESSAGE}"
                 )
--- a/src/plat/qemu-riscv-virt/config.cmake
+++ b/src/plat/qemu-riscv-virt/config.cmake
@@ -44,7 +44,7 @@ if(KernelPlatformQEMURiscVVirt)
                 string(
                     REGEX
                         MATCH
-                        "[0-9](\\.[0-9])+"
+                        "[0-9]+(\\.[0-9]+)+"
                         QEMU_VERSION
                         "${QEMU_STDOUT_MESSAGE}"
                 )


src/plat/qemu-arm-virt/config.cmake
```diff
@@ -107,7 +107,7 @@ if(KernelPlatformQEMUArmVirt)
                 string(
                     REGEX
                         MATCH
-                        "[0-9](\\.[0-9])+"
+                        "[0-9]+(\\.[0-9]+)+"
```
Indanz, line 110: The old regex would capture M.N version out of M.N.X, your new ones captures the full M.N.X. which is more than needed to handle qemu 10.n and isn't explained why it was done. `MIN_QEMU_VERSION` that we check against is "3.1.0", so this change is wanted, even though `VERSION_LESS` probably works either way. But please document all changes and the reason for them in the commit message.

Indanz (approved):

src/plat/qemu-arm-virt/config.cmake
josephg: Fair. It was a drive-by fix, and it seemed more strange to support XX.Y.Z but not XX.YY.ZZ. I'm happy to revert the secondary change or edit the commit message.

Indanz: Never mind, I misread the diff,  I thought you added the third `+` instead of the second one for some reason.

Indanz:
@lsf37, could you merge this when you have some time? I can't because of the failing style check.

PR #1464, opened 2025-05-21
arm,riscv: eliminate idle_thread function prologue

### LINKED ISSUE #519: make sure function prologue for idle_thread() is eliminated on Arm and RISC-V
We're currently relying on compiler optimisations to remove the function prologue for the idle thread. This does work correctly in the tests so far with the recommended tool chains, but with specific compiler version and flags, the optimisation seems not to happen and its absence [can lead to incorrect behaviour](https://sel4.discourse.group/t/kernel-crash-on-x86-but-not-on-x86-64/398).

PR #510 made this more robust for x86, we should do the same for arm and riscv.

midnightveil:
Create idle.S assembly files for each platform, as GCC seems to refuse
to obey `__attribute__((naked))` on AArch64 (GCC 13.2.0) and bails out;
so we can't use the same method as on x86.

Similar to #510 but for all other platforms. The idle_thread runs
without a stack and so cannot handle the stack prologue. This should
hopefully make the kernel rely less on FORCE_INLINE for this as well.

Fixes #519 

---

Before idle_thread (not identical compilers; from the microkit sdk)

```
board/maaxboard/release/elf/sel4.elf:   file format elf64-littleaarch64

Disassembly of section .text:

0000008040011250 <idle_thread>:
8040011250: d503207f    wfi
8040011254: 17ffffff    b       0x8040011250 <idle_thread>
8040011258: d503201f    nop
804001125c: d503201f    nop
```
```
board/maaxboard/debug/elf/sel4.elf:     file format elf64-littleaarch64

Disassembly of section .text:

00000080400113f0 <idle_thread>:
80400113f0: d503207f    wfi
80400113f4: 17ffffff    b       0x80400113f0 <idle_thread>
80400113f8: d503201f    nop
80400113fc: d503201f    nop
```
```
board/cheshire/debug/elf/sel4.elf:      file format elf64-littleriscv

Disassembly of section .text:

ffffffff80203b04 <idle_thread>:
ffffffff80203b04: 10500073      wfi
ffffffff80203b08: bff5  j       0xffffffff80203b04 <idle_thread>
```

After:

```
mmicrokit-sdk-2.0.1-dev/board/maaxboard/debug/elf/sel4.elf:      file format elf64-littleaarch64

Disassembly of section .text:

00000080400113e8 <idle_thread>:
80400113e8: d503207f    wfi
80400113ec: 17ffffff    b       0x80400113e8 <idle_thread>
```
```
microkit-sdk-2.0.1-dev/board/maaxboard/release/elf/sel4.elf:    file format elf64-littleaarch64

Disassembly of section .text:

000000804001125c <idle_thread>:
804001125c: d503207f    wfi
8040011260: 17ffffff    b       0x804001125c <idle_thread>
                ...
```
```
microkit-sdk-2.0.1-dev/board/cheshire/debug/elf/sel4.elf:       file format elf64-littleriscv

Disassembly of section .text:

ffffffff80203b04 <idle_thread>:
ffffffff80203b04: 10500073      wfi
ffffffff80203b08: bff5  j       0xffffffff80203b04 <idle_thread>
```
```
microkit-sdk-2.0.1-dev/board/cheshire/release/elf/sel4.elf:     file format elf64-littleriscv

Disassembly of section .text:

ffffffff80203a82 <idle_thread>:
ffffffff80203a82: 10500073      wfi
ffffffff80203a86: bff5  j       0xffffffff80203a82 <idle_thread>
```

There's negligible difference to the assembly (aside from the `nop` which I don't think should matter).

~~An issue that was brought up previously was that risc-v sets the stack pointer for the idle thread. I wasn't sure of the correctness of removing this because it seems to use the stack pointer in traps.S to determine the current core. (for SMP)~~ => Nevermind I forgot it loads from csscratch the sp so it should be fine to set it as 0 I think; I'll leave that for later.

### DIFF
--- a/src/arch/x86/idle.S
+++ b/src/arch/x86/idle.S
@@ -1,0 +1,15 @@
+/*
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <machine/assembler.h>
+
+.section .text, "ax"
+.code32
+
+BEGIN_FUNC(idle_thread)
+    1: hlt
+    jmp 1b
+END_FUNC(idle_thread)
--- a/include/arch/arm/armv/armv7-a/armv/machine.h
+++ b/include/arch/arm/armv/armv7-a/armv/machine.h
@@ -8,12 +8,6 @@
 
 #include <util.h>
 
-/* See idle_thread for an explanation as to why FORCE_INLINE is required here. */
-static inline void FORCE_INLINE wfi(void)
-{
-    asm volatile("wfi" ::: "memory");
-}
-
 static inline void dsb(void)
 {
     asm volatile("dsb" ::: "memory");
--- a/include/arch/arm/armv/armv8-a/64/armv/machine.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/machine.h
@@ -6,11 +6,6 @@
 
 #pragma once
 
-static inline void wfi(void)
-{
-    asm volatile("wfi" ::: "memory");
-}
-
 static inline void dsb(void)
 {
     asm volatile("dsb sy" ::: "memory");
@@ -42,4 +37,3 @@ static inline void isb(void)
 #define SYSTEM_READ_WORD(reg, v)  MRS(reg, v)
 #define SYSTEM_WRITE_64(reg, v)   MSR(reg, v)
 #define SYSTEM_READ_64(reg, v)    MRS(reg, v)
-
--- a/include/kernel/thread.h
+++ b/include/kernel/thread.h
@@ -165,7 +165,7 @@ void Arch_switchToIdleThread(void);
 void Arch_configureIdleThread(tcb_t *tcb);
 void Arch_activateIdleThread(tcb_t *tcb);
 
-void idle_thread(void);
+void NORETURN idle_thread(void);
 
 void configureIdleThread(tcb_t *tcb);
 void activateThread(void);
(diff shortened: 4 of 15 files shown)


src/arch/x86/idle.S
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <machine/assembler.h>
+
+.section .text, "ax"
+.code64
```
Indanz, line 10: This needs to be `.code32` if you want it to be common between 32-bit and 64-bit.

Indanz (approved):
Looks good to me. How does this affect verification?

Indanz:
> Sorry, not sure what relevance creating new files has with layout in memory? Like obviously it will affect it but so will the phase of the moon.

If you put it next to other assembly functions in one file you have more control. If it is just before code you know will be likely executed to handle traps, chances are high that that code will be in cache when the trap does happen.

E.g. for aarch64, you could squeeze it in the same cache line as `lower_el_irq`. That would avoid one cache miss when the wfi loop gets broken by an IRQ, reducing IRQ latency.

But like I said, there are many of these kind of optimisations that could be done, but haven't yet, so doing it now just for the idle task would be a bit weird.

midnightveil:
Proof test run: https://github.com/seL4/seL4/actions/runs/17057253139?pr=1464

Once again the only failure is the irrelevant RISCV MCS changes.

Looks like this had no proof impacts (as per previous message, I guess)

lsf37:
Yup, can confirm no verification impact.

PR #1465, opened 2025-05-21
Detect 'aarch64-elf-' toolchain

Ivan-Velickovic:
Already registered for RISC-V so makes sense to do the same for ARM.

### DIFF
--- a/gcc.cmake
+++ b/gcc.cmake
@@ -68,6 +68,7 @@ if("${CROSS_COMPILER_PREFIX}" STREQUAL "")
                 "aarch64-unknown-linux-gnu-"
                 "aarch64-none-linux-gnu-"
                 "aarch64-none-elf-"
+                "aarch64-elf-"
             )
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
Microkit seems to work with the toolchain, I can't test sel4test since I can't get it compiling because of https://github.com/seL4/seL4_libs/issues/97

Indanz (approved):

Indanz:
> Microkit seems to work with the toolchain, I can't test sel4test since I can't get it compiling because of [seL4/seL4_libs#97](https://github.com/seL4/seL4_libs/issues/97)

Surely newer releases have a more up-to-date binutils which doesn't mismatch and hence solves the problem?

Ivan-Velickovic:
See my comment here https://github.com/seL4/seL4_libs/issues/97#issuecomment-2897287205.

lsf37 (approved):

PR #1472, opened 2025-06-04
mcs: refactor decodeSchedControl_ConfigureFlags

michaelmcinerney:
Done in order to ensure overflow does not occur. More specifically, we call usToTicks on an argument only when it has been checked to be at most MAX_PERIOD_US, so that we know that overflow does not occur as a result of the conversion. We also rephrase the check on extra_refills. Rather than performing an addition, which may in extreme cases overflow, we perform a subtraction, which we know does not underflow, as a result of our invariants.

### DIFF
--- a/src/object/schedcontrol.c
+++ b/src/object/schedcontrol.c
@@ -92,22 +92,8 @@ static exception_t decodeSchedControl_ConfigureFlags(word_t length, cap_t cap, w
     }
 
     time_t budget_us = mode_parseTimeArg(0, buffer);
-    ticks_t budget_ticks = usToTicks(budget_us);
-    time_t period_us = mode_parseTimeArg(TIME_ARG_SIZE, buffer);
-    ticks_t period_ticks = usToTicks(period_us);
-    word_t extra_refills = getSyscallArg(TIME_ARG_SIZE * 2, buffer);
-    word_t badge = getSyscallArg(TIME_ARG_SIZE * 2 + 1, buffer);
-    word_t flags = getSyscallArg(TIME_ARG_SIZE * 2 + 2, buffer);
 
-    cap_t targetCap = current_extra_caps.excaprefs[0]->cap;
-    if (unlikely(cap_get_capType(targetCap) != cap_sched_context_cap)) {
-        userError("SchedControl_ConfigureFlags: target cap not a scheduling context cap");
-        current_syscall_error.type = seL4_InvalidCapability;
-        current_syscall_error.invalidCapNumber = 1;
-        return EXCEPTION_SYSCALL_ERROR;
-    }
-
-    if (budget_us > MAX_PERIOD_US || budget_ticks < MIN_BUDGET) {
+    if (budget_us > MAX_PERIOD_US) {
         userError("SchedControl_ConfigureFlags: budget out of range.");
         current_syscall_error.type = seL4_RangeError;
         current_syscall_error.rangeErrorMin = MIN_BUDGET_US;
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
No problems from my side.

src/object/schedcontrol.c
```diff
@@ -92,30 +92,38 @@ static exception_t decodeSchedControl_ConfigureFlags(word_t length, cap_t cap, w
     }
 
     time_t budget_us = mode_parseTimeArg(0, buffer);
+
+    if (budget_us > MAX_PERIOD_US) {
```
Indanz, line 96: Perhaps better to check `period_us` first and check for `budget_us > period_us` instead?

src/object/schedcontrol.c
```diff
@@ -92,30 +92,38 @@ static exception_t decodeSchedControl_ConfigureFlags(word_t length, cap_t cap, w
...
+    if (budget_ticks < MIN_BUDGET) {
         userError("SchedControl_ConfigureFlags: budget out of range.");
         current_syscall_error.type = seL4_RangeError;
         current_syscall_error.rangeErrorMin = MIN_BUDGET_US;
         current_syscall_error.rangeErrorMax = MAX_PERIOD_US;
         return EXCEPTION_SYSCALL_ERROR;
     }
 
-    if (period_us > MAX_PERIOD_US || period_ticks < MIN_BUDGET) {
+    if (period_ticks < MIN_BUDGET) {
```
Indanz, line 126: With a budget_us > period_us check this one isn't needed any more.

src/object/schedcontrol.c
```diff
@@ -131,16 +139,30 @@ static exception_t decodeSchedControl_ConfigureFlags(word_t length, cap_t cap, w
...
+    cap_t targetCap = current_extra_caps.excaprefs[0]->cap;
+    if (unlikely(cap_get_capType(targetCap) != cap_sched_context_cap)) {
+        userError("SchedControl_ConfigureFlags: target cap not a scheduling context cap");
+        current_syscall_error.type = seL4_InvalidCapability;
+        current_syscall_error.invalidCapNumber = 1;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    word_t extra_refills = getSyscallArg(TIME_ARG_SIZE * 2, buffer);
+    word_t max_refills = refill_absolute_max(targetCap);
```
Indanz, line 151: maybe introduce max_extra_refills and init it to max - MIN_REFILLS?

src/object/schedcontrol.c
michaelmcinerney: I think the idea here, which Gerwin can confirm, is that this `if` being here allows us to provide an error message for `period` which is independent of the value of `budget`.

src/object/schedcontrol.c
michaelmcinerney: I did initially introduce something similar, using the maximum size of an untyped, but this was more streamlined. We do have the proofs working with this formulation, so maybe our preference would be to keep this as is. Happy to change it if this idea is better somehow.

src/object/schedcontrol.c
lsf37: Yes, this was for better error reporting, which I thought we should prefer since this path isn't really performance critical anyway.

src/object/schedcontrol.c
lsf37: I think Indan means just locally, so you don't have the term `max_refills - MIN_REFILLS` in the code twice. Would be slightly nicer, but I think it's fine either way.

michaelmcinerney: I think I misread Indan's message and thought he was referring to a max int minus `MIN_REFILLS`. Sorry.

I see now. I don't feel very strongly about this, but my preference would be to leave it as it is: we already have many variables floating around here, so though it would save having to write `max_refills - MIN_REFILLS` twice, I think that is slightly better than having another similarly named variable for this. But as always, I'm happy to make this change if either of you feel more strongly about it.

lsf37: The main benefit would be readability (debatable) and removal of duplication. In the binary, I'm pretty sure the compiler is going to reuse the value in any case. So I'd be fine with leaving it.

src/object/schedcontrol.c
Indanz: How I look at it: Period must be <= MAX_PERIOD_US and budget must be >= MIN_BUDGET and <= period. Anything else follows from those requirements. Adding the same checks for the parameter that doesn't actually demand it is only making error reporting worse and the code unnecessarily verbose, because it's less clear what the actual requirements are.

src/object/schedcontrol.c
Indanz: Personally for C-only I would replace the `max_refills` with `max_extra_refills`, as `max_refills` isn't used elsewhere. But for verification that might be less nice, I don't know. No strong preferences, was just a suggestion.

michaelmcinerney:
Force pushed to rebase, so that I can test some things.

lsf37:
(rebased)

PR #1474, opened 2025-06-04
mcs: refactor finaliseCap to ease verification

michaelmcinerney:
This in particular introduces the function
schedContext_unbindReply, which is used within
finaliseCap, as well as invokeSchedContext_Unbind.

### DIFF
--- a/src/object/schedcontext.c
+++ b/src/object/schedcontext.c
@@ -143,10 +143,7 @@ static exception_t invokeSchedContext_Unbind(sched_context_t *sc)
 {
     schedContext_unbindAllTCBs(sc);
     schedContext_unbindNtfn(sc);
-    if (sc->scReply) {
-        sc->scReply->replyNext = call_stack_new(0, false);
-        sc->scReply = NULL;
-    }
+    schedContext_unbindReply(sc);
     return EXCEPTION_NONE;
 }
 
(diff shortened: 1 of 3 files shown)


michaelmcinerney:
Since this updates `invokeSchedContext_Unbind` as well, we should wait until the corresponding L4V PR for `finaliseCap_ccorres` is merged before merging this PR.

src/object/schedcontext.c
```diff
@@ -378,6 +375,25 @@ void schedContext_unbindNtfn(sched_context_t *sc)
     }
 }
 
+void schedContextMaybeUnbindNtfn(notification_t *ntfnPtr)
+{
+    sched_context_t *boundSchedContext;
+    boundSchedContext = (sched_context_t *)notification_ptr_get_ntfnSchedContext(ntfnPtr);
```
Indanz: Any reason why `SC_PTR` isn't used here?

And I'd prefer `sc` or boundSC, which would shorten the line perhaps enough to fit it on one line.

Indanz (approved):
Looks correct.

src/object/schedcontext.c
michaelmcinerney: No reason. I might have just copied this style from somewhere else. Maybe I'll go with

```
sched_context_t *sc = SC_PTR(notification_ptr_get_ntfnSchedContext(ntfn));
```

This lines occurs exactly in `maybeReturnSchedContext`, actually, so I guess we have prior art

lsf37: We should prefer the macro (`SC_PTR`) when we can.

michaelmcinerney:
I have updated the PR following Indan's suggestion, and have rebased, too.

PR #1476, opened 2025-06-05
boot: use explicit atomics and memory order for node_boot_lock and ksNumCPUs

ganboing:
Use __atomic_op(...) with explicit memory order for clarity. Avoid using
volatile keyword and arch-dependent fences. It's now easier to reason
about and the race-condition involving ksNumCPUs is fixed.


midnightveil:
What race condition does this fix?

At a glance it seems fine, _Atomic variables should give the necessary ordering semantics.

(To be clear, I'm not saying there isn't a race condition here, given I haven't looked at this in detail, what I'm wondering if what the execution trace which leads to a bug?)

src/arch/arm/kernel/boot.c
```diff
@@ -319,12 +319,11 @@ BOOT_CODE static void release_secondary_cpus(void)
 #endif
 
     /* Wait until all the secondary cores are done initialising */
-    while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
+    while (__atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL) != CONFIG_MAX_NUM_NODES) {
```
midnightveil: Why an atomic fetch add of 0 as opposed to an atomic load?

src/arch/arm/kernel/boot.c
```diff
@@ -319,12 +319,11 @@ BOOT_CODE static void release_secondary_cpus(void)
...
+        /* perform a memory acquire to get new values of ksNumCPUs, release for ksCurTime */
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
         NODE_STATE(ksCurTime) = getCurrentTime();
 #endif
-        /* perform a memory acquire to get new values of ksNumCPUs, release for ksCurTime */
```
midnightveil, line 326: What about the fence for ksCurTime  release? Where is that now?

src/arch/riscv/kernel/boot.c
```diff
@@ -153,37 +153,34 @@ BOOT_CODE static void init_plat(void)
...
+    word_t locked = __atomic_exchange_n(&node_boot_lock, 1, __ATOMIC_RELEASE);
+    /* Sanity check for a proper lock state. */
+    assert(!locked);
     /* At this point in time the primary core (executing this code) already uses
      * the seL4 MMU/cache setup. However, the secondary cores are still using
      * the elfloader's MMU/cache setup, and thus the update of node_boot_lock
      * may not be visible there if the setups differ. Currently, the mappings
      * match, so a barrier is all that is needed.
      */
-    fence_rw_rw();
```
midnightveil, line 180: What happened to all the cache/mmu state that this fence was for?

ganboing:
> What race condition does this fix?
> 
> At a glance it seems fine, _Atomic variables should give the necessary ordering semantics.
> 
> (To be clear, I'm not saying there isn't a race condition here, given I haven't looked at this in detail, what I'm wondering if what the execution trace which leads to a bug?)

Because the memory ordering is unclear. Also the use of volatile is just wrong. Now with explicit __atomic_op(...) and explicit memory ordering, is possible to reason and proof the code. The race-condition fixed is around `ksNumCPUs++`, which is definitely not atomic.

src/arch/riscv/kernel/boot.c
ganboing: The subsequent `while (__atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL) != CONFIG_MAX_NUM_NODES)` gives both acquire/release semantics. This synchronizes with secondary CPUs `__atomic_fetch_add(&ksNumCPUs, 1, __ATOMIC_RELEASE)`

src/arch/arm/kernel/boot.c
ganboing: It's merged into the above `__atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL)`

src/arch/arm/kernel/boot.c
ganboing: Because load can't provide `__ATOMIC_RELEASE`

ganboing: With this, the latter `__atomic_thread_fence(__ATOMIC_ACQ_REL)` is not needed

ganboing:
> What race condition does this fix?
> 
> At a glance it seems fine, _Atomic variables should give the necessary ordering semantics.
> 
> (To be clear, I'm not saying there isn't a race condition here, given I haven't looked at this in detail, what I'm wondering if what the execution trace which leads to a bug?)

The race-confition

ganboing:
Accidentally closed my PR. Reopened it. Sorry.

midnightveil:
Apologies, I somehow missed that some of the accesses were to `ksNumCPUs` which is not the same as `node_boot_lock.` This makes a bit more sense now.

src/arch/arm/kernel/boot.c
midnightveil: OK, that's a little unintuitive, but seems correct.

src/arch/arm/kernel/boot.c
midnightveil: OK, yes. This is doing 2 things in one.

src/arch/riscv/kernel/boot.c
midnightveil: I'm explicitly referring to the comment above (which probably lies outside the C memory model):

> However, the secondary cores are still using
> the elfloader's MMU/cache setup, and thus the update of node_boot_lock
> may not be visible there if the setups differ.

KurtWu10:
My main concern is whether atomic instructions provide necessary synchronisations for system features, including instruction caches and MMUs. In the case of AArch64, it is clear from the [model](https://github.com/herd/herdtools7/blob/7.58/herd/libdir/aarch64hwreqs.cat#L114-L133) that dmb barriers provide stronger guarantees than pure RMWs.

ganboing:
> My main concern is whether atomic instructions provide necessary synchronisations for system features, including instruction caches and MMUs. In the case of AArch64, it is clear from the [model](https://github.com/herd/herdtools7/blob/7.58/herd/libdir/aarch64hwreqs.cat#L114-L133) that dmb barriers provide stronger guarantees than pure RMWs.

I totally understand your concern. I was looking at herd7 for a while (for riscv), and yes to properly prove correctness, we definitely need to have much more reasoning about it. This commit is not intended to solve all the issues, but focus on fixing something very obvious. I.e., `ksNumCPUs` is not atomic, thus there's a race-condition on it when updating on secondary CPUs, not to mention the fact that there's no release/acquire semantics being conveyed through it. It's entirely possible that we need more synchronization/barriers apart from this. This patch ensures the memory ordering is no less stricter then before, if not more. Thus, I'm not introducing new bugs.

KurtWu10:
In this case, the removal of `__atomic_thread_fence(__ATOMIC_ACQ_REL);` could be an issue, I guess? For herd7 on RISC-V, it only talks about user-level features but not system features, so I am not sure about that.

src/arch/riscv/kernel/boot.c
ganboing: Here, `fence rw,rw` is achieved by the `__ATOMIC_ACQ_REL` memory order of `__atomic_fetch_add(&ksNumCPUs...`.  With `__ATOMIC_ACQ_REL`, it's going to be a amo.aqrl or similar. It ensures previous memory accesses can't "leak through" the __atomic_fetch_add, and subsequent memory accesses can't "move up" before the __atomic_fetch_add. The comment is actually confusing. I can rework that too.

ganboing:
> In this case, the removal of `__atomic_thread_fence(__ATOMIC_ACQ_REL);` could be an issue, I guess? For herd7 on RISC-V, it only talks about user-level features but not system features, so I am not sure about that.

The `__atomic_thread_fence(__ATOMIC_ACQ_REL)` is merged with the `__atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL)` above. That's why I'm using __atomic_fetch_add with 0. midnightveil actually asked the same question. Yes, there's no system features such as MMIO being modeled in herd7. Also, the official spec of RVWMO doesn't define the formal model involving MMIO/ifence/sfence.vma.... Quote: 
```
The interaction of the memory model with I/O memory, instruction fetches, FENCE.I, 
page-table walks, and SFENCE.VMA is not (yet) formalized. 
Some or all of the above may be formalized in a future revision of this specification
```

Thus, it's not trivial to reason. I find it helpful to first tie the special memory accesses into the local CPU's memory order, then the global memory order (if there's one. Thankfully ARMv8/RISCV have a global memory order because they are both multicopy-atomic). Again, the change only makes the memory order stricter, not less. It's totally possible that it's not strict enough. How we should definitively prove all these concurrency primitives would require a much broader discussion.

KurtWu10:
> The __atomic_thread_fence(__ATOMIC_ACQ_REL) is merged with the __atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL) above.

This sentence is exactly the sentence that I am concerning. In the case of Armv8, a `dmb` fence (but not ll/sc or rmw) is required to preserve the thread-local order w.r.t. many system features, as indicated in my previous link. Converting a dmb fence to a rmw weakens the ordering.

Multi-copy atomicity is not a concern here because the dmb fence can still provide the ordering for the outdated non-MCA Armv8.

For a realistic (but unrelated) example to demonstrate the issue, consider the [ISA2.F+dc-dmb+dsb-ic-dsb+ctrl-isb](https://github.com/KurtWu10/WMM/blob/766e3c3bd678059214125d350703da55c300540a/aarch64/ifetch/ISA2.F%2Bdc-dmb%2Bdsb-ic-dsb%2Bctrl-isb.litmus) litmus test for software thread migration in [Ben Simner's thesis](https://2plus2a.com/files/drafts/top-thesis-draft-2025-03-28-clean.pdf). Removing the `dmb sy` instruction (which is roughly equivalent to a SC fence) and converting the second `str` instruction in thread 0 to `stlr` (which is roughly equivalent to a SC store) allows the execution, because the `stlr` is weaker than `dmb` for the data cache maintenance (`dc`).

ganboing:
> > The __atomic_thread_fence(__ATOMIC_ACQ_REL) is merged with the __atomic_fetch_add(&ksNumCPUs, 0, __ATOMIC_ACQ_REL) above.
> 
> This sentence is exactly the sentence that I am concerning. In the case of Armv8, a `dmb` fence (but not ll/sc or rmw) is required to preserve the thread-local order w.r.t. many system features, as indicated in my previous link. Converting a dmb fence to a rmw weakens the ordering.
> 
> Multi-copy atomicity is not a concern here because the dmb fence can still provide the ordering for the outdated non-MCA Armv8.
> 
> For a realistic (but unrelated) example to demonstrate the issue, consider the [ISA2.F+dc-dmb+dsb-ic-dsb+ctrl-isb](https://github.com/KurtWu10/WMM/blob/766e3c3bd678059214125d350703da55c300540a/aarch64/ifetch/ISA2.F%2Bdc-dmb%2Bdsb-ic-dsb%2Bctrl-isb.litmus) litmus test for software thread migration in [Ben Simner's thesis](https://2plus2a.com/files/drafts/top-thesis-draft-2025-03-28-clean.pdf). Removing the `dmb sy` instruction (which is roughly equivalent to a SC fence) and converting the second `str` instruction in thread 0 to `stlr` (which is roughly equivalent to a SC store) allows the execution, because the `stlr` is weaker than `dmb` for the data cache maintenance (`dc`).

I think I've got your point now. Looks like the patch didn't weaken the memory order in C language level, but accidentally weakened it in asm level. In this case explicit `dmb` is probably required, as the language and compiler has no knowledge of system feature interactions. BTW, I just tried a simple program with `__atomic_thread_fence(__ATOMIC_ACQ_REL)`, and gcc 14.2 translates it to a `dmb ish`. I didn't read ARM spec well enough, but I suspect whether `ish` is strict enough, compared to `sy`. Even if true, the original code is still wrong, though, as there's no fence/atomic... in secondary CPUs to synchronize with this `__atomic_thread_fence` I can drop all the ARM code changes and focus on RISCV only. Let me know if any thoughts. Thanks.

KurtWu10:
> I didn't read ARM spec well enough, but I suspect whether `ish` is strict enough, compared to `sy`.

This is a known [limitation](https://github.com/herd/herdtools7/blob/7.58/herd/libdir/aarch64fences.cat#L47-L49) of the model. Other parts are fine for me.

ganboing:
Close this. Updated patchset incoming.

PR #1478, opened 2025-06-05
cleanup: use rootSlot to give a name to excapref0

midnightveil:
Non-functional change. This will probably affect verification, so definitely not expecting this to be merged anytime soon.

But it was very weird we had this rootSlot argument mark as UNUSED, and it took me a bit to realise exactly what excaprefs[0] was supposed to be.

### DIFF
--- a/src/object/untyped.c
+++ b/src/object/untyped.c
@@ -28,7 +28,7 @@ exception_t decodeUntypedInvocation(word_t invLabel, word_t length, cte_t *slot,
 {
     word_t newType, userObjSize, nodeIndex;
     word_t nodeDepth, nodeOffset, nodeWindow;
-    cte_t *rootSlot UNUSED;
+    cte_t *rootSlot;
     exception_t status;
     cap_t nodeCap;
     lookupSlot_ret_t lu_ret;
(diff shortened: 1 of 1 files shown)


lsf37:
Might actually be something that comes from verification, because the rule for extracaps expects a certain term pattern. Let me run the proof test.

midnightveil:
That proof failure seems unrelated (?). I saw a similar failure on the BIT macro PR.

127 is normally "not found in path". Is z3 not installed? (?)

lsf37:
Right, those failures are unrelated. Looks like the proof test in this repo is not using the ts branch of Isabelle, but the vanilla version. I recently change the AWS VMs over to Arm, because the x86 VMs were always getting cancelled. z3 needs to be a more recent version to work on Arm, which I have added on the ts branch.

Long story short, I need to do something in the setup first to fix the C proof tests. We also have the Isabelle2025 update in the queue, which would fix all this, but we might not want to wait for that.

lsf37:
#1481 should fix it.

midnightveil:
Huh, somehow the proofs passed.

Maybe that makes sense given the Haskell spec and the Isabelle spec both match the nicer C code, but I'm surprised it didn't break the link at all.

Anyhow, yay :)

lsf37:
> Huh, somehow the proofs passed.
> 
> Maybe that makes sense given the Haskell spec and the Isabelle spec both match the nicer C code, but I'm surprised it didn't break the link at all.

There is a bunch of automated rewriting at the beginning of most C proofs (not all) and also symbolic execution that tracks the content on local variables, and the combination of those might be doing what was needed here by accident already. It's a bit better than I would have expected, but I won't complain..

> Anyhow, yay :)

:-)

PR #1480, opened 2025-06-05
boot: use explicit atomics and memory ordering

ganboing:
Use __atomic_(...) with explicit memory order for clarity; Avoid using
volatile keyword and arch-dependent fences, and fix race-condition
around ksNumCPUs. Specifically:

a. node_boot_lock in arm/riscv is still atomic, but w/ explicit memory
   order for load/store. Per the C language spec, load/store, --/++,
   compound assignment on _Atomic variable use memory_order_seq_cst
   memory ordering. Thus, use equivalent and explicit memory order w/
   built-in atomic functions for clarity. Whether the memory ordering
   can be relaxed to acquire-release is left for future discussion.

b. ksNumCPUs in arm/riscv is now atomic to avoid race and have proper
   memory order specified. It's in reverse to node_boot_lock: BSP will
   wait and observe all APs coming out of initialization and all the
   side-effects. Note that APs do not rendezvous (although we can make
   them do). This is the same as before. The memory order chosen is
   acquire-release. Whether this is enough needs further reasoning, but
   it's at least much stricter than before, where we didn't have any
   memory ordering.

c. ksCurTime in all arch is now atomic w/ relaxed ordering in load/store
   This is because the read side in clock_sync_test is reading it purely
   for time check. There's no message passing or synchronization. Thus,
   given that C and all HW arch already provide coherence ordering for
   single atomic variable, it can be used with relaxed ordering. The
   previous use of acquire/release fences doesn't do anything, because,
   again, we are not synchronizing through ksCurTime + fence. I assume
   the original intent is to publish/grab the new values of ksCurTime
   as quickly as we can, but fence doesn't provide that functionality
   and actually we have no way of expediting the process.

Note:
1. For arm, the __atomic_thread_fence(__ATOMIC_ACQ_REL) at line 327 of
   src/arch/riscv/kernel/boot.c is kept as-is. After a discussion with
   Kurt Wu, this fence might accidentally provide more memory barriers
   than what the comment says. E.g., it lowers to a "dmb ish" on arm64
   with GCC 14.2. I'm leaving it there for further reasoning.

2. For x86, there's potential data race with smp_aps_index (volatile),
   and perhaps memory ordering issues. These issues could be masked by
   the fact the x86 has a very strict memory model. I leave the cleanup
   and reasoning of x86/boot code to future discussion and patches.

3. For riscv, GCC generates fences that also fence I/O. Refer to:
   "Use more conservative fences on RISC-V", Link:
   https://github.com/riscvarchive/riscv-gcc/pull/55
   Whether this is going to change (relaxed) in the future is unknown.
   Also this is a implementation (GCC) specific behavior. We should
   consider adding explicit I/O fences for correctness.

Signed-off-by: Bo Gan <ganboing@gmail.com>

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -34,7 +34,7 @@
  * spinning until the primary core has initialized all kernel structures and
  * then set it to 1.
  */
-BOOT_BSS static volatile _Atomic int node_boot_lock;
+BOOT_BSS static int node_boot_lock;
 #endif /* ENABLE_SMP_SUPPORT */
 
 BOOT_BSS static region_t reserved[NUM_RESERVED_REGIONS];
@@ -269,7 +269,7 @@ BOOT_CODE static void init_plat(void)
 BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 {
     /* need to first wait until some kernel init has been done */
-    while (!node_boot_lock);
+    while (!__atomic_load_n(&node_boot_lock, __ATOMIC_SEQ_CST));
 
     /* Perform cpu init */
     init_cpu();
@@ -288,7 +288,7 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
     NODE_LOCK_SYS;
 
     clock_sync_test();
-    ksNumCPUs++;
+    __atomic_fetch_add(&ksNumCPUs, 1, __ATOMIC_RELEASE);
 
     init_core_state(SchedulerAction_ResumeCurrentThread);
 
@@ -298,8 +298,8 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 BOOT_CODE static void release_secondary_cpus(void)
 {
     /* release the cpus at the same time */
-    assert(0 == node_boot_lock); /* Sanity check for a proper lock state. */
-    node_boot_lock = 1;
+    int locked = __atomic_exchange_n(&node_boot_lock, 1, __ATOMIC_SEQ_CST);
+    assert(!locked); /* Sanity check for a proper lock state. */
 
     /*
      * At this point in time the primary core (executing this code) already uses
@@ -318,12 +318,16 @@ BOOT_CODE static void release_secondary_cpus(void)
     plat_cleanInvalidateL2Cache();
 #endif
 
-    /* Wait until all the secondary cores are done initialising */
-    while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
+    /* Wait until all the secondary cores are done initialising
+     * The load-acquire on ksNumCPUs is paired with secondary cores' store-relese
+     * to ensure they've all finished initialization, and we (BSP) observes all
+     * side-effects from AP.
+     */
+    while (__atomic_load_n(&ksNumCPUs, __ATOMIC_ACQUIRE) != CONFIG_MAX_NUM_NODES) {
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
-        NODE_STATE(ksCurTime) = getCurrentTime();
+        __atomic_store_n(NODE_STATE(ksCurTime), getCurrentTime(), __ATOMIC_RELAXED);
 #endif
-        /* perform a memory acquire to get new values of ksNumCPUs, release for ksCurTime */
+        /* XXX Do we still need this? */
         __atomic_thread_fence(__ATOMIC_ACQ_REL);
     }
 }
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -27,7 +27,7 @@
  * spinning until the primary core has initialized all kernel structures and
  * then set it to 1.
  */
-BOOT_BSS static volatile _Atomic word_t node_boot_lock;
+BOOT_BSS static word_t node_boot_lock;
 #endif
 
 BOOT_BSS static region_t res_reg[NUM_RESERVED_REGIONS];
@@ -153,15 +153,13 @@ BOOT_CODE static void init_plat(void)
 #ifdef ENABLE_SMP_SUPPORT
 BOOT_CODE static bool_t try_init_kernel_secondary_core(word_t hart_id, word_t core_id)
 {
-    while (!node_boot_lock);
-
-    fence_r_rw();
+    while (!__atomic_load_n(&node_boot_lock, __ATOMIC_SEQ_CST));
 
     init_cpu();
     NODE_LOCK_SYS;
 
     clock_sync_test();
-    ksNumCPUs++;
+    __atomic_fetch_add(&ksNumCPUs, 1, __ATOMIC_RELEASE);
     init_core_state(SchedulerAction_ResumeCurrentThread);
     ifence_local();
     return true;
@@ -169,21 +167,25 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(word_t hart_id, word_t co
 
 BOOT_CODE static void release_secondary_cores(void)
 {
-    assert(0 == node_boot_lock); /* Sanity check for a proper lock state. */
-    node_boot_lock = 1;
+    word_t locked = __atomic_exchange_n(&node_boot_lock, 1, __ATOMIC_SEQ_CST);
+    /* Sanity check for a proper lock state. */
+    assert(!locked);
     /* At this point in time the primary core (executing this code) already uses
      * the seL4 MMU/cache setup. However, the secondary cores are still using
      * the elfloader's MMU/cache setup, and thus the update of node_boot_lock
      * may not be visible there if the setups differ. Currently, the mappings
-     * match, so a barrier is all that is needed.
+     * match, and a atomic store-release is used to synchronize with secondary
+     * CPUs' load-acquire to publish all preceding memory writes.
      */
-    fence_rw_rw();
 
-    while (ksNumCPUs != CONFIG_MAX_NUM_NODES) {
+    /* The load-acquire on ksNumCPUs is paired with secondary cores' store-relese
+     * to ensure they've all finished initialization, and we (BSP) observes all
+     * side-effects.
+     */
+    while (__atomic_load_n(&ksNumCPUs, __ATOMIC_ACQUIRE) != CONFIG_MAX_NUM_NODES) {
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
-        NODE_STATE(ksCurTime) = getCurrentTime();
+        __atomic_store_n(&NODE_STATE(ksCurTime), getCurrentTime(), __ATOMIC_RELAXED);
 #endif
-        __atomic_thread_fence(__ATOMIC_ACQ_REL);
     }
 }
 #endif /* ENABLE_SMP_SUPPORT */
--- a/src/arch/x86/kernel/smp_sys.c
+++ b/src/arch/x86/kernel/smp_sys.c
@@ -67,8 +67,7 @@ BOOT_CODE void start_boot_aps(void)
         /* wait for current AP to boot up */
         while (smp_aps_index == current_ap_index) {
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
-            NODE_STATE(ksCurTime) = getCurrentTime();
-            __atomic_thread_fence(__ATOMIC_ACQ_REL);
+            __atomic_store_n(&NODE_STATE(ksCurTime), getCurrentTime(), __ATOMIC_RELAXED);
 #endif
         }
     }
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -575,11 +575,14 @@ BOOT_CODE void clock_sync_test(void)
     ticks_t margin = usToTicks(CLOCK_SYNC_DELTA) + getTimerPrecision();
 
     assert(getCurrentCPUIndex() != 0);
-    t = NODE_STATE_ON_CORE(ksCurTime, 0);
+    /* Use relaxed load, as we are not synchronizing or message-passing through
+     * ksCurTime, and there's coherence order guarantee on same atomic variable.
+     * Therefore, we'll observe a monotonically increasing of NODE 0's ksCurTime.
+     * (Not going back)
+     */
+    t = __atomic_load_n(&NODE_STATE_ON_CORE(ksCurTime, 0), __ATOMIC_RELAXED);
     do {
-        /* perform a memory acquire to get new values of ksCurTime */
-        __atomic_thread_fence(__ATOMIC_ACQUIRE);
-        t0 = NODE_STATE_ON_CORE(ksCurTime, 0);
+        t0 = __atomic_load_n(&NODE_STATE_ON_CORE(ksCurTime, 0), __ATOMIC_RELAXED);
     } while (t0 == t);
     t = getCurrentTime();
     printf("clock_sync_test[%d]: t0 = %"PRIu64", t = %"PRIu64", td = %"PRIi64"\n",


ganboing:
@midnightveil  @KurtWu10 @Ivan-Velickovic Please help me review my 2nd attempt. Thanks.

KurtWu10:
(I am not familiar with the kernel and I am only writing my own thinking)

1.
> Per the C language spec, simple assignment and read does not provide acquire, release or sc semantics.

The original `node_boot_lock` is labelled by the `_Atomic` specifier, which says that it is an atomic type (section 6.7.2.4 of N1570 (of the C11 spec)). The ordering provided by the original code is SC, according to clause 9 of section 6.2.6.1:
> Loads and stores of objects with atomic types are done with `memory_order_seq_cst` semantics.

So the ordering in the newer code is indeed weaker than the original code.

2. The added atomic exchange in `release_secondary_cpus()` may slow down the kernel compiled in release mode. Whether this is a concern is beyond my scope.

And c is a good catch!

ganboing:
> (I am not familiar with the kernel and I am only writing my own thinking)
> 
> 
> > Per the C language spec, simple assignment and read does not provide acquire, release or sc semantics.
> 
> The original `node_boot_lock` is labelled by the `_Atomic` specifier, which says that it is an atomic type (section 6.7.2.4 of N1570 (of the C11 spec)). The ordering provided by the original code is SC, according to clause 9 of section 6.2.6.1:
> 
> > Loads and stores of objects with atomic types are done with `memory_order_seq_cst` semantics.
> 
> So the ordering in the newer code is indeed weaker than the original code.
> 
> 2. The added atomic exchange in `release_secondary_cpus()` may slow down the kernel compiled in release mode. Whether this is a concern is beyond my scope.
> 
> And c is a good catch!

Patch updated.

It Addressed comment 1. Sorry I was using cppreference, and the wording there isn't clear. Yes, you are right, _Atomic has the implicit memory_order_seq_cst built-in.

For 2, there may be some micro-arch level slow-down possible? I think it should be negligible compared with HW initialization and other stuff happening during boot.

I've also added Note .3 for riscv GCC's fence behavior.

ganboing:
@Indanz

src/arch/arm/kernel/boot.c
```diff
@@ -288,7 +288,7 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
     NODE_LOCK_SYS;
 
     clock_sync_test();
-    ksNumCPUs++;
+    __atomic_fetch_add(&ksNumCPUs, 1, __ATOMIC_RELEASE);
```
Indanz, line 291: This is done with the kernel lock taken, that's the `NODE_LOCK_SYS` above, so no further synchronisation is necessary, as we are the only writer. If it is, we should fix the global lock code, not work around it elsewhere.

That said, it doesn't hurt updating `ksNumCPUs` atomically, but there is no reason to. So doing it atomically will be confusing, as it implies it's needed. Either way, you should add a comment why the update is done atomically while within a critical section (either for consistency or reason X).

Strictly speaking doing the update atomically here will let core 0 know about it sooner, now it's only guaranteed to be seen when this core unlocks the global lock. But that is fine, because the first thing core 0 does after seeing ksNumCPUs being correct, is taking the global lock itself. The only reason why we have ksNumCPUs is to give other cores a chance to take the global lock.

src/arch/arm/kernel/boot.c
```diff
@@ -318,12 +318,16 @@ BOOT_CODE static void release_secondary_cpus(void)
...
+     */
+    while (__atomic_load_n(&ksNumCPUs, __ATOMIC_ACQUIRE) != CONFIG_MAX_NUM_NODES) {
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
-        NODE_STATE(ksCurTime) = getCurrentTime();
+        __atomic_store_n(NODE_STATE(ksCurTime), getCurrentTime(), __ATOMIC_RELAXED);
 #endif
-        /* perform a memory acquire to get new values of ksNumCPUs, release for ksCurTime */
+        /* XXX Do we still need this? */
         __atomic_thread_fence(__ATOMIC_ACQ_REL);
     }
```
Indanz, line 332: Agreed that doing an atomic load would be nicer than the thread fence, but as you've noticed, then you also need to do the ksCurTime explicitly atomically, so overall the code doesn't improve. As my SMP patches were already big,  I tried to minimise unnecessary code changes in the boot code.

Anyway, in your changes you lost release semantics (assuming you remove the fence, but if you don't, your other changes don't make sense), which is less strict than the current code, so you're weakening the memory ordering. I agree that that's probably fine, but it's also unnecessary. If you follow your reasoning, you could do the ksNumCPUs check also with relaxed memory ordering instead of acquire, so you're not being consistent.

All in all, you are making the memory ordering more complex, not simpler, so I don't like it. If you want stricter memory ordering on ksNumCPUs, then mark it as `_Atomic` and remove all redundant fence code. That actually would simplify the code and memory ordering.

src/arch/arm/kernel/boot.c
```diff
@@ -269,7 +269,7 @@ BOOT_CODE static void init_plat(void)
 BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 {
     /* need to first wait until some kernel init has been done */
-    while (!node_boot_lock);
+    while (!__atomic_load_n(&node_boot_lock, __ATOMIC_SEQ_CST));
```
Indanz, line 272: So for the one use case where C makes it easy to let the compiler generate the atomic code, you don't want to do it. I don't think this improves the code, it makes it less readable.

Perhaps we should remove `volatile` from node_boot_lock` to make it harder to miss the `_Atomic`.

src/arch/arm/kernel/boot.c
```diff
@@ -298,8 +298,8 @@ BOOT_CODE static bool_t try_init_kernel_secondary_core(void)
 BOOT_CODE static void release_secondary_cpus(void)
 {
     /* release the cpus at the same time */
-    assert(0 == node_boot_lock); /* Sanity check for a proper lock state. */
-    node_boot_lock = 1;
+    int locked = __atomic_exchange_n(&node_boot_lock, 1, __ATOMIC_SEQ_CST);
```
Indanz, line 301: We are the only writer, so an exchange is not necessary.

The assert checks that BSS gets zeroed properly, because if it wasn't debugging this is very difficult. It doesn't check locking correctness.

ganboing:
@Indanz 
Agree with a (`node_boot_lock`) -- I'll keep it _Atomic and simple, and remove volatile.

Partially disagree with b and disagree with c. See my explanation below. 

For b, I overlooked the `NODE_LOCK_SYS`. Yes, you are right, there's no data-race on `ksNumCPUs`, but only for all APs. For BSP, it's still being accessed in a racy way, because BSP is not in CS. Thus, I would argue it's better to use relaxed atomics, and it doesn't have to be RMW. I know in practice that compiler shouldn't do something crazy, such as on the AP side, doing `ksNumCPUs += 2`, and then `ksNumCPUs--` to achieve `ksNumCPUs++`, but still making it relaxed atomics guarantees that compiler only touches the variable when we instruct it to do so. I remember doing research during my M.S. and we were using thread sanitizer. This kind of pattern -- another thread reading plain variable outside of CS is for sure going to trigger a false-positive report in thread sanitizer. Relaxed atomics can serve as annotation for ASAN and tools alike to indicate we indeed want to do it in a racy way, and perhaps also ease the proof for concurrency correctness.

For c, the use of fence and acq_rel actually does nothing in the original code, because there's no atomics or fences on the AP side to pair with. Thus, there's actually no memory order in the original code. I.e., ksCurTime is not being read/written as atomic, and there's no fence on the reader side. I know your intent, but the fences and acq_rel doesn't help, and they are actually not needed. From cppreference's text about _Atomics:

```
Each atomic object has its own associated modification order, which is a total order of 
modifications made to that object. If, from some thread's point of view, modification A of
some atomic M [happens-before](https://en.cppreference.com/w/c/atomic/memory_order.html)
modification B of the same atomic M, then in the modification order of M, A occurs before B.

Note that although each atomic object has its own modification order, there is no single total
order; different threads may observe modifications to different atomic objects in different orders.
```
Then it talks about `read-read` `read-write` `write-read` and `write-write` coherence order. This should be also in the spec, see https://www.open-std.org/jtc1/sc22/wg14/issues/c11c17/issue0402.html Thus compiler will not combine multiple read or multiple write on the same variable, as long as they are using _Atomic or atomic primitives. Thus, relaxed ordering is already sufficient because we are not using `ksCurTime` as a lock or condition variable to synchronize. It's purely reading/writing the value in and out of itself. Based on these rules, monotonicity is guaranteed with relaxed atomics.

> All in all I'm uneasy to rely on relaxed doing the right thing if there is some form of communication from one core to another, like in these cases, and I'd rather use the stronger acquire-release semantics.

Again, It only applies to when we are message-passing through the `ksCurTime` variable. E.g., if it were a pointer to publish a structure, or a flag variable to publish other structures. Here, we are purely reading the value of `ksCurTime`, and therefore it's OK to use relaxed atomics. Also the original code didn't achieve any of the acquire/release semantics, because there's no paring of atomic op or fences on the reader side.

For keeping the `__atomic_thread_fence(__ATOMIC_ACQ_REL)` for arm, I would prefer removing it. Let me have some discussion with Kurt if I can to understand if there's additional barriers (for ordering access to system features) that needs to be in place. From C11 language level, it's useless. For the GCC generated fence on riscv that includes i/o fences, I hope seL4 kernel boot doesn't rely on this. BTW, I already found a case in sDDF where we most likely relied on this accidental feature:

https://github.com/au-ts/sddf/blob/70ccc26f645821e3d2b37cc6e8a869d6b74ef00c/drivers/network/dwmac-5.10a/ethernet.c#L96

The `THREAD_MEMORY_RELEASE` is just wrong. I'll have another issue/PR open to sDDF.

Lastly for the x86 case, `smp_aps_index++` is done before `NODE_LOCK_SYS`. Thus, I suspect there are some data-race issues. @amrzar or @Indanz Would you mind take a look?

KurtWu10:
- RISC-V's I/O fences are different from normal memory barriers (they use a different encoding, according to section 2.7 of the unprivileged spec). 

- `__atomic_thread_fence(__ATOMIC_ACQ_REL);` only acts as a compiler barrier on x86 ([godbolt](https://godbolt.org/z/oabeWqWWz), [ref](https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html)).

- > Thus compiler will not combine multiple read or multiple write on the same variable, as long as they are using _Atomic or atomic primitives.
  
  I *think* compilers can hoist non-volatile atomics ([ref1](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1489.htm), [ref2](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1284.htm)). It seems that llvm currently explicitly forbids this ([ref1](https://releases.llvm.org/20.1.0/docs/Atomics.html#atomics-and-ir-optimization), [ref2](https://github.com/llvm/llvm-project/blob/llvmorg-20.1.6/llvm/lib/Transforms/Scalar/LICM.cpp#L1171-L1174)), but I am not sure whether this is true at the C abstract machine level.

ganboing:
>   I _think_ compilers can hoist non-volatile atomics ([ref1](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1489.htm), [ref2](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1284.htm)). It seems that llvm currently explicitly forbids this ([ref1](https://releases.llvm.org/20.1.0/docs/Atomics.html#atomics-and-ir-optimization), [ref2](https://github.com/llvm/llvm-project/blob/llvmorg-20.1.6/llvm/lib/Transforms/Scalar/LICM.cpp#L1171-L1174)), but I am not sure whether this is true at the C abstract machine level.

I _think_ any sane compiler should not. Below is from GCC wiki: https://gcc.gnu.org/wiki/Atomic/GCCMM/Optimizations
> Relaxed mode operation has no synchronization side effects. This would indicate it could be treated like a normal shared memory access, but there appears to be a wrinkle. It is required to retain its ordering within a thread, and if I read the standard correct [n3242.1.10.19]
```
The four preceding coherence requirements effectively disallow compiler reordering of atomic operations
to a single object, even if both operations are relaxed loads. This effectively makes the cache coherence
guarantee provided by most hardware available to C++ atomic operations.
```

The four preceding coherence requirements that wiki refers to is the `read-read` `read-write` `write-read` and `write-write` coherence order I mentioned previously. Thus, I'm understanding the standard the same way as GCC and LLVM developers.

ganboing:
@KurtWu10 I see the disconnect now. The C++11 standard has the 4 coherence order in the text, but C11 standard doesn't. N1584 is there to fix that: https://www.open-std.org/jtc1/sc22/wg14/issues/c11c17/issue0402.html I don't have the access to the final C11 standard text, but the link above says:
```
Comment from WG14 on 2017-11-03:

Oct 2011 meeting

Committee Discussion
The consensus was that this is an oversight, and should be changed as recommended.
Feb 2012 meeting
```
Given the GCC/LLVM already does the right thing, I think it's safe to say we can expect any sane compiler to follow suit.

KurtWu10:
~~I guess the issue of non-volatile atomics can only be ruled out by having a proper forward progress model ([ref](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2809r3.html)), which talks about liveness properties rather than mere safety. The issue of coherence order is fixed in C17, but it is only a safety property but not liveness.~~ This seems to be not relevant here, as noted by Bo.

Note that if we are to strictly follow the C standard, it is already problematic for SC accesses due to some theoretical issues found in the C++ side ([ref1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0668r5.html), [ref2](https://cplusplus.github.io/LWG/lwg-active.html#3941)).

Indanz:
Well, point 4 of ref2 says: "The primary practical effect of that is that formal verification of C++ programs using relaxed atomics remains unfeasible."

Which strengthens my opinion of not using relaxed. The rest is basically saying that using relaxed or consumed together with acquire and release is problematic. That sequentially consistent ordering isn't as strong as defined isn't a problem for how people use it.

Another problem I have with relaxed is that it doesn't seem to give much reason to use it, compared to plain read and writes of global variables. Aligned accesses are architecturally guaranteed to be atomic, so why ever use relaxed? And global memory writes aren't allowed to be optimised away already.

> Yes, you are right, there's no data-race on ksNumCPUs, but only for all APs. For BSP, it's still being accessed in a racy way, because BSP is not in CS. 

Please use normal terminology, from another post it seems you mean the boot cpu core with "BSP" (not Board Support Package) and with APs you mean "Application Processors", but only other cores? I guess BSP is RISCV terminology, but AP sounds like Arm, not RISCV. It's very confusing.

Anyway, accessing things out of critical sections does not mean it is automatically racy. There needs to be either room for compilers to optimise things wrong, or concurrent access in such way that things can misbehave. You still haven't explained what can actually go wrong with how we use ksNumCPUs.

Your suggestions that using relaxed atomics is better doesn't make sense, as atomic updates do not avoid data races. They sometimes help for read-modify-write operations, but only when there are multiple writers, which isn't the case here. 
There isn't any requirement that ksNumCPUs need to be updated atomically, as all writers are serialised. It's a global variable, the compiler already is forced to update it. The only real requirement is that the boot core does not prematurely see ksNumCPUs == CONFIG_MAX_NUM_NODES.

> For c, the use of fence and acq_rel actually does nothing in the original code, because there's no atomics or fences on the AP side to pair with. Thus, there's actually no memory order in the original code. I.e., ksCurTime is not being read/written as atomic, and there's no fence on the reader side.

That is not how hardware works! Your view is too theoretical. Yes, the memory model describes behaviour with matching accesses, as that makes sense when describing memory ordering. But by hardware ordering is still guaranteed on the write side, it's not that the barriers suddenly disappear or are ignored because there are no matching reader barriers (the compiler and hardware don't know who the reader is).

That said, you are wrong about there being no matching fences. If we take RISCV as example, it does a `__atomic_thread_fence(__ATOMIC_ACQ_REL);` in `release_secondary_cores()` on the write side and on the read side we have a `__atomic_thread_fence(__ATOMIC_ACQUIRE)` in `clock_sync_test()`. Arm does the same in `release_secondary_cpus` and the clock sync code is shared. Considering you missed this, I feel you haven't looked at what the code is actually doing properly.

For any concurrency issues in code, you need to clearly state:
- What the problem is, what can go wrong.
- Why it is wrong.
- How it can happen (how to trigger, example concurrency flow).

This is not really different from any bug report, really, but with concurrency issues it is more important to be precise, as the issues can be less obvious. Vague theoretical hand waving as has been going on so far is not good enough.

I'll close this PR as I don't see this going further, as you've not addresses my points and have missed what the code is actually doing twice now.

Issue #1104, opened 2023-09-14
MCS fault handling issue when specifying period of a TCB.


Kswin01:
This issue is related to VPPI faults on AArch64, which uses the fault handling mechanisms to deliver virtual PPI's to a single-threaded VM. Specifically, it was observed in an seL4 system built using Microkit where a virtual-machine-monitor services a Linux virtual machine. The `microkit` branch of seL4 was used with the Odroid-C4 and qemu-arm-virt platforms.

When specifying a period different to the budget of the TCB for the virtual machine and using the VPPI handling mechanism, we non-deterministically hit an assert in the ```tcbReleaseEnqueue``` function where our TCB is seemingly already in the release queue, and potentially shouldn't be at this point in time. The particular assert is here: https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/object/tcb.c#L283

This issue doesn't occur when we set the period of the VM equal to our budget, and the thread is treated as a round robin thread.

My main suspicion is that the `postpone()` function, which in turn calls ```tcbReleaseEnqueue``` was getting invoked twice, but I have not been unable to completely validate this assumption. https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/kernel/thread.c#L568 

Postpone is called from three places mainly https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/kernel/thread.c#L614 
https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/kernel/thread.c#L128
and
https://github.com/seL4/seL4/blob/61a8ccef44eeb08ecd20ad40296340c8319adee5/src/object/schedcontext.c#L284 

So the interrupting thread may be unable to refill or have insufficient refill available when trying to end its timeslice, handle a reply transfer or attempt to resume.

The naive and inconclusive fix for this was to switch the assert to an if statement, and return if the tcb was already in the release queue rather than hitting the assert. As the `tcbReleaseEnqueue` function places the thread into the release queue, which it already is in, I assumed this to be an ok behavior. However, this lead to other issues, such as hitting asserts stating the the thread is un-scheduable. 

This was the snippet added to `tcbReleaseEnqueue`:
```
    if (thread_state_get_tcbInReleaseQueue(tcb->tcbState) == true) {
        // Thread is already in queue,no need to queue again
        return;
    }
```
[...]

PR #1486, opened 2025-06-19
Fix VPPI and VGIC Maintenance handling

Kswin01:
Add a check to ensure that we are not handling a VGIC/VPPI request if the current thread has  been placed in the release queue. This PR fixes: https://github.com/seL4/seL4/issues/1104.

### DIFF
--- a/src/arch/arm/object/vcpu.c
+++ b/src/arch/arm/object/vcpu.c
@@ -110,11 +110,16 @@ void VPPIEvent(irq_t irq)
      * correctly receive a fault IPC message. This may occur due to the
      * budget check that happens early in the handleInterruptEntry.
      *
-     * If the current thread does *not* have budget this interrupt is
-     * ignored for now. As it is a level-triggered interrupt it shall
-     * be re-raised (and not lost).
+     * If the current thread does *not* have budget, as indicated by its
+     * presence in the release queue, this interrupt is ignored for now.
+     * As it is a level-triggered interrupt it shall be re-raised
+     * (and not lost).
+     *
+     * Additionally, if we have already received a timeout fault due to
+     * our timeslice having ended, and we have a timeout fault handler for
+     * this thread, we do not want to overwrite that with our VCPU fault.
      */
-    if (thread_state_get_tcbQueued(NODE_STATE(ksCurThread)->tcbState)) {
+    if (!isSchedulable(NODE_STATE(ksCurThread))) {
         return;
     }
 #endif
@@ -141,7 +146,7 @@ void VGICMaintenance(void)
 
 #ifdef CONFIG_KERNEL_MCS
     /* See VPPIEvent for details on this check. */
-    if (thread_state_get_tcbQueued(NODE_STATE(ksCurThread)->tcbState)) {
+    if (!isSchedulable(NODE_STATE(ksCurThread))) {
         return;
     }
 #endif


lsf37:
Nice. How hard would it be to generate a test case for that? I guess reliably triggering an IRQ when a thread is in the release queue is not exactly easy.

Indanz (commented):
Congratulations on hunting this one down.

Now the main question is whether this is sufficient, or should we use `isSchedulable()` instead to catch other corner cases too?

`possibleSwitchTo()` also has a `thread_state_get_tcbInReleaseQueue()` check, but it's the only other one, so I wonder if that one is correct too.

Indanz:
> Nice. How hard would it be to generate a test case for that? I guess reliably triggering an IRQ when a thread is in the release queue is not exactly easy.

Release queue is for periodic tasks waiting for budget replenishment, so all you need is a task with small budget and very large period.

The problem is that sel4test doesn't has any IRQs for tests to use themselves, and in this case you can't steal the timer interrupt.

midnightveil:
> The problem is that sel4test doesn't has any IRQs for tests to use themselves, and in this case you can't steal the timer interrupt.

Extremely hacky, but QEMU has devices (e.g. the 'EDU' device) where it would be possible to create an IRQ on demand. It's likely possible to (ab)use that functionality. (And I did, to generate spurious IRQs on demand for another test).

Kswin01:
> Now the main question is whether this is sufficient, or should we use `isSchedulable()` instead to catch other corner cases too?

Yep I agree, `isSchedulable()` would suit this case as well. I've pushed a fix up, did some quick tests and it has the same behavior. 


> `possibleSwitchTo()` also has a `thread_state_get_tcbInReleaseQueue()` check, but it's the only other one, so I wonder if that one is correct too.

I think `if (target->tcbSchedContext != NULL && !thread_state_get_tcbInReleaseQueue(target->tcbState))` could be changed to just `if(isSchedulable(target))`. It's a more thorough check as it also ensures that the TCB is in a runnable state (although this is eventually checked in the scheduler logic I believe).

src/arch/arm/object/vcpu.c
```diff
@@ -110,11 +110,13 @@ void VPPIEvent(irq_t irq)
...
-     * ignored for now. As it is a level-triggered interrupt it shall
-     * be re-raised (and not lost).
+     * If the current thread does *not* have budget, as indicated by its
+     * presence in the release queue,  this interrupt is ignored for now.
+     * As it is a level-triggered interrupt it shall be re-raised
+     * (and not lost).
      */
-    if (thread_state_get_tcbQueued(NODE_STATE(ksCurThread)->tcbState)) {
+    if (thread_state_get_tcbQueued(NODE_STATE(ksCurThread)->tcbState) ||
+        !isSchedulable(NODE_STATE(ksCurThread))) {
```
Indanz:
```suggestion
    if (!isSchedulable(NODE_STATE(ksCurThread))) {
```
Actually, reading the original comment, it seems it always intended to check the release queue, not the scheduling queue, so replacing the original `thread_state_get_tcbQueued` with `thread_state_get_tcbInReleaseQueue` would be the minimal fix.

Whether the VCPU ran out of budget itself doesn't matter, as we're faulting and switching tasks anyway. It's just that the generic fault code doesn't expect the task to be already in the release queue, leading to problems like #1104. This is caused by the weird choice of communicating such VCPU IRQs as faults, as that breaks some assumption in generic code. So an alternative fix would be to make the generic code robust against tasks already in the release queue.

However, there is another problem: A task can only receive one fault, either a timeout fault or a VCPU fault, but not both at the same time. Checking for the release queue only works if the VCPU has no timeout handler. If it has, the timeout fault would be replaced with the VCPU fault, which isn't the intention. In case of a timeout fault, the VCPU state will be either `ThreadState_Inactive` or `ThreadState_BlockedOnReply`. So it does look like we want at least an additional `isRunnable()` check, which together is half of `isSchedulable()`. We should add this corner case to the comment.

I'm not sure if VCPU IRQs can arrive after we switched away from a VCPU. It seems we can, if an IRQ was pending when we switch away from the VCPU task, but perhaps that's prevented by the VCPU switching code. (Can pending GIC IRQs be delivered for the wrong VCPU on VCPU switch? I hope not.)

Kswin01: > I'm not sure if VCPU IRQs can arrive after we switched away from a VCPU. It seems we can, if an IRQ was pending when we switch away from the VCPU task, but perhaps that's prevented by the VCPU switching code. (Can pending GIC IRQs be delivered for the wrong VCPU on VCPU switch? I hope not.)

I'm not entirely convinced we are actually safeguarding against this. The VGIC maintenance code seems to check if the `GICH_MISR` has the `EOI` bit set to check if a VGIC maintenance has been requested. However, if it's not set, it makes the following assumption: `/* Assume that it was an EOI for a LR that was not present */`, and still delivers a fault. Would this be able to deliver to the wrong VCPU potentially then? Although that specific timing would probably be pretty rare. Additionally, the VGIC maintenance fault would eventually be delivered to the correct VCPU when it is scheduled next.

 In the case of a VPPI, the timer should only be running when that VCPU is active, so ***hopefully*** a VPPI is only delivered to the correct VCPU.

> So it does look like we want at least an additional `isRunnable()` check, which together is half of `isSchedulable()`. We should add this corner case to the comment.

I agree, I'll add that in.

src/arch/arm/object/vcpu.c
```diff
@@ -110,11 +110,19 @@ void VPPIEvent(irq_t irq)
...
+     * presence in the release queue, this interrupt is ignored for now.
+     * As it is a level-triggered interrupt it shall be re-raised
+     * (and not lost).
+     *
+     * Additionally, if we have already received a timeout fault due to
+     * our timeslice having ended, and we have a timeout fault handler for
+     * this thread, we do not want to overwrite that with our VCPU fault.
+     * Therefore, we check if the thread is runnable through isSchedulable()
+     * as if we have a timeout fault, the thread should be in a inactive
+     * or blocked on reply state, and not runnable.
```
Indanz:
```suggestion
```
No need to explain such details, just the reason for `isSchedulable()`, which you did. People can check the details themselves if they want (which they should do anyway instead of trusting the comment).

Indanz (approved):

src/arch/arm/object/vcpu.c
Indanz: > > I'm not sure if VCPU IRQs can arrive after we switched away from a VCPU. It seems we can, if an IRQ was pending when we switch away from the VCPU task, but perhaps that's prevented by the VCPU switching code. (Can pending GIC IRQs be delivered for the wrong VCPU on VCPU switch? I hope not.)
> 
> I'm not entirely convinced we are actually safeguarding against this.
>
> The VGIC maintenance code seems to check if the `GICH_MISR` has the `EOI` bit set to check if a VGIC maintenance has been requested. However, if it's not set, it makes the following assumption: `/* Assume that it was an EOI for a LR that was not present */`, and still delivers a fault. Would this be able to deliver to the wrong VCPU potentially then? Although that specific timing would probably be pretty rare. Additionally, the VGIC maintenance fault would eventually be delivered to the correct VCPU when it is scheduled next.
> 
> In the case of a VPPI, the timer should only be running when that VCPU is active, so _**hopefully**_ a VPPI is only delivered to the correct VCPU.

Kent had some comments about it, as I raised this before: https://github.com/seL4/seL4/pull/1183#discussion_r1980441717

I think we're okay, but haven't done a detailed review to double check it.

Indanz (approved):

PR #1487, opened 2025-06-25
HW Debug API: Better handling of BAS on ARM

alwin-joshy:
Watchpoint value registers should contain an 8-byte aligned virtual address (4-byte aligned are also supported, but this is deprecated). Ranges inside the doubleword pointed to by this address can be watched by setting a contigious range of bits inside the BAS field of the corresponding watchpoint control register i.e. 0th bit set for `addr`, 1st bit set for `addr + 1`.

The current implementation did not require that the address was aligned to eight bytes, which meant that watchpoints did not always work for small types such as characters and shorts, which are not always 4-byte aligned. In this patch, if a watchpoint with size < 8 is set, we align the address to the previous word boundary and calculate the BAS to trigger an exception for the range intended by the user.

I chose to implement it with the BAS calculation logic in the kernel because it is transparent to userspace, which passes in an arbitrary virtual address and size. Additionally, the ARMv8 manual states that _"The valid values for BAS are nonzero binary numbers all of whose set bits are contiguous. All other values are reserved and must not be used by software"_, meaning that some logic to validate that a user passes in a correct BAS would likely need to be added to the kernel anyway.

### DIFF
--- a/src/arch/arm/machine/debug.c
+++ b/src/arch/arm/machine/debug.c
@@ -290,13 +290,17 @@ void setBreakpoint(tcb_t *t,
     } else {
         dbg_wcr_t wcr;
 
+        /* Address has to be aligned to 8 byte boundary */
+        word_t offset = vaddr & 0x7;
+        vaddr &= ~0x7;
+
         writeWvrContext(t, bp_num, vaddr);
 
         /* Preserve reserved bits */
         wcr.words[0] = readWcrContext(t, bp_num);
         wcr = dbg_wcr_set_enabled(wcr, 1);
         wcr = dbg_wcr_set_pac(wcr, DBGWCR_PRIV_USER);
-        wcr = dbg_wcr_set_bas(wcr, convertSizeToArch(size));
+        wcr = dbg_wcr_set_bas(wcr, convertSizeToArch(size) << offset);
         wcr = dbg_wcr_set_lsc(wcr, convertAccessToArch(rw));
         wcr = dbg_wcr_set_watchpointType(wcr, 0);
         wcr = dbg_wcr_set_lbn(wcr, 0);
@@ -444,15 +448,24 @@ int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
     }
 
     if (reason == seL4_DataBreakpoint) {
+        /* Align the fault address to the previous 8-byte boundary */
+        word_t offset = vaddr & 0x7;
+        vaddr &= ~0x7;
         for (i = 0; i < seL4_NumExclusiveWatchpoints; i++) {
             dbg_wcr_t wcr;
-            word_t wvr = readWvrCp(i);
-
             wcr.words[0] = readWcrCp(i);
-            align_mask = convertArchToSize(dbg_wcr_get_bas(wcr));
-            align_mask = ~(align_mask - 1);
+            if (!dbg_wcr_get_enabled(wcr)) {
+                continue;
+            }
 
-            if (wvr != (vaddr & align_mask) || !dbg_wcr_get_enabled(wcr)) {
+            word_t wvr = readWvrCp(i);
+            if (vaddr != wvr) {
+                continue;
+            }
+
+            /* Check if the BAS set for this watchpoint corresponds
+             * to the faulting vaddr */
+            if ((dbg_wcr_get_bas(wcr) & BIT(offset)) == 0) {
                 continue;
             }
 


src/arch/arm/machine/debug.c
```diff
@@ -229,6 +229,18 @@ uint16_t getBpNumFromType(uint16_t bp_num, word_t type)
...
+static seL4_Word calculate_wcr_BAS(seL4_Word address, seL4_Word size)
+{
+    seL4_Word bas = 0;
+    seL4_Word set_address = address & ~0x7;
+
+    for (int i = address - set_address; i < (address - set_address) + size; i++) {
+        bas |= (1 << i);
+    }
+
+    return bas;
```
Indanz:
```suggestion
    seL4_Word offset = address & 0x7;
    seL4_Word mask = convertSizeToArch(size); // To handle default 0 as 4 instead of 0.

    return mask << offset;
```

But at this point, considering the dependency on set_address and address, it would make more sense to calculate offset in the caller and do `wcr = dbg_wcr_set_bas(wcr, convertSizeToArch(size) << offset);` below and get rid of this helper function altogether.

alwin-joshy: Yeah, that's much cleaner. I'll change it to that

src/arch/arm/machine/debug.c
```diff
@@ -449,10 +453,15 @@ int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
...
 
-            if (wvr != (vaddr & align_mask) || !dbg_wcr_get_enabled(wcr)) {
+            /* Align the fault address to the previous word */
+            if (!dbg_wcr_get_enabled(wcr) || (vaddr & ~0x7) != wvr) {
+                continue;
+            }
+
+            /* Check if the BAS set for this watchpoint corresponds to this vaddr */
+            int offset = vaddr & 0x7;
+            if (dbg_wcr_get_bas(wcr) & BIT(offset) == 0) {
```
Indanz:
```suggestion
            if ((dbg_wcr_get_bas(wcr) & BIT(offset)) == 0) {
```
Because of precedence order you need parentheses. I'd also use `!` instead of `== 0`, but that's personal preference.

alwin-joshy: Oops, had it right originally but messed it up when I changed something.

src/arch/arm/machine/debug.c
```diff
@@ -446,13 +450,21 @@ int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
...
-
             wcr.words[0] = readWcrCp(i);
-            align_mask = convertArchToSize(dbg_wcr_get_bas(wcr));
-            align_mask = ~(align_mask - 1);
 
-            if (wvr != (vaddr & align_mask) || !dbg_wcr_get_enabled(wcr)) {
+            if (!dbg_wcr_get_enabled(wcr)) {
+                continue;
+            }
+
```
Indanz, line 460: No need for such empty lines.

src/arch/arm/machine/debug.c
```diff
@@ -446,13 +450,21 @@ int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
...
+            /* Align the fault address to the previous word */
+            word_t wvr = readWvrCp(i);
+            if ((vaddr & ~0x7) != wvr) {
+                continue;
+            }
+
+            /* Check if the BAS set for this watchpoint corresponds to this vaddr */
+            int offset = vaddr & 0x7;
```
Indanz: You can calculate offset and align vaddr outside the loop (though compilers should be smart enough to do that already). But calculating offset first and then `vaddr &= 0x7;` seems cleaner.

Bonus points for cleaning up the `seL4_InstructionBreakpoint` case, as that requires the same logic, except aligned to 4 bytes instead of 8.

alwin-joshy:
> Assuming 8 byte aligned watchpoints are not an Aarch64-only thing, all seems fine to me. If it isn't, then you need to align to either 4 or 8 bytes, depending on WORD_SIZE.

Yep, it's the same on aarch64 and aarch32 according to the TRM.

src/arch/arm/machine/debug.c
alwin-joshy: So instruction breakpoints are a little different in that we only ever allow the BAS to be `0b1111` and we just need to check that the faulting address is somewhere in the 4 bytes following the `bvr`, in which case the existing logic works.

I'm wondering if this is even necessary for instruction breakpoints? Shouldn't the fault address always be 4-byte aligned anyway?

src/arch/arm/machine/debug.c
```diff
@@ -444,15 +448,23 @@ int getAndResetActiveBreakpoint(word_t vaddr, word_t reason)
...
-            word_t wvr = readWvrCp(i);
-
             wcr.words[0] = readWcrCp(i);
-            align_mask = convertArchToSize(dbg_wcr_get_bas(wcr));
-            align_mask = ~(align_mask - 1);
+            if (!dbg_wcr_get_enabled(wcr)) {
+                continue;
+            }
+
+            /* Align the fault address to the previous word */
```
Indanz: This comment should move too, or be changed. It doesn't make sense for 32-bit Arm to say `word`.
Aligning to 8 bytes also on 32-bit Arm is worth a comment I think.

src/arch/arm/machine/debug.c
Indanz: Not for Thumb mode, but not sure that's supported anyway.

I don't know the details, but what if you want to debug a buggy jump to an unaligned address? You may want to set a breakpoint there before trapping.

alwin-joshy: If you set the BAS to `0x3` or `0xc` (all other values are reserved iirc), you could have the breakpoint be set on a 2 byte range (for thumb mode), but I think the hardware does not otherwise allow you to set an instruction breakpoint on an non 4-byte aligned address.

Indanz: The Arm [docs](https://developer.arm.com/documentation/ddi0406/cb/Debug-Architecture/The-Debug-Registers/Register-descriptions--in-register-order/DBGBVR--Breakpoint-Value-Registers?lang=en) says:

> Bits[1:0], when register used for address comparison
>   Must be written as 0b00, otherwise the generation of Breakpoint debug events by this breakpoint is unpredictable.

So the programmed address must always be 4-byte aligned. It seems you're right that we should be able to remove those checks.

Indanz:
> > Assuming 8 byte aligned watchpoints are not an Aarch64-only thing, all seems fine to me. If it isn't, then you need to align to either 4 or 8 bytes, depending on WORD_SIZE.
> 
> Yep, it's the same on aarch64 and aarch32 according to the TRM.

How does that work if the BAS field is 4 bits, like on 32-bit Arm? That would mean breakpoints would only work on even addresses, but can't be set for odd addresses.

Edit: Gah, another documents:
> BAS, bits[12:5] or bits[8:5]
>    Byte address select. It is implementation defined whether a 4-bit or an 8-bit Byte address select field is implemented:
>        an 8-bit Byte address select field is DBGWCR[12:5]
>        a 4-bit Byte address select field is DBGWCR[8:5]. DBGWCR[12:9] is RAZ/WI.

How do we know what we're dealing with?

Edit2: [Elsewhere](https://developer.arm.com/documentation/ddi0406/cb/Debug-Architecture/Debug-Events/Watchpoint-debug-events/Byte-address-selection-and-masking-defined-by-the-DBGWCR?lang=en#Ciahgjfj)

> Debuggers can use the same programming model on implementations that support:
> 
>     an 8-bit Byte address select field, DBGWCR.BAS[7:0]
> 
>     a 4-bit Byte address select field, DBGWCR.BAS[3:0].
> 
> This is because, on an implementation that supports only a 4-bit Byte address select field, writes to [DBGWCR](https://developer.arm.com/documentation/ddi0406/cb/Debug-Architecture/The-Debug-Registers/Register-descriptions--in-register-order/DBGWCR--Watchpoint-Control-Registers?lang=en)[7:4] are ignored.

As far as I can find, there is no mention that addresses need to have 8-byte alignment even on 32-bit arm, they all talk about word alignment. Can you give a reference where it says that this is different?

Also, strictly speaking, shouldn't we also set MASK to exclude hits on bytes we're not interested in?

alwin-joshy:
My comments were based on the ARM architecture reference manual (https://developer.arm.com/documentation/ddi0487/latest/).

For both aarch64 and aarch32, the BAS field is 8 bits wide and there is a comment stating that _"If DBGWVR<n>[2] == 1, only BAS[3:0] are used and BAS[7:4] are ignored. Arm deprecates setting DBGWVR<n>[2] == 1."_, which if I'm not misunderstanding means that the address should be aligned to 8 bytes.

But I guess this is for aarch32 in ARMv8? Maybe it is implementation-defined as you say for ARMv7? Do we distinguish between the two in the kernel?

Indanz:
> My comments were based on the ARM architecture reference manual (https://developer.arm.com/documentation/ddi0487/latest/).

I know that. When I ask for a reference, I mean document version + chapter/section/page number, as searching in a 17k page document isn't trivial.

> For both aarch64 and aarch32, the BAS field is 8 bits wide

Where does it say that?

> and there is a comment stating that _"If DBGWVR[2] == 1, only BAS[3:0] are used and BAS[7:4] are ignored. Arm deprecates setting DBGWVR[2] == 1."_, which if I'm not misunderstanding means that the address should be aligned to 8 bytes.

Yes, but that comment is in "AArch64 System Register Descriptions, D24.3 Debug registers", so it applies to aarch64, not aarch32.

> But I guess this is for aarch32 in ARMv8? Maybe it is implementation-defined as you say for ARMv7? Do we distinguish between the two in the kernel?

I think that is the case, yes. I don't think we really distinguish between the two in the kernel, I might be wrong though. So for aarch32, we need to check/know whether 8 byte BAS is supported and only align that then. Otherwise, just use 32-bit alignment. But keeping it 4 byte aligned for aarch32 for now would be fine for a while, this is a long term deprecating thing I suspect.

src/arch/arm/machine/debug.c
```diff
@@ -290,13 +290,17 @@ void setBreakpoint(tcb_t *t,
     } else {
         dbg_wcr_t wcr;
 
+        /* Address has to be aligned to 8 byte boundary */
+        word_t offset = vaddr & 0x7;
+        vaddr &= ~0x7;
```
Indanz, line 295: TODO: Use proper alignment for aarch32.

Indanz (changes requested):

alwin-joshy:
Ah right, document number "ARM DDI 0487" and the specific references are D24.3.11 for aarch64 and G8.3.26 for aarch32. The comment about deprecation is the same for both.

Kswin01:
@kent-mcleod Where you able to take a look at these changes?

Issue #1085, opened 2023-08-03
Why are seL4_TCB_WriteRegisters/seL4_TCB_ReadRegisters not taking into account the count parameter?


Ivan-Velickovic:
Below is the libsel4 definition of `seL4_TCB_WriteRegisters` for AArch64:
```
seL4_TCB_WriteRegisters(seL4_TCB _service, seL4_Bool resume_target, seL4_Uint8 arch_flags, seL4_Word count, seL4_UserContext *regs)
{
	seL4_Error result;
	seL4_MessageInfo_t tag = seL4_MessageInfo_new(TCBWriteRegisters, 0, 0, 38);
	seL4_MessageInfo_t output_tag;
	seL4_Word mr0;
	seL4_Word mr1;
	seL4_Word mr2;
	seL4_Word mr3;

	/* Marshal and initialise parameters. */
	mr0 = (resume_target & 0x1ull) | ((arch_flags & 0xffull) << 8);
	mr1 = count;
	mr2 = regs->pc;
	mr3 = regs->sp;
	seL4_SetMR(4, regs->spsr);
	seL4_SetMR(5, regs->x0);
	seL4_SetMR(6, regs->x1);
	seL4_SetMR(7, regs->x2);
	seL4_SetMR(8, regs->x3);
	seL4_SetMR(9, regs->x4);
	seL4_SetMR(10, regs->x5);
	seL4_SetMR(11, regs->x6);
	seL4_SetMR(12, regs->x7);
	seL4_SetMR(13, regs->x8);
	seL4_SetMR(14, regs->x16);
	seL4_SetMR(15, regs->x17);
	seL4_SetMR(16, regs->x18);
	seL4_SetMR(17, regs->x29);
	seL4_SetMR(18, regs->x30);
	seL4_SetMR(19, regs->x9);
	seL4_SetMR(20, regs->x10);
	seL4_SetMR(21, regs->x11);
	seL4_SetMR(22, regs->x12);
	seL4_SetMR(23, regs->x13);
	seL4_SetMR(24, regs->x14);
	seL4_SetMR(25, regs->x15);
	seL4_SetMR(26, regs->x19);
	seL4_SetMR(27, regs->x20);
	seL4_SetMR(28, regs->x21);
	seL4_SetMR(29, regs->x22);
	seL4_SetMR(30, regs->x23);
	seL4_SetMR(31, regs->x24);
	seL4_SetMR(32, regs->x25);
	seL4_SetMR(33, regs->x26);
[...]

PR #1488, opened 2025-06-26
Use loop and respect the limit parameter for seL4_TCB_WriteRegisters and seL4_TCB_ReadRegisters

ZhanYF:
Use loop and respect the limit parameter for seL4_TCB_WriteRegisters and seL4_TCB_ReadRegisters.

This addresses issue #1085 and currently only affects arm64.

Diff for the generated sel4_client.h:
https://gist.github.com/ZhanYF/3b36cb35591352bd1d094b1724ef7993

I tested this with libvmm on Odroidc4 and qemu and it worked OK, please let me know if you have any feedback.

### DIFF
--- a/libsel4/tools/syscall_stub_gen.py
+++ b/libsel4/tools/syscall_stub_gen.py
@@ -698,10 +698,25 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
                 result.append("\tmr%d = %s;" % (i, input_expressions[i]))
             else:
                 result.append("\tmr%d = 0;" % i)
-        # Initialise buffered parameters
-        for i in range(num_mrs, len(input_expressions)):
-            result.append("\tseL4_SetMR(%d, %s);" % (i, input_expressions[i]))
-        result.append("")
+        TCB_writeRegister_can_loop = (
+            method_name == "WriteRegisters" and
+            interface_name == "seL4_TCB" and
+            num_mrs >= 2
+        )
+        if TCB_writeRegister_can_loop:
+            # First 2 registers are always passed by mrs, thus initialise buffered
+            # parameters only when requesting more than 2 registers
+            regs_by_mrs = num_mrs - 2
+            result.append("\tfor (seL4_Word i = %d; i < count; i++) {" % regs_by_mrs)
+            result.append("\t\tseL4_SetMR(i + 2, ((seL4_Word*)&%s)[i - %d]);" %
+                          (input_expressions[num_mrs], regs_by_mrs))
+            result.append("\t}")
+            result.append("")
+        else:
+            # Initialise buffered parameters
+            for i in range(num_mrs, len(input_expressions)):
+                result.append("\tseL4_SetMR(%d, %s);" % (i, input_expressions[i]))
+            result.append("")
 
     #
     # Generate the call.
@@ -747,15 +762,25 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
         for i in range(MAX_MESSAGE_LENGTH):
             if i < num_mrs:
                 source_words["w%d" % i] = "mr%d" % i
-            else:
+            elif method_name != "ReadRegisters" or num_mrs < 2:
                 source_words["w%d" % i] = "seL4_GetMR(%d)" % i
         unmashalled_params = generate_unmarshal_expressions(output_params, wordsize)
         for (param, words) in unmashalled_params:
             if param.type.pass_by_reference():
                 members = struct_members(param.type, structs)
-                for i in range(len(words)):
-                    result.append("\t%s->%s = %s;" %
-                                  (param.name, members[i], words[i] % source_words))
+                if method_name == "ReadRegisters" and num_mrs >= 2:
+                    for i in range(len(source_words)):
+                        result.append("\t%s->%s = %s;" %
+                                      (param.name, members[i], words[i] % source_words))
+                    result.append("\tfor (seL4_Word i = %d; i < count; i++) {" % num_mrs)
+                    result.append("\t\t((seL4_Word *) &%s->%s)[i - %d] = seL4_GetMR(i);" %
+                                  (param.name, members[num_mrs], num_mrs))
+                    result.append("\t}")
+
+                else:
+                    for i in range(len(words)):
+                        result.append("\t%s->%s = %s;" %
+                                      (param.name, members[i], words[i] % source_words))
             else:
                 if param.type.double_word:
                     result.append("\tresult.%s = ((%s)%s + ((%s)%s << 32));" %
@@ -764,10 +789,6 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
                 else:
                     for word in words:
                         result.append("\tresult.%s = %s;" % (param.name, word % source_words))
-
-    #
-    # }
-    #
     result.append("\treturn result;")
     result.append("}")
 


Indanz (commented):
Thanks for spending time on this.

Although your solution solves the initial issue for Aarch64, the problem applies to all architectures. I would be fine with writing the few first registers unnecessarily to get rid of the magic 2 and 4 numbers.

It only solves it for the syscall wrappers, it doesn't make things any easier for user space, which has the same problem of being forced to set each individual register one by one. Basically, the user space API is horrible, but we also don't want to break it.

Normally the solution for that would be to use an anonymous union. That would mean changing `seL4_UserContext`, wrap all existing fields in an unnamed union and add an union containing a `regs[]` array of the right size. Doing this is standard C11 and `seL4_UserContext` is only used by user space, so I think it should be fine.

lsf37:
> Normally the solution for that would be to use an anonymous union. That would mean changing `seL4_UserContext`, wrap all existing fields in an unnamed union and add an union containing a `regs[]` array of the right size. Doing this is standard C11 and `seL4_UserContext` is only used by user space, so I think it should be fine.

Sorry, libsel4 is in scope for verification in the future and that's exactly the kind of thing that would break it. So we should not do that.

We could add a second unrelated API function, though, that uses an array instead and just happens to call the same kernel invocation.

Indanz:
Surely that's a long-term project that will take years. By then we can remove the union and only keep the array. Or we just define two structures and add a wrapper that calls the verified one with a cast (and I'd suggest making the array version verified, as that is closer to the actual kernel code).

lsf37:
> Surely that's a long-term project that will take years. 

That doesn't mean we should let the code diverge from the verification subset. We've just spent extra work last year to bring it into the subset again. There is a solution without unions, I see no reason to not use that.

ZhanYF:
Thanks for your feedback! Would it be an acceptable solution if I port my changeset to other archs and get rid of the magic 2 and 4 numbers? I think the userspace API is still an issue but that can be addressed at a later stage.

Indanz:
> Thanks for your feedback! Would it be an acceptable solution if I port my changeset to other archs and get rid of the magic 2 and 4 numbers? I think the userspace API is still an issue but that can be addressed at a later stage.

I think so, yes.

Indanz:
When #1551 gets merged it may be easier to add a new, duplicate function that uses arrays.

ZhanYF:
@Indanz 

Hi Indan, I have reworked this PR to make it work on other architectures too, regarding to magic number, the offset of 2 is still needed to avoid override resume_target and count, as part of the definition for WriteRegister. I have documented why it's needed, so hopefully it's no longer confusing.

Does this look good to you?

libsel4/tools/syscall_stub_gen.py
```diff
@@ -747,15 +763,27 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
-                    result.append("\t%s->%s = %s;" %
-                                  (param.name, members[i], words[i] % source_words))
+                if method_name == "ReadRegisters":
+                    for i in range(len(source_words)):
+                        result.append("\t%s->%s = %s;" %
+                                      (param.name, members[i], words[i] % source_words))
+                    result.append("\tif (count > %d) {" % num_mrs)
+                    result.append("\t\tfor (seL4_Word i=0; i<count-%d; i++) {" % num_mrs)
+                    result.append("\t\t\t((seL4_Word *) &%s->%s)[i] = seL4_GetMR(i+%d);" %
+                                  (param.name, members[num_mrs], num_mrs))
```
Indanz: What about getting rid of the first if-check by making it part of the for-loop conditional:
```suggestion
                    result.append("\tfor (seL4_Word i=%d; i < count; i++) {" % num_mrs)
                    result.append("\t\t((seL4_Word*)&%s->%s)[i] = seL4_GetMR(i);" %
                                  param.name, members[0]))
```

libsel4/tools/syscall_stub_gen.py
```diff
@@ -691,17 +691,33 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
-        # Initialise in-register parameters
-        for i in range(num_mrs):
-            if i < len(input_expressions):
-                result.append("\tmr%d = %s;" % (i, input_expressions[i]))
-            else:
-                result.append("\tmr%d = 0;" % i)
-        # Initialise buffered parameters
-        for i in range(num_mrs, len(input_expressions)):
-            result.append("\tseL4_SetMR(%d, %s);" % (i, input_expressions[i]))
-        result.append("")
```
Indanz, line 704: Why did you remove the last `result.append("")`?

I think you may want to keep it, it's done elsewhere too for some reason. But put it after your new for-loop.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -691,17 +691,33 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
+          for i in range(num_mrs):
+              if i < len(input_expressions):
+                  result.append("\tmr%d = %s;" % (i, input_expressions[i]))
+          # First 2 registers are always passed by mrs, thus initialise buffered
+          # parameters only when requesting more than 2 registers
+          regs_by_mrs = num_mrs - 2
+          result.append("\tif (count > %d) {" % regs_by_mrs)
+          result.append("\t\tfor (seL4_Word i=0; i<count-%d; i++) {" % regs_by_mrs)
+          result.append("\t\t\tseL4_SetMR(i+%d, ((seL4_Word *) &%s)[i]);" %
+                        (num_mrs, input_expressions[num_mrs]))
```
Indanz: What about getting rid of the first if-check by making it part of the for-loop conditional:
```suggestion
          regs_by_mrs = num_mrs - 2
          result.append("\tfor (seL4_Word i = %d; i < count; i++) {" % regs_by_mrs)
          result.append("\t\tseL4_SetMR(i + 2, ((seL4_Word*)&%s)[i]);" %
                        input_expressions[0])
```

Indanz:
Over all looks good to me, much better to only have the magic 2 offset into the MRs which is indeed unavoidable, because it's the number of arguments used for other syscall parameters. With `regs_by_mrs` it's all fine now I think.

Indanz:
Could you post an updated diff too please?

Can you fix the CI errors? For Gitlint we currently get:

```
Checking the following commits:
bbe8a0ed Use loop and respect the limit parameter for seL4_TCB_WriteRegisters/ReadRegisters.

1: T1 Title exceeds max length (83>50): "Use loop and respect the limit parameter for seL4_TCB_WriteRegisters/ReadRegisters."
1: T3 Title has trailing punctuation (.): "Use loop and respect the limit parameter for seL4_TCB_WriteRegisters/ReadRegisters."
3: B1 Line exceeds max length (182>72): "Previously TCB_WriteRegisters/ReadRegisters ignored the count parameter and unconditionally copied the entire seL4_UserContext, this leads to unnecessary overhead for VMM operations."
Error: Action gitlint failed.
```

and the style complaints about the indentation being off.

ZhanYF:
@Indanz 

Thanks for your feedbacks! Change from the last version:

- Merged the first if-condition with the for-loop.
- There is only one message register for ia32-mcs, to not break it I added a condition check to bypass if num_mrs < 2.
- Check interface_name == "seL4_TCB", otherwise it changes seL4_X86_VCPU_WriteRegisters on x86, which does not take a `count` parameter, we don't any test for it yet and I only noticed it when manually reviewing the generated diff. I think the next step would be
https://github.com/seL4/sel4test/issues/115
- New sel4_client.h diff for arm64 and x86_64: https://gist.github.com/ZhanYF/76fcdaf9264fc7fba8cbc243ba0b03d3
- Fixed style/gitlint CI (Hopefully it will pass, GitHub CI doesn't allow me to reliablly check from my side)

Indanz:
Thanks!

>     * There is only one message register for ia32-mcs, to not break it I added a condition check to bypass if num_mrs < 2.

Great, I didn't know that. Let me see if I can fix that, because that's ridiculous.

> I think the next step would be
>       [Add tests for vCPU objects sel4test#115](https://github.com/seL4/sel4test/issues/115)

Yes, this is on my TODO list. However, this will be a lot of work because we don't want to run a real OS, but have our own test code as VM to do specific tests.

>     * New sel4_client.h diff for arm64 and x86_64: https://gist.github.com/ZhanYF/76fcdaf9264fc7fba8cbc243ba0b03d3

Thanks, that looks fine.

>     * Fixed style/gitlint CI (Hopefully it will pass, GitHub CI doesn't allow me to reliablly check from my side)

You can run the `tools/seL4/misc/style*` scripts locally too. They do modify the code, so make sure you either committed or staged all your own changes first before running them.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -691,17 +691,34 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
-            result.append("\tseL4_SetMR(%d, %s);" % (i, input_expressions[i]))
-        result.append("")
+        if (method_name == "WriteRegisters" and
+            interface_name == "seL4_TCB" and
+                num_mrs >= 2):
+            result.append("\t/* Marshal and initialise parameters with limiter. */")
+            # Initialise in-register parameters
+            for i in range(num_mrs):
+                if i < len(input_expressions):
+                    result.append("\tmr%d = %s;" % (i, input_expressions[i]))
```
Indanz: I liked your previous version where you didn't duplicate this bit better, like you do for ReadRegisters.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -691,17 +691,34 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
-                result.append("\tmr%d = %s;" % (i, input_expressions[i]))
-            else:
-                result.append("\tmr%d = 0;" % i)
-        # Initialise buffered parameters
-        for i in range(num_mrs, len(input_expressions)):
-            result.append("\tseL4_SetMR(%d, %s);" % (i, input_expressions[i]))
-        result.append("")
+        if (method_name == "WriteRegisters" and
+            interface_name == "seL4_TCB" and
+                num_mrs >= 2):
```
Indanz: If you create a local var to hold this expression, it's out of the way and you can use the shorter variable new instead below.

libsel4/tools/syscall_stub_gen.py
```diff
@@ -747,15 +764,25 @@ def generate_stub(arch, wordsize, interface_name, method_name, method_id, input_
...
                 members = struct_members(param.type, structs)
-                for i in range(len(words)):
-                    result.append("\t%s->%s = %s;" %
-                                  (param.name, members[i], words[i] % source_words))
+                if method_name == "ReadRegisters" and num_mrs >= 2:
+                    for i in range(len(source_words)):
+                        result.append("\t%s->%s = %s;" %
+                                      (param.name, members[i], words[i] % source_words))
+                    result.append("\tfor (seL4_Word i=%d; i<count; i++) {" % num_mrs)
+                    result.append("\t\t((seL4_Word *) &%s->%s)[i-%d] = seL4_GetMR(i);" %
```
Indanz:
```suggestion
                    result.append("\tfor (seL4_Word i = %d; i < count; i++) {" % num_mrs)
                    result.append("\t\t((seL4_Word *) &%s->%s)[i - %d] = seL4_GetMR(i);" %
```
Please add some space.

ZhanYF:
@Indanz 

Thanks for the new feedbacks, changes from the last version:
- removed duplicate logic in the if-section
- added some spaces to the generated C code
- moved loop check conditions into a local var
- new diff for x86/aarch64 (only style changes since the last version, as expected)
- https://gist.github.com/ZhanYF/0c3f93353329157c05df0999bc8e8530

Indanz (approved):
Good enough.

Indanz:
Thanks for your patience and dedication getting this merged!

PR #1491, opened 2025-07-04
manual/README.md: include API doc generation

lsf37:
Migrate documentation of the API docs generation process here from the docsite where it is too much out of view. Here it is closer to where it is needed and more likely to get updated when things change.

Compared to the docsite version, this has applied:

- standard wrap column
- markdown lints
- adjusted headings
- `object-api-sel4-arch.xml` etc renames
- include RISC-V in structure

### DIFF
--- a/manual/README.md
+++ b/manual/README.md
@@ -4,14 +4,199 @@
      SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
-seL4 Reference Manual
-=====================
+# seL4 Reference Manual
+
+## Build
 
 To build a PDF use
 
     make
 
-You will need recent version of `LaTeX` and `doxygen`.
+You will need recent version of `LaTeX` and `doxygen` as well as the seL4 python
+dependencies.
 
 The main source file is `manual.tex`, most of the text is in `parts/`,
 and most of the API reference is generated with `doxygen`.
+
+## API Generation
+
+The documentation of the seL4 API is automatically generated from
+comments in the source code. This section documents the process.
+
+### Types of API
+
+seL4 has two types of API, system calls and object invocations.
+
+- Top-level system calls are generally for message passing, for example `Call`,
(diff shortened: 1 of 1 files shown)


manual/README.md
```diff
@@ -4,14 +4,196 @@
...
-You will need recent version of `LaTeX` and `doxygen`.
+You will need recent version of `LaTeX` and `doxygen` as well as the seL4 python
+dependencies.
 
 The main source file is `manual.tex`, most of the text is in `parts/`,
 and most of the API reference is generated with `doxygen`.
+
+## API Generation
+
+The documentation of the seL4 API in them manual is automatically generated from
```
Indanz:
```suggestion
The documentation of the seL4 API in the manual is automatically generated from
```
Or leave out the "in the manual" altogether, as that's obvious from the context.

manual/README.md
```diff
@@ -4,14 +4,196 @@
...
+comments in the source code. This section documents the process.
+
+### Types of API
+
+seL4 has two types of API:
+
+- System Calls, mostly concerned with message-passing between threads. Some
+  examples are `Send` and `Recv`
```
Indanz: Aren't `Send` and `Recv` just object invocations on endpoints or notifications? Took me forever to figure that out.

Indanz (approved):

manual/README.md
lsf37: I've also always found that distinction a bit strange.`Call` etc are also just object invocations in the end. The only syscall that is not an object invocation is `Yield` (and maybe debug + benchmarking things). I don't think we should try to change that terminology in a transfer PR :-)

lsf37: Although, re-reading this part, it could be made at least a little clearer. Will attempt that.

lsf37: Alright, have tried that. Is that any better?

Indanz: Yes, that's better, thanks.

PR #1492, opened 2025-07-07
bitgield_gen: remove manual table of contents

lsf37:
Remove manual table of contents for rendering on the docsite, which will provide an automatic TOC.

(It was not that super useful on GitHub either)


Indanz:
Where on the docsite is/was the bitfield manual? I can't find it.

PR #1494, opened 2025-07-14
manual: markdown generation tweaks

lsf37:
- fix python string escape warnings
- minor python lints and typos
- use non-breaking space for types in markdown tables

The latter avoids breaking `seL4_Word *` into two lines in tables. Uses unicode `0u000A`, because `&nbsp;` does not always get rendered correctly and `0u000A` is just passed through.

### DIFF
--- a/manual/tools/parse_doxygen_xml.py
+++ b/manual/tools/parse_doxygen_xml.py
@@ -7,7 +7,7 @@
 
 """
 Script for generating latex from doxygen-generated xml files.
-The generatetd latex files are compatible with the seL4 manual.
+The generated latex files are compatible with the seL4 manual.
 """
 import argparse
 import sys
@@ -585,7 +585,7 @@ def generate_general_syscall_doc(generator, input_file_name, level, ref_dict):
     """
     Takes a path to a file containing doxygen-generated xml,
     and return a string containing latex suitable for inclusion
-    in the sel4 manual.
+    in the seL4 manual.
     """
 
     dir_name = os.path.dirname(input_file_name)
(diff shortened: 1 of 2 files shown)


manual/tools/parse_doxygen_xml.py
```diff
@@ -503,9 +503,15 @@ def generate_params(self, param_string):
...
+    def make_nbsp(self, string):
+        """Make all spaces non-breaking"""
+
+        # Use unicode non-breaking space. HTML &nbsp; is not valid in all Markdown flavours.
+        return string.replace(" ", "\u00a0")
+
     def generate_param_string(self, param_info, param_name):
         return "`%(type)s` | `%(name)s` | %(desc)s\n" % {
-            "type": self.get_text(param_info["type"], escape=False),
+            "type": self.make_nbsp(self.get_text(param_info["type"], escape=False)),
```
Indanz, line 514: Not sure I like this, what happens if you copy and paste these lines? I don't see the problem of breaking lines.

Indanz (approved):
Not sure about the non-breaking spaces commit, but rest is fine.

manual/tools/parse_doxygen_xml.py
lsf37: Without it, the line breaks like this even on wide screens:

```
seL4_Word | next column
*         |
```

lsf37: Let me check what happens when I copy the rendered text

lsf37: It gets copied as normal space from the browser, so should be fine for copy/pasting type info out of the docs.

Indanz: > Without it, the line breaks like this even on wide screens:

Well, that's a sign our text is too wide. Perhaps we have too much indentation?

lsf37: It's just markdown table layout -- the width of columns can't be controlled from just markdown, it depends on the size of the other columns and descriptions are long, so they will always overpower short type columns.

We could try to insert explicit layout, but then we should really be generating html instead of markdown.

Indanz: I looked on the API website and see what you mean. Non-breaking spaces is probably as good as it gets.

PR #1495, opened 2025-07-22
README: update links for new docsite structure

lsf37:
README.md:
- avoid using non-extension URLs, because they depend on how the site is hosted
- update links to point into the new structure

CHANGES.md:
- update releases link


corlewis (approved):
Looks good to me!

Issue #1008, opened 2023-04-04
Print memory regions as `[..)`


lsf37:
At boot time we print memory regions as `[a..b]` even though `b` is not included in the interval.

This has confused me a number of times in the past, I suspect it is the same for others.

Correct [interval notation](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals) would be `[a..b)`.

PR #1497, opened 2025-07-28
boot: check for size_bits=0 before calculating BIT(size_bits)

midnightveil:
This would previously create regions that were off-by-1 on the upper end with addresses such as 0x40001. This seems to have been fine, and just created an extra unnecessary memory region covering the extra bootinfo memory. But at the same time this was inconsistent with e.g. calculate_rootserver_size() which contained the check:

  size += extra_bi_size_bits > 0 ? BIT(extra_bi_size_bits) : 0;

and this created issues when I was experimenting with some boot code changes as the bootinfo size accounting was inconsistent.

### DIFF
--- a/include/kernel/boot.h
+++ b/include/kernel/boot.h
@@ -58,7 +58,7 @@ void create_domain_cap(cap_t root_cnode_cap);
 cap_t create_ipcbuf_frame_cap(cap_t root_cnode_cap, cap_t pd_cap, vptr_t vptr);
 word_t calculate_extra_bi_size_bits(word_t extra_size);
 void populate_bi_frame(node_id_t node_id, word_t num_nodes, vptr_t ipcbuf_vptr,
-                       word_t extra_bi_size_bits);
+                       word_t extra_bi_size);
 void create_bi_frame_cap(cap_t root_cnode_cap, cap_t pd_cap, vptr_t vptr);
 
 #ifdef CONFIG_KERNEL_MCS
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -416,7 +416,7 @@ static BOOT_CODE bool_t try_init_kernel(
     word_t extra_bi_size_bits = calculate_extra_bi_size_bits(extra_bi_size);
     v_region_t it_v_reg = {
         .start = ui_v_reg.start,
-        .end   = extra_bi_frame_vptr + BIT(extra_bi_size_bits)
+        .end   = extra_bi_frame_vptr + (extra_bi_size_bits > 0 ? BIT(extra_bi_size_bits) : 0)
     };
     if (it_v_reg.end >= USER_TOP) {
         /* Variable arguments for printf() require well defined integer types to
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -276,7 +276,7 @@ static BOOT_CODE bool_t try_init_kernel(
     word_t extra_bi_size_bits = calculate_extra_bi_size_bits(extra_bi_size);
     v_region_t it_v_reg = {
         .start = ui_v_reg.start,
-        .end   = extra_bi_frame_vptr + BIT(extra_bi_size_bits)
+        .end   = extra_bi_frame_vptr + (extra_bi_size_bits > 0 ? BIT(extra_bi_size_bits) : 0)
     };
     if (it_v_reg.end >= USER_TOP) {
         /* Variable arguments for printf() require well defined integer types
--- a/src/arch/x86/kernel/boot.c
+++ b/src/arch/x86/kernel/boot.c
@@ -142,7 +142,7 @@ BOOT_CODE bool_t init_sys_state(
 
     /* The region of the initial thread is the user image + ipcbuf and boot info */
     it_v_reg.start = ui_v_reg.start;
-    it_v_reg.end = ROUND_UP(extra_bi_frame_vptr + BIT(extra_bi_size_bits), PAGE_BITS);
+    it_v_reg.end = ROUND_UP(extra_bi_frame_vptr + (extra_bi_size_bits > 0 ? BIT(extra_bi_size_bits) : 0), PAGE_BITS);
 #ifdef CONFIG_IOMMU
     /* calculate the number of io pts before initialising memory */
     if (!vtd_init_num_iopts(num_drhu)) {
@@ -176,7 +176,7 @@ BOOT_CODE bool_t init_sys_state(
     populate_bi_frame(0, ksNumCPUs, ipcbuf_vptr, extra_bi_size);
     region_t extra_bi_region = {
         .start = rootserver.extra_bi,
-        .end = rootserver.extra_bi + BIT(extra_bi_size_bits)
+        .end = rootserver.extra_bi + (extra_bi_size_bits > 0 ? BIT(extra_bi_size_bits) : 0)
     };
 
     /* populate vbe info block */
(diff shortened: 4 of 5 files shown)


Indanz (dismissed):

lsf37:
This looks like a bug to me, there is a reason 2^0 is defined to be 1  -- either `end` is inclusive then it is now correct for 0 and wrong for all other cases, or the end is not inclusive, then it is now wrong for 0. Or is there something else further up in the code that subtracts 1? But it's still inconsistent between the 0 case and other cases.

midnightveil:
I don't think I understand. I'm pretty sure the accounting is done inclusively, that's how it's printed and it's how some of the checks are written to check overflow. If it was exclusive, then it still doesn't work anyhow. (*I think I've messed up the meaning of inclusive and exclusive, [0x1000...0x2000] i e. how it's written in prints is an inclusive range notation but actually the upper address is exclusive not inclusive)

If there's no extra bootinfo: then extra bootinfo is going to be exactly the inclusive end value of the bootinfo region.

e.g. user task range: [0x10000..0x30000]
bootinfo range: [0x30000..0x40000]

if extra bootinfo size = 0, then the bits is 0 (because we special case that too inside our essentially clog2 impl), and so the extra bootinfo range should be 0x40000..0x40000.

if the extra boot info is say, 0xf00 in size, then the bits is going to be 12, for a power of 2 size of 0x1000. then the extra bootinfo range is 0x40000..0x41000.

the overall range computed is then 0x10000..0x40000.

if we don't do this, then the BIT(12) is still 0x10000, but because BIT(0) is 1 then we get 0x40000...0x40001, which is wrong.

The issue is that BIT(clog2(0)) == 1, it's not consistent. If we use extra_bi_size instead of BIT(extra_bi_size_bits) then it'd be fine but then we'd have some mismatches in size when things are not power of 2.

EDIT: No, I see, if in this case we're excluding the upper address then 0x40000...0x40001 is an empty range (confusingly). However I think some of the later code doesn't think this is true ... I vaguely remember seeing an issue about how the boot code was confused between exclusive/inclusive range and flipped between them.

midnightveil:
The actual issue is probably then the get_n_paging function (and possibly others):

https://github.com/seL4/seL4/blob/f13f37a6a7bd74a43ca745f5a7565c063e74a633/include/kernel/boot.h#L122-L127

This rounds up the end value to the bits size; and so the [0x40000, 0x40001) empty range instead becomes [0x40000, 0x41000) which means an extra page is allocated.

I guess the boot code is probably really confused in various places and that this fix is more a "hot fix" that just papers over the problem more by making the empty range appear more empty to badly written other code.

EDIT again: Actually, no. Rust's clopen ranges define is_empty as start >= end, because if you have say [0, 1) then it's not an empty range but instead a range containing a singular value.

Another way of phrasing this is that $clog2 isn't invertible, we lose information between the 1/0 cases

Indanz:
> if extra bootinfo size = 0, then the bits is 0 (because we special case that too inside our essentially clog2 impl), and so the extra bootinfo range should be 0x40000..0x40000.

Yes, this is the problem: You can't encode 0 with a log value.

> EDIT: No, I see, if in this case we're excluding the upper address then 0x40000...0x40001 is an empty range (confusingly). 

No, it's not empty, it's one byte big, just as you would expect for something one byte big.

> This looks like a bug to me, there is a reason 2^0 is defined to be 1 -- either `end` is inclusive then it is now correct for 0 and wrong for all other cases, or the end is not inclusive, then it is now wrong for 0. Or is there something else further up in the code that subtracts 1? But it's still inconsistent between the 0 case and other cases.

Yes, it's inconsistent, but I think it's correct, because extra_bi_size_bits 0 means 0, not 1 here. It would be clearer if we skip the whole thing if `extra_bi_size_bits` is 0 instead of this.

axel-h:
That change make sense.

While looking though how `extra_bi_size_bits` is used, there seems another small quirk in https://github.com/seL4/seL4/blob/f13f37a6a7bd74a43ca745f5a7565c063e74a633/include/kernel/boot.h#L61 
`
void populate_bi_frame(node_id_t node_id, word_t num_nodes, vptr_t ipcbuf_vptr,
                       word_t extra_bi_size_bits);
`
should actually be
`
void populate_bi_frame(node_id_t node_id, word_t num_nodes, vptr_t ipcbuf_vptr,
                       word_t extra_bi_size);
`
But it's just the prototype definition that has the wrong name, we pass `extra_bi_size` everywhere.

lsf37:
Ok, so `extra_bi_size_bits` that we are doing this check on comes directly out of `calculate_extra_bi_size_bits`, which either returns something >= `seL4_PageBits ` or 0. I don't like it mathematically, but in this case indicating size 0 with bits = 0 does work, because size = 1 cannot happen. And you are catching the intention of `calculate_extra_bi_size_bits` correctly with the check.

So, assuming the interval is well-formed to begin with (inclusive or exclusive), we're either adding a full page (or a multiple of that) to the end or nothing. That does preserve whatever format it was expecting before.

lsf37 (approved):
You have convinced me.

lsf37:
I think we should add a comment, though (that `calculate_extra_bi_size_bits` returns 0 for size 0, and that non-empty sizes are always >= seL4_PageBits)

midnightveil:
> I think we should add a comment, though (that calculate_extra_bi_size_bits returns 0 for size 0, and that non-empty sizes are always >= seL4_PageBits

~~I think this is probably the case for now (we only use it to allocate extra bootinfo frames)~~

~~But there's some other code that compares it to the MinUntyped size (I think it's maybe_alloc_extra_bi).~~

Oh yeah there is round up

lsf37:
> `void populate_bi_frame(node_id_t node_id, word_t num_nodes, vptr_t ipcbuf_vptr, word_t extra_bi_size_bits);`
> should actually be
> `void populate_bi_frame(node_id_t node_id, word_t num_nodes, vptr_t ipcbuf_vptr, word_t extra_bi_size);`
> But it's just the prototype definition that has the wrong name, we pass `extra_bi_size` everywhere.

Agreed, that would be good to change. There's no bug, but the name in the prototype is misleading.

lsf37:
>  [0x1000...0x2000] i e. how it's written in prints is an inclusive range notation but actually the upper address is exclusive not inclusive)

I would really prefer if it printed `[0x1000...0x2000)` instead. People seem to not be used to that notation, but at least it wouldn't be actively misleading. Or we print `[0x1000...0x1FFF]`, that'd also be fine and looks maybe less weird to normal people.

midnightveil:
> I would really prefer if it printed [0x1000...0x2000) instead. People seem to not be used to that notation, but at least it wouldn't be actively misleading. Or we print [0x1000...0x1FFF]

There's also a C proposal for `0x1000..<0x2000`. (https://thephd.dev/_vendor/future_cxx/papers/C%20-%20Additional%20Half-Open%20Case%20Range%20Syntax.html)

But yeah, I'd be happy to put something through that changes it to be `)` (oh, I see https://github.com/seL4/seL4/issues/1008)

lsf37:
Thanks for the changes. I had forgotten about the issue #1008 :-) I could live with `[a..<b]`, that's also relatively clear.

Indanz:
> > [0x1000...0x2000] i e. how it's written in prints is an inclusive range notation but actually the upper address is exclusive not inclusive)
> 
> I would really prefer if it printed `[0x1000...0x2000)` instead. People seem to not be used to that notation, but at least it wouldn't be actively misleading.

We got that in high school, so it's not that strange to me, compared to any alternative weird notations. But then it should be `[0x1000, 0x2000)` I guess. I think in computer science they usually just use` 0x1000-0x2000` and it's almost always exclusive end so that `end - start` gives the size.

PR #1502, opened 2025-07-31
aarch64,vspace: move PUD/PGD typedefs to deprecated code

### LINKED ISSUE #1498: seL4_ARM_Page{Upper,Global]Directory typedefs should be moved to deprecated.h
https://github.com/seL4/seL4/blob/f13f37a6a7bd74a43ca745f5a7565c063e74a633/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/types.h#L11-L12

With the last release deprecating PageUpperDirectory/PageGlobalDirectory as aliases for PageTable then these should be moved to deprecated.h.

There's also a comment about PUD/PGD which I think is incorrect.

midnightveil:
This was missed in cb8ee83f0cd2268b1d6b1cd5ca30b031db5896c4; their use was removed from syscall_stub_gen but their typedefs remained.

Fixes #1498.

### DIFF
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
@@ -28,6 +28,10 @@
 #define seL4_PUDBits 13
 #define seL4_PUDIndexBits 10
 #define seL4_ARM_PageUpperDirectoryObject seL4_ARM_VSpaceObject
+#define seL4_ARM_PageUpperDirectory seL4_ARM_VSpace
+
+/* this was previously defined always, so keep that here for consistency */
+#define seL4_ARM_PageGlobalDirectory seL4_CPtr
 
 #else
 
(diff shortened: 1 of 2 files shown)


lsf37 (approved):

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
```diff
@@ -28,6 +28,10 @@
 #define seL4_PUDBits 13
 #define seL4_PUDIndexBits 10
 #define seL4_ARM_PageUpperDirectoryObject seL4_ARM_VSpaceObject
+#define seL4_ARM_PageUpperDirectory seL4_ARM_VSpace
+
+/* this was previously defined always, so keep that here for consistency */
+#define seL4_ARM_PageGlobalDirectory seL4_CPtr
```
Indanz, line 34: This can be `seL4_ARM_PageTable`. But as it all ends up the same, there's no type checking anyway and it doesn't really matter.

Indanz (approved):
Not entirely sure why we have typedefs in one and defines in the other, other than for consistency and historical reasons.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h
midnightveil: It's also never used (I think) when the VSpace is a PUD, so it really shouldn't matter.

PR #1504, opened 2025-08-01
asidpools: macro consistency

midnightveil:
In theory, this shouldn't affect any of the proofs, as the preprocessor output should be almost identical to previous (This is part of the motivation for changing the definition of `ASID_HIGH` to be consistent with the users of the macro).

There's also a bug (lying in wait, so not actually a bug) with the initial thread asid pool cap where we weren't actually creating it as an "asid base", but it was fine since it was 0 anyway.

### DIFF
--- a/include/arch/arm/arch/32/mode/object/structures.h
+++ b/include/arch/arm/arch/32/mode/object/structures.h
@@ -104,14 +104,13 @@ typedef struct asid_pool asid_pool_t;
 #define nASIDPools BIT(asidHighBits)
 
 #define ASID_LOW(a) (a & MASK(asidLowBits))
-#define ASID_HIGH(a) ((a >> asidLowBits) & MASK(asidHighBits))
+#define ASID_HIGH(a) ((a) >> asidLowBits)
 
 static inline cap_t CONST cap_small_frame_cap_set_capFMappedASID(cap_t cap, word_t asid)
 {
     cap = cap_small_frame_cap_set_capFMappedASIDLow(cap,
                                                     asid & MASK(asidLowBits));
-    return cap_small_frame_cap_set_capFMappedASIDHigh(cap,
-                                                      (asid >> asidLowBits) & MASK(asidHighBits));
+    return cap_small_frame_cap_set_capFMappedASIDHigh(cap, ASID_HIGH(asid));
 }
 
 static inline word_t CONST cap_small_frame_cap_get_capFMappedASID(cap_t cap)
@@ -124,8 +123,7 @@ static inline cap_t CONST cap_frame_cap_set_capFMappedASID(cap_t cap, word_t asi
 {
     cap = cap_frame_cap_set_capFMappedASIDLow(cap,
                                               asid & MASK(asidLowBits));
-    return cap_frame_cap_set_capFMappedASIDHigh(cap,
-                                                (asid >> asidLowBits) & MASK(asidHighBits));
+    return cap_frame_cap_set_capFMappedASIDHigh(cap, ASID_HIGH(asid));
 }
 
 static inline word_t CONST cap_frame_cap_get_capFMappedASID(cap_t cap)
@@ -432,3 +430,4 @@ static inline word_t PURE pte_ptr_get_pteType(pte_t *pte_ptr)
     }
 }
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
+
--- a/include/arch/arm/arch/32/mode/object/structures.h
+++ b/include/arch/arm/arch/32/mode/object/structures.h
@@ -99,7 +99,6 @@ typedef struct asid_pool asid_pool_t;
 
 #define HW_ASID_SIZE_BITS 1
 
-#define ASID_POOL_INDEX_BITS seL4_ASIDPoolIndexBits
 #define ASID_BITS (asidHighBits+asidLowBits)
 
 #define nASIDPools BIT(asidHighBits)
@@ -433,4 +432,3 @@ static inline word_t PURE pte_ptr_get_pteType(pte_t *pte_ptr)
     }
 }
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
-
--- a/include/arch/arm/arch/32/mode/object/structures.h
+++ b/include/arch/arm/arch/32/mode/object/structures.h
@@ -103,13 +103,12 @@ typedef struct asid_pool asid_pool_t;
 
 #define nASIDPools BIT(asidHighBits)
 
-#define ASID_LOW(a) (a & MASK(asidLowBits))
+#define ASID_LOW(a) ((a) & MASK(asidLowBits))
 #define ASID_HIGH(a) ((a) >> asidLowBits)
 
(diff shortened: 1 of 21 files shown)


include/arch/arm/arch/32/mode/object/structures.h
```diff
@@ -99,18 +99,21 @@ typedef struct asid_pool asid_pool_t;
...
 
 #define nASIDPools BIT(asidHighBits)
 
 #define ASID_LOW(a) (a & MASK(asidLowBits))
-#define ASID_HIGH(a) ((a >> asidLowBits) & MASK(asidHighBits))
+/* This *could* have been defined as ((a >> asidLowBits) & MASK(asidHighBits))
+ * but the compiler does not have enough information to be able to elide the
+ * mask operation, and ASIDs are only ever used internally to the kernel.
+ */
+#define ASID_HIGH(a) (a >> asidLowBits)
```
Indanz:
```suggestion
#define ASID_HIGH(a) ((a) >> asidLowBits)
```
If you're changing this anyway...

include/arch/arm/arch/32/mode/object/structures.h
```diff
@@ -433,4 +436,3 @@ static inline word_t PURE pte_ptr_get_pteType(pte_t *pte_ptr)
     }
 }
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
-
```
Indanz: This shouldn't be part of this commit.

include/arch/arm/arch/32/mode/object/structures.h
```diff
@@ -99,18 +99,21 @@ typedef struct asid_pool asid_pool_t;
...
 #define ASID_BITS (asidHighBits+asidLowBits)
 
 #define nASIDPools BIT(asidHighBits)
 
 #define ASID_LOW(a) (a & MASK(asidLowBits))
-#define ASID_HIGH(a) ((a >> asidLowBits) & MASK(asidHighBits))
+/* This *could* have been defined as ((a >> asidLowBits) & MASK(asidHighBits))
+ * but the compiler does not have enough information to be able to elide the
+ * mask operation, and ASIDs are only ever used internally to the kernel.
+ */
```
Indanz: Comment is misleading as it implies that the compiler would be allowed to elide the mask.

Before, top bits not part of the ASID were ignored, now we assume we only get a complete and valid (unsigned) ASID value passed, so the behaviour has actually changed.

src/arch/arm/32/kernel/vspace.c
```diff
@@ -591,7 +591,7 @@ BOOT_CODE void activate_kernel_vspace(void)
 BOOT_CODE void write_it_asid_pool(cap_t it_ap_cap, cap_t it_pd_cap)
 {
     asid_pool_t *ap = ASID_POOL_PTR(pptr_of_cap(it_ap_cap));
-    ap->array[IT_ASID] = PDE_PTR(pptr_of_cap(it_pd_cap));
+    ap->array[ASID_LOW(IT_ASID)] = PDE_PTR(pptr_of_cap(it_pd_cap));
```
Indanz, line 594: This change should be part of the bugfix commit.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -567,7 +567,7 @@ BOOT_CODE void write_it_asid_pool(cap_t it_ap_cap, cap_t it_vspace_cap)
                               , 0, false
 #endif
                           );
-    ap->array[IT_ASID] = asid_map;
+    ap->array[ASID_LOW(IT_ASID)] = asid_map;
```
Indanz, line 570: Same here.

src/arch/riscv/kernel/vspace.c
```diff
@@ -310,7 +310,7 @@ BOOT_CODE void activate_kernel_vspace(void)
 BOOT_CODE void write_it_asid_pool(cap_t it_ap_cap, cap_t it_lvl1pt_cap)
 {
     asid_pool_t *ap = ASID_POOL_PTR(pptr_of_cap(it_ap_cap));
-    ap->array[IT_ASID] = PTE_PTR(pptr_of_cap(it_lvl1pt_cap));
+    ap->array[ASID_LOW(IT_ASID)] = PTE_PTR(pptr_of_cap(it_lvl1pt_cap));
```
Indanz, line 313: Idem.

src/arch/x86/kernel/vspace.c
```diff
@@ -529,7 +529,7 @@ BOOT_CODE bool_t init_pat_msr(void)
 BOOT_CODE void write_it_asid_pool(cap_t it_ap_cap, cap_t it_vspace_cap)
 {
     asid_pool_t *ap = ASID_POOL_PTR(pptr_of_cap(it_ap_cap));
-    ap->array[IT_ASID] = asid_map_asid_map_vspace_new(pptr_of_cap(it_vspace_cap));
+    ap->array[ASID_LOW(IT_ASID)] = asid_map_asid_map_vspace_new(pptr_of_cap(it_vspace_cap));
```
Indanz, line 532: .

Indanz (approved):
Need to check all callers to double check dropping the mask is fine, but it probably is.

include/arch/arm/arch/32/mode/object/structures.h
midnightveil: Maybe I should have said, "as ASIDs are only ever used internally", not "and". 

I was trying to say that the behaviour being changed is OK. This was reflecting a comment by @lsf37 that adding this everywhere would be adding a bunch of masks everywhere that aren't actually needed (because it's replacing a bunch of spots that didn't have a mask).

midnightveil:
proof-test run: https://github.com/seL4/seL4/actions/runs/16712382410?pr=1504

Not sure what that RISC-V failure is about, I don't think I touched anything with IRQ caps or TCBs.

Indanz:
> proof-test run: https://github.com/seL4/seL4/actions/runs/16712382410?pr=1504
> 
> Not sure what that RISC-V failure is about, I don't think I touched anything with IRQ caps or TCBs.

Agreed it seems unrelated.

include/arch/arm/arch/32/mode/object/structures.h
Indanz: Just remove the comment altogether, it just confuses things as it is and doesn't add anything useful. Just say whatever you want to say in the commit message.

lsf37:
> proof-test run: https://github.com/seL4/seL4/actions/runs/16712382410?pr=1504
> 
> Not sure what that RISC-V failure is about, I don't think I touched anything with IRQ caps or TCBs.

It is indeed unrelated. We broke the MCS C proofs with merging the SGI and FPU PRs and forgetting that they are changing MCS as well, of course. We need to merge the proofs over to MCS, but that will take a while, because there is a large proof refactor going on that is making the merge hard.

In short: fine to ignore here.

include/arch/arm/arch/32/mode/object/structures.h
midnightveil: Removed.

PR #1505, opened 2025-08-04
bitfield_gen: improve readability of union funcs for sliced tags

midnightveil:
This makes it a bit easier to tell what's going on, I think, especially if these asserts trip off. 

Before:

    static inline uint64_t PURE
    pte_pte_table_ptr_get_pt_base_address(pte_t *pte_ptr) {
        uint64_t ret;
        assert(((pte_ptr->words[0] >> 0) & 0x400000000000003) ==
               0x3ull);

After:

    static inline uint64_t PURE
    pte_pte_table_ptr_get_pt_base_address(pte_t *pte_ptr) {
        uint64_t ret;
        /* fail if union does not have the expected tag */
        assert(((pte_ptr->words[0] >> 0) & 0x400000000000003) == 0x3ull /* sliced tag pte_pte_table */);

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -945,10 +945,10 @@ proof_templates = {
     'get_tag_spec': [
         '''lemma (in ''' + loc_name + ''') %(name)s_get_%(tagname)s_spec:
   "\\<forall>s. \\<Gamma> \\<turnstile> {s}
-       \\<acute>%(ret_name)s :== ''' \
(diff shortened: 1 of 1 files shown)


Indanz (approved):
Can't judge the generated proof changes, but the rest looks fine.

midnightveil:
Proofs shouldn't have changed aside from what the python stylechecker asked for.

PR #1506, opened 2025-08-20
arm: flush vcpu when switching and setting domain 

corlewis:
This avoids information about VCPU state being leaked across domains.


include/arch/arm/arch/object/vcpu.h
```diff
@@ -204,9 +206,12 @@ static inline VPPIEventIRQ_t irqVPPIEventIndex(irq_t irq)
 
 #else /* end of CONFIG_ARM_HYPERVISOR_SUPPORT */
 
-/* used in boot.c with a guard, use a marco to avoid exposing vcpu_t */
+/* used in boot.c with a guard, use a macro to avoid exposing vcpu_t */
```
Indanz: I'd remove the whole comment to be honest, because it is misleading. The compile error for exposing vcpu_t is easily fixed by always declaring `typedef struct vcpu vcpu_t;`. But then you'll run into the real issue of `vcpu_boot_init()` not existing if HYP is disabled.

corlewis:
> Conceptually it's very similar to the lazy FPU state loading/unloading, but somehow that code manages to re-use existing functions, while existing VCPU functions just does slightly the wrong things, making something new necessary.
> 
> Not entirely happy with the amount of extra wrapper code and fluff introduced, but unsure how to reduce it without bigger changes. If we had a cross-platform `CONFIG_HYPERVISOR_SUPPORT` setting and standardised the API, like we did for FPU, we wouldn't need `Arch_prepareNextDomain` and `Arch_prepareSetDomain`.

I completely agree with both of your points. `vcpu_flush` was the best that I could come up with for reusing existing functions, and the change was originally a fair bit smaller before I had to add all of the fluff to make it work for all configurations. A rewrite like you're suggesting would definitely be very nice.

include/arch/arm/arch/object/vcpu.h
corlewis: I'm happy to remove the comment, I have to admit that I wasn't even thinking about what it said, I just saw the typo and wanted to fix it.

Indanz (approved):

PR #1508, opened 2025-08-20
Make fastpath_restore match restore_user_context

Indanz:
Mostly call c_exit_hook() with global lock held.

It got more out of sync since the SMP fixes got merged, those missed the fastpath code.

### DIFF
--- a/include/arch/arm/arch/32/mode/fastpath/fastpath.h
+++ b/include/arch/arm/arch/32/mode/fastpath/fastpath.h
@@ -118,13 +118,13 @@ static inline int fastpath_reply_cap_check(cap_t cap)
 /** DONT_TRANSLATE */
 static inline void NORETURN FORCE_INLINE fastpath_restore(word_t badge, word_t msgInfo, tcb_t *cur_thread)
 {
-    NODE_UNLOCK;
-
     c_exit_hook();
 
 #ifdef ARM_CP14_SAVE_AND_RESTORE_NATIVE_THREADS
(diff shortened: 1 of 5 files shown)


axel-h:
Nice to see this getting aligned finally.  I have also been wondering about this and the guarantees the called functions in the restore process have actually.

midnightveil:
Maybe some of this is straightforward if you're more familiar with this code but I'd like to see a comment or somesuch explaining the ordering of these operations. These changes *seem* to make sense but I'm not confident enough to know the requirements of correctness here; which is where I think a comment somewhere would help.

Indanz:
> Maybe some of this is straightforward if you're more familiar with this code but I'd like to see a comment or somesuch explaining the ordering of these operations. These changes _seem_ to make sense but I'm not confident enough to know the requirements of correctness here; which is where I think a comment somewhere would help.

In the case of fastpath code, the order should preferably match the slow path code. So what I did was compare the `fastpath_restore` with `restore_user_context` implementations for each architecture and 32/64 bit combination and fixed the discrepancies.

As for the order in the slowpath code, some of it is obvious, others don't matter, and who knows about the rest. I agree comments would be great, but they should be in `restore_user_context()`, not just added to `fastpath_restore`.

lsf37:
Need to double-check the preprocess failure, but it should be fine.

kent-mcleod:
This also may impact the issue relating to this PR: https://github.com/seL4/seL4/pull/490

PR #1510, opened 2025-08-31
verification tweaks for supporting remaining Arm platforms

lsf37:
Small changes for supporting verification on the remaining Arm platforms. After this PR, all Arm platforms are supported by verification.

- add cmake verified configs for `omap3`, `am335c`, and `bcm2837`.
- make HAVE_SET_TRIGGER visible to verification by using `if` instead of `#ifdef` and wrapping it in `config_set`
- improve error message for Arm platforms that do not support SGI (where NUM_SGIS = 0). This also makes verification easier, because it shortcuts the range check reasoning for those platforms.
- complete the SGI interface for platforms that do not support SGI (use empty implementations that are not called). The verification tool chain needs a function body to be defined for the prototype even if the function is not called. 
- handle spurious IRQ in preemption: the information flow proof requires that interrupt kernel entry and interrupt at preemption behave the same. All platforms with verification support so far had an empty body for `handleSpuriousIRQ`, but the new platforms do not, so we do now need to add the call to satisfy the requirement. Factor out spurious IRQ condition into its own function to reduce code duplication.
- avoid `continue` statement for empty loop body, because the Isabelle end of the binary verification tool chain does not support it


include/drivers/irq/am335x.h
```diff
@@ -119,8 +119,20 @@ static inline void handleSpuriousIRQ(void)
     dsb();
 }
 
+void setIRQTrigger(irq_t irq, bool_t trigger)
+{
+    /* Unreachable; not supported on this platform. */
+    halt();
```
Indanz: Why the halt though? That just increases the binary size for no reason.

include/object/interrupt.h
```diff
@@ -21,7 +21,6 @@ void invokeIRQHandler_SetIRQHandler(irq_t irq, cap_t cap, cte_t *slot);
 void invokeIRQHandler_ClearIRQHandler(irq_t irq);
 void deletingIRQHandler(irq_t irq);
 void deletedIRQHandler(irq_t irq);
-void handleInterrupt(irq_t irq);
+static inline void maybeHandleInterrupt(void);
```
Indanz:
```suggestion
void maybeHandleInterrupt(void);
```

I'd keep the name the same or call it checkInterrupt and combine it with handleInterrupt, should be easy if `irqInvalid > maxIRQ`.

src/api/syscall.c
```diff
@@ -39,24 +39,14 @@
...
-    irq = getActiveIRQ();
-    if (IRQT_TO_IRQ(irq) != IRQT_TO_IRQ(irqInvalid)) {
-        handleInterrupt(irq);
-    } else {
-#ifdef CONFIG_IRQ_REPORTING
-        userError("Spurious interrupt!");
-#endif
-        handleSpuriousIRQ();
-    }
+    maybeHandleInterrupt();
```
Indanz: I assumed all this code duplication was there to make verification easier...

src/object/interrupt.c
```diff
@@ -295,3 +295,18 @@ void setIRQState(irq_state_t irqState, irq_t irq)
...
+    irq_t irq;
+
+    irq = getActiveIRQ();
+    if (IRQT_TO_IRQ(irq) != IRQT_TO_IRQ(irqInvalid)) {
+        handleInterrupt(irq);
+    } else {
+#ifdef CONFIG_IRQ_REPORTING
+        userError("Spurious interrupt!");
+#endif
+        handleSpuriousIRQ();
```
Indanz: The resetting IRQs that am335x and omap3 do does seem like a dubious thing to do, except if it works around an errata, but no errata is mentioned.

Indanz (approved):

include/drivers/irq/am335x.h
midnightveil: Per commit message:

> arm: default setlRQTrigger for non-GIC platform Provide a default dummy implementation for setIRQTrigger even for non-GIC platforms where this operation is n ot supported. Otherwise, the function is declared in the header, but has no implementation.
Implementing it as halt() means verificatio n will show that the function is not called.

Although I did think we had an UMREACHABLE macro

lsf37: It forces me to prove that the function is unreachable.

include/object/interrupt.h
lsf37: `irqInvalid = 0`, though, on RISC-V :-), and I do need the existing `handleInterrupt` to still exist, otherwise I need to change all proofs about it. Wrapping it in a new function means I don't have to break the old ones.

No problem removing the inline, if you think that's better, I just wanted to keep the binary diff as small as possible.

src/api/syscall.c
lsf37: In Haskell it's mentioned twice overall (once in IRQ kernel entry, and once in preemption). The preemption one got replicated multiple times in the C implementation, and this is attempting to at least mitigate that.

src/object/interrupt.c
lsf37: I've been wondering about that as well. I'll try to find the commit that introduces them, maybe there is some information there.

include/drivers/irq/am335x.h
lsf37: (We could also just make it empty, now that I've proved it, but if we ever change anything we won't be forced to prove that it's not reachable)

lsf37: Somehow @midnightveil's comment only popped up after I had replied -- the UNREACHABLE macro is only for the compiler, really, although staring at it now, maybe we could give it a spec that also forces us to prove that it is unreachable. I have the feeling things would already break if it was reachable, because it doesn't have a body, but making it explicit could be useful.

src/object/interrupt.c
lsf37: It does seem to at least not be an accident: [SELFOUR-154](https://sel4.atlassian.net/browse/SELFOUR-154)

(The assert in Haskell that the issue mentions no longer exists, verification deals with both paths now)

include/drivers/irq/am335x.h
Indanz: That sound like a better, more explicit approach than sprinkling `halt` everywhere.

This reminds me I should read commit messages more during review. I usually do, but not for quick reviews like this one.

include/object/interrupt.h
Indanz: If you make it inline, move the function to the header file or to syscall.c. But don't declare it inline and use it from different files please.

As everything gets combined into one big file it doesn't matter much probably, the compiler might generate identical code for both, so do whatever you prefer.

include/drivers/irq/am335x.h
midnightveil: > I have the feeling things would already break if it was reachable, because it doesn't have a body, but making it explicit could be useful.

For funsies, the compiler invokes undefined behaviour if code ever reaches __builtin_unreachable(), so actually what would probably happen if it's ever reachable it we cause the compiler invoke nasal demons (and probably delete the code path that the compiler thinks will arrive there).

Because it's a built-in and not actually a function; the lack of a body isn't an issue (the declaration was added to appease the C parser which doesn't know about it otherwise, and the compiler ignores it).

src/object/interrupt.c
Indanz: Ah, in OMAP's case it is just acking the IRQ. Probably the same for am335x too. So just misleading comments then.

include/object/interrupt.h
lsf37: (also easy to call it `checkInterrupt` -- the name `maybe` comes from the Haskell option type on `getActiveIRQ` and is supposed to express that it will handle the interrupt when it is valid. But `checkInterrupt` would express roughly the same, so that would be fine.)

Indanz: I'd prefer that.

include/drivers/irq/am335x.h
lsf37: Yes, it enables the compiler to make the assumption that it is actually unreachable and will let it do whatever that implies (like deleting statements that lead up to it). For the lack of body I meant that this would probably also force me to prove that it is not called, because the semantics of a missing function body on the verification side is that execution gets stuck in that case (and overall we have to show that execution never gets stuck). But I should be able to make that more explicit and not rely on that side effect of what properties we happen to prove.

include/object/interrupt.h
lsf37: Done.

include/drivers/irq/am335x.h
lsf37: Now using UNREACHABLE, which required a small tweak to the C parser, but is otherwise working well.

lsf37: Hm, apart from the fact that the CI check of course still has the old C parser and is failing. This will resolve itself when the verification PR is merged.

Indanz: Nice!

Indanz (approved):

Issue #1187, opened 2024-02-07
Should we add `-Wconversion` to compiler flags?


lsf37:
We just finished tracking down a bug in a PR where `ticks_t` was implicitly converted by the compiler to `word_t`, changing the value (overflowing 32 bit) and causing the kernel to crash.

The compiler currently doesn't warn for implicit C conversions that my change value or sign. Verification will catch these, but only when the code is actually in the verified subset (the example was for MCS, where the C proofs are still in progress and currently only for 64 bit, where `word_t` and `ticks_t` is equal).

@Xaphiosis points out that manually adding `-Wconversion` to the compiler flags would catch these much earlier. And arguably we should not have any implicit conversions in the kernel at all: either we have thought about it and want a conversion, then we should make it explicit, or we should avoid it, because it either dangerous (not verified) or increases verification cost (needs reasoning why the conversion is safe).

The downside is that there are currently quite a few warnings if we switch this on. On first sampling they look fine, i.e. are safe, but there are a few specifically on MCS that might not be.

So it'd be a bit of work and code churn to remove all warnings (including verification update), but I think it might be worth it.

What do people think?

kent-mcleod:
I think it's a good idea

Indanz:
Depends on how many of those "quite a few" warnings are duplicates, and if it are dozens or hundreds of unique warnings. If it's too many it's just a lot of work with not much gain: If people become too used fixing them without thinking, it may not prevent new bugs.

lsf37:
There are currently 85 such warnings on the master branch, many of which are versions of `unsigned conversion from 'int' to 'syscall_t'` at different points in `syscall.c` (all safe).

Some of the other ones are not obvious to me if they are correct or not. For instance is [this one](https://github.com/seL4/seL4/blob/76eee243238daab2e57b9a7527be2e9be82dd071/src/util.c#L127) always guaranteed to work? It converts everything to `unsigned` for both the `+` and then `*`, so either we lose the sign or we lose half of the value space. It's just a print utility function, but it would be very hard to spot.

Xaphiosis:
> Depends on how many of those "quite a few" warnings are duplicates, and if it are dozens or hundreds of unique warnings. If it's too many it's just a lot of work with not much gain: If people become too used fixing them without thinking, it may not prevent new bugs.

[...]

PR #1511, opened 2025-09-10
Runtime Domain Schedules

Indanz:
Implements [RFC-20](https://github.com/seL4/rfcs/pull/33).

TODO:
- [x] Document new syscalls.
- [x] Update existing Domains chapter.
- [x] Add tests to sel4test.
- [x] Check if generated code is reasonable.
- [x] Ease verification.

Test with https://github.com/seL4/sel4test/pull/151

### DIFF
--- a/include/arch/x86/arch/machine/cpu_registers.h
+++ b/include/arch/x86/arch/machine/cpu_registers.h
@@ -6,11 +6,13 @@
 
 #pragma once
 
+#define CR0_PE              BIT(0)  /* Protected Mode Enable */
 #define CR0_MONITOR_COPROC  BIT(1)  /* Trap on FPU "WAIT" commands. */
 #define CR0_EMULATION       BIT(2)  /* Enable OS emulation of FPU. */
 #define CR0_TASK_SWITCH     BIT(3)  /* Trap on any FPU usage, for lazy FPU. */
 #define CR0_NUMERIC_ERROR   BIT(5)  /* Internally handle FPU problems. */
 #define CR0_WRITE_PROTECT   BIT(16) /* Write protection in supervisor mode. */
+#define CR0_PG              BIT(31) /* Paging On */
 #define CR4_PCE             BIT(8)  /* Performance-Monitoring Counter enable. */
 #define CR4_OSFXSR          BIT(9)  /* Enable SSE et. al. features. */
 #define CR4_OSXMMEXCPT      BIT(10) /* Enable SSE exceptions. */
--- a/src/arch/x86/machine/cpu_identification.c
+++ b/src/arch/x86/machine/cpu_identification.c
@@ -109,7 +109,7 @@ BOOT_CODE static void x86_cpuid_amd_identity_initialize(cpu_identity_t *ci,
         ci->display.model = original.model;
     } else {
         ci->display.family = original.family + ci->display.extended_family;
-        ci->display.family = (ci->display.extended_model << 4u) + original.model;
+        ci->display.model = (ci->display.extended_model << 4u) + original.model;
     }
 }
 
--- a/src/arch/x86/object/vcpu.c
+++ b/src/arch/x86/object/vcpu.c
@@ -261,10 +261,6 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
         BIT(20) |   //Save guest IA32_EFER on exit
         BIT(21);    //Load host IA32_EFER
 #ifdef CONFIG_ARCH_X86_64
-#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
-    uint32_t entry_control_mask = 0;
-    entry_control_mask |= BIT(9); //Guest address-space size
-#endif
     exit_control_mask |= BIT(9); //Host address-space size
 #endif /* CONFIG_ARCH_X86_64 */
     /* Read out the fixed high and low bits from the MSRs */
@@ -298,6 +294,21 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
     cr0_low = x86_rdmsr_low(IA32_VMX_CR0_FIXED1_MSR);
     cr4_high = x86_rdmsr_low(IA32_VMX_CR4_FIXED0_MSR);
     cr4_low = x86_rdmsr_low(IA32_VMX_CR4_FIXED1_MSR);
+
+    /* "Unrestricted guest" is a userspace controlled setting that determines whether the VCPU
+     * may run in unpaged protected mode or in real-address mode. If the host CPU supports this feature, then
+     * CR0.PE and CR0.PG may be 0 while the VCPU is running: */
+
+    /* Intel SDM Combined Volumes
+     * Order Number: 325462-090US February 2026
+     * Chapter 26.8 "RESTRICTIONS ON VMX OPERATION"
+     * > Later processors support a VM-execution control called “unrestricted guest” (see Section 27.6.2).
+     * > If this control is 1, CR0.PE and CR0.PG may be 0 in VMX non-root operation (even if the capability
+     * > MSR IA32_VMX_CR0_FIXED0 reports otherwise). Such processors allow guest software to run in
+     * > unpaged protected mode or in real-address mode. */
+    if (secondary_control_low & BIT(7)) {
+        cr0_high &= ~(CR0_PE | CR0_PG);
+    }
 
     /* Check for VPID support */
     if (!(secondary_control_low & BIT(5))) {
@@ -349,9 +360,9 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
         return false;
     }
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
-    missing = (~entry_control_low) & entry_control_mask;
+    missing = (~entry_control_low) & BIT(9);
     if (missing) {
-        printf("vt-x: Unsupported entry control features %lx\n", (long)missing);
+        printf("vt-x: CPU does not support 64-bit guests\n");
         return false;
     }
 #endif /* CONFIG_X86_64_VTX_64BIT_GUESTS */
@@ -361,9 +372,6 @@ static bool_t BOOT_CODE init_vtx_fixed_values(bool_t useTrueMsrs)
     primary_control_high |= primary_control_mask;
     secondary_control_high |= secondary_control_mask;
     exit_control_high |= exit_control_mask;
-#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
-    entry_control_high |= entry_control_mask;
-#endif /* CONFIG_X86_64_VTX_64BIT_GUESTS */
 
     return true;
 }
@@ -400,6 +408,11 @@ static bool_t BOOT_CODE check_vtx_fixed_values(bool_t useTrueMsrs)
     uint32_t local_cr0_low = x86_rdmsr_low(IA32_VMX_CR0_FIXED1_MSR);
     uint32_t local_cr4_high = x86_rdmsr_low(IA32_VMX_CR4_FIXED0_MSR);
     uint32_t local_cr4_low = x86_rdmsr_low(IA32_VMX_CR4_FIXED1_MSR);
+
+    /* See "init_vtx_fixed_values" */
+    if (local_secondary_control_low & BIT(7)) {
+        local_cr0_high &= ~(CR0_PE | CR0_PG);
+    }
 
     /* We want to check that any bits that there are no bits that this core
      * requires to be high, that the BSP did not require to be high. This can
@@ -495,7 +508,15 @@ void vcpu_init(vcpu_t *vcpu)
     vmwrite(VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS, primary_control_high & primary_control_low);
     vmwrite(VMX_CONTROL_SECONDARY_PROCESSOR_CONTROLS, secondary_control_high & secondary_control_low);
     vmwrite(VMX_CONTROL_EXIT_CONTROLS, exit_control_high & exit_control_low);
-    vmwrite(VMX_CONTROL_ENTRY_CONTROLS, entry_control_high & entry_control_low);
+
+    uint64_t entry_control = entry_control_high & entry_control_low;
+#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
+    /* Put VCPU in IA-32e mode after VM entry for backward compatibility with previous versions of seL4.
+     * This bit can be zero'ed by userspace to enter in other modes. */
+    entry_control |= BIT(9);
+#endif /* CONFIG_X86_64_VTX_64BIT_GUESTS */
+    vmwrite(VMX_CONTROL_ENTRY_CONTROLS, entry_control);
+
     vmwrite(VMX_CONTROL_MSR_ADDRESS, (word_t)kpptr_to_paddr(&msr_bitmap_region));
     vmwrite(VMX_GUEST_CR0, vcpu->cr0);
     vmwrite(VMX_GUEST_CR4, cr4_high & cr4_low);
@@ -882,6 +903,9 @@ static exception_t decodeWriteVMCS(cap_t cap, word_t length, bool_t call, word_t
     case VMX_CONTROL_SECONDARY_PROCESSOR_CONTROLS:
         value = applyFixedBits(value, secondary_control_high, secondary_control_low);
         break;
+    case VMX_CONTROL_ENTRY_CONTROLS:
+        value = applyFixedBits(value, entry_control_high, entry_control_low);
+        break;
     case VMX_CONTROL_EXIT_CONTROLS:
         value = applyFixedBits(value, exit_control_high, exit_control_low);
         break;
@@ -1017,7 +1041,9 @@ static exception_t decodeReadVMCS(cap_t cap, word_t length, bool_t call, word_t 
     case VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
     case VMX_CONTROL_PIN_EXECUTION_CONTROLS:
     case VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS:
+    case VMX_CONTROL_SECONDARY_PROCESSOR_CONTROLS:
     case VMX_CONTROL_EXCEPTION_BITMAP:
+    case VMX_CONTROL_ENTRY_CONTROLS:
     case VMX_CONTROL_EXIT_CONTROLS:
     case VMX_GUEST_CR0:
     case VMX_GUEST_CR3:


lsf37:
Good to see the old domain tests failing -- that means we're actually doing something :). This should disappear once we've updated sel4test accordingly.

Indanz:
Main changes:
- Changed to 64-bit only version.
- Moved all domain code from tcb.c to src/object/domain.c
- Removed `activate` argument.

CHANGES.md
```diff
@@ -21,12 +21,17 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
+## Upcoming release: BREAKING
 
 ### Changes
 
+* Runtime domain schedule configuration.
+  See [RFC-20](https://sel4.github.io/rfcs/implemented/0200-domain-schedules.html).
 
 ### Upgrade Notes
+
+If using domains, replace your custom domain.c schedule with initialisation code in user space.
```
lsf37: We'll also need to update the capDL initialiser (I can do that part) and probably the CAmkES projects that use the domain scheduler. When that is done, we should mention that here, but for now we can leave the change log as you have it, I think.

src/object/domain.c
```diff
@@ -0,0 +1,185 @@
...
+
+    if (duration == 0 && domain != 0) {
+        userError("Domain Schedule Configure: Both domain and duration must be zero for end markers.");
+        current_syscall_error.invalidArgumentNumber = 1;
+        current_syscall_error.type = seL4_InvalidArgument;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    if (index == ksDomScheduleStart && duration == 0) {
+        userError("Domain Schedule Configure: Starting schedule's duration may not be zero.");
```
lsf37:
```suggestion
        userError("Domain Schedule Configure: Starting schedule's duration must not be zero.");
```

src/object/domain.c
```diff
@@ -0,0 +1,185 @@
...
+        userError("Domain Schedule Set Start: Invalid index.");
+        current_syscall_error.invalidArgumentNumber = 0;
+        current_syscall_error.type = seL4_RangeError;
+        current_syscall_error.rangeErrorMin = 0;
+        current_syscall_error.rangeErrorMax = domScheduleLength - 1;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    if (ksDomSchedule[index] == DSCHED_END_MARKER) {
+        userError("Domain Schedule Set Start: Starting schedule may not be an end marker.");
```
lsf37:
```suggestion
        userError("Domain Schedule Set Start: Starting schedule must not be an end marker.");
```

Should we use the same text in Schedule Configure? (end marker instead of duration 0)

libsel4/include/interfaces/object-api.xml
```diff
@@ -1352,6 +1352,81 @@
...
+            <description>
+                Configure the domain and duration of a scheduling entry.
+                A zero domain and duration means it is an end marker.
+                The domain scheduler will go back to the starting index when reaching
+                an end marker.
+                This makes it possible to configure multiple independent domain schedules
+                and switch between them atomically.
+
+                The scheduling entry will be modified immediately. However, if the currently active
+                index is modified, the change won't apply till the next time this entry is reached
```
lsf37:
```suggestion
                index is modified, the change will not apply until the next time this entry is reached
```

libsel4/include/interfaces/object-api.xml
```diff
@@ -1352,6 +1352,81 @@
...
+                by the domain scheduler.
+
+                After boot the first scheduling entry has domain zero and maximum duration, all other
+                entries will be end markers.
+                <docref>See <autoref label="sec:domains"/>.</docref>
+            </description>
+            <return><errorenumdesc/></return>
+            <param dir="in" name="index" type="seL4_Word" description="The scheduling entry to modify."/>
+            <param dir="in" name="domain" type="seL4_Uint8" description="The scheduling entry's new domain."/>
+            <param dir="in" name="duration" type="seL4_Time" description="Timeslice in ticks (milliseconds for non-MCS, timer ticks for MCS)."/>
```
lsf37:
```suggestion
            <param dir="in" name="duration" type="seL4_Time" description="Timeslice in ticks (kernel ticks for non-MCS, timer ticks for MCS)."/>
```

(since you can reconfigure kernel ticks to be longer/shorter)

CHANGES.md
Indanz: Shouldn't that be mentioned in the capDL and CAmkES changelogs instead of the kernel one? I'd say those fall under the custom domain.c mentioned already.

Probably should say `KernelDomainSchedule` instead of domain.c though.

libsel4/include/interfaces/object-api.xml
```diff
@@ -1352,6 +1352,81 @@
...
+                    Or, <texttt text="domain"/> is not less than <texttt text="KernelNumDomains"/>.
+                </description>
+            </error>
+        </method>
+        <method id="DomainScheduleSetStart" name="ScheduleSetStart" manual_label="domain_schedule_set_start">
+            <brief>
+                Change the starting index of the domain scheduler.
+            </brief>
+            <description>
+                The domain scheduler starts with the scheduler entry indicated by the starting index.
```
lsf37:
```suggestion
                The domain scheduler starts with the schedule entry indicated by the starting index.
```

lsf37 (commented):
This is looking good to me and is implementing the RFC as approved. Will leave out the tick for now so we don't accidentally merge before verification is finished.

CHANGES.md
lsf37: How those accommodate the new scheduler should be mentioned there yes, but the fact that user-level infrastructure is likely to hide this change from the user should be mentioned, because ideally CAmkES and Microkit users don't need to do anything.

lsf37:
> * [ ]  Rename `DomainSetSet` to `DomainSet`?

Given that the change is only breaking if you are using the kernel API directly, we should at least have a deprecation period for the old `DomainSetSet` if we change it.

CHANGES.md
Indanz: Microkit is an SDK based approach and has no support for domains, as that would require the user to recompile the kernel and hence the whole SDK. Now we're changing that, they can indeed add support for it if they want (overhead of just enabling domains seems to be around 3 cycles for the fast path and 30 cycles when scheduling happens, as measured by the seL4_Send benchmark on tx1).

The whole point of Microkit and camkes is to hide most kernel details, it seems silly to mention that nothing for those users changes for every kernel change. Also there is a chicken and egg problem here: Usually the kernel changes happen first, and user space frameworks are updated later. So we would have to guess here whether a kernel change affects users of Microkit of camkes.

Also, in this case, the timing changed from time to ticks for MCS and it seems likely that this change will propagate to the end user. I'll add this to the changelog.

lsf37: There is a patch for current Microkit that adds domain support (together with a patch to seL4 that allows updating the schedule through the linker) -- a few people are using that support and the interface there won't change.

The ticks/us change might propagate up, though.

src/model/statedata.c
```diff
@@ -84,15 +84,14 @@ compile_assert(irqCNodeSize, sizeof(intStateIRQNode) >= ((INT_STATE_ARRAY_SIZE)
...
 ticks_t ksDomainTime;
-#else
-word_t ksDomainTime;
-#endif
 
 /* An index into ksDomSchedule for active domain and length. */
 word_t ksDomScheduleIdx;
 
+/* Starting index of the domain schedule */
+word_t ksDomScheduleStart;
```
Xaphiosis, line 94: is it worth explaining a little more here what this signifies (resumes from here when end marker reached, right?)? Currently the comment says the nearly same thing as the variable name

include/object/domain.h
```diff
@@ -0,0 +1,32 @@
...
+#include <config.h>
+#include <types.h>
+#include <api/failures.h>
+
+/* The domain is stored in the top 8 bits, duration in the bottom 56 bits.
+ * Not using 64 bits for duration also avoids overflows. */
+typedef uint64_t dschedule_t;
+
+#define DSCHED_END_MARKER 0
+#define DOMAIN_0_MAX_DURATION (UINT64_MAX >> 8)
```
Xaphiosis: why is this domain *0* max duration and not domain max duration?

include/object/domain.h
```diff
@@ -0,0 +1,32 @@
...
+
+static inline dom_t dschedule_domain(dschedule_t sched)
+{
+    return sched >> 56;
+}
+
+static inline ticks_t dschedule_duration(dschedule_t sched)
+{
+    return sched & DOMAIN_0_MAX_DURATION;
```
Xaphiosis: feels like calling this something like `DOMAIN_DURATION_MASK` or something makes more sense; it's coincidental that the duration and the mask happen to be the same, but really it is a mask

@lsf37 do we care that this is not using the bitfield generator? we've had some issues in the past with people manually doing word arithmetic and masking causing us hiccups, because the generator spits out useful proofs and the direct way doesn't

libsel4/include/interfaces/object-api.xml
```diff
@@ -1352,6 +1352,81 @@
...
+            <error name="seL4_IllegalOperation">
+                <description>
+                    The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                </description>
+            </error>
+            <error name="seL4_InvalidArgument">
+                <description>
+                    The <texttt text="duration"/> is zero, but <texttt text="domain"/> is not.
+                    Or, <texttt text="index"/> points to the starting index, but <texttt text="duration"/> is zero.
+                    Or, <texttt text="duration"/> does not fit in 56 bits.
```
Xaphiosis, line 1402: wouldn't the user expect this last one to also be a range error?

libsel4/include/interfaces/object-api.xml
```diff
@@ -1352,6 +1352,81 @@
...
+            <brief>
+                Change the starting index of the domain scheduler.
+            </brief>
+            <description>
+                The domain scheduler starts with the schedule entry indicated by the starting index.
+                When reaching an end marker or the end of the schedules, the scheduler will go back
+                to the starting entry again. The starting index is zero after boot.
+
+                The currently active scheduling index will be changed to the new starting index
+                immediately, before returning this system call. This relinguishes all remaining
```
Xaphiosis: returning *from*?

src/object/domain.c
```diff
@@ -0,0 +1,185 @@
...
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+    setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
+    invokeDomainSetSet(TCB_PTR(cap_thread_cap_get_capTCBPtr(tcap)), domain);
+    return EXCEPTION_NONE;
+}
+
+static void invokeDomainScheduleConfigure(word_t index, dom_t domain, word_t duration)
+{
+    dschedule_t dom_and_duration = ((uint64_t)domain << 56) | duration;
```
Xaphiosis: this should be a `mk_dschedule`-style function living next to where the other accessors are; having the 56 and word arithmetic here feels... odd

include/object/domain.h
lsf37: It's a value you can directly write for the domain/duration word that has domain 0 and max duration.

include/object/domain.h
lsf37: Usually I would also push for the bitfield generator, but this one looks easy enough to do, because there is only a clean mask and `&` going on, no sign extending or other shifting.

lsf37: I also had the same thought about `DOMAIN_DURATION_MASK` initially, but then it becomes weird when we later write that value in the boot code as the full pair. Defining the same value with two different names is also weird. So any way we go, we end up with at least one use site that looks a bit strange. Maybe worth adding a comment, though.

libsel4/include/interfaces/object-api.xml
lsf37: Well spotted, this is in fact a range error in the code.

src/object/domain.c
lsf37: I'd be Ok with it being in here, since the whole thing is a one-liner, but looking at it again, the 56 should have a name at least. It shouldn't be a magic number.

include/object/domain.h
Xaphiosis: Ah, my brain parses it as "max duration for domain 0". A better name would be `DOMAIN_0_WITH_MAX_DURATION`

src/object/domain.c
Xaphiosis: Up to you. I'd prefer having accessors/constructors in one place. Can live with a `#define` for the magic number though. It's a spec rule + vcg vs csymbr + word arithmetic.

include/object/domain.h
Indanz: It are two properties encoded in 64-bits, encoded with domain first and duration next (in big-endian notation). I misuse the define later just for max duration, and that makes it more convenient to have the domain in the top bits.

If it's clearer/easier I could add a `dschedule_set(domain, duration)` function and use that with a separate max duration define instead.

include/object/domain.h
Indanz: The bitfield generator doesn't support 64-bit words on 32-bit platforms AFAIK. But it would be nice if it did so fields can cross 32-bit word boundaries on 32-bit platforms.

Indanz: Or to be more precise, the bitfield generator does support it via `base`, but I'm not sure the proofs it generate are valid for 32-bit platforms.

libsel4/include/interfaces/object-api.xml
Indanz: It can't be, range error is 32 bits on 32-bit platforms, so it wouldn't fit.

It was a range error in the previous version of the code, but not any more.

src/object/domain.c
Indanz: I'll add the constructor function and use that here and in the init code. Worst case it will be a macro.

include/object/domain.h
Xaphiosis: That might be clearer indeed. This and the mask/value overloading is confusing. Not sure about `_set` as it's not setting anything and isn't a set of anything, maybe some kind of `make`?

include/object/domain.h
Xaphiosis: After discussion with Gerwin, yeah, I agree with you: fine to not use bitfield gen on this one. I hadn't realised the 64-on-32bit limitations.

libsel4/include/interfaces/object-api.xml
Xaphiosis: Ok, so do the docs and code disagree then?

src/object/domain.c
Xaphiosis: That's fine I think. For something this small a macro will work as well as a function.

libsel4/include/interfaces/object-api.xml
lsf37: I might have hallucinated that -- I definitely tried to do this in the spec. Need to check if there is a cast there that is a downcast instead of the upcast from 56 that I thought.

src/object/domain.c
Indanz: The reason I didn't do it originally is because this function is the constructor function already, so it would be just adding extra indirection for not much gain. We could just call invokeDomainScheduleConfigure instead of initialising the array at boot.

include/object/domain.h
```diff
@@ -0,0 +1,43 @@
...
+
+extern dschedule_t ksDomSchedule[CONFIG_NUM_DOMAIN_SCHEDULES];
+
+/* The domain is stored in the top 8 bits, duration in the bottom 56 bits. */
+#define DSCHED_DOMAIN_BITS 8 // Maximum of 255 domains supported
+#define DSCHED_DOMAIN_SHIFT (64 - DSCHED_DOMAIN_BITS)
+#define DSCHED_MAX_DURATION (UINT64_MAX >> DSCHED_DOMAIN_BITS)
+#define DSCHED_DURATION_MASK DSCHED_MAX_DURATION
+
+static dschedule_t dschedule_make(dom_t domain, uint64_t duration)
```
lsf37:
```suggestion
static inline dschedule_t dschedule_make(dom_t domain, uint64_t duration)
```

libsel4/include/interfaces/object-api.xml
lsf37: I have hallucinated it, the spec and code both return `InvalidArgument`, as they should.

src/object/domain.c
```diff
@@ -0,0 +1,183 @@
...
+        current_syscall_error.invalidArgumentNumber = 0;
+        current_syscall_error.type = seL4_RangeError;
+        current_syscall_error.rangeErrorMin = 0;
+        current_syscall_error.rangeErrorMax = domScheduleLength - 1;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+
+    if (domain >= numDomains) {
+        userError("Domain Schedule Configure: Invalid domain.");
+        current_syscall_error.invalidArgumentNumber = 1;
```
Indanz: `invalidArgumentNumber` only gets propagated to user space for `seL4_InvalidArgument`.

src/object/domain.c
```diff
@@ -0,0 +1,183 @@
...
+        current_syscall_error.type = seL4_InvalidArgument;
+        current_syscall_error.invalidArgumentNumber = 1;
+        return EXCEPTION_SYSCALL_ERROR;
+    }
+    setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
+    invokeDomainSetSet(TCB_PTR(cap_thread_cap_get_capTCBPtr(tcap)), domain);
+    return EXCEPTION_NONE;
+}
+
+static void invokeDomainScheduleConfigure(word_t index, dom_t domain, word_t duration)
```
lsf37: Confirmation that reviews and tests are not that good at finding bugs:

```suggestion
static void invokeDomainScheduleConfigure(word_t index, dom_t domain, ticks_t duration)
```

Shouldn't we get a warning from the compiler for down casts?

midnightveil: No: https://github.com/seL4/seL4/issues/1187

lsf37: And it was even myself who opened that issue..

Indanz: The compiler only warns when you downcast a known value that doesn't fit without -Wconversion.

src/object/domain.c
lsf37: True, we should remove that. It's harmless, but not necessary.

Indanz: I'm a bit unsure what index to return in `current_syscall_error.invalidArgumentNumber`, should it be what we decode, or what the user would expect according to the syscall wrapper?

lsf37: It should always be the index that led to the decoding error. Still need to double check this, but I thin in this case we should not be setting it at all if it is not reported to the user.

Indanz: I know, I meant the other ones.

In the meantime, for some unfathomable reason, all kernel builds broke on my machine...

lsf37: 👍 They are all correct, as far as I can see.

lsf37: To actually answer your question, now that I understand it :-), it should be the index in the ABI, not the syscall wrapper.

lsf37:
The end marker change looks good, still need to update the proofs for it.

src/object/domain.c
```diff
@@ -0,0 +1,172 @@
...
+    return EXCEPTION_NONE;
+}
+
+static void invokeDomainScheduleSetStart(word_t index)
+{
+    ksDomScheduleStart = index;
+    /* End the current domain schedule and force an
+     * immediate switch to the new starting index: */
+    ksDomainTime = 0;
+    ksDomScheduleIdx = domScheduleLength - 1;
```
lsf37: That's still not quite right -- when we see in scheduling that the domain time is 0, we will first increase the index, and then check the new index for the end marker. That means, we need to set `ksDomScheduleIdx` to `domScheduleLength - 2` so that we reach the end marker at `domScheduleLength - 1` in the comparison in `nextDomain`.

```suggestion
    ksDomScheduleIdx = domScheduleLength - 2;
```

lsf37: We probably should have a static assert somewhere that the schedule length is always >= 2 so that it always fits one entry plus the end marker.

Indanz: Yes, I forgot to update this bit for the end marker change.

Issue #1191, opened 2024-02-09
arm64: silent asserts before mapping the kernel window


andybui01:
This might not be a big problem, but it's just something I've encountered while debugging the kernel.

Between the elfloader jumping to the kernel and mapping the kernel window, printing is effectively disabled as the kernel UART drivers use a UART_PPTR virtual address. This means that when an assert fails the user may experience kernel "hanging" without any definite reason/clues as to why.

The workaround for this is to change the UART VA to the PA value instead, as the elfloader identity mapping will still be active at this point (mapping the kernel window).

Just wondering if a debug early-UART implementation that makes use of the elfloader 1:1 mapping makes sense.

Side note: similarly, there is a small gap between `activate_kernel_vspace()` and `setVtable()`. In this gap, the elfloader vtable will no longer be useable, and we may miss exceptions.

Ivan-Velickovic:
Is this a duplicate of https://github.com/seL4/seL4/issues/540?

andybui01:
> Is this a duplicate of https://github.com/seL4/seL4/issues/540?

Woops, seems like it, I'm happy to close this.

Is there any real preference for a potential fix here? Since this seems to be a known thing I can work on a patch for this?

Indanz:
If using the 1:1 mapping is possible, your solution is better though.

Indanz:
Elfloader doesn't do a full 1:1 mapping any more, so if you want early printfs to work, the loader must map the UART at the virtual address the kernel expects, and then things just work for the kernel and no changes are needed.

That said, we should (re)move all printfs that are done too early though, as mentioned in #540.

PR #1513, opened 2025-09-19
boot: allow early printf() - before kernel vspace

midnightveil:
On ARM and RISC-V (not x86, which uses IO ports), the kernel's printf() call goes via the virtual address, which means that early prints before we setup virtual memory cause faults. Any issues at early stages will silently fail, e.g. various assert() that may go off.

We add a global variable that gets adjusted once the vspace exists so that the UART drivers can go through the physical memory address (since the kernel expects to be mapped 1:1 at boot time) early-on and then migrate to virtual address later. This makes debugging early-boot issues significantly easier on bare metal (without gdb).

Fixes: #540
Fixes: #1191

### DIFF
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -340,6 +340,8 @@ static BOOT_CODE bool_t try_init_kernel(
     word_t  dtb_size
 )
 {
+    printf("Bootstrapping kernel\n");
+
     cap_t root_cnode_cap;
     cap_t it_ap_cap;
     cap_t it_pd_cap;
@@ -369,14 +371,19 @@ static BOOT_CODE bool_t try_init_kernel(
     /* setup virtual memory for the kernel */
     map_kernel_window();
 
+#ifdef CONFIG_PRINTING
+    /* Switch printf() from using the physical UART address to the virtual
+     * UART address (in the kernel device region). printf() can no longer be
+     * called until activate_kernel_vspace() at the start of init_cpu() has
+     * finished. */
+    uart_pptr = UART_PPTR;
+#endif
+
     /* initialise the CPU */
     if (!init_cpu()) {
         printf("ERROR: CPU init failed\n");
         return false;
     }
-
-    /* debug output via serial port is only available from here */
-    printf("Bootstrapping kernel\n");
 
     /* initialise the platform */
     init_plat();
--- a/tools/hardware/outputs/c_header.py
+++ b/tools/hardware/outputs/c_header.py
@@ -12,7 +12,7 @@ from typing import Dict, List
 import hardware
 from hardware.config import Config
 from hardware.fdt import FdtParser
-from hardware.utils.rule import HardwareYaml
+from hardware.utils.rule import HardwareYaml, KernelRegionGroup
 
 
 HEADER_TEMPLATE = '''/*
@@ -69,6 +69,13 @@ static inline CONST word_t physBase(void)
 {% for (addr, macro) in sorted(kernel_macros.items()) %}
 #define {{ macro }} (KDEV_BASE + {{ "0x{:x}".format(addr) }})
 {% endfor %}
+
+{% if uart_region is not none %}
+/* UART DEVICE (early-printf) */
+{{ uart_region.get_macro() }}
+#define UART_PADDR ({{ "0x{:x}".format(uart_region.regions[0].base) }})
+{{ uart_region.get_endif() }}
+{% endif %}
 
 {% if len(kernel_regions) > 0 %}
 static const kernel_frame_t BOOT_RODATA kernel_device_frames[] = {
@@ -186,8 +193,8 @@ def get_interrupts(tree: FdtParser, hw_yaml: HardwareYaml) -> List:
 
 
 def create_c_header_file(config, kernel_irqs: List, kernel_macros: Dict,
-                         kernel_regions: List, physBase: int, physical_memory,
-                         outputStream):
+                         kernel_regions: List, physBase: int, uart_region: KernelRegionGroup,
+                         physical_memory, outputStream):
 
     jinja_env = jinja2.Environment(loader=jinja2.BaseLoader, trim_blocks=True,
                                    lstrip_blocks=True)
@@ -201,6 +208,7 @@ def create_c_header_file(config, kernel_irqs: List, kernel_macros: Dict,
             'kernel_macros': kernel_macros,
             'kernel_regions': kernel_regions,
             'physBase': physBase,
+            'uart_region': uart_region,
             'physical_memory': physical_memory})
     data = template.render(template_args)
 
@@ -215,12 +223,17 @@ def run(tree: FdtParser, hw_yaml: HardwareYaml, config: Config, kernel_config_di
     physical_memory, reserved, physBase = hardware.utils.memory.get_physical_memory(tree, config)
     kernel_regions, kernel_macros = get_kernel_devices(tree, hw_yaml, kernel_config_dict)
 
+    # FIXME: There is probably a better way to do this....
+    uart_region = next((r for r in kernel_regions if 'UART_PPTR' in r.labels), None)
+    assert uart_region is None or len(uart_region.regions) == 1
+
     create_c_header_file(
         config,
         get_interrupts(tree, hw_yaml),
         kernel_macros,
         kernel_regions,
         physBase,
+        uart_region,
         physical_memory,
         args.header_out)
 
--- a/include/machine/io.h
+++ b/include/machine/io.h
@@ -19,6 +19,11 @@ unsigned char kernel_getDebugChar(void);
 #include <arch/types.h>
 #include <stdarg.h>
 
+#ifdef CONFIG_ARCH_ARM
+/* See arm/machine/io.c for definition */
+extern pptr_t uart_pptr;
+#endif
+
 /* the actual output function */
 void kernel_putDebugChar(unsigned char c);
 
--- a/src/arch/arm/machine/io.c
+++ b/src/arch/arm/machine/io.c
@@ -9,6 +9,13 @@
 #include <drivers/uart.h>
 
 #ifdef CONFIG_PRINTING
+
+/**
+ * This is initially the PADDR for the early kernel boots, and we switch it
+ * out to use the PPTR once the kernel has mapped memory in.
+ */
+pptr_t uart_pptr = UART_PADDR;
+
 void kernel_putDebugChar(unsigned char c)
 {
     uart_console_putchar(c);
--- a/src/arch/riscv/kernel/boot.c
+++ b/src/arch/riscv/kernel/boot.c
@@ -199,6 +199,8 @@ static BOOT_CODE bool_t try_init_kernel(
     word_t  dtb_size
 )
 {
+    printf("Bootstrapping kernel\n");
+
     cap_t root_cnode_cap;
     cap_t it_pd_cap;
     cap_t it_ap_cap;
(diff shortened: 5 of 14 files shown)


src/arch/arm/kernel/boot.c
```diff
@@ -369,14 +371,20 @@ static BOOT_CODE bool_t try_init_kernel(
...
+    uart_pptr = UART_PPTR;
+#endif
+
     /* initialise the CPU */
     if (!init_cpu()) {
         printf("ERROR: CPU init failed\n");
         return false;
     }
 
     /* debug output via serial port is only available from here */
```
midnightveil: Need to remove this line; will do next time I push changes.

tools/hardware/outputs/c_header.py
```diff
@@ -215,12 +223,17 @@ def run(tree: FdtParser, hw_yaml: HardwareYaml, config: Config, kernel_config_di
     physical_memory, reserved, physBase = hardware.utils.memory.get_physical_memory(tree, config)
     kernel_regions, kernel_macros = get_kernel_devices(tree, hw_yaml, kernel_config_dict)
 
+    # FIXME: There is probably a better way to do this....
+    uart_region = next((r for r in kernel_regions if 'UART_PPTR' in r.labels), None)
```
midnightveil, line 227: Horrible.

src/arch/arm/kernel/boot.c
```diff
@@ -369,14 +371,20 @@ static BOOT_CODE bool_t try_init_kernel(
     /* setup virtual memory for the kernel */
     map_kernel_window();
 
+#ifdef CONFIG_PRINTING
+    /* Once the kernel vspace is active, switch uart to use virtual address.
+     * printf() can then no longer be used until activate_kernel_vspace() at the
+     * start of init_cpu(). */
```
Indanz: Comment is misleading: You're switching to the virtual address here, not once the kernel vspace is active.
```suggestion
    /* Switch printf() to use virtual UART address, it cannot be called until
     * activate_kernel_vspace() at the start of init_cpu() has finished. */
```

tools/hardware/outputs/c_header.py
Indanz: No idea what's happening here. But considering the physical address is already known and put into `kernel_device_frames[0].paddr`, it must be extracted already and known somewhere. Can't wherever that happens create the define?

Honestly, just finishing it out of `kernel_device_frames` would be better than making mysterious scripts even more mysterious.

tools/hardware/outputs/c_header.py
```diff
@@ -70,6 +70,13 @@
 #define {{ macro }} (KDEV_BASE + {{ "0x{:x}".format(addr) }})
 {% endfor %}
 
+{% if uart_region is not none %}
+/* UART DEVICE (early-printf) */
+{{ uart_region.get_macro() }}
+#define UART_PADDR ({{ "0x{:x}".format(uart_region.regions[0].base) }})
+{{ uart_region.get_endif() }}
+{% endif %}
+
```
Indanz, line 79: Isn't it easier to just do it for all macro's?

```suggestion
#define {{ macro }} (KDEV_BASE + {{ "0x{:x}".format(addr) }})
#define {{ macro }}_PADDR (KDEV_BASE + {{ "0x{:x}".format(paddr) }})
{% endfor %}

```
And update the macro script to add the physical address, wherever that happens.

Indanz (commented):
Approve of the general idea, but details to be worked out.

It looks SMP safe, as secondary cores wait till the first one releases them.

tools/hardware/outputs/c_header.py
midnightveil: The thing is most of the code is written as if one can have multiple physical regions and so this doesn't work for everything?

Indanz:
No idea why ZYNQ7000_debug_clang_32 [fails](https://github.com/seL4/seL4/actions/runs/17901806851/job/50896081850?pr=1513), but that needs to resolved first before merging this: 
```
    Enabling MMU and paging
    Warning: Could not infer GIC interrupt target ID, assuming 0.
    ERROR: DTB at [40000..42e47) exceeds PADDR_TOP (1ff00000)
    seL4 called fail at /github/workspace/kernel/src/arch/arm/kernel/boot.c:674 in function init_kernel, saying "ERROR: kernel init failed"
```

midnightveil:
Bingo...

```
ELF-loading image 'kernel' to 0
  
    paddr=[0..3ffff]
  
    vaddr=[e0000000..e003ffff]
  
    virt_entry=e0000000
```

Kernel is loaded to `0x_e000_0000`, but this also happens to be the UART address: https://github.com/seL4/seL4/blob/09e6c3f5e27607cd6f7854bdc5e87d829c48d500/tools/dts/zynq7000.dts#L260-L269

But for some reason the kernel identity-maps everything below the kernel first vaddr as identity, but everything above is direct mapped? Somehow I've become mistaken, I thought the kernel booted entirely identity-mapped (just... in virtual memory) until `activate_kernel_vspace()` happened. I guess we were just lucky that none of the platforms with UART above the kernel didn't happen to write into an address that was direct-mapped into non-memory, somehow. I guess this PR isn't entirely valid... only when the UART is below the kernel first vaddr. I don't understand why...

https://github.com/seL4/seL4_tools/blob/26f94df7e97a04f8ee49c1e94a0aa12036581f71/elfloader-tool/src/arch-arm/32/mmu.c#L13-L50

Indanz:
Does that mean zync7000's `seL4_UserTop 0xe0000000` is ~wrong~ unfortunate? On 32-bit it's a bit crowded in the address space because of the limited range.

Indanz:
> Kernel is loaded to `0x_e000_0000`, but this also happens to be the UART address:
> 
> https://github.com/seL4/seL4/blob/09e6c3f5e27607cd6f7854bdc5e87d829c48d500/tools/dts/zynq7000.dts#L260-L269

Good find.

> Somehow I've become mistaken, I thought the kernel booted entirely identity-mapped (just... in virtual memory) until `activate_kernel_vspace()` happened.

No, that's how Efloader boots.

The current kernel can't boot with identity mapping, that's why it needs Elfloader. There is no way other than making all code position independent to make that work for all kernel code in its current form (because the same code would be executed with different addresses, same for data accesses). So Elfloader needs to map the kernel in the same way as the kernel will map its memory in `activate_kernel_vspace()`. Basically doing more or less the same thing twice. This is one of many reasons why I want to make seL4 self-booting without Elfloader. Currently this is a big part of the kernel boot protocol which is entirely undocumented.

> I guess we were just lucky that none of the platforms with UART above the kernel didn't happen to write into an address that was direct-mapped into non-memory, somehow.

Elfloader has an if-check to handle the case where the UART is in memory overlapping with the kernel AFAIK, to prevent this from happening. The kernel didn't do printf until `activate_kernel_vspace()`, and at that point it got mapped to a sane virtual address.

> I guess this PR isn't entirely valid... only when the UART is below the kernel first vaddr. I don't understand why...

I think this PR makes the wrong assumption that there always is a 1:1 mapping for the physical UART address pre `activate_kernel_vspace()`. Might be enough to undo the whole idea, except if you find some work-around.

kent-mcleod:
What about having the uart mapped to a defined virtual address as part of the boot contract with the elfloader. The elfloader always has uart working when it switches to the kernel, and the elfloader sets up mappings for the kernel, so the elfloader could always map the uart registers to a compatible virtual address the kernel can use.

midnightveil:
Yes, I think that might be a good idea. It might require the elfloader to know about where the kernel expects UART to be mapped, but that's probably easy enough to do...

Indanz:
> What about having the uart mapped to a defined virtual address as part of the boot contract with the elfloader. The elfloader always has uart working when it switches to the kernel, and the elfloader sets up mappings for the kernel, so the elfloader could always map the uart registers to a compatible virtual address the kernel can use.

If doing this, the loader should map the UART to the final location the kernel will use after boot. And then there are no changes on the kernel side needed.

I think this is fine to do to make debugging easier, but I do not think the kernel should rely on the loader doing this correctly. Or the loader has a way to tell the kernel it mapped the UART or something. The boot protocol between the loader and the kernel is already obscure and badly documented enough, this would add another subtle requirement.

Edit: Perhaps add a kernel config option (default off, enabled when building with a compatible loader) and introduce `early_printf` and use that pre-init.

kent-mcleod:
With recent seL4_tools/elfloader changes to remove the uart mapping once the mmu is enabled, the gap between being able to write to the uart hardware has gotten bigger. 

> If doing this, the loader should map the UART to the final location the kernel will use after boot. And then there are no changes on the kernel side needed.

The loader already maps the kernel address space. Then the kernel remaps its own address space with the same translations essentially. Should we consider changing the kernel's init structure such that the initial vspace mappings are actually mapped by the elfloader and the kernel takes over with the mmu registers and kernel page tables already initialized? Seems like once we start getting the elfloader to premap the uart device into the same address the kernel expects, what's the limitations to having a vspace specification for the kernel that the loader can follow?

Indanz:
> With recent seL4_tools/elfloader changes to remove the uart mapping once the mmu is enabled, the gap between being able to write to the uart hardware has gotten bigger.

True, probably best to close this PR.

> > If doing this, the loader should map the UART to the final location the kernel will use after boot. And then there are no changes on the kernel side needed.
> 
> The loader already maps the kernel address space. Then the kernel remaps its own address space with the same translations essentially. Should we consider changing the kernel's init structure such that the initial vspace mappings are actually mapped by the elfloader and the kernel takes over with the mmu registers and kernel page tables already initialized? Seems like once we start getting the elfloader to premap the uart device into the same address the kernel expects, what's the limitations to having a vspace specification for the kernel that the loader can follow?

My plan was to make the kernel code position-independent and then there is no need for the loader to enable the MMU at all. And if the kernel is self-loading, there is no need for Elfloader either (chain loading would still be possible of course).

That said, many users have their own loader which does special stuff like more control over resource partitioning, so we probably need to have multiple boot protocols for the kernel:

- Self-booting with static, compile time info.
- Self-booting with DTB parsing.
- External loader, with Elfloader as reference implementation.

In the case of an extra loader I agree that it would be much simpler if it does the page table setup. Then the early seL4 boot code can use exactly the same ABI as the loaders do. Only the entry point would be different then, the binary could be otherwise the same.

kent-mcleod:
For early kernel printing, the boot loader could pass in a function pointer that the kernel could call similar to UEFI boot services or riscv SBI calls.

lsf37:
> For early kernel printing, the boot loader could pass in a function pointer that the kernel could call similar to UEFI boot services or riscv SBI calls.

Function pointers are not in the verification subset -- neither is printing, but I would suspect that the type would be visible unless we go to some effort to hide it.

kent-mcleod:
Kernel printing is only part of debug kernel anyway so...

lsf37:
> Kernel printing is only part of debug kernel anyway so...

Sure, but if you have to pass in the parameter from the boot loader, then that parameter is going to be in the function declaration.

kent-mcleod:
In other cases we've ifdef'd the different boot interface. RISCV SMP has additional parameters passed in via the elfloader. Kernel can also just treat it as a word_t, and then the print init code can cast it to the function pointer type.

lsf37:
Alright, those would all work if the rest stays hidden.

Indanz:
Elfloader doesn't do a full 1:1 mapping any more, so if you want early printfs to work, the loader must map the UART at the virtual address the kernel expects, and then things just work for the kernel and no changes are needed.

(That's also cleaner for the few platforms where the UART physical address clashes with a kernel's virtual address.)

PR #1515, opened 2025-10-07
Add support for Raspberry Pi 5B

Ivan-Velickovic:
This adds support for the Raspberry Pi 5B SBC which uses Cortex-A76 CPUs, a separate commit has been added for adding Cortex-A76 support.

This has been tested with sel4test in the following configurations:
`../init-build.sh -DPLATFORM=bcm2712`
`../init-build.sh -DPLATFORM=bcm2712 -DSMP=1`
`../init-build.sh -DPLATFORM=bcm2712 -DSMP=1 -DMCS=1`

which passed.

TODO: test hypervisor and release-mode combinations of sel4test.
TODO: put up docs PR.

### DIFF
--- a/libsel4/arch_include/arm/sel4/arch/constants_cortex_a76.h
+++ b/libsel4/arch_include/arm/sel4/arch/constants_cortex_a76.h
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <sel4/config.h>
+
+#if !defined(CONFIG_ARM_CORTEX_A76)
+#error CONFIG_ARM_CORTEX_A76 is not defined
+#endif
+
+/* Cortex-A76 Manual, Section 17.2.2 */
+#define seL4_NumHWBreakpoints           10
+#define seL4_NumExclusiveBreakpoints     6
+#define seL4_NumExclusiveWatchpoints     4
+
+#ifdef CONFIG_HARDWARE_DEBUG_API
+
+#define seL4_FirstBreakpoint             0
+#define seL4_FirstWatchpoint             6
+
+#define seL4_NumDualFunctionMonitors     0
+#define seL4_FirstDualFunctionMonitor    (-1)
+
+#endif /* CONFIG_HARDWARE_DEBUG_API */
--- a/src/plat/bcm2712/config.cmake
+++ b/src/plat/bcm2712/config.cmake
@@ -1,0 +1,31 @@
+#
+# Copyright 2025, UNSW
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
+declare_platform(bcm2712 KernelPlatformRpi5 PLAT_BCM2712 KernelArchARM)
+
+if(KernelPlatformRpi5)
+  declare_seL4_arch(aarch64)
+  set(KernelArmCortexA76 ON)
+  set(KernelArchArmV8a ON)
+  config_set(KernelARMPlatform ARM_PLAT rpi5)
+  set(KernelArmMachFeatureModifiers "+crc" CACHE INTERNAL "")
+  list(APPEND KernelDTSList "tools/dts/rpi5b.dts")
+  list(APPEND KernelDTSList "src/plat/bcm2712/overlay-rpi5.dts")
+  # The 2GB model is assumed, this can be changed with the KernelCustomDTSOverlay
+  # configuration option or adding support for other models like is done for
+  # bcm2711.
+  list(APPEND KernelDTSList "src/plat/bcm2712/overlay-rpi5-2gb.dts")
+
+  # - The clock frequency is 54 MHz as can be seen in bcm2712.dtsi in the
+  #   Linux Kernel under clk_osc, thus TIMER_FREQUENCY = 54000000.
+  # - MAX_IRQ is based on the GIC ITLinesNumber which reports 320.
+  declare_default_headers(
+    TIMER_FREQUENCY 54000000 MAX_IRQ 320 NUM_PPI 32 TIMER drivers/timer/arm_generic.h
+    INTERRUPT_CONTROLLER arch/machine/gic_v2.h KERNEL_WCET 10u)
+endif()
+
+add_sources(DEP "KernelPlatformRpi5" CFILES src/arch/arm/machine/gic_v2.c
+                                            src/arch/arm/machine/l2c_nop.c)
--- a/src/plat/bcm2712/overlay-rpi5-2gb.dts
+++ b/src/plat/bcm2712/overlay-rpi5-2gb.dts
@@ -1,0 +1,12 @@
+/*
+ * Copyright 2025, UNSW
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+    memory@0 {
+        device_type = "memory";
+        /* VideoCore memory lives at 0x3fc00000 onwards. */
+        reg = < 0x00000000 0x00000000 0x0 0x3fc00000>;
+    };
+};
--- a/configs/seL4Config.cmake
+++ b/configs/seL4Config.cmake
@@ -143,6 +143,7 @@ foreach(
         KernelArmCortexA55
         KernelArmCortexA57
         KernelArmCortexA72
+        KernelArmCortexA76
         KernelArchArmV7a
         KernelArchArmV7ve
         KernelArchArmV8a
@@ -194,6 +195,7 @@ config_set(KernelArmCortexA53 ARM_CORTEX_A53 "${KernelArmCortexA53}")
 config_set(KernelArmCortexA55 ARM_CORTEX_A55 "${KernelArmCortexA55}")
 config_set(KernelArmCortexA57 ARM_CORTEX_A57 "${KernelArmCortexA57}")
 config_set(KernelArmCortexA72 ARM_CORTEX_A72 "${KernelArmCortexA72}")
+config_set(KernelArmCortexA76 ARM_CORTEX_A76 "${KernelArmCortexA76}")
 config_set(KernelArchArmV7a ARCH_ARM_V7A "${KernelArchArmV7a}")
 config_set(KernelArchArmV7ve ARCH_ARM_V7VE "${KernelArchArmV7ve}")
 config_set(KernelArchArmV8a ARCH_ARM_V8A "${KernelArchArmV8a}")
@@ -232,6 +234,8 @@ elseif(KernelArmCortexA57)
   set(KernelArmCPU "cortex-a57" CACHE INTERNAL "")
 elseif(KernelArmCortexA72)
   set(KernelArmCPU "cortex-a72" CACHE INTERNAL "")
+elseif(KernelArmCortexA76)
+  set(KernelArmCPU "cortex-a76" CACHE INTERNAL "")
 endif()
 if(KernelArchARM)
   config_set(KernelArmMach ARM_MACH "${KernelArmMach}")
--- a/libsel4/sel4_plat_include/bcm2712/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/bcm2712/sel4/plat/api/constants.h
@@ -1,0 +1,9 @@
+/*
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#pragma once
+
+#include <sel4/config.h>
+#include <sel4/arch/constants_cortex_a76.h>
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -30,6 +30,9 @@ elseif(KernelArmCortexA72)
   # (https://developer.arm.com/documentation/100095/0001/memory-management-unit/about-the-mmu)
   set(KernelArmPASizeBits44 ON)
   math(EXPR KernelPaddrUserTop "(1 << 44)")
+elseif(KernelArmCortexA76)
+  set(KernelArmPASizeBits40 ON)
+  math(EXPR KernelPaddrUserTop "(1 << 40)")
 endif()
 config_set(KernelArmPASizeBits40 ARM_PA_SIZE_BITS_40 "${KernelArmPASizeBits40}")
 config_set(KernelArmPASizeBits44 ARM_PA_SIZE_BITS_44 "${KernelArmPASizeBits44}")
@@ -78,7 +81,7 @@ config_option(
   "Build as Hypervisor. Utilise ARM virtualisation extensions to build the kernel as a hypervisor"
   DEFAULT ${KernelSel4ArchArmHyp}
   DEPENDS
-    "KernelArmCortexA15 OR KernelArmCortexA35 OR KernelArmCortexA57 OR KernelArmCortexA53 OR KernelArmCortexA55 OR KernelArmCortexA72"
+    "KernelArmCortexA15 OR KernelArmCortexA35 OR KernelArmCortexA57 OR KernelArmCortexA53 OR KernelArmCortexA55 OR KernelArmCortexA72 OR KernelArmCortexA76"
 )
 
 config_option(KernelArmGicV3 ARM_GIC_V3_SUPPORT "Build support for GICv3" DEFAULT OFF)
@@ -234,7 +237,8 @@ if(KernelAArch32FPUEnableContextSwitch OR KernelSel4ArchAarch64)
 endif()
 
 if(KernelArmCortexA7 OR KernelArmCortexA8 OR KernelArmCortexA15 OR KernelArmCortexA35
-   OR KernelArmCortexA53 OR KernelArmCortexA55 OR KernelArmCortexA57 OR KernelArmCortexA72)
+   OR KernelArmCortexA53 OR KernelArmCortexA55 OR KernelArmCortexA57 OR KernelArmCortexA72
+   OR KernelArmCortexA76)
   # According to https://developer.arm.com/documentation/100095/0001/functional-description/about-the-cortex-a72-processor-functions/components-of-the-processor
   # the L1 instruction on the Cortex-A72 cache has a 64-byte cache line.
   # Thus, 6 bits are needed.
--- a/src/plat/bcm2712/overlay-rpi5.dts
+++ b/src/plat/bcm2712/overlay-rpi5.dts
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+	chosen {
+		seL4,elfloader-devices =
+		    "serial10",
+		    &{/psci},
+		    &{/timer};
+
+		seL4,kernel-devices =
+		    "serial10",
+		    &{/soc@107c000000/interrupt-controller@7fff9000},
+		    &{/timer};
+	};
+
+	soc@107c000000 {
+		interrupt-controller@7fff9000 {
+			/* vGIC maintenance IRQ necessary for hypervisor mode. */
+			interrupts = <0x01 0x09 0xf04>;
+		};
+	};
+
+    reserved-memory {
+        /delete-node/ linux,cma;
+    };
+};
--- a/tools/dts/rpi5b.dts
+++ b/tools/dts/rpi5b.dts
@@ -1,0 +1,564 @@
+/*
+ * Copyright Linux Kernel Team
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ *
+ * This file is derived from an intermediate build stage of the
+ * Linux kernel. The licenses of all input files to this process
+ * are compatible with GPL-2.0-only.
+ */
+/dts-v1/;
+
+/ {
+	compatible = "raspberrypi,5-model-b", "brcm,bcm2712";
+	#address-cells = <0x02>;
+	#size-cells = <0x02>;
+	interrupt-parent = <0x01>;
+	model = "Raspberry Pi 5";
+
+	clocks {
+
+		clk-osc {
+			compatible = "fixed-clock";
+			#clock-cells = <0x00>;
+			clock-output-names = "osc";
+			clock-frequency = <0x337f980>;
+			phandle = <0x18>;
+		};
+
+		clk-vpu {
+			compatible = "fixed-clock";
+			#clock-cells = <0x00>;
+			clock-frequency = <0x2cb41780>;
+			clock-output-names = "vpu-clock";
+			phandle = <0x0b>;
+		};
+
+		clk-uart {
+			compatible = "fixed-clock";
+			#clock-cells = <0x00>;
+			clock-frequency = <0x8ca000>;
+			clock-output-names = "uart-clock";
+			phandle = <0x0a>;
+		};
+
+		clk-emmc2 {
+			compatible = "fixed-clock";
+			#clock-cells = <0x00>;
+			clock-frequency = <0xbebc200>;
+			clock-output-names = "emmc2-clock";
+			phandle = <0x07>;
+		};
+	};
+
+	cpus {
+		#address-cells = <0x01>;
+		#size-cells = <0x00>;
+		phandle = <0x19>;
+
+		cpu@0 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a76";
+			reg = <0x00>;
+			enable-method = "psci";
+			d-cache-size = <0x10000>;
+			d-cache-line-size = <0x40>;
+			d-cache-sets = <0x100>;
+			i-cache-size = <0x10000>;
+			i-cache-line-size = <0x40>;
+			i-cache-sets = <0x100>;
+			next-level-cache = <0x02>;
+			phandle = <0x1a>;
+
+			l2-cache-l0 {
+				compatible = "cache";
+				cache-size = <0x80000>;
+				cache-line-size = <0x40>;
+				cache-sets = <0x400>;
+				cache-level = <0x02>;
+				cache-unified;
+				next-level-cache = <0x03>;
+				phandle = <0x02>;
+			};
+		};
+
+		cpu@1 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a76";
+			reg = <0x100>;
+			enable-method = "psci";
+			d-cache-size = <0x10000>;
+			d-cache-line-size = <0x40>;
+			d-cache-sets = <0x100>;
+			i-cache-size = <0x10000>;
+			i-cache-line-size = <0x40>;
+			i-cache-sets = <0x100>;
+			next-level-cache = <0x04>;
+			phandle = <0x1b>;
+
+			l2-cache-l1 {
+				compatible = "cache";
+				cache-size = <0x80000>;
+				cache-line-size = <0x40>;
+				cache-sets = <0x400>;
+				cache-level = <0x02>;
+				cache-unified;
+				next-level-cache = <0x03>;
+				phandle = <0x04>;
+			};
+		};
+
+		cpu@2 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a76";
+			reg = <0x200>;
+			enable-method = "psci";
+			d-cache-size = <0x10000>;
+			d-cache-line-size = <0x40>;
+			d-cache-sets = <0x100>;
+			i-cache-size = <0x10000>;
+			i-cache-line-size = <0x40>;
+			i-cache-sets = <0x100>;
+			next-level-cache = <0x05>;
+			phandle = <0x1c>;
+
+			l2-cache-l2 {
+				compatible = "cache";
+				cache-size = <0x80000>;
+				cache-line-size = <0x40>;
+				cache-sets = <0x400>;
+				cache-level = <0x02>;
+				cache-unified;
+				next-level-cache = <0x03>;
+				phandle = <0x05>;
+			};
+		};
+
+		cpu@3 {
+			device_type = "cpu";
+			compatible = "arm,cortex-a76";
+			reg = <0x300>;
+			enable-method = "psci";
+			d-cache-size = <0x10000>;
+			d-cache-line-size = <0x40>;
+			d-cache-sets = <0x100>;
+			i-cache-size = <0x10000>;
+			i-cache-line-size = <0x40>;
+			i-cache-sets = <0x100>;
+			next-level-cache = <0x06>;
+			phandle = <0x1d>;
+
+			l2-cache-l3 {
+				compatible = "cache";
+				cache-size = <0x80000>;
+				cache-line-size = <0x40>;
+				cache-sets = <0x400>;
+				cache-level = <0x02>;
+				cache-unified;
+				next-level-cache = <0x03>;
+				phandle = <0x06>;
+			};
+		};
+
+		l3-cache {
+			compatible = "cache";
+			cache-size = <0x200000>;
+			cache-line-size = <0x40>;
+			cache-sets = <0x800>;
+			cache-level = <0x03>;
+			cache-unified;
+			phandle = <0x03>;
+		};
+	};
+
+	psci {
+		method = "smc";
+		compatible = "arm,psci-1.0", "arm,psci-0.2";
+	};
+
+	reserved-memory {
+		ranges;
+		#address-cells = <0x02>;
+		#size-cells = <0x02>;
+		phandle = <0x1e>;
+
+		atf@0 {
+			reg = <0x00 0x00 0x00 0x80000>;
+			no-map;
+		};
+
+		linux,cma {
+			compatible = "shared-dma-pool";
+			size = <0x00 0x4000000>;
+			reusable;
+			linux,cma-default;
+			alloc-ranges = <0x00 0x00 0x00 0x40000000>;
+			phandle = <0x1f>;
+		};
+	};
+
+	soc@107c000000 {
+		compatible = "simple-bus";
+		ranges = <0x00 0x10 0x00 0x80000000>;
+		#address-cells = <0x01>;
+		#size-cells = <0x01>;
+		phandle = <0x20>;
+
+		mmc@fff000 {
+			compatible = "brcm,bcm2712-sdhci", "brcm,sdhci-brcmstb";
+			reg = <0xfff000 0x260 0xfff400 0x200>;
+			reg-names = "host", "cfg";
+			interrupts = <0x00 0x111 0x04>;
+			clocks = <0x07>;
+			clock-names = "sw_sdio";
+			mmc-ddr-3_3v;
+			vqmmc-supply = <0x08>;
+			vmmc-supply = <0x09>;
+			bus-width = <0x04>;
+			sd-uhs-sdr50;
+			sd-uhs-ddr50;
+			sd-uhs-sdr104;
+			phandle = <0x21>;
+		};
+
+		timer@7c003000 {
+			compatible = "brcm,bcm2835-system-timer";
+			reg = <0x7c003000 0x1000>;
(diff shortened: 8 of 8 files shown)


Ivan-Velickovic:
I just discovered that there are other IRQ controllers that do control IRQs for peripherals such as HDMI and the I2C controller. At the moment I do not plan to add the driver for these non-GIC IRQ controllers like there exists for the Raspberry Pi 4 (`bcm2836-armctrl-ic`).

Should I add these extra IRQ controller nodes to the kernel reserved regions in the overlay DTS or leave it as it is?

lsf37:
I was wondering why the RPi5 was suddenly so clean compared to the other ones.. If you leave them out now, we definitely need prominent pointers to what is missing and what to do if you need support for these.

Ivan-Velickovic:
> I was wondering why the RPi5 was suddenly so clean compared to the other ones.. If you leave them out now, we definitely need prominent pointers to what is missing and what to do if you need support for these.

I'll investigate tomorrow how complicated the extra IRQ controller drivers would be.

Indanz:
Aren't these just kind of IRQ multiplexers that end up sending IRQs to the GIC as usual? That is, the CPU has no direct support for them, you still need to manage those IRQs as any others.

If so, it's just a normal device for which user space needs to write drivers.

Ivan-Velickovic:
All the `interrupt-controller`s other than the GIC seem to have an SPI IRQ associated with the GIC from looking at the Device Tree.

Unfortunately I could not find any technical reference manual for BCM2712 so it's hard for me to confirm. Not sure how else I could check.

Indanz:
> All the `interrupt-controller`s other than the GIC seem to have an SPI IRQ associated with the GIC from looking at the Device Tree.

seL4 has GICv3 support, but no support for ITS and LPIs. It might be that that's needed to use some peripherals on this or future SoCs.

> Unfortunately I could not find any technical reference manual for BCM2712 so it's hard for me to confirm. Not sure how else I could check.

Maybe useful: https://www.reddit.com/r/raspberry_pi/comments/1ev61jg/datasheet_for_bcm2712/lisxc2b/

Ivan-Velickovic:
Okay with some more digging I believe that, like Indan said, these are just IRQ multiplexers and not directly attached the the CPU.

From [1]:

> This interrupt controller hardware is a second level interrupt controller that
is hooked to a parent interrupt controller: e.g: ARM GIC for ARM-based
platforms. It can be found on BCM7xxx products starting with BCM7120.

So I believe that the kernel does not need to have a driver for these IRQ controllers or reserve these regions.

> seL4 has GICv3 support, but no support for ITS and LPIs. It might be that that's needed to use some peripherals on this or future SoCs.

What I meant with my previous comment is that since they have IRQs with the GIC that may imply that they are multiplexing a single GIC IRQ into multiple peripheral IRQs. This SoC uses GICv2.

[1]: https://www.kernel.org/doc/Documentation/devicetree/bindings/interrupt-controller/brcm%2Cbcm7120-l2-intc.txt

src/plat/bcm2712/overlay-rpi5-2gb.dts
Ivan-Velickovic, line 1: I only have a 2GB model and with this VideoCore on the SoCs the main memory map is not so straight-forward.

Should I take an educated guess for the overlays of the other models based on the Pi 4B or leave them until someone else needs more than 2GB of RAM?

Indanz: Just leave it, people can add an overlay themselves if they have different amounts of memory, it's simple enough.

Linux specific reserved memory nodes like cma should be deletes.

Ivan-Velickovic: > Just leave it, people can add an overlay themselves if they have different amounts of memory, it's simple enough.

Sure thing.

> Linux specific reserved memory nodes like cma should be deletes.

Done.

Ivan-Velickovic:
I believe this is now ready for review, see PR description for some details.

libsel4/arch_include/arm/sel4/arch/constants_cortex_a76.h
```diff
@@ -0,0 +1,28 @@
...
+#define seL4_NumExclusiveBreakpoints     6
+#define seL4_NumExclusiveWatchpoints     4
+
+#ifdef CONFIG_HARDWARE_DEBUG_API
+
+#define seL4_FirstBreakpoint             0
+#define seL4_FirstWatchpoint             6
+
+#define seL4_NumDualFunctionMonitors     0
+#define seL4_FirstDualFunctionMonitor    (-1)
```
Ivan-Velickovic, line 26: The other constants are defined from the manual, but I do not understand how these are determined.

They appear to be the same for every ARM CPU.

Indanz:
> > seL4 has GICv3 support, but no support for ITS and LPIs. It might be that that's needed to use some peripherals on this or future SoCs.
> 
> What I meant with my previous comment is that since they have IRQs with the GIC that may imply that they are multiplexing a single GIC IRQ into multiple peripheral IRQs. This SoC uses GICv2.

Yes, but that's not efficient if you have multiple PCIE devices which are not part of the SoC. At some point we need to add ITS and LPI support.

src/plat/bcm2712/overlay-rpi5-2gb.dts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2025, UNSW
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+    /delete-node/ memory;
+
+    /* Memory for the 2GB RAM variant of the RPi5B, see overlay-pi5.dts for an
+     * explanation of the memory ranges.
```
Indanz: There is no such explanation there and it should be here anyway?

src/plat/bcm2712/config.cmake
```diff
@@ -0,0 +1,50 @@
...
+        # models.
+        set(RPI5_MEMORY "2048")
+    endif()
+
+    if("${RPI5_MEMORY}" STREQUAL "2048")
+        list(APPEND KernelDTSList "src/plat/bcm2712/overlay-rpi5-2gb.dts")
+    else()
+        message(FATAL_ERROR "Unsupported memory size given ${RPI5_MEMORY},
+                            supported memory sizes (in megabytes) are: 2048.")
+    endif()
```
Indanz: Ah, I understand your other question now. If you follow RPI4's way of handling multiple memory sizes and add the above weird stuff, you really should also add the overlays for the other memory sizes. It's known what sizes RPI5 comes with and the SoC is probably not used anywhere else. They come with 2, 4, 8 and 16 GB of RAM.

For a generic SoC it's okay to just add the maximum to the overlay and let users tweak it themselves. You can do that here too of course, but then don't have the above and just hardcode it in overlay-rpi5.dts with a comment to tell people to use `KernelCustomDTSOverlay` if they have a different amount.

src/plat/bcm2712/config.cmake
```diff
@@ -0,0 +1,50 @@
...
+
+    # - The clock frequency is 54 MHz as can be seen in bcm2712.dtsi in the
+    #   Linux Kernel under clk_osc, thus TIMER_FREQUENCY = 54000000.
```
Indanz: Please check if this is correct. That is, that Linux doesn't configure any prescaler in a different way than seL4 does and that 1 second actually takes 1 second on seL4 too.

src/plat/bcm2712/config.cmake
```diff
@@ -0,0 +1,50 @@
...
+    # - The GIC ITLinesNumber reports 320 however there are not enough bits to
+    #   encode the actual maximum IRQ so MAX_IRQ is based of the highest IRQ
+    #   from the Device Tree which is from '/soc@107c000000/mmc@fff000'.
```
Indanz: This is a strange comment, both 320 and 305 take the same amount of bits to encode, so why not use 320? And where is that bit limitation anyway? The tqma/imx8qxp platform has 511 IRQs and that didn't cause problems.

src/plat/bcm2712/overlay-rpi5-2gb.dts
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2025, UNSW
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+    /delete-node/ memory;
```
Indanz: I don't think it's necessary to delete a node you are redefining.

Ivan-Velickovic: Fixed.

src/plat/bcm2712/overlay-rpi5-2gb.dts
Ivan-Velickovic: Fixed.

src/plat/bcm2712/config.cmake
Ivan-Velickovic: By "not enough bits" I was referring to `ITLinesNumber`. Maybe I should get rid of that part of the comment and just say `MAX_IRQ is based of the highest IRQ from the Device Tree which is from '/soc@107c000000/mmc@fff000'`?

src/plat/bcm2712/config.cmake
Ivan-Velickovic: > For a generic SoC it's okay to just add the maximum to the overlay and let users tweak it themselves

I would prefer to have it default to the minimum to avoid potentially painful debugging as the kernel could use memory that does not exist on smaller memory models for kernel objects which leads to weird corruption issues.

The reason I left out the other overlays is because I believe the VideoCore memory that is present on RPi4 is not the exact same for RPi5 so I didn't want to add potentially wrong overlays since I cannot test anything other than 2GB right now.

I left this `RPI5_MEMORY` configuration option to make it easy to extend for people who definitely have 8GB/16GB etc working.

Do you want me to delete it and just hardcode 2GB?

src/plat/bcm2712/config.cmake
Indanz: Yes please. But why not just make it 320 if that is the maximum IRQ the SoC can produce?

Ivan-Velickovic: I thought it would be more accurate to make `MAX_IRQ` the maximum SPI that exists.

src/plat/bcm2712/config.cmake
Indanz: Minimum is fine too.

I'm not sure what VideoCore memory has to do with RAM. Normally there is a DDR controller on the SoC and its memory regions are defined in the datasheet, so known beforehand. VideoCore would hang off the GPU peripheral probably and also have a defined memory region. But RPIs are a bit weird, so who knows what's going on there. Anyway, it should be possible to find out what the memory layout is for bigger RPI5's.

My problem with `RPI5_MEMORY` is that it makes it harder, not easier to use different memory sizes as a user. So either add all of them now, or get rid of it.

src/plat/bcm2712/config.cmake
Indanz: My thinking is that the DTS may not be complete or that the high IRQ numbers are used for SoC specific IRQs instead of peripheral IRQs, so better to add them all now. It's only 15 more.

Ivan-Velickovic: Okay no worries, fixed.

src/plat/bcm2712/config.cmake
Ivan-Velickovic: Removed `RPI5_MEMORY` and left a comment in `config.cmake`.

libsel4/arch_include/arm/sel4/arch/constants_cortex_a76.h
Indanz: What do you mean? E.g. Cortex-A35 has `seL4_NumHWBreakpoints` set to 6.

Ivan-Velickovic: Ah, nevermind. I misunderstood what these were doing.

src/plat/bcm2712/overlay-rpi5-2gb.dts
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2025, UNSW
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+    memory@0 {
+        device_type = "memory";
+        /* VideoCore memory lives at 0x3fc00000 onwards. */
+        reg = < 0x00000000 0x00000000 0x0 0x3fc00000>;
```
Indanz, line 10: So the LPDDR is shared between CPU and GPU, and the GPU takes the top 1GB. Is this a software decision or decided by hardware?

Searching online, it seems it's configured in `config.txt`, which seems user configurable, but will configure that HW early on. Oh well, even more reason not to try to predefine all possible configurations.

Ivan-Velickovic:
@Indanz I will check the clock frequency tomorrow, I believe I've addressed all other feedback.

Indanz (approved):
All seems good.

With the SoC being bcm2712, you normally wouldn't expect big differences with the bcm2711. These seem much bigger changes than expected for a minor SoC revision increase.

src/plat/bcm2712/overlay-rpi5-2gb.dts
Ivan-Velickovic: > Searching online, it seems it's configured in config.txt, which seems user configurable, but will configure that HW early on. Oh well, even more reason not to try to predefine all possible configurations.

Yes, it can be configured by the user.

It is not clear to me whether the default is the same across all RPi5 models. I believe the default is most likely just decided on by the firmware at boot time.

Ivan-Velickovic: If someone reads this in the future and wants to add support for other models, the best thing to do I think is just run Linux and run `sudo cat /proc/meminfo` or `dtc -I fs -O dts /proc/device-tree`.

Indanz: It doesn't run U-Boot does it? If it does, `bdinfo` should help.

iiPythonx:
Out of curiosity, are there any instructions on how to get this running?

I've tried to run it on a 16gb Pi 5 by cloning both the rpi5 branch of the kernel and the bcm2712 branch of the libs/utils. Everything seems to build fine, the init launches, and ninja has no issues compiling it. However, copying it to my SD card with the dtb/config.txt/etc and then trying to load and launch it from within u-boot fails and restarts u-boot. 

Build commands:
```sh
../init-build.sh -DPLATFORM=bcm2712
ninja
```

The commands I used in U-Boot:
```
fatload mmc 0 0x10000000 sel4test-driver-image-arm-bcm2712
go 0x10000000
```

I'm unsure of if it's a U-Boot issue or a seL4 issue but I'm not sure.

midnightveil:
What exception information does uboot print?

I remember us getting it working using the microkit loader : https://github.com/seL4/microkit/tree/rpi5, but I also remember that at tve time there was no upstream Uboot (so we used the image from: https://github.com/au-ts/board_images_for_sel4)

iiPythonx:
> 

In particular, it gives nothing but a synchronous abort message:
```
U-Boot 2024.07 (Jan 01 1980 - 00:00:00 +0000)                                                                                                                                                                                              
                                                                                                                                                                                                                                           
DRAM:  1020 MiB (effective 16 GiB)                                                                                                                                                                                                         
RPI 5 Model B (0xe04171)                                                                                                                                                                                                                   
Core:  51 devices, 15 uclasses, devicetree: board                                                                                                                                                                                          
MMC:   mmc@fff000: 0, mmc@1100000: 1                                                                                                                                                                                                       
Loading Environment from FAT... Unable to read "uboot.env" from mmc0:1...                                                                                                                                                                  
In:    serial,usbkbd                                                                                                                                                                                                                       
Out:   serial,vidconsole                                                                                                                                                                                                                   
Err:   serial,vidconsole                                                                                                                                                                                                                   
PCIe BRCM: link up, 5.0 Gbps x4 (!SSC)                                                                                                                                                                                                     
Net:   eth0: ethernet@100000                                                                                                                                                                                                               
                                                                                                                                                                                                                                           
starting USB...                                                                                                                                                                                                                            
No USB controllers found                                                                                                                                                                                                                   
Hit any key to stop autoboot:  0                                                                                                                                                                                                           
U-Boot> fatload mmc 0 0x10000000 sel4test-driver-image-arm-bcm2712                                                                                                                                                                         
3512064 bytes read in 148 ms (22.6 MiB/s)                                                                                                                                                                                                  
U-Boot> go 0x10000000                                                                                                                                                                                                                      
## Starting application at 0x10000000 ...                                                                                                                                                                                                  
"Synchronous Abort" handler, esr 0x02000000                                                                                                                                                                                                
elr: ffffffffd04cd000 lr : 000000000000cc88 (reloc)                                                                                                                                                                                        
elr: 0000000010000000 lr : 000000003fb3fc88                                                                                                                                                                                                
x0 : 0000000000000001 x1 : 000000003f73e1c8                                                                                                                                                                                                
x2 : 000000003f73e1c8 x3 : 0000000010000000                                                                                                                                                                                                
x4 : 0000000000000030 x5 : 000000003f71abed                                                                                                                                                                                                
x6 : 000000000000000f x7 : 000000003f71aa28                                                                                                                                                                                                
x8 : 000000003f71b080 x9 : 00000000ffffffd0                                                                                                                                                                                                
x10: 0000000000000000 x11: 000000003f734e20                                                                                                                                                                                                
x12: 0000000000000000 x13: 0000000000000200                                                                                                                                                                                                
x14: 0000000000000000 x15: 000000003f71abed                                                                                                                                                                                                
x16: 0000000010000000 x17: 0000000000000000                                                                                                                                                                                                
x18: 000000003f72ed60 x19: 000000003f73e1c8                                                                                                                                                                                                
x20: 0000000010000000 x21: 0000000000000002                                                                                                                                                                                                
x22: 000000003f73e1c0 x23: 0000000000000002                                                                                                                                                                                                
x24: 000000003fbdb9fc x25: 0000000000000000                                                                                                                                                                                                
x26: 0000000000000000 x27: 000000003f73e220                                                                                                                                                                                                
x28: 000000003f73e1c0 x29: 000000003f71b080                                                                                                                                                                                                
                                                                                                                                                                                                                                           
Code: ffffffff ffffffff ffffffff ffffffff (464c457f)                                                                                                                                                                                       
Resetting CPU ...                                                                                                                                                                                                                          
                                                                                                                                                                                                                                           
resetting ...                                                                                                                                                                                                                              
                                                                                                                                                                                                                                           
  0.86 RPi: BOOTSYS release VERSION:ab8a9dde DATE: 2026/01/21 TIME: 13:38:47
...
```

I'm assuming I'm likely doing something wrong here, although I'm not too sure as to what. 
Is Microkit a requirement? I figured I would be able to boot a raw seL4 image just to see if it actually loads.

I also tried switching to the U-Boot images provided at https://github.com/au-ts/board_images_for_sel4 but still no dice unfortunately, same issue as with upstream.

Indanz:
> Is Microkit a requirement? I figured I would be able to boot a raw seL4 image just to see if it actually loads.

seL4 needs a loader to boot on Arm. Normally that's Elfloader, but Microkit has its own loader, same for rust-sel4.

I recommend either trying sel4test or a proper Microkit hello world example.

iiPythonx:
> I recommend either trying sel4test or a proper Microkit hello world example.

seL4test does give me an image file, but as explained above it does not boot for whatever reason.  
The other problem is trying to build Microkit on the `rpi5` branch produces some strange results:

```
$ python3 build_sdk.py --sel4=$(realpath ~/seL4/kernel) --skip-docs --boards rpi5b_2gb
    Finished `release` profile [optimized] target(s) in 0.02s
Building sel4: sel4_dir=PosixPath('/home/benjamin/seL4/kernel') root_dir=PosixPath('release/microkit-sdk-2.0.1-dev') build_dir=PosixPath('build/rpi5b_2gb/release/sel4') board=BoardInfo(name='rpi5b_2gb', arch=<KernelArch.AARCH64: 1>, gcc_cpu='cortex-a76', loader_link_address=268435456, kernel_options={'KernelPlatform': 'bcm2712', 'RPI5_MEMORY': 2048, 'KernelArmExportPCNTUser': True, 'KernelArmHypervisorSupport': True, 'KernelArmVtimerUpdateVOffset': False, 'KernelAllowSMCCalls': True, 'KernelIsMCS': True, 'KernelRootCNodeSizeBits': '17'}) config=ConfigInfo(name='release', debug=False, kernel_options={}, kernel_options_arch={})
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/benjamin/microkit/microkit/build/rpi5b_2gb/release/sel4/build
[31/31] Linking C executable kernel.elf
FAILED: [code=1] kernel.elf 
: && /usr/bin/aarch64-none-elf-gcc --sysroot=/home/benjamin/microkit/microkit/build/rpi5b_2gb/release/sel4/build -march=armv8-a+crc  -D__KERNEL_64__ -O3 -DNDEBUG -D__KERNEL_64__   -static -Wl,--build-id=none -Wl,-n -O2  -nostdlib  -fno-pic  -fno-pie  -mgeneral-regs-only   -Wl,-T /home/benjamin/microkit/microkit/build/rpi5b_2gb/release/sel4/build/linker.lds_pp CMakeFiles/kernel.elf.dir/src/arch/arm/armv/armv8-a/64/machine_asm.S.obj CMakeFiles/kernel.elf.dir/src/arch/arm/64/head.S.obj CMakeFiles/kernel.elf.dir/src/arch/arm/64/traps.S.obj CMakeFiles/kernel.elf.dir/src/arch/arm/64/idle.S.obj CMakeFiles/kernel.elf.dir/kernel_all.c.obj -o kernel.elf   && :
/usr/bin/aarch64-none-elf-ld: kernel.elf section `.bss' will not fit in region `ram'
/usr/bin/aarch64-none-elf-ld: ERROR: .data/.tdata flash size does not match RAM size
/usr/bin/aarch64-none-elf-ld: region `ram' overflowed by 86016 bytes
/usr/bin/aarch64-none-elf-ld: kernel.elf: section `.bss' can't be allocated in segment 3
LOAD: .boot .bss .stack .text
/usr/bin/aarch64-none-elf-ld: kernel.elf: section `.stack' can't be allocated in segment 3
LOAD: .boot .bss .stack .text
/usr/bin/aarch64-none-elf-ld: kernel.elf: section `._idle_thread' can't be allocated in segment 2
LOAD: .data ._idle_thread .boot.bss
/usr/bin/aarch64-none-elf-ld: kernel.elf: section `.boot.bss' can't be allocated in segment 2
LOAD: .data ._idle_thread .boot.bss
/usr/bin/aarch64-none-elf-ld: warning: kernel.elf has a LOAD segment with RWX permissions
/usr/bin/aarch64-none-elf-ld: final link failed: file truncated
collect2: error: ld returned 1 exit status
```

`~/seL4/projects/util_libs` is https://github.com/Ivan-Velickovic/util_libs/commit/137e14ec7d3f3de8401b755602dfeea810ac7c47
`~/seL4/kernel` is https://github.com/Ivan-Velickovic/seL4/commit/7dcd44a4b92ba230f8bf59531b497bca231c1d9e
`~/microkit/microkit` is https://github.com/seL4/microkit/commit/a0503ab38c2e3cae27b3feafc68d306fa03937b7

midnightveil:
re: the sel4test image

> ```
> U-Boot> fatload mmc 0 0x10000000 sel4test-driver-image-arm-bcm2712                                                                                                                                                                         
> 3512064 bytes read in 148 ms (22.6 MiB/s)                                                                                                                                                                                                  
> U-Boot> go 0x10000000                                                                                                                                                                                                                      
> ## Starting application at 0x10000000 ...                                                                                                                                                                                                  
> "Synchronous Abort" handler, esr 0x02000000  
> elr: ffffffffd04cd000 lr : 000000000000cc88 (reloc)                                                                                                                                                                                        
> elr: 0000000010000000 lr : 000000003fb3fc88                                                                                                                                                                                                                                                                          > ```

The fact that you have esr 0x02000000 tells us this is an invalid instruction fault. The elr is 0x10000000 so that means that it did so on the fist instruction.

I suspect the issue is that the sel4test-image-arm-bcm2712 is either uImage or ELF, not binary. If you run "file" on the image what does it report? You would then want to try either booti or bootelf.

iiPythonx:
It would indeed be an ELF file, although trying to load it with `bootelf` doesn't seem to do much:
```
U-Boot> fatload mmc 0 0x10000000 sel4test-driver-image-arm-bcm2712                                                                                                                                                                         
3512064 bytes read in 148 ms (22.6 MiB/s)                                                                                                                                                                                     
U-Boot> bootelf 0x10000000
U-Boot> 
```

Is there a way to produce a binary output from seL4test rather then ELF? I'm building my image as (mostly) described on https://docs.sel4.systems/projects/sel4test/, which is what's giving me the `sel4test-driver-image-arm-bcm2712` file.

midnightveil:
Yes: https://github.com/seL4/seL4_tools/blob/b82c586910bfd9da73c15cc4f9df5efd74bfd8d8/cmake-tool/helpers/application_settings.cmake#L17

U-Boot has an annoying default where bootelf only loads the image. There's an extra env var to set to actually boot that I can never remember the name of. (EDIT: https://docs.u-boot-project.org/en/latest/usage/environment.html see autostart here)

iiPythonx:
I unfortunately wasn't able to get bootelf working, even after setting the `autostart` environment variable.

I did, however, try adding `bcm2712` to the cmake file you linked which did in fact work for producing a binary build. Launching said build with `go` also worked fine, so seL4 is now up and running.

Thank you all for the help, I greatly appreciate it.

Indanz:
You can set the Elfloader boot image type via ccmake or during init with the `ElfloaderImage` setting.

The `go` command only works for binary images.

> U-Boot has an annoying default where bootelf only loads the image. There's an extra env var to set to actually boot 

Ah, that explains why I could never get it working. None of the info pages on the seL4 doc site mention this (I checked all of them).

Probably not the best idea to have ELF as default, especially if almost no board uses it.

iiPythonx:
Also, on a more unrelated note, is there anything blocking this from being merged into mainstream seL4? There's quite a few projects I'm working on that currently require patching to support the Pi 5 so having this upstream would be useful.

Indanz:
> Also, on a more unrelated note, is there anything blocking this from being merged into mainstream seL4?

I don't think so. I think I was waiting for all tests to complete before and later forgot about it. It's mostly that Ivan was busy with other things and didn't push for it. @midnightveil, could you test sel4test on the Pi 5B you have over there and see if it works? You'll need https://github.com/seL4/util_libs/pull/200.

midnightveil:
I think originally when we made this change there wss no upstream support for RPi5 in U-boot. This might have changed now. We had to hack it in with patches from mailing lists.

We probably want to add seL4 docs for this, specifically around different HW models (RAM size which I don't think we tested) and the GIC (due to the second-level interrupt controller).

I'll see if I can find the RPi5 somewhere (and try getting it in the machine queue). This is probably not going to the top of my priority list, though.

iiPythonx:
I will add that I have personally gotten sel4test running on a Pi 5 16gb model with no issues, and upstream u-boot does indeed work fine with the board.

midnightveil:
It looks like this port was done on Ivan's own Pi 5B, so we (Trustworthy Systems) don't have our own to test it on.

@iiPythonx Would you be able to add a PR to https://github.com/seL4/docs that documents how you set the board up?

Indanz:
All this PR need is a slight push by someone who has the actual hardware. If you could confirm that sel4bench also works, that would be great @iiPythonx.

iiPythonx:
Sure, I'll get to work on adding the documentation and setup a PR for that as soon as I can. I also didn't know sel4bench was a thing, I'll try to get that running as well.

iiPythonx:
I just went ahead and created https://github.com/seL4/docs/pull/340, it *should* be fine as far as I can tell with the documentation structure and whatnot.

As for sel4bench, it seems to run. This is the complete output (excuse minicom's random newlines in the output):
[minicom.log](https://github.com/user-attachments/files/29508232/minicom.log)

Here's also my completed output from sel4test:
[minicom.log](https://github.com/user-attachments/files/29508349/minicom.log)

midnightveil (approved):
I've rebased this and fixed the merge conflicts.

I think there's no outstanding major issues left.

midnightveil:
@iiPythonx Could you please recheck that the mainline sel4test-manifest and sel4bench-manifest repositories run for you now?

We are unsure how you got sel4bench working as our CI was failing to build until we fixed this: https://github.com/seL4/seL4_libs/pull/113.

iiPythonx:
Something worth noting is the PyPi package for `sel4-deps` does not currently include `protobuf` nor `grpcio-tools`, both of which seem to be required to build sel4test with ninja:
```
         **********************************************************************
         *** Could not import the Google protobuf Python libraries          ***
         ***                                                                ***
         *** Easiest solution is often to install the dependencies via pip: ***
         ***    pip install protobuf grpcio-tools                           ***
         **********************************************************************
    
  File "/home/benjamin/sel4test/build/nanopb/generator/protoc-gen-nanopb", line 7, in <module>
    from nanopb_generator import *
  File "/home/benjamin/sel4test/build/nanopb/generator/nanopb_generator.py", line 34, in <module>
    import google.protobuf.text_format as text_format
ModuleNotFoundError: No module named 'google'
--nanopb_out: protoc-gen-nanopb: Plugin failed with status code 1.
```

After you install them manually however (in my case `uv pip install sel4-deps protobuf grpcio-tools`), everything works.

> Could you please recheck that the mainline sel4test-manifest and sel4bench-manifest repositories run for you now

**seL4test:**

`../init-build.sh -DPLATFORM=bcm2712` (passed): 
[sel4test_blank.log](https://github.com/user-attachments/files/29847690/sel4test_blank.log)

`../init-build.sh -DPLATFORM=bcm2712 -DSMP=1` (passed):
[sel4test_smp.log](https://github.com/user-attachments/files/29847700/sel4test_smp.log)

`../init-build.sh -DPLATFORM=bcm2712 -DSMP=1 -DMCS=1` (passed):
[sel4test_smp_mcs.log](https://github.com/user-attachments/files/29847704/sel4test_smp_mcs.log)

This one gets a little interesting:
`../init-build.sh -DPLATFORM=bcm2712 -DSMP=1 -DMCS=1 -DRELEASE=TRUE`:
[sel4test_smp_mcs_release.log](https://github.com/user-attachments/files/29847899/sel4test_smp_mcs_release.log)

It's mostly the same as the prior ones, with the exception of the debug logging. However, as you can see from the log file, it seems the output text encoding or something gets mixed up midway through the test runs. It's obvious the end still says `All is well in the universe` based on the character lengths but I'm not sure what's causing this strange behavior or why it's only occuring on release mode.

**seL4bench:**

`../init-build.sh -DPLATFORM=bcm2712` (passed):
[sel4bench.log](https://github.com/user-attachments/files/29848058/sel4bench.log)

> We are unsure how you got sel4bench working as our CI was failing to build until we fixed this

I had to make the same file to get it to build, just forgot to mention that in the first comment. I assumed the events were the same as the A72, which seems to have been true thankfully.

midnightveil:
Cool, thanks!

lsf37:
> Something worth noting is the PyPi package for `sel4-deps` does not currently include `protobuf` nor `grpcio-tools`, both of which seem to be required to build sel4test with ninja:

Hm, I've been wondering whether we should include `protobuf`. The install instructions on https://docs.sel4.systems/projects/buildsystem/host-dependencies.html#base-dependencies do have `protobuf` and usually the package manager install is the preferred option. I'm not sure what happens when both are installed. I have not encountered needing to install `grpcio-tools` from python yet. Which system/distro are you using?


> **seL4test:**
> 
> `../init-build.sh -DPLATFORM=bcm2712` (passed): [sel4test_blank.log](https://github.com/user-attachments/files/29847690/sel4test_blank.log)
> 
> `../init-build.sh -DPLATFORM=bcm2712 -DSMP=1` (passed): [sel4test_smp.log](https://github.com/user-attachments/files/29847700/sel4test_smp.log)
> 
> `../init-build.sh -DPLATFORM=bcm2712 -DSMP=1 -DMCS=1` (passed): [sel4test_smp_mcs.log](https://github.com/user-attachments/files/29847704/sel4test_smp_mcs.log)
> 
> This one gets a little interesting: `../init-build.sh -DPLATFORM=bcm2712 -DSMP=1 -DMCS=1 -DRELEASE=TRUE`: [sel4test_smp_mcs_release.log](https://github.com/user-attachments/files/29847899/sel4test_smp_mcs_release.log)
> 
> It's mostly the same as the prior ones, with the exception of the debug logging. However, as you can see from the log file, it seems the output text encoding or something gets mixed up midway through the test runs. It's obvious the end still says `All is well in the universe` based on the character lengths but I'm not sure what's causing this strange behavior or why it's only occuring on release mode.

Thanks for checking all these. Is the output for the SMP_RELEASE build deterministically like this when you re-run it? E.g. switch happening at the same time in the output etc.

 
> **seL4bench:**
> 
> `../init-build.sh -DPLATFORM=bcm2712` (passed): [sel4bench.log](https://github.com/user-attachments/files/29848058/sel4bench.log)

Excellent, thanks!

iiPythonx:
> The install instructions on https://docs.sel4.systems/projects/buildsystem/host-dependencies.html#base-dependencies do have protobuf

Woops... I never actually saw that page so I assumed it was just a dependency missing from the Python metapackage.  
Also in that regard, I only installed `grpcio-tools` because the message suggested I do so, not because anything failed without it so I would imagine it's very much not required (my earlier statement saying both were needed was also based on the requirement message).

> Is the output for the SMP_RELEASE build deterministically like this when you re-run it? E.g. switch happening at the same time in the output etc.

That's correct, right now I've tested every combination of configuration options that I sent above (no options, SMP, MCS, SMP + MCS) and anytime `-DRELEASE=TRUE` is added to the mix sel4test will produce the same output on my Pi. It's always directly under test 142.

Indanz:
To find out whether it's bad input or a UART problem, could you add an `isascii()` check to `uart_write()` in projects/util_libs/libplatsupport/src/serial.c?

Test 142 is SERSERV_CLIENT_001 (the numbers don't mean much, please refer tests by name). That test only connects multiple clients to one console server concurrently via IPC, but it doesn't do anything special.

iiPythonx:
> To find out whether it's bad input or a UART problem, could you add an isascii() check to uart_write() in projects/util_libs/libplatsupport/src/serial.c?

Adding said `isascii()` check and making it not send any non-ASCII characters to the UART stream results in the same data being captured by minicom, that is the corrupted nonsense. By all indications it would appear to be a UART problem, but the question remains of why it would only happen on release mode (as no combination of other options affect it).

Also, taking logs of release mode on MCS + SMP, SMP only, and neither shows that it always occurs directly after the SERSERV_CLIENT_001 test, regardless of test number:
image

midnightveil:
I wouldn't be surprised if starting the serial server maybe allocates some untypeds in parts of physical memory one is not supposed to touch.

One thing I remember from earlier on is the CPU/GPU memory map is configurable per config.txt: https://github.com/seL4/seL4/pull/1515/files#r2444990188

Are you able to dump the dtb from UBoot (and Linux as well if you want) and share the decompiled dts? It would allow us to confirm if the memory map is the same. (I don't remember if you are also the 2GiB model).

The other thing to try would be to try disabling the SERVSRRV tests - is it specific to them or if removed will it eventually happen later?

iiPythonx:
> I don't remember if you are also the 2GiB model

I'm using the 16 GiB model.

> Are you able to dump the dtb from UBoot (and Linux as well if you want)

[linux_dts.txt](https://github.com/user-attachments/files/29978188/linux_dts.txt)
[uboot_dts.txt](https://github.com/user-attachments/files/29978190/uboot_dts.txt)

> try disabling the SERVSRRV tests

I completely removed the `serial_server.c` file from `projects/sel4test/apps/sel4test-tests/src/tests/` and did a fresh build, which resulted in it crashing in a different spot but the test name makes it seem like it's still a serial server test:
[output.log](https://github.com/user-attachments/files/29978333/output.log)

I'll try to figure out how the filtering tests by regex works, although I assumed just removing the file would be sufficient.

Issue #1175, opened 2024-01-29
Support full 48-bit PA on AArch64


heshamelmatary:
Currently the non-hyp AArch64 port only supports up to 47-bit Physical Address (PA) space due to how virtual memory mapping is implemented and a bitfield generator limitation with device untyped. It does not support 48-bit PAs accordingly. This issue was discussed in #1157 and #1163 when trying to add support for Morello platform that has 48-bit PA. The main (quoted) takeaways are:

- Device untyped represented in the seL4 capability still uses virtual canonical addresses (`capPtr `) even though they are not mapped. 
- The only way to relax that would be to find all places where that happens and to make special cases, comparing physical instead of virtual addresses in those case. This solution will require major verification efforts.
- One of the assumptions that the 64-bit kernel design currently makes is that the kernel window is large enough to hold all of addressable physical memory. With 48 bits of addressable virtual memory, 47-bit PA limit is probably the larges as there needs to be parts of the kernel virtual address space that isn't used for the direct mapping window. Getting larger than 47-bit would probably require a different approach for the kernel window mapping.
- In hyp mode (which support 48-bit PA), the kernel uses a separate page table for the kernel mappings. This could be a solution if armv8 would support that in non-hyp mode as well.

lsf37:
The 48 bit virtual addresses happen to match up with sizes on x64 and RISC-V 64 bit, which means that some of the generic bit field sizes (like those device untyped addresses) are [shared] (https://github.com/seL4/seL4/blob/64b1282accafdcc94f0087d25f2d51db1a9359ca/include/object/structures_64.bf) and 48 bit wide.

The sharing is mostly for convenience, it's probably possible to add another case there for 52, but we'd have to check if there is sufficient space in all caps for instance. Most look fine and have padding left, but `cnode_cap` looks like it might not have enough, at least without looking into further storage tricks.

With 3 cases for each pointer storage location it might be time to add an expression language to the bitfield generator, so that we can compute padding instead of having to add a preprocessor `IF` everywhere.

PR #1516, opened 2025-10-08
aarch64: use fixed virtual address for kernel address space

midnightveil:
This matches RISC-V which also does this (edit: actually, it doesn't, even though it maps the kernel ELF twice)

This is a precursor to some other boot protocol changes as a part of
the multikernel work where it is useful to be able to load the kernel
at different physical addresses using the same kernel image.

Previously, the kernel would run in the kernel physical memory window
at a platform-specific virtual address.

### DIFF
--- a/include/arch/arm/arch/64/mode/hardware.h
+++ b/include/arch/arm/arch/64/mode/hardware.h
@@ -77,13 +77,10 @@
  *            2^64 - 2^30 +-------------------+ PPTR_TOP
  *                        |                   |
  *                        |                   |
+ *                        |                   |
  *                        |  Physical Memory  |
  *                        |      Window       |
  *                        |                   |
- *                        |                   |
- *                        | ----------------- | KERNEL_ELF_TOP
- *                        |     Kernel ELF    |
- *                        | ----------------- | KERNEL_ELF_BASE (at `physBase()`)
  *                        |                   |
  *                        |                   |
  *            2^64 - 2^39 +-------------------+ PPTR_BASE
@@ -113,7 +110,11 @@
  *                      |     Unmapped     |
  *                      |                  |
  *                      |                  |
- *          2^64 - 2^30 +------------------+
+ *                      +------------------+ KERNEL_ELF_TOP (not fixed)
+ *                      |                  |
+ *                      |    Kernel ELF    |
+ *                      |                  |
+ *          2^64 - 2^30 +------------------+ KERNEL_ELF_BASE
  */
 
 /*
@@ -137,7 +138,7 @@
  *     -  The kernel expects to be able to create untypeds from 0 to
  *        KernelPaddrUserTop (CONFIG_PADDR_USER_TOP) which can be up to 2^44 in
  *        the current kernel. The kernel (proofs) expect addresses in these
- *        untypeds to be valid (i.e. canonical, not dereferencable) virtual
+ *        untypeds to be valid (i.e. canonical, not dereferenceable) virtual
  *        addresses, but allows device UT addresses to exceed ("overflow") the
  *        bounds of the physical memory window (PPTR_BASE...PPTR_TOP) as it will
  *        not dereference them itself. In the EL1 translation regime, as we use
@@ -203,10 +204,6 @@
  *                        |      Window       |
  *                        |                   |
  *                        |                   |
- *                        | ----------------- | KERNEL_ELF_TOP
- *                        |     Kernel ELF    |
- *                        | ----------------- | KERNEL_ELF_BASE (at `physBase()`)
- *                        |                   |
  *                        |                   |
  *            2^40 - 2^39 +-------------------+ PPTR_BASE
  *
@@ -228,7 +225,11 @@
  *                      |     Unmapped     |
  *                      |                  |
  *                      |                  |
- *          2^40 - 2^30 +------------------+
+ *                      +------------------+ KERNEL_ELF_TOP (not fixed)
+ *                      |                  |
+ *                      |    Kernel ELF    |
+ *                      |                  |
+ *          2^40 - 2^30 +------------------+ KERNEL_ELF_BASE
  *
  */
 
@@ -259,25 +260,36 @@
 /* For use by the linker (only integer constants allowed) */
 #define KERNEL_ELF_PADDR_BASE_RAW PHYS_BASE_RAW
 
-/* The base address in virtual memory to use for the kernel ELF mapping */
-#define KERNEL_ELF_BASE (PPTR_BASE_OFFSET + KERNEL_ELF_PADDR_BASE)
-/* For use by the linker (only integer constants allowed) */
-#define KERNEL_ELF_BASE_RAW (PPTR_BASE_OFFSET + KERNEL_ELF_PADDR_BASE_RAW)
-
-/* This is a page table mapping at the end of the virtual address space
- * to map objects with 4KiB pages rather than 2MiB large pages. */
+/* The KERNEL_ELF_BASE virtual address is placed at an arbitrary high memory
+ * address above PPTR_TOP. It is mapped using 2MiB pages.
+ * Arbitrarily, it is placed directly at PPTR_TOP.
+ */
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-#define KERNEL_PT_BASE UL_CONST(0x000000ffffe00000)  /* 2^40 - 2^21 */
+#define KERNEL_ELF_BASE UL_CONST(0x000000ffc0000000)  /* 2^40 - 2^30 */
 #else
-#define KERNEL_PT_BASE UL_CONST(0xffffffffffe00000)  /* 2^64 - 2^21 */
+#define KERNEL_ELF_BASE UL_CONST(0xffffffffc0000000)  /* 2^64 - 2^30 */
 #endif
 
-/* The base address in virtual memory to use for the kernel device
- * mapping region. These are mapped in the kernel page table. */
-#define KDEV_BASE KERNEL_PT_BASE
+/* For use by the linker (as it is shared with arm32) */
+#define KERNEL_ELF_BASE_RAW KERNEL_ELF_BASE
 
-/* The log buffer is placed before the device region */
-#define KS_LOG_PPTR (KDEV_BASE - UL_CONST(0x200000))
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* The base address in virtual memory to use for kernel devices, which are
+ * mapped using a PT containing 4K pages.  */
+#define KDEV_BASE UL_CONST(0x000000ffffe00000)  /* 2^40 - 2^21 */
+#else
+#define KDEV_BASE UL_CONST(0xffffffffffe00000)  /* 2^64 - 2^21 */
+#endif
+
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* The log buffer is placed before the device region using a 2MiB page. */
+#define KS_LOG_BASE UL_CONST(0x000000ffffc00000)  /* 2^40 - 2 * 2^21 */
+#else
+#define KS_LOG_BASE UL_CONST(0xffffffffffc00000)  /* 2^64 - 2 * 2^21 */
+#endif
+
+/* For consistency with the rest of the code. Changing the code might be a good idea. */
+#define KS_LOG_PPTR KS_LOG_BASE
 
 #ifndef __ASSEMBLER__
 /* All PPTR addresses must be canonical to be able to be stored in caps or objects.
--- a/include/machine.h
+++ b/include/machine.h
@@ -13,7 +13,11 @@
 /* When translating a physical address into an address accessible to the
  * kernel via virtual addressing we always use the mapping of the memory
  * into the physical memory window, even if the mapping originally
- * referred to a kernel virtual address. */
+ * referred to a kernel virtual address.
+ * As a result, this function should _not_ be used to modify memory if the
+ * addresses might lie within the kernel ELF mapping. See the comments attached
+ * to `map_kernel_elf_image()` on the AArch64 vspace code.
+ */
 static inline void *CONST ptrFromPAddr(paddr_t paddr)
 {
     return (void *)(paddr + PPTR_BASE_OFFSET);
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
                                                                                             attr_index);
 }
 
+/* Essentially, assert that the ELF, kernel devices, and log buffer are all in the same PUD.
+   In the current code this is the top index. This is arbitrary, the important part
+   is that they are all the same.
+ */
+compile_assert(elf_in_kernel_mappings, GET_KPT_INDEX(KERNEL_ELF_BASE,
+                                                     KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+compile_assert(kdev_in_kernel_mappings, GET_KPT_INDEX(KDEV_BASE,
+                                                      KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+compile_assert(log_buf_in_kernel_mappings, GET_KPT_INDEX(KS_LOG_BASE,
+                                                         KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+
+/**
+ *  Per ARM ARM DDI 0487 (version L.b), in B2.11 "Mismatched memory attributes",
+ *  when physical memory locations are accessed with mismatched attributes
+ *  (caching, memory type (device/normal), shareability), the coherency of these
+ *  memory locations may be lost, as well as various violations of read and
+ *  write orderings or atomic operations. Other ARM documents refer to the
+ *  situation in which two virtual addresses map to the same physical address as
+ *  "Memory aliasing" (e.g. Document ID 102376 0200_01_en).
+ *
+ *  Currently, the memory attributes used in the physical memory window does
+ *  match that of the kernel ELF mapping, so this is OK. At the same time,
+ *  the current physical window mapping does use NORMAL memory covering memory
+ *  that is not DRAM, which is questionable for newer processors with more
+ *  speculative execution. (FIXME)
+ *
+ *  Hence, the kernel should never access kernel ELF memory through the physical
+ *  memory window, i.e., never modify the result of `ptrFromPAddr()` when the
+ *  physical address lies within the kernel ELF. (Reads are fine if they
+ *  have been flushed appropriately per B2.11; this should be the case for
+ *  any paging structures).
+ *
+ *  As an addendum, LWN Article "ARM's multiply-mapped memory mess" specifies
+ *  that in ARMv6 specifications this kind of aliasing would be UNPREDICTABLE;
+ *  the current ARMv8 A-Profile specifications no longer contain this wording,
+ *  and instead specify the loss of coherency and other behaviours in B2.11.
+ */
+BOOT_CODE void map_kernel_elf_image(pte_t kernel_mapping_pds[])
+{
+    /** This is not strictly necessary, but to handle a non-aligned KERNEL_ELF_PADDR_BASE
+     *  we'd need to do a bit more logic and probably do what riscv64 does
+     *  where the *actual* KERNEL_ELF_BASE is rounded up to the alignment.
+     *  This is simpler and memory is cheap.
+     **/
+    compile_assert(elf_paddr_aligned, IS_ALIGNED(KERNEL_ELF_PADDR_BASE_RAW, seL4_LargePageBits));
+    compile_assert(elf_base_aligned, IS_ALIGNED(KERNEL_ELF_BASE, seL4_LargePageBits));
+
+    /* Please don't overlap with KDEV_BASE */
+    assert(KERNEL_ELF_TOP <= KDEV_BASE);
+    /* Please don't overlap with KS_LOG_PPTR */
+    assert(KERNEL_ELF_TOP <= KS_LOG_PPTR);
+
+    const word_t pd_start = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(2));
+    /* pd_end and ELF_TOP are both exclusive, so handle that correctly */
+    const word_t pd_end   = GET_KPT_INDEX(KERNEL_ELF_TOP - 1, KLVL_FRM_ARM_PT_LVL(2)) + 1;
+
+    paddr_t paddr = KERNEL_ELF_PADDR_BASE;
+    UNUSED pptr_t vaddr = KERNEL_ELF_BASE;
+
+    for (word_t pd = pd_start; pd < pd_end; pd++) {
+        assert(pd == GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(2)));
+
+        kernel_mapping_pds[pd] = pte_pte_page_new(
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+                                     /* XN */ 0,
+#else
+                                     /* UXN */ 1,
+#endif
+                                     /* page_base_address */ paddr,
+                                     /* global (nG) */ 0,
+                                     /* access flag (AF) */ 1,
+                                     /* shareability (SH) */ SMP_TERNARY(SMP_SHARE, 0),
+                                     /* AP */ APFromVMRights(VMKernelOnly),
+                                     /* AttrIndx */ NORMAL
+                                 );
+
+        paddr += BIT(seL4_LargePageBits);
+        vaddr += BIT(seL4_LargePageBits);
+    }
+}
+
 BOOT_CODE void map_kernel_window(void)
 {
 
@@ -253,8 +334,8 @@ BOOT_CODE void map_kernel_window(void)
     assert(IS_ALIGNED(PPTR_TOP, seL4_HugePageBits));
 
     /* place the PUD into the PGD */
-    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] = pte_pte_table_new(
-                                                                                 addrFromKPPtr(armKSGlobalKernelPUD));
+    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))]
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
 
     /* place all PDs except the last one in PUD */
     for (idx = GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(1)); idx < GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1));
@@ -283,6 +364,13 @@ BOOT_CODE void map_kernel_window(void)
                                                                                                                     );
         vaddr += BIT(seL4_LargePageBits);
     }
+
+
+    const word_t elf_pud_idx = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1));
+    pte_t *kernel_mapping_pds = armKSGlobalKernelPDs[elf_pud_idx];
+
+    /* map the kernel ELF */
+    map_kernel_elf_image(kernel_mapping_pds);
 
     /* put the PD into the PUD for device window */
     armKSGlobalKernelPUD[GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1))] = pte_pte_table_new(
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -369,18 +369,17 @@ BOOT_CODE void map_kernel_window(void)
     const word_t elf_pud_idx = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1));
     pte_t *kernel_mapping_pds = armKSGlobalKernelPDs[elf_pud_idx];
 
+    /* place the kernel mapping PDs in the kernel mapping PUD */
+    armKSGlobalKernelPUD[elf_pud_idx] = pte_pte_table_new(addrFromKPPtr(kernel_mapping_pds));
+
     /* map the kernel ELF */
     map_kernel_elf_image(kernel_mapping_pds);
 
-    /* put the PD into the PUD for device window */
-    armKSGlobalKernelPUD[GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1))] = pte_pte_table_new(
-                                                                                addrFromKPPtr(&armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][0])
-                                                                            );
-
-    /* put the PT into the PD for device window */
-    armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][BIT(PT_INDEX_BITS) - 1] = pte_pte_table_new(
-                                                                               addrFromKPPtr(armKSGlobalKernelPT)
-                                                                           );
+    /* put the PT into the PD for device window. It would be nice if
+       we could pass kernel_mapping_pds[511] to map_kernel_devices(), but it is
+       shared between AArch64 and AArch32, so... */
+    kernel_mapping_pds[BIT(PT_INDEX_BITS) - 1]
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPT));
 
     map_kernel_devices();
 }
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -337,16 +337,11 @@ compile_assert(physical_window_bottom_kernel_region,
 compile_assert(physical_window_top_kernel_region,
                GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
 
-BOOT_CODE void map_kernel_window(void)
+BOOT_CODE void map_kernel_physical_window(void)
 {
-
     paddr_t paddr;
     pptr_t vaddr;
     word_t idx;
-
-    /* place the PUD into the PGD */
-    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))]
-        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
 
     /* place all PDs except the last one in PUD */
     for (idx = GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(1)); idx < GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1));
@@ -375,7 +370,15 @@ BOOT_CODE void map_kernel_window(void)
                                                                                                                     );
         vaddr += BIT(seL4_LargePageBits);
     }
+}
 
+BOOT_CODE void map_kernel_window(void)
+{
+    /* place the PUD into the PGD */
+    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] =
+        pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
+
+    map_kernel_physical_window();
 
     const word_t elf_pud_idx = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1));
     pte_t *kernel_mapping_pds = armKSGlobalKernelPDs[elf_pud_idx];
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -314,24 +314,35 @@ BOOT_CODE void map_kernel_elf_image(pte_t kernel_mapping_pds[])
     }
 }
 
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* verify that the physical memory window as at the second entry of the PGD */
+compile_assert(physical_window_pgd_location,
+               GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
+#else
+/* verify that the physical memory window as at the last entry of the PGD */
+compile_assert(physical_window_pgd_location,
+               GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == BIT(PT_INDEX_BITS) - 1);
+#endif
+/* make sure that the PPTR_BASE and PPTR_TOP are huge paged aligned */
+compile_assert(physical_window_base_aligned, IS_ALIGNED(PPTR_BASE, seL4_HugePageBits));
+compile_assert(physical_window_top_aligned, IS_ALIGNED(PPTR_TOP, seL4_HugePageBits));
+
+/* the PPTR_BASE starts from the bottom of the kernel PUD */
+compile_assert(physical_window_bottom_kernel_region,
+               GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(1)) == 0);
+/* the PPTR_TOP then ends at the second-to-last entry of the PUD
+    note: because this is the exclusive value the assert is one higher than
+          you would expect.
+*/
+compile_assert(physical_window_top_kernel_region,
+               GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+
 BOOT_CODE void map_kernel_window(void)
 {
 
     paddr_t paddr;
     pptr_t vaddr;
     word_t idx;
-
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* verify that the kernel window as at the second entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
-#else
-    /* verify that the kernel window as at the last entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == BIT(PT_INDEX_BITS) - 1);
-#endif
-    assert(IS_ALIGNED(PPTR_BASE, seL4_LargePageBits));
-    /* verify that the kernel device window is 1gb aligned and 1gb in size */
-    assert(GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
-    assert(IS_ALIGNED(PPTR_TOP, seL4_HugePageBits));
 
     /* place the PUD into the PGD */
     armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))]
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -357,7 +357,7 @@ BOOT_CODE void map_kernel_physical_window(void)
         armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
                                                                                          KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-                                                                                                                        0, // XN
+                                                                                                                        1, // XN
 #else
                                                                                                                         1, // UXN
 #endif
--- a/tools/hardware/config.py
+++ b/tools/hardware/config.py
@@ -42,7 +42,15 @@ class ARMConfig(Config):
         elif sel4arch == 'aarch32':
             self.KERNEL_PHYS_ALIGN = 24
         else:
-            self.KERNEL_PHYS_ALIGN = 0
+            # On AArch64, `PADDR_BASE` is always zero, and only `KERNEL_ELF_PADDR_BASE`
+            # uses `PHYS_BASE`, which is influenced by this alignment.
+            # For simplicity, we require that `KERNEL_ELF_PADDR_BASE` is aligned
+            # to seL4_LargePageBits so that we can map it directly using 2MiB
+            # pages. This does however reserve memory from the true bottom of
+            # RAM up to this alignment (exposed as device UT), but this is no
+            # more than part of a single 2MiB page, which should be acceptable
+            # on AArch64.
+            self.KERNEL_PHYS_ALIGN = 21
 
     def get_kernel_phys_align(self) -> int:
         return self.KERNEL_PHYS_ALIGN
--- a/include/arch/arm/arch/kernel/vspace.h
+++ b/include/arch/arm/arch/kernel/vspace.h
@@ -20,6 +20,7 @@ cap_t create_mapped_it_frame_cap(cap_t pd_cap, pptr_t pptr, vptr_t vptr, asid_t 
                                  bool_t executable);
 
 void map_kernel_window(void);
+void map_kernel_elf_image(pte_t kernel_mapping_pds[]);
 void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_rights, vm_attributes_t vm_attributes);
 void activate_kernel_vspace(void);
 void write_it_asid_pool(cap_t it_ap_cap, cap_t it_pd_cap);
--- a/include/arch/arm/arch/kernel/vspace.h
+++ b/include/arch/arm/arch/kernel/vspace.h
@@ -21,6 +21,7 @@ cap_t create_mapped_it_frame_cap(cap_t pd_cap, pptr_t pptr, vptr_t vptr, asid_t 
 
 void map_kernel_window(void);
 void map_kernel_elf_image(pte_t kernel_mapping_pds[]);
+void map_kernel_physical_window(void);
 void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_rights, vm_attributes_t vm_attributes);
 void activate_kernel_vspace(void);
 void write_it_asid_pool(cap_t it_ap_cap, cap_t it_pd_cap);


include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
+ *    4-level page tables and 4K pages, for a total of 2^48 bits of translatable
+ *    virtual memory. This gives two addressable ranges of memory:
+ *
+ *    - the high portion, from 2^64 - 2^48 to 2^64 (0xffff000000000000 to 0xffffffffffffffff)
+ *    - the low portion,  from 0 to 2^48 (0x0 to 0xffffffffffff)
+ *
+ *    By convention, the kernel lives in the canonical high portion, and the
+ *    user lives in the canonical low portion. This distinction only really
+ *    matters in how the addresses are displayed to humans.
+      (FIXME: Does the hardware care at all?)
```
midnightveil: ~~Pointing out this FIXME.~~ Resolved, I somehow had missed that ARM had TTBR1 and TTBR0, I thought that was a FEAT_LVA feature.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
+ *           0x0 +-------------------+
+ *
+ *
+ *    The Kernel's PUD (`armKSGlobalKernelPUD`) is then split into several parts.
+ *
+ *    -  The physical memory window starts from the 0th PUD entry and extends
+ *       up to the second-to-last PUD entry. This covers 255GiB.
+ *
+ *       The physical memory window is mapped using Large Pages.
+         FIXME: Why not huge pages?
```
midnightveil: Ditto.

midnightveil:
I've added an extra commit documenting some concerns @kent-mcleod raised w/r/t the memory aliasing of the kernel ELF and the physical memory window when mapped with mismatched attributes. At this time the attributes are the same, so I believe there are no issues, but I've documented the architectural requirements for future readers.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,217 @@
...
- *                                       |
- *                         +-------------+
- *                         |
- *                         v
- *          2^64 +-------------------+
- *               |                   |
- *               |                   |     +------+
- *               | Kernel Page Table | --> |  PD  | ----+
- *               |                   |     +------+     |
+ *    In EL1, the 64th bit of the virtual address selects between TTBR1 and TTBR0,
```
midnightveil: Leaving this mostly as a TODO for myself and a note before merging: from some further discussion between Indan and myself, this is not strictly correct. Bit 55 selects this.

Probably should reference §D8.2.4 of the ARM ARM as well, which specifies this.

Looks like I wasn't the first to make this mistake, at least: https://lore.kernel.org/all/20240702091349.356008-1-kevin.brodsky@arm.com/t/#u

Unrelatedly it looks like the kernel relies on the bootloader to configure `TCR_EL1` and things like top-byte-ignore... which is a choice, I guess.

Indanz: The kernel really should configure `TCR_EL1` itself though, otherwise it is just assuming things.

midnightveil: https://github.com/seL4/seL4/issues/1517

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
 #include <arch/machine/hardware.h>
 #include <sel4/plat/api/constants.h>
 
-/* EL2 kernel address map:
```
Indanz: It's good to start with the non-HYP one, as that one is simpler, so approve of the re-ordering. However, because it's done in one commit, it becomes really hard to distinguish moves from changes, at least in the Github interface.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
- *                         v
- *          2^64 +-------------------+
- *               |                   |
- *               |                   |     +------+
- *               | Kernel Page Table | --> |  PD  | ----+
- *               |                   |     +------+     |
+ *    Per ARM ARM DDI 0487 (version L.b), in D8.2.4, VA bit[55] selects between
+ *    TTBR1_EL1 and TTBR0_EL1, the two Translation Table base registers. It is
+ *    conventionally (and seL4 follows this convention) that the "higher VA"
+ *    (TTBR1_EL1) is used for kernel memory and the "lower VA" (TTBR0_EL1) is
```
Indanz: Something like:
```suggestion
 *    TTBR1_EL1 and TTBR0_EL1, the two Translation Table base registers.
 *    seL4 follows the convention to use the "higher VA" (TTBR1_EL1) for kernel
 *    memory and the "lower VA" (TTBR0_EL1) is
```
(but with next sentence in the same style.)

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
+ *    used for userspace. seL4 does not enable the ARM "Address tagging" (D8.8)
+ *    feature (sometimes known as Top Byte Ignore) so virtual addresses need to
+ *    be sign extended in VA[63:56].
```
Indanz: seL4 should enable it though, no reason not to AFAIK. Better to follow Linux here.

include/arch/arm/arch/64/mode/hardware.h
Indanz: Good question. It was like this since the first public commit, so hard to find out.
I suspect they ran out of time to do it properly and fudged it all suboptimally to get it working?

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
+ *    The EL2 mode kernel uses TTBR0_EL2 which covers the range of
+ *    0x0 - 0x0000ffffffffffff (48 bits of vaddrspace).
+ *
+ *    The CPU on the TX1 only allows for 48-bits of addressable virtual memory.
+ *    Within the seL4 EL2 kernel's separate vaddrspace, it uses
+ *    the 512 GiB at the top of that 48 bits of addressable
+ *    virtual memory.
+ *
+ *    In EL2 there is no canonical-high portion of the address space since
+ *    address tagging is not supported in EL2. Therefore the kernel is linked
```
Indanz: I don't think address tagging has anything to do with it? There is just no `TTBR1_EL2`.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,222 @@
...
+ *    to the canonical lower portion of the address space (all the unused top bits
+ *    are set to 0, not 1). However, we get a separate set of page tables.
```
Indanz: Maybe better explain that in non-HYP, `TTBR0_EL1` and `TTBR1_EL1` configures both kernel and user space, while in HYP, the kernel has its own dedicated `TTBR0_EL2` register.

include/machine.h
```diff
@@ -13,7 +13,12 @@
 /* When translating a physical address into an address accessible to the
  * kernel via virtual addressing we always use the mapping of the memory
  * into the physical memory window, even if the mapping originally
- * referred to a kernel virtual address. */
+ * referred to a kernel virtual address.
+ *
+ * As a result, this function should _not_ be used to modify memory if the
+ * addresses might lie within the kernel ELF mapping.
+ * See the comments in `map_kernel_elf_image()` on AArch64.
```
Indanz: The mapping attributes should be compatible, in which case it should be fine.
But if you really want to check this, you should add an assertion to the function.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
```diff
@@ -252,6 +252,9 @@ SEL4_SIZE_SANITY(seL4_VSpaceEntryBits, seL4_VSpaceIndexBits, seL4_VSpaceBits);
 #endif
 
 #else
-/* First address in the virtual address space that is not accessible to user level */
+/* First address in the virtual address space that is not accessible to user level.
+ * This is 2^47 due to the untyped limitation of a maximum of 47 bits; see the
+ * compile_assert() 'ut_max_less_than_canonical' and 'ut_max_is_canonical'.
```
Indanz: This can't be true, UT have enough spare bits for 57 bits, so the limitation must be elsewhere.

What `ut_max_less_than_canonical` does is assuring that no overflows will happen in address range calculations by ensuring that maximum UT size is at least one bit less than the canonical range. It's just unfortunate that the canonical range is hard coded in that assertion.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+}
 
-    /* map the kernel window using large pages */
-    vaddr = PPTR_BASE;
-    for (paddr = PADDR_BASE; paddr < PADDR_TOP; paddr += BIT(seL4_LargePageBits)) {
-        armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
-                                                                                         KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
+BOOT_CODE void map_kernel_elf_image(const word_t kernel_mappings_pud_idx)
+{
+    compile_assert(elf_paddr_aligned, IS_ALIGNED(KERNEL_ELF_PADDR_BASE_RAW, seL4_LargePageBits));
```
Indanz: (This is the one failing for rpi4.)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+         * due to `physical_window_top_pud` and `physical_window_{base, top}_aligned`
+         * we can now fill out every PD in each PUD without worrying about edgecases */
+        for (word_t pd_idx = 0; pd_idx < BIT(PT_INDEX_BITS); pd_idx++) {
+
+            armKSGlobalKernelPDs[pud_idx][pd_idx]
+                = pte_pte_page_new(
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* verify that the kernel window as at the second entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
+                      /* XN */ 0,
```
Indanz: The no-execute bit can be set for the memory window now that we have a separate mapping for the kernel ELF.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+    assert(KERNEL_ELF_TOP <= KDEV_BASE);
+
+    /**
+     *  Per ARM ARM DDI 0487 (version L.b), in B2.11 "Mismatched memory attributes",
+     *  when physical memory locations are accessed with mismatched attributes
+     *  (caching, memory type (device/normal), shareability), the coherency of these
+     *  memory locations may be lost, as well as various violations of read and
+     *  write orderings or atomic operations. Other ARM documents refer to the
+     *  situation in which two virtual addresses map to the same physical address as
+     *  "Memory aliasing" (e.g. Document ID 102376 0200_01_en).
```
Indanz: Not sure this comment adds useful information here. Maybe good to mention where we do the mapping that it should avoid mismatching memory attributes, but I'd leave it at that.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+     *
+     *  Currently, the memory attributes used in the physical memory window does
+     *  match that of the kernel ELF mapping, so this is OK. At the same time,
+     *  the current physical window mapping does use NORMAL memory covering memory
+     *  that is not DRAM, which is questionable for newer processors with more
+     *  speculative execution.
```
Indanz: It's questionable for old in-order CPUs too because they have HW pre-fetchers. It's always wrong. We just get lucky because the distance is great enough.

It's probably fine if it stays within the DDR controller area and there is no backing RAM, but if there are devices mapped, the reads can have side effects.

Indanz (commented):
Please focus this PR on the fixed virtual address change and do the cleanups and improvements in a separate PR. Small, trivial changes are okay if they're combined, but they need to be relevant.

Updating the diagram is a border case, I think doing that in this PR is okay because you need to update it for this change anyway and it was wrong and misleading. But please split it up in two parts: One that updates it for this PR, and one that does the generic improvements. The order doesn't matter and the update necessary because of commit 58a2 is best done as part of that commit.

include/machine.h
midnightveil: Yes, but I'm being conservative.

I think I can add an assert here, I'll double check. I had the feeling it would have been hard but it should be doable...

include/arch/arm/arch/64/mode/hardware.h
midnightveil: This is unmodified comments (I didn't adjust HYP) but I can update this.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: This is unmodified comments (I didn't adjust HYP) but I can update this.

src/arch/arm/64/kernel/vspace.c
midnightveil: This is from the conversation I had with @kent-mcleod who agreed I should document these findings in the code.

Again, trying to be conservative with what the kernel needs vs what we have, since we don't *need* to rely on the fact that the kernel memory and phys window are mapped with same perms. E.g. if we decided to map it as device memory (which some of it actually) in case we wanted to mitigate speculation into device registers.

src/arch/arm/64/kernel/vspace.c
midnightveil: Yes.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
midnightveil: https://github.com/seL4/seL4/issues/1175

Hmm, OK I must have misunderstood this.

Will remove...

src/arch/arm/64/kernel/vspace.c
midnightveil: Fixed now by adding alignment.

midnightveil:
hw-test actions run is: https://github.com/seL4/seL4/actions/runs/18517280847/

looks like passing aside from unrelated boot failures.

I'll do more work to get the changes into shape probably tomorrow.

include/arch/arm/arch/64/mode/hardware.h
kent-mcleod: Comment was added in fbec286d97185011770b9343d8ac8e2f7a99455b.
I agree with Indan that it's because there's only a single TTBR0 in this EL2 VMSA

include/machine.h
Indanz: Well, actually, this is not so simple, because it is caller dependent. Most use it read-only or just want to have the pointer value, but some TLB code does modify memory via the pointer. And that can be page tables in the ELF mapping if it are the kernel's own page tables.

src/arch/arm/64/kernel/vspace.c
Indanz: With that in mind, practically it's mostly a problem for prefetching before the actual RAM into device memory. So in case of problems, perhaps check if there are devices somewhat near the start of SDRAM (thinking about the zcu106 here. Edit: But that was Aarch32 though).

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
Indanz: That issue is that it's not possible to handle a physical address the same size as the supported virtual address space size because you don't have enough bits to map anything else. That's true for any address size, including 32-bit.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: https://github.com/seL4/seL4/pull/1523/files#r2438356437

include/arch/arm/arch/64/mode/hardware.h
midnightveil: Split up in new PR.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: Resolved in new PR.

libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
midnightveil: Removed in new PR.

include/arch/arm/arch/64/mode/hardware.h
Indanz: I would have been happy with a separate commit in this PR, but new PR does make it easier to review.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
-    for (paddr = PADDR_BASE; paddr < PADDR_TOP; paddr += BIT(seL4_LargePageBits)) {
-        armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
-                                                                                         KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
+BOOT_CODE void map_kernel_elf_image(const word_t kernel_mappings_pud_idx)
+{
+    compile_assert(elf_paddr_aligned, IS_ALIGNED(KERNEL_ELF_PADDR_BASE_RAW, seL4_LargePageBits));
+    compile_assert(elf_base_aligned, IS_ALIGNED(KERNEL_ELF_BASE, seL4_LargePageBits));
+    /* asserts because ELF_TOP is dynamic */
+    assert(GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1)) == kernel_mappings_pud_idx);
+    assert(GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(1)) == kernel_mappings_pud_idx);
```
Indanz: Considering you know what it should be, wouldn't it make more sense to move the `armKSGlobalKernelPUD[kernel_mappings_pud_idx] = pte_pte_table_new(addrFromKPPtr(&armKSGlobalKernelPDs[kernel_mappings_pud_idx][0]));` here and remove the function argument?

Sadly the same can't be done for `map_kernel_devices` because it generic Arm code, so leaving it in `map_kernel_window` is more consistent. With that in mind, what about passing `&armKSGlobalKernelPDs[kernel_mappings_pud_idx]` as argument instead?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+     *  physical address might lie within the kernel ELF. (Reads are fine if they
+     *  have been flushed appropriately per B2.11; this should be the case for
+     *  any paging structures).
+     *
+     *  As an addendum, LWN Article "ARM's multiply-mapped memory mess" specifies
+     *  that in ARMv6 specifications this kind of aliasing would be UNPREDICTABLE;
+     *  the current ARMv8 A-Profile specifications no longer contain this wording,
+     *  and instead specify the loss of coherency and other behaviours in B2.11.
+     *
+     */
```
Indanz: I'd prefer if such big comments are outside the function body.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    const word_t kernel_elf_pd_start = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(2));
+    const word_t kernel_elf_pd_end   = GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(2));
```
Indanz: Please use shorter names for local variables. In this case, all PDs are about kernel ELF mappings, so drop that bit and just name it `pd_start` and `pd_end`.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    word_t paddr = KERNEL_ELF_PADDR_BASE;
+    word_t vaddr = KERNEL_ELF_BASE;
```
Indanz: Didn't we have specific typedefs for these? We have `pptr_t`/`paddr_t` (not sure when to use which though) and `vptr_t`.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+    for (word_t pd_idx = kernel_elf_pd_start; pd_idx < kernel_elf_pd_end; pd_idx++) {
```
Indanz: `i` instead of `pd_idx` would be clearer.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
-                                                                                                                    );
+                  /* page_base_address */ paddr,
+                  /* global (nG) */ 0,
+                  /* access flag (AF) */ 1,
+                  /* shareability (SH) */ SMP_TERNARY(SMP_SHARE, 0),
+                  /* AP */ APFromVMRights(VMKernelOnly),
+                  /* AttrIndx */ NORMAL
+              );
+
+        assert(pd_idx == GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(2)));
```
Indanz: You already checked the first index with above, I would move this one after the increments. Then you always check things before doing something.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
+
+    /* ============ 2: setup the physical memory window mappings ============ */
+    map_kernel_physical_window();
+
+    /* ========== 3: setup other kernel mappings (ELF, devices, etc) ======== */
+
+    /* put the PD into the PUD for kernel mappings */
+    armKSGlobalKernelPUD[kernel_mappings_pud_idx]
+        = pte_pte_table_new(addrFromKPPtr(&armKSGlobalKernelPDs[kernel_mappings_pud_idx][0]));
```
Indanz: Move this into `map_kernel_elf_image()`.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    map_kernel_elf_image(kernel_mappings_pud_idx);
 
     /* put the PT into the PD for device window */
-    armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][BIT(PT_INDEX_BITS) - 1] = pte_pte_table_new(
-                                                                               addrFromKPPtr(armKSGlobalKernelPT)
-                                                                           );
+    armKSGlobalKernelPDs[kernel_mappings_pud_idx][device_mapping_pd_idx]
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPT));
```
Indanz: I'd move this into `map_kernel_devices`, except that's shared with aarch32.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,68 +233,173 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
     }
+}
 
-    /* put the PD into the PUD for device window */
-    armKSGlobalKernelPUD[GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1))] = pte_pte_table_new(
-                                                                                addrFromKPPtr(&armKSGlobalKernelPDs[BIT(PT_INDEX_BITS) - 1][0])
-                                                                            );
+/* work around the fact that C doesn't have real constants */
+#define kernel_mappings_pud_idx (GET_KPT_INDEX(PPTR_TOP, KLVL_FRM_ARM_PT_LVL(1)))
+#define device_mapping_pd_idx (GET_KPT_INDEX(KDEV_BASE, KLVL_FRM_ARM_PT_LVL(2)))
```
Indanz: A `static const` is practically a real constant, or a non-static `const` with limited scope. So please declare these as local variables and use them only in this function.

And with the kernel C code al being combined during build, adding defines like these can have unexpected side effects, though unlikely with the names you've chosen.

src/arch/arm/64/kernel/vspace.c
midnightveil: There is a spot where we have two different indices and I didn't want to name them I & J so I tried being consistent.

src/arch/arm/64/kernel/vspace.c
midnightveil: Yes, I wish that was possible too.

src/arch/arm/64/kernel/vspace.c
midnightveil: static const wasn't enough to get this to compile for some of the CI builds on certain arches ("non-constant expression"). I don't know why.

open for other ways to deal with this ...

Indanz: That's weird, I checked all used defines and they all seem constant on Aarch64.

midnightveil: This commit: https://github.com/seL4/seL4/commit/34cf10180f1c32933fafd5c2fa0ca08e56a682e8 fixed the previous broken commit https://github.com/seL4/seL4/commit/853f87cbb9ce6760b71d2e3d9553cff129bdbc40 as you can see https://github.com/seL4/seL4/actions/runs/18336079121/job/52220453248

Indanz: Oh, of course. No, in that sense they're not constant expressions. I thought you meant it gave errors for the declaration itself.

src/arch/arm/64/kernel/vspace.c
midnightveil: Although perhaps that's not the correct solution, since the issue isn't that the physBase needs to be aligned but that the kernelElfBase does ..

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -173,42 +240,50 @@
...
+/* The base address in virtual memory to use for kernel devices, which are
+ * mapped using a PT containing 4K pages.  */
+#define KDEV_BASE UL_CONST(0x000000ffffe00000)  /* 2^48 - 2^21 */
+#else
+#define KDEV_BASE UL_CONST(0xffffffffffe00000)  /* 2^64 - 2^21 */
+#endif
+
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* The log buffer is placed before the device region using a 2MiB page. */
+#define KS_LOG_BASE UL_CONST(0x000000ffffc00000)  /* 2^48 - 2 * 2^21 */
```
midnightveil: apparently I renamed this from KS_LOG_PPTR which breaks benchmark builds. I should fix this before this is merged.

midnightveil:
I split up the commits a bit more, hopefully that's helpful @Indanz. I've also rebased after the docs changes, though this still includes #1543 just to reduce the merge effort. I'll unmark as draft once that's merged.

I have also put the fixed window changes before the cleanup, if we don't want any of the cleanups that should be easier.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -237,25 +238,36 @@
...
+/* The log buffer is placed before the device region using a 2MiB page. */
+#define KS_LOG_BASE UL_CONST(0x000000ffffc00000)  /* 2^40 - 2 * 2^21 */
+#else
+#define KS_LOG_BASE UL_CONST(0xffffffffffc00000)  /* 2^64 - 2 * 2^21 */
+#endif
 
-/* The log buffer is placed before the device region */
-#define KS_LOG_PPTR (KDEV_BASE - UL_CONST(0x200000))
+/* For consistency with the rest of the code. Changing the code might be a good idea. */
+#define KS_LOG_PPTR KS_LOG_BASE
```
Indanz, line 292: Well, `KS_LOG_BASE` is not used anywhere, why not remove it and only define `KS_LOG_PPTR`?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
                                                                                             attr_index);
 }
 
+/* Essentially, assert that the ELF, kernel devices, and log buffer are all in the same PUD.
+   In the current code this is the top index. This is arbitrary, the important part
+   is that they are all the same.
+ */
+compile_assert(elf_in_kernel_mappings, GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+compile_assert(kdev_in_kernel_mappings, GET_KPT_INDEX(KDEV_BASE, KLVL_FRM_ARM_PT_LVL(1))      == BIT(PT_INDEX_BITS) - 1);
+compile_assert(log_buf_in_kernel_mappings, GET_KPT_INDEX(KS_LOG_BASE, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
```
Indanz: Maybe add a define for that index and use that everywhere (also below) instead of `BIT(PT_INDEX_BITS) - 1`?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+ */
+BOOT_CODE void map_kernel_elf_image(pte_t kernel_mapping_pds[])
+{
+    /** This is not strictly necessary, but to handle a non-aligned KERNEL_ELF_PADDR_BASE
+     *  we'd need to do a bit more logic and probably do what riscv64 does
+     *  where the *actual* KERNEL_ELF_BASE is rounded up to the alignment.
+     *  TODO: Should we do this?
+     **/
+    compile_assert(elf_paddr_aligned, IS_ALIGNED(KERNEL_ELF_PADDR_BASE_RAW, seL4_LargePageBits));
+    compile_assert(elf_base_aligned, IS_ALIGNED(KERNEL_ELF_BASE, seL4_LargePageBits));
```
Indanz, line 281: I think it's fine to require that both the kernel elf physical load address and virtual address are aligned on 2MB.

Aren't these checks static? Probably nicer to move them next to the other ones above.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+    assert(GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+
+    /* Please don't overlap (pd_end with KDEV...) */
+    assert(KERNEL_ELF_TOP <= KDEV_BASE);
+
+    const word_t pd_start = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(2));
+    const word_t pd_end   = GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(2));
+
+    paddr_t paddr = KERNEL_ELF_PADDR_BASE;
+    pptr_t vaddr = KERNEL_ELF_BASE;
```
Indanz: New line please.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+     *  TODO: Should we do this?
+     **/
+    compile_assert(elf_paddr_aligned, IS_ALIGNED(KERNEL_ELF_PADDR_BASE_RAW, seL4_LargePageBits));
+    compile_assert(elf_base_aligned, IS_ALIGNED(KERNEL_ELF_BASE, seL4_LargePageBits));
+
+    /* Unfortunately, we can't check this at build time because KERNEL_ELF_TOP isn't constant.
+       This is equivalent to saying the kernel elf is <1GiB in size.
+       See `elf_in_kernel_mappings` above.
+    */
+    assert(GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
```
Indanz: This is already checked by the combination of `elf_in_kernel_mappings` and the next assert, so please remove.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    /* Please don't overlap (pd_end with KDEV...) */
+    assert(KERNEL_ELF_TOP <= KDEV_BASE);
```
Indanz: Suggestion: "Please don't overlap with KDEV_BASE."

However, that is not good enough, you also have to make sure it doesn't overlap with `KS_LOG_PPTR`.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    const word_t pd_start = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(2));
+    const word_t pd_end   = GET_KPT_INDEX(KERNEL_ELF_TOP, KLVL_FRM_ARM_PT_LVL(2));
```
Indanz: `pd_end` is defined by `ki_end`, which is 4k aligned, not 2MB aligned, which makes `pd_end` inclusive.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+    paddr_t paddr = KERNEL_ELF_PADDR_BASE;
+    pptr_t vaddr = KERNEL_ELF_BASE;
+    for (word_t pd = pd_start; pd < pd_end; pd++) {
```
Indanz, line 295: I think this should be `pd <= pd_end`, see above.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,87 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+        kernel_mapping_pds[pd] = pte_pte_page_new(
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+            /* XN */ 0,
+#else
+            /* UXN */ 1,
+#endif
+            /* page_base_address */ paddr,
+            /* global (nG) */ 0,
+            /* access flag (AF) */ 1,
+            /* shareability (SH) */ SMP_TERNARY(SMP_SHARE, 0),
```
Indanz: I think this should be inner shareable like on SMP for all our kernel mappings, to avoid any weird problems. (But probably not as part of this PR.)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -284,6 +365,13 @@ BOOT_CODE void map_kernel_window(void)
         vaddr += BIT(seL4_LargePageBits);
     }
 
+
+    const word_t kernel_mappings_pud_idx = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1));
+    pte_t *kernel_mapping_pds = armKSGlobalKernelPDs[kernel_mappings_pud_idx];
+
+    /* map the kernel ELF */
+    map_kernel_elf_image(kernel_mapping_pds);
```
Indanz: You should call it as `map_kernel_elf_image(armKSGlobalKernelPDs[kernel_mappings_pud_idx])` to conform to the calling convention you specified.

tools/hardware/config.py
```diff
@@ -45,7 +45,10 @@ def __init__(self, sel4arch, addrspace_max):
         elif sel4arch == 'aarch32':
             self.KERNEL_PHYS_ALIGN = 24
         else:
-            self.KERNEL_PHYS_ALIGN = 0
+            # AArch64 requires an alignment up to the size of seL4_LargePageBits,
+            # see the compile_assert `elf_paddr_aligned`.
+            # FIXME: We could not enforce this alignment?
+            self.KERNEL_PHYS_ALIGN = 21
```
Indanz, line 53: Probably good to say something like:

"On AArch64, `PADDR_BASE` is always zero and only `KERNEL_ELF_BASE` uses `PHYS_BASE_RAW`, which is influenced by this alignment. `KERNEL_ELF_BASE` must be aligned to seL4_LargePageBits."

Indanz (commented):
Looks almost ready to me.

include/arch/arm/arch/64/mode/machine/hardware.h
```diff
@@ -35,7 +35,8 @@
 
 #define KPT_LEVEL_SHIFT(n) (((PT_INDEX_BITS) * (((KPT_LEVELS) - 1) - (n))) + seL4_PageBits)
 #define GET_KPT_INDEX(addr, n)  (((addr) >> KPT_LEVEL_SHIFT(n)) & MASK(PT_INDEX_BITS))
-#define GET_KLVL_PGSIZE(n)      BIT(KPT_LEVEL_SHIFT((n)))
+#define GET_KLVL_PGSIZE_BITS(n) KPT_LEVEL_SHIFT(n)
+#define GET_KLVL_PGSIZE(n)      BIT(GET_KLVL_PGSIZE_BITS(n))
```
Indanz: What's the point of this?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -357,7 +357,7 @@ BOOT_CODE void map_kernel_physical_window(void)
         armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
                                                                                          KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-                                                                                                                        0, // XN
+                                                                                                                        1, // XN
 #else
                                                                                                                         1, // UXN
 #endif
```
Indanz, line 363: Hmm, shouldn't this be done in the other commit, so hyp and non-hyp behaviour stays in sync?

Indanz (commented):
Previous review was for the second commit only, this is for the rest (ignoring the doc one).

Still not sure about commit "nfc/cleanup: fill physical window using array indices", my feeling is that it only makes sense if you also switch to huge pages.

src/arch/arm/64/kernel/vspace.c
midnightveil: Yeah, that's what I had done. When I rebased that I removed it since it was somewhat ugly, but I can definitely go back to it.

src/arch/arm/64/kernel/vspace.c
midnightveil: If I did KERNEL_ELF_TOP - 1 as well that would work regardless of the alignment then yes?

Is there a better way to do that? It feels somewhat ugly.

(If it did get 2MiB align we make an extra unnecessary 2MiB. Maybe that's fine?)

src/arch/arm/64/kernel/vspace.c
midnightveil: I don't exactly understand why you say this currently doesn't match the "calling convention"? The types match and it's the same thing.?

tools/hardware/config.py
midnightveil: *KERNEL_ELF_PADDR_BASE, not KERNEL_ELF_BASE

Other comment is that this will reserve the memory from bottom of RAM up to the alignment because of how these scripts work, but that's probably fine.

include/arch/arm/arch/64/mode/machine/hardware.h
midnightveil: That is a very good question. I guess it's no longer used after I cleaned the code up, so I can probably remove it.

I did it because it's clearer the intent, but no longer using the intermediate one directly so I can drop that commit.

src/arch/arm/64/kernel/vspace.c
midnightveil: In what other commit? I'm confused, this is a standalone commit, and I'm not sure the relevance of hyp vs non-hyp here.

midnightveil:
> Still not sure about commit "nfc/cleanup: fill physical window using array indices", my feeling is that it only makes sense if you also switch to huge pages.

Well, the fun thing is the PUDs were already filled with indices, heh. So I was only really changing it for PDs.

I think actually after the rest of the cleanup I think it seems alright. Maybe I'll just explain it with some comments a bit more and drop the change.

src/arch/arm/64/kernel/vspace.c
Indanz: `map_kernel_elf_image()` takes a `param[]`, while kernel_mapping_pds is a `pte_t*`.

midnightveil: That's only because there isn't a nice way to type `pte_t kernel_mapping_pds[]` as a local variable in C, though. They're the same type.

tools/hardware/config.py
Indanz: Agreed.

src/arch/arm/64/kernel/vspace.c
Indanz: Sorry, I thought you set UXN to 1 in the other commit, but that's already 1 in master. Why was this different for hyp, any idea?

midnightveil: Because UXN is userspace execute never, I don't think we have control over the kernel execute never bits in hyp mode, i.e. it's always executable at the kernel level (IIRC).

src/arch/arm/64/kernel/vspace.c
midnightveil: I could make map_kernel_elf_image take a `pte_t *`?

This is really just arcane limitations of C here, annoyingly.

src/arch/arm/64/kernel/vspace.c
Indanz: If it's uglier than `BIT(PT_INDEX_BITS) - 1`, your define's name is definitely too long.

midnightveil: As in, having to do a `#define` because C doesn't support proper constants, is ugly.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: Probably could just. I'd prefer if it wasn't called `KS_LOG_PPTR` because we don't e.g., say `KERNEL_ELF_PPTR` but I can just remove this and define it as PPTR.

src/arch/arm/64/kernel/vspace.c
Indanz: Yes, strictly speaking also doing KERNEL_ELF_TOP - 1 is more correct (and then -1 on the same line). But the chance of ki_end falling exactly on a 2MB boundary is miniscule, and what's the harm? (Famous last words...)

No, making the binary 2MB bigger is not fine.

midnightveil: > chance of ki_end falling exactly on a 2MB boundary is miniscule, a

well, say, for instance, someone editing the linker script such that the kernel takes up exactly one 2MiB page :)

What do you mean by "then -1 on the same line"?

src/arch/arm/64/kernel/vspace.c
Indanz: If you define the array size, the compiler should give you a warning or error for this. Without size, it practically the same, though I believe there is a subtle difference, which I forgot and may not apply to function parameters anyway.

midnightveil: Hmm, I don't think so. C doesn't actually care because pointer decay. I think if you do `array_argument[static 512]` then it actually cares, but I don't know it it would check if you pass a pointer, only a differently-sized array directly.

My point is that it's impossible to make a local variable that is array-typed where the array isn't stored on the stack. (I believe; you can do a *pointer to an array-type* but that's different). The way you have to do it is with a pointer.

src/arch/arm/64/kernel/vspace.c
Indanz: Sorry, meant +1.
```
const word_t pd_end  = GET_KPT_INDEX(KERNEL_ELF_TOP - 1, KLVL_FRM_ARM_PT_LVL(2)) + 1;
```
and then keep using `<` in the compare.

src/arch/arm/64/kernel/vspace.c
Indanz: Ah, yes, hyp is using stage2 page tables only.

For non-hyp/stage1 we can set the "Privileged eXecute Never" bit now.

midnightveil:
> I think actually after the rest of the cleanup I think it seems alright. Maybe I'll just explain it with some comments a bit more and drop the change.

I've decided to do this.

Here's that patch for posterity.

[0001-nfc-cleanup-fill-physical-window-using-array-indices.patch](https://github.com/user-attachments/files/23466612/0001-nfc-cleanup-fill-physical-window-using-array-indices.patch)

midnightveil:
HW build run: https://github.com/seL4/seL4/actions/runs/19315663992/job/55246264645?pr=1516

Everything is building now OK, this is waiting on verification input.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+/**
+ *  Per ARM ARM DDI 0487 (version L.b), in B2.11 "Mismatched memory attributes",
+ *  when physical memory locations are accessed with mismatched attributes
+ *  (caching, memory type (device/normal), shareability), the coherency of these
+ *  memory locations may be lost, as well as various violations of read and
+ *  write orderings or atomic operations. Other ARM documents refer to the
+ *  situation in which two virtual addresses map to the same physical address as
+ *  "Memory aliasing" (e.g. Document ID 102376 0200_01_en).
+ *
+ *  Currently, the memory attributes used in the physical memory window does
```
JDuchniewicz, line 256:
```suggestion
 *  Currently, the memory attributes used in the physical memory window do
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+        paddr += BIT(seL4_LargePageBits);
+        vaddr += BIT(seL4_LargePageBits);
+    }
+}
 
-    /* place the PUD into the PGD */
-    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] = pte_pte_table_new(
-                                                                                 addrFromKPPtr(armKSGlobalKernelPUD));
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* verify that the physical memory window as at the second entry of the PGD */
```
JDuchniewicz, line 318:
```suggestion
/* verify that the physical memory window is the second entry of the PGD */
```

If I understand correctly ofc.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+compile_assert(physical_window_pgd_location,
+               GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
+#else
+/* verify that the physical memory window as at the last entry of the PGD */
```
JDuchniewicz, line 322: Same as above

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+        paddr += BIT(seL4_LargePageBits);
+        vaddr += BIT(seL4_LargePageBits);
+    }
+}
 
-    /* place the PUD into the PGD */
-    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] = pte_pte_table_new(
-                                                                                 addrFromKPPtr(armKSGlobalKernelPUD));
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+/* verify that the physical memory window is at the second entry of the PGD */
```
Indanz: Any idea why it isn't using the first entry?

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+    pptr_t vaddr = KERNEL_ELF_BASE;
 
+    for (word_t pd = pd_start; pd < pd_end; pd++) {
+        assert(pd == GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(2)));
+
+        kernel_mapping_pds[pd] = pte_pte_page_new(
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* verify that the kernel window as at the second entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
+                                     /* XN */ 0,
```
Indanz: I think we can set this to 1 too, as it controls execution at EL1 and EL0, not at EL2.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -283,16 +371,30 @@ BOOT_CODE void map_kernel_window(void)
...
+BOOT_CODE void map_kernel_window(void)
+{
+    /* place the PUD into the PGD */
+    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))]
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
+
+    map_kernel_physical_window();
+
+    const word_t kernel_mappings_pud_idx = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1));
+    pte_t *kernel_mapping_pds = armKSGlobalKernelPDs[kernel_mappings_pud_idx];
```
Indanz: To increase readability, it really helps to use short names for variables that are used for indexing. In this case it's the only PUD index used, so why not call it `pud_idx`? This is the map_kernel_window function, so from context it's already clear that this is the kernel mapping pud.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+
+        kernel_mapping_pds[pd] = pte_pte_page_new(
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    /* verify that the kernel window as at the second entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == 1);
+                                     /* XN */ 0,
 #else
-    /* verify that the kernel window as at the last entry of the PGD */
-    assert(GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0)) == BIT(PT_INDEX_BITS) - 1);
+                                     /* UXN */ 1,
```
Indanz, line 302: Hm, why is one 0 and the other 1 again? Both only control execute-never for lower levels than the kernel, so I'd expect both to be the same.

(The only practical effect it seems to have is whether code prefetching is allowed, but that might only apply to uncached memory.)

src/arch/arm/64/kernel/vspace.c
midnightveil: No, just that Kent made this change as it otherwise caused issues in the elfloader. a94d90598fec484636b14e1b8d3643156e1e27b1.

src/arch/arm/64/kernel/vspace.c
midnightveil: Previously discussed: https://github.com/seL4/seL4/pull/1516#discussion_r2496647326

src/arch/arm/64/kernel/vspace.c
Indanz: Yeah, I saw that commit, but it doesn't make sense.

kent-mcleod: It was less of a breaking change. The shifting of the kernel window changed the program header addresses. All the seL4 loaders that already expected to load the kernel to a region higher than the identity mapped region had two sets of page table structures. Making kernel window start from 0 would have meant the kernel would link to an identity mapping and that was a breaking change for the elfloaders that I didn't want to have to manage the fallout for at the time.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+                                     /* page_base_address */ paddr,
+                                     /* global (nG) */ 0,
+                                     /* access flag (AF) */ 1,
+                                     /* shareability (SH) */ SMP_TERNARY(SMP_SHARE, 0),
+                                     /* AP */ APFromVMRights(VMKernelOnly),
+                                     /* AttrIndx */ NORMAL
+                                 );
+
+        paddr += BIT(seL4_LargePageBits);
+        vaddr += BIT(seL4_LargePageBits);
```
midnightveil, line 313: These paddr/vaddr variables look like rebase remannds I need to fix.

lsf37: clang is complaining about `vaddr` being unused (correctly), although apparently only in release builds. `paddr` is used in the statement above.

midnightveil: Right, this actually isn't a rebase remmant. I'm not sure why this wasn't an issue before; but it's because it was only in the assert. I did do full hw-test runs before so not sure why it only failed recently, but anyway.

I've added the `UNUSED` attribute.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
 
-BOOT_CODE void map_kernel_window(void)
+/* Essentially, assert that the ELF, kernel devices, and log buffer are all in the same PUD.
+   In the current code this is the top index. This is arbitrary, the important part
+   is that they are all the same.
+ */
+compile_assert(elf_in_kernel_mappings, GET_KPT_INDEX(KERNEL_ELF_BASE,
+                                                     KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
+compile_assert(kdev_in_kernel_mappings, GET_KPT_INDEX(KDEV_BASE,
+                                                      KLVL_FRM_ARM_PT_LVL(1))      == BIT(PT_INDEX_BITS) - 1);
```
lsf37:
```suggestion
                                                      KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+ *  physical address might lie within the kernel ELF. (Reads are fine if they
+ *  have been flushed appropriately per B2.11; this should be the case for
+ *  any paging structures).
+ *
+ *  As an addendum, LWN Article "ARM's multiply-mapped memory mess" specifies
+ *  that in ARMv6 specifications this kind of aliasing would be UNPREDICTABLE;
+ *  the current ARMv8 A-Profile specifications no longer contain this wording,
+ *  and instead specify the loss of coherency and other behaviours in B2.11.
+ *
+ */
```
lsf37:
```suggestion
 */
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,28 +233,116 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+ *
+ *  Currently, the memory attributes used in the physical memory window does
+ *  match that of the kernel ELF mapping, so this is OK. At the same time,
+ *  the current physical window mapping does use NORMAL memory covering memory
+ *  that is not DRAM, which is questionable for newer processors with more
+ *  speculative execution. (FIXME)
+ *
+ *  Hence, the kernel should never access kernel ELF memory through the physical
+ *  memory window, i.e., never modify the result of `ptrFromPAddr()` when the
+ *  physical address might lie within the kernel ELF. (Reads are fine if they
```
lsf37:
```suggestion
 *  physical address lies within the kernel ELF. (Reads are fine if they
```

We should know whether it does or not, and we do know for all accesses in the kernel which region they are in.

lsf37 (approved):
Good to go from the verification side one the proofs in https://github.com/seL4/l4v/pull/1031 have passed.

src/arch/arm/64/kernel/vspace.c
Indanz: That still does the weird line split halfway GET_KPT_INDEX, making it hard to read. I'd go for an earlier new line and do something like:

```
compile_assert(kdev_in_kernel_mappings,
               GET_KPT_INDEX(KDEV_BASE, KLVL_FRM_ARM_PT_LVL(1)) == BIT(PT_INDEX_BITS) - 1);
```
But if we want to check that they are all equal, we could check that assumption directly with:

```
compile_assert(elf_and_kdev_in_same_pud,
               GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(1)) ==
               GET_KPT_INDEX(KDEV_BASE,       KLVL_FRM_ARM_PT_LVL(1)));
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,6 +233,88 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
...
+    compile_assert(elf_base_aligned, IS_ALIGNED(KERNEL_ELF_BASE, seL4_LargePageBits));
+
+    /* Please don't overlap with KDEV_BASE */
+    assert(KERNEL_ELF_TOP <= KDEV_BASE);
+    /* Please don't overlap with KS_LOG_PPTR */
+    assert(KERNEL_ELF_TOP <= KS_LOG_PPTR);
+
+    const word_t pd_start = GET_KPT_INDEX(KERNEL_ELF_BASE, KLVL_FRM_ARM_PT_LVL(2));
+    /* handle the fact that ELF_TOP is exclusive, not inclusive */
+    const word_t pd_end   = GET_KPT_INDEX(KERNEL_ELF_TOP - 1, KLVL_FRM_ARM_PT_LVL(2)) + 1;
```
Indanz: Really worth commenting here that `pd_end` is exclusive (because of the +1). That's very easy to miss, especially with the comment seemingly implying that we want an inclusive value.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -283,16 +371,30 @@ BOOT_CODE void map_kernel_window(void)
                                                                                                                     );
         vaddr += BIT(seL4_LargePageBits);
     }
+}
+
+BOOT_CODE void map_kernel_window(void)
+{
+    /* place the PUD into the PGD */
+    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))]
+        = pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
```
Indanz:
```suggestion
    armKSGlobalKernelPGD[GET_KPT_INDEX(PPTR_BASE, KLVL_FRM_ARM_PT_LVL(0))] =
        pte_pte_table_new(addrFromKPPtr(armKSGlobalKernelPUD));
```
very unusual in C to put the "=" on the next line when you break it up.

(But this monstrosity can be easily fixed by splitting it up and using local vars for the indices.)

Indanz (approved):
Other than some nitpicks all seems fine.

kent-mcleod:
@lsf37, When your proof tests have finished, can you enable this to perform the hardware test runs?

lsf37:
Proofs have passed (they do only say very little about this, the proof change is just the constant value of the KERNEL_ELF_BASE).

lsf37:
Given that this will only really be useful for multikernel together with additional work on top, we could also merge this just after the release if we want to make sure there are no other unforeseen secondary effects (beyond sel4test and sel4bench, which I have started).

It's basically ready to go after fixups, it's just a question of merging it before or after the release point.

midnightveil:
Yeah, I think it's worth leaving off until after. It doesn't really produce any benefits to have it now and we could have missed something that breaks everything.

Indanz:
I would merge this after release, as it doesn't directly benefit people right now while it introduces fairly big changes which are outside of verification. And as you said, to make it useful for multikernel additional changes are needed, this is just a prerequisite for those.

src/arch/arm/64/kernel/vspace.c
midnightveil: We'd still need another assert since there's 3 things in that top placement. Plus a few other spots assume `BIT(PT_INDEX_BITS) - 1` in this code. In theory this could also affect the loader, but so could all of this. (both Microkit and elfloader will continue working)

(applied Gerwin's suggestion)

src/arch/arm/64/kernel/vspace.c
midnightveil: Reworded.

src/arch/arm/64/kernel/vspace.c
midnightveil: Renaming to `elf_pud_idx`.

src/arch/arm/64/kernel/vspace.c
midnightveil: Moved.

PR #1518, opened 2025-10-15
cmake: Error message warning fix

kent-mcleod:
PLATFORM is not a variable defined by the Kernel CMake build and isn't always how the kernel platform is configured. For example the verified configs in configs/ just set KernelPlatform directly.

I was reviewing https://github.com/seL4/seL4/pull/1467 and noticed the PLATFORM may not always be the way that the user tried to configure the platform.

Try changing the platform name in `configs/AARCH64_verified.cmake` to an invalid name and build the config it would not print the right value out.

Because KernelPlatform is cleared by the config_choice on line 203 if it is invalid, we can save the old value and then use it to print out in the error string if the platform name was invalid.

### DIFF
--- a/configs/seL4Config.cmake
+++ b/configs/seL4Config.cmake
@@ -199,6 +199,7 @@ foreach(file ${result})
     include("${file}")
 endforeach()
 
+set(kernel_platform_origin "${KernelPlatform}")
 config_choice(KernelPlatform PLAT "Select the platform" ${kernel_platforms})
 
 # Verify that, as a minimum any variables that are used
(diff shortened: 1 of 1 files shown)


midnightveil (approved):
Fine by me, still works for my tests that I did in the previous PR locally.

Indanz (approved):
"origin" seems a weird name, compared to original, or "orig". But the error message is still misleading when KernelPlatform was set, but to an invalid value. Now you know the old value and whether it was empty or wrong, so you could print out different error messages that are less confusing.

configs/seL4Config.cmake
```diff
@@ -199,14 +199,15 @@ foreach(file ${result})
...
+set(kernel_platform_origin "${KernelPlatform}")
 config_choice(KernelPlatform PLAT "Select the platform" ${kernel_platforms})
 
 # Verify that, as a minimum any variables that are used
 # to find other build files are actually defined at this
 # point. This means at least: KernelPlatform KernelArch KernelWordSize
 
 if("${KernelPlatform}" STREQUAL "")
-    message(FATAL_ERROR "Variable 'KernelPlatform' is not set - is PLATFORM '${PLATFORM}' correct? \
+    message(FATAL_ERROR "Variable 'KernelPlatform' is not set - Was tested value:'${kernel_platform_origin}' correct? \
```
axel-h, line 210: Let's make the error message straight forward, don't try to explain internal details users are not aware of. We can explain more in a comment, which will help developers digging deeper into the scripts trying to analyze a specific issues.

```suggestion
    # value is cleared in case of an unsupported platform
    message(FATAL_ERROR "Unsupported KernelPlatform '${kernel_platform_origin}' \
```

midnightveil: This makes sense to do.

PR #1519, opened 2025-10-15
tools: keep align-reserved memory in 'devices'

midnightveil:
This is purely a bugfix, the kernel behaves exactly the same either way. This is only a difference in the generated `platform_gen` files to respect reality.

---

`align_memory()` in hardware.py both modifies the first normal memory region to adjust the base of it for alignment, and adds an extra reserved region to our list of reserved regions. This then feeds through `get_addrspace_exclude` which inverts the regions given and turns it into the available "device memory" at user-level.

    dev_mem = hardware.utils.memory.get_addrspace_exclude(
        list(reserved) + phys_mem + kernel_devs, config)

Anything as an argument to this is not given as "device memory" by the kernel. (It does not precisely match how the kernel works).

However, since `align_memory()` has adjusted both the phys_mem up (which *would* have added this region as "device" memory) but also added it to "reserved" region, which then made it disappear entirely, as "reserved" regions are not exposed to userspace.

However, this **does not** match the behaviour of the kernel, as it was not reserved, so this behaviour did not match the untypeds given to userspace. This commit solves this by removing the extra reserved region being added for that memory.

PR #1426 worked around this issue by removing the alignment on AArch64. Whilst this fixed the issue that microkit was seeing, it just masked the underlying issue. Reverting that PR, then applying this fix, results in the following platform_gen.yaml:

    devices:
    - end: 0x1000000 start: 0x0
    - end: 0xff800000 start: 0x3b400000
    - end: 0xff841000 start: 0xff801000
    - end: 0x100000000000 start: 0xff843000
    memory:
    - end: 0x3b400000 start: 0x1000000

Note especially the device region from 0x0 to 0x1000000; which is the combination of the 0x0 to 0x1000 reserved region, and the 0x1000 to 0x1000000 reserved by the kernel's alignment requirement. Previously, the platform_gen.yaml reported only the 0x0 to 0x1000 region,

    devices:
    - end: 0x1000 start: 0x0
    - end: 0xff800000 start: 0x3b400000
    - end: 0xff841000 start: 0xff801000
    - end: 0x100000000000 start: 0xff843000
    memory:
    - end: 0x3b400000 start: 0x1000000

I will be following this commit up with a PR to instead make the alignment-reserved region into a new memory region, since there's not any reason why userspace can't use this memory.

This has been tested on a few platforms with sel4test and with microkit on the pi4B.

### DIFF
--- a/tools/hardware/outputs/c_header.py
+++ b/tools/hardware/outputs/c_header.py
@@ -212,7 +212,7 @@ def run(tree: FdtParser, hw_yaml: HardwareYaml, config: Config, kernel_config_di
     if not args.header_out:
         raise ValueError('You need to specify a header-out to use c header output')
 
-    physical_memory, reserved, physBase = hardware.utils.memory.get_physical_memory(tree, config)
+    physical_memory, physBase = hardware.utils.memory.get_physical_memory(tree, config)
     kernel_regions, kernel_macros = get_kernel_devices(tree, hw_yaml, kernel_config_dict)
 
     create_c_header_file(
--- a/tools/hardware/outputs/json.py
+++ b/tools/hardware/outputs/json.py
@@ -56,10 +56,9 @@ def run(tree: FdtParser, hw_yaml: HardwareYaml, config: Config,
     if not args.json_out:
         raise ValueError('you need to provide a json-out to use the JSON output method')
 
-    phys_mem, reserved, _ = hardware.utils.memory.get_physical_memory(tree, config)
+    phys_mem, _ = hardware.utils.memory.get_physical_memory(tree, config)
     kernel_devs = get_kernel_devices(tree, hw_yaml, kernel_config_dict)
-    dev_mem = hardware.utils.memory.get_addrspace_exclude(
-        list(reserved) + phys_mem + kernel_devs, config)
+    dev_mem = hardware.utils.memory.get_addrspace_exclude(phys_mem + kernel_devs, config)
 
     create_json_file(dev_mem, phys_mem, args.json_out)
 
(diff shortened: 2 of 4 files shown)


kent-mcleod:
> However, this **does not** match the behaviour of the kernel, as it was not reserved, so this behaviour did not match the untypeds given to userspace. This commit solves this by removing the extra reserved region being added for that memory.

`get_addrspace_exclude()` isn't following the kernel's protocol properly. The `reserved` regions returned by `hardware.utils.memory.get_physical_memory()` aren't supposed to be reserved from device memory, just reserved from Kernel Memory. 

https://github.com/seL4/seL4/pull/1426 was addressing a valid issue that the aarch64 had an unnecessary alignment requirement that was only needed for aarch32 and this was causing a different `physBase` and `avail_p_regs` to be generated (the range  `[0x1000, 0x1000000)` was being excluded from being usable Kernel UT for the rpi platform). 

Some background that may have become lost is that this script is talking in device tree concepts in memory.py. `regions` correspond to the regions defined in the `memory` device tree node, while `reserved` regions correspond to the regions defined under the `/reserved-memory` node. Reserved regions are for describing DRAM regions that are reserved for other things in the platform like DMA regions for hardware accelerators. User space needs to be able to map them, but they can't be used for kernel Untyped.


`avail_p_regs` are the eligible regions on the platform that can become Normal untypeds.

The kernel boot code also has a list of 'reserved' regions, which is for regions that can't be given to user level in any UT cap, but this list is based on runtime calculations and is not related to the reserved regions returned by `hardware.utils.memory.get_physical_memory`.

kent-mcleod:
After discussing more with @midnightveil, it seems that there's no good reason for `get_physical_memory()` to be returning a reserved list because it's already been accounted for by subtraction from the `regions` list. And the only use of the returned reserved list is to incorrectly use it as a kernel reserved region list in `get_addrspace_exclude` which is the wrong assumption.

Indanz (approved):
Changes make sense to me, but I dread it breaking some other platform because of obscure reasons, e.g. RISCV. Would be good if we had a test that checks the generated memory regions between all output files.

PR #1520, opened 2025-10-15
asid_map_t,aarch32 feature (DRAFT)

kent-mcleod:
This is an updated patch set for aarch32 asid_map_t feature.

The asid_map_t feature associates additional metadata for a vspace root page table object by adding a larger entry in the seL4 asid pool for each asid ID. On aarch32 platforms there are less TLB hardware ASIDs than kernel vspace asids.
On aarch32 the kernel reserved a slot in the virtual address space root page table for holding the currently associated hw asid for the address space. 
With asid_map_t storing the hw_asid, this slot can be made available again. It returns up to 1MiB of memory back to user level that the kernel was reserving because the reserved slot at the end of the kernel mapping window.

### DIFF
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -82,8 +82,19 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
     }
 
 #ifdef CONFIG_ARCH_AARCH32
+     /* Need to test that the ASID is still valid */
+    asid_t asid = cap_page_directory_cap_get_capPDMappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
+    if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
+                 PD_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {
+        slowpath(SysCall);
+    }
+    /* Ensure the HW ASID is valid. */
+    if (unlikely(!asid_map_asid_map_vspace_get_hw_asid_valid(asid_map))) {
+        slowpath(SysCall);
+    }
     /* Get HW ASID */
-    stored_hw_asid = cap_pd[PD_ASID_SLOT];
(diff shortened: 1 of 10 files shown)


src/fastpath/fastpath.c
```diff
@@ -82,8 +82,19 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
...
+    asid_t asid = cap_page_directory_cap_get_capPDMappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
+    if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
+                 PD_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {
+        slowpath(SysCall);
+    }
+    /* Ensure the HW ASID is valid. */
+    if (unlikely(!asid_map_asid_map_vspace_get_hw_asid_valid(asid_map))) {
+        slowpath(SysCall);
+    }
```
Indanz, line 95: I don't understand the need for the first if-check: To me it seems that `asid_map_asid_map_vspace_get_hw_asid_valid()` should already catch `asid_map_get_type(asid_map) != asid_map_asid_map_vspace`, and that `PD_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd` should be impossible, considering both were looked up from `newVTable`.

src/fastpath/fastpath.c
```diff
@@ -361,8 +366,19 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
+    /* Get HW ASID */
+    stored_hw_asid.words[0] = asid_map_asid_map_vspace_get_hw_asid(asid_map);
```
Indanz, line 381: Maybe add a (static inline) helper function instead of repeating the same code.

Indanz (commented):
Looks like a good improvement, the old hack seems somewhat bizarre.

PR #1521, opened 2025-10-15
tools: give back alignment-reserved device memory as normal memory

midnightveil:
~~Also includes `kernel_phys_base` in the `platform_gen` files, so that tools don't need to read it out of the ELF file.~~ that's wrong, it's an assumption of aarch64

This is a followup to #1519 and as such includes its commit in this PR, please direct reviews of that commit to that PR instead. This PR includes a behavioural change in the kernel.

---

Taking the example I used in PR #1519 (raspberry pi after reverting
the alignment changes), the platform_gen is now:

    devices:
    - end: 0x1000
      start: 0x0
    - end: 0xff800000
      start: 0x3b400000
    - end: 0xff841000
      start: 0xff801000
    - end: 0x100000000000
      start: 0xff843000
    kernel_phys_base: 0x1000000
    memory:
    - end: 0x3b400000
      start: 0x1000

Userspace can now use the previously device [0x1000, 0x1000000) region as
normal memory.

For completeness, the elfloader bootlog:

    ELF-loader started on CPU: ARM Ltd. Cortex-A72 r0p3
      paddr=[10000000..102fa08f]
    No DTB passed in from boot loader.
    Looking for DTB in CPIO archive...found at 10038ab0.
    Loaded DTB from 10038ab0.
       paddr=[1237000..123dfff]
    ELF-loading image 'kernel' to 1000000
      paddr=[1000000..1236fff]
      vaddr=[ffffff8001000000..ffffff8001236fff]
      virt_entry=ffffff8001000000
    ELF-loading image 'rootserver' to 123e000
      paddr=[123e000..1633fff]
      vaddr=[400000..7f5fff]
      virt_entry=40e3f8
    Enabling MMU and paging

    Bootstrapping kernel
      [1000..3b400000)
      [ffffff8001000000..ffffff8001237000)
      [ffffff8001237000..ffffff800123d55f)
      [ffffff800123e000..ffffff8001634000)

### DIFF
--- a/include/arch/arm/arch/bootinfo.h
+++ b/include/arch/arm/arch/bootinfo.h
@@ -14,9 +14,10 @@
  * +1 for each available physical memory region (elements in avail_p_regs)
  * +1 for each MODE_RESERVED region, there might be none
  * +1 to allow the kernel to release its own boot data region
+ * +1 to allow the kernel to release memory below its alignment requirement
  * +1 for a possible gap between ELF images and rootserver objects
  */
-#define MAX_NUM_FREEMEM_REG (ARRAY_SIZE(avail_p_regs) + MODE_RESERVED + 1 + 1)
+#define MAX_NUM_FREEMEM_REG (ARRAY_SIZE(avail_p_regs) + MODE_RESERVED + 1 + 1 + 1)
 
 /* The regions reserved by the boot code are:
  * +1 for kernel
--- a/tools/hardware/utils/memory.py
+++ b/tools/hardware/utils/memory.py
@@ -83,20 +83,22 @@ def reserve_regions(regions: Set[Region], reserved: Set[Region]) -> Set[Region]:
     return ret
 
 
-def align_memory(regions: Set[Region], config: Config) -> List[Region]:
-    ''' Given a set of regions, sort them and align the first so that the
-    ELF loader will be able to load the kernel into it. Will return the
-    aligned memory region list, a set of any regions of memory that were
-    aligned out and the physBase value that the kernel will use. '''
+def find_aligned_kernel_phys_base(regions: Set[Region], config: Config) -> List[Region]:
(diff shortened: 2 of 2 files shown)


Indanz (commented):
The problem is that `physBase` as generated by the script is used differently per platform, so it has no definite meaning and is mostly confused with `KERNEL_ELF_PADDR_BASE`.

If talking about `physBase` as `PADDR_BASE`, giving back anything before that as normal memory makes zero sense.

I think the approach should be to give all normal memory to user space, but subtract reserved regions used by the kernel itself from that. If you do it that way, it is position independent with no weird corner cases or complexity.

But I think we should make the kernel position independent loading wise and then all this goes away, because it depends on where the kernel was loaded in memory, and that's then a user choice, as it should be.

midnightveil:
Yes, I'll add the additional note that this will likely break things on AArch32 which doesn't map any address below `physBase()` (becomes "PADDR_BASE"); whereas this is different on AArch64.

I think this is somewhat of a requirement for how AArch32 works, so unless we want to do this in an architecture-specific manner, I think this PR might not make sense?

Indanz:
> Yes, I'll add the additional note that this will likely break things on AArch32 which doesn't map any address below `physBase()` (becomes "PADDR_BASE"); whereas this is different on AArch64.

Aarch64 doesn't map anything below its physical base either. But physBase from the scripts becomes `KERNEL_ELF_PADDR_BASE` and it's not used for `PADDR_BASE`. It would make more sense to call this script variable elfBase, that would be less confusing.

> I think this is somewhat of a requirement for how AArch32 works, so unless we want to do this in an architecture-specific manner, I think this PR might not make sense?

No, I don't think this PR makes sense as it is, and it's hard to fix. The memory lost due to alignment should be returned as device memory anyway already, so it can still be used if necessary.

PR #1522, opened 2025-10-15
trivial: use KERNEL_ELF_TOP over ki_end

midnightveil:
Split out from PR 1516.

This is just being consistent.


Indanz:
It results in an extra cast, but it replaces an implicit one, and it's all in boot code or an assertion, so should be fine for verification.

Issue #957, opened 2022-12-08
AArch64 memory map diagram is wrong


lsf37:
During verification we were looking at requirements for the value of `PPTR_TOP`on AArch64 in EL2. 

It turns out that the actual `#define` is correct, or at least safe, but the comment and ASCII art diagram a) disagrees on the value and b) is incorrect.

We currently have

```
#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
#define PPTR_TOP UL_CONST(0x000000ffc0000000)
```

where `0x0000_00ff_c000_0000 = 2^40 - 2^30`

The diagram has `PPTR_TOP` at 2^48-2^30 instead. I assume from the comment at the top of the diagram that the 48 comes from the max width of virtual addresses.

This number would be too high, because in addition to respecting virtual address width, in EL2 the kernel window also needs to map one-to-one on to the intermediate physical address space, which is only 40 or 44 bits wide, depending on config (probably the reason for the 2^40 in the acutal PPTR_TOP definition).

So the actual constraint is that the translated kernel window does not go outside the interval `[ 0 .. 2^IPA_WIDTH (`.

lsf37:
I had an initial go at updating the diagram, but could then not trace where exactly the other parts of the diagram are defined that mention the 2^48 line and whether those should be updated as well or not, so it'd be good if someone who knows that layout better could have a look at that.

lsf37:
Right, #873 is the same issue, but I didn't find it when I checked. Since there is more description here, I'll add the comment from #873 here and close that one.

> Originally posted by @lsf37 in https://github.com/seL4/seL4/pull/869#pullrequestreview-1002088860
>
>> I think there might be another error in the picture for hypervisor mode. The picture puts PPTR_BASE at 2^48 - 2^39 (0xFF80_0000_0000), but the code puts it at 2^39 (0x80_0000_0000). @kent-mcleod I assume the code is correct here 
and the picture should be updated?
>
> See https://github.com/seL4/seL4/blob/master/include/arch/arm/arch/64/mode/hardware.h#L112-L155

PR #1523, opened 2025-10-17
aarch64: improve kernel address space layout docs

midnightveil:
To me, these diagrams have been very confusing, and were actually outdated/wrong, containing x86 PDPT naming and TLB bitmaps, which don't exist on AArch64.

Taken out of #1516.

### DIFF
--- a/include/arch/arm/arch/64/mode/hardware.h
+++ b/include/arch/arm/arch/64/mode/hardware.h
@@ -12,62 +12,111 @@
 #include <sel4/plat/api/constants.h>
 
 /*
+ * For information on the architectural details of the Armv8-A address translation,
+ * refer to the ARM "Armv8-A Address Translation", document ID 100940.
+ * https://developer.arm.com/documentation/100940/latest/
+ *
  * !defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
  *
- *          2^64 +-------------------+
- *               | Kernel Page PDPT  | --+
- *   2^64 - 2^39 +-------------------+ PPTR_BASE
- *               |    TLB Bitmaps    |   |
- *               +-------------------+   |
- *               |                   |   |
- *               |     Unmapped      |   |
- *               |                   |   |
- *   2^64 - 2^48 +-------------------+   |
- *               |                   |   |
- *               |   Unaddressable   |   |
- *               |                   |   |
- *          2^48 +-------------------+ USER_TOP
- *               |                   |   |
- *               |       User        |   |
- *               |                   |   |
- *           0x0 +-------------------+   |
- *                                       |
- *                         +-------------+
- *                         |
- *                         v
- *          2^64 +-------------------+
- *               |                   |
- *               |                   |     +------+
- *               | Kernel Page Table | --> |  PD  | ----+
- *               |                   |     +------+     |
- *               |                   |                  |
- *   2^64 - 2^30 +-------------------+ PPTR_TOP         |
- *               |                   |                  |
- *               |  Physical Memory  |                  |
- *               |       Window      |                  |
- *               |                   |                  |
- *               +-------------------+                  |
- *               |                   |                  |
- *               |                   |     +------+     |
- *               |    Kernel ELF     | --> |  PD  |     |
- *               |                   |     +------+     |
- *               |                   |                  |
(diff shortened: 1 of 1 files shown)


include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -11,155 +11,219 @@
...
+ *
+ *
+ *    The Kernel's PUD (`armKSGlobalKernelPUD`) is then split into several parts.
+ *
+ *    -  The physical memory window starts from the 0th PUD entry and extends
+ *       up to the second-to-last PUD entry. This covers 255GiB and contains the
+ *       virtual address for which the kernel ELF is mapped in to.
+ *
+ *       The physical memory window is mapped using Large Pages.
+         FIXME: Why not huge pages?
```
midnightveil: Linkback: https://github.com/seL4/seL4/pull/1516#discussion_r2412656515

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -83,10 +83,9 @@
  *    the 512 GiB at the top of that 48 bits of addressable
  *    virtual memory.
  *
- *    In EL2 there is no canonical-high portion of the address space since
- *    address tagging is not supported in EL2. Therefore the kernel is linked
- *    to the canonical lower portion of the address space (all the unused top bits
- *    are set to 0, not 1).
+ *    In EL2 there is only a TTBR1_EL2 register (unlike EL1 with TTBR0 and TTBR1)
```
Indanz: Only TTBR0_EL2, the missing one is TTBR1_EL2. FEAT_VHE did add TTBR1_EL2, we don't have support for that yet.

include/arch/arm/arch/64/mode/hardware.h
Indanz: Maybe for performance reasons. The TLB may not be shared and there may not be enough huge page entries in the TLB to cover both the kernel mapping and the memory used. But this seems unlikely though, that seemed to be an Intel specific design decision.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -12,62 +12,112 @@
...
+ *                        |                   |
+ *                        |                   |
+ *                        |                   |
+ *                        |  Physical Memory  |
+ *                        |      Window       |
+ *                        |                   |
+ *                        |                   |
+ *                        | ----------------- | KERNEL_ELF_TOP
+ *                        |     Kernel ELF    |
+ *                        | ----------------- | KERNEL_ELF_BASE (not fixed)
```
Indanz: It is compile time fixed though. Better say it's start of available memory/physBase.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -12,62 +12,112 @@
...
+ *                        |                   |
+ *                        |                   |
+ *            2^64 - 2^39 +-------------------+ PPTR_BASE
  *
+ *
+ *    This kernel mappings PD (code: `armKSGlobalKernelPDs[kernel_mappings_pud_idx]`)
```
Indanz: I'd say "The kernel mappings PD", otherwise it seems you're referring to the the PUD above.

Indanz:
Ready to be merged once the squash has happened.

It's better than it was, any further improvements (or fixes) we can do with a new PR.

midnightveil:
OK, squashed and removed the FIXME. This is ready to go then.

Indanz:
Please squash the second commit with the third one, both improve the docs. The second one introduces the TTBR1_EL2 error which is fixed in the third commit. (From the second commit's title I assumed it was changing text justification.)

lsf37:
Just wanted to say thanks for working on this, I've had an item to look at this on my todo list ever since we started the AArch64 verification.

midnightveil:
It looks like this PR is slightly wrong still (thanks @kent-mcleod), oops. Not sure how I didn't notice this... I swear I checked all the constants made sense.

Looks like for hyp the kernel mappings window is the second entry of the PGD, not the last one. So just some constants to be moved around and the diagrams for hyp to be adjusted.

https://github.com/seL4/seL4/issues/957

PR #1524, opened 2025-10-19
aarch32,boot: Fix patched entry code for aarch32

kent-mcleod:
Earlier patches add an additional dynamic reserved region defined by an earlier stage boot loader. But the asm for aarch32 was never updated.

Now the aarch32 head.S correctly passes through the bootloader arguments.

### DIFF
--- a/src/arch/arm/32/head.S
+++ b/src/arch/arm/32/head.S
@@ -78,10 +78,11 @@
 .section .boot.text, "ax"
 BEGIN_FUNC(_start)
     /*
-     * Get the dtb and dtb size from the elfloader stack. Do this first because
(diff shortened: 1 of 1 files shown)


Indanz (commented):
Code itself is correct. However, `init_kernel` has no extra device arguments, so this change achieves nothing.

midnightveil:
This is for the microkit branch, looks like Kent is adding arm32 support to microkit?

kent-mcleod:
> This is for the microkit branch, looks like Kent is adding arm32 support to microkit?

I'm not adding arm32 support to microkit, just fixing the kernel patch that microkit uses to also work on aarch32.

PR #1526, opened 2025-10-20
Prevent overmapping on aarch32, aarch64, x86

Kswin01:
This is continuing the work as outlined in PR #967. The original branch has been rebased, and merge issues resolved related to the aarch64 page tables refactor.

Test with: seL4/sel4test#148

### DIFF
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -1663,7 +1663,8 @@ createSafeMappingEntries_PTE
 #ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
         if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_large) {
 #else
-        if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_small) {
+        if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_small &&
+            pte_pte_small_ptr_get_contiguous_hint(ret.pte_entries.base)) {
 
 #endif
             if (!is_remap) {
@@ -2284,7 +2285,7 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         paddr_t capFBasePtr;
         cap_t pdCap;
         pde_t *pd;
-        asid_t asid, frame_asid;
+        asid_t asid;
         vm_rights_t capVMRights, vmRights;
         vm_page_size_t frameSize;
         vm_attributes_t attr;
@@ -2318,10 +2319,9 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         pd = PDE_PTR(cap_page_directory_cap_get_capPDBasePtr(
                          pdCap));
         asid = cap_page_directory_cap_get_capPDMappedASID(pdCap);
-        frame_asid = generic_frame_cap_get_capFMappedASID(cap);
 
         if (generic_frame_cap_get_capFIsMapped(cap)) {
-            if (frame_asid != asid) {
+            if (generic_frame_cap_get_capFMappedASID(cap) != asid) {
                 current_syscall_error.type = seL4_InvalidCapability;
                 current_syscall_error.invalidCapNumber = 1;
 


lsf37:
This still needs at least a TSC discussion if not an RFC. It breaks the API for systems that currently use overmapping.

Just because the RISC-V implementation follows the manual does not mean that the manual was correct. The ARM and x86 implementations both predate RISC-V and we have fixed so many misunderstandings in both the RISC-V implementation and in the manual over the years that it is absolutely open what the actual intention should be. Currently, it's just inconsistent and we should consciously decide what we want.

There is also still the question of whether anything should be done at all to the code. The manual can just state what the correct behaviour is for each architecture. There is no actual requirement for them to be the same even if consistency is generally preferable -- the question is who is going to come up with the funding for the verification of this change, it's too large to just do on the side.

Kswin01:
> This still needs at least a TSC discussion if not an RFC. It breaks the API for systems that currently use overmapping.

I'm in the process of writing an RFC up for this change, and will make a PR to the rfc repo for that soon. 

> I'm not convinced that we need is_remap

~~I agree, I removed that on aarch64. The `is_remap` is a remnant of the original PR that this is based on, and I'll remove this for other platforms.~~ Nevermind, `is_remap` seems to be necessary on other platforms due to the the structure of creating mappings (having different functions for different size pages). 

> The whole point of this is to prevent remapping, so when would is_remap be true and allowed?

My understanding of remap is that it essentially allows a permissions update, so if the supplied frame cap is mapped into the current address space, and doesn't exist in a different mapping within it, then we allow a remap of it (with potentially new permissions). This PR prevents overmapping, which is when we are attempting to over-write an existing mapping with an unmapped frame. So if re-map is true, then the check for over-mapping should be ignored (I believe that was the original intention behind Alwin's original PR).

src/arch/arm/32/kernel/vspace.c
```diff
@@ -1649,6 +1659,21 @@ createSafeMappingEntries_PTE
             }
         }
 
+        /* Check that we are not overwriting an existing mapping */
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+        if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_large) {
+#else
+        if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_small) {
```
Indanz: is this missing `&& pte_pte_small_ptr_get_contiguous_hint(ret.pte_entries.base)))`?

src/arch/arm/32/kernel/vspace.c
```diff
@@ -1607,6 +1607,16 @@ createSafeMappingEntries_PTE
             return ret;
         }
 
+        /* Check that we are not overwriting an existing mapping */
+        if (pte_ptr_get_pteType(ret.pte_entries.base) == pte_pte_small) {
+            if (!is_remap) {
+                userError("Virtual address (0x%"SEL4_PRIx_word") already mapped", vaddr);
+                current_syscall_error.type = seL4_DeleteFirst;
+                ret.status = EXCEPTION_SYSCALL_ERROR;
```
Indanz, line 1615: This is repeated so often, what about putting it in a separate function?

src/arch/arm/32/kernel/vspace.c
```diff
@@ -2271,9 +2318,10 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         pd = PDE_PTR(cap_page_directory_cap_get_capPDBasePtr(
                          pdCap));
         asid = cap_page_directory_cap_get_capPDMappedASID(pdCap);
+        frame_asid = generic_frame_cap_get_capFMappedASID(cap);
 
         if (generic_frame_cap_get_capFIsMapped(cap)) {
-            if (generic_frame_cap_get_capFMappedASID(cap) != asid) {
+            if (frame_asid != asid) {
```
Indanz: Perhaps nicer, but you're not re-using `frame_asid`, so this change is unnecessary.

src/arch/arm/32/kernel/vspace.c
```diff
@@ -2298,6 +2348,8 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
 
                 return EXCEPTION_SYSCALL_ERROR;
             }
+
+            is_remap = false;
```
Indanz, line 2352: You only need to set this once at the start of the `if (generic_frame_cap_get_capFIsMapped(cap))` block.

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1560,15 +1560,20 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
...
                 userError("ARMPageMap: Attempting to remap a frame that does not belong to the passed address space");
                 current_syscall_error.type = seL4_InvalidCapability;
                 current_syscall_error.invalidCapNumber = 1;
                 return EXCEPTION_SYSCALL_ERROR;
 
-            } else if (cap_frame_cap_get_capFMappedAddress(cap) != vaddr) {
+            }
+
+            /* The invoked frame cap is already mapped to a different address in this vspace  */
+            if (cap_frame_cap_get_capFMappedAddress(cap) != vaddr) {
```
Indanz, line 1576: No actual logical change here, just looks like it because the comment was wedged in.

src/arch/x86/64/kernel/vspace.c
```diff
@@ -1499,18 +1499,29 @@ static create_mapping_pdpte_return_t createSafeMappingEntries_PDPTE(paddr_t base
...
+
+
     ret.pdpte = makeUserPDPTEHugePage(base, attr, vmRights);
     ret.status = EXCEPTION_NONE;
     return ret;
 }
 
 exception_t decodeX86ModeMapPage(word_t label, vm_page_size_t page_size, cte_t *cte, cap_t cap,
-                                 vspace_root_t *vroot, vptr_t vaddr, paddr_t paddr, vm_rights_t vm_rights, vm_attributes_t vm_attr)
+                                 vspace_root_t *vroot, vptr_t vaddr, paddr_t paddr, vm_rights_t vm_rights, vm_attributes_t vm_attr, asid_t frame_asid)
```
Indanz: Isn't that supposed to be an `is_remap`? Where did `frame_asid` come from?

src/arch/x86/64/kernel/vspace.c
```diff
@@ -1476,7 +1476,7 @@ typedef struct create_mapping_pdpte_return create_mapping_pdpte_return_t;
 
 static create_mapping_pdpte_return_t createSafeMappingEntries_PDPTE(paddr_t base, word_t vaddr, vm_rights_t vmRights,
                                                                     vm_attributes_t attr,
-                                                                    vspace_root_t *vspace)
+                                                                    vspace_root_t *vspace, asid_t frame_asid)
```
Indanz: This seems to be actually `is_remap`, see other comment.

src/arch/x86/64/kernel/vspace.c
```diff
@@ -1499,18 +1499,29 @@ static create_mapping_pdpte_return_t createSafeMappingEntries_PDPTE(paddr_t base
         return ret;
     }
 
+    /* Check that we are not overwriting an existing mapping */
+    if (pdpte_ptr_get_page_size(ret.pdptSlot) == pdpte_pdpte_1g) {
+        if (pdpte_pdpte_1g_ptr_get_present(ret.pdptSlot) && frame_asid == asidInvalid) {
```
Indanz: Probably:
```suggestion
        if (pdpte_pdpte_1g_ptr_get_present(ret.pdptSlot) && !is_remap) {
```

Indanz (commented):
So really, the changes can be limited to just an extra check per mapping invocation that checks the info in the current page table entry for presence and current physical address. It's just unfortunate that on some archs this requires a check per page type.

src/arch/arm/32/kernel/vspace.c
Kswin01: Yep it seems to be that way, apologies I wasnt too familiar with aarch32 hypervisor support.

src/arch/arm/64/kernel/vspace.c
Kswin01: Would you prefer to keep this as before?

src/arch/arm/32/kernel/vspace.c
Kswin01: Explicitly setting to either true or false in these cases reads nicer to me

src/arch/arm/32/kernel/vspace.c
Kswin01: The only thing we'd be able to put in that function is the userError call and setting the current syscall error. The type of ret changes between pte and pde annoyingly

src/arch/arm/64/kernel/vspace.c
Indanz: No. It was just a comment for other reviewers so they know it's okay.

src/arch/arm/32/kernel/vspace.c
Indanz: Well, you can also move the if-check and let it return the status.

PR #1527, opened 2025-10-21
x86: pass the EFI system table pointer to root task

cl91:
Pass the EFI system table pointer supplied by the bootloader in the multiboot2 information structure to the root task, in the form of an extra bootinfo structure. The EFI system table remains (partially) valid after the boot loader has exited the EFI services, and contains pointers to firmware tables such as SMBIOS tables, which userspace tasks can then map using standard seL4 APIs.

### DIFF
--- a/libsel4/include/sel4/bootinfo_types.h
+++ b/libsel4/include/sel4/bootinfo_types.h
@@ -106,6 +106,7 @@ typedef enum {
     SEL4_BOOTINFO_HEADER_X86_FRAMEBUFFER    = 4,
     SEL4_BOOTINFO_HEADER_X86_TSC_FREQ       = 5, /* frequency is in MHz */
     SEL4_BOOTINFO_HEADER_FDT                = 6, /* device tree */
+    SEL4_BOOTINFO_HEADER_X86_EFI_SYSTBL_PTR = 7, /* EFI system table pointer on x86 */
     /* Add more IDs here, the two elements below must always be at the end. */
     SEL4_BOOTINFO_HEADER_NUM,
     SEL4_FORCE_LONG_ENUM(seL4_BootInfoID)
--- a/src/arch/x86/kernel/boot.c
+++ b/src/arch/x86/kernel/boot.c
@@ -94,7 +94,9 @@ BOOT_CODE bool_t init_sys_state(
     acpi_rsdp_t      *acpi_rsdp,
     seL4_X86_BootInfo_VBE *vbe,
     seL4_X86_BootInfo_mmap_t *mb_mmap,
-    seL4_X86_BootInfo_fb_t *fb_info
+    seL4_X86_BootInfo_fb_t *fb_info,
+    uint32_t      efi_systbl_ptr32,
+    uint64_t      efi_systbl_ptr64
 )
 {
     cap_t         root_cnode_cap;
@@ -131,6 +133,12 @@ BOOT_CODE bool_t init_sys_state(
     }
     if (fb_info && fb_info->addr) {
         extra_bi_size += sizeof(seL4_BootInfoHeader) + sizeof(*fb_info);
+    }
+    if (efi_systbl_ptr64) {
+        extra_bi_size += sizeof(seL4_BootInfoHeader) + sizeof(uint64_t);
+    }
+    if (efi_systbl_ptr32) {
+        extra_bi_size += sizeof(seL4_BootInfoHeader) + sizeof(uint32_t);
     }
 
     word_t mb_mmap_size = sizeof(seL4_X86_BootInfo_mmap_t);
@@ -224,6 +232,26 @@ BOOT_CODE bool_t init_sys_state(
         extra_bi_offset += sizeof(header);
         *(uint32_t *)(extra_bi_region.start + extra_bi_offset) = tsc_freq;
         extra_bi_offset += 4;
+    }
+
+    /* populate EFI system table pointer block */
+    if (efi_systbl_ptr64) {
+        seL4_BootInfoHeader header;
+        header.id = SEL4_BOOTINFO_HEADER_X86_EFI_SYSTBL_PTR;
+        header.len = sizeof(header) + sizeof(uint64_t);
+        *(seL4_BootInfoHeader *)(extra_bi_region.start + extra_bi_offset) = header;
+        extra_bi_offset += sizeof(header);
+        *(uint64_t *)(extra_bi_region.start + extra_bi_offset) = efi_systbl_ptr64;
+        extra_bi_offset += sizeof(uint64_t);
+    }
+    if (efi_systbl_ptr32) {
+        seL4_BootInfoHeader header;
+        header.id = SEL4_BOOTINFO_HEADER_X86_EFI_SYSTBL_PTR;
+        header.len = sizeof(header) + sizeof(uint32_t);
+        *(seL4_BootInfoHeader *)(extra_bi_region.start + extra_bi_offset) = header;
+        extra_bi_offset += sizeof(header);
+        *(uint32_t *)(extra_bi_region.start + extra_bi_offset) = efi_systbl_ptr32;
+        extra_bi_offset += sizeof(uint32_t);
     }
 
 #ifdef CONFIG_KERNEL_MCS
--- a/src/arch/x86/kernel/boot_sys.c
+++ b/src/arch/x86/kernel/boot_sys.c
@@ -179,7 +179,9 @@ static BOOT_CODE bool_t try_boot_sys_node(cpu_id_t cpu_id)
             &boot_state.acpi_rsdp,
             &boot_state.vbe_info,
             &boot_state.mb_mmap_info,
-            &boot_state.fb_info
+            &boot_state.fb_info,
+            boot_state.efi_ptr32,
+            boot_state.efi_ptr64
         )) {
         return false;
     }
@@ -629,6 +631,18 @@ static BOOT_CODE bool_t try_boot_sys_mbi2(
         if (tag->type == MULTIBOOT2_TAG_CMDLINE) {
             char const *const cmdline = (char const * const)(behind_tag);
             cmdline_parse(cmdline, &cmdline_opt);
+        } else if (tag->type == MULTIBOOT2_TAG_EFI32) {
+            if (tag->size - sizeof(*tag) == sizeof(uint32_t)) {
+                boot_state.efi_ptr32 = *(uint32_t *)behind_tag;
+                printf("32-bit EFI system table at physical address=0x%x\n",
+                       boot_state.efi_ptr32);
+            }
+        } else if (tag->type == MULTIBOOT2_TAG_EFI64) {
+            if (tag->size - sizeof(*tag) == sizeof(uint64_t)) {
+                boot_state.efi_ptr64 = *(uint64_t *)behind_tag;
+                printf("64-bit EFI system table at physical address=0x%llx\n",
+                       boot_state.efi_ptr64);
+            }
         } else if (tag->type == MULTIBOOT2_TAG_ACPI_1) {
             if (ACPI_V1_SIZE == tag->size - sizeof(*tag)) {
                 memcpy(&boot_state.acpi_rsdp, (void *)behind_tag, tag->size - sizeof(*tag));
--- a/include/arch/x86/arch/kernel/boot.h
+++ b/include/arch/x86/arch/kernel/boot.h
@@ -38,7 +38,9 @@ bool_t init_sys_state(
     acpi_rsdp_t      *acpi_rsdp,
     seL4_X86_BootInfo_VBE *vbe,
     seL4_X86_BootInfo_mmap_t *mb_mmap,
-    seL4_X86_BootInfo_fb_t *fb_info
+    seL4_X86_BootInfo_fb_t *fb_info,
+    uint32_t      efi_systbl_ptr32,
+    uint64_t      efi_systbl_ptr64
 );
 
 bool_t init_cpu(
--- a/include/arch/x86/arch/kernel/boot_sys.h
+++ b/include/arch/x86/arch/kernel/boot_sys.h
@@ -30,6 +30,8 @@ typedef struct boot_state {
     seL4_X86_BootInfo_VBE vbe_info; /* Potential VBE information from multiboot */
     seL4_X86_BootInfo_mmap_t mb_mmap_info; /* memory map information from multiboot */
     seL4_X86_BootInfo_fb_t fb_info; /* framebuffer information as set by bootloader */
+    uint32_t     efi_ptr32;   /* 32-bit EFI system table pointer */
+    uint64_t     efi_ptr64;   /* 64-bit EFI system table pointer */
 } boot_state_t;
 
 extern boot_state_t boot_state;
--- a/include/arch/x86/arch/kernel/multiboot2.h
+++ b/include/arch/x86/arch/kernel/multiboot2.h
@@ -48,6 +48,8 @@ enum multiboot2_tags {
     MULTIBOOT2_TAG_MODULE  = 3,
     MULTIBOOT2_TAG_MEMORY  = 6,
     MULTIBOOT2_TAG_FB      = 8,
+    MULTIBOOT2_TAG_EFI32   = 11, /* 32-bit EFI system table pointer */
+    MULTIBOOT2_TAG_EFI64   = 12, /* 64-bit EFI system table pointer */
     MULTIBOOT2_TAG_ACPI_1  = 14,
     MULTIBOOT2_TAG_ACPI_2  = 15,
 };


src/arch/x86/kernel/boot.c
```diff
@@ -226,6 +233,25 @@ BOOT_CODE bool_t init_sys_state(
...
+        extra_bi_offset += sizeof(uint64_t);
+    } else if (efi_systbl_ptr32) {
+        seL4_BootInfoHeader header;
+        header.id = SEL4_BOOTINFO_HEADER_EFI_SYSTBL_PTR;
+        header.len = sizeof(header) + sizeof(uint32_t);
+        *(seL4_BootInfoHeader *)(extra_bi_region.start + extra_bi_offset) = header;
+        extra_bi_offset += sizeof(header);
+        *(uint32_t *)(extra_bi_region.start + extra_bi_offset) = efi_systbl_ptr32;
+        extra_bi_offset += sizeof(uint32_t);
+    }
```
Indanz, line 255: Considering you're not passing both the 32 and 64 bits value, it seems much simpler if you always pass the value as 64-bits, even when it's 32-bits. That would simplify the code a fair bit and user space can check the top 32-bits to see whether it's a 32-bit or 64-bit value.

src/arch/x86/kernel/boot_sys.c
```diff
@@ -628,6 +630,18 @@ static BOOT_CODE bool_t try_boot_sys_mbi2(
         if (tag->type == MULTIBOOT2_TAG_CMDLINE) {
             char const *const cmdline = (char const * const)(behind_tag);
             cmdline_parse(cmdline, &cmdline_opt);
+        } else if (tag->type == MULTIBOOT2_TAG_EFI32) {
+            if (tag->size - sizeof(*tag) == sizeof(uint32_t)) {
+                boot_state.efi_ptr32 = *(uint32_t *)behind_tag;
+                printf("32-bit EFI system table at physical address=0x%x\n",
+                       boot_state.efi_ptr32);
+            }
```
Indanz, line 639: If you only keep one 64-bit value and if bootloaders are allowed to pass both, you probably need an extra check to see if the 64-bit value wasn't set yet before accidentally overwriting it with the 32-bit value here (assuming that's the behaviour you want).

Indanz (commented):
Seems a reasonable addition.

src/arch/x86/kernel/boot_sys.c
cl91: I'm not sure if the bootloader is allow to pass both, but my concern is sometimes 64-bit platforms can ship with 32-bit firmware (and 64-bit firmware can boot 32-bit OS, at least on x86), so I think the safest option is to pass the information from the bootloader verbatim, and let the user space task decide what to do with it. I've modified the commit to pass both if the bootloader has specified both (and fixed a bug where the extra bootinfo size is calculated incorrectly).

cl91: By the way I just tested booting 32-bit seL4 under a 64-bit firmware (bootloader is grub2). The system table pointer passed by grub2 is 64-bit. The EFI system table pointed to by the pointer is a 64-bit table (meaning that pointers and EFI_HANDLEs in the table are all 64-bit values). So we definitely need a way for the root task to know if the EFI system table pointer is 32 bit or 64 bit. I don't have a 64-bit machine that ships with 32-bit firmware, but I imagine the need in that case is similar.

Indanz: That just confirms that the bootloader only passes one value, in which case there is no point is passing both. Just always pass the value as 64-bits to user space, like I suggested above. User space can check whether the top 32-bits are zero to see if it's a 32-bit value or a 64-bit one.

cl91: > Just always pass the value as 64-bits to user space, like I suggested above. User space can check whether the top 32-bits are zero to see if it's a 32-bit value or a 64-bit one.

That's incorrect because the firmware can place the 64bit table below 4GB in physical memory (in fact this is often the case on x86, to make booting 32bit OS easier), making this case indistinguishable from a 32-bit table. The only way of reliably detecting the firmware bitness is from the pointer size passed in by the bootloader.

Indanz: Okay, you're saying that users need the tag type to distinguish between 32-bit and 64-bit tables. If that is the case, you should add separate SEL4_BOOTINFO_HEADER_EFI_SYSTBL32 and SEL4_BOOTINFO_HEADER_EFI_SYSTBL64 or something.

However, I doubt that there is such need, because if there is a practical difference, it will be discoverable in the tables themselves, like in the case SMBIOS. EFI is suppose to be cross platform after all. Using different tags only tells you whether GRUB was [compiled](https://elixir.bootlin.com/grub/grub-2.12/source/grub-core/loader/multiboot_mbi2.c#L885) in 32-bit mode or 64-bit mode, the passed content is the same. If there is a need to know whether the system booted in 32-bit or 64-bit mode (good luck finding any x86 EFI systems that boots in 32-in mode), I think there should be a better way than checking these tags.

GRUB has different tags because the size and format is different.

cl91: > Okay, you're saying that users need the tag type to distinguish between 32-bit and 64-bit tables. If that is the case, you should add separate SEL4_BOOTINFO_HEADER_EFI_SYSTBL32 and SEL4_BOOTINFO_HEADER_EFI_SYSTBL64 or something.

Well the seL4_BootInfoHeader::len (minus the header size) specifies the pointer size so there is no need to have two separate bootinfo type.

> However, I doubt that there is such need, because if there is a practical difference, it will be discoverable in the tables themselves, like in the case SMBIOS. EFI is suppose to be cross platform after all.

Funny enough, the EFI system table doesn't in fact tell you the bitness (at least not directly --- one can infer it from the table size and the firmware revision, but that's cumbersome (there are over ten UEFI revisions) and not very bug-proof or future-proof (the UEFI specs explicitly says the table size can change in the future)). I'd argue it's a design oversight of "U"EFI.

By the way (it's not entirely related to this discussion, just a fact) it's often the case a 64-bit UEFI firmware actually supplies a 32-bit SMBIOS entrypoint structure :sweat_smile: .

> Using different tags only tells you whether GRUB was [compiled](https://elixir.bootlin.com/grub/grub-2.12/source/grub-core/loader/multiboot_mbi2.c#L885) in 32-bit mode or 64-bit mode

The bitness in which GRUB is compiled is precisely the bitness of the UEFI firmware, because 32-bit UEFI cannot load 64-bit EFI executables and vise versa.

> If there is a need to know whether the system booted in 32-bit or 64-bit mode (good luck finding any x86 EFI systems that boots in 32-in mode), I think there should be a better way than checking these tags.

Well the main use case is booting a 32-bit kernel on a 64-bit firmware (or bootloader). There is no way of reliably knowing the bitness of the firmware without the EFI system pointer size passed by grub (you can only guess based on the table size and firmware revision, and pray the firmware developer actually bothered to populate these fields correctly).

Yes 32-bit EFI firmware is uncommon on consumer 64-bit systems but they do appear in embedded systems.

I'd argue since the principle of microkernels encourages delegating the policy and decision making to the user space, fully passing the information that grub has supplied is a better approach here.

Indanz: Thanks for the detailed answer.

> > Okay, you're saying that users need the tag type to distinguish between 32-bit and 64-bit tables. If that is the case, you should add separate SEL4_BOOTINFO_HEADER_EFI_SYSTBL32 and SEL4_BOOTINFO_HEADER_EFI_SYSTBL64 or something.
> 
> Well the seL4_BootInfoHeader::len (minus the header size) specifies the pointer size so there is no need to have two separate bootinfo type.

That is not the point, it are two different things, so they should have different names.

> By the way (it's not entirely related to this discussion, just a fact) it's often the case a 64-bit UEFI firmware actually supplies a 32-bit SMBIOS entrypoint structure 😅 .

UEFI firmware has the bitness in its type in the first few bytes of the image, so firmware at least is well-defined.

> > Using different tags only tells you whether GRUB was [compiled](https://elixir.bootlin.com/grub/grub-2.12/source/grub-core/loader/multiboot_mbi2.c#L885) in 32-bit mode or 64-bit mode
> 
> The bitness in which GRUB is compiled is precisely the bitness of the UEFI firmware, because 32-bit UEFI cannot load 64-bit EFI executables and vise versa.

This is unexpected. If EFI doesn't care about being compatible, why should seL4? That is, user space can know at compile time what system it runs on.

> > If there is a need to know whether the system booted in 32-bit or 64-bit mode (good luck finding any x86 EFI systems that boots in 32-in mode), I think there should be a better way than checking these tags.
> 
> Well the main use case is booting a 32-bit kernel on a 64-bit firmware (or bootloader). There is no way of reliably knowing the bitness of the firmware without the EFI system pointer size passed by grub (you can only guess based on the table size and firmware revision, and pray the firmware developer actually bothered to populate these fields correctly).
> 
> Yes 32-bit EFI firmware is uncommon on consumer 64-bit systems but they do appear in embedded systems.

But on embedded systems user space knows what system it is running on, there is no need for seL4 to tell it.
Edit: Also, why would you go for 32-bit code instead of 64-bit on embedded systems? 

> I'd argue since the principle of microkernels encourages delegating the policy and decision making to the user space, fully passing the information that grub has supplied is a better approach here.

Agreed, but up to a point: If there is no practical need, then we don't have to go out of our way to add extra code for whatifs.

cl91: > This is unexpected. If EFI doesn't care about being compatible, why should seL4? That is, user space can know at compile time what system it runs on.

Well a 32-bit x86 OS can run on both 32-bit and 64-bit x86 machines, so it cannot know the bitness of the platform or firmware at compile time. This is in fact a pretty common use case (think of running 32-bit Win10 or debian on x64 systems), especially on systems with less than 4GB of RAM. If anything, it's a useful test for software correctness, especially drivers.

Indanz: What about having one SEL4_BOOTINFO_HEADER_EFI_SYSTBL_PTR tag and always encode it 64-bits, but separately convey the information that it's a 64-bit machine running 32-bit OS?

cl91: > having one SEL4_BOOTINFO_HEADER_EFI_SYSTBL_PTR tag and separately convey the information that it's a 64-bit machine running 32-bit OS

Isn't this functionally exactly the same as my implementation (but more contrived)? The size of the bootinfo struct gives the bitness of the firmware (which always matches the bitness of the bootloader, hence the size of the EFI system table pointer passed by grub --- this is required by UEFI). The bootinfo struct is after all a variable-length structure where a common header encodes the actual length of the struct.

Indanz: Of course it's functionally the same. But instead of having to check a random size of a random field to determine what mode GRUB booted in, the info is told explicitly to the user instead. However, I suppose the only place people need to know is when accessing EFI tables, so having it here makes sense.

Indanz (approved):

Indanz:
(FYI, waiting on a decision whether this API change requires an RFC or is small enough to get merged as is. Another review by someone more familiar with the x86 code would also help.)

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -106,6 +106,7 @@ typedef enum {
     SEL4_BOOTINFO_HEADER_X86_FRAMEBUFFER    = 4,
     SEL4_BOOTINFO_HEADER_X86_TSC_FREQ       = 5, /* frequency is in MHz */
     SEL4_BOOTINFO_HEADER_FDT                = 6, /* device tree */
+    SEL4_BOOTINFO_HEADER_EFI_SYSTBL_PTR     = 7, /* EFI system table pointer */
```
kent-mcleod: This probably should have `X86` to signify that it belongs to the set of bootinfo fields generated by only the x86 boot code.

kent-mcleod:
By my interpretation of the RFC conditions, this shouldn't require an RFC. Extended bootinfo types is intended for extensions like this as the `seL4_BootInfoID` enum is intended to be sparse and open to extensions that add new information fields.

lsf37:
> By my interpretation of the RFC conditions, this shouldn't require an RFC. Extended bootinfo types is intended for extensions like this as the `seL4_BootInfoID` enum is intended to be sparse and open to extensions that add new information fields.

Alright, if Kent is happy with it, I'm fine with it as well. It's true that this is a data structure that is explicitly intended to be extensible.

libsel4/include/sel4/bootinfo_types.h
cl91: > This probably should have `X86` to signify that it belongs to the set of bootinfo fields generated by only the x86 boot code.

I though so too originally but later decided to go with only EFI_SYSTBL_PTR because other architectures are starting to ship UEFI firmware increasingly (especially ARM64 laptops, as UEFI is a requirement of shipping Windows 11), so although seL4 doesn't yet support UEFI on anything other than x86 at this point, potentially this tag can be used for other architectures eventually (similar to how the `FDT` tag is shared for both arm and riscv).

kent-mcleod: That would be committing the other architectures to the same semantics as what x86 has before there's any platforms/implementations for them. It's currently only implemented as an x86 bootinfo type, so the name should carry that information

cl91: Ok I have changed the tag name to include X86. I don't know If existing tag number can be changed or not. If it can, perhaps FDT should be moved after the X86_EFI_SYSTBL_PTR so all X86 bootinfo IDs stay together.

Indanz: No, that is part of the libsel4 API, you can't change it without breaking backward compatibility.

cl91: > No, that is part of the libsel4 API, you can't change it without breaking backward compatibility.

It would be a source-compatible change --- only breaking binary compatibility. The upcoming release appears to be a breaking release, so binary compatibility is already not possible.

cl91: > No, that is part of the libsel4 API, you can't change it without breaking backward compatibility.

It only breaks binary compatibility and is still a source-compatible change. If the next release is a source-compatible or a breaking release, changing this would not introduce any further incompatibility.

midnightveil:
I think this is ready to be merged, it just needs to be rebased @cl91?

midnightveil:
Thanks. I'll wait for proofs... unless anyone has objections.

Indanz (approved):

lsf37:
Let me have brief look

lsf37:
Technically, this does need an RFC, it's a user interface change. It looks straightforward and small enough to do without, but we should give other people in the @seL4/tsc a chance to object.

Please leave it open another week (say 4 September, because of the summit), after that I'm happy for it to be merged.

(Just using the protocol we've decided on, with pinging the TSC on those small kinda-need-RFC changes)

midnightveil:
Per previous messages, I thought it was agreed to skip it:

https://github.com/seL4/seL4/pull/1527#issuecomment-3518884028

(Or have we changed our mind?)

lsf37:
> Per previous messages, I thought it was agreed to skip it:
> 
> [#1527 (comment)](https://github.com/seL4/seL4/pull/1527#issuecomment-3518884028)
> 
> (Or have we changed our mind?)

No change of mind, from my side it's still fine, we just had agreed since then to ping the TSC on such changes where it's not entirely clear if there should be an RFC. Then if anyone thinks it does need further discussion we can do something about it.

(Many people on the TSC don't have the bandwidth to monitor all PRs, but if they get a ping, they can have a look.)

src/arch/x86/kernel/boot_sys.c
lsf37: I'm still not entirely convinced by this argument. Everywhere else in seL4 the tag fully determines the payload type. Here we are deviating from that. Is that really so necessary?

lsf37 (commented):
If not an RFC, this change should at least be documented in CHANGES.md. Do we also have boot info blocks in the manual? If yes, it should be documented there, too.

src/arch/x86/kernel/boot_sys.c
cl91: @lsf37 I guess a better way to explain this is that there are two pieces of information we need to communicate to the root task, ie the pointer itself and the bitness of the EFI tables pointed to by the pointer (ie. whether the information inside the EFI tables are 32-bit or 64-bit). The latter may be different from the bitness of the platform or the OS (eg. when running 32-bit seL4 on a 64-bit machine, or running 64-bit seL4 on a 32-bit UEFI firmware). The payload size communicates to the root task the bitness of the EFI tables, hence the need for the payload size to be varied.

cl91:
> If not an RFC, this change should at least be documented in CHANGES.md. Do we also have boot info blocks in the manual? If yes, it should be documented there, too.

@lsf37 The manual doesn't seem to have any documentation on other extra bootinfo blocks. Also, `SEL4_BOOTINFO_HEADER_PADDING` is still mentioned there, but my understanding is that it has been removed in the latest master (this was the rebase conflict I needed to resolve).

lsf37:
> > If not an RFC, this change should at least be documented in CHANGES.md. Do we also have boot info blocks in the manual? If yes, it should be documented there, too.
> 
> @lsf37 The manual doesn't seem to have any documentation on other extra bootinfo blocks. Also, `SEL4_BOOTINFO_HEADER_PADDING` is still mentioned there, but my understanding is that it has been removed in the latest master (this was the rebase conflict I needed to resolve).

Oh, good point, we should definitely check the PADDING entry in the manual (it's not removed as a concept, just not currently emitted anywhere, but I don't remember if we checked the manual properly). If the other ones are not in the manual, then docs in CHANGES.md are enough for now.

I guess we should add an issue to properly document all the potential extra boot info items in the manual. There's not much point in providing info from the kernel if we don't tell people what it is and how to use it.

midnightveil:
Is it worth defining an seL4_BootInfo_EfiTablePtr structure? I know for other value-sized things we don't, but it does make it a bit more annoying - you have to manually skip the header to get to the value. Whereas if you have the value struct you have the header in it.

cl91:
> Is it worth defining an seL4_BootInfo_EfiTablePtr structure? I know for other value-sized things we don't, but it does make it a bit more annoying - you have to manually skip the header to get to the value. Whereas if you have the value struct you have the header in it.

I had originally considered this but later decided against it because it would mask the fact that one needs to explicitly check the length field in the header to determine whether the firmware gave us a 32-bit EFI table or a 64-bit EFI table. It is not always the case that the bitness of the EFI tables matches the bitness of the OS.

PR #1528, opened 2025-10-22
Tensor-g3

M1llpex:


Indanz:
So just out of curiosity, can you confirm that that code is AI generated?

In that case, there's some copyrights missing.

Issue #1529, opened 2025-10-22
Only one of `ARCH_{ARM,RISCV,ARCH_X86}` ever appears in `gen_config.json`


nspin:
Most kernel config options appear in `gen_config.json` whether or not they are enabled. They are associated with the value `true` if enabled and as `false` otherwise. `KernelSel4Arch`, for example, behaves this way.

`KernelArch` on the other hand, does not behave this way. For example, when configured for AArch64, `ARCH_ARM` is present in `gen_config.json` with the value `true`, but `ARCH_RISCV` and `ARCH_X86` are absent, instead of present with the value `false`.

Unlike C `#ifdefs` where the use of an undefined macro is allowed, the `#[sel4_cfg]` macro in Rust requires that the config keys exist. That means that the following will fail to compile on AArch64:

```rust
#[sel4_cfg(ARCH_RISCV)]
fn foo() {}
```

We are using the following verbose workaround:

```rust
#[sel4_cfg(any(ARCH_RISCV32, ARCH_RISCV64)]
fn foo() {}
```

The issue seems to be related to the use of choice conditions in the definition of the `KernelArch` option:

https://github.com/seL4/seL4/blob/f9eb65c9a586ccacb5d2583c1cf1882212bb592a/configs/seL4Config.cmake#L69-L76

Indanz:
Considering this is very Rust specific, can't you solve this when parsing gen_config.json?

That said, don't you want to check for `ARCH == riscv` instead of checking the arch-specific defines?

But I'd be happy to merge any PR you have that fixes this.

nspin:
> That said, don't you want to check for ARCH == riscv instead of checking the arch-specific defines?

I prefer `#sel4_cfg[ARCH_RISCV]` over `#sel4_cfg[ARCH = "riscv"]` because the former benefits more from compile-time checking. For example, the following mistake would compile: `#sel4_cfg[ARCH = "riscu"]`.

nspin:
I found a fix in `helpers.cmake`: https://github.com/seL4/seL4/pull/1531

PR #1531, opened 2025-10-23
cmake: record all disabled options

nspin:
Fixes https://github.com/seL4/seL4/issues/1529#issuecomment-3431655078

Before this change, options that were hidden from the `cmake-gui` due to unsatisfied `config_choice` conditions were not recorded in `gen_config.{yaml,json,h}`. After this change, these hidden options are recorded as disabled.

### DIFF
--- a/tools/helpers.cmake
+++ b/tools/helpers.cmake
@@ -31,7 +31,10 @@ function(get_absolute_source_or_binary output input)
     if(NOT EXISTS "${test}")
         get_filename_component(test "${input}" ABSOLUTE BASE_DIR "${CMAKE_CURRENT_BINARY_DIR}")
     endif()
-    set("${output}" "${test}" PARENT_SCOPE)
+    set("${output}"
+        "${test}"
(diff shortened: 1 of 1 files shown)


nspin:
It appears that CI's formatting expectations for CMake files have changed since the last time `helpers.cmake` was modified. That check is complaining about parts of the file unrelated to my change.

Indanz:
> It appears that CI's formatting expectations for CMake files have changed since the last time `helpers.cmake` was modified. That check is complaining about parts of the file unrelated to my change.

That's why I hate style checkers so much, especially the CMake one. And the suggestions are all over the place, no consistency at all.

Anyway, I can't merge this PR because of this, need to wait for @lsf37.

lsf37:
We should add a style fix commit on this file before we merge, otherwise we're going to have this every time.

(That is also the official policy: update style in the files we touch)

lsf37 (approved):

PR #1532, opened 2025-10-24
Add SBI cap

Ivan-Velickovic:
Implements RFC-22 [1].

This implementation was done with Gerwin Klein.

This has been tested 32-bit and 64-bit RISC-V platforms.

For 64-bit, the HiFive P550 and QEMU virt were tested.
For 32-bit, the only supported platform is QEMU virt so I have
not been able to test this implementation on 32-bit hardware.

[1]: https://github.com/seL4/rfcs/pull/35

### DIFF
--- a/include/arch/riscv/arch/32/mode/object/structures.bf
+++ b/include/arch/riscv/arch/32/mode/object/structures.bf
@@ -58,6 +58,16 @@ block asid_pool_cap {
     field       capType         4
 }
 
+#ifdef CONFIG_ALLOW_SBI_CALLS
+block sbi_cap {
+    field capSBIFIDBadge 30
+    field capSBIEIDBadged 1
+    field capSBIFIDBadged 1
+    field capSBIEIDBadge 28
+    field capType 4
+}
+#endif
+
 -- NB: odd numbers are arch caps (see isArchCap())
 tagged_union cap capType {
     mask 4 0xe
@@ -75,6 +85,9 @@ tagged_union cap capType {
     -- 4-bit tag arch caps
     tag frame_cap           1
     tag page_table_cap      3
+#ifdef CONFIG_ALLOW_SBI_CALLS
+    tag sbi_cap             5
+#endif
     tag asid_control_cap    11
     tag asid_pool_cap       13
 
--- a/include/arch/riscv/arch/64/mode/object/structures.bf
+++ b/include/arch/riscv/arch/64/mode/object/structures.bf
@@ -64,6 +64,17 @@ block asid_pool_cap {
     field_high  capASIDPool     37
 }
 
+#ifdef CONFIG_ALLOW_SBI_CALLS
+block sbi_cap {
+    field capSBIFIDBadge 64
+    field capType 5
+    field capSBIEIDBadge 28
+    field capSBIEIDBadged 1
+    field capSBIFIDBadged 1
+    padding 29
+}
+#endif
+
 -- NB: odd numbers are arch caps (see isArchCap())
 tagged_union cap capType {
     -- 5-bit tag caps
@@ -88,6 +99,9 @@ tagged_union cap capType {
     tag page_table_cap      3
     tag asid_control_cap    11
     tag asid_pool_cap       13
+#ifdef CONFIG_ALLOW_SBI_CALLS
+    tag sbi_cap             15
+#endif
 }
 
 ---- Arch-independent object types
--- a/include/arch/riscv/arch/object/objecttype.h
+++ b/include/arch/riscv/arch/object/objecttype.h
@@ -43,5 +43,42 @@ static inline CONST bool_t Arch_isIRQControlDescendant(cap_t cap)
  */
 static inline CONST bool_t Arch_isMDBParentOf(cap_t cap_a, cap_t cap_b, bool_t firstBadged)
 {
-    return true;
+    switch (cap_get_capType(cap_a)) {
+#ifdef CONFIG_ALLOW_SBI_CALLS
+    case cap_sbi_cap: {
+        if (!cap_sbi_cap_get_capSBIEIDBadged(cap_a)) {
+            /* If there is no EID badge then the cap cannot be badged. */
+            return true;
+        }
+
+        if (!cap_sbi_cap_get_capSBIEIDBadged(cap_b)) {
+            return false;
+        }
+
+        word_t eid_a_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_a);
+        word_t eid_b_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_b);
+
+        if (eid_a_badge != eid_b_badge) {
+            return false;
+        }
+
+        if (!cap_sbi_cap_get_capSBIFIDBadged(cap_a)) {
+            return !firstBadged;
+        }
+
+        if (!cap_sbi_cap_get_capSBIFIDBadged(cap_b)) {
+            return false;
+        }
+
+        word_t fid_a_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_a);
+        word_t fid_b_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_b);
+        return (fid_a_badge == fid_b_badge && !firstBadged);
+        break;
+    }
+#endif
+
+    default:
+        return true;
+        break;
+    }
 }
--- a/include/arch/riscv/arch/object/structures.h
+++ b/include/arch/riscv/arch/object/structures.h
@@ -145,6 +145,23 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
 
 static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 {
+#ifdef CONFIG_ALLOW_SBI_CALLS
+    switch (cap_get_capType(derivedCap)) {
+    case cap_sbi_cap:
+        if (!cap_sbi_cap_get_capSBIEIDBadged(derivedCap) || !cap_sbi_cap_get_capSBIFIDBadged(derivedCap)) {
+            return true;
+        }
+
+        return (cap_sbi_cap_get_capSBIFIDBadge(derivedCap) !=
+            cap_sbi_cap_get_capSBIFIDBadge(srcCap) ||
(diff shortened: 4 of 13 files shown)


src/arch/riscv/config.cmake
```diff
@@ -50,6 +50,12 @@ config_option(
     DEPENDS "KernelArchRiscV"
 )
 
+config_option(
+    KernelAllowSBICalls ALLOW_SBI_CALLS "Allow user-space to make SBI calls. TODO"
```
Ivan-Velickovic, line 54: Noting TODO to fix.

src/arch/riscv/object/sbi.c
```diff
@@ -0,0 +1,107 @@
+#include <config.h>
+
+#ifdef CONFIG_ALLOW_SBI_CALLS
+
+#include <arch/object/sbi.h>
+
+// TODO: duplicated because I don't know the right inlcude for libsel4 types
+// TODO: maybe just make this an internal type
```
Ivan-Velickovic, line 8: Noting TODOs to fix.

include/arch/riscv/arch/32/mode/object/structures.bf
```diff
@@ -58,6 +58,16 @@ block asid_pool_cap {
...
 }
 
+#ifdef CONFIG_ALLOW_SBI_CALLS
+block sbi_cap {
+    field capSBIFIDBadge 30
+    field capSBIEIDBadged 1
+    field capSBIFIDBadged 1
+    field capSBIEIDBadge 28
+    field capType 4
+}
```
Indanz, line 68: Consider using underscores? It's very hard to distinguish the fields currently.

As mentioned in the RFC, we need at least 31 bits for EID. FID all seem to be sequential starting from 0, so we don't actually need that many bits to encode them.

include/arch/riscv/arch/64/mode/object/structures.bf
```diff
@@ -64,6 +64,17 @@ block asid_pool_cap {
...
 
+#ifdef CONFIG_ALLOW_SBI_CALLS
+block sbi_cap {
+    field capSBIFIDBadge 64
+    field capType 5
+    field capSBIEIDBadge 28
+    field capSBIEIDBadged 1
+    field capSBIFIDBadged 1
+    padding 29
+}
```
Indanz, line 75: Same comments apply here.

include/arch/riscv/arch/object/objecttype.h
```diff
@@ -43,5 +43,42 @@ static inline CONST bool_t Arch_isIRQControlDescendant(cap_t cap)
  */
 static inline CONST bool_t Arch_isMDBParentOf(cap_t cap_a, cap_t cap_b, bool_t firstBadged)
 {
-    return true;
+    switch (cap_get_capType(cap_a)) {
+#ifdef CONFIG_ALLOW_SBI_CALLS
+    case cap_sbi_cap: {
+        if (!cap_sbi_cap_get_capSBIEIDBadged(cap_a)) {
+            /* If there is no EID badge then the cap cannot be badged. */
+            return true;
```
Indanz, line 51: This comment is very misleading, it makes it sound like false should be returned.

I guess what was meant is that if cap A has no badge, cap B must be either a copy or a badged version of A.

include/arch/riscv/arch/object/structures.h
```diff
@@ -145,6 +145,23 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
...
+    case cap_sbi_cap:
+        if (!cap_sbi_cap_get_capSBIEIDBadged(derivedCap) || !cap_sbi_cap_get_capSBIFIDBadged(derivedCap)) {
+            return true;
+        }
+
+        return (cap_sbi_cap_get_capSBIFIDBadge(derivedCap) !=
+            cap_sbi_cap_get_capSBIFIDBadge(srcCap) ||
+            cap_sbi_cap_get_capSBIEIDBadge(derivedCap) !=
+            cap_sbi_cap_get_capSBIEIDBadge(srcCap)
+            );
```
Indanz, line 159: What about revoking a derivedCap with EID+FID set via a srcCap that has only EID set?

Was the earlier `cap_sbi_cap_get_capSBIFIDBadged` check supposed to be on srcCap instead?

include/arch/riscv/arch/object/structures.h
```diff
@@ -145,6 +145,23 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
...
+
+    default:
+        return false;
+    }
+#endif
     return false;
```
Indanz, line 165: Nicer to do it the Arm way and always have the switch, no final return and the ifdef per case.

libsel4/arch_include/riscv/interfaces/object-api-arch.xml
```diff
@@ -367,5 +384,24 @@
         </method>
 
     </interface>
+        <interface name="seL4_RISCV_SBI" manual_name="SBI" cap_description="Capability to allow threads to make SBI (Supervisor Binary Interface) calls.">
```
Indanz, line 387:
```suggestion
        <interface name="seL4_RISCV_SBI" manual_name="SBI" cap_description="Capability to allow threads to make Supervisor Binary Interface calls.">
```

libsel4/arch_include/riscv/sel4/arch/types.h
```diff
@@ -60,6 +61,22 @@ typedef struct seL4_UserContext_ {
...
+    seL4_Word a4;
+    seL4_Word a5;
+    seL4_Word a6;
+    seL4_Word a7;
+} seL4_RISCV_SBIContext;
+
+typedef struct seL4_RISCV_SBIRet_ {
+    seL4_Word error;
+    seL4_Word value;
+} seL4_RISCV_SBIRet;
```
Indanz, line 78: Just a suggestion, but wouldn't it make more sense to use the same construction as `seL4_ARM_Page_GetAddress` and combine seL4's errors with SBI errors?

SBI errors are negative, seL4's errors are positive numbers, both use 0 for no error. Or alternatively, use a specific seL4 error code for SBI errors and return the SBI error in `value`.

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -28,7 +28,8 @@ enum seL4_RootCNodeCapSlots {
     seL4_CapSMMUCBControl       = 13, /* global SMMU CB controller cap, null cap if not supported */
     seL4_CapInitThreadSC        = 14, /* initial thread's scheduling context cap, null cap if not supported */
     seL4_CapSMC                 = 15, /* global SMC cap, null cap if not supported */
-    seL4_NumInitialCaps         = 16
+    seL4_CapSBI                 = 16, /* global SBI cap, null cap if not supported */
+    seL4_NumInitialCaps         = 17
```
Indanz, line 32:
```suggestion
    seL4_CapSBI                 = seL4_CapSMC, /* global SBI cap, null cap if not supported */
    seL4_NumInitialCaps         = 16
```
Because it's the same thing with a different name.

src/arch/riscv/object/objecttype.c
```diff
@@ -52,6 +59,25 @@ deriveCap_ret_t Arch_deriveCap(cte_t *slot, cap_t cap)
...
+            if (cap_sbi_cap_get_capSBIEIDBadge(cap)) {
+                if (cap_sbi_cap_get_capSBIFIDBadge(cap)) {
+                    return cap_null_cap_new();
+                } else {
+                    cap_t badged_cap = cap_sbi_cap_set_capSBIFIDBadge(cap, data);
+                    return cap_sbi_cap_set_capSBIFIDBadged(badged_cap, 1);
+                }
+            } else {
+                cap_t badged_cap = cap_sbi_cap_set_capSBIEIDBadge(cap, data);
+                return cap_sbi_cap_set_capSBIEIDBadged(badged_cap, 1);
```
Indanz, line 76: What about invalid badge values because they wouldn't fit? I suppose silently ignoring them makes sense, but in debug builds it probably triggers an assert. Is that what we want though? I guess the same happens for e.g. CNode guards.

src/arch/riscv/object/sbi.c
```diff
@@ -0,0 +1,107 @@
...
+    ret.value = a1;
+
+    return ret;
+}
+
+static inline exception_t invokeSBICall(word_t *buffer, bool_t call)
+{
+    word_t i;
+    seL4_Word args[NUM_SBI_REGS];
+    word_t *ipcBuffer;
```
Indanz, line 43: Missing newline.

Indanz:
First simplification that can be done: Considering legacy extensions are not supported, we can get rid of `capSBIEIDBadged` and check badge value of zero instead. Improvement is limited because we still need to check `capSBIFIDBadge`, so not entirely sure this will improve things.

include/arch/riscv/arch/object/objecttype.h
```diff
@@ -43,5 +43,42 @@ static inline CONST bool_t Arch_isIRQControlDescendant(cap_t cap)
...
+            return !firstBadged;
+        }
+
+        if (!cap_sbi_cap_get_capSBIFIDBadged(cap_b)) {
+            return false;
+        }
+
+        word_t fid_a_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_a);
+        word_t fid_b_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_b);
+        return (fid_a_badge == fid_b_badge && !firstBadged);
```
Indanz, line 75: I have the feeling that all these checks can be replaced with just `return !firstBadged;`.

Because `firstBadged` is used we already assume we follow the derivation list in-order and are not comparing random caps, and it's used to catch the much harder case of multiple badges that happen to have the same value. The same check can be used to catch all cases with mismatching badge values, so there is no need to check for that explicitly.

include/arch/riscv/arch/object/structures.h
Indanz: Definitely looks like there is a `!cap_sbi_cap_get_capSBIFIDBadged(srcCap)` check missing somewhere though, because the code here assumes it is set.

Ivan-Velickovic:
Closing as an alternative implementation has been accepted for the RFC.

PR #1534, opened 2025-10-27
sel4.h: add missing vmenter.h include

Ivan-Velickovic:
I may be missing something but my understanding is that sel4.h should include everything from libsel4 so that users only have to include sel4.h.

When I was working on an x86-64 VMM I ran into this header missing.

### DIFF
--- a/libsel4/include/sel4/sel4.h
+++ b/libsel4/include/sel4/sel4.h
@@ -27,4 +27,7 @@
 #include <sel4/arch/constants.h>
 #include <sel4/plat/api/constants.h>
 
+#ifdef CONFIG_VTX
+#include <sel4/arch/vmenter.h>
+#endif
 


libsel4/include/sel4/sel4.h
```diff
@@ -27,4 +27,7 @@
 #include <sel4/arch/constants.h>
 #include <sel4/plat/api/constants.h>
 
+#ifdef CONFIG_VTX
```
Ivan-Velickovic, line 30: Actually I realise there’s probably a better place for this rather than here

Ivan-Velickovic: I think arch-specific `constants.h` is probably the best place? This whole header doesn't really seem to conform to the same style as all the other constants though, but I think that's a different problem.

Indanz:
> I may be missing something but my understanding is that sel4.h should include everything from libsel4 so that users only have to include sel4.h.

Should it though? What about mapping.h? That's not included by default as far as  I can tell either.

The enum is called `kernel_fault_context_t` while it only applies to VMENTER return codes. That on its own is enough reason not to include it by default in my opinion.

libsel4/include/sel4/sel4.h
Indanz: The enum would belong in libsel4/arch_include/x86/sel4/arch/types.h, while the define would be in seL4/libsel4/arch_include/x86/sel4/arch/constants.h.

Ivan-Velickovic:
> Should it though? What about mapping.h? That's not included by default as far as I can tell either.

I don't know what the policy is then, I'm used to libsel4 having enums defined describing the state of the IPC buffer in the case of fault handling (e.g what values at what message register) since where the kernel places data related to the fault affects user-space as well.

On ARM there's the `seL4_Fault_*` labels and then and enum for each kind of fault that can occur, so I imagined that `vmenter.h` is used for the same purpose (for example, this function: https://github.com/seL4/seL4_projects_libs/blob/0e06bebf4a3f0c21317c6877e0079f637e021562/libsel4vm/src/arch/x86/vm.c#L144).

The seL4 manual (section 10.2.5.1, VM Enter) explicitly mentions these defines.

Ivan-Velickovic:
So if the manual mentions these either this header should be included by default or the manual is wrong.

Indanz:
I have no preference either way, leaving it as it is, or adding the includes.

Indanz:
The header file needs to be cleaned up first before it can be unconditionally included, but that will break the API. So I think the only way forwards is to move and add the cleaned up definitions to other files and keep the unclean existing stuff in vmenter.h with a deprecated warning.

Indanz:
See last comment how to move this forwards. In the meantime, user space can do:

```
#ifdef CONFIG_VTX
#include <sel4/arch/vmenter.h>
#endif
```

PR #1537, opened 2025-10-30
am335/ompa3/bcm2837: compare irq & maxIRQ with <=

midnightveil:
Noticed this by accident.

I'm guessing these aren't covered by proof?

Otherwise it looks like maskInterrupt wouldn't do anything for the "maxIRQ" IRQ.

### DIFF
--- a/include/arch/arm/arch/machine/gic_v3.h
+++ b/include/arch/arm/arch/machine/gic_v3.h
@@ -23,6 +23,11 @@
 
 #include "gic_common.h"
 
+/**
+ * Per §4.8 of ARM GICv3/v4 Spec (IHI0069 Hb):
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Great stuff. These are indeed behind the machine interface for the proofs, so we don't see them. The hw tests passed for the platforms that changed, so this is good to merge from my side.

Have you reviewed other uses of `maxIRQ` for these platforms? If no, we should do that, in case this was not a typo that was copied/pasted, but an actual misunderstanding of what `maxIRQ` means.

midnightveil:
Yep, it's just these 2. The rest of the uses use it correctly.

lsf37:
Excellent, thanks for checking.

Issue #1536, opened 2025-10-30
kernel interrupt handling: (benign) race between creating the irq handler cap and binding it to a notification


midnightveil:
This came up from some discussions with multikernel, specifically SGIs (when we serialise the kernel boots).

> I think it's actually a design issue in the kernel that there's a race between creating the IRQ handler and binding it to a notification
> The bind invocation should return a flag telling you if an IRQ was delivered already
> — @kent-mcleod 

invokeIRQControl looks like this:

```
exception_t invokeIRQControl(irq_t irq, cte_t *handlerSlot, cte_t *controlSlot)
{
    setIRQState(IRQSignal, irq);
    cteInsert(cap_irq_handler_cap_new(IRQT_TO_IDX(irq)), controlSlot, handlerSlot);

    return EXCEPTION_NONE;
}
```

`setIRQState` then calls `maskInterrupt(false, ...)` which causes our interrupt handling code to enable this particular interrupt.

This in itself is fine, however, if we get an interrupt between the time we have created the IRQ Handler cap and bound it, we get the following print:

```
    case IRQSignal: {
        /* Merging the variable declaration and initialization into one line
         * requires an update in the proofs first. Might be a c89 legacy.
         */
        cap_t cap;
        cap = intStateIRQNode[IRQT_TO_IDX(irq)].cap;
        if (cap_get_capType(cap) == cap_notification_cap &&
            cap_notification_cap_get_capNtfnCanSend(cap)) {
            sendSignal(NTFN_PTR(cap_notification_cap_get_capNtfnPtr(cap)),
                       cap_notification_cap_get_capNtfnBadge(cap));
        } else {
#ifdef CONFIG_IRQ_REPORTING
            printf("Undelivered IRQ: %d\n", (int)IRQT_TO_IRQ(irq));
#endif
        }
```

If this happens, we need to ACK that IRQ, but the kernel doesn't expose this information. This means we need to unconditionally ACK the IRQ when a driver has started to re-enable it.

This is somewhat of a footgun that we've run into occasionally (mostly, our CI breaking). We've got a note in our developer docs about it: https://github.com/au-ts/sddf/blob/0911c90aa7c39a26de52f566b422cc530605bd4e/docs/developing.md#code, also see PRs: https://github.com/au-ts/sddf/pull/461

Either:

- When binding a notification to an IRQ, making that notification active if there was an active IRQ, so that the IRQ was not lost 
[...]

PR #1539, opened 2025-10-30
Enable IRQ when binding to a notification

Indanz:
And disable when unbinding.

This solves issue #1536.


Indanz:
Seems very straightforward and overall a simplification which is more user friendly in a backward compatible way.

Except if I'm missing something, the only downside is that it causes some minor verification work.

midnightveil (approved):
Oh, that's nice. It also seems to have removed some complexity with the delete/deletingIRQHandler too, which is great. I love it when this sort of improvement makes the code nicer.

Indanz:
> Oh, that's nice. It also seems to have removed some complexity with the delete/deletingIRQHandler too, which is great. I love it when this sort of improvement makes the code nicer.

I like that too, but when the obvious solution is better and simpler than what's actually implemented, it makes me highly suspicious that there is some subtle or obscure reason for the current implementation to be as it is.

lsf37:
Hm, yes, this is suspiciously simple. I'll see if I can find something in the earlier commit history that suggests why we did not do this in the first place.

lsf37:
*Insert rant about useless commit messages here.*

I can find a set of commits in C/Haskell/proofs that introduce the split into deleting/deleted from 2008, but I can't find anything as to why we did this (I can find that Tom did this on the proof side, and Philip on Haskell/C). We do have an invariant that if an IRQ handler cap exists, the corresponding IRQ must be enabled. Maybe that is related, because that is broken by the commit here.

src/object/interrupt.c
```diff
@@ -178,12 +179,14 @@ void invokeIRQHandler_ClearIRQHandler(irq_t irq)
 {
     cte_t *irqSlot;
 
+    setIRQState(IRQInactive, irq);
+
     irqSlot = intStateIRQNode + IRQT_TO_IDX(irq);
     /** GHOSTUPD: "(True, gs_set_assn cteDeleteOne_'proc (-1))" */
     cteDeleteOne(irqSlot);
```
lsf37: This breaks the invariant that when an IRQHandlerCap exists in an IRQ slot, the corresponding IRQ must be enabled. `cteDelete` and its variations, including the simpler `cteDeleteOne` are not functions where it would be easy to deal with temporarily broken invariants. If it is acceptable to do the `setIRQState` after `cteDeleteOne`, we might be Ok.

src/object/interrupt.c
```diff
@@ -192,11 +195,6 @@ void deletingIRQHandler(irq_t irq)
     cteDeleteOne(slot);
 }
```
lsf37, line 199: This is where it breaks. If someone revokes the IRQHandler cap, or it is being deleted for some other reason, the IRQ is now not set to Inactive.

I'm still not 100% sure why are doing this, but we are avoiding setting any machine state inside finaliseCap. I.e. the post-cap-deletion split is not particular to IRQ handler caps, it is also used for IOPorts, for instance.

src/object/interrupt.c
midnightveil: That invariant being broken is somewhat the point here, it would need to become "IRQ enabled whenever irq handler and bound notification".

src/object/interrupt.c
midnightveil: Annoying. This feels like the same as the mapping page tables issue where life would be a lot easier if we could enforce some kind of cleanup ordering.

---

The other alternative here would be that we add some state to the irq handler cap (we have the space, it's 1 bit) which stores if we got an undelivered IRQ, and then when binding we set the notification to the active state. Would that be easier from the proofs side?

src/object/interrupt.c
lsf37: Actually, I read this wrong. The invariant is going to be broken in any case if we switch this behaviour to whether a notification is registered or not, so the ordering here doesn't matter for that.

The ordering will matter, though if we switch that invariant from talking about the existence of IRQHandler caps, to the presence of notification caps in the the IRQ slots. So far that the comment is still relevant.

But with that comes a whole host of problems -- when you are deleting any notification cap now (inside `finaliseCap`), you now need to know where that cap lives and whether it might be sitting in the IRQ CNode. If yes, you need to do the IRQ disabling thing (which probably does need the split for the proofs, so we'd need to figure out something for notification caps there). This introduces a new dependency of the behaviour of caps on their location, which does not sound like a good idea to me. E.g. the whole long-running recursive delete procedure shuffles caps around. It won't shuffle anything into or out of the IRQ CNode, but I really would not want to need to prove that.

src/object/interrupt.c
lsf37: That doesn't work unfortunately -- you can't find the IRQHandler cap when an IRQ comes in. The IRQHandler cap is somewhere in someone's CSpace.

We could add a flag to the slot in the IRQ CNode, but that would make the IRQ CNode no longer a CNode, which breaks a lot of other things. We could keep a bitfield around, of course.

midnightveil: Ah. We've already got a table with the `intStateIRQTable`. There's a comment somewhere about how seL4 ignores the ARM "active + pending" state and reduces it to just "active/inactive".

We could probably store the state there without too much issue with a bitmask since there's enough bits, though probably from a proofs side it makes sense for it to be a whole another table (I'm guessing? otherwise we need to adjust some functions).

lsf37: I think we could deal with using free bits of the `intStateIRQTable` in the proofs. In the specs we'd make them separate, but encoding two tables into one with distinct bits should not be a problem.

lsf37: (famous last words :-))

midnightveil: From the perspective of easing the proof burden (since I doubt you're being paid for this) and a (extremely small, like maxIrq/64 words) memory overhead, would it beasier to just use 2 separate tables and make an issue for combining them? Or should I just try doing a combined one and seeing how it goes?

From a software side it's somewhat easy either way...

lsf37: I'd go with a combined one. If I'm spending free time on it, I'd rather only do it once.

src/object/interrupt.c
Indanz: I don't think it's possible for the final cap to be in the IRQ CNode, that's always a derived cap, never the final, I think. Does that help?

src/object/interrupt.c
Indanz: > This is where it breaks. If someone revokes the IRQHandler cap, or it is being deleted for some other reason, the IRQ is now not set to Inactive.

Ah, yes. I missed that we don't explicitly unbind the notification from the IRQHandle. So the old code has to stay to catch this corner case. But it doesn't mean we can't do the other improvements though.

An alternative is to do the cleanup when removing the notification cap (the one in `intStateIRQNode`), but that would require an extra bit in the notification cap to tell it that it was inserted in the IRQ slot, otherwise it can't know. Not sure that's simpler than what we had, but seems simpler than what you two are talking about.

Edit: Never mind, the notification doesn't know what IRQ it is bound to, so we'd need to hack the badge as IRQ number too to have all information, which gets altogether too ugly. Edit2: Never mind, we need the badge to know which bits to set.

What we had is already proven though, I'd say let's keep that.

Indanz:
New push with the old cleanup code left intact, but only enabling IRQs when setting a notification and disabling IRQs when clearing the IRQHandler.

src/object/interrupt.c
Indanz: I switched the order in any case, now it's symmetrical with invokeIRQHandler_SetIRQHandler.

Indanz:
With this change the new invariant could be something like: "IRQs are disabled if no corresponding IRQ handler cap exists and enabled when there is a notification registered". And deliberately saying nothing about the IRQ enable state when there is no notification registered, but there is an IRQhandler present.

Summary for future readers why this code is so weird:

There is no dependency tracking between IRQhandlers and notifications, the only state is the notification cap copy in the global `intStateIRQNode` CNode and the global `intStateIRQTable` table.

Deleting/revoking the Notification will clear the cap copy (but keep the IRQ enabled and the IRQ state unchanged, leading to "Undelivered IRQ" messages in debug builds. Also because of this it's impossible to proof the IRQs are only enabled when there is a valid notification registered. We'd need proper IRQ cleanup at the notification side to solve this).

When deleting the IRQHander the right to use the IRQ is being revoked, so the notification cap should be removed from `intStateIRQNode`. We also want IRQs without an IRQHandle to be disabled, hence the cleanup I did before isn't allowed.

src/object/interrupt.c
lsf37: > I don't think it's possible for the final cap to be in the IRQ CNode, that's always a derived cap, never the final, I think. Does that help?

It is absolutely possible for the final cap to be in the IRQ CNode, you just need to delete the parent cap.

lsf37:
> With this change the new invariant could be something like: "IRQs are disabled if no corresponding IRQ handler cap exists and enabled when there is a notification registered". And deliberately saying nothing about the IRQ enable state when there is no notification registered, but there is an IRQhandler present.

It looks like that would be true, yes. I'm not sure yet if it will be strong enough for what the invariant is used for, let me try to track that down.

src/object/interrupt.c
Indanz: But wouldn't that also delete/revoke the derived caps? Or is that a special case of some cap types? Or am I just confused and it never works like that for any caps?

lsf37:
First results of tracing where that invariant is used (still incomplete, there are over 150 occurrences in just the invariant proofs):

For acknowledging an interrupt, we currently require that `interrupt_states s irq != IRQInactive`. With the new invariant we no longer know this, and it can in fact be violated, because acknowledging an interrupt only requires the `IRQHandler` cap. If you call Ack before setting a notification, the condition will be violated, because the presence of that cap alone no longer guarantees that the interrupt is not inactive.

We could fix this by checking for the presence of a notification cap in the IRQ node, because if we have the handler cap, we can locate the slot. This would either create a new error condition for the user or would be silently skipped. I think it should create an error, because if you're acknowledging interrupts before you can receive them, something is probably wrong.

The other occurrences so far point at a bit of trouble with the way we will need to phrase the new invariant. Currently, deleting caps does not violate the invariant, because it only says "if handler cap exists, then X". So if you remove a handler cap, that remains trivially true. If we now say, "if no handler cap exists, then Y", we will need to show that corresponding cleanup has happened already. But maybe we don't need that direction, and only need the presence of the notification.

Tying interrupt state to the presence of notifications will make a bunch of things harder, because we will now need to know where caps are when we do any cap operation, but I think we have most of those conditions already present for normal cases of copy/move/swap/mint/mutate. cteInsert will gain another special case, although we can hopefully remove or at least reduce the current special case for IRQHandler caps (maybe). 

Will keep looking.

Indanz:
Thanks for the update.

> For acknowledging an interrupt, we currently require that `interrupt_states s irq != IRQInactive`. With the new invariant we no longer know this, and it can in fact be violated, because acknowledging an interrupt only requires the `IRQHandler` cap. If you call Ack before setting a notification, the condition will be violated, because the presence of that cap alone no longer guarantees that the interrupt is not inactive.
> 
> We could fix this by checking for the presence of a notification cap in the IRQ node, because if we have the handler cap, we can locate the slot.

Seems easier to check for `interrupt_states s irq != IRQInactive` directly.

> This would either create a new error condition for the user or would be silently skipped. I think it should create an error, because if you're acknowledging interrupts before you can receive them, something is probably wrong.

I don't think so, the end state is as the user wants it to be, so I wouldn't return an error. Currently the call doesn't return an error when there is nothing to ack on a hw level either, this wouldn't be much different. Edit: Not exactly the same, there probably will be a notification set at least when the user acks.

Also acking first is currently needed sometimes to avoid the problem mentioned in the issue, all that code would trigger an error.

> The other occurrences so far point at a bit of trouble with the way we will need to phrase the new invariant. Currently, deleting caps does not violate the invariant, because it only says "if handler cap exists, then X". So if you remove a handler cap, that remains trivially true. If we now say, "if no handler cap exists, then Y", we will need to show that corresponding cleanup has happened already. But maybe we don't need that direction, and only need the presence of the notification.

Well, if that direction isn't trivial, I wouldn't bother with it.

> Tying interrupt state to the presence of notifications will make a bunch of things harder, because we will now need to know where caps are when we do any cap operation, but I think we have most of those conditions already present for normal cases of copy/move/swap/mint/mutate.

The IRQ notification cap can't be modified via those operations, is that what you mean?

> cteInsert will gain another special case, although we can hopefully remove or at least reduce the current special case for IRQHandler caps (maybe).

If it's easier, what about only proving that the IRQ gets enabled after a `SetIRQHandler` call (and disabled after `ClearIRQHandler`), and not tying it to the cap state at all? Because with the current version the IRQ doesn't get disabled when the cap disappears, so no need to track that in the proofs either.

midnightveil:
I feel like it would still be easier to do as Gerwin and I discussed earlier (https://github.com/seL4/seL4/pull/1539#discussion_r2485063130), where we track whether or not we got an undelivered IRQ, and set the notification active when binding. This wouldn't be the *best* solution perhaps, but it seems like it would be significantly less proof burden?

Indanz:
> I feel like it would still be easier to do as Gerwin and I discussed earlier ([#1539 (comment)](https://github.com/seL4/seL4/pull/1539#discussion_r2485063130)), where we track whether or not we got an undelivered IRQ, and set the notification active when binding. This wouldn't be the _best_ solution perhaps, but it seems like it would be significantly less proof burden?

Of course, and doing nothing at all would be even less work. The problem of missing IRQs is easy to work around by ACKing the IRQ first, or initialising the device differently, so it's not a big deal. Your alternative solution would only solve that particular corner case and probably still introduce some verification work.

I'm trying to make the behaviour sane. To me the bigger issue is that IRQs are enabled too early, not missing the initial one.

I don't like the idea that you can get interrupts for IRQs that are not being handled. Because IRQcontrol cap can't be copied, if you do any kind of delegation or partitioning, you are forced to create all IRQhandles at boot-up and delegate those instead. If of those hundreds IRQs, you only actually use a few, any others could theoretically fire at any moment for unknown reasons (hopefully only immediately, but who knows). This is not something you want to happen in a real-time system.

src/object/interrupt.c
lsf37: You can delete a single cap without revoking its children. (I.e. there's delete and there is revoke).

lsf37:
I actually like the solution here, because it makes IRQ enabling and notification setting atomic and gets us nice semantics. I have the feeling we probably would have done this from the start if we had thought of making it non-symmetric. Tying it only to the notification gives you the right semantics, but the deletion case is terrible (needs more storage somewhere and makes notifications complex). Deletion is simple when you tie it to the IRQHandler, but using IRQHandlers for enabling gets you the race condition. 

Having the IRQ binding atomic and the unbinding "lazy" looks to me like it will solve both problems without additional storage and have nice semantics.

I've finished looking through the abstract invariants and have not uncovered any new issues. However, the invariant is mentioned 300 times in the refinement proofs, so I should check that out as well to see if anything is lurking there. I think not, the high count is probably just because the refinement proofs have not yet factored out generic proofs from arch proofs (currently ongoing). But I should be sure before I try to pull the whole proof through.

src/object/interrupt.c
Indanz: Is that also true for UT? Because I remember having to keep a lot of caps around. Might have been the page caps though, I can't remember. Probably the mappings disappeared if I didn't keep the page caps around, and it percolated in my mind to all cap types.

lsf37: You can always delete a single cap. If that single cap happens to be the last cap to an object, then cleanup happens (the `finaliseCap` function) -- that is a different, additional process and might affect further caps in the system recursively, e.g. the notification cap in the IRQ slot for IRQHandler caps, or the contents of CNodes for a CNode cap, or just other kernel state like the mapping that a page cap is responsible for.

The recursion is bounded in depth, but in time it is only bounded by the number of caps in the system. It could affect a lot of other caps and state. It's the main reason the kernel has preemption -- most of the rest of the API is O(1) (until MCS introduced sorted queues, which theoretically makes IPC unbounded, but I'm not annoyed by that at all :-)).

Indanz: (You can keep MCS O(1) with careful design, also for the release queue.)

lsf37: From user level, sure, but it is no longer guaranteed by the kernel (where it was before). I don't like kernel properties that depend on well-behaving users. That's not really a kernel then, that's just a library. It's not quite as black and white as that, I know. Still, I don't like it.

Indanz: Agreed. I'm thinking about how to combine all the queues into one big priority heap structure somehow, but haven't though of a way yet. (Combining all schedule queues and release queue, including all queues per priority into one.)

Issue #1327, opened 2024-10-03
Pre-RFC: Give seL4 TCB fault message types unique IDs across kernel configurations


kent-mcleod:
Context: When a thread that has a fault handling endpoint faults, the kernel will attempt to deliver a Call-message over the endpoint. Another thread can handle the fault and resume the faulting thread by receiving the fault message and replying to the message with a response.  There are different fault message types because there are different ways that a thread can fault. Additionally, with hypervisor extensions, a thread with a VCPU attached will have additional fault types specific to hardware virtualization. The MCS configuration also provides additional fault types. Debug exception support also adds additional fault types. When receiving a fault message, the label inside the seL4_MsgInfo_t word contains an ID for the type of fault.  These values are assigned as part of the kernel ABI.

Problem: Within the same architecture, the same fault types can have different IDs. Fault handling code is required to know what configuration the kernel is on in order to decode the fault message.  Users have to make a choice between maintaining different binaries for different configurations or having to handle the different definitions at runtime when wanting to use common libraries/binaries for an architecture across different kernel configurations.

A change we should make is to assign these IDs: (https://github.com/seL4/seL4/blob/master/libsel4/arch_include/arm/sel4/arch/shared_types.bf#L9) for each architecture to be unique across kernel configurations, so for example the VMFault type has the same ID on MCS and on non-mcs.

This would be a breaking change to the ABI, it could be put behind a configuration flag though if needed.

kent-mcleod:
One question is whether this change is worth making.
Cost of the change:
- Verification impact likely low, but not zero.
- Binary incompatible change (although probably source compatible).

Benefits of the change:
- Make it easier to write down the fault API in different programming languages because definitions don't end up needing config macros.

Points against:
- fault handling code likely already needs to know the kernel configuration differences to handle faults properly.

Indanz:
I'm in favour.

Considering that `DebugException` is always 4, I think it may have been a mistake to make the fault numbers different for MCS. There is no reason why VMFault is 5 on non-MCS and 6 on MCS. The Timeout fault should have been 6. Arch specific faults should be at the end and unique. I don't think fault numbers need to be sequential, otherwise the missing 4 would be a problem for RISCV.

lsf37:
From the verification side I don't see any problems with this. It will be some update work, but should be fairly minimal.

Longer term this sounds like the right direction to me. Short term it might be an annoyance for users, because it is at least a binary breaking change, but it should still be source compatible.

[...]

PR #1542, opened 2025-10-31
Proposal: Can we make invocation numbers flat?

kent-mcleod:
Configuration dependent invocations cause invocation numbers to be unstable across kernel configurations within the same kernel release. This means libsel4.h clients need to be recompiled for different kernel configurations such as hyp vs non-hyp.

By adjusting the tooling to not #ifdef the invocation labels from the enum definitions, the numbers assigned for each label are consistent across kernel configurations.

MCS changes the parameters of the invocations from non-mcs and so we try assign different invocation labels for this. This still prevents using the same libsel4 for both MCS and non-MCS kernels...

When decoding CNode invocations, the kernel assumes all CNode operations within the starting label and ending label are consecutive. This isn't possible if invocation IDs become sparse. This consecutive assumption should be easy to resolve though.

Eg: 
```
enum invocation_label {
    InvalidInvocation,
    UntypedRetype,
    TCBReadRegisters,
    TCBWriteRegisters,
    TCBCopyRegisters,
// if !defined(CONFIG_KERNEL_MCS)
    TCBConfigure,
// endif
// if defined(CONFIG_KERNEL_MCS)
    TCBConfigureMCS,
// endif
    TCBSetPriority,
    TCBSetMCPriority,
// if !defined(CONFIG_KERNEL_MCS)
    TCBSetSchedParams,
// endif
// if defined(CONFIG_KERNEL_MCS)
    TCBSetSchedParamsMCS,
// endif
// if defined(CONFIG_KERNEL_MCS)
    TCBSetTimeoutEndpoint,
// endif
    TCBSetIPCBuffer,
// if !defined(CONFIG_KERNEL_MCS)
    TCBSetSpace,
// endif
// if defined(CONFIG_KERNEL_MCS)
    TCBSetSpaceMCS,
// endif
    TCBSuspend,
    TCBResume,
    TCBBindNotification,
    TCBUnbindNotification,
// if (!defined(CONFIG_KERNEL_MCS) && defined(CONFIG_ENABLE_SMP_SUPPORT))
    TCBSetAffinity,
// endif
// if defined(CONFIG_HARDWARE_DEBUG_API)
    TCBSetBreakpoint,
// endif
// if defined(CONFIG_HARDWARE_DEBUG_API)
    TCBGetBreakpoint,
// endif
// if defined(CONFIG_HARDWARE_DEBUG_API)
    TCBUnsetBreakpoint,
// endif
// if defined(CONFIG_HARDWARE_DEBUG_API)
    TCBConfigureSingleStepping,
// endif
    TCBSetTLSBase,
    TCBSetFlags,
    CNodeRevoke,
    CNodeDelete,
    CNodeCancelBadgedSends,
    CNodeCopy,
    CNodeMint,
    CNodeMove,
    CNodeMutate,
    CNodeRotate,
// if !defined(CONFIG_KERNEL_MCS)
    CNodeSaveCaller,
// endif
    IRQIssueIRQHandler,
    IRQAckIRQ,
    IRQSetIRQHandler,
    IRQClearIRQHandler,
    DomainSetSet,
// if defined(CONFIG_KERNEL_MCS)
    SchedControlConfigureFlags,
// endif
// if defined(CONFIG_KERNEL_MCS)
    SchedContextBind,
// endif
// if defined(CONFIG_KERNEL_MCS)
    SchedContextUnbind,
// endif
// if defined(CONFIG_KERNEL_MCS)
    SchedContextUnbindObject,
// endif
// if defined(CONFIG_KERNEL_MCS)
    SchedContextConsumed,
// endif
// if defined(CONFIG_KERNEL_MCS)
    SchedContextYieldTo,
// endif
    nInvocationLabels
};
```

### DIFF
--- a/libsel4/include/interfaces/object-api.xml
+++ b/libsel4/include/interfaces/object-api.xml
@@ -221,7 +221,7 @@
                 </description>
             </error>
         </method>
-        <method id="TCBConfigure" name="Configure" manual_name="Configure (MCS)" manual_label="tcb_configure_mcs">
+        <method id="TCBConfigureMCS" name="Configure" manual_name="Configure (MCS)" manual_label="tcb_configure_mcs">
             <condition><config var="CONFIG_KERNEL_MCS"/></condition>
             <brief>
                 Set the parameters of a TCB
@@ -353,7 +353,7 @@
                 </description>
             </error>
         </method>
-        <method id="TCBSetSchedParams" name="SetSchedParams" manual_name="Set Sched Params (MCS)" manual_label="tcb_setschedparams_mcs">
+        <method id="TCBSetSchedParamsMCS" name="SetSchedParams" manual_name="Set Sched Params (MCS)" manual_label="tcb_setschedparams_mcs">
             <condition><config var="CONFIG_KERNEL_MCS"/></condition>
             <brief>
                 Change a thread's priority, maximum controlled priority, scheduling context
@@ -487,7 +487,7 @@
             </error>
         </method>
 
-        <method id="TCBSetSpace" name="SetSpace" manual_name="Set Space (MCS)" manual_label="tcb_setspace_mcs">
+        <method id="TCBSetSpaceMCS" name="SetSpace" manual_name="Set Space (MCS)" manual_label="tcb_setspace_mcs">
             <condition><config var="CONFIG_KERNEL_MCS"/></condition>
             <brief>
                 Set the fault endpoint, CSpace and VSpace of a thread
--- a/src/object/tcb.c
+++ b/src/object/tcb.c
@@ -835,6 +835,11 @@ static exception_t decodeSetFlags(cap_t cap, word_t length, bool_t call, word_t 
     return EXCEPTION_NONE;
 }
 
+#ifdef CONFIG_KERNEL_MCS
+#define TCBConfigure TCBConfigureMCS
+#define TCBSetSchedParams TCBSetSchedParamsMCS
+#define TCBSetSpace TCBSetSpaceMCS
+#endif
 /* The following functions sit in the syscall error monad, but include the
  * exception cases for the preemptible bottom end, as they call the invoke
  * functions directly.  This is a significant deviation from the Haskell
--- a/tools/invocation_header_gen.py
+++ b/tools/invocation_header_gen.py
@@ -40,11 +40,11 @@ enum invocation_label {
     InvalidInvocation,
     {%- for label, condition in invocations %}
     {%- if condition %}
-#if {{condition}}
+// if {{condition}}
     {%- endif %}
     {{label}},
     {%- if condition %}
-#endif
+// endif
     {%- endif %}
     {%- endfor %}
     nInvocationLabels
@@ -69,7 +69,7 @@ enum sel4_arch_invocation_label {
             {%- if loop.first %}
 #error "First sel4_arch invocation label cannot be conditional"
             {%- endif %}
-#if {{condition}}
+// if {{condition}}
         {%- endif %}
         {%- if loop.first %}
     {{label}} = nInvocationLabels,
@@ -77,7 +77,7 @@ enum sel4_arch_invocation_label {
     {{label}},
         {%- endif %}
         {%- if condition %}
-#endif
+// endif
         {%- endif %}
     {%- endfor %}
     {%- if invocations|length == 0 %}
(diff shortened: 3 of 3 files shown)


kent-mcleod:
At this stage I'm just wondering if there would be a large impact to a change like this?

My motivation is to be able have CONFIG_ARM_HYPERVISOR_SUPPORT or CONFIG_HARDWARE_DEBUG_API  on/off without having to rebuild apps that include libsel4 headers. 

Being able to switch between MCS and non-MCS without recompiling is not a goal I have because application logic would have to also sufficiently change to account for the different Kernel model.

Indanz:
I think this would be a step in the right direction.

> MCS changes the parameters of the invocations from non-mcs and so we try assign different invocation labels for this. This still prevents using the same libsel4 for both MCS and non-MCS kernels...

Why give them different labels then, other than to get it compiling?

> When decoding CNode invocations, the kernel assumes all CNode operations within the starting label and ending label are consecutive. This isn't possible if invocation IDs become sparse. This consecutive assumption should be easy to resolve though.

I don't see how this assumption would break with your proposed change, it's still sequential. Only `CNodeSaveCaller` is special and that's already handled by the `CNODE_LAST_INVOCATION` define.

> Being able to switch between MCS and non-MCS without recompiling is not a goal I have because application logic would have to also sufficiently change to account for the different Kernel model.

Not much though, and it's limited to only a few places, mostly init. But init is the one that's easy to do dynamically, just an if instead of an ifdef.

midnightveil:
I was going to say there was already an issue about this: https://github.com/seL4/seL4/issues/1327 but it's slightly different and turns out it was opened by you.

The comments there are probably the same too. I think it's a good change and would definitely make other languages support be much easier too. But also yes even for C not having to recompile would be nice.

lsf37:
The only problem can I see  is that the compiler no longer warns you if you are using an invocation label that the current config does not support. If we're happy to make that trade-off, I don't think anything stands in the way of doing this.

It will need a verification update, but as I said on #1327, it should be manageable.

Indanz:
> The only problem can I see is that the compiler no longer warns you if you are using an invocation label that the current config does not support. If we're happy to make that trade-off, I don't think anything stands in the way of doing this.

If the syscall wrappers are not generated then the user still gets a compile error.

eternaleye:
Another option might be to introduce a degree of separation - have the userspace headers define them as the values of linker symbols, and emit a library based on the kernel configuration that satisfies them with the concrete value, if one is available.

This would then promote attempting to use an unsupported invocation to a linker error that reports the symbol's name (a usability improvement) if the provider library is statically linked, or a load-time error if it's dynamic.

Furthermore, because the _value_ of the linker symbol is used (not its _target_), relocations _should_ result in this being zero overhead at runtime.

This would avoid the need to modify the kernel at all, while allowing any userspace code to function with at most a relink (in the static case; in the dynamic case, it'd Just Work) on any kernel that provides the invocations it uses.

midnightveil:
I have a suggestion, which would also resolve the CNode ID sparseness/consecutive issue:

We should allocate some fixed number (say 0x1000) of invocations per cap type, and then group the invocation numbers by cap type.

For instance:

```
enum invocationLabel {
  Invalid,
  // any other debug ones i suppose 

  UntypedRetype = 0x1000,
  
  TCBSetSpace = 0x2000,
  TCBConfigure = 0x2001,
  TCBBlah = 0x2002,

  CNodeMint = 0x3000,
  CNodeCopy = 0x3001,
  // etc
}
```

And we assume these as fixed identifiers.

This keeps invocations on CNode contiguous within it's type (which should have no code changes). Having these large spacing (not that large compared to 32 bits or whatever) means we can easily add new syscalls or variations without losing this grouping property.

The value 'nInvocationLabels' currently at the end of the enum does lose it's meaning: but it's not actually used anywhere in the seL4 codebase.

We could likely do the same for the syscall IDs. (EDIT: This would not be possible without code changes, as currently we generate "syscall_min" (https://github.com/seL4/seL4/blob/ae11cf1450d810b4ecac1c151dbfcfceb9d96d35/tools/syscall_header_gen.py#L21 and for reasons unknown to me the trap code for syscalls does a "syscall in range" check before doing a switch statement which could just do the same thing. which would just slow down syscalls afaict).

(to be clear I would say specify the numbers in the XML and generate the C as before, just with numbers)

lsf37:
I'd be fine with that. We keep adding cap types, but as long as we're careful not to allocate too much for each cap type, that shouldn't be a problem.

Indanz:
There isn't a consecutive issue though, because all the CNode ones stay consecutive even after Kent's change.

That said, grouping by ID can solve other issues, like having the same syscall numbers for different archs and across kernel versions.

Currently the arch-specific invocations get appended, so you have basically multiple sets that get combined into one. If you do grouping like Julia proposes you can keep doing that, but if you have a global flat list, all arch specific invocation numbers must be in there too to keep the numbers stable across kernel versions.

Invocation numbers come from the message info label, so for 32-bits that's limited to 20 bits in total. That's enough, but we have to be careful not to waste it too much.

PR #1543, opened 2025-11-03
aarch64: fixup address space layout docs for hyp

### LINKED ISSUE #957: AArch64 memory map diagram is wrong

midnightveil:
Fixes #957

I removed the notes about TK1 since it's not just TK1.

### DIFF
--- a/include/arch/arm/arch/64/mode/hardware.h
+++ b/include/arch/arm/arch/64/mode/hardware.h
@@ -45,9 +45,7 @@
  *         2^64 - 2^48 +-------------------+                --+
  *                     |                   |
  *                     |                   |
- *                     |                   |
  *                     |   Unaddressable   |
- *                     |                   |
  *                     |                   |
  *                     |                   |
  *                2^48 +-------------------+                --+
@@ -79,7 +77,6 @@
  *            2^64 - 2^30 +-------------------+ PPTR_TOP
  *                        |                   |
  *                        |                   |
- *                        |                   |
  *                        |  Physical Memory  |
  *                        |      Window       |
  *                        |                   |
@@ -106,11 +103,11 @@
  *                      |                  |
  *                      |  Kernel Devices  | --> (PT)
  *                      |                  |
- *          2^64 - 2^21 +------------------+ KDEV_BASE
+ *          2^64 - 2^21 +------------------+ KDEV_BASE / KERNEL_PT_BASE
  *                      |                  |
  *                      |    Log Buffer    |
  *                      |                  |
- *      2^64 - 2 * 2^21 +------------------+ KS_LOG_BASE
+ *      2^64 - 2 * 2^21 +------------------+ KS_LOG_PPTR
  *                      |                  |
  *                      |                  |
  *                      |     Unmapped     |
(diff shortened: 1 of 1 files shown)


include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -127,69 +124,56 @@
...
- *    The CPU on the TX1 only allows for 48-bits of addressable virtual memory.
- *    Within the seL4 EL2 kernel's separate vaddrspace, it uses
- *    the 512 GiB at the top of that 48 bits of addressable
- *    virtual memory.
+ *    The layout of the kernel's virtual address space is very similar to
+ *    non-hyp, but the kernel address space is distinct to userland (EL1/EL0).
  *
- *    In EL2 there is only a TTBR0_EL2 register (unlike EL1 with TTBR0 and TTBR1)
- *    so the kernel is linked in the low 2^48 bits of the virtual address space,
- *    or "lower VA" in EL1 parlance.
```
Indanz, line 137: It's good to mention the missing TTBR1, so  I would keep that part, but perhaps make it clearer it's private to the kernel and not shared with user space, like for non-HYP.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -127,69 +124,56 @@
...
+ *    The major difference to non-hyp mode, is that the kernel mem region is
+ *    located in the first entry of the TTBR0_EL2 page table (PGD) rather than
+ *    the topmost. This is because:
```
Indanz, line 135: Isn't it in the second entry instead of the first?

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -127,69 +124,56 @@
...
  *
- *    The memory map used by the EL2 kernel's separate address space
- *    ends up looking something like this:
+ *     -  The ut_max_less_than_canonical assert, as we can't allow overflows
+ *        in untypeds. (see commit a94d90598fec484636b14e1b8d3643156e1e27b1)
```
Indanz: That's not the reason, that assert is there for the same reason. The commit message and code comments aren't that clear either though.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -261,9 +245,9 @@
 /* This is a page table mapping at the end of the virtual address space
  * to map objects with 4KiB pages rather than 4MiB large pages. */
```
Indanz: Large pages are 2MB on Aarch64.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: Uh, yeah. zero-indexing confuses once again.

include/arch/arm/arch/64/mode/hardware.h
midnightveil: I would appreciate some clarification here because I don't really understand the message of the commit nor the reason for the assert in the UT personally.

Indanz: The main thing is:

> For non-hyp, PPTR_BASE is in the top part of the address space and device untyped
>   addresses are allowed to be large enough to overflow and be in the bottom half of
>   the address space.  However, when the kernel is in EL2 it is not possible to safely
>   overflow without going into address ranges that are non-canonical.

The check makes sure that device untyped addresses don't overflow into unreachable address range for the kernel.

So far so good. What I don't understand is why it's fine for non-hyp to overflow, that's not explained anywhere.

I suspect it's because how the kernel uses it, it will underflow back into the original value and everything is okay. And the reason that this doesn't work for hyp is because when it overflows, the number becomes 49 bits big and doesn't fit in the cap any more. Or something like that.

Edit: Or it's just a funny feature that one untyped cap can cover both the beginning and the start of an address range on non-hyp?

midnightveil: > I suspect it's because how the kernel uses it, it will underflow back into the original value and everything is okay. And the reason that this doesn't work for hyp is because when it overflows, the number becomes 49 bits big and doesn't fit in the cap any more. Or something like that

There's the mention of canonical addreses, maybe it's because an overflow in the 0xffffffXXXXXX range overflows into the 0x00000XXXXX range, which is still a valid canonical address. For hyp, since we're not at the top of the 64 bits range, an overflow off the end of the 0x0000000XXXXX range goes into the 0x000001XXXXX range which is non-canonical. Maybe?

Why overflows /this matter at all, I don't understand.

---

So I guess the reason is that if we don't use the top PGD entry, then an overflow remains within the canonical range, as we're not right at the top of the 48 bits.? If I'm correct I think I understand the comment in the commit message now, though I don't understand *why*, just *what*.

midnightveil: I've adjusted this. Still not sure on it.

lsf37: Overall, we do need all addresses, even the ones in device-untypeds, to be canonical (they don't get dereferenced, but they do get compared to other addresses). So if a large physical address could make the kernel-virtual address non-canonical, we'd need to prevent that. 

I'm not sure what the original commit intended, but from the proof perspective, only EL2 (hyp) has a proof for AArch64 and will actually force wellformedness requirements. That might explain why we assert something only for hyp. It does not mean the non-hyp case is actually correct (although it could well be). I'm not entirely sure an untyped that is overflowing on the virtual address space side is Ok. It might be. Depends on whether the isMDBParentOf checks and all other checks that talk about the end address of an Untyped assume that start < end.

include/arch/arm/arch/64/mode/hardware.h
```diff
@@ -127,69 +124,64 @@
...
- *    |                                   |    which is the rest of
- *    |                                   |    the lower 256 TiB.
- *    |                                   |    |
- *    |                                   |    v
- *    +-----------------------------------+ <- 0x0
+ *     -  the kernel requires that the PPTR addresses stored in cap slots are
+ *        valid (i.e. canonical); whilst this was fine for non-hyp as overflows
+ *        of device UTs (starting from PPTR_BASE) go into the "lower VA" range
+ *        which is allowed; it is not fine for hyp as overflows above 2^48 go
+ *        into the non-canonical range. Moving the PPT_RBASE down to the second
```
lsf37:
```suggestion
 *        into the non-canonical range. Moving the PPTR_BASE down to the second
```

include/arch/arm/arch/64/mode/hardware.h
midnightveil: discussion conclusion:


oh, wait, I think I see.

the kernel expects to be able to create untypeds from 0 to KernelPaddrUserTop (CONFIG_KERNEL_PADDR_USER_TOP), which can be as much as PA44
if PADDR_TOP lies in the top address range in non-hyp, the overflow in the addresses will overflow down the the 0...2^48 range, which are still valid non-canonical addresses (note: the kernel won't dereference these, and the overlapping checks use start+size)
however in EL2 because we can't liver in the upper VA range, the top entry of the PGD would be (2^48 - 2^40 ... 2^48). For device untypeds up in the 44 bit range these would exceed the maxiumum lower canonical address 2^48 - 1, and reach into the non-canonical range.
as the proofs require that all untyped addresses are valid vritual addresses (even if not dereferenced), this is invalid
so we move it down to the bottom

Indanz: Pretty much. But to me it's still unclear whether it's just a theoretical problem (non-canonical address) or also a practical problem (address doesn't fit in cap's compressed pointer format or address gets mangled after overflow in a way that isn't undone when underflowing). Maybe it's both.

midnightveil: I've updated the description with that now.

> But to me it's still unclear whether it's just a theoretical problem (non-canonical address) or also a practical problem (address doesn't fit in cap's compressed pointer format or address gets mangled after overflow in a way that isn't undone when underflowing)

It sounds to me at the very least something the proofs use? Might be wrong there.

I'm happy with this description now.

Issue #1540, opened 2025-10-31
incorrect spurious interrupt reporting following commit 0f78e4e95776c56afac077a5d9971314353e4b19


midnightveil:
0f78e4e95776c56afac077a5d9971314353e4b19 as part of #1510 by @lsf37 refactors out both `handleInterruptEntry` and `handleSyscall` to call a `checkInterrupt` function, which prints out if there's a spurious interrupt.

https://github.com/seL4/seL4/blob/0f78e4e95776c56afac077a5d9971314353e4b19/src/api/syscall.c#L38-L51

**However**, since this was called around `mcsPreemptionPoint`:

https://github.com/seL4/seL4/blob/0f78e4e95776c56afac077a5d9971314353e4b19/src/api/syscall.c#L524-L537

This means that every time we take a syscall and *there is no interrupt* we print out a "spurious interrupt" message. This means we print a *lot* of spurious interrupts, one for every syscall.

```
Running test SCHED0007 (Test IPC priorities)
Test SCHED0007 passed
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @4236fc]: Spurious interrupt!>>
        </testcase>
        <testcase classname="sel4test" name="SCHED0008">
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @41a1f8]: Spurious interrupt!>>
<<seL4(CPU 0) [checkInterrupt/47 T0xffffff80fffd4800 "sel4test-driver" @41a1f8]: Spurious interrupt!>>
Running test SCHED0008 (Test changing prio while in endpoint queues results in correct message order)
[...]

PR #1544, opened 2025-11-04
mcs: hotfix for spurious IRQs due to time preempt

midnightveil:
This is a hotfix for #1540.

I'm not going to say this is a *fix* for that issue since it feels like a bit of a hack, but at the moment #1510 is causing a large amount of (clearly incorrect) spurious IRQ warnings on MCS, that's also breaking the CI.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -327,9 +327,16 @@ static exception_t handleInvocation(bool_t isCall, bool_t isBlocking)
                               isBlocking, isCall, buffer);
 #endif
 
+#ifdef CONFIG_KERNEL_MCS
+    if (unlikely(status == EXCEPTION_PREEMPTED || status == EXCEPTION_TIME)) {
+        return status;
+    }
+#else
     if (unlikely(status == EXCEPTION_PREEMPTED)) {
         return status;
     }
+#endif
+
 
     if (unlikely(status == EXCEPTION_SYSCALL_ERROR)) {
         if (isCall) {
@@ -531,7 +538,13 @@ exception_t handleSyscall(syscall_t syscall)
             ret = handleInvocation(false, true, false, false, getRegister(NODE_STATE(ksCurThread), capRegister));
             if (unlikely(ret != EXCEPTION_NONE)) {
                 mcsPreemptionPoint();
+#ifdef CONFIG_KERNEL_MCS
+                if (ret == EXCEPTION_PREEMPTED) {
+                    checkInterrupt();
+                }
+#else
                 checkInterrupt();
+#endif
             }
 
             break;
@@ -540,7 +553,13 @@ exception_t handleSyscall(syscall_t syscall)
             ret = handleInvocation(false, false, false, false, getRegister(NODE_STATE(ksCurThread), capRegister));
             if (unlikely(ret != EXCEPTION_NONE)) {
                 mcsPreemptionPoint();
+#ifdef CONFIG_KERNEL_MCS
+                if (ret == EXCEPTION_PREEMPTED) {
+                    checkInterrupt();
+                }
+#else
                 checkInterrupt();
+#endif
             }
             break;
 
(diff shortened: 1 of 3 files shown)


src/api/syscall.c
```diff
@@ -327,9 +327,16 @@ static exception_t handleInvocation(bool_t isCall, bool_t isBlocking)
                               isBlocking, isCall, buffer);
 #endif
 
+#ifdef CONFIG_KERNEL_MCS
+    if (unlikely(status == EXCEPTION_PREEMPTED || status == EXCEPTION_TIME)) {
```
Indanz, line 331: What about 
```suggestion
    if (unlikely(status >= EXCEPTION_PREEMPTED) {
```
and use that for both?

src/api/syscall.c
```diff
@@ -531,7 +538,13 @@ exception_t handleSyscall(syscall_t syscall)
             ret = handleInvocation(false, true, false, false, getRegister(NODE_STATE(ksCurThread), capRegister));
             if (unlikely(ret != EXCEPTION_NONE)) {
```
Indanz, line 539: #Probably want to change this to `status >= EXCEPTION_PREEMPTED` too everywhere.

Indanz (commented):
Maybe an easier fix would be to pass an argument to checkInterrupt() to tell it whether it was called via interrupt or not. It would be unused in verification builds, and in debug builds it would suppress the printf.

src/api/syscall.c
lsf37: Hm, yes, you're right, this is annoying. We now need to check for this everywhere where we just had `EXCEPTION_PREEMPTED` before.

lsf37:
> Maybe an easier fix would be to pass an argument to checkInterrupt() to tell it whether it was called via interrupt or not. It would be unused in verification builds, and in debug builds it would suppress the printf.

If it is only visible for debug builds that would work, but it does feel strange because I was trying to get rid of this very argument on the verification side (so I get that equality between preemption and entry). It's probably less invasive, though, and would work until we have a better solution for MCS preemption. So, yes, I'd vote for trying that out.

src/api/syscall.c
midnightveil: Fortunately it's just these 5 places, as far as I can tell.

The solution where we only print depending on a debug-only argument to spurious IRQ kinda works. though it means it calls handleSpuriousIrq inappropriately for time preempts; whether or not it's ok idk.

it would fix the prints though

lsf37: It'd be equivalent to the current behaviour. I don't think it should be doing that in the long run -- we shouldn't check for IRQ when we know there isn't one, but maybe that at least documents in the code that we have looked at this. Maybe add a comment with a link to the issue.

Indanz:
> > Maybe an easier fix would be to pass an argument to checkInterrupt() to tell it whether it was called via interrupt or not. It would be unused in verification builds, and in debug builds it would suppress the printf.
> 
> If it is only visible for debug builds that would work, but it does feel strange because I was trying to get rid of this very argument on the verification side (so I get that equality between preemption and entry). It's probably less invasive, though, and would work until we have a better solution for MCS preemption. So, yes, I'd vote for trying that out.

The better solution is removed all this horrible interrupt checking around syscalls as implemented by https://github.com/seL4/seL4/commit/8ae457a4d40b8360404fd0c89bb042631e0140e0 (part of [Endpoint Budget Threshold](https://github.com/Indanz/seL4/pull/2)).

I should open a PR and test the slowdown for long-running operations on the platform with the biggest kernel entry/exit time + schedule overhead compared to the configured WCET time.

But any slowdown can easily be fixed by implementing the optimisation from https://github.com/seL4/seL4/pull/1363/commits/5cfa3943ce431bb4abf0a5a7ba3a2bf8eb1349a0: Increase `CONFIG_RESET_CHUNK_BITS` from 8 to 14 and call `preemptionPoint()` once instead of doing the above `CONFIG_MAX_NUM_WORK_UNITS_PER_PREEMPTION` times.

src/api/syscall.c
midnightveil: I don't like it, because it makes it really hard to find uses of the enum values; the actual values of them don't really have meaning in this case. It's /fine/ but my preference is for the simpler, and (I highly doubt) any significantly less performant ||.

Plus it probably breaks the non-MCS proof (simply because it's changed).

Indanz: The `||` would cost one or two more instructions, while the `>=` would be free. No idea what's easier to proof.

midnightveil: Yes, but we generally haven't optimised seL4 to that level of precision. I doubt that the extra few cycles would matter compared to the actual overheads of the operations involved.

midnightveil:
Closing in favour of #1546.

(Doing it this way so the commits for this solution are still here, in case people want to reference it. Pushing to this branch would lose that history).

PR #1546, opened 2025-11-04
mcs: hotfix for incorrect spurious IRQ reporting

midnightveil:
This is a hotfix for issue #1540, an a less intrusive alternative to #1544.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -34,8 +34,19 @@
 #include <mode/machine/debug.h>
 #endif
 
-
+/**
+ * FIXME: This is a temporary hack to prevent the printing of incorrect
+ *        spurious interrupt warnings on MCS when checkInterrupt() is called
+ *        following preemptionPoint and the reason is because running
+ *        out of sufficient budget, rather than an active IRQ.
+ *        See issue https://github.com/seL4/seL4/issues/1540 and
+ *        https://github.com/seL4/seL4/pull/1544.
+ **/
+#ifdef CONFIG_IRQ_REPORTING
+static inline void checkInterrupt(bool_t was_interrupt_entry)
+#else
 static inline void checkInterrupt(void)
+#endif
 {
     irq_t irq;
 
@@ -44,11 +55,18 @@ static inline void checkInterrupt(void)
         handleInterrupt(irq);
     } else {
 #ifdef CONFIG_IRQ_REPORTING
-        userError("Spurious interrupt!");
+        if (was_interrupt_entry) {
+            userError("Spurious interrupt!");
+        }
 #endif
         handleSpuriousIRQ();
     }
 }
+
+#ifndef CONFIG_IRQ_REPORTING
+/** Part of the temporary hack above **/
+#define checkInterrupt(was_interrupt_entry) checkInterrupt()
+#endif
 
 /* The haskell function 'handleEvent' is split into 'handleXXX' variants
  * for each event causing a kernel entry */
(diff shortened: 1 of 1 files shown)


src/api/syscall.c
```diff
@@ -34,8 +34,19 @@
 #include <mode/machine/debug.h>
 #endif
 
-
+/**
+ * FIXME: This is a temporary hack to prevent the printing of incorrect
+ *        spurious interrupt warnings on MCS when checkInterrupt() is called
+ *        following preemptionPoint and the reason is because of the time
+ *        expiry, rather than an active IRQ.
```
Indanz: Reason is budget becoming less than `MIN_BUDGET`, not time expiry.

src/api/syscall.c
```diff
@@ -34,8 +34,19 @@
...
+ *        See issue https://github.com/seL4/seL4/issues/1540 and
+ *        https://github.com/seL4/seL4/pull/1544.
+ **/
+#ifdef CONFIG_IRQ_REPORTING
+static inline void checkInterrupt(bool_t was_interrupt_entry)
+#else
 static inline void checkInterrupt(void)
+#endif
```
Indanz, line 49: Isn't it easier to always have the argument, even when unused? This is kind of horrible.

midnightveil: I don't want to break verification for this change.

src/api/syscall.c
midnightveil: Is that not the same?

Indanz: To me "time expiry" sounds like pending timer or something. But the reason why not checking for interrupts (including timer interrupt) is sufficient for MCS, is because they want to do the MIN_BUDGET check before running for possibly WCET time again.
```suggestion
 *        following preemptionPoint and the reason is because running
 *        out of sufficient budget, rather than an active IRQ.
```

src/api/syscall.c
Indanz: But it will break MCS verification anyway, better to keep the changes between MCS and non-MCS minimal, at least that's my thinking. Assuming that's possible, of course.

midnightveil: Well, my previous change broke MCS, but not non-MCS. This one breaks neither until we solve the actual issue (as IRQ reporting is only on when printing is enabled which is only on for non-verification)

Indanz: Ah, I missed that for some reason. Good point. Okay, what about adding the following after the `checkInterrupt()` definition:

```
#ifndef CONFIG_IRQ_REPORTING
/* Hide */
#define checkInterrupt(was_interrupt_entry) checkInterrupt()
#endif
```
so the damage is limited to one place and not all call sites?

midnightveil: Sure, I can do that... but this is supposed to be a hotfix, and being ugly is more of a feature (it makes it obvious there's something to be fixed if someone reads this code). Whereas hiding it makes it look like this is just how it's supposed to be? It's not supposed to remain for long.

But, to be fair, I do get annoyed with the amount of `#ifdef` in various places around the kernel, so I'm ambivalent (makes it hard to understand sometimes).

lsf37: I like Indan's hacky preprocessor suggestion -- we do have a large FIXME at the declaration site and an open issue for it, so reducing the `ifdef` pain a bit sounds admissible to me. It still won't look like it is supposed to be this way.

midnightveil:
I've done that now.

lsf37 (approved):

PR #1547, opened 2025-11-06
SBI cap and CNode Minting with long badge 

Ivan-Velickovic:
This is an alternative implementation to the original PR
for RFC-22.

Instead of having a unique 'double badging' implementation as done in https://github.com/seL4/seL4/pull/1532, this implementation adds more general functionality to mint a cap with two word-sized badges instead of one with a CNodeMintLong invocation.

This is generally less complexity than the original PR as `Arch_isMDBParentOf` and `Arch_updateCapData` has the same logic as other caps such as SMC and endpoints.

Putting this up as a draft so people can look at the difference between the two approaches.

### DIFF
--- a/include/arch/riscv/arch/64/mode/object/structures.bf
+++ b/include/arch/riscv/arch/64/mode/object/structures.bf
@@ -68,10 +68,8 @@ block asid_pool_cap {
 block sbi_cap {
     field capSBIFIDBadge 64
     field capType 5
-    field capSBIEIDBadge 28
-    field capSBIEIDBadged 1
-    field capSBIFIDBadged 1
-    padding 29
+    field capSBIEIDBadge 32
+    padding 27
 }
 #endif
 
--- a/include/arch/riscv/arch/object/objecttype.h
+++ b/include/arch/riscv/arch/object/objecttype.h
@@ -14,6 +14,7 @@
 
 deriveCap_ret_t Arch_deriveCap(cte_t *slot, cap_t cap);
 cap_t CONST Arch_updateCapData(bool_t preserve, word_t data, cap_t cap);
+cap_t CONST Arch_updateCapDataLong(bool_t preserve, word_t data, word_t data2, cap_t cap);
 cap_t CONST Arch_maskCapRights(seL4_CapRights_t cap_rights_mask, cap_t cap);
 finaliseCap_ret_t Arch_finaliseCap(cap_t cap, bool_t final);
 bool_t CONST Arch_sameRegionAs(cap_t cap_a, cap_t cap_b);
@@ -46,33 +47,16 @@ static inline CONST bool_t Arch_isMDBParentOf(cap_t cap_a, cap_t cap_b, bool_t f
     switch (cap_get_capType(cap_a)) {
 #ifdef CONFIG_ALLOW_SBI_CALLS
     case cap_sbi_cap: {
-        if (!cap_sbi_cap_get_capSBIEIDBadged(cap_a)) {
-            /* If there is no EID badge then the cap cannot be badged. */
-            return true;
-        }
-
-        if (!cap_sbi_cap_get_capSBIEIDBadged(cap_b)) {
-            return false;
-        }
-
         word_t eid_a_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_a);
         word_t eid_b_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_b);
+        word_t fid_a_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_a);
+        word_t fid_b_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_b);
 
         if (eid_a_badge != eid_b_badge) {
             return false;
         }
 
-        if (!cap_sbi_cap_get_capSBIFIDBadged(cap_a)) {
-            return !firstBadged;
-        }
-
-        if (!cap_sbi_cap_get_capSBIFIDBadged(cap_b)) {
-            return false;
-        }
-
-        word_t fid_a_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_a);
-        word_t fid_b_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_b);
-        return (fid_a_badge == fid_b_badge && !firstBadged);
+        return (eid_a_badge == eid_b_badge && fid_a_badge == fid_b_badge && !firstBadged);
         break;
     }
 #endif
--- a/include/arch/riscv/arch/32/mode/object/structures.bf
+++ b/include/arch/riscv/arch/32/mode/object/structures.bf
@@ -60,10 +60,8 @@ block asid_pool_cap {
 
 #ifdef CONFIG_ALLOW_SBI_CALLS
 block sbi_cap {
-    field capSBIFIDBadge 30
-    field capSBIEIDBadged 1
-    field capSBIFIDBadged 1
-    field capSBIEIDBadge 28
+    field capSBIEIDBadge 32
+    field capSBIFIDBadge 28
     field capType 4
 }
 #endif
--- a/include/arch/riscv/arch/object/structures.h
+++ b/include/arch/riscv/arch/object/structures.h
@@ -148,10 +148,6 @@ static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 #ifdef CONFIG_ALLOW_SBI_CALLS
     switch (cap_get_capType(derivedCap)) {
     case cap_sbi_cap:
-        if (!cap_sbi_cap_get_capSBIEIDBadged(derivedCap) || !cap_sbi_cap_get_capSBIFIDBadged(derivedCap)) {
-            return true;
-        }
-
         return (cap_sbi_cap_get_capSBIFIDBadge(derivedCap) !=
             cap_sbi_cap_get_capSBIFIDBadge(srcCap) ||
             cap_sbi_cap_get_capSBIEIDBadge(derivedCap) !=
--- a/include/object/objecttype.h
+++ b/include/object/objecttype.h
@@ -21,6 +21,7 @@ bool_t CONST hasCancelSendRights(cap_t cap);
 bool_t CONST sameRegionAs(cap_t cap_a, cap_t cap_b);
 bool_t CONST sameObjectAs(cap_t cap_a, cap_t cap_b);
 cap_t CONST updateCapData(bool_t preserve, word_t newData, cap_t cap);
+cap_t CONST updateCapDataLong(bool_t preserve, word_t newData, word_t newData2, cap_t cap);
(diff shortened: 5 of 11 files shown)


include/arch/riscv/arch/object/objecttype.h
```diff
@@ -43,5 +44,25 @@ static inline CONST bool_t Arch_isIRQControlDescendant(cap_t cap)
...
+        word_t eid_a_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_a);
+        word_t eid_b_badge = cap_sbi_cap_get_capSBIEIDBadge(cap_b);
+        word_t fid_a_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_a);
+        word_t fid_b_badge = cap_sbi_cap_get_capSBIFIDBadge(cap_b);
+
+        if (eid_a_badge != eid_b_badge) {
+            return false;
+        }
+
+        return (eid_a_badge == eid_b_badge && fid_a_badge == fid_b_badge && !firstBadged);
```
Indanz, line 59: This is missing the unbadged cap_a case. I think the above if-check should be `if (!eid_a) return true;`.

Also, checking just `!firstBadged` would probably be sufficient. But if you do want to compare all fields, you could do a structure compare: `return cap_a == cap_b && !firstBadged;`. I guess the CParser doesn't support that though.

libsel4/arch_include/riscv/interfaces/object-api-arch.xml
```diff
@@ -367,5 +384,24 @@
...
 
     </interface>
+        <interface name="seL4_RISCV_SBI" manual_name="SBI" cap_description="Capability to allow threads to make SBI (Supervisor Binary Interface) calls.">
+        <method id="RISCVSBICall" name="Call" manual_name="SBI Call" manual_label="sbi_call">
+            <brief>
+                Specify SBI call
+            </brief>
+            <description>
+                Register a6 is for the SBI function ID (FID).
+                Register a7 is for the SBI extension ID (EID).
```
Indanz, line 394: These shouldn't be necessary any more.

Indanz (commented):
Main comment is that I wouldn't introduce new syscalls, nor `updateCapDataLong()` , but add a new, optional argument to the existing syscalls on 32-bit platforms and make capData 64-bits everywhere.

CNodeMint and perhaps CNodeMutate can then check for the upper 32-bit in the optional extra parameter.

This is much less change to the code and should be less work for verification too. The tricky bit is to generate the syscall wrapper for the invocation with the extra argument, and only for 32-bit platforms.

include/arch/riscv/arch/object/objecttype.h
Indanz: Or introduce a function that returns the 64-bit badge value of EID+FID and use that instead. Maybe that's cleaner, then all the juggling happens in that function for 32-bit only.

include/arch/riscv/arch/64/mode/object/structures.bf
```diff
@@ -64,6 +64,15 @@ block asid_pool_cap {
     field_high  capASIDPool     37
 }
 
+#ifdef CONFIG_ALLOW_SBI_CALLS
+block sbi_cap {
+    field capSBIFIDBadge 64
+    field capType 5
+    field capSBIEIDBadge 32
```
Indanz, line 71: I'd combine these two into one 64-bit badge value on 64-bit.

libsel4/arch_include/riscv/interfaces/object-api-arch.xml
lsf37: They are there to be compatible with the SBI calling convention

include/arch/riscv/arch/object/objecttype.h
lsf37: > This is missing the unbadged cap_a case. I think the above if-check should be if (!eid_a) return true;

Yes, that first `if` there is wrong. I'm pretty sure I saw Ivan change this to a 0 check on his screen, maybe an undo went one step too far.

> Also, checking just !firstBadged would probably be sufficient. 

No, that's not sufficient. If the badges are different, the caps belong to different trees.

> But if you do want to compare all fields, you could do a structure compare: return cap_a == cap_b && !firstBadged;. I guess the CParser doesn't support that though.

That would indeed look nicer and since there is nothing else in those caps, it would work. Actually that would also work for endpoints and notifications, because we're only calling that function if the object pointers are equal. Now that I'm saying that, I don't think it's nicer any more, though. The current version makes explicit what we actually want to know (the badges). Comparing the full caps would work and looks neat, but it's not really what the check is about, it just happens to coincide. I don't think we should do that.

include/arch/riscv/arch/64/mode/object/structures.bf
lsf37: What would become better if we did this? The bitfield declarations are there to do encoding for us, and we want to be able to tell EID and FID apart for error reporting and calling convention. If we put everything into one value, we'd just have to encode/decode manually.

lsf37:
> Main comment is that I wouldn't introduce new syscalls, nor `updateCapDataLong()` , but add a new, optional argument to the existing syscalls on 32-bit platforms and make capData 64-bits everywhere.

That was actually our first implementation and works fine from the kernel side, but it becomes very awkward in libsel4 and the syscall stub generation. Changing the latter would have been a much larger change, or at least we didn't see how to make it a small one.

Otherwise I do agree, I'd actually like that better, because it is less code and proof duplication. The duplication is not so terrible in the proof (it's annoying, but not hard), but it'd still be nicer to avoid.

> CNodeMint and perhaps CNodeMutate can then check for the upper 32-bit in the optional extra parameter.

We left out CNodeMutate, because it cannot be used to set badges (only CNode cap data or other data that does not affect the cap derivation tree). Given that it's additional proof work I'd leave that feature until there is a cap type that can actually use it.

> This is much less change to the code and should be less work for verification too. The tricky bit is to generate the syscall wrapper for the invocation with the extra argument, and only for 32-bit platforms (or do it always in case we need badges greater than 64-bits in the future).

If we do it, we should do it generically, otherwise we gain a lot of ifdefs.

libsel4/arch_include/riscv/interfaces/object-api-arch.xml
lsf37: Ah, you mean, because the info is in the cap? We still do support all calls with the unbadged SBI cap.

Indanz: Oh yes, I forgot about that.

include/arch/riscv/arch/64/mode/object/structures.bf
Indanz: I think everything becomes simpler if we do that, especially on 64-bits. Maybe I'm wrong and it turns out worse.

You don't have to print out the EID and FID separately when printing out the error. Just say badge value `x` doesn't match provided EID `y` and FID `z`. That would actually be more useful because then people can see that they need to swap the values when setting the badge.

Indanz:
> > Main comment is that I wouldn't introduce new syscalls, nor `updateCapDataLong()` , but add a new, optional argument to the existing syscalls on 32-bit platforms and make capData 64-bits everywhere.
> 
> That was actually our first implementation and works fine from the kernel side, but it becomes very awkward in libsel4 and the syscall stub generation. Changing the latter would have been a much larger change, or at least we didn't see how to make it a small one.

What happens if you have the long version with the same `CNodeMint` ID in the XML, does that break the parser? I can have a look, see if I can find an easy work-around or fix for this limitation.

It would be a good feature to have, because it's useful to be able to do this, it came up before.

include/arch/riscv/arch/64/mode/object/structures.bf
lsf37: I don't see it. We'd get a lot of ifdef between 64 and 32 bit if we do that and I don't see anything that becomes simpler. We'd have one == instead of two in two places, but we'd have to do bit shifting to compute the value we need to compare to and the user would need to understand that and do the same bit shifting when setting the badge. The user would also have different APIs between 64 and 32 bit versions. That does not seem simpler to me.

lsf37:
> What happens if you have the long version with the same `CNodeMint` ID in the XML, does that break the parser? I can have a look, see if I can find an easy work-around or fix for this limitation.

I think the parser gets confused, but we have not actually just tried it.


> It would be a good feature to have, because it's useful to be able to do this, it came up before.

It did, and if we can make that work without too much trouble, that'd be nice.

include/arch/riscv/arch/object/objecttype.h
Indanz: > > Also, checking just !firstBadged would probably be sufficient.
> 
> No, that's not sufficient. If the badges are different, the caps belong to different trees.

But wouldn't firstBadged be set for cap_b then? Example derivation tree:

1) Mint a new cap:
cap_0 -> cap_b(firstBadged)
2) Mint another cap from the zero cap:
cap_0 -> cap_a(firstBadged) -> cap_b(firstBadged)
3) Copy the newest cap:
cap_0 -> cap_a(firstBadged) -> cap_a'(!firstBadged) -> cap_b(firstBadged)
4) Copy again:
cap_0 -> cap_a(firstBadged) -> cap_a'(!firstBadged) -> cap_a''(!firstBadged) -> cap_b(firstBadged)

Other than for the zero cap, I don't see any transitions that aren't caught with a !firstBadge check, badge value doesn't matter.

And this is because like I said, if you can handle the corner case where the badge values happen to be the same by coincidence, you don't need to check the badge values any more. The whole thing works correctly if everyone is using the same badge value. In other words, it adds no extra information and can be ignored.

If I'm right, this is true for all cap types and would mean you could have a cap type independent `isMDBParentOf()`, at least as far as badging is concerned. The more exotic cap types would still need their own additional checks, but all the simple ones could be just this one check. But it's late here, so I'm probably wrong.

corlewis:
> What happens if you have the long version with the same `CNodeMint` ID in the XML, does that break the parser? I can have a look, see if I can find an easy work-around or fix for this limitation.
> 
> It would be a good feature to have, because it's useful to be able to do this, it came up before.

If I'm understanding correctly, it did indeed come up before and you already fixed it as part of some work that unfortunately never quite managed to be finished. See https://github.com/seL4/seL4/pull/889/commits/c2f716984b987bf918b1e0db1fa94ef33b308832 for that fix.

Indanz:
> If I'm understanding correctly, it did indeed come up before and you already fixed it as part of some work that unfortunately never quite managed to be finished. See [c2f7169](https://github.com/seL4/seL4/commit/c2f716984b987bf918b1e0db1fa94ef33b308832) for that fix.

Thanks for finding this! It's part of: https://github.com/seL4/seL4/pull/889

It works because only the enum handling is broken for duplicate IDs, and by keeping track of what we defined, we avoid adding duplicates. I forgot why checking the condition was necessary, but that's the part that would disappear if Kent has his way.

Would be nice to avoid the duplication of the API description in the XML and have optional arguments understood by the script directly, so it generates multiple versions for all combinations. But this is more work and can always be added later.

PR #1548, opened 2025-11-06
x86: fix VCPU object creation

matneutrality:
VCPU objects are "physical" objects: they do use memory areas allocated from Untyped. Details and implications shared with @lsf37  some time ago.

### DIFF
--- a/include/arch/x86/arch/object/structures.h
+++ b/include/arch/x86/arch/object/structures.h
@@ -112,7 +112,7 @@ compile_assert(vtd_pt_size_sane, VTD_PT_INDEX_BITS + VTD_PTE_SIZE_BITS == seL4_I
 #define EPT_PT_PTR(r)    ((ept_pte_t *)(r))
 #define EPT_PT_REF(p)    ((word_t)(p))
 
-#define VCPU_PTR(r)       ((vcpu_t *)(r))
+#define VCPU_PTR(r)       ((struct vcpu *)(r))
 #define VCPU_REF(p)       ((word_t)(p))
 
 #endif /* CONFIG_VTX */
@@ -225,6 +225,9 @@ static inline bool_t CONST cap_get_archCapIsPhysical(cap_t cap)
         return true;
 
 #ifdef CONFIG_VTX
+    case cap_vcpu_cap:
+        return true;
+
     case cap_ept_pt_cap:
         return true;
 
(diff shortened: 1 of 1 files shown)


matneutrality:
> What are the practical consequences of this change? That might warrant a changelog entry.
> 
> Reading the code a bit, it seems that without this fix `sameRegionAs`, `isMDBParentOf` and `ensureNoChildren` all fail. Which means that revoking an untyped will fail for VCPUs and all other objects allocated before that from the same UT region. That seems quite catastrophic.

Yep, that, and it has catastrophic consequences on object creation too.

Indanz:
> Yep, that, and it has catastrophic consequences on object creation too.

Could you elaborate? That's the one case I would expect it to work, some people are using this, so I'd expect them to run into issues if even object creation has problems.

matneutrality:
> > Yep, that, and it has catastrophic consequences on object creation too.
> 
> Could you elaborate? That's the one case I would expect it to work, some people are using this, so I'd expect them to run into issues if even object creation has problems.

There is a security angle here, @lsf37 has the details.

include/arch/x86/arch/object/structures.h
```diff
@@ -112,7 +112,7 @@ compile_assert(vtd_pt_size_sane, VTD_PT_INDEX_BITS + VTD_PTE_SIZE_BITS == seL4_I
 #define EPT_PT_PTR(r)    ((ept_pte_t *)(r))
 #define EPT_PT_REF(p)    ((word_t)(p))
 
-#define VCPU_PTR(r)       ((vcpu_t *)(r))
+#define VCPU_PTR(r)       ((struct vcpu *)(r))
```
lsf37, line 115: Why are we changing this declaration?

lsf37:
Thanks for putting up the PR, @Ivan-Velickovic was about to put up one up, but this is easier. The hard bit is actually the changelog entry, but I did some more research on that yesterday, so I think I have enough data now.

The bug is indeed catastrophic: it means that any user with an untyped cap can crash the kernel in VT-x configurations (which are unverified, but still). It can be triggered just with retype if you do things in the right (or wrong..) order. It's been there since the introduction of VT-x support in seL4 in 2016 (released in version 4.0.0).

CAmkES systems with x86 VMM do not trigger it, but the Microkit loader did trigger it. Not sure if the new Rust capDL loader also does, but in any case, it needs to be fixed.

include/arch/x86/arch/object/structures.h
matneutrality: If I recall well I tried to include `vcpu.h` but the compiler screamed at me so that was a cheap workaround.

lsf37: It seems to build for me without this change, which makes sense -- `structures.h` is already (transitively) included almost everywhere, so you should not need to include `vcpu.h` anywhere new.

lsf37: Sorry, never mind, I reverted to much, I see the problem now. I hate the preprocessor.

lsf37: Alright, we're already doing the same trick on Arm, and rearranging things so that `vcpu_t` becomes visible in `structures.h` is super annoying (I had thought we needed it the other way around). So let's keep this as you have it.

Indanz: Sorry, should have mentioned why this is as it is when reviewing to save you the trouble.

PR #1549, opened 2025-11-06
x86_64: Fix boot code error reporting

matneutrality:
Writing too quickly to a UART risks loosing bytes when its internal FIFO fills up. Also lines should be terminated with "\r\n" for compatibility with 1975 hardware.

### DIFF
--- a/src/arch/x86/64/head.S
+++ b/src/arch/x86/64/head.S
@@ -33,8 +33,13 @@
 .code32
 
 BEGIN_FUNC(print_string)
-    movw $0x3f8, %dx
 1:
+    movw $0x3f8+5, %dx
+2:
+    inb  (%dx), %al
+    test $0x20, %al
+    jz   2b
+    sub  $5, %dx
     movb (%ebx), %al
     outb %al, %dx
     incl %ebx
@@ -63,7 +68,7 @@ BEGIN_FUNC(huge_page_check)
     call hang
 
 huge_page_error_string:
-    .string "Huge page not supported by the processor"
+    .string "Huge page not supported by the processor\r\n"
     .set huge_page_error_size, . - huge_page_error_string
 END_FUNC(huge_page_check)
 #endif /* CONFIG_HUGE_PAGE */
@@ -137,7 +142,7 @@ BEGIN_FUNC(pcid_check)
     call hang
 
 pcid_error_string:
-    .string "PCIDs not supported by the processor"
+    .string "PCIDs not supported by the processor\r\n"
     .set pcid_error_size, . - pcid_error_string
 END_FUNC(pcid_check)
 
@@ -155,7 +160,7 @@ BEGIN_FUNC(invpcid_check)
     call  hang
 
 invpcid_error_string:
-    .string "INVPCID instruction not supported by the processor"
+    .string "INVPCID instruction not supported by the processor\r\n"
     .set invpcid_error_size, . - invpcid_error_string
 END_FUNC(invpcid_check)
 #endif /* CONFIG_SUPPORT_PCID */
@@ -174,7 +179,7 @@ BEGIN_FUNC(syscall_check)
     call  hang
 
 syscall_error_string:
-    .string "SYSCALL/SYSRET instruction not supported by the processor"
+    .string "SYSCALL/SYSRET instruction not supported by the processor\r\n"
     .set syscall_error_size, . - syscall_error_string
 END_FUNC(syscall_check)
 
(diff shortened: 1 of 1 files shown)


Indanz:
I guess you're the first to test it on real UARTs instead of QEMU.

matneutrality:
> I guess you're the first to test it on real UARTs instead of QEMU.

That thought has crossed my mind quite often lately.

Indanz:
Actually, the CI system is using real UARTs, so it has been used. Maybe that hardware just has bigger UART buffers?

(I say real, but probably UART to USB, so who knows what speed they are. Maybe CI logs tell us.)

axel-h:
Nice catch. 

Changes look ok. If it's just the few strings, adding "\r\n" there should to keep thing simple. Otherwise the UART output should add the "\r" when encountering a "\n", that is how all the other dirvers work, and the strings just have the "\n". Or change the function to be "println_string", so it appends the "\r\n" automatically and things are done where they matter.

> compatibility with 1975 hardware

This is not really the reason. It's well defined for terminal output what "\r" and "\n" do, and also what you can do with that.

matneutrality:
> Changes look ok. If it's just the few strings, adding "\r\n" there should to keep thing simple. Otherwise the UART output should add the "\r" when encountering a "\n", that is how all the other dirvers work, and the strings just have the "\n". Or change the function to be "println_string", so it appends the "\r\n" automatically and things are done where they matter.

That'd be conceptually cleaner yes, but TBH the code change would probably be larger than the few additional newlines, something like this (untested):

```asm
BEGIN_FUNC(uart_wait_ready)
    add  $5, %dx
1:
    inb  (%dx), %al
    test $0x20, %al
    jz   1b
    sub  $5, %dx
    ret
END_FUNC(uart_wait_ready)

BEGIN_FUNC(printnl_string)
    movw $0x3f8, %dx
1:
    call uart_wait_ready
    movb (%ebx), %al
    outb %al, %dx
    incl %ebx
    decl %ecx
    jnz  1b

    call uart_wait_ready
    movb $0x0a, %al
    outb %al, %dx

    call uart_wait_ready
    movb $0x0d, %al
    outb %al, %dx
    ret
END_FUNC(printnl_string)
```

I don't think it it would be better in this particular case but if you insist I could give it a go.

> > compatibility with 1975 hardware
> 
> This is not really the reason. It's well defined for terminal output what "\r" and "\n" do, and also what you can do with that.

Sure, I guess I'm more than a bit bothered that we've inherited the need for **two** characters to indicate newlines from ancient typewriters and teletype systems. But yeah let's not open that box please :)

axel-h:
split would be more like this (code untested). And you want to print CR-LF (0x0d 0x0a) and not LF-CR (which is what some of our broken UART driver unfortunately do. Reminds me to check is this has been fixed everywhere....)

```asm
BEGIN_FUNC(uart_putchar)
    movw $0x3f8+5, %dx
1:
    push %rax
    inb  (%dx), %al
    test $0x20, %al
    jz   1b
    pop %rax
    sub  $5, %dx
    outb %al, %dx
    ret
END_FUNC(uart_putchar)

BEGIN_FUNC(println_string)
1:
    movb (%ebx), %al
    call uart_putchar
    incl %ebx
    decl %ecx
    jnz  1b
    movb $0x0d, %al /* CR */
    call uart_putchar
    movb $0x0a, %al /* LF */
    call uart_putchar
    ret
END_FUNC(println_string)
```

matneutrality:
> split would be more like this (code untested). And you want to print CR-LF (0x0d 0x0a) and not LF-CR (which is what some of our broken UART driver unfortunately do. Reminds me to check is this has been fixed everywhere....)
> 
> ```assembly
> BEGIN_FUNC(uart_putchar)
>     movw $0x3f8+5, %dx
> 1:
>     push %rax
>     inb  (%dx), %al
>     test $0x20, %al
>     jz   1b
>     pop %rax
>     sub  $5, %dx
>     outb %al, %dx
>     ret
> END_FUNC(uart_putchar)
> 
> BEGIN_FUNC(println_string)
> 1:
>     movb (%ebx), %al
>     call uart_putchar
>     incl %ebx
>     decl %ecx
>     jnz  1b
>     movb $0x0d, %al /* CR */
>     call uart_putchar
>     movb $0x0a, %al /* LF */
>     call uart_putchar
>     ret
> END_FUNC(println_string)
> ```

Thanks, this is indeed cleaner. I'll try to find some time tomorrow to have a go at it.

Indanz:
Please, that function is only called from assembly in that one file, keep it simple.

(That is, what you had was fine, all the rest is bloatware.)

matneutrality:
Simple & effective versus proper & elegant. Both have their merits. Let me know how you want it :)

Indanz:
Once the compile errors are fixed, can you bring this branch up-to-date? Then after the tests have ran I can merge it.

matneutrality:
I don't understand why CI fails. The failing tests seem to be ARM/RISC-V builds and I doubt these x86 changes can affect them.

Indanz:
> I don't understand why CI fails. The failing tests seem to be ARM/RISC-V builds and I doubt these x86 changes can affect them.

I don't see any build failures, only HW test failures, but those were unrelated to your changes (console down etc.).

PR #1550, opened 2025-11-06
changelog: log vtx defect

lsf37:
The bug itself has been fixed in #1548


lsf37:
I would actually like to credit @matneutrality with finding and fixing this. @matneutrality would the following be fine for you in the Changelog entry?

```
Thanks to Mathieu Mirmont from Neutrality for reporting and fixing this defect.
```

lsf37:
Acknowledgement now added.

PR #1551, opened 2025-11-10
tools: support multiple methods with the same id

Indanz:
Useful to implement multiple API definitions for syscalls with optional arguments.

Useful for e.g. PR #1547 and PR #889 or any future API expansion with optional arguments added to existing syscalls.

### DIFF
--- a/tools/invocation_header_gen.py
+++ b/tools/invocation_header_gen.py
@@ -147,10 +147,18 @@ def parse_xml(xml_file):
         sys.exit(-1)
 
     invocation_labels = []
+    invocation_ids = set()
     for method in doc.getElementsByTagName("method"):
-        invocation_labels.append((str(method.getAttribute("id")),
-                                  str(condition_to_cpp(method.getElementsByTagName("condition")))))
-
+        label = str(method.getAttribute("id"))
+        condition = str(condition_to_cpp(method.getElementsByTagName("condition")))
+        uid = label + condition
+        dup = method.getAttribute("duplicate")
+        if uid not in invocation_ids:
+            invocation_ids.add(uid)
+            invocation_labels.append((label, condition))
+        elif dup not in ("true", "1"):
+            print("Error: Implicit duplicate id '%s' in xml file" % label, file=sys.stderr)
+            sys.exit(-1)
     return invocation_labels
 
 
--- a/libsel4/tools/sel4_idl.xsd
+++ b/libsel4/tools/sel4_idl.xsd
@@ -40,6 +40,7 @@
         <xsd:attribute name="name" type="xsd:string" use="required" />
         <xsd:attribute name="manual_name" type="xsd:string" use="optional" />
         <xsd:attribute name="cap_description" type="xsd:string" use="optional" />
+        <xsd:attribute name="duplicate" type="xsd:boolean" use="optional" />
     </xsd:complexType>
 
 
--- a/tools/invocation_json_gen.py
+++ b/tools/invocation_json_gen.py
@@ -32,7 +32,7 @@ def xml_to_json_invocations(xml, gen_config, counter, invocations_dict, ):
     for method in xml.getElementsByTagName("method"):
         label = str(method.getAttribute("id"))
         exists = condition_to_bool(method.getElementsByTagName("condition"), gen_config)
-        if exists:
+        if exists and label not in invocations_dict:
             invocations_dict[label] = counter
             counter += 1
 


tools/invocation_header_gen.py
```diff
@@ -147,9 +147,13 @@ def parse_xml(xml_file):
         sys.exit(-1)
 
     invocation_labels = []
+    invocation_ids = set()
     for method in doc.getElementsByTagName("method"):
-        invocation_labels.append((str(method.getAttribute("id")),
-                                  str(condition_to_cpp(method.getElementsByTagName("condition")))))
+        mid = str(method.getAttribute("id"))
```
midnightveil: I think naming this `method_id` would help a bit...

`mid` reads like "middle" to me.

tools/invocation_header_gen.py
```diff
@@ -147,9 +147,13 @@ def parse_xml(xml_file):
...
+        mc = str(condition_to_cpp(method.getElementsByTagName("condition")))
+        if mid+mc not in invocation_ids:
+            invocation_ids.add(mid+mc)
+            invocation_labels.append((mid, mc))
```
midnightveil: I think the correctness of adding multiple of the same label ("id") is non-obvious here and it would be worth it to explain it more...

You somewhat described it in 1547, but I don't understand how exactly duplicate enum values would be OK, precisely?

Like

```
enum invocation_label {
    InvalidInvocation,
    {%- for label, condition in invocations %}
    {%- if condition %}
#if {{condition}}
    {%- endif %}
    {{label}},
    {%- if condition %}
#endif
    {%- endif %}
    {%- endfor %}
    nInvocationLabels
};
```

would produce an enum like?

```
enum invocation_label {
   Invalid,
   LabelOne,
   LabelOne,
   // etc
}
```

I don't understand how that works, unless the conditions are disjoint?

midnightveil (changes requested):

tools/invocation_header_gen.py
Indanz: It made perfect sense at the time! (But I had the same first impression when reading it again.) What about `idc`? ;-)

tools/invocation_header_gen.py
Indanz: I can't remember whether I needed to add that to make it work, or that it fixes a theoretical problem. I suspect it's the former, but not sure.

It handles the case where there are multiple labels with the same name, but different conditions. Of course, if #1542 gets implemented, this won't be needed any more.

Example:

```
enum invocation_label {
   Invalid,
#if A
   LabelOne,
#endif
#if B
   LabelOne,
#endif
   // etc
}
```

tools/invocation_header_gen.py
midnightveil: What is that supposed to mean? id condition, I guess?

The other alternative is ae use the same naming as in the iteration in the Jinja templates: label and condition.

Indanz: That's what I ended up doing, and uid for the combination of them.

tools/invocation_header_gen.py
midnightveil: Right, I think when I initially read this I was confused.

Of course there won't be duplicate IDs for the same config because the existing XML don't have duplicates.

This just coalesces duplicates in the enum together (that have the same condition): which is exactly the *point* of this PR.

We won't have duplicates from conflicting configs in the output because if we did there would be an issue right now.

---

with #1542 that then allows by removing the conditions to unify invocations across all configs, which does require unique for all configd.

---

This PR would then remove any issues with "duplicate" syscall ID definitions by accident, but I don't think that's super easy to work around since we do also want to allow it.

midnightveil (approved):
I don't know why CI was failing but I understand what this is doing now.

I'll see if this rerun fails. I suspect we should also do (at least) a hw-build before merging.

tools/invocation_header_gen.py
Indanz: If they use the same label and it works because it's on different objects by accident, then that would be unfortunate and undetected. However, that's also a use case we would want to allow though, but only if it's done deliberately.

Maybe we could add a tag to indicate duplication is deliberate, but not sure what the best way would be.

Indanz:
Okay, I changed it so to get duplicate ids, you have to add an attribute `duplicate="true"` to make it explicit. It's not perfect as it doesn't catch the case where one without that attribute gets added first and a duplicate marked as such is parsed later, but so be it.

I also fixed the generated JSON, duplicates incremented the ID before, now they're ignored.

Now if I add a dummy:

```xml
        <method id="TCBResume" duplicate="true" name="Resume2" manual_label="tcb_resume2">
            <brief>
                Resume a thread, test duplicate IDs
            </brief>
        </method>
```

The `diff -pru before-kernel/ kernel/` shows no changes and `diff -pru before-libsel4/ libsel4/` shows the expected:

```diff
diff -pru before-libsel4/include/interfaces/sel4_client.h libsel4/include/interfaces/sel4_client.h
--- before-libsel4/include/interfaces/sel4_client.h	2026-06-23 14:01:57.638651520 +0100
+++ libsel4/include/interfaces/sel4_client.h	2026-06-23 14:09:06.547780426 +0100
@@ -3494,6 +3494,50 @@ seL4_TCB_Resume(seL4_TCB _service)
 }
 
 /**
+ * @xmlonly <manual name="Resume2" label="tcb_resume2"/> @endxmlonly
+ * @brief @xmlonly Resume a thread, test duplicate IDs @endxmlonly
+ * @param[in] _service Capability to the TCB which is being operated on.
+ * @return @xmlonly <errorenumdesc/> @endxmlonly
+ */
+LIBSEL4_INLINE seL4_Error
+seL4_TCB_Resume2(seL4_TCB _service)
+{
+	seL4_Error result;
+	seL4_MessageInfo_t tag = seL4_MessageInfo_new(TCBResume, 0, 0, 0);
+	seL4_MessageInfo_t output_tag;
+	seL4_Word mr0;
+	seL4_Word mr1;
+	seL4_Word mr2;
+	seL4_Word mr3;
+
+	/* Marshal and initialise parameters. */
+	mr0 = 0;
+	mr1 = 0;
+	mr2 = 0;
+	mr3 = 0;
+
+	/* Perform the call, passing in-register arguments directly. */
+	output_tag = seL4_CallWithMRs(_service, tag,
+		&mr0, &mr1, &mr2, &mr3);
+	result = (seL4_Error) seL4_MessageInfo_get_label(output_tag);
+
+	/* Unmarshal registers into IPC buffer on error. */
+	if (result != seL4_NoError) {
+		seL4_SetMR(0, mr0);
+		seL4_SetMR(1, mr1);
+		seL4_SetMR(2, mr2);
+		seL4_SetMR(3, mr3);
+#ifdef CONFIG_KERNEL_INVOCATION_REPORT_ERROR_IPC
+		if (seL4_CanPrintError()) {
+			seL4_DebugPutString(seL4_GetDebugError());
+		}
+#endif
+	}
+
+	return result;
+}
+
+/**
  * @xmlonly <manual name="Bind Notification" label="tcb_bindnotification"/> @endxmlonly
  * @brief @xmlonly Binds a notification object to a <obj name="TCB"/> @endxmlonly
  * 
```

Issue #1541, opened 2025-10-31
plat_SGITargetValid on GICv3 is nonsensical


midnightveil:
https://github.com/seL4/seL4/blob/3977e12b8b19afd1cb88322cb7e3747b94fc10df/src/arch/arm/machine/gic_v3.c#L360-L363

On GICv3, a valid SGI target is really just a valid MPIDR value. This check effectively limits the number of valid SGI targets to 16, which means that you can only target MPIDRs between 0x0 and 0x10. This is somewhat limited, especially when any boards with multiple affinity clusters will have much higher bits set.

Indanz:
Yes, that's wrong and should just return always true on GICv3. The only MPIDR value the kernel knows about is its own, but sending SGIs to itself is useful for testing, so we don't want to limit that either.

midnightveil:
Closing this.

PR #1552, opened 2025-11-13
SGI, GICv3: Fix plat_SGITargetValid

Indanz:
Otherwise it is not possible to send SGIs to all cores on platforms with non-contiguous MPIDR values or more than 16 cores.

16 is the limit of the target list, which is only relevant when sending an SGI to multiple targets at once.
    
Update the API documentation to reflect what's actually happening.

This fixes issue #1541.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 * Change fault based FPU context switching to a TCB flag based approach:
   New system call `seL4_TCB_SetFlags` and new flag `seL4_TCBFlag_fpuDisabled`.
   See [RFC-18](https://sel4.github.io/rfcs/implemented/0180-fpu-switching.html).
+* Multiple SMP related bugs fixed.
 
 ### Platforms
 


src/arch/arm/machine/gic_v3.c
```diff
@@ -359,7 +359,8 @@ BOOT_CODE void cpu_initLocalIRQController(void)
 
 bool_t plat_SGITargetValid(word_t target)
 {
-    return target < GIC_SGI_NUM_TARGETS;
+    /* MPIDR values are not guaranteed to be contiguous */
+    return target <= UINT32_MAX;
```
midnightveil, line 364: This forces the Aff3 value to be 0.

The encoding seL4 uses it the MPIDR_AFF_MASK which matches the layout of the MPIDR value.

Note that original there was Aff2,Aff1,Aff0 in the MPIDR but it was extended with an Aff3 value in the higher 32 bits.

Look at `MPIDR_AFF3` specifically which is bits 32-40.

We could either enforce that

- `(target & ~MPDIR_AFF_MASK) == 0` (no non-aff bits set)
-  allow arbitrary MPIDR values (which include other bits such as `U` or `MT`); we are extracting only the bits we need from it when writing to the GICv3 so this is also fine. I'd prefer this.

Alternatively we could force userspace into this 32-bit packing of the 4 mpidr values, but if so we need to change how we generate the MPIDR value we write to the GIC to match the hardware layout, this seems like a bad idea to me.

Indanz: The target as given by the user is a concatenation of Aff0+Aff1+Aff2+Aff3, not the MPIDR register value directly, see `sgir_word_from_args()`.

I thought to use MPIDR as a shorthand for Aff0+Aff1+Aff2+Aff3, but that clearly doesn't work. I'll fix the misleading comment...

Indanz: I'll fix the XML description too.

libsel4/arch_include/arm/interfaces/object-api-arch.xml
```diff
@@ -800,10 +800,10 @@
...
             </description>
             <param dir="in" name="irq" type="seL4_Word" description="The SGI INTID (0-15) that can be signalled."/>
-            <param dir="in" name="target" type="seL4_Word" description="The node ID that will be
-            targeted. 0-7 for GICv2 and 0-31 for GICv3. Targets within that range that are not
-            supported by the hardware will be ignored. For example, on a GICv3 board with 4 CPUs, the
-            capability for target 13 can be created, but signals to it will have no effect."/>
+            <param dir="in" name="target" type="seL4_Word" description="The node ID that will be targeted.
+            0-7 for GICv2 and the affinity value for GICv3 (concatenation of Aff0+Aff1+Aff2+Aff3 from MPIDR).
+            Targets within that range that are not supported by the hardware will be ignored. For example, on a GICv2
+            board with 4 CPUs, the capability for target 5 can be created, but signals to it will have no effect."/>
```
Indanz, line 806: Or should I say "concatenation of Aff3+Aff2+Aff1+Aff0" instead? That would follow big-endian hex notation.

src/arch/arm/machine/gic_v3.c
Indanz: > if so we need to change how we generate the MPIDR value we write to the GIC to match the hardware layout, this seems like a bad idea to me.

On sane hardware the packed version would go from 0 to the maximum core number, so it's nice and easy to work with (basically logical id == target id). If using insane hardware, you get pain either way.

midnightveil: ARM tells implementors that the Affinity values should be used as such? So if you have 2 clusters of say 64 CPUs in a NUMA-ish configuration you'll get non-contiguous values (0-6 of the 8 aff0 bits, and 1 of the aff1 bits). That doesn't seem silly to me.

It is true that a lot of the simple 4 core boards we have do ranges like 0,0x100,0x200,0x300 but that doesn't mean the structure itself is silly.

It's not like a 64-bit argument is particularly difficult, and I seem to recall the Aff3 is hard coded to 0 for 32-bit systems anyway (I may be misremembering the spec here, Aff3 was a more recent addition to MPIDR).

I have seen a fair amount of ARM SMP code that assumes that e.g. PSCI calls take indices 0-n and not MPIDRs, hoping that it's contiguous is just not true and packing them provides... next to zero benefit afaict?

Instead of being able to just pass a (maybe masked) MPIDR of another core you instead have to bitshuffle them around.

Indanz: Well, it says:
```
GICv3 uses affinity routing to identify connected PEs and to route interrupts to a specific PE or
group of PEs. The affinity of a PE is represented as four 8-bit fields:
<affinity level 3>.<affinity level 2>.<affinity level 1>.<affinity level 0>
```
So using affinity makes sense for interrupt routing which we do for SGIs.

However, you are right that expecting contiguous values is unrealistic for GICv3, as that doesn't seem to be a goal at all for Arm, so I probably was too optimistic about that and perhaps using MPIDR  directly would be more practical for the user. But shifting a non-zero Aff3 down isn't the end of the world either though.

midnightveil:
Yeah, I'm fine with it. Not my preferred API but that's a preference, it should work.

lsf37:
This revealed an incompleteness of the preprocess test (which should have failed here, see #1557). I'll need to update the proofs.

I'll likely want to change the comparison for GICv2 into a `<=` as well so I don't have to deal with a different word proof for different platforms. Using `<` is not an option for GICv3, because that would produce an overflow on 32 bit platforms.

Indanz:
> I'll likely want to change the comparison for GICv2 into a `<=` as well so I don't have to deal with a different word proof for different platforms. Using `<` is not an option for GICv3, because that would produce an overflow on 32 bit platforms.

My fault, I should have updated the GICv2 version to use `<=` as well and retained `GIC_SGI_NUM_TARGETS`, although maybe renamed to `GIC_SGI_MAX_TARGET` or something. But I was too focussed on minimising the change.

lsf37:
No worries, the preprocess check should have alerted us that it might break the proof, then we'd have seen this earlier. All done now on the code side, I just still need to update the preprocess check

PR #1553, opened 2025-11-14
Add support for rock3b

JDuchniewicz:
Add support for Radxa rock3b board based on rk3568 SoC.

Test with: https://github.com/seL4/seL4_tools/pull/241, https://github.com/seL4/util_libs/pull/204


Indanz (approved):
Seems straightforward.

Have you ran sel4test and sel4bench for all configurations? (SMP off and on, MCS on and off.)

Indanz:
Are you planning on adding the rk3568 to CI?

Indanz:
Is kernel debug printing working for you? You probably need to add your UART's compat string to tools/hardware.yml and the driver if it's missing (and also add a driver to the kernel if it's a new one).

JDuchniewicz:
> Are you planning on adding the rk3568 to CI?

Yes, the board will be added to the testing equipment rack.

JDuchniewicz:
@Indanz - the kernel debug printing is working. Ran all configurations of sel4test and all configurations of sel4bench (running that allowed me to catch a bug in periodic mode setting which was not caught anywhere else: sel4test/sddf). All is passing. I uncovered a new issue which reproduces on other boards when enabling debug logs: https://github.com/seL4/sel4bench/issues/60


As a sidenote - SMP benchmark takes a long time to finish and I initially thought it hanged. Maybe there should be some README's etc with median run times or something similar to guide people unfamiliar with how long they take to run.

lsf37:
> As a sidenote - SMP benchmark takes a long time to finish and I initially thought it hanged. Maybe there should be some README's etc with median run times or something similar to guide people unfamiliar with how long they take to run.

Run times change all the time, but a warning that some tests take long would make sense. You can also check the existing CI logs for run times (switching on time stamps in the log view on GitHub).

Indanz:
> As a sidenote - SMP benchmark takes a long time to finish and I initially thought it hanged. Maybe there should be some README's etc with median run times or something similar to guide people unfamiliar with how long they take to run.

As far as I remember, the only long tests is one of the SMP tests. But even then it's much faster than sel4test, which has a very slow memory test where it maps and touches all available memory.

How long does the test take for you? It shouldn't be more than a few minutes IIRC. If it takes much longer, there may be an SMP bug (or HW problem).

JDuchniewicz:
Under 10 minutes, but I was fairly sure the bench was stuck somewhere. Maybe there should be some info for the user that some tests might take a while to finish.

Indanz:
How long does sel4test take?

lsf37:
To merge the platform PRs, we also need a PR to [seL4/docs](https://github.com/seL4/docs). You need to add a new page to the `Hardware/` directory there. See e.g. cheshire.md for a recent example.

lsf37:
Oh, and an addition to https://github.com/seL4/ci-actions/blob/master/seL4-platforms/platforms.yml -- you can set `no_hw_test: true` until the board(s) are available in the machine queue. (How long until we can expect them to show up there?)

Indanz:
> To merge the platform PRs, we also need a PR to [seL4/docs](https://github.com/seL4/docs). You need to add a new page to the `Hardware/` directory there. See e.g. cheshire.md for a recent example.

Personally I'm very bad at that, apologies. I might be the reason why this is being required now. :-/

lsf37:
It's always been required, but people keep forgetting about checking it.

JDuchniewicz:
> How long does sel4test take?

Took me less than 1,5 minute @Indanz 

Also, will focus on finishing up rock3b PRs on Monday - got sidetracked by some performance regression. 

> Oh, and an addition to https://github.com/seL4/ci-actions/blob/master/seL4-platforms/platforms.yml -- you can set `no_hw_test: true` until the board(s) are available in the machine queue. (How long until we can expect them to show up there?) 

Set that up (in https://github.com/seL4/ci-actions/pull/426) - for it to appear in machine queue there are some cleanups necessary as we have run out of space for the physical boards (!) @lsf37

JDuchniewicz:
Added docs in https://github.com/seL4/docs/pull/311 @lsf37

PR #1554, opened 2025-11-18
complete SGI cap docs

lsf37:
- add SGIs to change log
- add missing error conditions in API docs
- add note on target check limitations

This documents the current state of the implementation and should be mostly independent of #1552.

Also adds a trivial update commit that all Arm platforms now have proof support.

### DIFF
--- a/libsel4/arch_include/arm/interfaces/object-api-arch.xml
+++ b/libsel4/arch_include/arm/interfaces/object-api-arch.xml
@@ -798,12 +798,19 @@
                 The resulting capability can be invoked like a notification capability that supports
                 only signal/send. SGIs can be received by IRQ notification objects on the target
                 core like other IRQs. See also <autoref label="sec:interrupts"/>.
+
+                Note that the kernel only checks architectural limits for SGI IRQ id and target. It
+                does not know whether the corresponding target core exists. Depending on hardware
+                implementation, signalling a non-existent target may create an unrecoverable SError.
+                This means it is the responsibility of the developer to not issue capabilities for
+                targets that do not exist.
             </description>
             <param dir="in" name="irq" type="seL4_Word" description="The SGI INTID (0-15) that can be signalled."/>
             <param dir="in" name="target" type="seL4_Word" description="The node ID that will be targeted.
             0-7 for GICv2 and the affinity value for GICv3 (concatenation of Aff3+Aff2+Aff1+Aff0 from MPIDR).
-            Targets within that range that are not supported by the hardware will be ignored. For example, on a GICv2
-            board with 4 CPUs, the capability for target 5 can be created, but signals to it will have no effect."/>
+            Targets within that range that are not supported by thde hardware will be not be rejected. For example,
+            on a GICv2 board with 4 CPUs, the capability for target 5 can be created. The result of later signalling
+            this target depends on the hardware implementation."/>
 
             <param dir="in" name="root" type="seL4_CNode" description="CPtr to the CNode that forms the root of the destination CSpace. Must be at a depth equivalent to the wordsize."/>
             <param dir="in" name="index" type="seL4_Word" description="CPtr to the destination slot. Resolved from the root of the destination CSpace."/>
@@ -822,6 +829,7 @@
             <error name="seL4_IllegalOperation">
                 <description>
                     The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                    Or, SGIs are not supported on this platform.
                 </description>
             </error>
             <error name="seL4_InvalidCapability">
@@ -833,6 +841,12 @@
                 <description>
                     The value of <texttt text="irq"/> or <texttt text="target"/> is out of range.
                     Or, <texttt text="depth"/> is invalid <docref>(see <autoref label="s:cspace-addressing"/>)</docref>.
+                </description>
+            </error>
+            <error name="seL4_InvalidArgument">
+                <description>
+                    The SGI target is not supported on this GIC. Note that this only checks
+                    architectural limits, not the presence of the target core on the current board.
                 </description>
             </error>
         </method>


CHANGES.md
```diff
@@ -59,6 +59,19 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
+* Support SGIs (software generated interrupts) on platforms with GICv2 or GICv3 in non-SMP configurations. SGIs are
+  intended for signalling other cores when seL4 is used in a multi-kernel setup. On Arm, this is implemented with a new
+  capability `SGISignal`, which can be created from `IRQControl` capabilities with the new IRQ control invocation
+  `ARMIRQIssueSGISignal` for a specific IRQ and target core. The resulting `SGISignal` capability can be invoked like a
+  notification capability that supports only signal/send. SGIs can be received by IRQ notification objects on the target
+  core like other IRQs.
+
+  Note that the kernel only checks architectural limits for IRQ and target. It does not know whether the corresponding
+  target core exists. Depending on hardware implementation signalling a non-existent target may create an unrecoverable
+  S-Error. This means it is the responsibility of the developer to not issue capabilities for targets that do not exist.
```
Indanz:
```suggestion
  SError. This means it is the responsibility of the developer to not issue capabilities for targets that do not exist.
```

Indanz: Also, it's only unrecoverable because seL4 panics if it gets an SError, if it just ignored it or passed the information to user space everything would be fine.

It's only an asynchronous system error because it is raised by hardware after a device memory write which can't be coupled to the responsible CPU instruction. So it would be fine to ignore. The problem is, theoretically there may exist other, actually fatal hardware errors that are also raised as SErrors which _are_ unrecoverable and the kernel can't distinguish those.

But the same is true for normal IRQs already: Some SoCs generate an IRQ on unrecoverable ECC error. So in that regard (implementation defined behaviour), SError is no different than an IRQ: Both are asynchronous notifications, potentially signalling a fatal system error.

midnightveil: Does it actually generate an SError or documented to do so?

(also, S-errors can be recoverable. newer ARM impleentations have RAS which allow certain kinds of Serrors to be recovered)

lsf37: We didn't find anything in the docs for what happens for invalid targets, so we have to assume that it might generate one. It will generate one in the kernel (if it does), and that will be unrecoverable.

lsf37: Point is, it may or may not, so if you want to be sure that it does not, you should make sure you are only using valid targets.

lsf37:
(rebased over #1552)

CHANGES.md
lsf37: None of that changes the fact that the current implementation of seL4 expects SError not to happen in kernel code. We're not going to change that for this release, so I don't see what in that warning is wrong. It is entirely up to user level to not screw this up, and people need to know about that.

lsf37:
> If you would be consistent with this level of warning, you should add the same SError warning to the Arm memory map syscall.

Where in memory map can we get an SError? For device untyped that is not mapped to memory? Excluding that is the responsibility of the board port, not the developer. I guess mapping a page that is MMIO and accessing parts of it that are not backed by the device could generate one. The SError there occurs in user code, though. Still bad, but the point is that kernel calls should not be able to produce them.

We should definitely add a warning if there are additional cases where an SError can be triggered in the kernel.

lsf37:
> I think it's worded too strongly and repeatedly

I wasn't sure about the repetition in the change log, happy to just remove it there.

> I would have left it at saying that sending an SGI to a target that doesn't exist gives implementation defined behaviour, so should be avoided.

That's not strong enough. The kernel is verified not to produce errors during kernel execution. Any exception to that needs to be documented loudly, especially if it just becomes an assumption in the formal model that depends on user-level behaviour.

> The exact behaviour may be described in all the GIC implementations (Corelink GIC n00 e.a.), there aren't too many of those, so we can check what those say instead of freaking out people unnecessarily. I'll have a look if I have time.

If we can verify what the actual behaviour is supposed to be that would of course be better. If it is a possibility, though, I do in fact want people to be freaked out about it so they are avoiding it.

lsf37:
(rebased; removed duplicate warning in CHANGES; S-Error -> SError)

CHANGES.md
```diff
@@ -86,6 +86,15 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
 #### Arm
 
+* Support SGIs (software generated interrupts) on platforms with GICv2 or GICv3 in non-SMP configurations. SGIs are
+  intended for signalling other cores when seL4 is used in a multi-kernel setup. On Arm, this is implemented with a new
+  capability `SGISignal`, which can be created from `IRQControl` capabilities with the new IRQ control invocation
+  `ARMIRQIssueSGISignal` for a specific IRQ and target core. The resulting `SGISignal` capability can be invoked like a
+  notification capability that supports only signal/send. SGIs can be received by IRQ notification objects on the target
+  core like other IRQs.
+
+  See also [RFC-17](https://sel4.github.io/rfcs/implemented/0170-multikernel-ipi-api.html)
```
Xaphiosis, line 96: Will people get confused by IPI in the target link? Probably doesn't matter

libsel4/arch_include/arm/interfaces/object-api-arch.xml
```diff
@@ -798,12 +798,19 @@
                 The resulting capability can be invoked like a notification capability that supports
                 only signal/send. SGIs can be received by IRQ notification objects on the target
                 core like other IRQs. See also <autoref label="sec:interrupts"/>.
+
+                Note that the kernel only checks architectural limits for SGI IRQ id and target. It
+                does not know whether the corresponding target core exists. Depending on hardware
+                implementation, signalling a non-existent target may create an unrecoverable SError.
```
Xaphiosis, line 804: is SError a known term in this documentation?

libsel4/arch_include/arm/interfaces/object-api-arch.xml
```diff
@@ -798,12 +798,19 @@
...
+                targets that do not exist.
             </description>
             <param dir="in" name="irq" type="seL4_Word" description="The SGI INTID (0-15) that can be signalled."/>
             <param dir="in" name="target" type="seL4_Word" description="The node ID that will be targeted.
             0-7 for GICv2 and the affinity value for GICv3 (concatenation of Aff3+Aff2+Aff1+Aff0 from MPIDR).
-            Targets within that range that are not supported by the hardware will be ignored. For example, on a GICv2
-            board with 4 CPUs, the capability for target 5 can be created, but signals to it will have no effect."/>
+            Targets within that range that are not supported by the hardware will be not be rejected. For example,
+            on a GICv2 board with 4 CPUs, the capability for target 5 can be created. The results of signalling a
+            non-existing target depend on the hardware implementation."/>
```
Xaphiosis: possible alternative: "The result ... depends on ..."

libsel4/arch_include/arm/interfaces/object-api-arch.xml
```diff
@@ -822,6 +829,7 @@
             <error name="seL4_IllegalOperation">
                 <description>
                     The <texttt text="_service"/> is a CPtr to a capability of the wrong type.
+                    Or, SGIs are not supported on this platform.
```
Xaphiosis, line 832: Would `Either ..., or SGIs ...` make more sense here?

lsf37: The `Or, ` is a thing in the documentation so far (for enumerating different classes of causes), so I didn't want to mess with that.

libsel4/arch_include/arm/interfaces/object-api-arch.xml
lsf37: It's an Arm spec term, and since it's Arm specific documentation, I think we can assume it. Happy to add things if people feel different, though.

CHANGES.md
lsf37: Not much we can do about there if we don't want to launch into an explanation of IPI/SGI etc. The RFC text does explain, though. So the confusion should only last until then (of course nobody ever clicks on the link and reads things).

lsf37:
I'll merge this for now. Happy to reword if Indan finds out that none of the existing implementations actually raises an SError.

Indanz:
> > If you would be consistent with this level of warning, you should add the same SError warning to the Arm memory map syscall.
> 
> Where in memory map can we get an SError? For device untyped that is not mapped to memory? Excluding that is the responsibility of the board port, not the developer. I guess mapping a page that is MMIO and accessing parts of it that are not backed by the device could generate one.

E.g. i.MX GPIO peripherals generate a bus error when you write to a read-only register. And that arrives as an SError at the CPU level, even though the write is done via uncached memory.

> The SError there occurs in user code, though. Still bad, but the point is that kernel calls should not be able to produce them.

Currently the kernel handles them equally and halts on both, except if `CONFIG_AARCH64_SERROR_IGNORE` is set.

But yes, agreed that the kernel shouldn't generate them, ideally.

> We should definitely add a warning if there are additional cases where an SError can be triggered in the kernel.

SError's are asynchronous in nature, usually caused by writes. E.g. if the GPIO read-only register was written via cached memory, the SError can happen any time when the cache line gets evicted. That can happen in kernel mode too. For that reason you can't really treat SErrors arriving in kernel mode differently than the ones arriving in user mode.

Any peripheral can generate it in theory, that's why halting by default is too strict, because usually it's just some peripheral being difficult. Only exception is if the error is generated by the memory controller or some other peripheral that affects memory integrity. But e.g. ECC unrecoverable errors can arrive via normal IRQs too (at least they do on i.MX). This is a very badly documented area.

That's why I added the SError kernel behaviour user config option in the first place and why others disabled the halt for the tx platform. Sadly the SError config option is broken and I don't know how to fix it. It was made a platform config setting only instead of user configurable, as originally intended.

PR #1559, opened 2025-11-19
arm,gic_v2: minor rephrasing for verification

lsf37:
The proofs are shared between GICv2 and GICv3, which means we want the term structure to be the same for key functions that are visible to the proofs (but constants are Ok to change).

~(on draft while I'm updating the proofs)~
Now ready for review.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,15 +43,41 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 ### Changes
 
 * Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
-* riscv: Change default cmake options KernelRiscvExtF and KernelRiscvExtD from OFF to ON.
-  Except for RISCV32 with LLVM clang enabled will default both to OFF.
+* Added: build exports an `invocations_all.json` file with syscall invocation numbers.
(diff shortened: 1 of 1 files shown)


src/arch/arm/machine/gic_v2.c
```diff
@@ -177,7 +177,8 @@ BOOT_CODE void cpu_initLocalIRQController(void)
 
 bool_t plat_SGITargetValid(word_t target)
 {
-    return target < GIC_SGI_NUM_TARGETS;
+    /* written as <= so that the term is the same as in gic_v3 for the proofs */
+    return target <= GIC_SGI_NUM_TARGETS - 1;
```
midnightveil, line 181: Given that `GIC_SGI_NUM_TARGETS` only exists for GICv2 now, I think it makes more sense for it to be `GIC_SGI_MAX_TARGET 7` instead.

lsf37: I had the same impulse, but to do that properly, I'd have to update the name also in the specs and proofs and that's a lot more than the one line. I don't think it's misleading to keep it as NUM_TARGETS, so the minus one doesn't really hurt.

midnightveil: OK

PR #1562, opened 2025-11-20
Disable clock sync test on QEMU RISC-V

Ivan-Velickovic:
This will sometimes fail when using QEMU RISC-V virt with multiple cores, make the behaviour consistent as QEMU ARM virt.

Given that QEMU is not intended to be a cycle accurate simulator, I imagine that is why this does not reliably succeed.


Indanz:
> It consistently fails on some real boards as well, but that is something for after the release.

Which boards are that?

RISC-V already has 6 us margin, and the specification guarantees maximum of 1 us difference, so more than 5 us overhead to call into M-mode is ridiculous.

PR #1563, opened 2025-11-21
Fix TCB size for SMP + benchmark config

Ivan-Velickovic:
In the definition of tcb_t, there's extra fields if there's SMP or BENCHMARK_TRACK_UTILISATION, looks like having both on is not enough for 11 bits.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,7 +42,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 ### Changes
 
-* Added `zynqmp` and `rpi4` to the set of verified AArch64 configs.
+* All Arm platforms supported by seL4 now have a verified configuration.
 * Added: build exports an `invocations_all.json` file with syscall invocation numbers.
 * Added option for suppressing generation of config output files in build process
 * Added `--skip-unchanged` option to build process to prevent time stamp update for config output files when these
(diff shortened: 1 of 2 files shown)


Ivan-Velickovic:
I only ran into this on AArch64, but I only was building this kind of configuration of the kernel in AArch64 and RISC-V 64-bit.

Ivan-Velickovic:
Yes, but I'm wondering if I need a `CONFIG_KERNEL_MCS` in the condition as well, since Microkit always uses MCS. I'll play around with sel4test to figure out what exact combination trips the assert.

Indanz (approved):

midnightveil:
I can reproduce this without MCS:

```
cmake -GNinja -DKernelArmHypervisorSupport=ON -DKernelBenchmarks=track_utilisation -DKernelDebugBuild=OFF -DKernelMaxNumNodes=4 -DKernelPlatform=qemu-arm-virt -DKernelSel4Arch=aarch64 -DKernelVerificationBuild=OFF -DQEMU_MEMORY=2048 -S /home/julia/ts/seL4 -B . -DCROSS_COMPILER_PREFIX=aarch64-none-elf-
```

midnightveil:
But actually, only when `KernelVerificationBuild=OFF` is specified: (well, duh, because without that it ignores the benchmarks being turned on). Hypervisor OFF also makes it go away, same as SMP.

```
cmake -GNinja -DKernelArmHypervisorSupport=ON -DKernelBenchmarks=track_utilisation -DKernelDebugBuild=OFF -DKernelMaxNumNodes=4 -DKernelPlatform=qemu-arm-virt -DKernelSel4Arch=aarch64 -DQEMU_MEMORY=2048 -S /home/julia/ts/seL4 -B . -DCROSS_COMPILER_PREFIX=aarch64-none-elf-
```

lsf37:
Thanks for checking. It sounds like we can merge this when the tests have passed.

Ivan-Velickovic:
> Thanks for checking. It sounds like we can merge this when the tests have passed.

Yep, it looks like hypervisor support being on as well is what triggers it. I think we've got the right combination of config options now.

PR #1567, opened 2025-12-02
libsel4: fix incorrect seL4_VMEnter() description

dreamliner787-9:
`SEL4_VMENTER_CALL_CONTROL_ENTRY_MR` refers to the VM Entry Interruption-Information Field of the VMCS [1]. Not the VM-Entry Controls, which are two different things.

I was tempted to change it to something like `SEL4_VMENTER_CALL_INTERRUPT_INFO_MR` but that would be a breaking change so I have not. Could I please have some inputs on this?

[1]: [src/arch/x86/object/vcpu.c: vcpu_update_state_sysvmenter()](https://github.com/seL4/seL4/blob/4f7d7b7376a180183ae6de3b2c5fe620260a71c2/src/arch/x86/object/vcpu.c#L1075)

### DIFF
--- a/src/arch/x86/object/vcpu.c
+++ b/src/arch/x86/object/vcpu.c
@@ -1092,7 +1092,8 @@ void vcpu_sysvmenter_reply_to_user(tcb_t *tcb)
     setMR(tcb, buffer, SEL4_VMENTER_CALL_EIP_MR, vmread(VMX_GUEST_RIP));
     setMR(tcb, buffer, SEL4_VMENTER_CALL_CONTROL_PPC_MR, vmread(VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS));
 
-    setMR(tcb, buffer, SEL4_VMENTER_CALL_CONTROL_ENTRY_MR, vmread(VMX_CONTROL_ENTRY_INTERRUPTION_INFO));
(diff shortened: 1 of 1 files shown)


Indanz:
I propose to change it to `SEL4_VMENTER_CALL_INTERRUPT_INFO_MR` like you suggest and also add a backward compatible `SEL4_VMENTER_CALL_CONTROL_ENTRY_MR` which triggers a deprecation warning. Because the current name is just too misleading.

Probably with `SEL4_DEPRECATE_MACRO()` in libsel4/include/sel4/deprecated.h.

lsf37:
Adding an alias + deprecation warning sounds like a good solution to me.

Indanz (approved):

Indanz:
Can you bring your branch up-to-date so I can merge it?

dreamliner787-9:
Yep all done, I've also fixed the style issue so hopefully the CI will pass now. Sorry it took a while as I was caught up in other things and forgot about this.

Indanz:
> Yep all done, I've also fixed the style issue so hopefully the CI will pass now. Sorry it took a while as I was caught up in other things and forgot about this.

No Problem! It's just that I can't bring your tree up-to-date as I don't have the permissions as the repo is in a different org.

Issue #1569, opened 2025-12-10
Performance degradation after #1325 (FPU Switching)


JDuchniewicz:
In our performance graphs we just recently realized that the performance benchmarks (for ARM and x86 platforms) report too high cycle counts after a certain point of time (#1325 merge). 

Although proper disabling of FPU in the server for the IPC benchmarks [#IPC](https://github.com/seL4/sel4bench/pull/62) reduced the cycle counts, a residual amount of overhead (40 cycles on Armv8a A57/Tx1/Jetson platform) remained.

After deeper investigation I found that on ARM the culprit were the `isb` calls and that the `enable/disableFpu` calls were done frequently. Removing these `isb`s reduced the overhead to just around 4-13 cycles that we can attribute to setting up the special registers.

Further, following @KurtWu10's reasoning it turns out that the number of `isb`s can be further reduced but it would require some platform-specific macros to be hoisted to platform-agnostic code in https://github.com/seL4/seL4/blob/2bef61b9da4c200ada22efdc5d42461576cab7a7/src/machine/fpu.c#L15

I am pasting the table with my measurements done using sel4bench checked out at 3 points in time (before #1325/after and the latest master on the sel4bench manifest (2bef61b):

### For same vspace: `true`
| Version   | Before FPU changes<br>(FPU disabled) - db233b | No FPU disabled - 22f816a | FPU disabled - 22f816a | FPU client only - 22f816a | FPU client only - isb[1] - 2bef61b | FPU client only - isb[2] - 2bef61b |
| --------- | --------------------------------------------- | ------------------------- | ---------------------- | ------------------------- | -------------------------- | -------------------------- |
| Call      | 413(10)                                       | 639(0)                    | 443(0)                 | 443(0)                    | 417(0)                     | 417(0)                     |
| ReplyRecv | 410(2.9)                                      | 637(0)                    | 450(0)                 | 448(0)                    | 422(0)                     | 422(0)                     |

### For same vspace: `false`
| Version   | Before FPU changes<br>(FPU disabled) - db233b | No FPU disabled - 22f816a | FPU disabled - 22f816a | FPU client only - 22f816a | FPU client only - isb[1] - 2bef61b | FPU client only - isb[2] - 2bef61b |
| --------- | --------------------------------------------- | ------------------------- | ---------------------- | ------------------------- | -------------------------- | -------------------------- |
[...]

PR #1570, opened 2025-12-14
Aarch64: Drop ISBs in enableFpu/disableFpu

Indanz:
The kernel is not doing any FPU operations after disabling the FPU, so there is never a need for an ISB there. Returning to user space counts as a synchronising context switch and ensures the FPU enable or disable has completed when user space starts executing.

An ISB after enabling the FPU is only needed in case the FPU was disabled for the kernel and the kernel wants to do FPU operations like saving or restoring the FPU state. The ISB ensures that the FPU enable system register write is finished before FPU using instructions are executed.

For Aarch64 non-HYP it is possible to disable the FPU for user space (EL0), but keep it enabled for the kernel (EL1). This is done by setting CPACR_EL1.FPEN to 1, which the code actually already did.

Aarch64 with virtualisation enabled is more complicated and needs higher level changes to achieve the same, which aren't done here, the change is limited to removing the ISB in disableFpu().

This partially fixes issue #1569.

### DIFF
--- a/include/arch/arm/arch/64/mode/machine/fpu.h
+++ b/include/arch/arm/arch/64/mode/machine/fpu.h
@@ -102,7 +102,11 @@ static inline void enableFpuEL01(void)
     MRS("cpacr_el1", cpacr);
     cpacr |= (3 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
-    isb();
+    /*
+     * No ISB because we rely on context switch for synchronisation
+     * of the FPU enable for EL0 and disableFpuEL0() keeps the FPU
+     * enabled for the kernel itself, which runs in EL1.
+     */
 }
 
 /* Enable the FPU to be used without faulting.
@@ -126,7 +130,7 @@ static inline void enableTrapFpu(void)
     MRS("cptr_el2", cptr);
     cptr |= (BIT(10) | BIT(31));
     MSR("cptr_el2", cptr);
-    isb();
+    /* No ISB because we rely on context switch for synchronisation */
 }
 
 /* Disable FPU access in EL0 */
@@ -135,9 +139,10 @@ static inline void disableFpuEL0(void)
     word_t cpacr;
     MRS("cpacr_el1", cpacr);
     cpacr &= ~(3 << CPACR_EL1_FPEN);
+    /* Keep FPU enabled for the kernel */
     cpacr |= (1 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
-    isb();
+    /* No ISB because we rely on context switch for synchronisation */
 }
 
 /* Disable the FPU so that usage of it causes a fault */
--- a/src/arch/arm/64/machine/fpu.c
+++ b/src/arch/arm/64/machine/fpu.c
@@ -17,7 +17,8 @@ BOOT_CODE bool_t fpsimd_init(void)
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
         enableFpuEL01();
     }
-
+    /* Non-HYP Kernel assumes FPU is always enabled for EL1: Make sure it is */
+    isb();
     return true;
 }
 #endif /* CONFIG_HAVE_FPU */


Indanz:
~~The ISB in `setCurrentUserVSpaceRoot()` can probably be removed too.~~ Nope, causes crashes. Not sure why though, I don't see any dependent operation happening during normal task switching. Edit2: Just removing the ISB seems fine for `switchToThread_fp()`, so I must have missed another corner case somewhere.

Keeping the FPU enabled for the kernel for Aarch64+HYP is much [harder](https://github.com/seL4/seL4/issues/1569#issuecomment-3642842085). The main problem is that the kernel's model doesn't match reality, especially when virtualisation is enabled.

A better fit would be to save and restore the FPU enable/disable register as part of the normal context switch code, like RISC-V already does, and give enableFpu() and disableFpu() the target thread as parameter, so they know for which task something is done. And instead of calling them during task switch, only call them during TCB flag changes.

include/arch/arm/arch/64/mode/machine/fpu.h
```diff
@@ -126,7 +127,7 @@ static inline void enableTrapFpu(void)
     MRS("cptr_el2", cptr);
     cptr |= (BIT(10) | BIT(31));
     MSR("cptr_el2", cptr);
-    isb();
+    /* No ISB because we count on context switch for synchronisation */
```
midnightveil: "count on" makes is sound more like hoping IMO. 

More accurately:

```suggestion
    /* For the direct writes to the `CPACR_EL1` register to be observed, we need a
       'Context Synchronization event' (see ARM DDI 0487 M.a §D24.1.2.2 / pD24-8367) to occur.
       We rely on 'Exception return' to cause explicit synchronization and omit an `ISB`
       as (a) the ISB has a ~30 cycle cost, and (b) the kernel does not access the
       FPU after disabling it.
```

include/arch/arm/arch/64/mode/machine/fpu.h
```diff
@@ -102,7 +102,8 @@ static inline void enableFpuEL01(void)
     MRS("cpacr_el1", cpacr);
     cpacr |= (3 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
-    isb();
+    /* No ISB because we count on context switch for synchronisation */
+    /* FPU is always enabled for the kernel, no extra need for an ISB here. */
```
midnightveil: I'd add a note that this function is called when the kernel runs in EL1 - should the function be renamed to `enableFPUEL0` and made to just to `(2 << CPACR_EL1_FPEN)`, just for EL0.

EDIT: Reading `disableFPUEL0`, this isn't the easiest to see what's going on... CPACR_EL1_FPEN is a 2-bit field, `0b11` (i.e. 3) is EL1/EL0 enable, `0b01` (i.e. 1) is EL1 enable EL0 disable. But that's not really relevant to this PR so much as I just find this slightly confusing.

(it's not particularly clear that this is the case, IMO).

The other thing is that this is called by `fpsimd_init()` during kernel boot, and there I think we *do* need an `ISB`? Which comes down to I guess it being named `enableFpuEL01` for a reason, when semantically we *mostly* use it as `enableFPUEL0`...?

Indanz: I'd rather keep setting all flags as we want it here than assuming it was set how we want it before and just toggling one bit.

Yes, took me a while too to realise that the name is very literal and that enabeFpuEL01 means it enables the FPU for both EL0 and EL1, and that disableFpuEL0 only disables it for EL0. Maybe it should be called diableFpuEL0_enableFpuEL1() instead.

> The other thing is that this is called by fpsimd_init() during kernel boot, and there I think we do need an ISB? Which comes down to I guess it being named enableFpuEL01 for a reason, when semantically we mostly use it as enableFPUEL0...?

The commit adding that is a bit clearer than the code itself:
```
commit f563916076d0fcda672cf3bf3d26ba98ba34b064
Author: Yanyan Shen <yanyan.shen@data61.csiro.au>
Date:   Wed Mar 28 10:54:44 2018 +1100

    armv8/hyp: Enable FPU in EL0/1 when HYP is on
    
    The disableFpu() for EL2 traps any FPU instruction to EL2 even
    if the FPU is enabled for EL0/1 in CPACR_EL1. For EL2, we
    use CPTR_EL2 to enable/disable trapping FPU instructions.
    Therefore, we enable FPU for EL0/1 by default so that no additional
    exceptions will be generated when we enable FPU access in CPTR_EL2.
```

Basically with HYP we use CPTR_EL2 to enable/disable the FPU instead of CPACR_EL1, so the code enables the FPU in CPACR_EL1 with the non-HYP version of fpuEnable().

(Still no need for an ISB.)

include/arch/arm/arch/64/mode/machine/fpu.h
Indanz: That's too verbose. I'll add it to the commit message though.

midnightveil: Maybe chop off the reasoning why if you want to shorten it?

I'd somewhat prefer overly verbose & why it's correct to be in the source...

include/arch/arm/arch/64/mode/machine/fpu.h
Indanz: > The other thing is that this is called by `fpsimd_init()` during kernel boot, and there I think we _do_ need an `ISB`?

I changed my mind, it would be prudent to add an ISB there. There are plenty other ISBs between this function and when we eventually try to touch the FPU, but ignoring those, strictly speaking we are counting on the cpu having FPU enabled at bootup, as the initial task will have it enabled and it's the first task we try to run.

Indanz:
Some results for tx1 (with the FPU disable sel4bench, non-HYP, non-MCS):

| Benchmark | Cycles Before | After | Speedup |
|--------|--------|--------|----|
| seL4_Call | 443 | 380 | 17% |
| seL4_Call (FPU) | 632 | 556 | 12% |
| seL4_ReplyRecv | 453 | 387 |17% |
| seL4_ReplyRecv (FPU) | 640 | 565 | 13% |
| IRQ path cycle count | 709 | 673 | 5% |
| Signal to process of higher prio | 1142 | 1042 | 12% |

With just dropping the ISBs in the FPU enable/disable code the results were Call 416, Call (FPU) 591, ReplyRecv 422.
So the last comit saved a further 30 cycles, as expected for one ISB.

I added some debug info and with these changes we have no ISBs in the fast path left, at least for aarch64 non-HYP.

Indanz:
I added `setCurrentUserVSpaceRoot_fp()`, but if it's easier for verification, I can instead add an extra parameter to the existing `setCurrentUserVSpaceRoot()` which tells it whether to do the ISB or not.

lsf37:
> I added `setCurrentUserVSpaceRoot_fp()`, but if it's easier for verification, I can instead add an extra parameter to the existing `setCurrentUserVSpaceRoot()` which tells it whether to do the ISB or not.

I think that would be preferable, yes, at least if it doesn't make anything slower.

Indanz:
> > I added `setCurrentUserVSpaceRoot_fp()`, but if it's easier for verification, I can instead add an extra parameter to the existing `setCurrentUserVSpaceRoot()` which tells it whether to do the ISB or not.
> 
> I think that would be preferable, yes, at least if it doesn't make anything slower.

It's an inline function, so the binary will be the same either way. But the fast path will be different than the slow path now though, not sure if the proofs can deal with that.

Because of the unstable CI system I'm not sure if failures are because of code problems or because of CI fragility. I think all tests passed at one point, but I'd like to see a couple of successful runs just to be sure. If easier, we can merge the no ISB in FPU code first and I can open a new PR for the second commit which removes it from the fastpath.

lsf37:
> > > I added `setCurrentUserVSpaceRoot_fp()`, but if it's easier for verification, I can instead add an extra parameter to the existing `setCurrentUserVSpaceRoot()` which tells it whether to do the ISB or not.
> > 
> > 
> > I think that would be preferable, yes, at least if it doesn't make anything slower.
> 
> It's an inline function, so the binary will be the same either way. But the fast path will be different than the slow path now though, not sure if the proofs can deal with that.

Oh, I didn't pick up on that. That's not going to work at all. It must do exactly the same machine operations, otherwise it does not satisfy its basic correctness requirement.

Why is it possible to leave out the isb on the fast path and not the slow path? Could we leave it out on the slow path as well if we find ourselves in the same state the fast path establishes?
 
> Because of the unstable CI system I'm not sure if failures are because of code problems or because of CI fragility. I think all tests passed at one point, but I'd like to see a couple of successful runs just to be sure. If easier, we can merge the no ISB in FPU code first and I can open a new PR for the second commit which removes it from the fastpath.

I'm bringing a new USB hub to UNSW today which may fix that 🤞 (it's the same bunch of boards that are hanging of the same old hub that are having problems).

Indanz:
> > It's an inline function, so the binary will be the same either way. But the fast path will be different than the slow path now though, not sure if the proofs can deal with that.
> 
> Oh, I didn't pick up on that. That's not going to work at all. It must do exactly the same machine operations, otherwise it does not satisfy its basic correctness requirement.
> 
> Why is it possible to leave out the isb on the fast path and not the slow path? Could we leave it out on the slow path as well if we find ourselves in the same state the fast path establishes?

Well, I think it's possible on the slow path as well, it just didn't work when I tried doing that though. But I'm not sure if that's because of CI failure or real, I'll have to put more time into testing this to know what exactly goes wrong where and why. Practically it's harder to do because they don't call `setCurrentUserVSpaceRoot` directly, but via armv_contextSwitch* or setVMRoot and I need to check all users of those too, which will take time.

That's why I proposed dropping that commit and merging the ISB drops in FPU enable/disable first.

> I'm bringing a new USB hub to UNSW today which may fix that 🤞 (it's the same bunch of boards that are hanging of the same old hub that are having problems).

That gives hope that things will improve soon. Before know I had no idea why things were randomly failing.

Indanz:
Okay, dropped the setCurrentUserVSpaceRoot commit, limiting this to just dropping ISBs in FPU enable/disable functions. I'll open another PR for the other changes.

lsf37:
> Okay, dropped the setCurrentUserVSpaceRoot commit, limiting this to just dropping ISBs in FPU enable/disable functions. I'll open another PR for the other changes.

Thank you, yes, let's look at that change in isolation.

lsf37:
> > I'm bringing a new USB hub to UNSW today which may fix that 🤞 (it's the same bunch of boards that are hanging of the same old hub that are having problems).
> 
> That gives hope that things will improve soon. Before know I had no idea why things were randomly failing.

Still a bit early to tell, but it seems to have improved the situation. The only failures I've seen since the swap are boards that are not booting.

lsf37 (approved):
This looks good and should be behind the machine interface for the proofs.

PR #1571, opened 2025-12-14
SMP: Only migrate when needed

Indanz:
schedContext_donate() is performance critical and should be fast,
but migrateTCB() assumes that there is a change in affinity and
calls fpuRelease() unconditionally, causing unnecessary FPU
saving and restoring.

Also a small cleanup: Use fpuRelease() instead of open-coding the same.

Further fixes issue #901.

### DIFF
--- a/include/arch/arm/arch/64/mode/machine/fpu.h
+++ b/include/arch/arm/arch/64/mode/machine/fpu.h
@@ -102,7 +102,11 @@ static inline void enableFpuEL01(void)
     MRS("cpacr_el1", cpacr);
     cpacr |= (3 << CPACR_EL1_FPEN);
     MSR("cpacr_el1", cpacr);
-    isb();
+    /*
+     * No ISB because we rely on context switch for synchronisation
+     * of the FPU enable for EL0 and disableFpuEL0() keeps the FPU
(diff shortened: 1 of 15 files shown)


src/model/smp.c
```diff
@@ -12,6 +12,9 @@
 
 void migrateTCB(tcb_t *tcb, word_t new_core)
 {
+    if (new_core == tcb->tcbAffinity) {
```
midnightveil, line 15: My one thought was "Is it better to do this check here or in the users of `migrateTCB`, and upgrade this to an `assert(new_core != tcb->tcbAffinity)`" (in the case of e.g. adding and removing from queues unnecessarily).

Looking at the uses of `migrateTCB`, there's only one spot where we already had a similar check, (`invokeSchedControl_ConfigureFlags`) but that check happens *after* this operation and likely depends on it. So I think the answer is: no, actually.

Indanz: That's what I had before in commit 26fbe2625. The only performance critical user is `schedContext_donate`. But it felt wrong because it doesn't seem correct to migrate to the same core.

lsf37 (approved):

PR #1574, opened 2025-12-16
FPU: Fix task enabling FPU for itself

Indanz:
There is no switchToThread() call when a task toggles seL4_TCBFlag_fpuDisabled for itself, so the FPU stays disabled even when it was just enabled.

The disable case was handled properly as fpuRelease() does the right thing already.

### DIFF
--- a/src/object/tcb.c
+++ b/src/object/tcb.c
@@ -802,9 +802,12 @@ static void invokeSetFlags(tcb_t *thread, word_t clear, word_t set, bool_t call)
     thread->tcbFlags = flags;
 
 #ifdef CONFIG_HAVE_FPU
-    /* Save current FPU state before disabling FPU: */
     if (flags & seL4_TCBFlag_fpuDisabled) {
+        /* Save current FPU state before disabling FPU: */
         fpuRelease(thread);
+    } else if (thread == cur_thread) {
+        /* Restore FPU here as switchToThread() won't be called: */
+        lazyFPURestore(thread);
     }
 #endif
     if (call) {


Indanz:
So the question is why we didn't find this and what we should change to avoid missing bugs like this in the future.

The sel4test tests did test whether enabling and disabling the FPU actually did work, but only for another task.

I expected verification to find bugs like these though, or isn't there an invariant that says the FPU is enabled when a task is running and `seL4_TCBFlag_fpuDisabled` flag isn't set?

lsf37:
There currently is no invariant for that, but we could add one and probably should. 

The behaviour doesn't violate any of the high level properties, so nothing so far was tripping this up as wrong behaviour. The code does what the spec says (which is "don't enable the FPU if it is for the same thread"), because we made the same mistake in both. 

The way to prevent these is to add a new property that expresses what we want. I think adding an invariant that expresses that the FPU should be enabled when a thread is running (i.e. after kernel exit) if and only if the FPU flag in the TCB is set will do it.

We'll first just fix the proofs for this change and then add the invariant.

Indanz:
> The way to prevent these is to add a new property that expresses what we want. I think adding an invariant that expresses that the FPU should be enabled when a thread is running (i.e. after kernel exit) if and only if the FPU flag in the TCB is set will do it.

Like I said [elsewhere](https://github.com/seL4/seL4/pull/1570#issuecomment-3650025977), the current code isn't at the optimal abstraction level, so I would wait with this before we know what the best path forward is. All archs with virtualisation go through great trouble getting FPU to work properly, and RISC-V which has no virtualisation yet also doesn't match the model.

I suspect the minimal change is to give enable/disable FPU the task as parameter and only call it during flag changes instead of context switches. Then RISC-V doesn't need to check the global flag when restoring user context and x86 and Arm virtualisation code doesn't need to fight with the FPU context switch code. But I haven't tried this out yet to see whether it would work as hoped.

midnightveil:
> The behaviour doesn't violate any of the high level properties, so nothing so far was tripping this up as wrong behaviour. The code does what the spec says (which is "don't enable the FPU if it is for the same thread"), because we made the same mistake in both.

Is there no high-level property like 

"if we don't rerun the scheduler then the final state of the kernel is identical to if we did call chooseTheead (which looks at the scheduler state and eventually calls switchToThread)?"

The only difference should be one of performance / noop operations (in theory, I think). I suspect that this might be... hard to prove, but I would have assumed this is how the fastpaths are phrased? (are they more... specific?)

Indanz:
> Is there no high-level property like
> 
> "if we don't rerun the scheduler then the final state of the kernel is identical to if we did call chooseTheead (which looks at the scheduler state and eventually calls switchToThread)?"

There is an invariant that checks that the task running is always the highest priority ready task, but that's about it I think.

In this case schedule() wouldn't call switchToThread() anyway because `NODE_STATE(ksSchedulerAction) == SchedulerAction_ResumeCurrentThread`, so it wouldn't have caught this bug.

lsf37:
> Is there no high-level property like
> 
> "if we don't rerun the scheduler then the final state of the kernel is identical to if we did call chooseTheead (which looks at the scheduler state and eventually calls switchToThread)?"

No, there is nothing like that.

It's not really a high-level property -- it talks about functions and implementation details of the kernel. The one I was talking about would be an API-level property, connecting an API concept to machine state and talking about a global invariant instead of the result of certain functions inside the kernel. Those are much easier to phrase and therefore easier to be sure that they express what we want.
 
> The only difference should be one of performance / noop operations (in theory, I think). 

I doubt that it would be true for all aspects of kernel state, but it's not impossible. Judging from how much people get subtly wrong in the fastpath implementations before verification, those kinds of equivalences are hard to think about.

> I suspect that this might be... hard to prove, but I would have assumed this is how the fastpaths are phrased? (are they more... specific?)

Yes, the only property of that kind we have for specific functions is the fast path and it is indeed a terrible proof. Not necessarily very hard, but messy and full of reasoning about reordering of executions under very specific conditions. 

It's not that those kinds of equivalences can't be useful or high level -- for the fast path it is a pretty good correctness criterion. But they do tend to be messy to prove in the kernel.

corlewis:
> There is an invariant that checks that the task running is always the highest priority ready task, but that's about it I think.

We actually don't even have that, although we've talked about proving it in the past. The only high-level scheduler property that is proven is for non-MCS seL4 and says that the scheduler does not 'forget' about any runnable threads, i.e. that any runnable threads are either the current thread, are about to be switched to, or are in a scheduling queue.

corlewis (approved):
Looks good to me and I've updated the proofs in https://github.com/seL4/l4v/pull/946.

Thanks for finding this and apologies for not noticing when writing the specification. It's often easier to spot things like this at that level but this one is subtle enough that we all missed it as well.

Indanz:
> Looks good to me and I've updated the proofs in [seL4/l4v#946](https://github.com/seL4/l4v/pull/946).

Great!

> Thanks for finding this and apologies for not noticing when writing the specification. It's often easier to spot things like this at that level but this one is subtle enough that we all missed it as well.

I feel equally guilty, let's share the blame 50-50.

PR #1576, opened 2025-12-17
mcs: handle endpoint and ntfn queues uniformly

michaelmcinerney:
- Introduce functions to append and dequeue to/from endpoint and notification queues, in order to make verification feasible. Handle linked list manipulations with the same functions that were previously used for the ready and release queues, together with a new function that allows for a new item to be inserted into the queue after another item that is already in the queue.

- Remove tcbEPNext and tcbEPPrev pointers from the MCS version, and instead use tcbSchedNext and tcbSchedPrev pointers, given that no thread can be simultaneously in any two of the endpoint, notification, ready, or release queues.

### DIFF
--- a/include/object/tcb.h
+++ b/include/object/tcb.h
@@ -128,42 +128,62 @@ void remoteTCBStall(tcb_t *tcb);
 #define SCHED_APPEND_CURRENT_TCB    tcbSchedAppend(NODE_STATE(ksCurThread))
 
 #ifdef CONFIG_KERNEL_MCS
-/* Add TCB into the priority ordered endpoint queue */
-static inline tcb_queue_t tcbEPAppend(tcb_t *tcb, tcb_queue_t queue)
+
+static inline bool_t PURE higher_than_tcb_prio(tcb_t *tcb, prio_t priority)
 {
-    /* start at the back of the queue as FIFO is the common case */
-    tcb_t *before = queue.end;
-    tcb_t *after = NULL;
-
-    /* find a place to put the tcb */
-    while (unlikely(before != NULL && tcb->tcbPriority > before->tcbPriority)) {
-        after = before;
-        before = after->tcbEPPrev;
-    }
-
-    if (unlikely(before == NULL)) {
-        /* insert at head */
-        queue.head = tcb;
-    } else {
-        before->tcbEPNext = tcb;
-    }
-
-    if (likely(after == NULL)) {
-        /* insert at tail */
-        queue.end = tcb;
-    } else {
-        after->tcbEPPrev = tcb;
-    }
-
-    tcb->tcbEPNext = after;
-    tcb->tcbEPPrev = before;
-
-    return queue;
+    return tcb != NULL && priority > tcb->tcbPriority;
 }
 
-tcb_queue_t tcbEPDequeue(tcb_t *tcb, tcb_queue_t queue);
+/* Find the rightmost TCB in the given queue that has a priority which is
+   strictly greater than the given priority */
+static tcb_t *find_tcb_with_higher_prio(tcb_queue_t queue, prio_t priority)
+{
+    tcb_t *tcb = queue.end;
+
+    while (higher_than_tcb_prio(tcb, priority)) {
+        tcb = tcb->tcbSchedPrev;
+    }
+
+    return tcb;
+}
+
+/* Insert a TCB into a queue immediately after another item in the queue
+   (the queue must initially contain at least two items) */
+static inline void tcb_queue_insert_after(tcb_t *tcb, tcb_t *before)
+{
+    tcb_t *after;
+    after = before->tcbSchedNext;
+
+    tcb->tcbSchedPrev = before;
+    tcb->tcbSchedNext = after;
+
+    after->tcbSchedPrev = tcb;
+    before->tcbSchedNext = tcb;
+}
+
+/* Add TCB into the priority ordered endpoint or notification queue */
+static inline tcb_queue_t tcbAppend(tcb_t *tcb, tcb_queue_t queue)
+{
+    prio_t priority = tcb->tcbPriority;
+    tcb_queue_t new_queue = queue;
+
+    if (tcb_queue_empty(queue) || priority > queue.head->tcbPriority) {
+        new_queue = tcb_queue_prepend(queue, tcb);
+    } else {
+        if (queue.end->tcbPriority >= priority) {
+            new_queue = tcb_queue_append(queue, tcb);
+        } else {
+            tcb_t *before;
+            before = find_tcb_with_higher_prio(queue, priority);
+            tcb_queue_insert_after(tcb, before);
+        }
+    }
+
+    return new_queue;
+}
 
 #else
+
 tcb_queue_t tcbEPAppend(tcb_t *tcb, tcb_queue_t queue);
 tcb_queue_t tcbEPDequeue(tcb_t *tcb, tcb_queue_t queue);
 
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -465,7 +465,7 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
     } else {
 #ifdef CONFIG_KERNEL_MCS
         /* Update queue. */
-        tcb_queue_t queue = tcbEPAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
+        tcb_queue_t queue = tcbAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
         endpoint_ptr_set_epQueue_head_np(ep_ptr, TCB_REF(queue.head));
         endpoint_ptr_mset_epQueue_tail_state(ep_ptr, TCB_REF(queue.end), EPState_Recv);
 #else
--- a/src/object/endpoint.c
+++ b/src/object/endpoint.c
@@ -28,7 +28,6 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
     case EPState_Idle:
     case EPState_Send:
         if (blocking) {
-            tcb_queue_t queue;
 
             /* Set thread state to BlockedOnSend */
             thread_state_ptr_set_tsType(&thread->tcbState,
@@ -47,10 +46,16 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
             scheduleTCB(thread);
 
             /* Place calling thread in endpoint queue */
+#ifdef CONFIG_KERNEL_MCS
+            tcbEPAppend(thread, epptr, EPState_Send);
+#else
+            tcb_queue_t queue;
             queue = ep_ptr_get_queue(epptr);
             queue = tcbEPAppend(thread, queue);
             endpoint_ptr_set_state(epptr, EPState_Send);
             ep_ptr_set_queue(epptr, queue);
+#endif /* CONFIG_KERNEL_MCS */
+
         }
         break;
 
@@ -66,12 +71,16 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
         assert(dest);
 
         /* Dequeue the first TCB */
+#ifdef CONFIG_KERNEL_MCS
+        tcbEPDequeue(dest, epptr);
+#else
         queue = tcbEPDequeue(dest, queue);
         ep_ptr_set_queue(epptr, queue);
 
         if (!queue.head) {
             endpoint_ptr_set_state(epptr, EPState_Idle);
         }
+#endif /* CONFIG_KERNEL_MCS */
 
         /* Do the transfer */
         doIPCTransfer(thread, epptr, badge, canGrant, dest);
@@ -165,7 +174,6 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
         switch (endpoint_ptr_get_state(epptr)) {
         case EPState_Idle:
         case EPState_Recv: {
-            tcb_queue_t queue;
 
             if (isBlocking) {
                 /* Set thread state to BlockedOnReceive */
@@ -187,10 +195,15 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
 #endif
 
                 /* Place calling thread in endpoint queue */
+#ifdef CONFIG_KERNEL_MCS
+                tcbEPAppend(thread, epptr, EPState_Recv);
+#else
+                tcb_queue_t queue;
                 queue = ep_ptr_get_queue(epptr);
                 queue = tcbEPAppend(thread, queue);
                 endpoint_ptr_set_state(epptr, EPState_Recv);
                 ep_ptr_set_queue(epptr, queue);
+#endif /* CONFIG_KERNEL_MCS */
             } else {
                 doNBRecvFailedTransfer(thread);
             }
@@ -213,12 +226,16 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
             assert(sender);
 
             /* Dequeue the first TCB */
+#ifdef CONFIG_KERNEL_MCS
+            tcbEPDequeue(sender, epptr);
+#else
             queue = tcbEPDequeue(sender, queue);
             ep_ptr_set_queue(epptr, queue);
 
             if (!queue.head) {
                 endpoint_ptr_set_state(epptr, EPState_Idle);
             }
+#endif /* CONFIG_KERNEL_MCS */
 
             /* Get sender IPC details */
             badge = thread_state_ptr_get_blockingIPCBadge(&sender->tcbState);
@@ -321,7 +338,6 @@ void cancelIPC(tcb_t *tptr)
     case ThreadState_BlockedOnReceive: {
         /* blockedIPCCancel state */
         endpoint_t *epptr;
-        tcb_queue_t queue;
 
         epptr = EP_PTR(thread_state_ptr_get_blockingObject(state));
 
@@ -329,6 +345,10 @@ void cancelIPC(tcb_t *tptr)
         assert(endpoint_ptr_get_state(epptr) != EPState_Idle);
 
         /* Dequeue TCB */
+#ifdef CONFIG_KERNEL_MCS
+        tcbEPDequeue(tptr, epptr);
+#else
+        tcb_queue_t queue;
         queue = ep_ptr_get_queue(epptr);
         queue = tcbEPDequeue(tptr, queue);
         ep_ptr_set_queue(epptr, queue);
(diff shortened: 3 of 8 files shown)


include/object/tcb.h
```diff
@@ -128,42 +128,68 @@ void remoteTCBStall(tcb_t *tcb);
...
-        /* insert at head */
-        queue.head = tcb;
-    } else {
-        before->tcbEPNext = tcb;
-    }
+/* Find the rightmost TCB in a queue that has a priority which is strictly
+   greater than the given priority */
+static tcb_t *find_tcb_with_higher_prio(tcb_t *tcb, prio_t new_priority)
+{
+    tcb_t *before = tcb;
```
Indanz:
```suggestion
static inline tcb_t *find_tcb_with_higher_prio(tcb_t *before, prio_t new_priority)
{
```
Or give it a queue as parameter instead of a tcb.

include/object/tcb.h
```diff
@@ -128,42 +128,68 @@ void remoteTCBStall(tcb_t *tcb);
...
+/* Insert a TCB into a queue immediately after another item in the queue
+   (the queue must initially contain at least two items) */
+static inline void tcb_queue_insert_after(tcb_t *tcb, tcb_t *before)
+{
+    tcb_t *after;
+    after = before->tcbSchedNext;
+
+    assert(after != NULL);
+    assert(after != before);
+
```
Indanz: No need for this, I don't think it adds anything. You don't check that `before != NULL` either.

include/object/tcb.h
```diff
@@ -128,42 +128,68 @@ void remoteTCBStall(tcb_t *tcb);
 #define SCHED_APPEND_CURRENT_TCB    tcbSchedAppend(NODE_STATE(ksCurThread))
 
 #ifdef CONFIG_KERNEL_MCS
-/* Add TCB into the priority ordered endpoint queue */
-static inline tcb_queue_t tcbEPAppend(tcb_t *tcb, tcb_queue_t queue)
+
+static inline bool_t PURE higher_than_tcb_prio(tcb_t *tcb, prio_t new_priority)
```
Indanz:
```suggestion
static inline bool_t PURE higher_than_tcb_prio(tcb_t *tcb, prio_t priority)
```

include/object/tcb.h
```diff
@@ -128,42 +128,68 @@ void remoteTCBStall(tcb_t *tcb);
...
+            new_queue = tcb_queue_append(queue, tcb);
+        } else {
+            tcb_t *before;
+            before = find_tcb_with_higher_prio(queue.end, new_priority);
+            tcb_queue_insert_after(tcb, before);
+        }
+    }
+
+    return new_queue;
+}
```
Indanz:
```suggestion
static inline tcb_queue_t tcbIPCAppend(tcb_queue_t queue, tcb_t *tcb)
{
    prio_t priority = tcb->tcbPriority;

    if (higher_than_tcb_prio(queue.head, priority)) {
        queue = tcb_queue_prepend(queue, tcb);
    } else {
        if (queue.end->tcbPriority >= priority) {
            queue = tcb_queue_append(queue, tcb);
        } else {
            tcb_t *before = find_tcb_with_higher_prio(queue.end, priority);
            tcb_queue_insert_after(tcb, before);
        }
    }
    return queue;
}
```
The end check could be replaced with `!higher_than_tcb_prio(queue.end, priority)`, but that would add an extra NULL check. Anyway, not sure if using higher_than_tcb_prio() is better, but if not here, then why have it at all?

src/fastpath/fastpath.c
```diff
@@ -175,12 +175,16 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
...
+#ifdef CONFIG_KERNEL_MCS
+    tcbEPDequeue_fp(dest, ep_ptr);
+#else
     endpoint_ptr_set_epQueue_head_np(ep_ptr, TCB_REF(dest->tcbEPNext));
     if (unlikely(dest->tcbEPNext)) {
         dest->tcbEPNext->tcbEPPrev = NULL;
     } else {
         endpoint_ptr_mset_epQueue_tail_state(ep_ptr, 0, EPState_Idle);
     }
+#endif
```
Indanz: I'd rather you rename tcbEPNext/tcbEPPrev for non-MCS and MCS to just tcbNext/tcbPrev and keep code that is the same the same.

src/fastpath/fastpath.c
```diff
@@ -465,7 +474,7 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
     } else {
 #ifdef CONFIG_KERNEL_MCS
         /* Update queue. */
-        tcb_queue_t queue = tcbEPAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
+        tcb_queue_t queue = tcbIPCAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
```
Indanz: I prefer tcbAppend, because "IPC" is always ambiguous whether that includes notifications or not. E.g. in the manual it are separate chapters.

src/object/endpoint.c
```diff
@@ -487,11 +535,45 @@ void cancelBadgedSends(endpoint_t *epptr, word_t badge)
...
+    tcb_queue_t queue;
+    tcb_queue_t new_queue;
+
+    queue = ep_ptr_get_queue(epptr);
+    new_queue = tcbIPCAppend(thread, queue);
+    ep_ptr_set_queue(epptr, new_queue);
+
+    /* if the queue was originally empty, update the state of
+     * the endpoint according to the flag that was passed in */
+    if (tcb_queue_empty(queue)) {
```
Indanz: The original code did this unconditionally, which seems much cheaper than doing an extra if-check. Was the old code wrong?

src/object/endpoint.c
```diff
@@ -47,10 +46,16 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
...
             /* Place calling thread in endpoint queue */
+#ifdef CONFIG_KERNEL_MCS
+            tcbEPAppend(thread, epptr, false);
+#else
+            tcb_queue_t queue;
             queue = ep_ptr_get_queue(epptr);
             queue = tcbEPAppend(thread, queue);
             endpoint_ptr_set_state(epptr, EPState_Send);
             ep_ptr_set_queue(epptr, queue);
+#endif /* CONFIG_KERNEL_MCS */
```
Indanz, line 57: This makes the MCS and non-MCS code diverge unnecessarily. It's also very confusing to have two totally different `tcbEPAppend` functions. Can't you keep this code the same for both?

src/object/notification.c
```diff
@@ -240,47 +248,66 @@ void receiveSignal(tcb_t *thread, cap_t cap, bool_t isBlocking)
...
 }
 
+#ifdef CONFIG_KERNEL_MCS
+static inline void removeAndRestartNTFNQueuedThread(tcb_t *thread, notification_t *ntfnPtr)
+{
+    tcbNTFNDequeue(thread, ntfnPtr);
+    setThreadState(thread, ThreadState_Restart);
+    if (sc_sporadic(thread->tcbSchedContext)) {
+        /* We know that the thread can't have the current SC as its own SC as
+         * this point as it should still be associated with the current thread,
```
Indanz:
```suggestion
        /* We know that the thread can't have the current SC as its own SC at
         * this point as it should still be associated with the current thread,
```

Indanz (commented):
I don't understand how a unified version can end up with 200 more lines of code than the old version. :-(

I don't like how this makes MCS diverge more from non-MCS, in seemingly unnecessary ways.

All in all the overall result doesn't look better, so I'm surprised it helps verification. Moving MCS code to their own functions seems like a good cleanup that probably helps verification, but not sure about the rest.

Indanz:
>     * Remove tcbEPNext and tcbEPPrev pointers from the MCS version, and instead use tcbSchedNext and tcbSchedPrev pointers, given that no thread can be simultaneously in any two of the endpoint, notification, ready, or release queues.

The same is true for non-MCS though, so why not to the same there, just with different (non-ordering) list functions?

lsf37:
> > ```
> > * Remove tcbEPNext and tcbEPPrev pointers from the MCS version, and instead use tcbSchedNext and tcbSchedPrev pointers, given that no thread can be simultaneously in any two of the endpoint, notification, ready, or release queues.
> > ```
> 
> The same is true for non-MCS though, so why not to the same there, just with different (non-ordering) list functions?

You were too fast, I meant to write a comment explaining all this :-)

The reason everything is `#ifdef` for now is so that we don't have to update the proofs for master. Doing this change in the queues was a longer multi-month project and we currently don't have that kind of time for the master proofs (where it would be even longer, because it affects more proofs -- MCS is currently only done for RISCV). We also have multiple other proofs in progress on master (e.g. domains, infoflow, arch-split etc), some of which would be majorly affected, especially the arch split of the refinement proofs.

So we completely agree, it would be much nicer to do the same queue unification on master and the general reasoning does work there, but we can't do it right now and we can't wait with MCS until they are all done, because that would be a lot of delay on MCS.

What we can do is make this change for MCS only for now, set up the proof framework there and then transition it over to master when those other projects are no longer affected. At that point most of the `#ifdefs` can either disappear again, or at least be hidden inside interface functions.

This PR will actually require some rework on the master proofs already (so we cannot merge it immediately), but hopefully only about a week's worth of work.

Indanz:
> So we completely agree, it would be much nicer to do the same queue unification on master and the general reasoning does work there, but we can't do it right now and we can't wait with MCS until they are all done, because that would be a lot of delay on MCS.
> 
> What we can do is make this change for MCS only for now, set up the proof framework there and then transition it over to master when those other projects are no longer affected. At that point most of the `#ifdefs` can either disappear again, or at least be hidden inside interface functions.
> 
> This PR will actually require some rework on the master proofs already (so we cannot merge it immediately), but hopefully only about a week's worth of work.

None of that is a good argument why not to rename tcbEPNext/tcbEPPrev for non-MCS and MCS to just tcbNext/tcbPrev and keep code that is the same the same. If you want to hide the rename you can add a define for non-MCS. What I object to is having exactly the same code with the only difference being the name of the next/prev pointers.

lsf37:
> None of that is a good argument why not to rename tcbEPNext/tcbEPPrev for non-MCS and MCS to just tcbNext/tcbPrev and keep code that is the same the same. If you want to hide the rename you can add a define for non-MCS. What I object to is having exactly the same code with the only difference being the name of the next/prev pointers.

Let me look into that. I thought we tried, but failed because types are different. But there have been a few iterations of this and that might not be true any more. If we can reduce the noise without causing proof rework on master, by all means we should.

include/object/tcb.h
michaelmcinerney: To me, although your suggestion is equivalent, it makes more sense to have the initial TCB that is "fixed", and to then have a separate variable that is updated during the loop. But I'm no C programmer. I'd be happy to change it if your suggestion is more idiomatic or is somehow more efficient.

I do only use this function once, and with the end of the queue, so I could also just give it a queue. That bothers me in a way, because I'd like to give the function the least information possible, so a tcb is better than a queue.

include/object/tcb.h
michaelmcinerney: `higher_than_tcb_prio` is used really just to get the `while` loop in `find_tcb_with_higher_prio` into the right shape for our rules: the correspondence rules we have for while loops are very sensitive to the parsed output and it's very difficult to manage this without a dedicated function for the loop condition. 

I'd rather not use `higher_than_tcb_prio` in this function because it would make the conditions here a bit less transparent.

src/fastpath/fastpath.c
michaelmcinerney: I will make it `tcbAppend`. The downside is we then have a function called `tcb_queue_append` which is the more generic function that performs the linked list update, and then this function `tcbAppend`.

src/object/endpoint.c
michaelmcinerney: This is done so that we can pass in a boolean flag as to whether we'd like to new state to be send or receive, and makes the function more generic. Before, the update was just hard coded into the blob.

michaelmcinerney:
One approach we can take that would get rid of the new function `tcbEPDequeue_fp` in `fastpath.h` is to simply introduce a `#define tcbEPNext tcbSchedNext` and `#define tcbEPPrev tcbSchedPrev` within MCS. I'll push an update to the PR with this change shortly.

Gerwin and I can't see any further improvements to increase the uniformity of the definitions. Is there anything else you had in mind?

include/object/tcb.h
Indanz: If you do that, I at least expect that `tcb` would be used elsewhere for something, so I have to keep it in mind. While if you re-use the variable, there is only one to keep track of.

In this case you return either `tcb` or another one with higher priority before that, so they are in that sense the same and re-using the same variable makes a lot of sense to me.

But no strong opinion either way, it was just a suggestion.

The name `before` isn't really correct and a bit confusing though, so I'd keep it named `tcb`, even when re-using it.

> I do only use this function once, and with the end of the queue, so I could also just give it a queue. That bothers me in a way, because I'd like to give the function the least information possible, so a tcb is better than a queue.

That's good habit which I agree with. But if you assume that `tcb` is the end of a queue and not somewhere in the middle of a queue, then you're not really giving it less information, you are just assuming more.

src/fastpath/fastpath.c
Indanz: Well, we had that downside already, but with slightly different names, this doesn't change the problem, perhaps just makes it slightly more obvious.

src/object/endpoint.c
Indanz: Wouldn't it make more sense to pass the new endpoint state directly, to avoid the if?

And if the new function does exactly the same, why not use it for non-MCS? The proof update required for this must be pretty much the same as what you need to do for MCS anyway.

michaelmcinerney: Good point, I will just pass in the new endpoint state. Thanks. I was following the abstract spec a bit too closely, here.

We wouldn't be able to use this new function for non-MCS without some sort of update to the mainline proofs. I could investigate just how long that might take, but my guess is that we wouldn't want to consider that right now.

include/object/tcb.h
michaelmcinerney: Alright, I'll go with your suggestion of passing in a queue and using the variable `tcb` throughout the function. Thanks

src/object/endpoint.c
lsf37: The proof update for MCS took multiple months, and they won't work the same for master, we definitely do not have the time for it.

src/object/endpoint.c
```diff
@@ -487,11 +535,45 @@ void cancelBadgedSends(endpoint_t *epptr, word_t badge)
...
+    ep_ptr_set_queue(epptr, new_queue);
+
+    /* if the queue was originally empty, update the state of
+     * the endpoint according to the flag that was passed in */
+    if (tcb_queue_empty(queue)) {
+        if (isRecv) {
+            endpoint_ptr_set_state(epptr, EPState_Recv);
+        } else {
+            endpoint_ptr_set_state(epptr, EPState_Send);
+        }
```
lsf37: Not sure if this too annoying in CRefine, but this would be shorter and at least to me more readable.

```suggestion
        endpoint_ptr_set_state(epptr, isRecv ? EPState_Recv : EPState_Send)
```

lsf37: We could also still do this unconditionally, without checking if the queue is empty first -- at least if we know that the thread state will only be overwritten with the same thing, which I think we do. I'm not convinced that is really faster, though.

@michaelmcinerney @Indanz what do you think of that?

lsf37: Actually, I think Indan is right, it would be faster. We'd still have the conditional `isRecv` every time, but at the call site that is a constant, so the compiler can remove it.

src/object/endpoint.c
```diff
@@ -487,11 +535,45 @@ void cancelBadgedSends(endpoint_t *epptr, word_t badge)
...
+    }
```
lsf37: Ok, @michaelmcinerney, so the proposal would be this instead:

```suggestion
    /* Update the state of the endpoint according to the flag that was passed in. If the queue
     * was previously non-empty this must be the same state the endpoint is currently in. */
    endpoint_ptr_set_state(epptr, isRecv : EPState_Recv ? EPState_Send);
```

lsf37: If we do what Indan suggested in the beginning and pass in the thread state directly, it would become just

```
endpoint_ptr_set_state(epptr, ep_state)
```

(assuming that new thread state parameter is `ep_state`). This would then be fully equivalent to what the code before did.

src/object/endpoint.c
lsf37: The `#ifdefs` are now reduced a bit, but they are still there where the queue handling code was manually inlined before. We will do this for master as well and then they can all disappear again, but if we want to get the MCS proofs done in any reasonable time frame we need to do this detour first.

lsf37:
(rebased)

lsf37 (approved):

midnightveil:
BTW: this has broken builds with KernelSignalFastpath & MCS set (not verification builds).

lsf37:
> BTW: this has broken builds with KernelSignalFastpath & MCS set (not verification builds).

Hm, looks like we missed the signal fast path, yes. Looking into it.

Ivan-Velickovic:
> Hm, looks like we missed the signal fast path, yes. Looking into it.

Also the TCBBits size seems to be excessive in certain configurations (e.g benchmarking) which is affecting Microkit. Looking into it now, I guess the MCS TCB size is now smaller for all configurations.

lsf37:
> Also the TCB seems to be excessive in certain configurations (e.g benchmarking) which is affecting Microkit. Looking into it now, I guess the MCS TCB size is now smaller for all configurations.

TCB size has only decreased in this PR, and should not have increased in any other recent commit.

midnightveil:
> TCB size has only decreased in this PR, and should not have increased in any other recent commit, so this is an issue.

```c
compile_assert(tcb_size_sane,
               BIT(TCB_SIZE_BITS) >= sizeof(tcb_t))
compile_assert(tcb_size_not_excessive,
               BIT(TCB_SIZE_BITS - 1) < sizeof(tcb_t))
```

There's a code assert for that, "not_excessive" which fails.

Ivan-Velickovic:
> TCB size has only decreased in this PR, and should not have increased in any other recent commit.

Sorry, typo in my comment. Meant to say that the `seL4_TCBBits` size is now excessive:

```
/Users/ivanv/ts/microkit/seL4/include/assert.h:67:36: error: static assertion failed: "tcb_size_not_excessive"
   67 | #define compile_assert(name, expr) _Static_assert(expr, #name);
```

lsf37:
Right. Which configuration is that? Just MCS + benchmarking enabled? The TCB size probably got small enough that we now need to decrease TCB_SIZE_BITS there.

We really do not seem to have enough configs in the regression test, although that one should actually fire in sel4bench.

Edit: it did not, the build in https://github.com/seL4/sel4bench/actions/runs/26268432916/job/77316616906 succeeded, i.e. it must be more than just benchmarking on + MCS on.

Ivan-Velickovic:
> i.e. it must be more than just benchmarking on + MCS on.

I think benchmarking + hyp + MCS.

lsf37:
> > i.e. it must be more than just benchmarking on + MCS on.
> 
> I think benchmarking + hyp + MCS.

Looks like it is SMP + HYP + MCS + benchmarking: #1675

PR #1577, opened 2025-12-18
manual: fix incorrect x86 EPT VMAttributes type

dreamliner787-9:
On x86, the caching attributes encoding used for ordinary page-table entries differ from those used for Extended Page Tables (EPT) in second-stage translation (1). The manual previously implied they were the same. This change corrects that description to prevent subtly incorrect mappings when following the manual.

(1): https://github.com/seL4/seL4/blob/master/libsel4/arch_include/x86/sel4/arch/types.h

### DIFF
--- a/manual/parts/vspace.tex
+++ b/manual/parts/vspace.tex
@@ -285,13 +285,15 @@ RISC-V               & \autoref{group__riscv__seL4__RISCV__ASIDPool} \\
 \end{tabularx}
 
 \section{Mapping Attributes}
-A parameter of type \texttt{seL4\_ARM\_VMAttributes}, \texttt{seL4\_x86\_VMAttributes}
-, or \texttt{seL4\_RISCV\_VMAttributes} is used to specify the cache behaviour of the
-page being mapped. Possible values for Arm that can be bitwise OR'd together are
-shown in \autoref{tbl:vmattr_arm} \ifxeightsix. An enumeration of valid values
-for IA-32 and x64 are shown in \autoref{tbl:vmattr_x86}\fi. Possible values for RISC-V that
-can be bitwise OR'd together are shown in \autoref{tbl:vmattr_riscv}. Mapping attributes
-can be updated on existing mappings using the Map invocation with the same virtual address.
+A parameter of type \texttt{seL4\_ARM\_VMAttributes}, \texttt{seL4\_X86\_VMAttributes},
+\texttt{seL4\_X86\_EPT\_VMAttributes}, or \texttt{seL4\_RISCV\_VMAttributes} is used to specify the
+cache behaviour of the page being mapped. Possible values for Arm that can be bitwise OR'd together
+are shown in \autoref{tbl:vmattr_arm} \ifxeightsix. An enumeration of valid values for IA-32 and x64
+are shown in \autoref{tbl:vmattr_x86}. An enumeration of valid values for IA-32 and x64 Extended
+Page Table (EPT) used in second-level translation are shown in \autoref{tbl:vmattr_x86_ept}.
+Possible values for RISC-V that can be bitwise OR'd together are shown in
+\autoref{tbl:vmattr_riscv}. Mapping attributes can be updated on existing mappings using the Map
+invocation with the same virtual address.
 
 \begin{table}[htb]
   \begin{center}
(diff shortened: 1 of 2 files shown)


manual/parts/vspace.tex
```diff
@@ -285,13 +285,15 @@ \subsection{ASID Pool}
...
 
 \section{Mapping Attributes}
-A parameter of type \texttt{seL4\_ARM\_VMAttributes}, \texttt{seL4\_x86\_VMAttributes}
-, or \texttt{seL4\_RISCV\_VMAttributes} is used to specify the cache behaviour of the
-page being mapped. Possible values for Arm that can be bitwise OR'd together are
-shown in \autoref{tbl:vmattr_arm} \ifxeightsix. An enumeration of valid values
-for IA-32 and x64 are shown in \autoref{tbl:vmattr_x86}\fi. Possible values for RISC-V that
-can be bitwise OR'd together are shown in \autoref{tbl:vmattr_riscv}. Mapping attributes
-can be updated on existing mappings using the Map invocation with the same virtual address.
+A parameter of type \texttt{seL4\_ARM\_VMAttributes}, \texttt{seL4\_x86\_VMAttributes},
```
midnightveil: you missed a lowercase x86 here

dreamliner787-9: Whoops, fixed, thank you.

Indanz (approved):

lsf37 (approved):

libsel4/sel4_arch_include/x86_64/interfaces/object-api-sel4-arch.xml
```diff
@@ -39,7 +39,7 @@
             </description>
             <param dir="in" name="pml4" type="seL4_X64_PML4" description="Capability to the VSpace which will contain the mapping."/>
             <param dir="in" name="vaddr" type="seL4_Word" description="Virtual address at which to map page."/>
-            <param dir="in" name="attr" type="seL4_X86_VMAttributes">
+            <param dir="in" name="attr" type="seL4_X86_EPT_VMAttributes">
```
lsf37: Is this change correct? It's inside `<method id="X86PDPTMap" name="Map">`, so I would have expected `seL4_X86_VMAttributes` to be right.

Ivan-Velickovic: It's not an EPT invocation so I think it's wrong, I've added a commit to revert this change.

lsf37:
(rebased)

libsel4/sel4_arch_include/x86_64/interfaces/object-api-sel4-arch.xml
Indanz: This is actually quite hard to check.

seL4_X86_VMAttributes is:
```c
typedef enum {
    seL4_X86_Default_VMAttributes = 0,
    seL4_X86_WriteBack = 0,
    seL4_X86_WriteThrough = 1,
    seL4_X86_CacheDisabled = 2,
    seL4_X86_Uncacheable = 3,
    seL4_X86_WriteCombining = 4,
    SEL4_FORCE_LONG_ENUM(seL4_X86_VMAttributes),
} seL4_X86_VMAttributes;
```

X86PDPTMap calls `vmAttributesFromWord()` to translate the argument, which just turns it into a `vm_attributes_t`, which has the following bits:

```
block vm_attributes {
    padding         61           -- (iz: 29 bits for 32-bit x86)
    field x86PATBit 1
    field x86PCDBit 1
    field x86PWTBit 1
}
```
so at first glance the two don't seem to be equal.

- x86PWTBit is write-through, that matches seL4_X86_WriteBack/seL4_X86_WriteThrough.
- x86PCDBit is cache-disabled, that matches seL4_X86_CacheDisabled. seL4_X86_Uncacheable seems to be cache disabled + write through instead of write-back.
- x86PATBit is just "pat" though and at first glance doesn't seem to be write-combining.
 
Turns out it's an indirect page table attribute thing, see https://wiki.osdev.org/Paging#PAT. The actual page table attributes are looked up from `IA32_PAT_MSR` by indexing with PAT + PCD + PWT. So for PAT enabled systems, the actual meaning of the page table bits is determined by the `IA32_PAT_MSR` configuration.

`init_pat_msr()` configures `IA32_PAT_MSR` to what we want, which matches `seL4_X86_VMAttributes`, so in the end it all works out correctly.

PR #1580, opened 2026-01-07
bitfields: reduce #ifdef use with new field_ptr command

lsf37:
- allow simple arithmetic expressions in bitfield specifications (`+`, `-`, `*`, `/`, `%`, `word_size`, `canonical`)
- add new command
  ```
  field_ptr(align) name size
  ```
  which is an abbreviation for
  ```
  padding size - canonical + align
  field_high name canonical - align
  ```
  The `(align)` part is optional and defaults to 0 when left out.
- use the above to remove `BF_CANONICAL_RANGE` and related `#ifdefs` in bitfield definitions

The change produces identical code and proof output to before, and all preprocess checks should be passing.

We could go further in refactoring some of the bitfield definitions to become more regular, but I have left that for later in favour of producing identical output. It may now also be possible to share more between the generic structures_32/64.bf files, but I have also left that for later to reduce noise.

There are more things we could add to the language front end, like an if/then/else statement, letting the user define constants, and functions such as `size(field_name)`, but this change here produces a good first return for effort.

### DIFF
--- a/include/object/structures_64.bf
+++ b/include/object/structures_64.bf
@@ -5,135 +5,135 @@
 --
 
 block null_cap {
-    padding 64
+    padding        64
 
-    field capType 5
-    padding word_size - 5
+    field capType  5
+    padding        word_size - 5
 }
 
 block untyped_cap {
-    field capFreeIndex canonical_size
-    padding word_size - canonical_size - 1 - 6
-    field capIsDevice 1
-    field capBlockSize 6
+    field capFreeIndex  canonical_size
+    padding             word_size - canonical_size - 1 - 6
+    field capIsDevice   1
+    field capBlockSize  6
 
-    field capType 5
-    field_ptr capPtr word_size - 5
+    field capType       5
+    field_ptr capPtr    word_size - 5
 }
 
 block endpoint_cap(capEPBadge, capCanGrantReply, capCanGrant, capCanSend,
                    capCanReceive, capEPPtr, capType) {
-    field capEPBadge 64
+    field capEPBadge       64
 
-    field capType 5
+    field capType          5
     field capCanGrantReply 1
-    field capCanGrant 1
-    field capCanReceive 1
-    field capCanSend 1
-    field_ptr capEPPtr word_size - 5 - 4
+    field capCanGrant      1
+    field capCanReceive    1
+    field capCanSend       1
+    field_ptr capEPPtr     word_size - 5 - 4
 }
 
 block notification_cap {
-    field capNtfnBadge 64
+    field capNtfnBadge       64
 
-    field capType 5
-    field capNtfnCanReceive 1
-    field capNtfnCanSend 1
-    field_ptr capNtfnPtr word_size - 5 - 2
+    field capType            5
+    field capNtfnCanReceive  1
+    field capNtfnCanSend     1
+    field_ptr capNtfnPtr     word_size - 5 - 2
 }
 
 #ifdef CONFIG_KERNEL_MCS
 block reply_cap {
-    field capReplyPtr 64
+    field capReplyPtr       64
 
-    field capType 5
-    field capReplyCanGrant 1
-    padding word_size - 5 - 1
+    field capType           5
+    field capReplyCanGrant  1
+    padding                 word_size - 5 - 1
 }
 
 block call_stack(callStackPtr, isHead) {
-    padding word_size - 1 - 48
-    field isHead 1
-    field_ptr callStackPtr 48
+    padding                 word_size - 1 - 48
+    field isHead            1
+    field_ptr callStackPtr  48
 }
 #else
 block reply_cap(capReplyCanGrant, capReplyMaster, capTCBPtr, capType) {
-    field capTCBPtr 64
+    field capTCBPtr         64
 
-    field capType 5
-    padding word_size - 5 - 2
-    field capReplyCanGrant 1
-    field capReplyMaster 1
+    field capType           5
+    padding                 word_size - 5 - 2
+    field capReplyCanGrant  1
+    field capReplyMaster    1
 }
 #endif
 
 -- The user-visible format of the data word is defined by cnode_capdata, below.
 block cnode_cap(capCNodeRadix, capCNodeGuardSize, capCNodeGuard,
                 capCNodePtr, capType) {
-    field capCNodeGuard 64
+    field capCNodeGuard       64
 
-    field capType 5
-    field capCNodeGuardSize 6
-    field capCNodeRadix 6
-    field_ptr(1) capCNodePtr 47
+    field capType             5
+    field capCNodeGuardSize   6
+    field capCNodeRadix       6
+    field_ptr(1) capCNodePtr  47
 }
 
 block thread_cap {
-    padding 64
+    padding              64
 
-    field capType 5
-    field_ptr capTCBPtr word_size - 5
+    field capType        5
+    field_ptr capTCBPtr  word_size - 5
 }
 
 block irq_control_cap {
-    padding 64
+    padding        64
 
     field capType  5
-    padding word_size - 5
+    padding        word_size - 5
 }
 
 block irq_handler_cap {
 #ifdef ENABLE_SMP_SUPPORT
-    field capIRQ 64
+    field capIRQ   64
 #else
-    padding word_size - 12
-    field capIRQ 12
+    padding        word_size - 12
+    field          capIRQ 12
 #endif
 
     field capType  5
-    padding word_size - 5
+    padding        word_size - 5
 }
 
 block zombie_cap {
-    field capZombieID     64
+    field capZombieID    64
 
-    field capType         5
-    padding               word_size - 5 - 7
-    field capZombieType   7
+    field capType        5
+    padding              word_size - 5 - 7
+    field capZombieType  7
 }
 
 block domain_cap {
-    padding 64
+    padding        64
 
-    field capType 5
-    padding word_size - 5
+    field capType  5
+    padding        word_size - 5
 }
 
 #ifdef CONFIG_KERNEL_MCS
 block sched_context_cap {
-    field_ptr capSCPtr 48
-    field capSCSizeBits 6
-    padding 10
+    field_ptr capSCPtr   48
+    field capSCSizeBits  6
+    padding              10
 
-    field capType 5
-    padding       59
+    field capType        5
+    padding              59
 }
 
 block sched_control_cap {
-    field core    64
+    field core     64
 
-    field capType 5
-    padding       59
+    field capType  5
+    padding        59
 }
 #endif
 
@@ -141,36 +141,36 @@ block sched_control_cap {
 
 -- Endpoint: size = 16 bytes
 block endpoint {
-    field epQueue_head 64
-    field_ptr(2) epQueue_tail word_size - 2
-    field state 2
+    field epQueue_head         64
+    field_ptr(2) epQueue_tail  word_size - 2
+    field state                2
 }
 
 -- Async endpoint: size = 32 bytes (64 bytes on mcs)
 block notification {
 #ifdef CONFIG_KERNEL_MCS
-    padding 3 * word_size
-    field_ptr ntfnSchedContext word_size
+    padding                     3 * word_size
+    field_ptr ntfnSchedContext  word_size
 #endif
 
-    field_ptr ntfnBoundTCB word_size
+    field_ptr ntfnBoundTCB      word_size
 
-    field ntfnMsgIdentifier word_size
+    field ntfnMsgIdentifier     word_size
 
-    field_ptr ntfnQueue_head word_size
+    field_ptr ntfnQueue_head    word_size
 
-    field_high ntfnQueue_tail canonical_size
-    padding word_size - canonical_size - 2
-    field state 2
+    field_high ntfnQueue_tail   canonical_size
+    padding                     word_size - canonical_size - 2
+    field state                 2
 }
 
 -- Mapping database (MDB) node: size = 16 bytes
 block mdb_node {
-    field_ptr(2) mdbNext word_size - 2
-    field mdbRevocable 1
-    field mdbFirstBadged 1
+    field_ptr(2) mdbNext  word_size - 2
+    field mdbRevocable    1
+    field mdbFirstBadged  1
 
-    field mdbPrev word_size
+    field mdbPrev         word_size
 }
 
 -- Thread state data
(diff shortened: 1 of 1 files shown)


tools/bitfield_gen.md
```diff
@@ -158,6 +158,31 @@ means, not necessarily all of the combinations above will appear in the
...
+been written as follows, assuming a `base 32` definition for `word_size.
+
+```bf_gen
+block VMFault {
+    field     address           word_size
+
+    field     FSR               5
+    padding                     7
+    field     instructionFault  1
+    padding                     word-size - (5 + 7 + 1) - 4
```
lsf37:
```suggestion
    padding                     word_size - (5 + 7 + 1) - 4
```

include/object/structures_64.bf
```diff
@@ -8,32 +8,17 @@ block null_cap {
...
-#if BF_CANONICAL_RANGE == 48
-    field capFreeIndex 48
-    padding 9
-#elif BF_CANONICAL_RANGE == 39
-    field capFreeIndex 39
-    padding 18
-#else
-#error "Unspecified canonical address range"
-#endif
+    field capFreeIndex canonical
```
midnightveil: personally I find this rather confusing, I read canonical here as a keyword like `field` or such as opposed to a constant that becomes a number. same when added to numbers, the expressions like `field xxxx 5 + 3`.. breaks my brain a bit. 

suggestion: enforce brackets around compind expressions (`field xxx (5+3)` and use a symbol to denote constants like `field capFreeIndex @canonical`.

I don't exactly want to bikeshed this, and I don't think my suggestions entirely improve this perfectly but it's a bit hard to distinguish when we have more alphabetical text on the same line now.

tools/bitfield_gen.md
```diff
@@ -227,6 +252,96 @@ that stores 36 bits (on a 64 bit base platform with 48 significant bits).
...
+### Field Ptr
+
+Often we want to store a a pointer with some known alignment (i.e. not all bits
+need to be stored), and pad out to a specific size. The `field_ptr` abbreviation
+makes this convenient:
+
+```bf_gen
+block some_block {
+    ...
+    field_ptr(align) some_name size
```
midnightveil, line 264: OK, I see why this was added but I really don't like it. 

Especially the part where it contains implicit padding. 

I like the part where we can store a pointer of $base size with an alignment, that's neat, but then also having padding up to size... that makes it hard to track what's going on. I'd rather have to compute the padding field here in this case.

(and adding field_ptr with alignment should really just replace all the field_high IMO which is only used for pointers I think anyway?)

If we're going the path of having implicit padding I think it would make more sense to add an expandable padding which grows to fill up to the word (caveat: might need an extra boundary marker so can know which field to stop allowing past, for the commonish case of "padding; other_field; capType".

midnightveil (commented):
some thoughts...

include/object/structures_64.bf
Indanz: I think those are good suggestions.

Edit: Perhaps it's tricky to enforce parentheses by the parser, but we can at least use them everywhere.

include/object/structures_64.bf
```diff
@@ -8,32 +8,17 @@ block null_cap {
...
+    padding word_size - canonical - 1 - 6
```
Indanz:
```suggestion
    field capFreeIndex word_size - 1 - 6
```
Just noting it here as future improvement, what you have now retains the same bit locations.

include/object/structures_64.bf
```diff
@@ -203,80 +142,35 @@ block sched_control_cap {
...
-#elif BF_CANONICAL_RANGE == 39
-    field_high ntfnQueue_tail 39
-    padding 23
-#else
-#error "Unspecified canonical address range"
-#endif
+    field_ptr ntfnQueue_head word_size
+
+    field_high ntfnQueue_tail canonical
+    padding word_size - canonical - 2
```
Indanz:
```suggestion
    field_high ntfnQueue_tail word_size - 2
```
Future improvement.

tools/bitfield_gen.md
```diff
@@ -227,6 +252,96 @@ that stores 36 bits (on a 64 bit base platform with 48 significant bits).
 
 [Architecture Parameters]: #architecture-parameters-and-canonical-pointers
 
+### Field Ptr
+
+Often we want to store a a pointer with some known alignment (i.e. not all bits
```
Indanz:
```suggestion
Often we want to store a pointer with some known alignment (i.e. not all bits
```

tools/bitfield_gen.md
```diff
@@ -227,6 +252,96 @@ that stores 36 bits (on a 64 bit base platform with 48 significant bits).
...
+base 64(39,1)
+
+block mdb_node {
+    padding 25
+    field_high mdbNext 37
+    field mdbRevocable 1
+    field mdbFirstBadged 1
+
+    field mdbPrev 64
+}
```
Indanz, line 343: This bit doesn't really add anything useful I think, perhaps better to leave it out, because it is kind of obvious. The same is true for word_size and you don't have an elaborate example there either.

tools/bitfield_gen.md
```diff
@@ -594,17 +714,29 @@ canonical_spec ::= "(" INTLIT "," ("0"|"1") ")"
...
+       | "-" expr | "(" expr ")" | INTLIT
+       | "word_size" | "canonical"
 ```
 
 Forward references to names that are declared later in the file are allowed.
+Multiplicative operators "*", "/", "%" bind stronger than "+" and "-". Binary
+operators are left-associative. Unary minus binds stronger than all of the
+other operators. The constants "word_size" and "canonical" are only available
+in "field", "field_high", and "padding" expressions and refer to the most recent
+corresponding "base" declaration.
```
Indanz:
```suggestion
other operators. The constants "word_size" and "canonical" are only available
in "field", "field_high", "field_ptr" and "padding" expressions and refer to the most recent
corresponding "base" declaration.
```
However, you've said this already above, so not sure if it's worth repeating here.

tools/bitfield_gen.py
```diff
@@ -213,17 +225,31 @@ def p_fields_empty(t):
...
+    padding_size = ('-', t[4], "canonical")
+    field_size = "canonical"
+    t[0] = t[1] + [(None, padding_size, False), (t[3], field_size, True)]
+
+
+def p_fields_field_ptr_align(t):
+    """fields : fields FIELD_PTR LPAREN int_expr RPAREN IDENTIFIER int_expr"""
+    padding_size = ('+', ('-', t[7], "canonical"), t[4])
+    field_size = ('-', "canonical", t[4])
+    t[0] = t[1] + [(None, padding_size, False), (t[6], field_size, True)]
```
Indanz, line 250:
```suggestion
def p_fields_field_ptr_align(t):
    """fields : fields FIELD_PTR LPAREN int_expr RPAREN IDENTIFIER int_expr"""
    align = t[4]
    padding_size = ('+', ('-', t[7], "canonical"), align) # size - canonical + align
    field_size = ('-', "canonical", align)
    t[0] = t[1] + [(None, padding_size, False), (t[6], field_size, True)]
```
To maintain sanity somewhat.

tools/bitfield_gen.py
```diff
@@ -87,17 +87,29 @@ def var_name(name, base):
...
 tokens = reserved + ('IDENTIFIER', 'INTLIT', 'LBRACE', 'RBRACE',
-                     'LPAREN', 'RPAREN', 'COMMA')
+                     'LPAREN', 'RPAREN', 'COMMA', 'PLUS', 'MINUS',
+                     'MULT', 'DIV', 'MOD')
+
+precedence = (
+    ('left', 'PLUS', 'MINUS'),
+    ('left', 'MULT', 'DIV', 'MOD'),
+    ('right', 'UMINUS')
+)
```
Indanz, line 102: This isn't used anywhere, is it some special property that the tool knows about?

tools/bitfield_gen.py
```diff
@@ -289,14 +315,78 @@ def p_masks_empty(t):
...
+
+
+def eval_expr(t, const_map={}):
+    if isinstance(t, int):
+        return t
+    if isinstance(t, str):
+        if t in const_map:
+            return const_map[t]
+        raise ValueError(f"Unknown constant '{t}'")
+    if len(t) == 2:
```
Indanz, line 371: Should we double check here that t[0] == '-'?

tools/bitfield_gen.md
Indanz: I disagree. It shows the intention very clearly: Defining a pointer field of a certain size. You can see in one glance how big the whole thing is in total. The padding it generates is the padding needed to hide the differences in canonical sizes.

Doing that manually would be very ugly and would greatly increase the instances of `canonical` and `word_size` used everywhere, which isn't nice either. While now it's minimal (all explicit usages of `canonical` can be replaced by shifting the fields, because it's a matter of swapping the order of padding and the pointer).

Yes, it can also generate additional padding, and maybe doing that makes it harder to see what's going on, but you don't have to use it like that. This is no difference than other field types being bigger than necessary to avoid adding manual padding, so your objection applies to all types, not just field_ptr.

Enforcing no unnecessary padding for field_ptr could be done by the parser, but is a bit tricky as it depends on the maximum size of `canonical` for that block, not the currently configured one. Only way around that is to add it as an option to `base()`, which we can do in the future if that really improves things. But usually there isn't much else in the same word, otherwise the pointer wouldn't fit, so I think it's fine to let field_ptr add the necessary padding, instead of adding it explicitly everywhere.

> (and adding field_ptr with alignment should really just replace all the field_high IMO which is only used for pointers I think anyway?)

Yes, more or less. But it would be awkward for page table structures which know exactly what they have already and replacing those uses doesn't simplify anything.

> If we're going the path of having implicit padding I think it would make more sense to add an expandable padding which grows to fill up to the word (caveat: might need an extra boundary marker so can know which field to stop allowing past, for the commonish case of "padding; other_field; capType".

Probably true, but tricky to implement, with less gain. Let's start with getting rid of all the horrible BF_CANONICAL_RANGE ifdefs first, those really make the code much less readable.

Indanz (approved):
Thanks for doing this! Other than some nitpicks it looks good to me.

Its great to get rid of those horrible BF_CANONICAL_RANGE ifdefs.

include/object/structures_64.bf
lsf37: Sorry, I don't like either of these. I'm coming from ML-like languages and enforcing unnecessary parentheses is anathema to me. I have never seen a case where it made things better. And adding symbols for constants just makes them look like something special, but they are not.

I think this could relatively easily be made readable in the current syntax by using white space to separate and align the size field as it was originally intended. I don't think we should enforce this in the parser, though.

include/object/structures_64.bf
lsf37: That would not be correct in the sense that setting `capFreeIndex` to size `canonical` enforces an invariant on the max size of `capFreeIndex` in storage. If you relax that, you'd need to explicitly prove that whatever was stored there does not exceed that size and is therefore safe to use in the rest of code. With the size constraint, you get that proof for free.

include/object/structures_64.bf
lsf37: Yes, I specifically didn't do that one to keep it identical. Unless there is some manual stuff going on with notification encoding in the fast path it would be safe to change this from the proof side.

tools/bitfield_gen.md
lsf37: I still think this is useful, I specifically wanted examples with explicit numbers to confirm intuition for people who read this for the first time. We've thought through this now and therefore it looks all obvious, but if you come at it fresh (not just this part, but the whole bitfield generator) this construct mixes a lot of the concepts that people tend to get confused about.

tools/bitfield_gen.md
lsf37: > However, you've said this already above, so not sure if it's worth repeating here.

I can't see in the GitHub UI which section this is in, but if this is the syntax reference part, then it should be repeated. It's a slightly arbitrary restriction, that's why I think it's important to document. (The cause is that implementing the evaluation inside TaggedUnion is super annoying in the current code structure, and it's very unlikely to be useful there, so I left it)

tools/bitfield_gen.py
lsf37: Yes, this is a yacc thing, like `tokens` (they are all used). I'll add a comment.

tools/bitfield_gen.py
lsf37: Probably better style to do that, yes. It is currently guaranteed by the parser, but checking would make it more robust for extensions later.

include/object/structures_64.bf
lsf37: I think what makes this look weird is that `canonical` doesn't look like a size and therefore the brain goes "keyword". How about we rename it to `canonical_size`?

tools/bitfield_gen.md
lsf37: > OK, I see why this was added but I really don't like it.
> 
> Especially the part where it contains implicit padding.

I do see where you are coming from, but if you look at what this replaces, then "store a pointer, fill with padding to size x" is exactly what we want to be doing all over the place in bf structs that have to accommodate different pointer sizes. 

I briefly looked at a `padding fill` command while doing this, but it's problematic: you would either have to enforce that only one of these existed per word in the block, or you'd have to give it some semantics like "the first one fills, the rest become 0". All of that makes it very hard to see what offsets other fields are at, which is what you usually want to know. It would also make it non-obvious where word boundaries are in a block, i.e. which word the fill belongs to.

With `field_ptr`, all of that is stable, because it says what overall space it takes -- it removes precisely the variability that you get from varying pointer sizes over different architectures and makes those slots constant width, which was previously very hard to see. And it removes code duplication that people can get wrong, of course. It does make it less obvious how much storage the pointer itself gets and therefore whether there is potentially space left to use, but as an abstraction "field of size x that stores a pointer" I think it has nice properties.

> (and adding field_ptr with alignment should really just replace all the field_high IMO which is only used for pointers I think anyway?)

If we removed the padding, then we could instead add an alignment parameter to `field_high` which effectively would be the same thing, yes. But I don't think we should remove the padding, and therefore `field_high` is still needed as the primitive. It also enables you to make padding explicit if that is important for the use case. E.g. if there is a block where space is tight and you want to make explicit that the padding is 0 for the largest canonical pointer sizes or something like that.

 Indan writes:
> Yes, it can also generate additional padding, and maybe doing that makes it harder to see what's going on, but you don't have to use it like that. This is no difference than other field types being bigger than necessary to avoid adding manual padding, so your objection applies to all types, not just field_ptr.

Good point, we do actually use `field` like that a whole bunch of times. I don't particularly like that usage, but I don't think the tool should forbid it.

midnightveil: I think I'm fine with this on further consideration. Thinking it about it like this:

```
field_ptr(«align») «fieldName» [padded to] «size»
```

I'd mark this thread as resolved but it seems like I can't.

include/object/structures_64.bf
midnightveil: Maybe renaming helps a bit? I don't know how much though.

I'd have to see what adding space between the field name and the size of that helps, but it might do so (almsot certainly instead of brackets for expressions).

That, or it gets made clear but having syntax highlighting in the end.

(This probably doesn't really matter that much in the end. If you disagree I'm happy if you just ignore this personally - it's not worth blocking this over).

tools/bitfield_gen.md
lsf37: I might try to use that phrasing in the docs, I think that is helpful

include/object/structures_64.bf
Indanz: Sure, but we do that kind of stuff everywhere else, I don't see why `capFreeIndex` is different or special. How is this solved for other pointers?

tools/bitfield_gen.md
Indanz: Ah, I missed that the repetition was deliberate and the whole point was to show that both 48 and 39 canonical sizes can be described the same. With that in mind, I suggest swapping the blocks around and changing the text to "is also equivalent to". (And to keep it consistent, perhaps swap the blocks around earlier too.)

tools/bitfield_gen.md
Indanz: Maybe say it's not available in tagged unions instead of repeating the same text, to make your point clearer? If that is what this is about, because it's not entirely clear to me what restriction you mean.

tools/bitfield_gen.py
Indanz: Because it's the only 'right' precedence? In that case, I think it's fine to leave as-is. No point in checking the impossible.

tools/bitfield_gen.md
lsf37: It's also in `base` declarations and potentially other places. The restriction is just that constants are not allowed in those expressions. The evaluation of constants needs context (what the current base/canonical bits are), and that context is not available during the parse run yet, so it has to be first stored, then evaluated later. I did that reshuffling for the `Block` class, but it's not worth doing all that refactoring for the other places that admit expressions, because you don't usually need to refer to those constants there.

tools/bitfield_gen.py
lsf37: Ok, I feel lazy today, will leave it :-)

include/object/structures_64.bf
lsf37: It mostly depends on whether the proofs currently care about the max size or not (or if they are already forced to prove a separate invariant). For this one it is currently used.

include/object/structures_64.bf
```diff
@@ -5,278 +5,172 @@
...
-    padding 23
-#else
-#error "Unspecified canonical address range"
-#endif
-    field state 2
+    field_ptr ntfnQueue_head    word_size
+
+    field_high ntfnQueue_tail   canonical_size
+    padding                     word_size - canonical_size - 2
+    field state                 2
```
lsf37, line 164: I've added a commit with white-space only change that shows how it looks like aligned (and have renamed to `canonical_size`, which hopefully we don't have to mention very often). I think this is fairly readable now.

midnightveil: Yes, that's much better. :)

lsf37: Excellent.

If it wasn't for the annoying preprocessor stuff, it would be fairly easy to write a pretty-printer/formatter for this. I guess it could still be pretty easy, but it would be a slightly different parser. I'm almost tempted..

lsf37: On second thought, writing a formatter would still be quite a bit of work, so probably not..

lsf37:
If there are no further comments, I would merge this. We can add more .bf file cleanup things in a separate PR if we feel like it.

midnightveil:
I'm happy.

tools/bitfield_gen.md
```diff
@@ -227,6 +252,100 @@ that stores 36 bits (on a 64 bit base platform with 48 significant bits).
...
+    ...
+}
+```
+
+The `(align)` parameter can be left out and is equivalent to `field_ptr(0)`. The
+constant `canonical_size` in the expression above stands for the number of
+significant bits (the canonical size) of a pointer according to the most recent
+`base` definition. For example the long form
+
+```bfgen
```
Indanz: Why is this one bfgen instead of bf_gen like the others?

tools/bitfield_gen.md
```diff
@@ -227,6 +252,100 @@ that stores 36 bits (on a 64 bit base platform with 48 significant bits).
...
+}
+```
+
+The `field_ptr` abbreviation is especially useful for sharing the same block
+definition under different `base` declarations, e.g. for different
+architectures. For example, with a `base 64(39,1)` the `field_ptr` definition
+text remains the same, but expands to a different `field_high`/`padding`
+combination that achieves the same overall width:
+
+```bfgen
```
Indanz: Same here.

tools/bitfield_gen.md
lsf37: because it's a typo :-), thanks for spotting it

It does not actually have any effect at the moment, because GitHub doesn't have a highlighter for it yet, but I was planning to add one in the future.

Issue #1566, opened 2025-11-27
clock sync test fails on SMP x86 whe running with QEMU


Ivan-Velickovic:
Similar to QEMU ARM/RISC-V, `ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT` should not be on for x86 QEMU. But since it's x86 the kernel doesn't know it's running on QEMU when you build the kernel for a generic x86 CPU.

According to my colleague @dreamliner787-9 there's no reliable way of looking at CPUID to tell if you're being virtualised by QEMU.

I'm wondering if anyone has any other ideas for how we could skip the test at run-time if we detect we're running on QEMU, I'm still getting familiar with the x86 architecture myself.

Indanz:
Usually x86 is fast enough to pass the test even on QEMU because it has a 1us margin.

Ivan-Velickovic:
It happens to me on both a Linux x86-64 host and macOS ARM64 host regularly.

I've put up https://github.com/seL4/seL4/pull/1581, I guess alternatively we could try increase the margin but skipping the test seems more reliable.

PR #1581, opened 2026-01-09
x86: skip clock_sync_test on QEMU

Ivan-Velickovic:
Try detect if we are running as a guest based on the CPUID and skip clock_sync_test if we are.

Similar as previous patches for skipping QEMU on ARM/RISC-V but at run-time instead of build-time.

Tested with qemu-system-x86_64 using `-smp 4`, the test is now skipped with TCG QEMU as well as `-accel kvm`.

Closes https://github.com/seL4/seL4/issues/1566.

### DIFF
--- a/include/arch/x86/arch/machine.h
+++ b/include/arch/x86/arch/machine.h
@@ -96,6 +96,23 @@
 
 #define IA32_PRED_CMD_MSR                   0x49
 
+/*
+ * CPUID bits for detecting the kernel is running as a guest.
+ *
+ * These bits correspond to the strings "KVMKVMKVM" for KVM and "TCGTCGTCGTCG"
+ * for QEMU's Tiny Code Generator (TCG).
+ *
+ * https://docs.kernel.org/virt/kvm/x86/cpuid.html.
+ */
+#define KVM_CPUID_SIGNATURE 0x40000000
+
+#define CPUID_TCG_EBX 0x54474354
+#define CPUID_TCG_ECX 0x43544743
+#define CPUID_TCG_EDX 0x47435447
+#define CPUID_KVM_EBX 0x4b4d564b
+#define CPUID_KVM_ECX 0x564b4d56
+#define CPUID_KVM_EDX 0x4d
+
 word_t PURE getRestartPC(tcb_t *thread);
 void setNextPC(tcb_t *thread, word_t v);
 
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -578,6 +578,21 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
 }
 
 #ifdef ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT
+BOOT_CODE static bool_t hypervisor_present(void)
+{
+#ifdef CONFIG_ARCH_X86
+    uint32_t ebx = x86_cpuid_ebx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t ecx = x86_cpuid_ecx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t edx = x86_cpuid_edx(KVM_CPUID_SIGNATURE, 0);
+
+    if ((ebx == CPUID_KVM_EBX && ecx == CPUID_KVM_ECX && edx == CPUID_KVM_EDX)
+        || (ebx == CPUID_TCG_EBX && ecx == CPUID_TCG_ECX && edx == CPUID_TCG_EDX)) {
+        return true;
+    }
+#endif
+    return false;
+}
+
 BOOT_CODE void clock_sync_test(void)
 {
     ticks_t t, t0;
@@ -593,7 +608,16 @@ BOOT_CODE void clock_sync_test(void)
     t = getCurrentTime();
     printf("clock_sync_test[%d]: t0 = %"PRIu64", t = %"PRIu64", td = %"PRIi64"\n",
            (int)getCurrentCPUIndex(), t0, t, t - t0);
-    assert(t0 <= margin + t && t <= t0 + margin);
+    /*
+     * The test does not consistently work if we are in a virtual machine (e.g
+     * within QEMU) because the measurement cannot distinguish between
+     * interrupted clock reads and out-of-sync clocks.
+     */
+    if (hypervisor_present()) {
+        printf("clock_sync_test[%d]: disabled, detected running as VM\n", (int)getCurrentCPUIndex());
+    } else {
+        assert(t0 <= margin + t && t <= t0 + margin);
+    }
 }
 #endif
 


src/kernel/boot.c
```diff
@@ -580,6 +580,20 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
...
+     * clock_sync_test does not consistently work on QEMU, try and detect if we
+     * are running as a guest and skip the test.
+     */
+    uint32_t ebx = x86_cpuid_ebx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t ecx = x86_cpuid_ecx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t edx = x86_cpuid_edx(KVM_CPUID_SIGNATURE, 0);
+
+    if ((ebx == CPUID_KVM_EBX && ecx == CPUID_KVM_ECX && edx == CPUID_KVM_EDX)
+        || (ebx == CPUID_TCG_EBX && ecx == CPUID_TCG_ECX && edx == CPUID_TCG_EDX)) {
+        return;
```
axel-h: Given this only runs of `ENABLE_SMP_CLOCK_SYNC_TEST_ON_BOOT` is set, you may want to print a note here that this is deliberately skipped because QEMU was detected.
```
printf("clock_sync_test[%d]: disabled, QEMU was detected\n", (int)getCurrentCPUIndex());
```

Ivan-Velickovic: Sure thing, done.

axel-h:
> The purpose of the clock sync test is to test whether our hardware assumptions are correct. Those go out of the window when running virtualised in any form anyway.

I wonder if this is something that is really up to what is required from the virtualization. So I am a bit lost now what we expect to hold to avoid running into real functional problems vs. having to accept that in some well known configurations some things are just not guaranteed any longer, but still the system has to work in this cases.

Indanz:
> I wonder if this is something that is really up to what is required from the virtualization. So I am a bit lost now what we expect to hold to avoid running into real functional problems vs. having to accept that in some well known configurations some things are just not guaranteed any longer, but still the system has to work in this cases.

With virtualisation, there can be long running scheduling latencies, while the test code assumes it's the only thing running and can't distinguish between just a very slow read of the clock and actually being out of sync, so we can't easily measure this. Instead of adding work-around, just disabling the clock sync test is fine. At this point we could remove the clock sync test altogether, but then if there is new hardware with truly out of sync clock, we won't detect it.

In some corner cases MCS used to read the time cross-core (e.g. when moving a task from another core), however, the cross core timestamp is never up-to-date, so this was wrong, see https://github.com/seL4/seL4/issues/854. Once that got fixed and local core time was used, we do rely on the clocks being synchronous on all cores. Any clocking error shows itself as wrong time bookkeeping for tasks that are affected. In the worst case they get a full period too much or too little charged (assuming the clocks are somewhat in sync). If the clocks are not in sync at all you effectively can't move tasks across cores safely.

Ivan-Velickovic:
> The same is true for running virtualised as a guest on real hardware, not just QEMU, so I suggest broadening the check.

I don't know if there is a generic way to test if the kernel is being virtualised, did you have something in mind?

src/kernel/boot.c
```diff
@@ -580,6 +580,21 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
...
+    uint32_t ebx = x86_cpuid_ebx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t ecx = x86_cpuid_ecx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t edx = x86_cpuid_edx(KVM_CPUID_SIGNATURE, 0);
+
+    if ((ebx == CPUID_KVM_EBX && ecx == CPUID_KVM_ECX && edx == CPUID_KVM_EDX)
+        || (ebx == CPUID_TCG_EBX && ecx == CPUID_TCG_ECX && edx == CPUID_TCG_EDX)) {
+        printf("clock_sync_test[%d]: disabled, QEMU was detected\n", (int)getCurrentCPUIndex());
+        return;
+    }
+#endif
```
Indanz, line 592: Can you put this into its own helper function and use it to skip the assert only instead of the whole test? It's still good to see what the clock differences are. And make the text more broad, like "KVM or QEMU detected".

Just define it to something empty for non-x86.

Indanz:
> > The same is true for running virtualised as a guest on real hardware, not just QEMU, so I suggest broadening the check.
> 
> I don't know if there is a generic way to test if the kernel is being virtualised, did you have something in mind?

No, but you're already checking for KVM and assuming it's QEMU, while it also applies if it is actually a KVM guest. I didn't find anything similar for Xen, VMware and others though.

include/arch/x86/arch/machine.h
```diff
@@ -96,6 +96,23 @@
 
 #define IA32_PRED_CMD_MSR                   0x49
 
+/*
+ * CPUID bits for detecting the kernel is running as a guest.
+ *
+ * These bits correspond to the strings "KVMKVMKVM" for KVM and "TCGTCGTCGTCG"
+ * for QEMU's TCG.
```
Indanz: Can you spell it out to "QEMU's Tiny Code Generator (TCG)"?

src/kernel/boot.c
```diff
@@ -578,6 +578,22 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
...
+    uint32_t ebx = x86_cpuid_ebx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t ecx = x86_cpuid_ecx(KVM_CPUID_SIGNATURE, 0);
+    uint32_t edx = x86_cpuid_edx(KVM_CPUID_SIGNATURE, 0);
+
+    if ((ebx == CPUID_KVM_EBX && ecx == CPUID_KVM_ECX && edx == CPUID_KVM_EDX)
+        || (ebx == CPUID_TCG_EBX && ecx == CPUID_TCG_ECX && edx == CPUID_TCG_EDX)) {
+        return true;
+    }
+#endif
+
```
Indanz: Please remove this empty line. It cuts the function in two floating parts which jars the brain.

src/kernel/boot.c
```diff
@@ -593,7 +609,16 @@ BOOT_CODE void clock_sync_test(void)
     t = getCurrentTime();
     printf("clock_sync_test[%d]: t0 = %"PRIu64", t = %"PRIu64", td = %"PRIi64"\n",
            (int)getCurrentCPUIndex(), t0, t, t - t0);
-    assert(t0 <= margin + t && t <= t0 + margin);
+
```
Indanz: Same for this line. There is absolutely no need to add such empty lines, they don't improve readability.

src/kernel/boot.c
```diff
@@ -593,7 +609,16 @@ BOOT_CODE void clock_sync_test(void)
...
+    /*
+     * clock_sync_test does not consistently work when being virtualised (e.g within QEMU),
+     * skip the assert if we are in a virtual machine.
```
Indanz:
```suggestion
     * The test does not consistently work if we are in a virtual machine (e.g within QEMU), because
     * the measurement cannot distinguish between interrupted clock reads and out-of-sync clocks.
```

Indanz:
I can't merge it until it's up-to-date. :-/

PR #1584, opened 2026-01-15
debug: align debug_dumpScheduler output

midnightveil:
Using tabs gives viewer-defined alignment (4 or 8 tabs), and even fixing that the alignment was off as the max sizes of various values did not match with the actual sizes.

The only case where we can see non-alignment now is if the name of a thread is longer than ~40 characters.

Also, print the domain when we have multiple domains.

Now it looks like:

```
Dumping all tcbs!
Name                                       State              IP                   Prio   Core   InReleaseQueue
----------------------------------------------------------------------------------------------------------------
idle_thread                                idle               0                       0      0   no
sel4test-driver                            running            0x456604              255      0   no



```

### DIFF
--- a/include/api/debug.h
+++ b/include/api/debug.h
@@ -77,7 +77,6 @@ static inline void debug_printUserState(void)
 
 static inline void debug_printTCB(tcb_t *tcb)
 {
-    printf("%40s\t", TCB_PTR_DEBUG_PTR(tcb)->tcbName);
     char *state;
     switch (thread_state_get_tsType(tcb->tcbState)) {
     case ThreadState_Inactive:
@@ -113,10 +112,23 @@ static inline void debug_printTCB(tcb_t *tcb)
         fail("Unknown thread state");
     }
 
-    word_t core = SMP_TERNARY(tcb->tcbAffinity, 0);
-    printf("%15s\t%p\t%20lu\t%lu", state, (void *) getRestartPC(tcb), tcb->tcbPriority, core);
+    /* 40: (arbitrary) max print-length of name
+     * 16: max length of state
+     * 18: length of `0x + max 64 bit number'
+     * 4:  length of "Prio" header
+     * 4:  length of "Core" header
+     * 4:  length of "Dom" header, but +1 for consistency with Prio/Core headers
+     * 14: length of "InReleaseQueue" header */
+    printf("%-40s   %-16s   %-18p   %4lu", TCB_PTR_DEBUG_PTR(tcb)->tcbName, state,
+           (void *) getRestartPC(tcb), tcb->tcbPriority);
+#ifdef CONFIG_ENABLE_SMP_SUPPORT
+    printf("   %4lu", tcb->tcbAffinity);
+#endif
+#if CONFIG_NUM_DOMAINS > 1
+    printf("   %4lu", tcb->tcbDomain);
+#endif
 #ifdef CONFIG_KERNEL_MCS
-    printf("\t%lu", (word_t) thread_state_get_tcbInReleaseQueue(tcb->tcbState));
+    printf("   %-14s", thread_state_get_tcbInReleaseQueue(tcb->tcbState) ? "yes" : "no");
 #endif
     printf("\n");
 }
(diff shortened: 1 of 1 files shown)


Indanz:
That bugged me for a while, I had a quick fix somewhere locally, but forgot where.

> Now it looks like:
> 
> ```
> Dumping all tcbs!
> Name                                        State              IP                   Prio    Dom   InReleaseQueue
> ----------------------------------------------------------------------------------------------------------------
>                               idle_thread               idle                    0      0      0               no
>                           sel4test-driver            running             0x456ac8    255      0               no
> ```

State and IP don't align properly, so I guess you forgot to adjust the header.

It would all be more readable if we would limit task names to a reasonably short something,
40 characters is too much, I'd go for 20 or so.

include/api/debug.h
```diff
@@ -113,20 +113,36 @@ static inline void debug_printTCB(tcb_t *tcb)
...
+    printf("   %16s   %18p   %4lu", state, (void *) getRestartPC(tcb), tcb->tcbPriority);
+#ifdef CONFIG_ENABLE_SMP_SUPPORT
+    printf("   %4lu", tcb->tcbAffinity);
+#endif
+#if CONFIG_NUM_DOMAINS > 1
+    printf("   %4lu", tcb->tcbDomain);
+#endif
 #ifdef CONFIG_KERNEL_MCS
-    printf("\t%lu", (word_t) thread_state_get_tcbInReleaseQueue(tcb->tcbState));
+    printf("   %14s", thread_state_get_tcbInReleaseQueue(tcb->tcbState) ? "yes" : "no");
```
Indanz:
```suggestion
    printf("   %4s", thread_state_get_tcbInReleaseQueue(tcb->tcbState) ? "yes" : "no");
```
14 seems excessive.

include/api/debug.h
```diff
@@ -113,20 +113,36 @@ static inline void debug_printTCB(tcb_t *tcb)
...
     printf("\n");
 }
 
 static inline void debug_dumpScheduler(void)
 {
     printf("Dumping all tcbs!\n");
-    printf("Name                                    \tState          \tIP                  \t Prio \t Core%s\n",
-           config_set(CONFIG_KERNEL_MCS) ?  "\t InReleaseQueue" : "");
-    printf("--------------------------------------------------------------------------------------\n");
+    printf("Name                                        State              IP                   Prio%s%s%s\n",
```
Indanz: You probably want to use `%Ns` for the headers with the same widths as in the other prints, so it's easier to align them properly.

include/api/debug.h
midnightveil: It's to align with the header.

midnightveil:
> State and IP don't align properly, so I guess you forgot to adjust the header.

They do? Unless you want them right aligned instead of left aligned. Consider the "blocked on reply" state case and an IP well into the 64 bit range.

Indanz:
> They do? Unless you want them right aligned instead of left aligned.

I suppose they kinda do, but it doesn't look like it because everything else is right aligned, making it look messy and wrong to me.

Normally you'd align the columns content the same as their headers. In this case that would make sense for Name, State and InReleaseQueue, but not for the rest.

midnightveil:
Yeah, I think it does make more sense for these to be left aligned.

include/api/debug.h
midnightveil: Done. Unfortunately can't do this for the `-`.

midnightveil:
I have another PR that I will make to make `debug_dumpScheduler` print out scheduling information of each core that I used in #1662 that I think I should upstream too, as currently only printing the current core is annoying.

But I'll do this after this one.

lsf37:
Can you post a final "this is how it now looks" for posterity?

midnightveil:
I updated the PR description.

lsf37 (approved):
Nice!

Indanz (approved):
Big improvement, just some nitpicks.

- Style guide prefers /* */ comments.
- I'd prefer all the comments in one place for better overview instead of scattered around.
- `len('Prio')` looks very weird in C code. The C version of this would be `strlen("Prio")`.
- Why the random three spaces as prefix instead of just one or none? That looks weird and makes things wider than necessary.

Indanz:
> I have another PR that I will make to make `debug_dumpScheduler` print out scheduling information of each core that I used in #1662 that I think I should upstream too, as currently only printing the current core is annoying.
> 
> But I'll do this after this one.

I'm fine with you doing it now too, main work is checking the output and you need to do that already.

The two times I used this function I ran into the current core only problem, very annoying.

midnightveil:
> Why the random three spaces as prefix instead of just one or none? That looks weird and makes things wider than necessary.

None would mean it looks like NamePrioIP. One looks too bunched up. 
3 is what it would be if you had vertical bars between columns.

ACK on the others, I'll fix tomorrow.

> I'm fine with you doing it now too, main work is checking the output and you need to do that already.

If I was starting this PR from scratch I would've, but this is basically done right now so I'll do separately.

PR #1592, opened 2026-02-12
Add support for stm32mp2 SoC family

bruelc:
Hello, 

This is the first patch in a series to introduce STM32MP2 support to the kernel and VM CAmkES projects
Depends on PR [Add an option to restrict physical address mappings in elfloader](https://github.com/seL4/seL4_tools/pull/244)

### Description
The STM32MP Soc family is based on ARM Cortex-A35 and M33 cores. seL4 uses the A35 as the primary boot core with TF-A. Resources are managed by a RIF an accessed by OP-TEE with SMC calls. 
Boot loader is currently managed with BL33 u-boot 

The kernel is loaded at 0x84000000, after the 64MB of reserved secure DDR starting from 0x80000000

### Changes
Add STM32MP2 definitions, UART driver, and stm32m257f-ev1 evaluation board device tree.

### Documentation
- Will be provided with a seL4 documentation PR.
- STM32MP2 TRM and STM32MP257-EV1 board technical references can be found in [stm32mp257f-ev1](https://www.st.com/en/evaluation-tools/stm32mp257f-ev1.html#documentation)

### Testing
tested with **-DPLATFORM=stm32mp2 -DAARCH64=1** with the configuration options **-DSMP=on -DNUM_NODE=2 -DARM_HYP=TRUE**

for **RELEASE** and **DEBUG** builds:

- seL4 test
- seL4 bench
- CAmkES vm_minimal and vm_cross_connector examples

Thank you for your comments,


src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,44 @@
...
+static const int uart1_irqs[] = {UART1_IRQ, -1};
+
+#define UART_DEFN(devid) {          \
+    .id      = UART##devid,         \
+    .paddr   = UART##devid##_PADDR, \
+    .size    = BIT(12),             \
+    .irqs    = uart##devid##_irqs,  \
+    .init_fn = &uart_init           \
+}
+
```
midnightveil: Just so you're aware: [the seL4 kernel code style](https://sel4.systems/Contribute/style.html#verification-requirements) applies, which excludes the existence of function pointers as they are outside the verification subset

src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * Contains the definition for all character devices on this platform.
+ * Currently this is just a simple patch.
+ */
+
+#include "../../chardev.h"
```
midnightveil: These include paths should follow the usual convention (e.g. `#include <plat/xxx.h>`), not these `../` paths. Also you include `chardev.h` twice.

src/plat/stm32mp2/chardev.c
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2026, STMicroelectronics
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+/**
+ * Contains the definition for all character devices on this platform.
+ * Currently this is just a simple patch.
+ */
```
midnightveil: This is obvious, the comment can be removed.

libsel4/sel4_plat_include/stm32mp2/sel4/plat/api/constants.h
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#pragma once
+
+#include <sel4/config.h>
+#include <sel4/arch/constants_cortex_a35.h>
+
+#ifdef CONFIG_ARCH_AARCH32
+/* First address in the virtual address space that is not accessible to user level */
```
midnightveil: These comments are probably unnecessary. The `#ifdef` itself might be too, since that should be solved by the CMake platforms.

src/plat/stm32mp2/config.cmake
```diff
@@ -0,0 +1,56 @@
...
+#
+
+declare_platform(stm32mp2 KernelPlatformMP2 PLAT_STM32MP2 KernelSel4ArchAarch64)
+
+set(c_configs PLAT_STM32MP25_EV1)
+set(
+  cmake_configs
+  KernelPlatformMP25_EV1
+)
+set(plat_lists stm32mp257f-ev1)
```
midnightveil, line 11: We have a different way of handling this usually, have a look at the Zcu102/Zynqmp platforms where I think there's a KernelARMPlatforms that can have subplatforms. We don't tend to use CMake like this (for better or for worse)

bruelc: thank you for your review Julia, I will address your comments in a v2

src/plat/stm32mp2/chardev.c
midnightveil: Now that I look at the util_libs PR, I don't know if any of these headers exist in the kernel? Or are even needed; you never include chardev.c and it seems to be identical to that in libutils.

libsel4/sel4_plat_include/stm32mp2/sel4/plat/api/constants.h
```diff
@@ -0,0 +1,17 @@
...
+#error "AARCH32 is unsupported"
```
Indanz, line 13: BTW, it's possible to add 32-bit support to any 64-bit only platforms if anyone wants it, it's just not done for some platforms because whoever paid for the platform port didn't need 32-bit support and  it saved time and effort not getting 32-bit to work.

In this case it's ST adding support, so it might make more sense to add 32-bit support from the start, but up to you.

src/plat/stm32mp2/chardev.c
Indanz, line 1: This file isn't necessary and should be deleted. libsel4utils needs it, but they have a different infrastructure. The kernel just uses hard-coded UARTs for debugging only.

src/plat/stm32mp2/config.cmake
```diff
@@ -0,0 +1,56 @@
...
+    list(GET c_configs ${index} c_config)
+    list(GET cmake_configs ${index} cmake_config)
+    config_set(KernelMP2Platform ARM_PLAT ${KernelMP2Platform})
+    config_set(${cmake_config} ${c_config} ON)
+    list(APPEND KernelDTSList "tools/dts/${KernelMP2Platform}.dts")
+    list(APPEND KernelDTSList "src/plat/stm32mp2/overlay-stm32mp2.dts")
+
+    set(KernelArmCortexA35 ON)
+    set(KernelArchArmV8a ON)
+    set(KernelArmGicV2 ON)
```
Indanz: It's very strange that this SoC comes with a GICv2 instead of a GICv3, but checking the datasheet, that's indeed the case.

src/plat/stm32mp2/config.cmake
```diff
@@ -0,0 +1,56 @@
...
+    set(KernelArmExportPCNTUser ON)
```
Indanz: This should be done by the user and not by the platform (we made this mistake in the past for the tqma board).

src/plat/stm32mp2/config.cmake
```diff
@@ -0,0 +1,56 @@
...
+
+    message(STATUS "KernelMP2Platform is ${KernelMP2Platform}")
+    config_set(KernelARMPlatform MACH "stm32mp2")
+
+    declare_default_headers(
+        TIMER_FREQUENCY 40000000
+        MAX_IRQ 383
```
Indanz: I think you have to add 32 to this to account for the PPIs.

src/plat/stm32mp2/overlay-stm32mp2.dts
```diff
@@ -0,0 +1,28 @@
...
+		    "serial0",
+		    &{/psci};
+
+		seL4,kernel-devices =
+		    "serial0",
+		    &{/interrupt-controller@4ac00000},
+		    &{/timer};
+	};
+
+        /* vgic maintenance interrupt */
```
Indanz: Indent is off, maybe mixing spaces and tabs?

tools/dts/stm32mp257f-ev1.dts
```diff
@@ -0,0 +1,2875 @@
...
+
+	reserved-memory {
+		#address-cells = <0x02>;
+		#size-cells = <0x02>;
+		ranges;
+
+		fw@80000000 {
+			compatible = "shared-dma-pool";
+			reg = <0x00 0x80000000 0x00 0x4000000>;
+			no-map;
```
Indanz, line 2876: Only add memory as reserved if it really reserved and not a choice made by Linux, like the DMA pool probably.

Users can always add their own reserved regions via the `KernelCustomDTS` option.

Indanz (commented):
Thanks for adding this. Needs some work, but nothing major.

Is there any chance that ST is willing to provide one or two STM32MP2 dev boards to the seL4 Foundation so that they can add it to CI? It's unusual enough that having it in the testing loop would benefit everyone.

bruelc:
> Thanks for adding this. Needs some work, but nothing major.

Thank you for your very precise review and suggestions. 

> Is there any chance that ST is willing to provide one or two STM32MP2 dev boards to the seL4 Foundation so that they can add it to CI? It's unusual enough that having it in the testing loop would benefit everyone.

I will ask, but a positive outcome is not guaranteed.

tools/dts/stm32mp257f-ev1.dts
bruelc: this cames from the dtb from the linux build. I thought it was easier to leave it untouched and just use the overlays ?

src/plat/stm32mp2/config.cmake
bruelc: Ok I was not sure about the PPIs here, thought I have seen a +32 elsewhere in the code. I'll check

libsel4/sel4_plat_include/stm32mp2/sel4/plat/api/constants.h
bruelc: maybe another PR, for now we don't have a use case for a 32bit kernel. Our use case runs an aarch64 virtualized linux

src/plat/stm32mp2/config.cmake
bruelc: why ? I find it strange to not having those SoC definitions hardcoded, as they cannot be modified.

Indanz: I only meant `KernelArmExportPCNTUser`, which is user settable.

(I really dislike that there are settings users can't modify. I don't think there is a good reason for having those, other than working around limitations of a crappy config system.)

src/plat/stm32mp2/overlay-stm32mp2.dts
```diff
@@ -0,0 +1,28 @@
...
+	interrupt-controller@4ac00000 {
+		interrupt-parent = <0x05>;
+		interrupts = <0x01 0x09 0x308>;
+	};
+
+	vm-memory@b0000000 {
+		reg = <0x0 0xb0000000 0x0 0x20000000>;
```
Indanz: I missed this.

First, if you want to reserve memory, you need to use `no-map` to mark it as "device memory" for the seL4 kernel.

Secondly, you want to do this in the camkes vm project which needs this (via `KernelCustomDTS`), not here. This should be suitable for all users, including those that don't use Linux.

Effectively the above will get merged with the memory one in the main DTS file and be a no-op, so if it works as-is then you don't need it anyway.

tools/dts/stm32mp257f-ev1.dts
Indanz: Yes, agreed it should be removed in overlay-stm32mp2.dts, but I don't see it there yet.

src/plat/stm32mp2/config.cmake
```diff
@@ -0,0 +1,56 @@
...
+    set(KernelArmCortexA35 ON)
+    set(KernelArchArmV8a ON)
+    set(KernelArmGicV2 ON)
+    set(KernelArmExportPCNTUser ON)
+
+    message(STATUS "KernelMP2Platform is ${KernelMP2Platform}")
+    config_set(KernelARMPlatform MACH "stm32mp2")
+
+    declare_default_headers(
+        TIMER_FREQUENCY 40000000
```
Indanz: Just double checking, but is this supposed to be 40 MHz?

(If you get this wrong everything will work fine, except timing will be off, which is only noticeable in MCS for periodic tasks and either lower performance because of short kernel tick periods and the context switch overhead, or high latencies because the periods are too long for non-MCS. The MCS test that prints tick, tock in sel4test should print them every second.)

tools/dts/stm32mp257f-ev1.dts
bruelc: Sorry I didn't explain correctly. This node is in the linux original dts. 
https://github.com/torvalds/linux/blob/master/arch/arm64/boot/dts/st/stm32mp257f-ev1.dts
 I prefer to leave it untouched.

src/plat/stm32mp2/config.cmake
bruelc: it should be OK. This value is derived from the HSE crystal in clock tree by TF-A. 
Now, I can add a platform init sanity check to check the cntfrq_el0 against TIMER_FREQUENCY.

edit: There is a check in initGenericTimer(), but it does't panic

tools/dts/stm32mp257f-ev1.dts
Indanz: Why should we apply Linux policy decisions to seL4? Shouldn't that be up to the user?

src/plat/stm32mp2/config.cmake
Indanz: I forgot about that check. All good then.

tools/dts/stm32mp257f-ev1.dts
bruelc: To me, a device tree describes the hardware, not a policy. In this particular case, what matters is reserving the first 64MB (this is the Secure zone).
I thought we should not touch the original DTS and leave the customizations to the overlay if needed; that simplifies maintenance a lot if the original DTS is updated.

Indanz: > To me, a device tree describes the hardware, not a policy.

Linux device trees often have fixed DMA regions for devices that require 32-bit physical addresses or big contiguous allocations, I thought this was one of those. That's why I said to only reserve it if it's really reserved.

> In this particular case, what matters is reserving the first 64MB (this is the Secure zone).

If we can't use this, then yes, it should stay reserved. But it is named "shared-dma-pool", so I didn't expect it to be the secure-only memory.

> I thought we should not touch the original DTS and leave the customizations to the overlay if needed; that simplifies maintenance a lot if the original DTS is updated.

Yes, I never said otherwise. When I said to remove it, I meant via an entry in the overlay, not directly from the input DTS file.

bruelc: > > To me, a device tree describes the hardware, not a policy.
> 
> 
> 
> Linux device trees often have fixed DMA regions for devices that require 32-bit physical addresses or big contiguous allocations, I thought this was one of those. That's why I said to only reserve it if it's really reserved.
> 
> 
> 
> > In this particular case, what matters is reserving the first 64MB (this is the Secure zone).
> 
> 
> 
> If we can't use this, then yes, it should stay reserved. But it is named "shared-dma-pool", so I didn't expect it to be the secure-only memory.
> 
I admit it s not clear why this area is named shared-dma-pool. Not in the scope of seL4 but this is interesting. I will check with the maintainer. 
In between if for clarity it is best to replicate a reserved area in the overlay I ll do

Indanz: Better to fix it upstream than directly here, I wouldn't bother replicating it. It's as much a problem there as here.

bruelc: After discussing with the Linux maintainer for this part, he confirmed that the name "shared-dma-pool" is not necessary here (probably historical). So, it is not an issue for the seL4 inclusion, aside from causing the confusion you raised.
We are fine using "no-map" without an overlay; in case of an upgrade, the DTS will include it without the name.

bruelc:
> Thanks for adding this. Needs some work, but nothing major.
> 
> Is there any chance that ST is willing to provide one or two STM32MP2 dev boards to the seL4 Foundation so that they can add it to CI? It's unusual enough that having it in the testing loop would benefit everyone.

Good news! I have received a positive response. Could you please send me the delivery information and phone numbers by email?

Indanz (approved):

bruelc:
please hold-on any merge. I have a regression while running the tests in RELEASE mode.

bruelc:
> please hold-on any merge. I have a regression while running the tests in RELEASE mode.

pfff. back to "All is well in the universe" with -DRELEASE=TRUE for sel4test and sel4bench

For the record, the regressions was due to a missing update event necessary to commit the prescaler's timer and the use of USART1 not clocked. Changed to USART2 already clocked by TF-A

Indanz:
> For the record, the regressions was due to a missing update event necessary to commit the prescaler's timer and the use of USART1 not clocked. Changed to USART2 already clocked by TF-A

For the record, it's fine for UART drivers to count on early boot code to have enabled the UART clock/power, it's okay to assume that the same UART will be used as during the boot process.

Indanz:
Is everything ready to be merged, or are there any outstanding issues that anyone is aware of? It all looks ready to be merged to me, except perhaps the camkes VM one, will postpone merging that till there is consensus.

bruelc:
Final functional checks are fine. I ve pushed a camkes-vm-images revision without the binaries. (Kept them on another stm32mp2 fork branch)

Indanz:
Can you squash that last commit and force push?

lsf37:
The current CI failures are unrelated, we're in the process of updating the GitHub CI actions and there are some issues to sort out first. 

Edit: not entirely unrelated it turns out -- the style check fails for the cmake file,  because we have adjusted the formatting rules slightly in the meantime. When the other CI failures are sorted out, I can push a fixup commit for this, no need to do anything at this point.

midnightveil:
@lsf37 Was there anything blocking this being merged now that CI is working again?

lsf37:
Unfortunately CI isn't really working again yet (there are some RISCV+MCS+release failures to sort out still). Otherwise this is ready to go.

I'd like to wait until we have a full green hardware CI run before we merge.

Indanz:
I merged everything I'm aware of, except https://github.com/seL4/camkes-vm-images/pull/13 because I don't know what the camkes VM policy is. Thank you for your patience @bruelc, let me know if I missed anything.

bruelc:
Thank you for your support Indan, I'll do a recheck from master.

bruelc:
> I merged everything I'm aware of, except [seL4/camkes-vm-images#13](https://github.com/seL4/camkes-vm-images/pull/13) because I don't know what the camkes VM policy is. Thank you for your patience @bruelc, let me know if I missed anything.

https://github.com/seL4/microkit/pull/446 seems missing. I just rebased it. 

 There are some build errors in the CI:

CMake Error at configs/seL4Config.cmake:175 (message):
  Variable 'KernelPlatform' is not set - is PLATFORM '' correct? Validplatforms are
  'allwinnerA20;am335x;apq8064;ariane;bcm2711;bcm2837;cheshire;hifive-p550;exynos4;exynos5;fvp;hifive;hikey;imx6;imx7;imx8mq-evk;imx8mm-evk;imx8mp-evk;imx93;maaxboard;odroidc2;odroidc4;omap3;pc99;polarfire;qemu-arm-virt;qemu-riscv-virt;quartz64;rk3568;rocketchip;rockpro64;bananapi-f3;spike;star64;tk1;tqma8xqp1gb;tx1;tx2;zynq7000;zynqmp'

but I supposed it is because a missing reference to the latest kernel ?

PR #1593, opened 2026-02-13
libsel4: make thread-local ipc buffer optional

midnightveil:
At the moment, the seL4 microkit does not setup TLS variable support (nor does it intend to). The workaround has been to `#define __thread` (blank) before including `<sel4/sel4.h>` in `<microkit.h>`, which causes issues if `<microkit.h>` is included *after* `<sel4/sel4.h>`, often with obscure linker errors to `__emutls_**` symbols.

Instead we add a libsel4 config option that allows us to build it with __thread copied out. We introduce an LIBSEL4_THREAD_LOCAL macro in a similar way to the existing LIBSEL4_INLINE macro.

### DIFF
--- a/libsel4/CMakeLists.txt
+++ b/libsel4/CMakeLists.txt
@@ -28,7 +28,8 @@ config_choice(
 )
 
 config_option(
-    LibSel4StubsUseIPCBufferOnly LIB_SEL4_STUBS_USE_IPC_BUFFER_ONLY
+    LibSel4StubsUseIPCBufferOnly
+    LIB_SEL4_STUBS_USE_IPC_BUFFER_ONLY
     "use only IPC buffer for syscalls. When generating syscall wrappers, only use the \
     IPC buffer for marshalling and unmarshalling arguments. Without this option set, \
     arguments will be passed in registers where possible for better performance."
@@ -62,8 +63,7 @@ gen_invocation_header(
 
 gen_invocation_header(
     OUTPUT sel4_arch_include/${KernelSel4Arch}/sel4/sel4_arch/invocation.h
-    XML
-        "${CMAKE_CURRENT_SOURCE_DIR}/sel4_arch_include/${KernelSel4Arch}/interfaces/object-api-sel4-arch.xml"
+    XML "${CMAKE_CURRENT_SOURCE_DIR}/sel4_arch_include/${KernelSel4Arch}/interfaces/object-api-sel4-arch.xml"
     LIBSEL4 SEL4ARCH
 )
 
@@ -73,8 +73,7 @@ gen_invocation_header(
     LIBSEL4 ARCH
 )
 
-set(
-    source_header_dirs
+set(source_header_dirs
     "${CMAKE_CURRENT_SOURCE_DIR}/include"
     "${CMAKE_CURRENT_SOURCE_DIR}/arch_include/${KernelArch}"
     "${CMAKE_CURRENT_SOURCE_DIR}/sel4_arch_include/${KernelSel4Arch}"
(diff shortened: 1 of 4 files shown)


Indanz:
Yes, please. However, please go one step further and make it possible to provide a user defined macro which returns the PC buffer pointer and get rid of `__sel4_ipc_buffer` altogether if that macro is defined.

Use case: For pre-build Arm binaries with no TLS support, it would be nice if we can use the TLS register ourselves for the IPC buffer, so that threads can do kernel calls, even if they don't have TLS.

libsel4/src/sel4_bootinfo.c
```diff
@@ -7,11 +7,11 @@
 #include <sel4/sel4.h>
 
 #ifdef CONFIG_KERNEL_INVOCATION_REPORT_ERROR_IPC
-__thread char __sel4_print_error = CONFIG_LIB_SEL4_PRINT_INVOCATION_ERRORS;
+LIBSEL4_THREAD_LOCAL char __sel4_print_error = CONFIG_LIB_SEL4_PRINT_INVOCATION_ERRORS;
```
Indanz, line 10: This one should just disappear, together with `seL4_SetPrintError` and `seL4_CanPrintError`. The latter is used in generated code, but I can't find any callers of `seL4_SetPrintError` anywhere on Github. But that's for another PR maybe.

Indanz (approved):
Let's start with this first.

The only alternative to this would be to force users to pass `-D__thread=` as global compile options, but it should be up to the user whether ipc_buffer is a __thread or not.

libsel4/src/sel4_bootinfo.c
midnightveil: I found this searching it, which is interesting because it relates to your earlier comment: https://github.com/cl91/NeptuneOS/blob/0452b7644aa08a548550346ccc337df07bbae835/build.sh#L139-L157 about replacing the entire feature.

I wonder if it makes sense to make seL4_GetIPCBuffer defined entirely by the user. But this seems like a bigger change.

Indanz: Currently libsel4 makes user policy decisions about the ipc_buffer, so of course having an option to let users define their own `seL4_GetIPCBuffer()` instead makes a lot of sense.

But to avoid the problem you ran into, it either needs to be a global compile option (e.g. `--include` which includes a header file defining seL4_GetIPCBuffer), or the header file to include needs to be the config option and then sel4.h does something like `#include CONFIG_GET_IPC_BUFFER_FILE`.

Indanz: I wish the Neptune people would have opened an issue asking for proper support for what they're doing instead of hacking it together though. :-/

kent-mcleod:
> Yes, please. However, please go one step further and make it possible to provide a user defined macro which returns the PC buffer pointer and get rid of `__sel4_ipc_buffer` altogether if that macro is defined.

Or have a build time option to codegen functions that take the IPC buffer as a function argument so that there's an option for neither global or TLS variable, but just a variable on the thread's stack.

Indanz:
> > Yes, please. However, please go one step further and make it possible to provide a user defined macro which returns the PC buffer pointer and get rid of `__sel4_ipc_buffer` altogether if that macro is defined.
> 
> Or have a build time option to codegen functions that take the IPC buffer as a function argument so that there's an option for neither global or TLS variable, but just a variable on the thread's stack.

Although that makes a lot of sense, seL4 complicates thing by requiring you to register the IPC buffer beforehand.
What I personally did was put the IPC buffer at the bottom of the stack, so it's not totally unused. People wanting it on the stack can do the same, with the advantage that the location is known and can't cross a page boundary by accident.

But really, people should configure the IPC buffer less and use the `*withMR` syscalls more instead. It's only needed for some syscalls, not for IPC.

kent-mcleod:
> Although that makes a lot of sense, seL4 complicates thing by requiring you to register the IPC buffer beforehand.

I just meant storing the pointer to the IPC buffer on the stack instead of a global/thread global.

But yea, top of the stack is a common place to allocate the IPC buffer itself too.

midnightveil:
I believe most of the comments are about making this more generic, see #1598. Whilst I do agree with most of them, they are a bit more work.

Aside from that, I don't think there's any other comments about this that were raised.

libsel4/CMakeLists.txt
```diff
@@ -43,6 +43,14 @@ config_string(
     UNQUOTE
 )
 
+config_option(
+    LibSel4UseThreadLocals LIB_SEL4_USE_THREAD_LOCALS
+    "mark the __sel4_ipc_buffer variable as thread local using '__thread'. \
+     This requires thread local support to exist and is useful if multiple \
```
Indanz:
```suggestion
     This requires thread local storage support and is useful if multiple \
```

Indanz (approved):

libsel4/CMakeLists.txt
midnightveil: Done.

Indanz:
Can't merge because of the failing style check. Of course unrelated to your changes... :-(

midnightveil:
> Can't merge because of the failing style check. Of course unrelated to your changes... :-(

Fixed up now.

Issue #1591, opened 2026-02-11
x86 `KernelIOMMU` set even when `KernelVerificationBuild` is set to true


Ivan-Velickovic:
Microkit has `KernelVerificationBuild` set to true when building in release mode. However, in `gen_config.h` I still see `CONFIG_IOMMU` enabled. I don't exactly know what's going on since the configuration option has `DEFAULT_DISABLED OFF`, so I would have assumed that if the depends condition `DEPENDS "KernelPlatPC99; NOT KernelVerificationBuild"` evaluates to false then the IOMMU option would also be OFF.

Ivan-Velickovic:
When I enter the build directory and do `ccmake .` I do see `KernelVerificationBuild: ON` although I do not see `KernelPlatPC99` although I'm not sure if it is supposed to be there. I'm not an expert on CMake so I'm not quite sure how to debug this.

Indanz:
For projects using easy-settings, not setting `-DVERIFICATION=ON` will overrule any manually set `KernelVerificationBuild`, just like it does for release and debug and printing etc.

It's very annoying and unexpected, they should be just an easy shorthand when set, but do nothing when those easy settings aren't used.

So double check that `KernelVerificationBuild` is actually set to ON in the generated CMakeCache.txt, I suspect it isn't.

Ivan-Velickovic:
In `CMakeCache.txt` I see both `KernelVerificationBuild:BOOL=ON` and `KernelPlatPC99:INTERNAL=ON`.

Ivan-Velickovic:
I see this in my `CMakeCache.txt` though:
```
KernelIOMMU:INTERNAL=OFF
KernelIOMMU_DISABLED:INTERNAL=TRUE
```

Indanz:
Do you have a CMakeCache.txt with inconsistent results?

If not, do you have a CMakeCache.txt which is inconsistent with the generated CONFIG_IOMMU defines in the same build directory?

midnightveil:
This is basically the same issue as #1565, not sure why new issue.

Anyway, the issue is the ordering of the `config_option` statements and their interaction with default values — because the "DEFAULT" is evaluated at the time of the `config_option` declaration, if we have no value for `KernelVerificationBuild` (e.g. we are evaluating `KernelIOMMU` in the arch configs which happens before all the global things).

Note that CMake is still an imperative language, ordering matters; it's not lazily evaluated.

Copy the same steps from #1565:

```
[...]

Issue #1565, opened 2025-11-26
KernelAllowSMCCalls can be turned on in verification builds


Ivan-Velickovic:
```
git clone https://github.com/seL4/seL4
mkdir build
cd build
cmake -GNinja -DKernelAllowSMCCalls=1 -DKernelPlatform=qemu-arm-virt -DKernelSel4Arch=aarch64 -DQEMU_MEMORY=2048 -S ../seL4 -B . -DCROSS_COMPILER_PREFIX=aarch64-none-elf-
```

If you look at `gen_config/kernel/gen_config.h` you'll see:
```
#define CONFIG_ALLOW_SMC_CALLS  1
```

However, the CMake configuration has this:
https://github.com/seL4/seL4/blob/c0749e622753a8819e3b3a6a23a93f1b039e4f32/src/arch/arm/config.cmake#L218

which I thought should mean that since `CONFIG_VERIFICATION_BUILD` is on by default (looking at `gen_config.h` again) that it should not be possible for me to have this option turned on.

This will soon become irrelevant anyways since @lsf37 said that this config option will be allowed for verification builds but it would be nice to know why this doesn't work as intended in case it affects other configuration options.

kent-mcleod:
It looks like `DEFAULT_DISABLED OFF` is missing from the option declaration which should lead it to be set to the OFF value when verification build is enabled.

midnightveil:
That doesn't make sense because of `config_option`:

```
    if("${CONFIG_DEFAULT_DISABLED}" STREQUAL "")
        set(CONFIG_DEFAULT_DISABLED "${CONFIG_DEFAULT}")
    endif()
```

PR #1594, opened 2026-02-16
config: Fix evaluation ordering for configs

Kswin01:
Currently, there are two configs, `KernelVerificationBuild` and `KernelMaxNumNodes` that are used in arch specific config cmake files, but are not evaluated in the top level config.cmake before we include the arch specific ones. Also add error checks to make sure that these dependencies are defined when entering arch specific configs.

We ran into this issue when building x86 Microkit systems. By default, Microkit didn't specify KernelVerificationBuild being true in release mode. However, due to evaluating the arch specific configs before setting the default for KernelVerificationBuild, we inadvertently enable the IOMMU:
```
config_option(
    KernelIOMMU IOMMU "IOMMU support for VT-d enabled chipset"
    DEFAULT ON
    DEPENDS "KernelPlatPC99; NOT KernelVerificationBuild"
    DEFAULT_DISABLED OFF
)
```

`config_option` evaluates unset variables to false, which results in this unexpected behavior.

I'm not sure if I've missed any other configs that suffer the same issues, but please let me know.

Fixes https://github.com/seL4/seL4/issues/1565. Using the steps to replicate issues, we now correctly get:
```
/* disabled: CONFIG_ALLOW_SMC_CALLS */
```

Also closes: https://github.com/seL4/seL4/issues/1591.

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -398,24 +398,13 @@ config_option(
     DEPENDS "NOT KernelBinaryVerificationBuild"
 )
 
-if(NOT KernelArchARM)
-    # This depends on KernelArmHypervisorSupport, the default will only
-    # be set if we are building for ARM configs. If building for other architectures
-    # this is problematic. We can assume that it is false at this point
-    set(KernelArmHypervisorSupport OFF CACHE INTERNAL "" FORCE)
-endif()
-
-if(NOT KernelArchX86)
-    # Similar to above, KernelSkimWindow is only defined in x86 configs
-    set(KernelSkimWindow OFF CACHE INTERNAL "" FORCE)
-endif()
 config_option(
     KernelDangerousCodeInjection DANGEROUS_CODE_INJECTION
     "Adds a system call that allows users to specify code to be run in kernel mode. \
     Useful for profiling."
     DEFAULT OFF
     DEPENDS
-        "NOT KernelArmHypervisorSupport;NOT KernelVerificationBuild;NOT KernelPlatformHikey;NOT KernelSkimWindow"
+        "KernelDangerousCodeInjectionSupported"
 )
 
 config_option(
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -265,6 +265,16 @@ else()
     config_set(KernelArmHasTlbLock ARM_HAS_TLB_LOCK OFF)
 endif()
 
+if(NOT KernelArmHypervisorSupport AND NOT KernelVerificationBuild AND NOT KernelPlatformHikey)
+    set(KernelDangerousCodeInjectionSupported
+        ON
+        CACHE INTERNAL "")
+else()
+    set(KernelDangerousCodeInjectionSupported
+        OFF
+        CACHE INTERNAL "")
+endif()
+
 add_sources(
     DEP "KernelArchARM"
     PREFIX src/arch/arm
--- a/src/arch/x86/config.cmake
+++ b/src/arch/x86/config.cmake
@@ -387,6 +387,16 @@ endif()
 if(KernelSel4ArchX86_64 AND NOT KernelFSGSBaseInst)
     set(KernelSetTLSBaseSelf ON)
 endif()
+if(NOT KernelSkimWindow AND NOT KernelVerificationBuild)
+    set(KernelDangerousCodeInjectionSupported
+        ON
+        CACHE INTERNAL "")
+else()
+    set(KernelDangerousCodeInjectionSupported
+        OFF
+        CACHE INTERNAL "")
+endif()
+
 
 add_sources(
     DEP "KernelArchX86"
--- a/src/arch/riscv/config.cmake
+++ b/src/arch/riscv/config.cmake
@@ -124,6 +124,16 @@ set(KernelHardwareDebugAPIUnsupported
     CACHE INTERNAL ""
 )
 
+if(NOT KernelVerificationBuild)
+    set(KernelDangerousCodeInjectionSupported
+        ON
+        CACHE INTERNAL "")
+else()
+    set(KernelDangerousCodeInjectionSupported
+        OFF
+        CACHE INTERNAL "")
+endif()
+
 add_sources(
     DEP "KernelArchRiscV"
     PREFIX src/arch/riscv


src/arch/arm/config.cmake
```diff
@@ -10,6 +10,10 @@ if(KernelSel4ArchAarch32)
     set_property(TARGET kernel_config_target APPEND PROPERTY TOPLEVELTYPES pde_C)
 endif()
 
+if (NOT KernelVerificationBuild)
+    message(FATAL_ERROR "KernelVerificationBuild used but not defined!")
+endif()
```
Indanz: How does this work? Below it depends on `NOT KernelVerificationBuild`.

src/arch/x86/config.cmake
```diff
@@ -26,6 +26,14 @@ if(KernelSel4ArchX86_64)
     )
 endif()
 
+if (NOT KernelMaxNumNodes)
+    message(FATAL_ERROR "KernelMaxNumNodes used but not defined!")
+endif()
```
Indanz: This works because it is always defined to 1 or more.

src/arch/x86/config.cmake
```diff
@@ -26,6 +26,14 @@ if(KernelSel4ArchX86_64)
...
+
+if (NOT KernelVerificationBuild)
+    message(FATAL_ERROR "KernelVerificationBuild used but not defined!")
+endif()
```
Indanz: But this isn't. You have not explicitly check whether it is is undefined instead of doing a logical check.

```suggestion
if (NOT DEFINED KernelVerificationBuild)
    message(FATAL_ERROR "KernelVerificationBuild used but not defined!")
endif()
```

Bette to check for NOT DEFINED everywhere as that makes the intention clearer.

src/arch/x86/config.cmake
Kswin01: This default for KernelMaxNumNodes was set after we evaluated the arch specific configs. Is it also defined elsewhere?

src/arch/x86/config.cmake
midnightveil: Was there a particular reason you did not modify config_option to check if the values were defined when evaluating it's dependencies?

Kswin01: I didn't want to change config_option as I wasn't entirely sure if I missed any other options that relied on this behavior. I can change config_option of you want, but I'm guessing there might want to be a bit of discussion around that

Indanz: Why? It seems a great idea, it should find most users of undefined configs with no downsides. It doesn't find explicit if-checks, but it's better than nothing. It should have found this particular issue at least.

src/arch/x86/config.cmake
Indanz: I meant, just checking it with NOT instead of NOT DEFINED works by chance.

src/arch/x86/config.cmake
```diff
@@ -26,6 +26,14 @@ if(KernelSel4ArchX86_64)
     )
 endif()
 
+if(NOT DEFINED KernelMaxNumNodes)
+    message(FATAL_ERROR "KernelMaxNumNodes used but not defined!")
+endif()
+
+if(NOT DEFINED KernelVerificationBuild)
+    message(FATAL_ERROR "KernelVerificationBuild used but not defined!")
+endif()
```
Indanz: These checks are superfluous if you add them to config_option/config_choice/etc.

Kswin01:
I've now added the check for if a dependency has been defined to config options. This lead to additional issues which are described in the commit message, as there were several cases where we were depending on the previous behavior of config options

Kswin01:
When adding an error case to the config option function, we get alot of errors where we are relying on the depend expression to evaluate to false. An example of this is in `config.cmake`:
```
config_option(
    KernelDangerousCodeInjection DANGEROUS_CODE_INJECTION
    "Adds a system call that allows users to specify code to be run in kernel mode. \
    Useful for profiling."
    DEFAULT OFF
    DEPENDS
        "NOT KernelArmHypervisorSupport;NOT KernelVerificationBuild;NOT KernelPlatformHikey;NOT KernelSkimWindow"
)
```

This relies on `KernelArmHypervisorSupport`, which will only be set to a default of false if building for ARM architectures, and `KernelSkimWindow` which is the same except for x86.

There's alot of other small issues as well that have cropped up from this, I'm going to print a warning from `config_options` instead of an error and try and work from there

tools/helpers.cmake
```diff
@@ -282,6 +282,19 @@ function(config_option optionname configname doc)
         # Check the passed in dependencies. This loop and logic is inspired by the
         # actual cmake_dependent_option code
         foreach(test ${CONFIG_DEPENDS})
+            # This is assuming that the only keywords that are in a depend string
+            # are or equal and not. Equal and not will be stripped away, and or
+            # replaced with ; so we can iterate through the list correctly.
+            string(REGEX REPLACE "OR" ";" depend_vars "${test}")
+            string(REGEX REPLACE "EQUAL" "" depend_vars "${depend_vars}")
+            string(REGEX REPLACE "NOT" "" depend_vars "${depend_vars}")
```
Indanz:
```suggestion
            string(REGEX REPLACE " OR " ";" depend_vars "${test}")
            string(REGEX REPLACE " EQUAL [0-9A-Za-z_]*" "" depend_vars "${depend_vars}")
            string(REGEX REPLACE "NOT " "" depend_vars "${depend_vars}")
```
Otherwise you end up with a lonely " 1" for the only user left of EQUAL: `DEPENDS "${KernelNumDomains} EQUAL 1"`, and without the spaces, anything with OR or NOT as substring matches by accident.

tools/helpers.cmake
```diff
@@ -526,8 +539,13 @@ function(config_choice optionname configname doc)
...
         else()
-            # Remove this config as it's not valid
+            # Ensure that we set this option to false, and remove the previous
+            # entry from the cache
             unset(${option_cache} CACHE)
+            set(${option_cache}
+                OFF
+                CACHE
+                INTERNAL "" FORCE)
             cfg_str_add_disabled(local_config_string ${option_config})
```
Indanz, line 548: This is not obvious, and CMake variable handling is tricky enough already, so I'd like a good explanation of why this is as it is. Why unset and set the same option?

If the intention is to get the variable defined so the config_option checks doesn't trigger, that's fine, but can't we do that locally instead of polluting the cache?

Kswin01: Yes the intention behind the set is to ensure that was due to some dependencies on `KernelArch...` and `KernelPlat...`, which were being unset here. This caused configs such as this to fail: 
```
config_option(
    KernelDebugDisablePrefetchers
    DEBUG_DISABLE_PREFETCHERS
    "On ia32 platforms, this option disables the L2 hardware prefetcher, the L2 adjacent \
    cache line prefetcher, the DCU prefetcher and the DCU IP prefetcher. On the cortex \
    a53 this disables the L1 Data prefetcher."
    DEFAULT OFF
    DEPENDS "KernelArchX86 OR KernelPlatformHikey"
)
```

I tried setting it locally first, but that didn't change the behavior, whereas adding it to the cache did. There's no need for unset now

config.cmake
```diff
@@ -398,6 +398,17 @@ config_option(
...
+    # This depends on KernelArmHypervisorSupport, the default will only
+    # be set if we are building for ARM configs. If building for other architectures
+    # this is problematic. We can assume that it is false at this point
+    set(KernelArmHypervisorSupport OFF CACHE INTERNAL "" FORCE)
+endif()
+
+if(NOT KernelArchX86)
+    # Similar to above, KernelSkimWindow is only defined in x86 configs
+    set(KernelSkimWindow OFF CACHE INTERNAL "" FORCE)
+endif()
```
kent-mcleod: Rather than defining arch specific configs in the common config.cmake file, can you define a Supported flag like what we do with the hardware debug config: `KernelHardwareDebugAPIUnsupported` (but in the positive rather than negative). The architecture configs can then set the KernelDangerousCodeInjectionSupported and then the option can be enabled if the architecture config supports it.

PR #1595, opened 2026-02-17
aarch64,smc: reduce #ifdef use

lsf37:
Since SMC caps are now part of the verified code base, there is no longer any need to hide the code when CONFIG_ALLOW_SMC_CALLS is not configured.

Remove the corresponding #ifdefs where possible, replace them by  #ifdef CONFIG_ARCH_AARCH64 where they were in mode-generic arm files. Move smc.h/.c to mode-specific directories.

The only remaining use of #ifdef CONFIG_ALLOW_SMC_CALLS is in boot code, either providing or not providing the initial SMC cap. With this change, verification now applies to both configurations.

We could go one step further and remove the config option to have it always on, but that would be an interface change, so I've left that for now.


lsf37:
Looks like I need to bump the Isabelle version for MCS proofs, but RISC-V was not affected anyway so we can ignore that result. Will push a style fixup when the proofs check is finished.

Indanz (approved):
I'm strongly in favour of removing the config option now. I think the next release will be breaking in one way or the other anyway, however, removing the config option doesn't break anything for anyone, it's a backward compatible change. It's still good to mention the change clearly in case people don't want any SMC cap, in which case they now have to delete it. But long-term we don't want unnecessary config options, and the sooner we remove it, the less people are affected.

lsf37:
> I'm strongly in favour of removing the config option now. I think the next release will be breaking in one way or the other anyway, however, removing the config option doesn't break anything for anyone, it's a backward compatible change. It's still good to mention the change clearly in case people don't want any SMC cap, in which case they now have to delete it. But long-term we don't want unnecessary config options, and the sooner we remove it, the less people are affected.

I would not be opposed. (Removing the ALLOW_SMC_CALLS option completely and always leave it on).  @seL4/tsc any opinions on that before I do this?

midnightveil:
> however, removing the config option doesn't break anything for anyone

It's probably worth leaving in a `config_set` so that people who did maybe an `#ifdef SMC` aren't broken by this change?

lsf37:
> > however, removing the config option doesn't break anything for anyone
> 
> It's probably worth leaving in a `config_set` so that people who did maybe an `#ifdef SMC` aren't broken by this change?

Good point. I don't see a good way to do that, but it would be nice if this could somehow result in a deprecation warning so that we can more smoothly fully remove it in a few releases.

midnightveil:
We have a way to deprecate macros: `SEL4_DEPRECATE_MACRO`. We could probably do like

```c
#ifdef CONFIG_AARCH64
#define CONFIG_ARM_SMC SEL4_DEPRECATE_MACRO(1)
#endif
```

in `deprecated.h` or some such.

lsf37:
> We have a way to deprecate macros: SEL4_DEPRECATE_MACRO.

Cool, if that fires on `#ifdef ALLOW_SMC_CALLS` that would work.

Now that I think about it, this may be annoying. If I have user-level code that I want to be able to use with multiple seL4 versions (maybe after suitable wrappers), you might need the `#ifdef` for older ones, but you still get the warning all the time. Maybe that's too obscure a use case and not worth worrying about, though.

Indanz:
> Now that I think about it, this may be annoying. If I have user-level code that I want to be able to use with multiple seL4 versions (maybe after suitable wrappers), you might need the `#ifdef` for older ones, but you still get the warning all the time. Maybe that's too obscure a use case and not worth worrying about, though.

But then you want a way to suppress those deprecated warnings somehow, because what you said is a valid point which applies to all deprecated features. And I still think people would prefer an explicit deprecation warning than having to debug a broken system if a config option silently disappears. So perhaps:

```C
#if defined(CONFIG_AARCH64) && !defined(CONFIG_ARM_SMC)
#define CONFIG_ARM_SMC SEL4_DEPRECATE_MACRO(1)
#endif
```
Then people can do a version check and define CONFIG_ARM_SMC themselves to avoid this warning.
Hm, except the seL4 version isn't exported?

lsf37:
Ok, I've added a second commit now that does the always-on and deprecation thing.

It occurs to me that this might still be breaking if someone reads the `gen_config.json` file, though, because they would conclude that the option is not set and the cap does not exist where in fact it does exist.

Happy to go back on that if we don't want it after all.

(The previous state, before the second commit is definitely non-breaking)

midnightveil:
HRM, this would break the microkit:

https://github.com/seL4/microkit/blob/cce4a8d8c6cd66d49a0d4f0630249d0556a7f58e/tool/microkit/src/main.rs#L496

Probably would break rust-seL4 too. 

Might be not worth it. I guess a `config_set` and eschewing the configuration *option* would retain that. 

But maybe it's not worth it at that point.

lsf37:
It seems it already breaks sel4test if you look at the simulation results in the PR checks. I think I'd also be happiest with not doing this for now.

We could do the non-breaking change now and add a deprecation warning for turning off `KernelAllowSMCCalls` in cmake for two releases or so, then people have time to prepare and we can make a small breaking change down the line.

kent-mcleod:
> I would not be opposed. (Removing the ALLOW_SMC_CALLS option completely and always leave it on). @seL4/tsc any opinions on that before I do this?

I'm against removing the config option. The static compilation options are part of the "Don’t pay for what you don’t use" design principle. If a system isn't going to be performing SMC calls, yes it could delete the cap at runtime, but it's still carrying around the program code for implementing the SMC cap behavior.

lsf37:
> > I would not be opposed. (Removing the ALLOW_SMC_CALLS option completely and always leave it on). @seL4/tsc any opinions on that before I do this?
> 
> I'm against removing the config option. The static compilation options are part of the "Don’t pay for what you don’t use" design principle. If a system isn't going to be performing SMC calls, yes it could delete the cap at runtime, but it's still carrying around the program code for implementing the SMC cap behavior.

I think we're converging on not removing the option for now, but should point out that we do need to pay the price of having the code in any case if we want both options to have a proof. These are just a few instructions though, and no runtime data and no impact on any performance critical paths.

The only difference is whether the initial cap will show up in the initial CSpace or not.

kent-mcleod:
> I think we're converging on not removing the option for now, but should point out that we do need to pay the price of having the code in any case if we want both options to have a proof.

If the code isn't used, but we need to include it anyway then this seems like it could be a problem with the way things are structured somewhere.

lsf37:
> > I think we're converging on not removing the option for now, but should point out that we do need to pay the price of having the code in any case if we want both options to have a proof.
> 
> If the code isn't used, but we need to include it anyway then this seems like it could be a problem with the way things are structured somewhere.

I don't think it's a code structure problem. It's too big a feature to support generically in the proof as a config, it affects too many invariants and intermediate results. We need to decide to either have it or not have it. There is no runtime performance cost or security difference, so I don't think we are really paying anything. The only difference is marginally increased code size.

In that sense it's like SGI -- that's also always on. The only reason SMC ever was an option in the first place was because the proofs could not support it when it was introduced.

kent-mcleod:
> We could do the non-breaking change now and add a deprecation warning for turning off `KernelAllowSMCCalls` in cmake for two releases or so, then people have time to prepare and we can make a small breaking change down the line.

If the CMake config variable or header file define were being used as feature flags, then isn't the change you're making removing the case where the flag is disabled? So moving forward the flag can only be set to true and doesn't need to be deprecated.

kent-mcleod (approved):

Indanz:
> > We could do the non-breaking change now and add a deprecation warning for turning off `KernelAllowSMCCalls` in cmake for two releases or so, then people have time to prepare and we can make a small breaking change down the line.
> 
> If the CMake config variable or header file define were being used as feature flags, then isn't the change you're making removing the case where the flag is disabled? So moving forward the flag can only be set to true and doesn't need to be deprecated.

@kent-mcleod, are you happy with the current:

```C
#if defined(CONFIG_ARCH_AARCH64) && !defined(CONFIG_ALLOW_SMC_CALLS)
#define CONFIG_ALLOW_SMC_CALLS SEL4_DEPRECATE_MACRO(1)
#endif
```
Or would you prefer to first force set the CMake config option to ON and do the above in the future?

lsf37:
I have rebased, resolved conflicts, and dropped the second commit again that made this a breaking change.

We're back to the version that is no behaviour change at all, it just removes `#ifdef`. It has a few more lines of code in the kernel for non-SMC configs than before that are unreachable because the cap is not provided, but that is minimal and comparable to other caps that are e.g. not provided by the initialiser.

Indanz (approved):

midnightveil (approved):

lsf37:
The ARM proof failed because the l4v manifest hadn't updated to the last commit yet, so I'll merge.

PR #1604, opened 2026-03-06
Support for Kria K26 SOM as zynqmp overlay

SeedRizvi:
Register `kria_k26` as an overlay on ZCU102 in the build system. Made with support from @potanin .

*This is part of a set of related PRs across microkit, seL4, sDDF, and seL4-docs to add Kria K26 support.*

- `src/plat/zynqmp/config.cmake`: add `KernelPlatformZynqmpKriaK26` to `cmake_configs` and `kria_k26` to `plat_lists`, creating `CONFIG_PLAT_ZYNQMP_KRIA_K26`
- Overlay DTS `overlay-zynqmp-kria-k26.dts` onto `zynqmp.dts`:
  - UART: Serial1 instead of Serial0 
  - Different `memory@0` region for K26.

### DIFF
--- a/src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
+++ b/src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2026, Skykraft
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
@@ -32,8 +32,6 @@
 	};
 
     chosen {
-		bootargs = "earlycon";
-		stdout-path = "serial1:115200n8";
 		seL4,elfloader-devices =
 		    "serial1",
 		    &{/psci},


src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
```
midnightveil: This copyright doesn't seem right.

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+/ {
+	amba {
+		/* These blocks remove the duplicate interrupt which causes issues
```
midnightveil, line 9: Is the ethernet etc the same as the zcu102 or is it different that this needs to be copied?

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
```diff
@@ -0,0 +1,47 @@
...
+	};
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00 0x00 0x00 0x7ff00000 0x08 0x00 0x00 0x80000000>;
+	};
+
+    chosen {
+		bootargs = "earlycon";
+		stdout-path = "serial1:115200n8";
```
midnightveil: These are linux things, no?

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
```diff
@@ -0,0 +1,47 @@
...
+
+		ethernet@ff0e0000 {
+			interrupts = < 0x00 0x3f 0x04 >;
+		};
+	};
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00 0x00 0x00 0x7ff00000 0x08 0x00 0x00 0x80000000>;
+	};
```
midnightveil, line 32: I guess this is the important thing here, as well as the serial?

This is what is it in the ZynqMP variant:

```
	memory@0 {
		device_type = "memory";
		reg = < 0x00 0x00 0x00 0x80000000 0x08 0x00 0x00 0x80000000 >;
	};
```

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
SeedRizvi: It's the same as ZCU102, but I misspoke in the original PR. Since this is an overlay on the `zynqmp.dts`, I figured this block would still be needed?  (`seL4/src/plat/zynqmp/config.cmake` applies this `overlay-zynqmp-kria` now instead of `overlay-zynqmp`)

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
SeedRizvi: Precisely

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
SeedRizvi: You're right, forgot to remove those. Just waiting on something internal and I'll push the changes.

src/plat/zynqmp/overlay-zynqmp-kria-k26.dts
SeedRizvi: Just waiting on something internal and then I'll push the fixed changes.

Indanz:
The reason I added the `KernelCustomDTS` config option is so that users can easily tweak memory and UART settings for their use case, _without_ the need to add another platform seL4.

So considering none of this is needed, I don't think we should spend time on adding every module or board based on an already supported SoC. If the particular board is widespread enough and the existing platform already supports multiple flavours, adding another one is fine. I don't think that's the case here.

Of course, if UART support is missing for the alternative UART, it's fine to add support for that to user space libraries.

midnightveil:
> The reason I added the KernelCustomDTS config option is so that users can easily tweak memory and UART settings for their use case, without the need to add another platform seL4

I think the one case where this is not true is KernelArchArmPlatform which can't be overridden easily externally? Which is probably not a kernel concern but a lot of userspace programs (e.g. the microkit) uses it to distinguish between boards.

---

However yes we have done this before for seL4 microkit.

Indanz:
> I think the one case where this is not true is KernelArchArmPlatform which can't be overridden easily externally? Which is probably not a kernel concern but a lot of userspace programs (e.g. the microkit) uses it to distinguish between boards.

IMHO user space programs should avoid using kernel configuration options as much as possible, but have their own instead. `KernelARMPlatform` isn't used by the kernel itself as far as I can tell and shouldn't be there at all I think.

The user tells everyone what they want and the build system should turn that into the correct configuration for the software so it does the right thing. What that is for the seL4 kernel is independent from the way to tell Microkit that.

If Microkit wants to know the detailed kernel configuration, then whatever info the kernel uses out of DTS should be exported somehow and that should be used as input for Microkit. Not use `KernelARMPlatform` and then make wrong assumptions based on that.

SeedRizvi:
Just read this thread. Should I just close this PR and instead just keep the microkit/sddf PRs, with updated documentation for the requirement of an overlay in e.g. microkit/docs/manual.md? @Indanz

Indanz:
> Should I just close this PR and instead just keep the microkit/sddf PRs, with updated documentation for the requirement of an overlay in e.g. microkit/docs/manual.md? @Indanz

That would be my preference, yes. But that's just my opinion though.

You don't have to, you can also give counter arguments why I am wrong and merging this would be better.

kent-mcleod:
I'm also in the same boat as Indan. The build interfaces should enable the platform to be used without kernel changes, but there are also lots of other zynqmp boards in the kernel config and the kria would feel left out.

SeedRizvi:
The alternative I can suggest is to only keep the overlay in this PR, and revert all the other changes. That way Kria is still supported out of the box for Microkit? This would be viable now due to [this microkit commit](https://github.com/seL4/microkit/commit/27dc7914b24539214f9672d4714604f18067e079).

Indanz:
> The alternative I can suggest is to only keep the overlay in this PR, and revert all the other changes. That way Kria is still supported out of the box for Microkit?

That doesn't make any sense though. So either nothing or all.

Microkit can have the overlay if it needs it for a board, no reason it should be here.

> This would be viable now due to [this microkit commit](https://github.com/seL4/microkit/commit/27dc7914b24539214f9672d4714604f18067e079).

If Microkit had support for overlays itself, and would pass that to the kernel, then none of this would be needed.

Because just like it doesn't make sense to add explicit support for all the boards and flavours of boards out there to seL4, while just SoC support is sufficient, the same is true for Microkit.

Being user friendly is fine, but in this case you achieve that by making it easy for users to choose UART and the memory size, without preconfiguring all permutations out there.

Ivan-Velickovic:
> If Microkit had support for overlays itself, and would pass that to the kernel, then none of this would be needed.

We do that already for another platform, we can do the same thing for the Kria then.

Ivan-Velickovic:
> Being user friendly is fine, but in this case you achieve that by making it easy for users to choose UART and the memory size, without preconfiguring all permutations out there.

Except you can't really do that since seL4 hard-codes all of this information, so if you want a pre-built kernel you have to have all permutations. We could package multiple kernels into a single image and choose the right one at runtime based on the runtime DTB but that still isn't ideal.

Indanz:
> > Being user friendly is fine, but in this case you achieve that by making it easy for users to choose UART and the memory size, without preconfiguring all permutations out there.
> 
> Except you can't really do that since seL4 hard-codes all of this information, so if you want a pre-built kernel you have to have all permutations. We could package multiple kernels into a single image and choose the right one at runtime based on the runtime DTB but that still isn't ideal.

I'm talking about compile-time, you are talking runtime. Yes, we have a fair way to go to be truly runtime flexible. But first step is stamping out as much board-specific code as possible and limiting it to SoCs.

My long-term plan is to split seL4 boot into a self-booting part that replaces Elfloader and does basic DTB parsing to figure out UARTs, memory etc. to make it runtime dynamic, and a secondary boot stage that is fully platform independent and does the initial task setup etc.

However, Microkit doesn't need this: Only debug builds use UARTs, Microkit can support any UART it wants in user space. Same for memory: As long as the board has at last the compiled in amount of memory, you can still use any additional memory in Microkit as device memory. No need to recompile the kernel for each permutation.

For UARTs, we could add a debug print facility that logs messages to memory which can be printed out to a real UART (or send over the network) by user space. That doesn't help with bootup problems, but to debug those you probably need to recompile anyway.

Ivan-Velickovic:
> But first step is stamping out as much board-specific code as possible and limiting it to SoCs.

Yep, that's fair enough. I agree.

> However, Microkit doesn't need this: Only debug builds use UARTs, Microkit can support any UART it wants in user space. Same for memory: 

Yes that's right, only debug builds need to know the specific UART. The Microkit loader sets this at build time but so does the kernel. At the moment I don't think we could do anything about this part unless we were fine with not having kernel prints on debug builds?

So my understanding is if we have two platforms with the exact same setup except for default UART, we would still need to re-compile the kernel (at least for debug configurations). That seems preferable over not having kernel debug prints.

> As long as the board has at last the compiled in amount of memory, you can still use any additional memory in Microkit as device memory. No need to recompile the kernel for each permutation.

That's true and users can do this already, but it requires the user to know that they are working with more 'normal memory' than the kernel is aware about. If they were to create a memory region from device memory it wouldn't be zeroed out right? So there might be some assumptions in user-code that are broken by doing this. This memory would not be able to be used for kernel objects, but in practice this is probably not a likely issue for people to run into.

The other issue I can think of is that if you have say a kernel configured for 2GB memory and it's the same platform but it has 4GB of memory, if there are holes in that extra 2GB range, the user needs to be aware of it. This is not the case when the kernel is configured for the right amount of memory since it is aware of the holes. The Raspberry Pi 4B is one concrete example of this case.

I agree that it is undesirable to have to recompile the kernel for each permutation, but at the moment as long as the kernel still only sees part of all main memory as being usable as main memory, it doesn't seem much more convenient for users.

Let me know if I'm missing anything/misunderstood.

Indanz:
> So my understanding is if we have two platforms with the exact same setup except for default UART, we would still need to re-compile the kernel (at least for debug configurations). That seems preferable over not having kernel debug prints.

Currently, yes.

> That's true and users can do this already, but it requires the user to know that they are working with more 'normal memory' than the kernel is aware about. If they were to create a memory region from device memory it wouldn't be zeroed out right?

Correct.

> So there might be some assumptions in user-code that are broken by doing this. This memory would not be able to be used for kernel objects, but in practice this is probably not a likely issue for people to run into.

Well, for Microkit people usually don't do any of this, this is all done by the loader and the loader can easily zero memory before mapping it for compatibility and make sure it allocates all kernel objects from the correct UT.

The people doing stuff manually should know what they're doing, you just have to document this somewhere. The section about passing UT to PD's would make sense, or in the board's documentation or something.

> The other issue I can think of is that if you have say a kernel configured for 2GB memory and it's the same platform but it has 4GB of memory, if there are holes in that extra 2GB range, the user needs to be aware of it. This is not the case when the kernel is configured for the right amount of memory since it is aware of the holes. The Raspberry Pi 4B is one concrete example of this case.

How is this different than knowing which device memory to use, or knowing which memory is actually SDRAM?
If you have a crappy system with weird, random holes, then you have to deal with it, or use a less crappy system.

PR #1605, opened 2026-03-09
libsel4: include vmenter.h if VTX is enabled

dreamliner787-9:
`seL4_VMEnter()`, declared in `<sel4/syscalls.h>`, takes three arguments in message registers. The indices for those arguments are defined in `<sel4/arch/vmenter.h>`, but that header is not included transitively.

Include `<sel4/arch/vmenter.h>` from `<sel4/syscalls.h>` when `CONFIG_VTX` is enabled so userspace code including `<sel4/sel4.h>` does not need an extra include.

### DIFF
--- a/libsel4/include/sel4/syscalls.h
+++ b/libsel4/include/sel4/syscalls.h
@@ -21,6 +21,10 @@
 #include "syscalls_master.h"
 #endif
 
+#ifdef CONFIG_VTX
+#include <sel4/arch/vmenter.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


midnightveil:
Is there a difference between this and https://github.com/seL4/seL4/pull/1534?

dreamliner787-9:
> Is there a difference between this and #1534?

Ahh whoops I forgot that already existed. Thanks for pointing out will close this one.

Issue #275, opened 2020-11-01
remove "arm_hyp" as aarch32 kernel architecture and just use flags like on aarch64


axel-h:
Follow-up from a comment from @kent-mcleod in https://github.com/seL4/seL4/pull/274/files#r515668574: `arm_hyp` was needed as a new kernel architecture as back then the scripts that produced a kernel that was given to L4V could only handle changing the kernel architecture, so it was used to mean "aarch32 + hyp" extensions. Now it would be possible to completely remove it and follow the same pattern as `aarch64`. Just need to find a good time to pay the cost of deprecating the old config.

kent-mcleod:
We already checked in the past that this wouldn't have any verification impacts as long as the same kernel_all_cpp.c file was produced in the end (which it would) and their scripts use a different config option for selecting which proofs to use.  
The main work is in going through all of the downstream projects that depend on the arm_hyp config option

bennoleslie:
This would be a great thing to get done. The current scenario is somewhat confusing for new comers.

Is there a specific list of "downstream projects" ?

kent-mcleod:
This is the list of TS projects that have tests in CI: https://docs.sel4.systems/processes/release-process.html#versioned-manifests

PR #1606, opened 2026-03-09
gh-275: Remove arm_hyp KernelSeL4Architecture def

kent-mcleod:
Bring build config in line with aarch64 where
KernelArmHypervisorSupport is the option used to designate hypervisor support.

Just an initial commit to explore the breakages.


tools/hardware/config.py
```diff
@@ -41,6 +41,7 @@ def __init__(self, sel4arch, addrspace_max):
         # Page sizes differ on arm_hyp which is why the alignment changes even though
         # both arm_hyp and arm are 32-bit.
         if sel4arch == 'arm_hyp':
+            #FIX ME
```
kent-mcleod, line 44: This needs an API change to the tool to indicate whether the kernel is running in hyp or not address space translation regime. This is probably something that wants to be known by aarch64 too.

Issue #548, opened 2021-09-04
(minor) Only call cmake_minimum_required from CMakeLists.txt


bennoleslie:
Currently cmake_minimum_required` is called from most `.cmake` files. The exceptions are

* `FindseL4.cmake`
* `configs/*.cmake`

The versions are either 3.8.2 or 3.7.2 (mostly 3.8.2).

It seem unnecessary to call this in all the included `.cmake` files.

Indeed the documentation: https://cmake.org/cmake/help/v3.8/command/cmake_minimum_required.html explicitly states:

> Note Call the cmake_minimum_required() command at the beginning of the top-level CMakeLists.txt file even before calling the project() command. It is important to establish version and policy settings before invoking other commands whose behavior they may affect. See also policy CMP0000.

Of course it doesn't say that calling it multiple times causes any problems as such.

For a user, I find it confusing that we have multiple versions throughout, and also seems unnecessary to state the version repeated in every file like this. It seems like we have less duplication by following the guidelines and just calling it once in `CMakeLists.txt`

As with the other minor enhancements suggested I don't feel that strongly about this, but if there is agreement (or not disagreement) I'll create the PR.

kent-mcleod:
Yea I think that this can be called from just CMakeLists.txt and maybe other files that are imported by other CMake projects (flags.cmake and helpers.cmake).  I'm not sure whether cmake_minimum_required() is supposed to be used to guard against incorrect CMake tool versions, or just to configure CMake policies though.

bennoleslie:
I find the documentation ambiguous on that point as well. I think it does both?

Leoncoci:
I simply deleted the ones that made me unhappy

kent-mcleod:
Reading the documentation again, cmake_minimum_required() also calls cmake_policy(version) which mutates the existing policy context that's in scope. https://cmake.org/cmake/help/v3.22/command/cmake_policy.html#cmake-policy-stack explains that there's a new policy scope created when ever a new directory or CMake file is loaded or included, but not for function calls. My guess would be that calling a function included by another module would cause that function to evaluate in the caller's policy scope.

kent-mcleod:
cmake_minimum_required() probably should be left in the files that are allowed to be included from other projects, but could be removed from files that are more "private" and not expected to be imported from other projects.
These files are all included by other CMake contexts and should have the same min version definition. All other files don't need the min version setting:
- tools/flags.cmake
- tools/helpers.cmake
[...]

PR #1607, opened 2026-03-10
gh-548: Update use of cmake_minimum_required cmd

kent-mcleod:
Only call the command on CMake files that may be the first file evaluated from the project. Remove from all other files.

### DIFF
--- a/FindseL4.cmake
+++ b/FindseL4.cmake
@@ -11,22 +11,17 @@ set(KERNEL_CONFIG_PATH "${CMAKE_CURRENT_LIST_DIR}/configs/seL4Config.cmake" CACH
 mark_as_advanced(KERNEL_PATH KERNEL_HELPERS_PATH KERNEL_CONFIG_PATH)
 
 macro(sel4_import_kernel)
-    add_subdirectory(${KERNEL_PATH} ${CMAKE_BINARY_DIR}/kernel)
+  add_subdirectory(${KERNEL_PATH} ${CMAKE_BINARY_DIR}/kernel)
 endmacro()
 
 macro(sel4_import_libsel4)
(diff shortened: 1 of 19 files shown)


kent-mcleod:
Am I going to get in trouble if I merge without fixing the CMake style?

Indanz:
> Am I going to get in trouble if I merge without fixing the CMake style?

Not from me. The suggested CMake style is horrible and doesn't make any sense, so I hate it with a passion.

kent-mcleod:
How about this: 
- https://github.com/seL4/seL4_tools/pull/249
- https://github.com/seL4/seL4/pull/1612

kent-mcleod:
~ 450 lines of reduce CMake without any semantic changes feels pretty good. 

I reviewed all the style changes make sense.

There's one area where the multiline help messages from cmake_option declarations don't get indented because there actually line continued from the previous line. I left them as is because the whitespace is already a bit messed up anyway and the resulting output in tools like `ccmake` is unchanged.

Indanz:
> ~ 450 lines of reduce CMake without any semantic changes feels pretty good.

Agreed.

> There's one area where the multiline help messages from cmake_option declarations don't get indented because there actually line continued from the previous line. I left them as is because the whitespace is already a bit messed up anyway and the resulting output in tools like `ccmake` is unchanged.

The reason all those help texts are messed up when viewed in `ccmake` is because we have a style checker that enforces indentation and max line lengths and CMake not having native multi-line text support.

kent-mcleod:
> The reason all those help texts are messed up when viewed in `ccmake` is because we have a style checker that enforces indentation and max line lengths and CMake not having native multi-line text support.

The checker no longer forces indentation of strings that continue onto a newline. The subsequent lines after the escaped newline can start at line offset 0 and the style checker is happy, and the render in the CMakeCache.txt is correct.

PR #1608, opened 2026-03-10
gh-557: Add userError and comment about truncate

### LINKED ISSUE #557: Why is length silently truncated when there is no IPC buffer?
This bit of code from `syscall.c:handleInvocation`

```
    if (unlikely(length > n_msgRegisters && !buffer)) {
        length = n_msgRegisters;
    }
```

has me stumped. I think, it is checking if there is no IPC buffer, and if the length is too long (more than fits in registers) it is truncated the length.

The check seems valid to me. I don't get why truncate and not simply stop everything right there?

It's almost certainly going to result in a truncated error when it hits the cap/object specific code, so why not exit much sooner?

Could we put a `userError` there? It is certainly helpful to the user to know "hey you screwed up and no longer have an IPC buffer" vs "you are passing not enough MRs for this particular call".

Thanks!

kent-mcleod:
Explain that the kernel truncates instead of errors deliberately. Add a userError to help debugging this case.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -345,6 +345,11 @@ static exception_t handleInvocation(bool_t isCall, bool_t isBlocking)
     /* Syscall error/Preemptible section */
     length = seL4_MessageInfo_get_length(info);
     if (unlikely(length > n_msgRegisters && !buffer)) {
+        /* If no IPC buffer is present the kernel truncates the maximum message length to n_msgRegisters.
+         * The kernel truncates rather than returns because not all message transfer points in the kernel
+         * are allowed to return an error.
+         */
+        userError("Warning: No IPC buffer for thread. Truncating message length to: %d.", n_msgRegisters);
         length = n_msgRegisters;
     }
 #ifdef CONFIG_KERNEL_MCS


midnightveil (approved):

lsf37:
Should we call this a warning instead of an error? It is explicitly not an error after all.

kent-mcleod:
> Should we call this a warning instead of an error? It is explicitly not an error after all.

Not sure what you mean here. The userError function in the kernel doesn't implement error levels.

lsf37:
> > Should we call this a warning instead of an error? It is explicitly not an error after all.
> 
> Not sure what you mean here. The userError function in the kernel doesn't implement error levels.

That's mostly the problem. Most other output produced by it is an actual error message. So seeing a message like this pop up will indicate to people that an error is happening, which it is not. We could just prefix the message with "Warning: " for now. I don't think we want to actually implement error levels, although that would probably be the proper solution.

midnightveil:
> That's mostly the problem. Most other output produced by it is an actual error message. So seeing a message like this pop up will indicate to people that an error is happening, which it is not. We could just prefix the message with "Warning: " for now. I don't think we want to actually implement error levels, although that would probably be the proper solution.

Is this not always an error?

If one has no IPC buffer then as long as your syscall fits within the number of message registers - i.e. the length in msginfo is < n_msgRegisters then this error won't print.

If one tries to use more registers than that then

- they're immediately going to get a kernel error (if an invocation wants a certain number of arguments - none of our invocations have a variable number of arguments) due to the message length. In this case this error helps 

- they're doing a endpoint invocation between userland threads, in which case you silently have zero registers if you don't check the length, and so this error helps.

Is there a case where this is not useful? Because there should never be a number of registers specified for a syscall where one didn't use it. 

I can't see a case where it's spurious and not an actual error, or at least - not a user level that has set an improper length even if it uses less registers than it says it does.

I can see a situation where the error message is confusing because it assumes that not having an IPC buffer is in and of itself an error, which is not necessarily true, but it is the cause of the limit. (I might reword it to "message has been truncated: without an IPC buffer you can only set a message length up to n_msgRegisters").

lsf37:
> Is this not always an error?

No, it's explicitly not an error by design. We could easily make it one if we wanted to, but that would not be just a message, it'd return an error code. I'm not saying there shouldn't be a message, it should just not be an error message.
 
> If one tries to use more registers than that then
> 
> * they're immediately going to get a kernel error (if an invocation wants a certain number of arguments - none of our invocations have a variable number of arguments) due to the message length. In this case this error helps

Agreed, but it would help the same way if it said "Warning"

> * they're doing a endpoint invocation between userland threads, in which case you silently have zero registers if you don't check the length, and so this error helps.

It does really not -- if we wanted to prevent that, we should fail with error code, not print a message and continue. We decided not to fail and to assume that user land knows how to deal with it. A warning is perfectly fine (because you are absolutely right that there probably is something wrong), but implying that it is an error is inconsistent.

> Is there a case where this is not useful? Because there should never be a number of registers specified for a syscall where one didn't use it.

A warning here is almost always useful I think. I'm sure we can construct cases where the truncation is not a problem, but I do agree that a well-designed user space is unlikely to hit it on purpose.

midnightveil:
OK, I think we all agree then.

Whilst it's called `userError` the actual printed output never says anything about an 'error', e.g.

```
<<seL4(CPU 0) [decodeUntypedInvocation/204 T0xff152400 "rootserver" @80493f6]: Untyped Retype: Insufficient memory (1 * 8388608 bytes needed, 0 bytes available).>>
```

I'm fine with a `Warning:` prefix if you think that's clearer.

<details><summary>so as to not derail this further</summary>
The above happens at the start of a lot of the sel4_test runs, I think you can say that is a warning too, it's intended. But that's really irrelevant.
</details>

lsf37:
> OK, I think we all agree then.

👍 
 
 > so as to not derail this further

but let's :-)

> The above happens at the start of a lot of the sel4_test runs, I think you can say that is a warning too, it's intended. 

I think they are actually problematic, because a lot of people think these are error messages and the test is not working.

kent-mcleod:
I just went with how the discussion ended up at: https://github.com/seL4/seL4/issues/557#issuecomment-1937657229

Indanz (approved):

src/api/syscall.c
```diff
@@ -332,6 +332,10 @@ static exception_t handleInvocation(bool_t isCall, bool_t isBlocking)
     /* Syscall error/Preemptible section */
     length = seL4_MessageInfo_get_length(info);
     if (unlikely(length > n_msgRegisters && !buffer)) {
+        // If no IPC buffer is present the kernel truncates the maximum message length to n_msgRegisters.
+        // The kernel truncates rather than returns because not all message transfer points in the kernel
+        // are allowed to return an error and so the common behavior is specified to truncate.
```
Indanz:
```suggestion
    /* If no IPC buffer is present the kernel truncates the maximum message length to n_msgRegisters.
     * The kernel truncates rather than returns because not all message transfer points in the kernel
     * are allowed to return an error.
     */
    if (unlikely(length > n_msgRegisters && !buffer)) {
```

Indanz:
So the main downside is that if user space ever does this deliberately, it will get spammed with this warning while it's harmless. The chance of this is very low though, it's much more likely that this will help debugging problems for users forgetting to configure an IPC buffer.

kent-mcleod:
There is a kernel config for writing userError messages to the ipc buffer instead of printing them out. And then a libsel4 api for enabling or disabling printing the errors at runtime: seL4_SetPrintError(), seL4_GetPrintError().

Indanz:
I can't find any users of `seL4_SetPrintError`, has it ever been used by anyone?

kent-mcleod:
It was added for converting seL4test to not spew errors that weren't errors. But then the student working on it left and the sel4test part of the work was abandoned. Probably on a branch somewhere. But when I merged the feature, I'd checked that it was working as expected.

lsf37:
Rebased and addressed review comment. If the proof check is passing, this can be merged.

PR #1609, opened 2026-03-10
Draft: Remove Clean to point of unification from page table operations

kent-mcleod:
On all ARM cpus that we support apart from cortex-a8, the memory model feature register indicates that cache clean is not required before TLB maintenance operations: https://developer.arm.com/documentation/ddi0464/e/System-Control/Register-descriptions/Memory-Model-Feature-Register-3?lang=en

### DIFF
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -226,7 +226,7 @@ static inline void invalidateLocalTLB_EL1(void)
 
 static inline void invalidateLocalTLB(void)
 {
-    dsb();
+    dsb_ish();
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
         invalidateLocalTLB_EL2();
         dsb();
@@ -234,7 +234,7 @@ static inline void invalidateLocalTLB(void)
     } else {
         asm volatile("tlbi vmalle1");
     }
-    dsb();
+    dsb_ish();
     isb();
 }
 
@@ -242,17 +242,17 @@ static inline void invalidateLocalTLB_ASID(asid_t asid)
 {
     assert(asid < BIT(16));
 
-    dsb();
+    dsb_ish();
     asm volatile("tlbi aside1, %0" : : "r"(asid << 48));
-    dsb();
+    dsb_ish();
     isb();
 }
 
 static inline void invalidateLocalTLB_VAASID(word_t mva_plus_asid)
 {
-    dsb();
+    dsb_ish();
     asm volatile("tlbi vae1, %0" : : "r"(mva_plus_asid));
-    dsb();
+    dsb_ish();
     isb();
 }
 
@@ -261,7 +261,7 @@ static inline void invalidateLocalTLB_VAASID(word_t mva_plus_asid)
 static inline void invalidateLocalTLB_VMALLS12E1(void)
 {
     asm volatile("tlbi vmalls12e1");
-    dsb();
+    dsb_ish();
     isb();
 }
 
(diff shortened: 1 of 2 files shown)


Indanz:
So I suspect the only reason they were added for AArch64 is because of SMMU support, where it is not guaranteed the SMMU is cache coherent.

This can be fixed by putting cache cleaning operations behind a config and a SMMU check. This does require that a VSpace is bound to SMMU first before it gets a valid ASID, otherwise the clean hasn't been done for previous mappings. This would be a breaking change.

Indanz:
(We may need to map all memory as inner-shareable for this to work for some platforms.)

kent-mcleod:
> (We may need to map all memory as inner-shareable for this to work for some platforms.)

I've only been able to find examples that claim you need ish dsb barriers (in the sense that non-shareable barriers are insufficient), so it does seem like ish is the case for memories accessed by hardware page table walkers even on unicore.

midnightveil:
> I've only been able to find examples that claim you need ish dsb barriers (in the sense that non-shareable barriers are insufficient), so it does seem like ish is the case for memories accessed by hardware page table walkers even on unicore.

Yes, see [discussion here](https://github.com/seL4/seL4/issues/1625), non-shareable memory is not made coherent by hardware for different observers (and page table walker is a separate observer to the PE).

kent-mcleod:
> Yes, see [discussion here](https://github.com/seL4/seL4/issues/1625), non-shareable memory is not made coherent by hardware for different observers (and page table walker is a separate observer to the PE).

The most direct explanation I've found in the past is here: https://developer.arm.com/documentation/ddi0487/maa/-Part-K-Appendixes/-Appendix-K11-Barrier-Litmus-Tests/-K11-5-Cache-and-TLB-maintenance-instructions-and-barriers/-K11-5-3-TLB-maintenance-instructions-and-barriers?lang=en#babggdaf

Indanz:
Yes, but that also says:

> The same code sequence can be used in a multiprocessing system. The Armv8 and later architectures require a PE that executes a TLB maintenance instruction to execute a DSB instruction to ensure completion of the maintenance operation. This ensures that the TLB maintenance instruction is complete on all PEs in the Inner Shareable shareability domain.

That seL4 itself is configured for one core does not mean that it's not running on a multiprocessing system! As far as I know, all platforms supported by seL4 have multiple cores.

To me it seems that on a multiprocessing system you have to map memory as inner shareable, even if you only use one core, otherwise it's not part of the Inner Shareable shareability domain.

But it's all ambiguous, as it explicitly mentions L1 cache for some reason. What about L2 or L3 caches? And does this mean the TLB snoops the cache even if not mapped inner-shareable?

midnightveil:
(unrelated) I've been wondering if when we do page overmapping on AArch64 whether we need a break-before-make sequence for a little while...

Indanz:
> (unrelated) I've been wondering if when we do page overmapping on AArch64 whether we need a break-before-make sequence for a little while...

The chapters "K14.5.3.4 Using break-before-make when updating translation table entries" and "D8.17.1 Using break-before-make when updating translation table entries" are quite clear about that I think.

Indanz:
Would be fun to compare the sel4bench memory mapping benchmark results with these changes.

Indanz:
So a lot of 32-bit failures, but some 64-bit ones too.

kent-mcleod:
> So a lot of 32-bit failures, but some 64-bit ones too.

Might need to make the changes more incrementally to be able to fix the failures.
1. Changing the mapping type to ISH instead of NSH. AFAIK, this shouldn't require making any additional barrier changes.

Issue #1601, opened 2026-03-03
AArch64/hypervisor: why trapping CPACR_EL1 accesses for FPU?


heshamelmatary:
Hello,

I am wondering what's the history/reason why AArch64's FPU handling in EL1/EL0 [traps access](https://github.com/seL4/seL4/blob/efd9426e1771b6cdbcc7e58d3fd58d3df77c337e/include/arch/arm/arch/64/mode/machine/fpu.h#L131) to `CPACR_EL1`, by setting CPTR_EL2.TCPAC (bit 31)? Isn't `CPTR_EL2.TFP` (bit 10) enough?

midnightveil:
Have a look at this comment and associated discussion: https://github.com/seL4/seL4/issues/1569#issuecomment-3642842085

Indanz:
> Have a look at this comment and associated discussion: [#1569 (comment)](https://github.com/seL4/seL4/issues/1569#issuecomment-3642842085)

To clarify, the context is about keeping FPU always enabled the FPU for the kernel and only disable/enable it for lower levels, so the kernel can do FPU context switches without having to first enable the FPU, with the expensive ISB necessary for that.

Indanz:
> I am wondering what's the history/reason why AArch64's FPU handling in EL1/EL0 [traps access](https://github.com/seL4/seL4/blob/efd9426e1771b6cdbcc7e58d3fd58d3df77c337e/include/arch/arm/arch/64/mode/machine/fpu.h#L131) to `CPACR_EL1`, by setting CPTR_EL2.TCPAC (bit 31)? Isn't `CPTR_EL2.TFP` (bit 10) enough?

Good question. It's probably a left-over from the 32-bit Arm port.

Edit: Never mind, I remember now: My removed/edited comment only applies when `ELIsInHost(EL2)`, but that's only true for `HCR_EL2.E2H == 1`. I really hate how Arms make the documentation worse by mixing things up like that. :-(

Indanz:
Actually, you added the lazy FPU save/restore code for AArch32 in 2017, or was that another Hesham Almatary? Why are you asking us, we should be asking you this, you were around when amrzar added the AArch64 code.

heshamelmatary:
> Edit: Never mind, I remember now: My removed/edited comment only applies when ELIsInHost(EL2), but that's only true for HCR_EL2.E2H == 1. I really hate how Arms make the documentation worse by mixing things up like that. :-(

[...]

PR #1611, opened 2026-03-10
aarch64,fpu,hyp: do not trap CPACR_EL1 acceses

heshamelmatary:
The AArch64 FPU trap helpers were modifying both CPTR_EL2.TFP and CPTR_EL2.TCPAC when enabling or disabling FPU trapping. However, only TFP controls trapping of FP/SIMD instructions from EL0/EL1 to EL2.

TCPAC traps accesses to CPACR_EL1 and is unrelated to FPU instruction execution. Trapping CPACR_EL1 is unnecessary for seL4’s FPU handling and can interfere with other architectural features that are controlled via CPACR_EL1.

In particular, CPACR_EL1 is an Architectural Feature Access Control Register and is not limited to FP/SIMD. For example, Morello uses CPACR_EL1.CEN to enable CHERI instructions. Trapping CPACR_EL1 via TCPAC would therefore require special handling or could break guests that legitimately modify CPACR_EL1 for non-FPU features.

Since seL4 only needs to trap FP/SIMD instructions if FPU is disabled per TCB/VM, it is sufficient to control CPTR_EL2.TFP alone. Remove the modification of TCPAC from enableTrapFpu()/disableTrapFpu() and leave CPACR_EL1 accesses untrapped.

This aligns the implementation with the architectural intent of the AArch64 FP/SIMD trapping model and avoids assuming that CPACR_EL1 is used solely for FPU control.

See Issue #1601 for more context and discussion.

### DIFF
--- a/include/arch/arm/arch/64/mode/machine/fpu.h
+++ b/include/arch/arm/arch/64/mode/machine/fpu.h
@@ -90,7 +90,8 @@ static inline void disableTrapFpu(void)
 {
     word_t cptr;
     MRS("cptr_el2", cptr);
-    cptr &= ~(BIT(10) | BIT(31));
+    /* Clear the TFP bit */
+    cptr &= ~BIT(10);
     MSR("cptr_el2", cptr);
     isb();
 }
@@ -128,7 +129,8 @@ static inline void enableTrapFpu(void)
 {
     word_t cptr;
     MRS("cptr_el2", cptr);
-    cptr |= (BIT(10) | BIT(31));
+    /* Set the TFP bit */
+    cptr |= BIT(10);
     MSR("cptr_el2", cptr);
     /* No ISB because we rely on context switch for synchronisation */
 }


lsf37:
This should not have been merged without a proof run. I think it will be fine, but please don't do this kind of thing, we now have 5 commits on top of this that all fail the preprocess check.

Indanz:
Sorry about that. This code seemed deep into arch specific code, so I assumed it was behind the machine interface and wouldn't cause any verification problems.

lsf37:
You were right in this case, but's it's not a sure thing, some of those are visible. The easiest thing to do when you think it's going to pass is to add the `proof-test` label to the PR. That will run the C part of the proofs on the PR and if that passes we should be fine to merge. They don't check all platforms because that would be too expensive to run, but they do check all architectures -- if it's something board-specific only, then maybe do ping one of the verification people (e.g. me), but those are rare.

The important part after that is to not leave too long a gap for bumping the verification manifest, so that a) all proofs are actually checks, and b) subsequent PRs and commits don't fail the preprocess test and people start ignoring it.

kent-mcleod:
For what's it worth, we were only merging prs weren't changing the preprocess diff beyond the fpu change.

lsf37:
It's alright, nothing actually bad happened, I just want to make sure this is not the normal state. There is a bit more automation that I can add for bumping the verification manifest, so the preprocess test gets back in sync more quickly in this kind of situation, but I need to do some GitHub action wrangling for that first.

PR #1612, opened 2026-03-10
test style

kent-mcleod:


CMakeLists.txt
```diff
@@ -661,49 +535,41 @@ set(CPPExtraFlags "-I${CMAKE_CURRENT_BINARY_DIR}/generated")
...
+        EXTRA_FLAGS -CC -P "${CPPExtraFlags}")
 
 add_custom_command(
-    OUTPUT circular_includes_valid
-    COMMAND ${PYTHON3} ${CIRCULAR_INCLUDES} --ignore kernel_all_copy.c < kernel_all.i
-    COMMAND touch circular_includes_valid
-    DEPENDS kernel_i_wrapper kernel_all.i
-)
+  OUTPUT circular_includes_valid COMMAND ${PYTHON3} ${CIRCULAR_INCLUDES} --ignore kernel_all_copy.c
+                                         < kernel_all.i COMMAND touch circular_includes_valid
```
kent-mcleod, line 565: This second COMMAND seems broken, but the other cases of add_custom_command seem to organize in a nicer way.

kent-mcleod:
Turned out a lot better than I thought

CMakeLists.txt
Indanz: Is there a way to allow the original without triggering a style error?

The insistence of the style checker of forcing a particular line breaking is what makes it so horrible. It would be nice if that was still a user choice, at least to some extend.

kent-mcleod: I think in this specific case, it's losing track of the internal structure of the arguments within each COMMAND sequence. Maybe quoting the `<` char is enough to resolve this, but who knows.

kent-mcleod: Something that would make this easier to manage is getting the style tools to only apply to the diff rather than the entire file. Then in cases where the tool doesn't do something you like, its' easier to  just ignore it for the particular commit and it won't come up on all subsequent PRs.

kent-mcleod:
I don't plan to merge this, it's just illustrative of the impact of changing the cmake style tool config.

I expect that the style updates would continue to happen as files are changed in future commits.

PR #289, opened 2020-12-07
Correctly invalidate I-cache for VIPT caches in hyp mode

bennoleslie:
CPU cores that use a VIPT I-cache are explicitly configured as
part of the kernel config.

In the case where the kernel is configured for VIPT I-cache and
hypervisor mode the I-cache invalidate range is implemented using
a full I-cache flush. This is required as there is no practical
way in which the kernel can invalidate a specific range.

See: https://developer.arm.com/documentation/ddi0500/j/System-Control/AArch64-register-descriptions/Cache-Type-Register?lang=en for confirmation that A53 is VIPT.

(See also [SELFOUR-2830])

[SELFOUR-2830]: https://sel4.atlassian.net/browse/SELFOUR-2830

PR #1619, opened 2026-03-17
arm64: start PPTR window mapping from physBase()

hlyytine:
I'm porting seL4 to NVIDIA Orin boards (tegra234). They have a hole between 0x40008000 - 0x80000000 which must not be mapped as normal cacheable, or else Cortex-A78 speculative page table walker gives you random unrecoverable RAS errors.

This is also how it is done on 32-bit ARM.

### DIFF
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -233,11 +233,31 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
                                                                                             attr_index);
 }
 
+static BOOT_CODE void map_kernel_window_range(paddr_t start, paddr_t end)
+{
+    pptr_t vaddr = (pptr_t)ptrFromPAddr(start);
+    for (paddr_t paddr = start; paddr < end; paddr += BIT(seL4_LargePageBits)) {
+        armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
+                                                                                         KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+                                                                                                                        0, // XN
+#else
+                                                                                                                        1, // UXN
+#endif
+                                                                                                                        paddr,
+                                                                                                                        0,                        /* global */
+                                                                                                                        1,                        /* access flag */
+                                                                                                                        SMP_TERNARY(SMP_SHARE, 0),        /* Inner-shareable if SMP enabled, otherwise unshared */
+                                                                                                                        0,                        /* VMKernelOnly */
+                                                                                                                        NORMAL
+                                                                                                                    );
+        vaddr += BIT(seL4_LargePageBits);
+    }
+}
+
 BOOT_CODE void map_kernel_window(void)
 {
 
-    paddr_t paddr;
-    pptr_t vaddr;
     word_t idx;
 
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
@@ -265,23 +285,8 @@ BOOT_CODE void map_kernel_window(void)
     }
 
     /* map the kernel window using large pages */
-    vaddr = PPTR_BASE;
-    for (paddr = PADDR_BASE; paddr < PADDR_TOP; paddr += BIT(seL4_LargePageBits)) {
-        armKSGlobalKernelPDs[GET_KPT_INDEX(vaddr, KLVL_FRM_ARM_PT_LVL(1))][GET_KPT_INDEX(vaddr,
-                                                                                         KLVL_FRM_ARM_PT_LVL(2))] = pte_pte_page_new(
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-                                                                                                                        0, // XN
-#else
-                                                                                                                        1, // UXN
-#endif
-                                                                                                                        paddr,
-                                                                                                                        0,                        /* global */
-                                                                                                                        1,                        /* access flag */
-                                                                                                                        SMP_TERNARY(SMP_SHARE, 0),        /* Inner-shareable if SMP enabled, otherwise unshared */
-                                                                                                                        0,                        /* VMKernelOnly */
-                                                                                                                        NORMAL
-                                                                                                                    );
-        vaddr += BIT(seL4_LargePageBits);
+    for (idx = 0; idx < ARRAY_SIZE(avail_p_regs); idx++) {
+        map_kernel_window_range(avail_p_regs[idx].start, avail_p_regs[idx].end);
     }
 
     /* put the PD into the PUD for device window */


Indanz:
On AArch32 `physBase()` is used as `PADDR_BASE`.
On AArch64 `physBase()` is only used for `KERNEL_ELF_PADDR_BASE`, and `PADDR_BASE` is hard-coded to 0. That means that on AArch64 `physBase()` is not the physical memory base address, which is very confusing.

Although both look the same, they are not really. That said, the script setting `physBase()` looks at the first physical normal memory address and doesn't know that it will only be used for the kernel's own location, so practically it ends up having the value you want here.

[Alex Potanin](https://mattermost.trustworthy.systems/sel4-external/messages/@potanin) also ran into this issue recently for an Orion Nano, his work can be found at https://fiducia-lang.github.io/blog/sel4-jetson-orin-nano.html (but there is a fair bit of AI generated nonsense code over there, so beware). The fix I suggested on Mattermost was to only map the actual memory present and implemented in [this](https://github.com/seL4/seL4/commit/7c1f8d9c3de833985cb068153266fc3faefed14b) commit (which needs some cleanup), that solved the problem for him. Maybe good to exchange information for other issues either of you have ran into.

Your solution works in case such problematic mappings happen before actual SDRAM, but don't solve the problem for speculative accesses past the actual SDRAM. My preferred solution is to use 1GB mappings that cover the known SDRAM area. I'll try to write a patch tomorrow.

hlyytine:
I realized we already have array of RAM regions there.

edit: and actually seems that I was ignorant enough to not read Alex's commit...

src/arch/arm/64/kernel/vspace.c
```diff
@@ -233,11 +233,31 @@ BOOT_CODE void map_kernel_frame(paddr_t paddr, pptr_t vaddr, vm_rights_t vm_righ
                                                                                             attr_index);
 }
 
+static BOOT_CODE void map_kernel_window_range(paddr_t start, paddr_t end)
+{
+    pptr_t vaddr = (pptr_t)ptrFromPAddr(start);
+    for (paddr_t paddr = start; paddr < end; paddr += BIT(seL4_LargePageBits)) {
```
midnightveil, line 239: If the motivation for this change is that we don't want to map in reserved memory, only mapping large pages seems like it would still cause issues depending on the memory layout; i.e. we should also enforce that `avail_p_regs` has appropriate alignment as well. (and the bits and pieces chopped off by alignment can become seL4 "device" memory, and so don't need to be mapped into seL4)

Say for instance, if memory looks like this:

```
0...0x10000000   RAM
0x10000000...0x16000000    unused/device
0x16000000...0x17ff0000   RAM
0x17ff0000...0x18000000   Reserved for secure world
```

Then large pages of size `0x20000` would cover that secure-world protected memory, and so speculative read/writes could access that memory and cause faults.

Similarly with any true device memory.

Indanz: You make a valid point which is good to be aware of.

True device memory won't be in the same 1GB region as RAM, so it's only a problem for secure world memory. Assuming secure world memory is at least large page size aligned is a fair assumption I think (and easily fixed with an overlay if it isn't).

lsf37:
How does this interact with the kernel using its own mappings for cache maintenance operations in hyp mode?

So far, the check for these just checks whether that address is in the kernel window, because the entire kernel window is mapped: 

https://github.com/seL4/seL4/blob/3d449dba1523dab949b936f187a66333ce41625d/src/arch/arm/64/kernel/vspace.c#L1764-L1770

This check will now succeed for regions that are unmapped in the kernel, and if I remember correctly, these could be regions that are available to the user as device untypeds (e.g. device MMIO). I'm unclear on whether there might be a kind of region that is not mapped in the kernel after this, and is accessible to user level via device untyped, and that the user might legitimately need/want to do cache maintenance operations on. Maybe not (because if yes, it should really just be RAM, no?).

If not, we "just" need to extend this check such that it never uses a cache maintenance operation outside the mapped area, otherwise the kernel crashes. One solution would be a page table walk of the kernel page tables. From the verification side, this would be annoying (we'd be forced to model more about the kernel page tables, which maybe we should anyway), but it would be possible. The walk should probably be in software so that we can reason about it, although I'm not completely sure about this part.

If there are legitimate cases where cache maintenance would be needed on kernel IPAs that are not mapped, then we'd need to do something like dynamically map in that page, do the cache maintenance, and unmap that page again.

In either case, there would be performance impact for cache maintenance in hyp mode, although it is probably dominated by the cache flush anyway, so maybe not a problem.

midnightveil:
>  I'm unclear on whether there might be a kind of region that is not mapped in the kernel after this, and is accessible to user level via device untyped, and that the user might legitimately need/want to do cache maintenance operations on. Maybe not (because if yes, it should really just be RAM, no?).

One case here would be I think shared memory in a Multikernel setup; since there we use device so that the kernel neber shared "normal" memory. Others would include the alignment-reserved memory on e.g. RPI4, as well as something like TCM memory 

However, I'm honestly not sure what is the point of the kernel cache maintenance operations, if you can just do it from userspace yourself. Admittedly, only clean & invalidate, but I don't know if I've ever seen a use for invalidate-only, and besides even if it's faster in some circumstance it would probably be not true because of syscsll costs.

lsf37:
> > I'm unclear on whether there might be a kind of region that is not mapped in the kernel after this, and is accessible to user level via device untyped, and that the user might legitimately need/want to do cache maintenance operations on. Maybe not (because if yes, it should really just be RAM, no?).
> 
> One case here would be I think shared memory in a Multikernel setup; since there we use device so that the kernel neber shared "normal" memory. Others would include the alignment-reserved memory on e.g. RPI4, as well as something like TCM memory

Right, yes, the multikernel setup definitely has plain RAM that would be in the kernel window but is device untyped. So we do need something more than just a check.


> However, I'm honestly not sure what is the point of the kernel cache maintenance operations, if you can just do it from userspace yourself. Admittedly, only clean & invalidate, but I don't know if I've ever seen a use for invalidate-only, and besides even if it's faster in some circumstance it would probably be not true because of syscsll costs.

IIRC the point was that you want to be able to prevent cache maintenance operations for components you don't trust. I think Kent had a draft PR or issue somewhere with making the accessibility of cache maintenance ops by user level a runtime TCB flag (like the FPU flag) instead of a build time flag. If we had that, we could probably remove it from the cap API and just control whether you are allowed to use the instruction or not. Not sure of performance impact, but that does sound like a solution that would get rid of all of these "may crash in kernel space" problems, because they will then be a user-level fault which can be handled.

Indanz:
> How does this interact with the kernel using its own mappings for cache maintenance operations in hyp mode?

Good point. Badly. :-)
 
For non-hyp we switch the page table to the one that is used for the mapping of the page, and we check that the page is mapped, so everything is fine. Not so for EL2 as you point out.

> I'm unclear on whether there might be a kind of region that is not mapped in the kernel after this, and is accessible to user level via device untyped

Yes, there will be.

> If there are legitimate cases where cache maintenance would be needed on kernel IPAs that are not mapped, then we'd need to do something like dynamically map in that page, do the cache maintenance, and unmap that page again.

This seems the safest option, if we do want to support this feature.

> However, I'm honestly not sure what is the point of the kernel cache maintenance operations, if you can just do it from userspace yourself. Admittedly, only clean & invalidate, but I don't know if I've ever seen a use for invalidate-only

Agreed. Because invalidate-only is inherently dangerous, I think on most actual implementations it silently behaves as clean & invalidate when executed from EL0.

> I think Kent had a draft PR or issue somewhere with making the accessibility of cache maintenance ops by user level a runtime TCB flag (like the FPU flag) instead of a build time flag.

Yes, we have:
- [Pre-RFC: Arm: Add TCB policy support for controlling user level access to certain system registers](https://sel4.discourse.group/t/pre-rfc-arm-add-tcb-policy-support-for-controlling-user-level-access-to-certain-system-registers/507)
- [RFC-16: New capability for the PMU](https://github.com/seL4/rfcs/pull/22)

> If we had that, we could probably remove it from the cap API and just control whether you are allowed to use the instruction or not.

This would be my preferred solution. Cache maintenance syscalls is very awkward and weird for EL2, as they all work on virtual addresses and the kernel can't use the correct page table, leading to issues like #289.

That said, it would be less fine grained than the existing page cap based granularity (counter argument is the above issue though).

> Not sure of performance impact, but that does sound like a solution that would get rid of all of these "may crash in kernel space" problems, because they will then be a user-level fault which can be handled.

It will cost two extra instructions per system register on context switch. So ideally, most useful options are located in only a few registers. Now that we have `seL4_TCB_SetFlags()`, it's easy to add more flags.

Indanz:
> > If there are legitimate cases where cache maintenance would be needed on kernel IPAs that are not mapped, then we'd need to do something like dynamically map in that page, do the cache maintenance, and unmap that page again.
> 
> This seems the safest option, if we do want to support this feature.

Actually, this won't be safe to do, for the same reasons why we do not want to map that memory in the first place: Accessing it may not be safe. That includes doing cache operations on them. But even if not those, while we temporarily map the page, speculative accesses could be done.

So the only (simple) safe option is to fail the operation when done on memory not mapped by an EL2 kernel.

midnightveil:
Even for EL1, that seems like a bad idea in its current state.

If performing cache maintenance operations on what is architecture-device memory can cause faults / SError then we shouldn't allow them either. And since there's no guarantee that's not the case unless it's seL4-normal memory (well, who knows what happens if you get ECC errors while seL4 is running).[^1]

It looks like not being able to expose "Cache Invalidate" to userland isn't a big issue; from reading the ARM specification VMs operating at EL1 that try to perform a cache invalidate gets silently upgraded to cache clean & invalidate.

> The AArch64 Data cache invalidate instructions, DC IVAC and DC ISW, at EL1 and EL0, and the AArch32 Data cache invalidate instructions DCIMVAC and DCISW, at EL1, perform a cache clean as well as a cache invalidation
> if both of the following apply:
> — The value of HCR.VM is 1.
> — The instruction is executed in Non-secure state, or EL3 is not implemented.


[^1]: Note also that architecturally cache maintenance on non-cacheable/device memory (as the page table entries) still affect caches and have defined behaviour:
     > For VA-based cache maintenance instructions, the instruction operates on the caches regardless of the memory type
and cacheability attributes marked for the memory address in the VMSA translation table entries. This means that
the effects of the cache maintenance instructions can apply regardless of:
    > Whether the address accessed:
    > — Is Normal memory or Device memory.
    > — Has the Cacheable attribute or the Non-cacheable attribute

midnightveil:
For page/frame-based cache maintenance operations, we should use the same checks as we do for IPC buffer and otherwise, i.e. 'cap_frame_cap_get_capFIsDevice' must not be true.

For the VSpace ones, we can also do a similar check, as we already only allow invocations local to a single page.

This would probably be the least intrusive change, and not require extra lookups.

KurtWu10:
Executing `DC IVAC` has an undefined behaviour when running at EL0, according to the [ASL](https://support.arm.com/documentation/ddi0601/2026-06/AArch64-Instructions/DC-IVAC--Data-or-unified-Cache-line-Invalidate-by-VA-to-PoC). In practice, it triggers a fault.

midnightveil:
Not just in practice, `UNDEFINED` means generate an exception. (unless accompanying verbiage exists about UNPREDICTABLE etc). But that's a separate concern really, because it happens *anyway*.

image

kent-mcleod:
> Executing `DC IVAC` has an undefined behaviour when running at EL0, according to the [ASL](https://support.arm.com/documentation/ddi0601/2026-06/AArch64-Instructions/DC-IVAC--Data-or-unified-Cache-line-Invalidate-by-VA-to-PoC). In practice, it triggers a fault.

`DC IVAC` is not an instruction that is defined to exist at EL0. Generally when I've been referring to user level cache ops, it's just the ones enabled by setting `SCTLR_EL1.UCI` to 1 (`DC CVAU`, `DC CIVAC`, `DC CVAC`, or `IC IVAU`)

Indanz:
> Even for EL1, that seems like a bad idea in its current state.
> 
> If performing cache maintenance operations on what is architecture-device memory can cause faults / SError then we shouldn't allow them either.

In that case it can already cause SErrors by merely writing to the memory, cache maintenance instructions don't change anything here. Only a direct fault would change things, but if we have write permissions in the page tables, the cache flush instructions won't themselves fault. So I think EL1 is fine.

I guess the above also applies to EL2 with its own temporary mapping, hm. So perhaps it's not as dire as I thought.

midnightveil:
> In that case it can already cause SErrors by merely writing to the memory, cache maintenance instructions don't change anything here. 

Yes, but not in kernel mode. seL4 isn't designed to take faults in kernel mode, all sorts of things could break.

Indanz:
> > In that case it can already cause SErrors by merely writing to the memory, cache maintenance instructions don't change anything here.
> 
> Yes, but not in kernel mode. seL4 isn't designed to take faults in kernel mode, all sorts of things could break.

SError is asynchronous, it can arrive at any time and the kernel will halt no matter whether it arrives in user mode or kernel mode.

With direct fault I meant a synchronous fault.

kent-mcleod:
> SError is asynchronous, it can arrive at any time and the kernel will halt no matter whether it arrives in user mode or kernel mode.

SError is still maskable by the exception level that it's routed to be delivered to.

lsf37:
> > SError is asynchronous, it can arrive at any time and the kernel will halt no matter whether it arrives in user mode or kernel mode.
> 
> SError is still maskable by the exception level that it's routed to be delivered to.

We might already be doing this, but I think the following would work with kernel verification assumptions on faults:

- mask SError while in kernel mode (EL1 when kernel in EL1, EL2 when kernel in EL2)
- unmask/allow SError in user mode (incl EL1 when kernel in EL2)

If I understand the machine behaviour correctly, this would mean that the kernel continues executing normally when an SError is generated by e.g. speculative execution (we do prove that we're not touching things ourselves that are not memory), but at context switch, when the SError is unmasked, we would immediately take the exception.

I think that is the behaviour I've been seeing on the TX2 verification build, only that the kernel's reaction to SError in user mode there is to halt, which we should change.

In that scenario, we'd still need to figure out what to do with SErrors raised in user mode (generated by either user or kernel), but it would not change the verification assumptions on how code execution flows. Ignoring it + debug print does not actually sound as bad as I thought in the past.

There should eventually be a better mechanism, but I'm not sure what. The async nature of it is very annoying. We could e.g. have a global SError handler, or one per domain or even one per current thread or something like that, but all of them would introduce a global information flow that we cannot control unless we can put some bounds of the async delay. I.e. if I generate an SError just before context switch, the kernel masks before it is raised, then unmasks when the next domain runs, we have a SError raised in domain A delivered to domain B. Is there a way to clear/ignore them on domain switch? (or maybe even to re-inject them later?). I.e. we could on kernel exit umask, take the exception, and try to avoid specific scenarios. If we do this carefully, that could work from the verification perspective similarly to normal IRQs. I.e. if it looks to C like sequential execution of one machine interface function, it'd be fine for the C model.

Indanz:
> If I understand the machine behaviour correctly, this would mean that the kernel continues executing normally when an SError is generated by e.g. speculative execution (we do prove that we're not touching things ourselves that are not memory), but at context switch, when the SError is unmasked, we would immediately take the exception.

That should be possible, yes.

> In that scenario, we'd still need to figure out what to do with SErrors raised in user mode

IOMMU/SMMU have a similar out-of-bound fault problem. There at least the fault could be traced back to the source and delivered to a specific task, but that might be more hassle than it's worth it (and might be SoC specific).

Maybe have a syscall to bind global faults to an endpoint or TCB (if that's easier)?

> (generated by either user or kernel), but it would not change the verification assumptions on how code execution flows. Ignoring it + debug print does not actually sound as bad as I thought in the past.

The only problem is if the SError was caused by a catastrophic system error, e.g. system glitches, uncorrectable ECC errors or other undefined behaviours. But this isn't different than bit-flips caused by cosmic rays, almost no one is queueing up for more expensive ECC memory, so deferring the SError till user space should be good enough. We can always leave the current behaviour as a user config option (as it initially was intended).

> There should eventually be a better mechanism, but I'm not sure what. The async nature of it is very annoying.

Yes.

> We could e.g. have a global SError handler, or one per domain or even one per current thread or something like that, but all of them would introduce a global information flow that we cannot control unless we can put some bounds of the async delay. I.e. if I generate an SError just before context switch, the kernel masks before it is raised, then unmasks when the next domain runs, we have a SError raised in domain A delivered to domain B. Is there a way to clear/ignore them on domain switch? (or maybe even to re-inject them later?). 

It doesn't matter whether this is possible or not, because the main problem is that we don't know which domain caused an SError. Code in domain A can trigger an SError that only arrives when we run domain B. Example: Write to read-only device registers in cached memory. The SError will only happen when that cache line gets evicted, which could happen any moment, including hours later.

Arm has RAS extensions that give more info, like [`DISR_EL1`](https://support.arm.com/documentation/100403/0201/register-descriptions/aarch64-system-registers/disr-el1--deferred-interrupt-status-register--el1) and the `ESB`, error synchronisation barrier instruction, but I don't think the kernel can figure out which domain the error belongs to based on virtual or physical address.

I suspect the only practical solution is to have one global error handler where all async errors arrive.

But generally speaking, communication via SError is a bit silly. Deliberate ones would involve device memory access to a device that has system-wide side-effects via SError. That's no difference than other devices causing global observable state changes (e.g. CPU throttling because of busy looping causing higher temperatures, or DMA).

PR #1620, opened 2026-03-17
arm64: replace set/way cache ops with VA-based ops

hlyytine:
I bumped this while porting seL4 to NVIDIA tegra234. 

On modern aarch64, cache ops by set/way are flaky at best. Since we know the VA range during boot, we can use VA-based cache clean/invalidate.

### DIFF
--- a/src/arch/arm/armv/armv8-a/64/cache.c
+++ b/src/arch/arm/armv/armv8-a/64/cache.c
@@ -4,105 +4,41 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+#include <config.h>
+#include <arch/machine.h>
 #include <arch/machine/hardware.h>
+#include <kernel/boot.h>
 
-static inline void cleanByWSL(word_t wsl)
+static void get_boot_flush_region(word_t *start, word_t *end)
 {
-    asm volatile("dc csw, %0" : : "r"(wsl));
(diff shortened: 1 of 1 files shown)


src/arch/arm/armv/armv8-a/64/cache.c
```diff
@@ -4,105 +4,41 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+#include <config.h>
+#include <arch/machine.h>
 #include <arch/machine/hardware.h>
+#include <kernel/boot.h>
```
Indanz, line 10: You should mark the `cleanInvalidate_*` functions as `BOOT_CODE` if you make them depend on boot-only code.

src/arch/arm/armv/armv8-a/64/cache.c
```diff
@@ -4,105 +4,41 @@
...
-                               (s << lbits) | (l << 1));
-                }
-            }
-        }
+    if (rootserver.paging.end != 0) {
+        /* After init_freemem: flush up to rootserver allocations + margin */
+        *end = rootserver.paging.end + (2 * 1024 * 1024);
+    } else {
+        /* Early boot: rootserver not yet initialized, flush kernel image */
+        *end = (word_t)ki_end + (2 * 1024 * 1024);
```
Indanz, line 21: Can as well always do this if we're guessing anyway.

PR #1621, opened 2026-03-17
arm64: fix cache maintenance completion barriers

hlyytine:
ARM reference manual says `DSB` must be used after these cache ops, not `DMB`. Actually `DSB ISH` would be enough but since `dsb()` (which expands to `DSB SY`) is used everywhere else, let us use it as well.

### DIFF
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -280,19 +280,19 @@ void lockTLBEntry(vptr_t vaddr);
 static inline void cleanByVA(vptr_t vaddr, paddr_t paddr)
 {
     asm volatile("dc cvac, %0" : : "r"(vaddr));
-    dmb();
+    dsb();
 }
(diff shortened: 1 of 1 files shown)


Indanz:
Reference please. In chapter K14 Barrier Litmus Tests, especially section "K14.5.1 Data cache maintenance instructions", for `dc cvac` only DMBs are used, no DSBs.

I think DSB is only needed when a `dc` is followed by a dependent `ic`, to force the clean to finish before the instruction chase operation starts. So perhaps `invalidateByVA_I` and `invalidate_I_PoU` should have an extra `DSB` before doing `ic`, just to be sure. But most callers already do the DSB, so the convention seems to be it's the caller's responsibility.

Indanz:
There may some missing barrier somewhere else, but it's not in these places, so closing the PR.

In other words, if this fixes something, it hides the missing barrier bug instead of properly fixing it.

PR #1622, opened 2026-03-17
arm64: Use Point of Coherence page table cache maintenance ops

hlyytine:
I bumped into this while porting seL4 to NVIDIA tegra234 with `ARM_HYPERVISOR_SUPPORT` (seL4 running at EL2).

ARM reference manual says that data cache coherency is guaranteed by HW, but when walking the page tables, it's a different matter: stage-1 page tables are walked from PoU, while stage-2 page tables are walked from PoC. Since modifying page tables is not that hot path, let us do cache clean/invalidate to PoC even if we are running at EL1, much cleaner code.

When I changed code to use PoC, sel4test caught a new corner case, the commit message on flushing cache on final cap delete explains it all.

With this and sel4#1621, sel4#1620 and sel4#1619 I finally managed to run sel4test on tegra234 without any errors.

### DIFF
--- a/include/arch/arm/arch/machine.h
+++ b/include/arch/arm/arch/machine.h
@@ -53,8 +53,8 @@ static inline void clearMemory(word_t *ptr, word_t bits)
 static inline void clearMemory_PT(word_t *ptr, word_t bits)
 {
     memzero(ptr, BIT(bits));
-    cleanCacheRange_PoU((word_t)ptr, (word_t)ptr + BIT(bits) - 1,
-                        addrFromPPtr(ptr));
+    cleanInvalidateCacheRange_RAM((word_t)ptr, (word_t)ptr + BIT(bits) - 1,
+                                  addrFromPPtr(ptr));
 }
 
 #ifdef ENABLE_SMP_SUPPORT
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1028,7 +1028,7 @@ void unmapPageTable(asid_t asid, vptr_t vptr, pte_t *target_pt)
     /* If we found a pt then ptSlot won't be null */
     assert(ptSlot != NULL);
     *ptSlot = pte_pte_invalid_new();
-    cleanByVA_PoU((vptr_t)ptSlot, pptr_to_paddr(ptSlot));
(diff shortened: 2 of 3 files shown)


Indanz:
> ARM reference manual says that data cache coherency is guaranteed by HW, but when walking the page tables, it's a different matter: stage-1 page tables are walked from PoU, while stage-2 page tables are walked from PoC.

Argh, that's horrible, why would they do that? Reference please.

> Since modifying page tables is not that hot path, let us do cache clean/invalidate to PoC even if we are running at EL1, much cleaner code.

Maybe, but I think I disagree with the invalidate. That's good for debugging, but I don't see what advantage it has otherwise. On the downside, it will slow down future mappings.

> When I changed code to use PoC, sel4test caught a new corner case, the commit message on flushing cache on final cap delete explains it all.

I'm puzzled this can happen at all. At that point no one should have the page table active and the whole VSpace was invalidated by ASID/VMID, so why would the hw page table walker even look at the stale page table? Or is none of this done for revoke?

You clearly ran into this, so there must be a problem here. I just want to get to the bottom of it so we do the proper fix instead of accidentally working around the problem.

I'll have another look tomorrow, it's getting late here.

Indanz:
The Arm manual states:

_"The Armv8 and later architectures require that translation table walks look in the data or unified caches at L1, so such systems do not require data cache cleaning."_

The section "D8.2.6.1 Ordering of memory accesses from translation table walks" does not add an exception for stage 2 translations.

I have the impression you're chasing a rabbit down the rabbit hole instead of nailing the cause of your problems.

hlyytine:
You are definitely right. Period.

Longer story: I thought I had it derived from architecture manual but failed to reproduce it after night's sleep. Turns out my build system built a UP kernel instead of SMP, meaning NSH shareability was used instead of ISH as used with SMP. NSH on UP would be completely OK, except that I found out UEFI uses ISH mappings and most likely powers on all cores before they go to WFI/WFE, and ISH-influenced snoop units were wreaking havoc when my kernel was (inadvertently) running with UP/NSH. Hence the disorientation.

Had to fix build system to allow GICs for more than 8 cores, but that's another story and PR. Closing this one.

Indanz:
Yes, that NSH for unicore is plain wrong I think, we should always use ISH.

PR #1623, opened 2026-03-19
x86: Fix AMD CPU model assignment typo

cuiweixie:
When original.family >= 0xF, the extended model was incorrectly assigned to ci->display.family instead of ci->display.model. This matches the Intel implementation and AMD CPUID specification.


midnightveil (approved):

midnightveil:
Your commit title fails the gitlint checks.

cuiweixie:
just git commit --amend to retrigger ci.

Indanz:
As neither display.model, nor display.family is used anywhere, not exported to use space, this change shouldn't affect runtime behaviour, so no need for a CHANGES update or anything like that.

Indanz:
Please don't add "Made-with: Cursor" to your commit messages, that's just advertisement for some product that shouldn't be there.

It wasn't in your original commit, that's why I missed it before merging.

PR #1624, opened 2026-03-19
scheduleTCB: simplify scheduler action logic

michaelmcinerney:
The call to rescheduleRequired within scheduleTCB
is performed only when the scheduler action is
ResumeCurrentThread, and therefore will only set
the scheduler action to ChooseNewThread. This
removes the call to rescheduleRequired and
replaces it with the direct update to the
scheduler action.

### DIFF
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -586,7 +586,8 @@ void scheduleTCB(tcb_t *tptr)
     if (tptr == NODE_STATE(ksCurThread) &&
         NODE_STATE(ksSchedulerAction) == SchedulerAction_ResumeCurrentThread &&
         !isSchedulable(tptr)) {
-        rescheduleRequired();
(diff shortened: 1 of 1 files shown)


src/kernel/thread.c
```diff
@@ -589,7 +589,7 @@ void scheduleTCB(tcb_t *tptr)
     if (tptr == NODE_STATE(ksCurThread) &&
         NODE_STATE(ksSchedulerAction) == SchedulerAction_ResumeCurrentThread &&
         !isSchedulable(tptr)) {
-        rescheduleRequired();
+        NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
```
lsf37, line 590: Happy with the code change, but we should probably put parts of the commit message as a comment here so people know why we are not calling `reschedule_required` as usual.

lsf37 (approved):

Indanz (approved):

PR #1628, opened 2026-03-27
Add explicit type conversions

### LINKED ISSUE #1187: Should we add `-Wconversion` to compiler flags?

kent-mcleod:
This is in preparation for turning on -Wconversion:
- fastpath.h: Cast to register_t from a + b is safe
- fastpath.h: Cast to hw_asid_t is safe from bitfield.
- machine.h: Change clearMemory to take word_t type. This is aligned with arm.
- util.c: int cast to match up with compiler builtins
- util.c: long cast in str_to_long()
- fastpath.c: syscall_t cast for negative syscall numbers
- tcb.c: intentionally cast to breakpoint number type.
- untyped.c: Retype chunk to word_t.

### DIFF
--- a/include/arch/arm/arch/32/mode/fastpath/fastpath.h
+++ b/include/arch/arm/arch/32/mode/fastpath/fastpath.h
@@ -47,7 +47,7 @@ static inline void FORCE_INLINE switchToThread_fp(tcb_t *thread, pde_t *cap_pd, 
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
         vcpu_switch(thread->tcbArch.tcbVCPU);
     }
-    hw_asid = pde_pde_invalid_get_stored_hw_asid(stored_hw_asid);
+    hw_asid = (hw_asid_t)pde_pde_invalid_get_stored_hw_asid(stored_hw_asid);
     armv_contextSwitch_HWASID(cap_pd, hw_asid);
 
 #ifdef CONFIG_BENCHMARK_TRACK_UTILISATION
--- a/include/arch/riscv/arch/fastpath/fastpath.h
+++ b/include/arch/riscv/arch/fastpath/fastpath.h
@@ -91,7 +91,7 @@ static inline void fastpath_copy_mrs(word_t length, tcb_t *src, tcb_t *dest)
     /* assuming that length < n_msgRegisters */
     for (i = 0; i < length; i ++) {
         /* assuming that the message registers simply increment */
-        reg = msgRegisters[0] + i;
+        reg = (register_t)(msgRegisters[0] + i);
         setRegister(dest, reg, getRegister(src, reg));
     }
 }
--- a/include/arch/x86/arch/64/mode/fastpath/fastpath.h
+++ b/include/arch/x86/arch/64/mode/fastpath/fastpath.h
@@ -103,7 +103,7 @@ static inline void fastpath_copy_mrs(word_t length, tcb_t *src, tcb_t *dest)
     /* assuming that length < n_msgRegisters */
     for (i = 0; i < length; i ++) {
         /* assuming that the message registers simply increment */
-        reg = msgRegisters[0] + i;
+        reg = (register_t)(msgRegisters[0] + i);
         setRegister(dest, reg, getRegister(src, reg));
     }
 }
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -39,7 +39,7 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
      * saved fault. */
     if (unlikely(fastpath_mi_check(msgInfo) ||
                  fault_type != seL4_Fault_NullFault)) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 
     /* Lookup the cap */
@@ -48,7 +48,7 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
     /* Check it's an endpoint */
     if (unlikely(!cap_capType_equals(ep_cap, cap_endpoint_cap) ||
                  !cap_endpoint_cap_get_capCanSend(ep_cap))) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 
     /* Get the endpoint address */
@@ -60,13 +60,13 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
 
     /* Check that there's a thread waiting to receive */
     if (unlikely(endpoint_ptr_get_state(ep_ptr) != EPState_Recv)) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 
     /* ensure we are not single stepping the destination in ia32 */
 #if defined(CONFIG_HARDWARE_DEBUG_API) && defined(CONFIG_ARCH_IA32)
     if (unlikely(dest->tcbArch.tcbContext.breakpointState.single_step_enabled)) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 #endif
 
@@ -78,7 +78,7 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
 
     /* Ensure that the destination has a valid VTable. */
     if (unlikely(! isValidVTableRoot_fp(newVTable))) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 
 #ifdef CONFIG_ARCH_AARCH32
@@ -101,12 +101,12 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
     asid_map_t asid_map = findMapForASID(asid);
     if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
                  VSPACE_PTR(asid_map_asid_map_vspace_get_vspace_root(asid_map)) != cap_pd)) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
     /* Ensure the vmid is valid. */
     if (unlikely(!asid_map_asid_map_vspace_get_stored_vmid_valid(asid_map))) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
     }
     /* vmids are the tags used instead of hw_asids in hyp mode */
     stored_hw_asid.words[0] = asid_map_asid_map_vspace_get_stored_hw_vmid(asid_map);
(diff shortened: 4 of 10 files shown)


include/arch/arm/arch/32/mode/fastpath/fastpath.h
```diff
@@ -47,7 +47,7 @@ static inline void FORCE_INLINE switchToThread_fp(tcb_t *thread, pde_t *cap_pd,
     if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
         vcpu_switch(thread->tcbArch.tcbVCPU);
     }
-    hw_asid = pde_pde_invalid_get_stored_hw_asid(stored_hw_asid);
+    hw_asid = (hw_asid_t)pde_pde_invalid_get_stored_hw_asid(stored_hw_asid);
```
Indanz, line 50: But almost all callers of `pde_pde_invalid_get_stored_hw_asid` use the return value as `hw_asid_t`, and none cast it. How come those don't cause problems with -Wconversion?

We can add an inline wrapper that does the cast and call that instead. It would be great if the BF generator would support type directly instead of always returning word_t though. That would automatically fix all the proofs too.

include/arch/riscv/arch/fastpath/fastpath.h
```diff
@@ -91,7 +91,7 @@ static inline void fastpath_copy_mrs(word_t length, tcb_t *src, tcb_t *dest)
     /* assuming that length < n_msgRegisters */
     for (i = 0; i < length; i ++) {
         /* assuming that the message registers simply increment */
-        reg = msgRegisters[0] + i;
+        reg = (register_t)(msgRegisters[0] + i);
```
Indanz, line 94: In this case it makes more sense to make `i` register_t too instead of this cast.

include/arch/x86/arch/64/mode/fastpath/fastpath.h
```diff
@@ -103,7 +103,7 @@ static inline void fastpath_copy_mrs(word_t length, tcb_t *src, tcb_t *dest)
     /* assuming that length < n_msgRegisters */
     for (i = 0; i < length; i ++) {
         /* assuming that the message registers simply increment */
-        reg = msgRegisters[0] + i;
+        reg = (register_t)(msgRegisters[0] + i);
```
Indanz, line 106: Same here.

src/fastpath/fastpath.c
```diff
@@ -39,7 +39,7 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
      * saved fault. */
     if (unlikely(fastpath_mi_check(msgInfo) ||
                  fault_type != seL4_Fault_NullFault)) {
-        slowpath(SysCall);
+        slowpath((syscall_t)SysCall);
```
Indanz, line 42: This is silly. We have an enum syscall, and then do a `typedef word_t syscall_t`.

What if we do `typedef enum syscall syscall_t` instead? And do the cast to word_t where it is actually needed?

Probably breaks the proofs all over the place though. But the proofs must have casts everywhere too now, which can't be fun either. It's not only converting from potentially 32-bit to 64-bit, but also from signed to unsigned. What a mess.

Alternatively, just fix up slowpath, if that's the only place this gives warnings?

src/object/untyped.c
```diff
@@ -257,7 +257,7 @@ static exception_t resetUntypedCap(cte_t *srcSlot)
         srcSlot->cap = cap_untyped_cap_set_capFreeIndex(prev_cap, 0);
     } else {
         for (offset = ROUND_DOWN(offset - 1, chunk);
-             offset != - BIT(chunk); offset -= BIT(chunk)) {
+             offset != (word_t)(-(sword_t)BIT(chunk)); offset -= BIT(chunk)) {
             clearMemory(GET_OFFSET_FREE_PTR(regionBase, offset), chunk);
             srcSlot->cap = cap_untyped_cap_set_capFreeIndex(prev_cap, OFFSET_TO_FREE_INDEX(offset));
```
Indanz, line 262: I would add and `word_t end = (word_t)(-(sword_t)BIT(chunk));` variable and use it here.
Then it's also easy to add a comment for this very unusual construction.

Alternatively, you can avoid all casts with `~BIT(chunk) + 1`, at this point it's equally unreadable...

include/arch/arm/arch/32/mode/fastpath/fastpath.h
kent-mcleod: This PR only addresses the wconversion errors in the files that it's updating. Gerwin wanted to minimize the proof impact with each change, so the other callers (eg in vspace.c) changes just aren't in this PR.

include/arch/riscv/arch/fastpath/fastpath.h
kent-mcleod: Not for the proofs.

src/fastpath/fastpath.c
kent-mcleod: (@lsf37).

lsf37: I think we can do that (make `slowpath` take `word_t` instead and cast inside). It might need an update, but there's also a chance that automation will be fine with it.

src/object/untyped.c
```diff
@@ -257,7 +257,7 @@ static exception_t resetUntypedCap(cte_t *srcSlot)
         srcSlot->cap = cap_untyped_cap_set_capFreeIndex(prev_cap, 0);
     } else {
         for (offset = ROUND_DOWN(offset - 1, chunk);
-             offset != - BIT(chunk); offset -= BIT(chunk)) {
+             offset != (word_t)(-(sword_t)BIT(chunk)); offset -= BIT(chunk)) {
```
lsf37: I'm not sure this change is safe and I'm not clear on whether it is actually needed. This is in a verified function, so it is already controlled for overflow (either there is overflow possible and it is safe, or there is no overflow possible). If we cast to `sword_t` we are reducing the word width and are introducing more overflow.

It is possible that both are safe, because `chunk` probably has to be less than max untyped size bits, which is sufficiently below word size that it doesn't matter.

lsf37: Actually, isn't the whole point here that it does underflow and wrap? I'd rather have that explicitly in a wrapping type than have to reason about signed/unsigned conversion for that case.

src/object/untyped.c
lsf37: Neither of them is nice in the proofs, although we can convert `~BIT(chunk) + 1` easily enough if we really need to. The point is that on a wrapping type, underflow with `-` still satisfies almost all of the algebraic laws of `-` so comparing with `- v` very easily gives you exactly one `v` step after count down from `0` and behaves nicely.

Underflow and overflow have defined behaviour on unsigned types. Do we really have to work around them?

lsf37:
If people are happy with my two suggestions above, I can take over this PR and push those after rebase.

src/object/untyped.c
Indanz: The work around is for `-BIT(chunk)`, which is negating an unsigned integer. But if I try doing this myself, the compiler doesn't complain.

@kent-mcleod, is this bit really needed? What compiler did you use?

lsf37: According to the C standard (as I understand it :-)), `- (1ul << n)` should be perfectly fine if `n` fits into the word type. unsigned types are explicitly defined to wrap on over and underflow.

midnightveil: Left shifts greater than the bit count of LHS are always UB, regardless of if the LHS is unsigned.

(not sure if this is what you mean)
(the actual subtraction I would agree)

lsf37: The `n` is fine, we prove that it is less than word_bits. The question is just the `-`.

midnightveil: Yes, then should be fine.

kent-mcleod: Building the configs/AARCH64_verified.cmake configuration:
```
-- Found GCC with prefix aarch64-linux-gnu-
-- The C compiler identification is GNU 12.2.0

```
```
/tmp/tmp.aE9ZuTJUnV/sel4/src/object/untyped.c: In function ‘resetUntypedCap’:
/tmp/tmp.aE9ZuTJUnV/sel4/src/object/untyped.c:261:66: error: conversion to ‘word_t’ {aka ‘long unsigned int’} from ‘int’ may change the sign of the result [-Werror=sign-conversion]
  261 |             clearMemory(GET_OFFSET_FREE_PTR(regionBase, offset), chunk);
      |                                                                  ^~~~~
```

kent-mcleod: So then changing the `int` type for chunk to word_t was the first change.

kent-mcleod: Looks like the `sword_t` cast is too much.

kent-mcleod: Especially since it's not on `offset -= BIT(chunk)` as well.

src/object/tcb.c
```diff
@@ -754,7 +754,7 @@ static exception_t decodeUnsetBreakpoint(cap_t cap, word_t *buffer)
     syscall_error_t error;
 
     tcb = TCB_PTR(cap_thread_cap_get_capTCBPtr(cap));
-    bp_num = getSyscallArg(0, buffer);
+    bp_num = (uint16_t)getSyscallArg(0, buffer);
```
midnightveil, line 757: Is this what we want? It's obviously safe for the kernel but surely this should actually be an error to userspace.

Indanz: Probably not, as the whole point of this is to find potential bugs, not hide them blindly.

midnightveil: Shall we do this as a separate issue/pr then?

Indanz: First step is to push the cast further down the call chain. And then either add a check or combine it with an existing one. Whether we do that here or in a separate PR depends on how big the changes are and how many.

If we do it in separate PRs, we should merge those first before merging this one.

lsf37: As far as possible, this PR should not have proof impact. Adding a check will definitely have proof impact. I'd rather collect these in issues and work through those separately, otherwise it will take very long before anything at all happens.

midnightveil: This is in unverified breakpoints code. But issue or PR is fine by me, just wanted to flag in case we do want to fix it.

PR #1629, opened 2026-03-27
aarch64: avoid clearing reserved SCTLR_EL2 bits

### LINKED ISSUE #1024: aarch64,hyp,boot: SCTLR_EL2 initialization uses reserved bits
The following code could be problematic because it uses register field definitions (In particular, `SA0`, `EE` and `E0E`) that are valid for SCTLR_EL1, but not for SCTLR_EL2:
```c
#define CR_BITS_SET    (CR_ALIGN_SET | \
                        CR_L1_ICACHE_SET | \
                        CR_L1_DCACHE_SET | \
                        BIT(CONTROL_M))

#define CR_BITS_CLEAR  (CR_ALIGN_CLEAR | \
                        CR_L1_ICACHE_CLEAR | \
                        CR_L1_DCACHE_CLEAR | \
                        BIT(CONTROL_SA0) | \
                        BIT(CONTROL_EE) | \
                        BIT(CONTROL_E0E))
```

```c
#define SCTLR   sctlr_el2
#else
#define SCTLR   sctlr_el1
#endif
```

```asm
    mrs     x4, SCTLR
    ldr     x19, =CR_BITS_SET
    ldr     x20, =CR_BITS_CLEAR
    orr     x4, x4, x19
    bic     x4, x4, x20
    msr     SCTLR, x4

```

kent-mcleod:
SA0 (bit 4) and E0E (bit 24) are SCTLR_EL1 fields that are reserved in SCTLR_EL2. Guard these bits so they are only cleared when not building with CONFIG_ARM_HYPERVISOR_SUPPORT.

Fixes: seL4/seL4#1024


Indanz:
Hm, according to [this](https://arm.jonpalmisc.com/latest_sysreg/AArch64-sctlr_el2) they are valid for both SCTLR_EL1 and SCTLR_EL2. And the ARM DDI 0487 pdf confirms this.

kent-mcleod:
> Hm, according to [this](https://arm.jonpalmisc.com/latest_sysreg/AArch64-sctlr_el2) they are valid for both SCTLR_EL1 and SCTLR_EL2. And the ARM DDI 0487 pdf confirms this.

Only if the Effective value of HCR_EL2.{E2H, TGE} is {1, 1}, otherwise they are RES1 which means it should follow Should-Be-One-or-Preserved policy. kernel doesn't currently set HCR_EL2.E2H to 1.

Indanz:
Ah, that's what `ELIsInHost(EL2)` means, very confusing. (I keep forgetting this.)

For SA0 it indeed says it's RES1, but E0E is always defined. It's RES0 or RES1 if only one mode is supported, but clearing it for little-endian is always correct.

Why are we clearing SA0 anyway? That seems quite unusual for AArch64.

So I don't see anything wrong with clearing E0E, but I think we should never clear SA0 and instead always set it.

kent-mcleod:
You're right that E0E is RES0 not RES1. In any case, when booting the SCTLR_EL2, the kernel should be preserving their values which is a defined recommendation for both RES0 and RES1 register fields.

Whether SCTLR_EL1 should be setting SA0 to 0 is a separate issue. (we could change to leave stack checking enabled. One thing I'm not clear on is whether an unaligned SP at EL0 is always wrong, or if trapping unaligned access is just some policy that the CPU provides

kent-mcleod:
We could enable SA0 always, and if there's any issues down the line with user tasks needing unaligned access we can revisit?

Indanz:
> You're right that E0E is RES0 not RES1. In any case, when booting the SCTLR_EL2, the kernel should be preserving their values which is a defined recommendation for both RES0 and RES1 register fields.

But we don't know for sure it is RES0 or RES1, if the endianness is configurable, then we do need to do that here.
And as we expect little-endian, going for that should be always safe, as all our code won't run anyway if we're wrong.
The same is true for EE.

> One thing I'm not clear on is whether an unaligned SP at EL0 is always wrong, or if trapping unaligned access is just some policy that the CPU provides

It seems always wrong for AArch64, I thought it was part of the system ABI. But that's controlled with SCTLR_EL1.SA0, not this one.

So new proposal:
- Remove SA0 from the SCTLR_EL2 set to keep it unchanged.
- Probably want to set SA0 for SCTLR_EL1, but that would change behaviour, so for now keep it as it is. (Can be done in a separate commit if we want.)
- Treat E0E the same as EE, so either clear both, or neither.

Probably should synchronise behaviour of `SCTLR_EL1_NATIVE`, which doesn't set and clear bits, but overwrites everything with what it wants.

lsf37:
This took me a lot of spec reading, but I think I understand now and I agree with Indan's proposal.

If the core does not support little-endian, nothing will work anyway, so it is fine to assume RES0, and to always write 0 for E0E and EE for consistency. I assume things would fail long before we get  here anyway. Is that correct? (If not we could check the value and explicitly error here)

> * Remove SA0 from the SCTLR_EL2 set to keep it unchanged.

Agreed.

> * Probably want to set SA0 for SCTLR_EL1, but that would change behaviour, so for now keep it as it is. (Can be done in a separate commit if we want.)

Agree to not change for now.

> * Treat E0E the same as EE, so either clear both, or neither.

I would say clear both.

> Probably should synchronise behaviour of `SCTLR_EL1_NATIVE`, which doesn't set and clear bits, but overwrites everything with what it wants.

It sounds like we should do that, yes, but I would prefer that in a different PR, because this will have verification impact (very small but not zero -- the changes here we can just merge).

Generally, comments like [A57 default value](https://github.com/seL4/seL4/blob/44b2b946fc51fc6ebdaf126ff72b9e073575dc81/include/arch/arm/armv/armv8-a/64/armv/vcpu.h#L46) are a bit scary. We should double-check what is architecture defined and what is implementation defined and mark that there. The 32-bit `SCTLR_DEFAULT` are even worse, they are just hex values without comment.

Indanz:
> Generally, comments like [A57 default value](https://github.com/seL4/seL4/blob/44b2b946fc51fc6ebdaf126ff72b9e073575dc81/include/arch/arm/armv/armv8-a/64/armv/vcpu.h#L46) are a bit scary. We should double-check what is architecture defined and what is implementation defined and mark that there. The 32-bit `SCTLR_DEFAULT` are even worse, they are just hex values without comment.

Easy enough to check: https://arm.jonpalmisc.com/latest_sysreg/AArch64-sctlr_el1

But most fields are "On a Warm reset, this field resets to an architecturally UNKNOWN value." Furthermore, usually there is some other code running before us, so even if it was, we can't count on it being what we want. A lot of the field are of the form "if feature X supported, then this, otherwise RES0/RES1". On the upside, all fields are defined now, so no future surprises from these register at least.

If we control EL1 we should set all fields to what we want and support, not toggle some bits we know and leave the rest untouched. And same for EL2 and SCTLR_EL2 (which is also fully [defined](https://arm.jonpalmisc.com/latest_sysreg/AArch64-sctlr_el2) now).

PR #1633, opened 2026-03-30
aarch32,vspace: check stale mapping on page flush

kent-mcleod:
A frame cap with stale ASID/virtual address mapping info could be used to flush incorrect memory regions via ARMPageClean_Data, ARMPageInvalidate_Data, ARMPageCleanInvalidate_Data, or ARMPageUnify_Instruction. Add a stale mapping check using resolveVAddr to verify the page table entry still matches the cap's frame size and physical base address before authorising the flush.

This mirrors the AArch64 EL1 fix from PR #1282 for the AArch32 non-hypervisor configuration.

See also #1281

### DIFF
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -2439,6 +2439,22 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+        /* When not in hypervisor mode, we flush via user virtual addresses and
+           need to make sure the mapping info in the cap is not stale. */
+        {
+            resolve_ret_t resolve_ret = resolveVAddr(pd.pd, vaddr);
+            if (unlikely(!resolve_ret.valid ||
+                         resolve_ret.frameSize != generic_frame_cap_get_capFSize(cap) ||
+                         resolve_ret.frameBase != addrFromPPtr((void *)generic_frame_cap_get_capFBasePtr(cap)))) {
+                userError("Page Flush: Attempting to use cap with stale mapping information.");
+                current_syscall_error.type = seL4_InvalidCapability;
+                current_syscall_error.invalidCapNumber = 0;
+                return EXCEPTION_SYSCALL_ERROR;
+            }
+        }
+#endif
+
         start = getSyscallArg(0, buffer);
         end =   getSyscallArg(1, buffer);
 
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -2457,6 +2457,20 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         page_size = 1 << pageBitsForSize(generic_frame_cap_get_capFSize(cap));
         page_base = addrFromPPtr((void *)generic_frame_cap_get_capFBasePtr(cap));
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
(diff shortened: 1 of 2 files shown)


src/arch/arm/32/kernel/vspace.c
```diff
@@ -2439,6 +2439,22 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
...
         }
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+        /* When not in hypervisor mode, we flush via user virtual addresses and
+           need to make sure the mapping info in the cap is not stale. */
+        {
+            resolve_ret_t resolve_ret = resolveVAddr(pd.pd, vaddr);
+            if (unlikely(!resolve_ret.valid ||
+                         resolve_ret.frameSize != generic_frame_cap_get_capFSize(cap) ||
+                         resolve_ret.frameBase != addrFromPPtr((void *)generic_frame_cap_get_capFBasePtr(cap)))) {
```
Indanz: I'd move this check below them to avoid the duplication. Then you can use `page_base` directly. Probably not worth it to avoid calling `generic_frame_cap_get_capFSize` twice though.

src/arch/arm/32/kernel/vspace.c
```diff
@@ -2439,6 +2439,22 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+        /* When not in hypervisor mode, we flush via user virtual addresses and
+           need to make sure the mapping info in the cap is not stale. */
+        {
```
Indanz: No need for this extra block.

Indanz (approved):
Looks good to me, just some nitpicks.

src/arch/arm/32/kernel/vspace.c
lsf37: That would be nicer for the proofs as well. I.e. moving the entire new check below where the local variables `page_size` and `page_base` are set (moving those up would disturb ARM_HYP as well, so moving this down is better).

lsf37 (commented):
This looks good from my side. Will need a small verification update.

lsf37:
I've rebased and added the new errors to the API docs.

src/arch/arm/32/kernel/vspace.c
```diff
@@ -2455,6 +2455,20 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         page_size = 1 << pageBitsForSize(generic_frame_cap_get_capFSize(cap));
         page_base = addrFromPPtr((void *)generic_frame_cap_get_capFBasePtr(cap));
 
+#ifndef CONFIG_ARM_HYPERVISOR_SUPPORT
+        /* When not in hypervisor mode, we flush via user virtual addresses and
+           need to make sure the mapping info in the cap is not stale. */
+        resolve_ret_t resolve_ret = resolveVAddr(pd.pd, vaddr);
+        if (unlikely(!resolve_ret.valid ||
+                     resolve_ret.frameSize != generic_frame_cap_get_capFSize(cap) ||
+                     resolve_ret.frameBase != page_base)) {
```
lsf37, line 2466: I think this might be wrong, `frameBase` as returned from `resolve_vaddr` should be a kernel virtual pointer, not a physical address, but `page_base` at this point already has `addrFromPPtr` applied.

lsf37: no, ignore me, it's correct. `resolve_vaddr` looks up in PTEs/PDEs, not caps.

lsf37:
Verification PR is now up at https://github.com/seL4/l4v/pull/1035. Once that has passed and is approved, we can merge.

PR #1636, opened 2026-04-02
arm,cmake: make verification and release equal

lsf37:
Make verification and release builds equal by setting the default for `KernelArmHypEnableVCPUCP14SaveAndRestore` to `OFF`. The option is still available to set in non-verification builds, only the default changes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,12 +21,19 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 ---
 
-## Upcoming release: BINARY COMPATIBLE
+## Upcoming release: BREAKING
 
 ### Changes
 
+#### ARM
+
+* change default for `KernelArmHypEnableVCPUCP14SaveAndRestore` to `OFF` to provide equal defaults for
+  verification and release builds. This affects AArch32/hyp only.
 
 ### Upgrade Notes
+
+AArch32 with hypervisor mode enabled only: if the option `KernelArmHypEnableVCPUCP14SaveAndRestore` is required, manually set it to `ON` in the build.
+
 ---
 
 ## 15.0.0 2026-03-31: BREAKING
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -97,7 +97,7 @@ config_option(
     context for performance (or other) reasons, we can just turn them off \
     and trap them instead, and have the VCPUs' accesses to CP14 \
     intercepted and delivered to the VM Monitor as fault messages"
-  DEFAULT ON
+  DEFAULT OFF
   DEPENDS "KernelSel4ArchArmHyp;NOT KernelVerificationBuild"
   DEFAULT_DISABLED OFF)
 


lsf37:
Tested in https://github.com/seL4/camkes-vm-examples/pull/96 (which changes no code) and appears to have no effect on our VM examples. Since it always can be switched on per project, I vote for merging this.

See also the discussion in https://github.com/seL4/ci-actions/issues/380 -- for the cases where verification and release configs are equal, it would be enough to test only one of them, which could speed up CI significantly and reduce load on the test hardware.

lsf37:
We might want to have this on by default for DEBUG builds. I currently just have it to OFF, but we could also provide `${KernelDebugBuild}`.

Edit: if we do want it on by default for debug builds, we have to reorder things a bit, because `KernelDebugBuild` only gets set after the file `src/arch/arm/config.cmake` is included. If we move `KernelDebugBuild` and `KernelVerificationBuild` up to before the include, I think it would work.

Probably doesn't matter if `ApplyCommonReleaseVerificationSettings` is used, but I don't think we are using that in the verified configs or are generally guaranteed that it is being used.

CHANGES.md
```diff
@@ -21,12 +21,19 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
 
-## Upcoming release: BINARY COMPATIBLE
+## Upcoming release: BREAKING
 
 ### Changes
 
+#### ARM
+
+* change default for `KernelArmHypEnableVCPUCP14SaveAndRestore` to `OFF` to provide equal defaults for
+  verification and release builds. This affects AArch32/hyp only.
```
Indanz, line 31: Hm, libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h:173 uses `ARM_HYP_ENABLE_VCPU_CP14_SAVE_AND_RESTORE`, but that's only set by `KernelArmHypEnableVCPUCP14SaveAndRestore`, which depends on `KernelSel4ArchArmHyp`, which seems 32-bit only.

Indanz (approved):

Indanz:
> We might want to have this on by default for DEBUG builds. 

I'm against that, because although this is for debugging, it is an entirely different kind of debug than that's normally enabled by DEBUG build. We have:

- Kernel debug build which enables assertions and prints in the kernel. Controlled with `KernelDebugBuild`.
- Enabling user space debug API. Controlled with `HardwareDebugAPI`.
- Enabling VCPUS access to HW debugging. Controlled with `KernelArmHypEnableVCPUCP14SaveAndRestore` for 32-bit Arm.

Let's not conflate those.

lsf37:
> I'm against that, because although this is for debugging, it is an entirely different kind of debug than that's normally enabled by DEBUG build.

I would be happy with that, it's also the less noisy option for users to deal with, because it doesn't change DEBUG configs at all.

CHANGES.md
lsf37: I agree, at least when comparing CMake output, there only is a difference on Arm 32-bit with hyp on. It does say AArch32/hyp in the diff. Did you mean something else in the diff?

Indanz: No, it's just that at fist glance it seems this also affects 64-bit Arm, as the define is used in AArch64 code and in common Arm code. So it gives entirely the wrong impression. My comment was just confirming that from review it seems indeed 32-bit Arm only, even though the code suggests otherwise.

midnightveil: Isn't `ARM_HYP_CP14_SAVE_AND_RESTORE_VCPU_THREADS` used in AArch64 state? And that is set depending on `CONFIG_ARM_HYP_ENABLE_VCPU_CP14_SAVE_AND_RESTORE`?

Which feels like it should be usable for AArch64 hypervisor (see the `debug_conf.h` file).

The fact that it wasn't able to be enabled from the CMake seems like it is a *bug*. So this change note seems correct, but that config should also work for AArch64?

lsf37: We should probably look into that, yes. If so, then the config has never been tested for AArch64, which means we should review properly what the code does when enabled and check that seL4test runs AArch64 tests for it as well.

Indanz: Yes, it does look like it should be set for AArch64 too when debugging is enabled.

@alwin-joshy, did we miss something here when we worked on https://github.com/seL4/seL4/pull/1206?

midnightveil: I copied this into an issue so we can follow up on it: https://github.com/seL4/seL4/issues/1644

PR #1637, opened 2026-04-05
start on python type annotations 

lsf37:
I'm planning to incrementally add python type annotations until most (maybe all) files in the seL4 repo pass pylance strict checking mode.

For now, only 3 smaller files. Apart from adding type, there are the following changes:

- umm.py:
  - change python2 `iteritems()` to python3 `items()`
-  reciprocal.py
   - replace `xrange` by `range` for python3
   - point out that you need 128-bit intermediate for 64-bit division by multiplication
- condition.py
   - add some comments
   - rename `helper` to something more useful
   - add typing assertions where the code makes assumptions about the shape of the XML tree, and document them.
      Apart from these added assertions, there should be no behaviour change in any of the files.

### DIFF
--- a/tools/condition.py
+++ b/tools/condition.py
@@ -3,13 +3,29 @@
 # SPDX-License-Identifier: BSD-2-Clause or GPL-2.0-only
 #
 
-def remove_ws_comments(node):
+# Utility functions for processing condition tags in the seL4 xml files
+
+from __future__ import annotations
+
+from typing import List, Sequence, cast
+from xml.dom.minidom import Element
+
+
+def remove_ws_comments(node: Element) -> None:
+    """Remove whitespace and comment nodes from the children of the given node.
+    Assume being called on a condition tag that only has one child element apart
+    from whitespace and comments."""
     for child in list(node.childNodes):
         if child.nodeType == node.COMMENT_NODE or (child.nodeType == node.TEXT_NODE and len(child.data.strip()) == 0):
             node.removeChild(child)
+    # childNodes are of type _ElementChildren where
+    # _ElementChildren: TypeAlias = Element | ProcessingInstruction | Comment | Text | CDATASection
+    # At most one node of type Element should be left. Length is checked elsewhere, only assert type here.
+    assert all(isinstance(child, Element) for child in node.childNodes)
 
 
-def condition_to_cpp(conditions):
+def condition_to_cpp(conditions: Sequence[Element]) -> str:
+    """Convert the given condition tags list (0 or 1 tags) to a C preprocessor expression."""
     n = len(conditions)
     # Expect zero or one <condition> tag in the conditions list.
     assert n <= 1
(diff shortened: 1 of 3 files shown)


lsf37:
Well, great, stuff is breaking immediately. Switching to draft (will fix)..

lsf37:
Right, it's python versions. I was using 3.10 locally and some of the typing support was only added in 3.10. I'll see if I can make it work with 3.9.

Given that 3.9 has reached end-of-life in Oct last year, we should probably look at that issue we already have open for upgrading the docker containers before I do more files.

Indanz:
@axel-h started doing this years ago in #797, and also relevant #1205. Granted, that PR got polluted by other cleanups as you noted in a comment, but some commits should still be useful.

lsf37:
> @axel-h started doing this years ago in #797, and also relevant #1205. Granted, that PR got polluted by other cleanups as you noted in a comment, but some commits should still be useful.

Yes, I basically want to do these more carefully, incrementally a few files at a time, without any behaviour changes.

I did end up fixing one bug in error reporting in `reciprocal.py` now, though, because I switched from pylance to mypy (which is the one we had decided to use at the TSC meeting), and that flagged a string fmt error.

tools/umm.py
```diff
@@ -5,14 +5,30 @@
 # SPDX-License-Identifier: BSD-2-Clause
 #
 
+from __future__ import annotations
+
 import sys
 from functools import reduce
+from typing import Callable, Dict, IO, Iterable, Iterator, Tuple, TypeVar, Union
+from typing_extensions import TypeAlias
```
midnightveil, line 13: Is this package in sel4-deps already? I don't see it.

midnightveil (commented):
Seems unobjectionable to me, aside from the one comment below.

tools/condition.py
```diff
@@ -23,45 +39,49 @@ def condition_to_cpp(conditions):
...
     if text.startswith(prefix):
         return text[len(prefix):]
     return text
 
 
-def expr_to_bool(expr, values) -> bool:
+def expr_to_bool(expr: Element, values: dict[str, bool]) -> bool:
+    """Evaluate the given condition expression in the provided values environment."""
+    # Cast is safe, because the expr element comes out of remove_ws_comments
+    child_elems: list[Element] = cast(List[Element], list(expr.childNodes))
```
Indanz, line 71: Cast is a no-op in Python, so it's always "safe".

But why add type annotation if you are going to cast anyway? Seems to defeat the purpose a bit.

tools/reciprocal.py
```diff
@@ -13,14 +13,18 @@
 
 # for details on how this script works,
 # see Hacker's delight, Chapter 10, unsigned division.
-from math import floor, ceil
+
+# Note that you need an intermediate value that is 128 bits wide for the
+# multiplication if want to guarantee absence of overflow when you have a 64 bit
+# divisor.
```
Indanz:
```suggestion
# multiplication if want to guarantee absence of overflow when you have a 64 bit
# divisor.
```
```suggestion
# multiplication if want guaranteed absence of overflow when you have a
#  64 bit divisor.
```

Indanz (approved):

tools/umm.py
lsf37: It is not, we should probably add it. We should probably also at some point declare what the minimum supported python version is (we only need this for 3.9 I think, and CI at least now has 3.12 or 3.13).

tools/condition.py
lsf37: I meant safe in the sense of "conservatively correct, they will actually be of that type".

The type of expr.Childnode from just XML is way too general for the consumers of child_elems to type check, so I need to tell it at some point that I know more than what the XML type says. I guess we could do an `isInstance` check instead if we wanted it to be able to fail, but I didn't really want to introduce runtime failure potential. 

(So, basically, the cast is annotation that makes the rest type check)

lsf37: I think I misunderstood the comment earlier. The cast is necessary for the annotation to type check, because the type of `expr.childNodes` is `Node`, not `Any`. If it was just `Any`, the annotation without cast would work, but since it already has a type and we're claiming we know more, it seems to need the cast.

tools/reciprocal.py
Indanz: I spotted the missing "you" initially, but then for got to add it. Maybe better avoid it altogether:

```suggestion
# multiplication for guaranteed absence of overflow when you have a
# 64 bit divisor.
```

lsf37: yes, that's nicer

Issue #1148, opened 2023-12-08
x86 VM + SMP Issue


chrisguikema:
Hello,

A while back, I made this [PR](https://github.com/seL4/seL4/pull/118) that fixed an issue we were having with passthrough interrupts on an x86 SMP VM configuration. We've tried with MCS and the Round Robin scheduler, and the issue is present with both.

Here is a simplified version of our configuration (The actual version gives multiple hardware resources and cores to each VM)

![image](https://github.com/seL4/seL4/assets/21088707/1e55cb59-fefc-4d8d-add8-ffd329296a65)

Each VM is isolated on its own core with a fault handler:

```
Dumping all tcbs on Core 1!
Name                                              State               IP                  Prio  Core
----------------------------------------------------------------------------------------------------------------------
vm0:control                                       blocked on reply    0x40e363              100 1 
vm0:fault_handler                                 blocked on recv     0x40b1cb              255 1 
idle_thread                                       idle                0xffffffff80806781    0   1 
```

We noticed that one of the VMs would begin to experience a time drift. Naturally, I assumed it was a problem with the Time Server, or the HPET emulation layer. However, I noticed that there were a bunch of times when the calls to `init_timer_completed()` would return 0. The time server locks all shared resources around a mutex, so this would only make sense if the Time Server notification was being received multiple times. 

Looking at @kent-mcleod's comments on the closed PR, he mentioned that pending events would be delivered multiple times. This seems to match the behavior that I'm currently seeing. However, when I remove that patch, the VMs print sporadically, and passthrough interrupts don't work at all, which causes our system to break.

![image](https://github.com/seL4/seL4/assets/21088707/ce9598cb-5924-4312-905d-12d432954ab1)

So I began to look into the problem a bit. I'm not an expert in the (Bound) Notification path, so please correct me if I'm wrong.

Specifically for SMP VMs, there are two places in the kernel where the Notification is delivered:

1. When the VMEnter syscall is triggered:

```
    if (syscall == SysVMEnter && NODE_STATE(ksCurThread)->tcbArch.tcbVCPU) {
        vcpu_update_state_sysvmenter(NODE_STATE(ksCurThread)->tcbArch.tcbVCPU);
        if (NODE_STATE(ksCurThread)->tcbBoundNotification
            && notification_ptr_get_state(NODE_STATE(ksCurThread)->tcbBoundNotification) == NtfnState_Active) {
[...]

PR #1641, opened 2026-04-09
x86,SMP: Fix VMCheckBoundNotification IPI handling

Indanz:
When a notification is bound to an IRQ that arrives on a different core than where the VCPU is running, x86 uses a special IPI to notify the other core about this. (For performance reasons you would try to avoid this setup. Nevertheless, it should work.)

When an IpiRemoteCall_VMCheckBoundNotification arrives during a VM exit, the notification reply set by VMCheckBoundNotification() gets overwritten by the handleVmexit() reply, leading to lost notification events. This happens when VMCheckBoundNotification() gets called by the IPI handling code within NODE_LOCK_SYS.

As there is no way to postpone the VM exit handling and the IPI code doesn't know whether it races with a VM exit, doing nothing if the current task is the target is the safest choice:

Either the IPI itself caused a VM exit, or there was a VM exit happening already.

To handle the first case, explicitly call VMCheckBoundNotification() in handleVmexit(). This must be done while holding the kernel lock, as the other core can release the lock any moment after it received our IPI reply.

In the latter case, pending notifications will be detected and returned to user space by the next seL4_VMEnter() call.

Fix tested by Alessandro Legnani.

Resolves issue #1148.


Indanz:
Just in case people wonder: The IPI is always sent, even if the VM is not currently scheduled, because `Arch_leaveVMAsyncTransfer` needs to run on the same CPU.

We do need an IPI to trigger a VM exit, so it's not safe to move the current task check to the IPI sending core.

Indanz:
(I should add a test for this case to sel4test. Edit: Never mind, we don't have VMs in seL4 test yet.)

Furao (approved):
Looks good

lsf37 (approved):

PR #1642, opened 2026-04-10
suspend: move call to setThreadState

michaelmcinerney:
The call to setThreadState within suspend is
moved to the end of the function, in order to
preserve invariants about the queues.


Indanz (approved):
Trivial change.

lsf37 (approved):

PR #1647, opened 2026-04-29
x86/vcpu: expose VMX-preemption timer scale

dreamliner787-9:
Currently, the kernel permits userspace VMMs to enable and use the VMX-preemption timer by allowing writes to the following VMCS fields of a VCPU object:
- `VMX_GUEST_PREEMPTION_TIMER_VALUE` (count)
- `VMX_CONTROL_PIN_EXECUTION_CONTROLS` (timer enable bit)
- `VMX_CONTROL_EXIT_CONTROLS` (save current count on context switch bit)

It also forwards the timer expiry event as a VM exit.

But it does not tell userspace how fast the timer will count down, making it impossible to use. This PR exposes the timer's scale via VCPU MSR read invocation. There are no breaking changes.

For background, the VMX-preemption timer is useful for emulating various x86 timer devices in a VMM. Using the host's physical HPET to back guest timers emulation is inefficient because programming it requires [expensive MMIO accesses](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_real_time/7/html/reference_guide/chap-timestamping#Reading_hardware_clock_sources), and there is no other ubiquitous, high-resolution timer device available on x86. In contrast, the VMX-preemption timer is a dedicated, per-vCPU hardware feature that is cheaply programmed via the VMCS, driven by the TSC, and seamlessly context-switched for each vCPU. Consequently, the VMX-preemption timer provides vastly superior precision and lower virtualisation overhead compared to relying on the shared, high-latency HPET for all of your vCPUs.


src/arch/x86/object/vcpu.c
```diff
@@ -1256,9 +1264,30 @@ BOOT_CODE bool_t vtx_init(void)
...
+    } else {
+        vmx_preemption_timer_scale = 0;
+    }
+
     return true;
 }
 
+BOOT_CODE uint8_t vtx_get_preemption_timer_scale(void)
+{
+    return vmx_preemption_timer_scale;
```
midnightveil: Why bother adding this function if all it does is load a global?

src/arch/x86/object/vcpu.c
```diff
@@ -79,6 +79,10 @@ static uint32_t cr0_low;
 static uint32_t cr4_high;
 static uint32_t cr4_low;
 
+/* Cached value for the VMX-preemption timer scale, the timer will count down
+ * every time bit `vmx_preemption_timer_scale` in the TSC changes due to a TSC increment. */
+static uint8_t vmx_preemption_timer_scale;
```
midnightveil: BOOT_BSS?

src/arch/x86/object/vcpu.c
dreamliner787-9: I felt like extern'ing it in `arch/x86/kernel/boot.c` might be bad.

src/arch/x86/object/vcpu.c
```diff
@@ -1256,9 +1264,30 @@ BOOT_CODE bool_t vtx_init(void)
...
+        /* Intel SDM Combined Volumes
+         * Order Number: 325462-090US February 2026
+         * APPENDIX A VMX CAPABILITY REPORTING FACILITY "MISCELLANEOUS DATA"
+         * > The IA32_VMX_MISC MSR (index 485H) consists of the following fields:
+         * > • Bits 4:0 report a value X that specifies the relationship between
+         * > the rate of the VMX-preemption timer and that of the timestamp counter
+         * > (TSC). Specifically, the VMX-preemption timer (if it is active) counts down
+         * > by 1 every time bit X in the TSC changes due to a TSC increment.
+         */
+        vmx_preemption_timer_scale = vmx_misc_msr_get_preemption_timer_scale(vmx_misc);
```
midnightveil: Can this ever be 0? It seems like it could be. In that case it would mean 1:1 count ratio and be valid whereas the code would treat it as invalid

include/arch/x86/arch/object/structures.bf
```diff
@@ -214,6 +214,11 @@ block vmx_basic_msr {
     field vmcs_revision     31
 }
 
+block vmx_misc_msr {
+    padding                        59
```
midnightveil: Is this actually padding or is it just bits we don't care about?

Can you check if the bit fields allow the padding bits to be nonzero? Otherwise you're stuffing possibly non-zero values into a field that expects to be zero. (And in that case why not just `& 0b1111` tbh)

libsel4/include/sel4/bootinfo_types.h
```diff
@@ -99,13 +99,14 @@ SEL4_COMPILE_ASSERT(
...
-    SEL4_BOOTINFO_HEADER_X86_TSC_FREQ       = 5, /* frequency is in MHz */
-    SEL4_BOOTINFO_HEADER_FDT                = 6, /* device tree */
+    SEL4_BOOTINFO_HEADER_PADDING             = 0,
+    SEL4_BOOTINFO_HEADER_X86_VBE             = 1,
+    SEL4_BOOTINFO_HEADER_X86_MBMMAP          = 2,
+    SEL4_BOOTINFO_HEADER_X86_ACPI_RSDP       = 3,
+    SEL4_BOOTINFO_HEADER_X86_FRAMEBUFFER     = 4,
+    SEL4_BOOTINFO_HEADER_X86_TSC_FREQ        = 5, /* frequency is in MHz */
+    SEL4_BOOTINFO_HEADER_FDT                 = 6, /* device tree */
+    SEL4_BOOTINFO_HEADER_X86_VMX_TIMER_SCALE = 7,
```
midnightveil: Why did this need reformatting? Surely this should be a oneline diff.

if the CI needs it make it a separate commit to update formatting

src/arch/x86/object/vcpu.c
dreamliner787-9: Oh yea that's true, it could be zero.

include/arch/x86/arch/object/structures.bf
dreamliner787-9: It's just bits we don't care about.

libsel4/include/sel4/bootinfo_types.h
dreamliner787-9: Because otherwise it would look like this:
```c
typedef enum {
    SEL4_BOOTINFO_HEADER_PADDING            = 0,
    SEL4_BOOTINFO_HEADER_X86_VBE            = 1,
    SEL4_BOOTINFO_HEADER_X86_MBMMAP         = 2,
    SEL4_BOOTINFO_HEADER_X86_ACPI_RSDP      = 3,
    SEL4_BOOTINFO_HEADER_X86_FRAMEBUFFER    = 4,
    SEL4_BOOTINFO_HEADER_X86_TSC_FREQ       = 5, /* frequency is in MHz */
    SEL4_BOOTINFO_HEADER_FDT                = 6, /* device tree */
    SEL4_BOOTINFO_HEADER_X86_VMX_TIMER_SCALE = 7,
...
```

Indanz (changes requested):
This is not the right approach for exposing virtualisation specific information to VMMs. It's in the wrong place, it's awkward to use and it does not scale.

For one thing, you should expose the whole register at once, not one field at a time. If there are bits that should not be exposed (unlikely), then mask them off.

This information should be exposed via `seL4_X86_VCPU_ReadMSR`.

Indanz (approved):

Indanz:
Currently `seL4_X86_VCPU_ReadMSR` is behind `CONFIG_X86_64_VTX_64BIT_GUESTS`, no idea why personally. But if necessary we can always change that in the future and limit some registers with that config options instead of the whole syscall.

lsf37 (approved):
Fine from my side. It looks like "maintainers can update" is not set, so you'll have to rebase/update on your end before we can merge.

PR #1648, opened 2026-05-04
manual: Reorg riscv32/64 hardware objects

yf13:
This makes rv32/64 section the same level  as those for x86 and arm32/64.


manual/parts/vspace.tex
```diff
@@ -30,6 +30,7 @@ \subsection{Hardware Virtual Memory Objects}
 When referring to it generically, we call this top-level paging structure the VSpace object. The
 seL4 object type that implements the VSpace object is architecture dependent. For instance on
 AArch32, a VSpace is represented by the PageDirectory object and on x64 by a PML4 object.
+On RISC-V the same structure is used for all levels thus the VSpace is a \obj{PageTable} object.
```
Indanz: You want this in the RISC-V 32 bit and 64 bit subsubsections, that's where details should go. (I'd rather remove the previous line than adding another detail that shouldn't be here.)

Indanz (commented):
Well spotted.

An alternative fix is to go one level deeper in subsections, but this seems like a better approach.

Indanz (approved):

manual/parts/vspace.tex
```diff
@@ -125,13 +125,10 @@ \subsubsection{AArch64}
...
 \end{tabularx}
 
-\subsection{RISC-V}
-
-RISC-V provides the same paging structure for all levels, \obj{PageTable}. This means the VSpace
-object is here also implemented by the \obj{PageTable} object.
-
 \subsubsection{RISC-V 32-bit}
 
+On RISC-V the same structure is used for all levels thus VSpace is a \obj{PageTable} object.
```
lsf37: Maybe I suggest active voice instead?


```suggestion
On RISC-V, all levels use the same structure. This means VSpace is a \obj{PageTable} object.
```

manual/parts/vspace.tex
```diff
@@ -143,6 +140,8 @@ \subsubsection{RISC-V 32-bit}
 
 \subsubsection{RISC-V 64-bit}
 
+On RISC-V the same structure is used for all levels thus VSpace is a \obj{PageTable} object.
```
lsf37: (if we change the other one, then also here:)


```suggestion
On RISC-V, all levels use the same structure. This means VSpace is a \obj{PageTable} object.
```

lsf37 (approved):
Indeed, thanks for spotting that. Minor language tweak suggestion only.

PR #1651, opened 2026-05-06
mcs: refactor some functions to ease verification

michaelmcinerney:
- separate the update of the release queue in tcbReleaseEnqueue from the setting of ksReprogramTimer, so that verification may reason uniformly abbout the update of the queue.

- use local variables in sendIPC and receiveIPC to avoid repeated accesses of the heap.

- move the update of the replyTCB field in receiveIPC so that the updates related to the thread state are grouped together, to more closely align with the Haskell specification.

### DIFF
--- a/src/object/endpoint.c
+++ b/src/object/endpoint.c
@@ -93,12 +93,13 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
             schedContext_donate(thread->tcbSchedContext, dest);
         }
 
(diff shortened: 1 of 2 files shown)


Indanz (approved):
If it makes the proofs any easier, feel free to get rid of `ksCurSC`, it can always be reached via `ksCurrentTask` anyway.

lsf37 (approved):

Issue #1650, opened 2026-05-06
compiler issue with gcc-14 and `seL4_Yield`


lsf37:
I need some help with this one. I have a fix for the problem, but I'm not sure it's a good fix and I think this might be a compiler bug.

The issue is that with gcc-14 we're seeing the test SCHED0011 failing for release+MCS+gcc on RISC-V.  It produces

```
Starting test 105: SCHED0011
Cap fault from [SCHED0011] in phase send
PC = 0x37766
CPtr = 0x78e8e2074
Register of root thread in test (may not be the thread that faulted)
Register dump:
pc:	0x37766
ra:	0x37764
sp:	0x10011530
gp:	0x7b108
s0:	0x10011a60
s1:	0x78e8e2074
s2:	0x10011ab0
s3:	0xa
s4:	0x5d75c7f
s5:	0x6146580
s6:	0x10011a60
s7:	0x0
s8:	0x0
s9:	0x0
s10:	0x0
s11:	0x0
a0:	0x78e8e2074
a1:	0x10000000
a2:	0x0
a3:	0x78e8e2074
a4:	0x15
a5:	0x0
a6:	0x0
a7:	0xffffffffffffffff
t0:	0x0
t1:	0x2
t2:	0x10013290
t3:	0x10fc4fd8
t4:	0x3
t5:	0x1
t6:	0x3a
tp:	0x1b5da0
	Failure: result == SUCCESS at line 296 of file /host/projects/sel4test/apps/sel4test-driver/src/testtypes.c
	Error: result == SUCCESS at line 217 of file /host/projects/sel4test/apps/sel4test-driver/src/main.c
```

If I look at address 0x37766 in the image For GCC-14 and for GCC-8, we do in fact have a difference and the GCC-14 code is wrong. There is a loop with the following body:

GCC-14:
```
37760: jal sel4test_timestamp
37764: mv s1,a0
37766: ecall 
```

With GGC-8 it is:
```
37760: jal sel4test_timestamp
37764: mv s1,a0
37766: li a7,-11
37768: ecall
```

`-11` is the syscall number for `Yield`. Inside `sel4test_timestamp` there is a `seL4_Call` which produces a7 = -1 (`seL4_SysCall`). So in the GCC-14 version, the `ecall` at `37766` does another `seL4_SysCall` with junk arguments, hence the cap fault (also shows a7=-1).

[...]

PR #1654, opened 2026-05-07
libsel4/riscv: avoid gcc 14.2 miscompilation

lsf37:
a7 is not changed by seL4_Yield, so the original declaration is correct.

However, in some loops GCC 14.2 with -O3 may drop the load to a7 if a7 is not declared as clobbered in the assembly block. This is a problem if *other* code does write to a7. For some reason GCC 14.2 does not recognise those other loads. GCC 14.3 and GCC 15 both work as expected.

The problem manifests in SCHED0011 in sel4test.

This change works around the GCC 14.2 problem because GCC 14.2 is the standard Debian trixie compiler and it is likely that people will hit the problem even if we say that GCC 14.2 should not be used. The workaround puts the load into the assembly block and declares it as being clobbered.

(Edit: workaround changed during the discussion below)

See also the discussion in #1650

I'm planning to update the default compiler in docker to something else, so that we're using a more reliable toolchain in general, but we still need to work around the issue for gcc 14.2.

### DIFF
--- a/libsel4/arch_include/riscv/sel4/arch/syscalls.h
+++ b/libsel4/arch_include/riscv/sel4/arch/syscalls.h
@@ -687,8 +687,8 @@ LIBSEL4_INLINE_FUNC seL4_MessageInfo_t seL4_NBSendWaitWithMRs(seL4_CPtr dest, se
 
 LIBSEL4_INLINE_FUNC void seL4_Yield(void)
 {
-    register seL4_Word scno asm("a7") = seL4_SysYield;
-    asm volatile("ecall" :: "r"(scno));
+    riscv_sys_null(seL4_SysYield);
+    asm volatile("" ::: "memory");
 }
 
 #ifdef CONFIG_KERNEL_MCS


lsf37:
We may want to do the same for `riscv_sys_null`, it is likely to have the same issue (will let the tests finish first).

midnightveil (approved):

Indanz (approved):

libsel4/arch_include/riscv/sel4/arch/syscalls.h
```diff
@@ -687,8 +687,13 @@ LIBSEL4_INLINE_FUNC seL4_MessageInfo_t seL4_NBSendWaitWithMRs(seL4_CPtr dest, se
 
 LIBSEL4_INLINE_FUNC void seL4_Yield(void)
 {
-    register seL4_Word scno asm("a7") = seL4_SysYield;
-    asm volatile("ecall" :: "r"(scno));
+    /* Not written as "register .. asm .." to work around miscompilaton in gcc 14.2.
```
Indanz:
```suggestion
    /* Not written as "register .. asm .." to work around miscompilation in gcc 14.2.
```

libsel4/arch_include/riscv/sel4/arch/syscalls.h
```diff
@@ -687,8 +688,13 @@ LIBSEL4_INLINE_FUNC seL4_MessageInfo_t seL4_NBSendWaitWithMRs(seL4_CPtr dest, se
 
 LIBSEL4_INLINE_FUNC void seL4_Yield(void)
 {
-    register seL4_Word scno asm("a7") = seL4_SysYield;
-    asm volatile("ecall" :: "r"(scno));
+    /* Not written as "register .. asm .." to work around miscompilation in gcc 14.2.
```
axel-h: Could make this call `riscv_sys_null(seL4_SysYield)` now to remove the duplication. This is also what the arm code does. Plus there is a `asm volatile("" ::: "memory");` barrier there, which also seems not a bad thing to state for a general yield operation.

lsf37: I do like the idea of `riscv_sys_null(seL4_SysYield)` to avoid duplication. I've not added the memory clobber declaration for now, because I got differing input for that and it's not directly related to the issue I'm trying to fix.

In general, I'd agree, though. Even if seL4 itself doesn't change memory in this call, other threads explicitly can, even on a uniprocessor system. We wouldn't want other volatile memory accesses be moved over the yield. On the other hand, `yield` is not really a synchronisation point and wouldn't make sense to be used as one in a multicore system, so in the end it probably doesn't really matter.

lsf37: > On the other hand, yield is not really a synchronisation point and wouldn't make sense to be used as one in a multicore system, so in the end it probably doesn't really matter.

Ok, I've changed my mind. Arm and x86 both do this, and not doing it on RISC-V is just inconsistent. I'll change it to be consistent.

lsf37:
Going off Axel's feedback I've changed the workaround: the `riscv_sys_null` call is unchanged and the `seL4_Yield` call is now the same as it is on Arm and x86 (inline call to `sys_null` with and additional memory clobber declaration).

This fixes the gcc 14.2 misbehaviour and leads to the same assembly code as before. If we change `riscv_sys_null` to include a `mv` instruction (for the `sys` parameter), the code changes, and we would then get two instructions to load the syscall number into a7, which is not nice.

I'll leave this sitting until tomorrow so @Indanz and @midnightveil who have already approved have a chance to take another look.

midnightveil:
Looks like the hardware failures seem to be related to conserver, at the 9:57 mark in 2 of the runs I see that the conserver appeared to drop off until 10:59 or so. And the other was a boot failure.

midnightveil (approved):
Should we not add a comment as to why we are adding the memory barrier? Or maybe it should be part of `sys_null`? 

I still don't understand why `::: "memory"` is needed; and it does just feel like a GCC bug still.

However, in the interest of fixing our CI...

lsf37:
> Should we not add a comment as to why we are adding the memory barrier? Or maybe it should be part of `sys_null`?

It would be strange to add a comment here, but not in the other architectures, given that they now all have the same implementation of seL4_Yield. I'll merge this to unblock CI, but I'll try out what happens on other architectures if I remove it there. It might produce the same issue, actually.

 
> I still don't understand why `::: "memory"` is needed; and it does just feel like a GCC bug still.

I'm close to certain that this is a GCC bug (esp since 14.3 no longer does this).

It's more a question of generally making a true statement to the compiler, which is that after `ecall` returns, the memory may have changed. To the compiler it doesn't matter whether the kernel might have changed it or another thread, just that it may have changed after the instruction. Of course memory can change pretty much at any point when other threads are running, but people might be using Yield explicitly to give other threads the chance to change memory, so I do think it's reasonably to add here.

It shouldn't be necessary to do this for this specific issue, it just happens to remove the trigger for it.

midnightveil:
> It would be strange to add a comment here, but not in the other architectures, given that they now all have the same implementation of seL4_Yield. I'll merge this to unblock CI, but I'll try out what happens on other architectures if I remove it there. It might produce the same issue, actually.

Hopefully. Though I guess it could be a RISC-V specific bug. 
(But yes, I meant for all architectures).

> It's more a question of generally making a true statement to the compiler, which is that after ecall returns, the memory may have changed. To the compiler it doesn't matter whether the kernel might have changed it or another thread, just that it may have changed after the instruction. Of course memory can change pretty much at any point when other threads are running, but people might be using Yield explicitly to give other threads the chance to change memory, so I do think it's reasonably to add here.

That's true.

Indanz (commented):
The new code doesn't make sense to me. `riscv_sys_null` does exactly the same as the old code does, so would trigger the same bug as before, if not for the memory barrier added.

But the memory barrier makes no conceptual sense, and it is a much bigger hammer than marking a register as clobbered, as it forces the compiler to reload memory reads it did before the yield, so it will cause extra memory accesses and instructions. In other words, it can make code that runs after the yield returns slower for no reason.

The only thing it has going for it is consistency with other archs.

The other caller of `riscv_sys_null` is `seL4_BenchmarkNullSyscall`. High chance that that's broken with GCC 14.2 too.

Indanz:
> Hopefully. Though I guess it could be a RISC-V specific bug. (But yes, I meant for all architectures).

The O3 miscompilations in GCC 14.2 that were fixed in 14.3 are RISC-V only.

> > It's more a question of generally making a true statement to the compiler, which is that after ecall returns, the memory may have changed. To the compiler it doesn't matter whether the kernel might have changed it or another thread, just that it may have changed after the instruction. Of course memory can change pretty much at any point when other threads are running, but people might be using Yield explicitly to give other threads the chance to change memory, so I do think it's reasonably to add here.

The problem is that you do need some kind of synchronisation or way to reload the memory reads, other than seL4_Yield. Now seL4_Yield may hide programming bugs. And adding a memory clobber isn't free because it does forces the compiler to generate instructions to reload all variables loaded from memory.

axel-h:
You have point, that the current change does not really look like it solve the issue. However, since we don't really know what the compiler bug was, it's hard to argue what would be a proper fix - it's basically anything that works. I tried to look though the GCC changelog, but so far could not find anything. There is something in that direction mentioned for x86, where the optimizer is too aggressive in loops. What I also found are discussions that indicate there is a wide concensus that things like
```
register seL4_Word scno asm("a7") = seL4_SysYield;
asm volatile("ecall" :: "r"(scno));
```
are the proper way to do syscalls and that is the only thing where the `register ... asm("a7") ...` variables are guaranteed to work as expected. Seems this has even been clarified in the GCC docs some time ago.

For the memory barrier, I agree that this has a downside for performance. But it's good practice for a library to be conservative and align the behavior to what happens on other architectures. Also, I think there are unfortunately many real world use cases where yield is used (wrongly, sloppily ...) for some polling or synchronization, so the barrier helps. For other use cases, I wonder if performance for yielding is really a concern. If somebody is really concerned, the lib does not have to be used, but the call could be implemented directly.

Indanz:
> I tried to look though the GCC changelog, but so far could not find anything. There is something in that direction mentioned for x86, where the optimizer is too aggressive in loops.

The list for fixes in gcc 14.3 is: https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=14.3

Relevant for RISC-V and O3 that I found:
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117594
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118950
- https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119114

> are the proper way to do syscalls and that is the only thing where the `register ... asm("a7") ...` variables are guaranteed to work as expected. Seems this has even been clarified in the GCC docs some time ago.

Yes, and when it does not work because of compiler bug, the minimal change would have been to mark a7 as clobbered, not adding a separate asm statement that adds a memory clobber. If we do want a memory clobber, it should have been part of the syscall asm at least.

The chance of memory clobber fixing it by accident is much higher than for marking a7 clobbered. Which means that on slightly different code the original bug can pop up again, as memory clobber in theory doesn't affect register allocation and usage directly, but only indirectly.

So the amount of pot luck of it working with memory clobber feels a lot higher than for a direct a7 clobber.

> For other use cases, I wonder if performance for yielding is really a concern. If somebody is really concerned, the lib does not have to be used, but the call could be implemented directly.

The performance price paid of the memory clobber is after yield returns, not before. If you yielded to get deterministic behaviour, you can't use it like that any more because all memory accesses will be done after the yield, introducing jitter you may have wanted to avoid. So this kind of argument can go either way, as it is always use case specific.

lsf37:
I agree that this is all still suspect, but it doesn't change the fact that if we decide that memory clobber is not the thing to do, then we need to do it everywhere not just locally in RISC-V that yet again happened to do its own thing for no reason. I'd rather have it consistent first and then decide properly what should be done in general.

We can still change it, I'm not opposed to investigating further. For now we can at least still merge PRs again.

For what it's worth, the memory clobber does not seem to lead to GCC spilling all variables. It produces exactly the same sequence

```
jal sel4test_timestamp
mv s1,a0
li a7,-11
ecall
```

as GCC 8 did before. The variable that stores the result of sel4test_timestamp does not get spilled to memory before the `ecall`. (This is not the whole loop body, just the bit that varied between compilers, it goes on to do a second `sel4test_timestamp` and subtract etc, all in registers).

PR #1655, opened 2026-05-08
mcs: idle threads should have maximum period

midnightveil:
mcs: idle threads should have maximum period

Setting this to the boot thread time slice makes no sense, as the only case where this would matter not being the maximum is if round-robin is occurring. This just means when you're in the idle thread you're constantly getting interrupts every 5ms, for no reason at all. This would mess with any power savings you'd see from an idle core in `wfi`, for example.

I'd argue that BOOT_THREAD_TIME_SLICE sort of doesn't matter either for MCS (and it should be instead `getMaxTicksToUs()` as well); userland can configure this if it wants instead of the kernel parameter existing. But I suppose it doesn't matter that much and might break existing userspace.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -484,7 +484,7 @@ BOOT_CODE void create_idle_thread(void)
         SMP_COND_STATEMENT(NODE_STATE_ON_CORE(ksIdleThread, i)->tcbAffinity = i);
 #ifdef CONFIG_KERNEL_MCS
         configure_sched_context(NODE_STATE_ON_CORE(ksIdleThread, i), SC_PTR(&ksIdleThreadSC[SMP_TERNARY(i, 0)]),
-                                usToTicks(CONFIG_BOOT_THREAD_TIME_SLICE * US_IN_MS));
+                                MAX_PERIOD_TICKS);
         SMP_COND_STATEMENT(NODE_STATE_ON_CORE(ksIdleThread, i)->tcbSchedContext->scCore = i;)
         NODE_STATE_ON_CORE(ksIdleSC, i) = SC_PTR(&ksIdleThreadSC[SMP_TERNARY(i, 0)]);
 #endif
--- a/include/kernel/sporadic.h
+++ b/include/kernel/sporadic.h
@@ -46,7 +46,8 @@
  * than 2^62. */
 #define MAX_PERIOD_US (getMaxUsToTicks() / 8)
 #endif /* CONFIG_KERNEL_STATIC_MAX_PERIOD_US != 0 */
-#define MAX_RELEASE_TIME (UINT64_MAX - 5 * usToTicks(MAX_PERIOD_US))
+#define MAX_PERIOD_TICKS (usToTicks(MAX_PERIOD_US))
+#define MAX_RELEASE_TIME (UINT64_MAX - 5 * MAX_PERIOD_TICKS)
 
 /* Short hand for accessing refill queue items */
 static inline refill_t *refill_index(sched_context_t *sc, word_t index)


Indanz:
Agreed in principle that the idle thread could have maximum time slice.

However, in practice people can configure user space tasks with the same priority as the idle task, and then round-robin does happen.

midnightveil:
> However, in practice people can configure user space tasks with the same priority as the idle task, and then round-robin does happen.

Idle-thread never appears in the scheduler queues, and only gets run when no other tasks are schedulable. So whilst it is labelled with priority 255 it's more like 256.

Indanz (commented):
`getMaxUsToTicks` returns `time_t`, not ticks. It gives the maximum value you can pass to `UsToTicks`. The input here is the value in ticks.

`UsToTicks(getMaxUsToTicks())` would be correct, but also silly. You actually want just `UINT64_MAX`, or even `UINT64_MAX / 8` for consistency with `MAX_PERIOD_US`.

Edit: There is also `getMaxTicksToUs`.

Indanz:
> Idle-thread never appears in the scheduler queues, and only gets run when no other tasks are schedulable. So whilst it is labelled with priority 255 it's more like 256.

Ah, good point. Yes, in that case, it's silly.

Indanz:
> I'd argue that BOOT_THREAD_TIME_SLICE sort of doesn't matter either for MCS (and it should be instead `getMaxUsToTicks()` as well); userland can configure this if it wants instead of the kernel parameter existing. But I suppose it doesn't matter that much and might break existing userspace.

Agreed with not changing it now as it may break existing user space. But we can change the behaviour when we change the MCS ABI from us to ticks instead.

midnightveil:
Fixed to use `getMaxTicksToUs()` instead of `getMaxUsToTicks()`.

src/kernel/boot.c
```diff
@@ -533,7 +533,7 @@ BOOT_CODE tcb_t *create_initial_thread(cap_t root_cnode_cap, cap_t it_pd_cap, vp
 
     /* initialise TCB */
 #ifdef CONFIG_KERNEL_MCS
-    configure_sched_context(tcb, SC_PTR(rootserver.sc), usToTicks(CONFIG_BOOT_THREAD_TIME_SLICE * US_IN_MS));
+    configure_sched_context(tcb, SC_PTR(rootserver.sc), getMaxTicksToUs());
```
Indanz: You don't want this. If you do, you should remove the CONFIG_BOOT_THREAD_TIME_SLICE config option altogether. But that would require an RFC as that can break a lot of existing user space, so I don't think it's worth it on its own.

midnightveil: I'm not sure how that happened. I even mention in the PR description that I wasn't going to change it.

Indanz: That's life in a nutshell.

midnightveil: (Fixed)

Indanz (approved):

midnightveil:
From the test failures:

```
  seL4 failed assertion 'next_interrupt >= NODE_STATE(ksCurTime)' at /github/workspace/kernel/src/kernel/thread.c:623 in function setNextInterrupt
  ```
  
I think this means this should probably be using `usToTicks(MAX_PERIOD_US)` or something. 

(This is not planned for the next release)

**Comment by midnightveil** [comment]
Huh, from the rt branch of l4v, this is already supposed to be the case. Probably should check if it's the case for the refills, too. From the looks of things `configure_sched_context -> refill_new` sets `rTime` to `ksCurTime`...

https://github.com/seL4/l4v/blob/8b61de771de8d576bd3dce327e9cb692b41875a8/spec/abstract/AARCH64/Init_A.thy#L89-L94

```thy
     idle_sc_ptr \<mapsto> SchedContext (default_sched_context
                                   \<lparr>sc_tcb := Some idle_thread_ptr,
                                    sc_refills := [\<lparr>r_time = 0, r_amount = MAX_PERIOD\<rparr>,
                                                   \<lparr>r_time = MAX_PERIOD, r_amount = 0\<rparr>],
                                    sc_budget := MAX_PERIOD,
                                    sc_refill_max := MIN_REFILLS\<rparr>) 0,
                                 
```

(`definition "MAX_PERIOD = us_to_ticks MAX_PERIOD_US"`)

lsf37:
> Huh, from the rt branch of l4v, this is already supposed to be the case. Probably should check if it's the case for the refills, too. From the looks of things `configure_sched_context -> refill_new` sets `rTime` to `ksCurTime`...
> 
> https://github.com/seL4/l4v/blob/8b61de771de8d576bd3dce327e9cb692b41875a8/spec/abstract/AARCH64/Init_A.thy#L89-L94

Init_A is not a real state, it's just an example to make sure the invariant predicates aren't vacuous, but it doesn't cover all of them. In particular I don't think they have scheduling predicates.

(Edit: although we should really change that and make sure we got everything. @michaelmcinerney could you have a look at what we're missing?)

michaelmcinerney (approved):

michaelmcinerney:
> > Huh, from the rt branch of l4v, this is already supposed to be the case. Probably should check if it's the case for the refills, too. From the looks of things `configure_sched_context -> refill_new` sets `rTime` to `ksCurTime`...
> > https://github.com/seL4/l4v/blob/8b61de771de8d576bd3dce327e9cb692b41875a8/spec/abstract/AARCH64/Init_A.thy#L89-L94
> 
> Init_A is not a real state, it's just an example to make sure the invariant predicates aren't vacuous, but it doesn't cover all of them. In particular I don't think they have scheduling predicates.
> 
> (Edit: although we should really change that and make sure we got everything. @michaelmcinerney could you have a look at what we're missing?)

We do show that the initial state does satisfy `valid_sched`, as well as many of the other smaller invariants we have that are true at kernel entry and exit. The proofs begin here

https://github.com/seL4/l4v/blob/rt/proof/refine/RISCV64/Refine.thy#L400

midnightveil:
> Init_A is not a real state, it's just an example to make sure the invariant predicates aren't vacuous, but it doesn't cover all of them. In particular I don't think they have scheduling predicates.

OK, my mistake then. I was just curious about whether anything in the proofs care about the configuration of the idle thread precisely.

It looks like Michael is happy with this. Should I do a proof-test before merging?

midnightveil:
Proofs succeeded here: https://github.com/seL4/seL4/actions/runs/30313075751
hw-tests succeeded.

PR #1661, opened 2026-05-14
debug: add get thread affinity syscall

midnightveil:
As per title. 

Useful for https://github.com/seL4/sel4test/pull/155.

### DIFF
--- a/libsel4/arch_include/arm/sel4/arch/syscalls.h
+++ b/libsel4/arch_include/arm/sel4/arch/syscalls.h
@@ -645,6 +645,21 @@ LIBSEL4_INLINE_FUNC void seL4_DebugNameThread(seL4_CPtr tcb, const char *name)
 }
 
 #if CONFIG_ENABLE_SMP_SUPPORT
+LIBSEL4_INLINE_FUNC seL4_Word seL4_DebugGetThreadAffinity(seL4_CPtr tcb)
+{
+    seL4_Word unused0 = 0;
+    seL4_Word unused1 = 0;
+    seL4_Word unused2 = 0;
+    seL4_Word unused3 = 0;
+    seL4_Word unused4 = 0;
+
+    seL4_Word affinity = 0;
+
+    arm_sys_send_recv(seL4_SysDebugGetThreadAffinity, tcb, &affinity, 0,
+                      &unused0, &unused1, &unused2, &unused3, &unused4, 0);
+    return affinity;
+}
+
 LIBSEL4_INLINE_FUNC void seL4_DebugSendIPI(seL4_Uint8 target, unsigned irq)
 {
     arm_sys_send(seL4_SysDebugSendIPI, target, irq, 0, 0, 0, 0);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -28,6 +28,9 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 ### Changes
 
+* Add `seL4_DebugGetThreadAffinity()` syscall to return the current affinity of a thread
+  in SMP configurations.
+
 ### Platforms
 
 * Add support for the stm32mp2 SoC family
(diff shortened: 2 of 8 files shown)


lsf37:
Adding a syscall does require an RFC, even if it is just a small one like this.

midnightveil:
> Adding a syscall does require an RFC, even if it is just a small one like this.

Even if it's just debug-only, and not a "proper" release-mode syscall?

lsf37:
> > Adding a syscall does require an RFC, even if it is just a small one like this.
> 
> Even if it's just debug-only, and not a "proper" release-mode syscall?

I guess debug-only is a gray area. Strictly speaking, yes (imagine someone wanting to add a large new debug feature).

It also seems pretty obvious to me that this one here is trivial and is clearly useful debug info, so maybe we can skip the overhead if nobody in the @seL4/tsc has any objections?

libsel4/arch_include/arm/sel4/arch/syscalls.h
```diff
@@ -645,6 +645,21 @@ LIBSEL4_INLINE_FUNC void seL4_DebugNameThread(seL4_CPtr tcb, const char *name)
...
+    seL4_Word unused0 = 0;
+    seL4_Word unused1 = 0;
+    seL4_Word unused2 = 0;
+    seL4_Word unused3 = 0;
+    seL4_Word unused4 = 0;
+
+    seL4_Word affinity = 0;
+
+    arm_sys_send_recv(seL4_SysDebugGetThreadAffinity, tcb, &affinity, 0,
+                      &unused0, &unused1, &unused2, &unused3, &unused4, 0);
```
Indanz, line 659: Why not just pass one `&unused` for all parameters?

Indanz:
For really simple ones like this, the RFC process seems a bit heavy handed and would waste a lot of time for everyone, but currently it's the only procedure we have. Removing existing debug features would require an RFC, so for symmetry it makes sense that one would be needed to add any new ones. But using RFCs for simple things like this instead of normal PRs seems inefficient.

Can't we add an RFC tag for PRs like this? Any TSC member can request a full RFC at any time anyway if they disagree with the fast track mini-RFC approach. To make sure all TSC members are aware of the PR, we can request a review from seL4/tsc. Would that work for everyone?

Indanz (approved):

libsel4/arch_include/arm/sel4/arch/syscalls.h
axel-h: Seems this follows the pattern used everywhere else. I'd leave such a change for a follow-up PR then and do this everywhere.

lsf37:
> Can't we add an RFC tag for PRs like this? Any TSC member can request a full RFC at any time anyway if they disagree with the fast track mini-RFC approach. To make sure all TSC members are aware of the PR, we can request a review from seL4/tsc. Would that work for everyone?

I like that approach. We can discuss more and make it official at the next TSC meeting. For now I'll leave this open for at least a week so people can request more time or object if they want to.

Overall, the RFC process is also for making sure the rest of the community has a chance to comment (not only the TSC), but I think for very small ones like this a label would suffice.

midnightveil:
> I like that approach. We can discuss more and make it official at the next TSC meeting. For now I'll leave this open for at least a week so people can request more time or object if they want to.

Have we heard anything back?

lsf37:
Ok, nobody voiced any concerns. Let's go ahead.

lsf37:
Could you add a CHANGELOG entry, please? It is picked up correctly for the API reference in the manual, so that part is fine.

lsf37 (approved):

midnightveil:
> Could you add a CHANGELOG entry, please? It is picked up correctly for the API reference in the manual, so that part is fine.

Done. Is that enough information?

lsf37:
> > Could you add a CHANGELOG entry, please? It is picked up correctly for the API reference in the manual, so that part is fine.
> 
> Done. Is that enough information?

Yes, perfect. Happy for you to merge.

midnightveil:
Well, that broke the build. Looks like x86 SMP wasn't in the build set and I had an off-by-1 typo.

Indanz:
> Well, that broke the build. Looks like x86 SMP wasn't in the build set and I had an off-by-1 typo.

Apologies, I should have caught that when reviewing.

Opened PR #1694 to fix this.

Issue #138, opened 2019-07-11
MCS - TCB Affinity Modification Bug


chrisguikema:
I've noticed an issue on the MCS branch with changing a threads core by invoking the `seL4_SchedControl_Configure` system call. Essentially, if a thread is running on the new core, the TCB will migrate properly. However, if no thread is active on the new core, the TCB migrates, but a scheduler dump shows the TCB as "running", but its instruction pointer is the `syscall` instruction, and everything stalls out. This behavior is also seen with passive servers.

For example, the camkes-vm can be configured such that the Serial Server provides passive interfaces for `putchar` and `getchar`. With a single VM, this functions as expected. However, with multiple VMs on different cores, the same behavior is encountered. The passive server is shown to be "Running" at the `syscall` instruction. This obviously causes issues, because the calling thread has donated its scheduling context, and can no longer run.

By reverting the changes in this commit, https://github.com/seL4/seL4/commit/ef47fa6e3c842112bbe8fbfc0783af9f78088a4c, I can get the passive servers to run on QEMU, but not on real hardware (Optiplex, i7). 

Do you have any insight on where I can look or what is happening? I can attempt to spend some time and fix the issue.

chrisguikema:
As an update, it appears that the system behaves differently on QEMU v. Hardware even when not functioning.

This is a (slightly modified) dump scheduler state of the system on QEMU:
```
Dumping all tcbs on Core 1!
Name                                              State               IP                  Prio 	Core 	HasSchedContext InReleaseQueue
--------------------------------------------------------------------------------------------------------------------------------------
serial:processed_putchar                running             0x406778            100	1	1		0
vm0:control                                    blocked on reply    0x41b19a       100	1	0		0
```
So at this point on QEMU, the context has been donated, but is stuck.

And on hardware:
```
Dumping all tcbs on Core 0!
--------------------------------------------------------------------------------------------------------------------------------------
serial:processed_putchar                          running             0x406778            100   0       0               0

Dumping all tcbs on Core 1!
Name                                              State               IP                  Prio  Core    HasSchedContext InReleaseQueue
--------------------------------------------------------------------------------------------------------------------------------------
vm0:control                                       blocked on reply    0x41b19a            100   1       1               0
```

[...]

PR #1662, opened 2026-05-14
SMP: fix MCS scheduling bugs with passive servers and core migration

### LINKED ISSUE #941: MCS+SMP: Missing remoteQueueUpdate somewhere?

midnightveil:
An alternate to #986 and fixes #941.

Test with https://github.com/seL4/sel4test/pull/155.

Merge after: https://github.com/seL4/seL4/pull/1661.

---

Some notes:

1. I can't tell if `NODE_STATE(ksCurSC)->scCore == getCurrentCPUIndex()` should also be
an invariant. The current SC only seems to be updated at the end of schedule,
unlike ksCurThread which is updated in a lot of places; but there is a fair amount
of code which looks at `NODE_STATE(ksCurSC)` throughout the kernel that I can't tell
if it would be correct.

2. The idea behind this is that `migrateTCB` is always called to set the `tcbAffinity`, hence it is the most appropriate place to also cause IPIs if necessary. It's possible it some cases it might be less performant than necessary (but only in terms of local operations, not excess IPIs) if it turns out the function that calls `migrateTCB` does some work that causes nothing to happen. However, considering SMP is unverified this seems acceptable because of how complex the scheduling is.

-----

There's quite a few commits in here, some of which need to be cleaned up. Broadly included:

1. The contents of #1661.
2. Some improvements to `debug_dumpScheduler` that makes it print out information of remote cores (this should be a separate PR).
3. Addition of an 'Invariant: the current thread always belongs to the current core.' throughout the code. See the relevant commit messages for how I determined the placement of these asserts; strictly speaking the most important one is in `schedule()` but if that goes off it can be hard to tell where the issue happened.
4. Addition of a 'if a thread, ksSchedulerAction belongs to the current core' invariant.
5. Because of the invariants, omit some remoteQueueUpdate calls when we know a thread is already on the current core
6. Moving the `isSchedulable()` check on `possibleSwitchTo` to be a pre-condition, since the code varies a lot whether it considers this to be the case or not, and personally I like it more as a precondition. As part of this commit I also added a variety of `TODO` or `XXX` comments where I don't quite understand details of the MCS scheduler, specifically in that `assert()` sometimes checks `isSchedulable` but sometimes is `refill_ready` and `refill_sufficient` and sometimes other parts. I don't know if these are exactly equivalent, as sometimes I made changes or rearranged these, but (a) I'd like to know why and (b) could it be possible to test this?
7. Disable the signal fastpath for the migration of the current core.

---

I don't think I discovered any other issues other than the previously identified case in #941 regarding SMP+MCS.

(Related: https://github.com/seL4/seL4/issues/138)

### DIFF
--- a/include/model/statedata.h
+++ b/include/model/statedata.h
@@ -120,6 +120,10 @@ extern paddr_t ksUserLogBuffer;
 #define SchedulerAction_ResumeCurrentThread ((tcb_t*)0)
 #define SchedulerAction_ChooseNewThread ((tcb_t*) 1)
 
+#define SchedulerAction_IsCandidateThread(schedulerAction) \
+    ((schedulerAction) != SchedulerAction_ResumeCurrentThread && \
+     (schedulerAction) != SchedulerAction_ChooseNewThread)
+
 #define MODE_NODE_STATE(_state)    MODE_NODE_STATE_ON_CORE(_state, getCurrentCPUIndex())
 #define ARCH_NODE_STATE(_state)    ARCH_NODE_STATE_ON_CORE(_state, getCurrentCPUIndex())
 #define NODE_STATE(_state)         NODE_STATE_ON_CORE(_state, getCurrentCPUIndex())
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -670,6 +670,15 @@ void NORETURN fastpath_signal(word_t cptr, word_t msgInfo)
         crossnode = true;
     }
 
+#if defined(ENABLE_SMP_SUPPORT) || defined(CONFIG_KERNEL_MCS)
+    /* If we would cause our current thread to be migrated between cores,
+     * or between domains, go to the slowpath, as this would result in
+     * scheduling on our core/domain. */
+    if (crossnode && dest == NODE_STATE(ksCurThread)) {
+        slowpath(SysSend);
+    }
+#endif
+
     /*  Point of no return */
 #ifdef CONFIG_BENCHMARK_TRACK_KERNEL_ENTRIES
     ksKernelEntry.is_fastpath = true;
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -465,7 +465,7 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
     } else {
 #ifdef CONFIG_KERNEL_MCS
         /* Update queue. */
-        tcb_queue_t queue = tcbEPAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
+        tcb_queue_t queue = tcbAppend(NODE_STATE(ksCurThread), ep_ptr_get_queue(ep_ptr));
         endpoint_ptr_set_epQueue_head_np(ep_ptr, TCB_REF(queue.head));
         endpoint_ptr_mset_epQueue_tail_state(ep_ptr, TCB_REF(queue.end), EPState_Recv);
 #else
@@ -689,7 +689,7 @@ void NORETURN fastpath_signal(word_t cptr, word_t msgInfo)
         cancelIPC_fp(dest);
     } else {
         /* Dequeue dest from the notification queue */
-        ntfn_queue_dequeue_fp(dest, ntfnPtr);
+        tcbNTFNDequeue(dest, ntfnPtr);
     }
 
     /* Wake up the signalled thread and transfer badge */
--- a/src/fastpath/fastpath.c
+++ b/src/fastpath/fastpath.c
@@ -706,7 +706,9 @@ void NORETURN fastpath_signal(word_t cptr, word_t msgInfo)
         if (NODE_STATE(ksCurThread)->tcbPriority > dest->tcbPriority || crossnode) {
             SCHED_ENQUEUE(dest);
         } else {
-            SCHED_APPEND(dest);
+            /* We know that this thread must be on the current core */
+            SMP_COND_STATEMENT(assert(sc->scCore == getCurrentCPUIndex()));
+            tcbSchedAppend(dest);
         }
     }
 
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -186,6 +186,11 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
     if (receiver->tcbSchedContext && isRunnable(receiver)) {
         sched_context_t *sc = receiver->tcbSchedContext;
         if ((refill_ready(sc) && refill_sufficient(sc, 0))) {
+            // TODO: is (refill_ready(sc) && refill_sufficient(sc, 0)
+            //         <=> sc_active)?
+            // If not, add a test to sel4test for this (unless guaranteed
+            // otherwise by this function)
+            assert(isSchedulable(receiver));
             possibleSwitchTo(receiver);
         } else {
             if (validTimeoutHandler(receiver) && fault_type != seL4_Fault_Timeout) {
@@ -468,8 +473,17 @@ void chooseThread(void)
 void switchToThread(tcb_t *thread)
 {
 #ifdef CONFIG_KERNEL_MCS
-    assert(thread->tcbSchedContext != NULL);
-    assert(!thread_state_get_tcbInReleaseQueue(thread->tcbState));
+    // XX: Could these two just be isSchedulable? That adds an extra sc_active()
+    //     call, but it also seems like most uses of switchToThread either
+    //     have an if(isSchedulable())) check or assert that it is the case.
+    //     We have isSchedulable() and refill_sufficient and refill_ready from
+    //     the asserts at tcbSchedAppend() often as well.
+    //     This isn't captured by sel4test when I change it, it notices no
+    //     difference, so either it's OK or sel4test is missing behaviours.
+    // 470⋮ 472│#ifdef CONFIG_KERNEL_MCS
+    // 471⋮    │    assert(thread->tcbSchedContext != NULL);
+    // 472⋮    │    assert(!thread_state_get_tcbInReleaseQueue(thread->tcbState));
+    assert(isSchedulable(thread));
     assert(refill_sufficient(thread->tcbSchedContext, 0));
     assert(refill_ready(thread->tcbSchedContext));
 #endif
@@ -520,6 +534,7 @@ void setMCPriority(tcb_t *tptr, prio_t mcp)
 #ifdef CONFIG_KERNEL_MCS
 void setPriority(tcb_t *tptr, prio_t prio)
 {
+    // XX: We don't use the possibleSwitchTo here?
     switch (thread_state_get_tsType(tptr->tcbState)) {
     case ThreadState_Running:
     case ThreadState_Restart:
@@ -568,23 +583,26 @@ void setPriority(tcb_t *tptr, prio_t prio)
 void possibleSwitchTo(tcb_t *target)
 {
 #ifdef CONFIG_KERNEL_MCS
-    if (target->tcbSchedContext != NULL && !thread_state_get_tcbInReleaseQueue(target->tcbState)) {
+    // XXX: This seems very similar to the isSchedulable check here.
+    //      See also my comment on switchToThread; we usually will also see
+    //      an assert on the candidate in schedule() that it is schedulable,
+    //      so why does this one omit sc_active?
+    // if (target->tcbSchedContext != NULL && !thread_state_get_tcbInReleaseQueue(target->tcbState)) {
+    assert(isSchedulable(target));
 #endif
-        if (ksCurDomain != target->tcbDomain
-            SMP_COND_STATEMENT( || target->tcbAffinity != getCurrentCPUIndex())) {
-            SCHED_ENQUEUE(target);
-        } else if (NODE_STATE(ksSchedulerAction) != SchedulerAction_ResumeCurrentThread) {
-            /* Too many threads want special treatment, use regular queues. */
-            rescheduleRequired();
-            tcbSchedEnqueue(target);
-            /* We know that this thread must be on the current core */
-            SMP_COND_STATEMENT(assert(target->tcbAffinity == getCurrentCPUIndex()));
-        } else {
-            NODE_STATE(ksSchedulerAction) = target;
-        }
-#ifdef CONFIG_KERNEL_MCS
+
+    if (ksCurDomain != target->tcbDomain
+        SMP_COND_STATEMENT( || target->tcbAffinity != getCurrentCPUIndex())) {
+        SCHED_ENQUEUE(target);
+    } else if (NODE_STATE(ksSchedulerAction) != SchedulerAction_ResumeCurrentThread) {
+        /* Too many threads want special treatment, use regular queues. */
+        rescheduleRequired();
+        tcbSchedEnqueue(target);
+        /* We know that this thread must be on the current core */
+        SMP_COND_STATEMENT(assert(target->tcbAffinity == getCurrentCPUIndex()));
+    } else {
+        NODE_STATE(ksSchedulerAction) = target;
     }
-#endif
 
 #ifdef ENABLE_SMP_SUPPORT
     /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
@@ -763,6 +781,10 @@ static void tcbReleaseDequeue(void)
     SMP_COND_STATEMENT(assert(awakened->tcbAffinity == getCurrentCPUIndex()));
     /* threads HEAD refill should always be >= MIN_BUDGET */
     assert(refill_sufficient(awakened->tcbSchedContext, 0));
+    // TODO: Why only refill_sufficient, and not refill_ready.
+    // TODO: Refill ready guaranteed by release_q_non_empty_and_ready()
+    // TODO: is ready + sufficient <=> sc_active? if not so, add sel4test.
+    assert(isSchedulable(awakened));
     possibleSwitchTo(awakened);
 }
 
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -374,6 +374,9 @@ void schedule(void)
 #ifdef ENABLE_SMP_SUPPORT
     /* Invariant: the current thread always belongs to the current core. */
     assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
 #endif
 
 #ifdef CONFIG_KERNEL_MCS
@@ -579,6 +582,11 @@ void possibleSwitchTo(tcb_t *target)
     }
 #endif
 
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
+#endif
 }
 
 void setThreadState(tcb_t *tptr, _thread_state_t ts)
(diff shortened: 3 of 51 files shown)


midnightveil:
Obviously, this needs some work before merging, but I wanted thoughts on this approach.

It might also be possible to remove the queue update inside migrateTCB if we look at its callers which probably do their own enqueues anyway, but I took the simple-and-correct approach .

src/object/schedcontrol.c
```diff
@@ -71,6 +71,12 @@ static exception_t invokeSchedControl_ConfigureFlags(sched_context_t *target, wo
         }
     }
 
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core.
+     * sel4test: SCHED0022 */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
+#endif
+
```
Indanz, line 83: If you check at all places where ksCurThread changes, then these other random checks don't add any value I think.

What you actually want to check is that if the target was currently running on another core, then it isn't any more when we're done here. However, because of `remoteTCBStall` that shouldn't be possible.

The commit message says:

"Note: ipiStallCoreCallback does a similar thing to what my plan is with this invariant, that is, switch to idle thread and then reschedule."

That is the most horrible part of all SMP support. :-(

include/model/statedata.h
```diff
@@ -120,6 +120,10 @@ extern paddr_t ksUserLogBuffer;
 #define SchedulerAction_ResumeCurrentThread ((tcb_t*)0)
 #define SchedulerAction_ChooseNewThread ((tcb_t*) 1)
 
+#define SchedulerAction_IsCandidateThread(schedulerAction) \
+    ((schedulerAction) != SchedulerAction_ResumeCurrentThread && \
+     (schedulerAction) != SchedulerAction_ChooseNewThread)
```
Indanz, line 125: Please move the `NODE_STATE(ksSchedulerAction)` part into this macro too while you're at it. You never call the macro with any other argument.

Or leave the code as it is and add a macro to check your assertion, as it's quite ugly and all over the place now.

(I also think that the whole scheduler action "optimisation" only complicates code and slows things down over all.)

src/kernel/thread.c
```diff
@@ -371,6 +376,16 @@ static void scheduleChooseNewThread(void)
 
 void schedule(void)
 {
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
```
Indanz, line 381: The only place ksCurThread changes is below and in the fast path code. Better to add this check to `lazyFPURestore`, as that's also called by `switchToThread_fp`, which is per arch.

src/kernel/thread.c
```diff
@@ -708,6 +716,12 @@ void timerTick(void)
 
 void rescheduleRequired(void)
 {
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
```
Indanz, line 773: I would prefer to only have checks after `ksSchedulerAction` gets changed, same for ksCurThread.

src/fastpath/fastpath.c
```diff
@@ -706,7 +706,9 @@ void NORETURN fastpath_signal(word_t cptr, word_t msgInfo)
         if (NODE_STATE(ksCurThread)->tcbPriority > dest->tcbPriority || crossnode) {
             SCHED_ENQUEUE(dest);
         } else {
-            SCHED_APPEND(dest);
+            /* We know that this thread must be on the current core */
+            SMP_COND_STATEMENT(assert(sc->scCore == getCurrentCPUIndex()));
```
Indanz, line 719: There's an explicit (stronger) `crossnode` check just above, no need for this assert.

src/kernel/thread.c
```diff
@@ -377,6 +377,8 @@ void schedule(void)
     /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
     assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
            NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
+    /* These invariants mean we don't need any remoteQueueUpdate calls, and
+       thus can call tcbSchedEnqueue/tcbSchedAppend directly. */
```
Indanz, line 386: This comment makes sense in a diff view context. I fear it will be somewhat confusing when seen in the code and will only add noise, because it's distracting and doesn't give additional information.

src/kernel/thread.c
```diff
@@ -574,7 +576,9 @@ void possibleSwitchTo(tcb_t *target)
         } else if (NODE_STATE(ksSchedulerAction) != SchedulerAction_ResumeCurrentThread) {
             /* Too many threads want special treatment, use regular queues. */
             rescheduleRequired();
-            SCHED_ENQUEUE(target);
+            tcbSchedEnqueue(target);
+            /* We know that this thread must be on the current core */
+            SMP_COND_STATEMENT(assert(target->tcbAffinity == getCurrentCPUIndex()));
```
Indanz: The first if checks for `target->tcbAffinity != getCurrentCPUIndex()`, this assert really doesn't add anything.

src/kernel/thread.c
```diff
@@ -186,6 +186,11 @@ void doReplyTransfer(tcb_t *sender, tcb_t *receiver, cte_t *slot, bool_t grant)
     if (receiver->tcbSchedContext && isRunnable(receiver)) {
         sched_context_t *sc = receiver->tcbSchedContext;
         if ((refill_ready(sc) && refill_sufficient(sc, 0))) {
+            // TODO: is (refill_ready(sc) && refill_sufficient(sc, 0)
+            //         <=> sc_active)?
```
Indanz, line 190: No, I think `sc_active` just means it's configured in a workable way so that the thread can run at all on it (but not necessarily now or any time soon).

Indanz (commented):
Reviewed till commit "[WIP] move check in possibleSwitchTo to a precondition", will check the rest later.

src/kernel/thread.c
midnightveil: But it's not the only place that ksCurThread changes affinity.

I'm happy to remove some of these asserts: a lot of them were me adding them to make sure I'd not missed cases. Specifically, adding in `schedule` was usually enough but meant all the different causes collapsed to this one assert and you couldn't tell where it came from.

But at the same time they're useful.

src/kernel/thread.c
midnightveil: I was consistent, mostly because I messed it up once or twice. But yes, when we have a preceding if statement checking it we can definitely remove it.

src/kernel/thread.c
Indanz: > But it's not the only place that ksCurThread changes affinity.

And that's why it's fine to add checks at those places too.

But at the start of schedule is just a random place like anywhere else and doesn't add value beyond initial debugging. Although it's not totally random, it would catch affinity changes for the current thread eventually, so if you don't add checks to the affinity changing places, keeping this one is fine. But then I'd rather have this check before and after changing ksCurThread, to keep it together.

src/fastpath/fastpath.c
midnightveil: Sure.

The placement and quantity of these asserts is not that important; and in theory we could drop them entirely. 

Whether or not we want that as an invariant is a different Q entirely.

src/object/endpoint.c
```diff
@@ -101,7 +101,10 @@ void sendIPC(bool_t blocking, bool_t do_call, word_t badge,
         if (sc_sporadic(dest_sc) && dest_sc != NODE_STATE(ksCurSC)) {
             refill_unblock_check(dest_sc);
         }
-        possibleSwitchTo(dest);
+        // XX: What happens if not isSchedulable()?
+        if (isSchedulable(dest)) {
```
Indanz, line 124: The only extra check this adds compared to the checks in `possibleSwitchTo` is `sc_active`. But that only checks `sc->scRefillMax > 0`, which must be true, otherwise the above existing assertion `refill_sufficient(dest_sc, 0)` would fail.

Edit: I somehow missed that you removed the checks in `possibleSwitchTo()`. I now understand your precondition comment. I hate the code duplication this introduces, so I rather keep it in `possibleSwitchTo()` and I don't see any gain from keeping it outside the function.

src/object/endpoint.c
```diff
@@ -258,8 +261,14 @@ void receiveIPC(tcb_t *thread, cap_t cap, bool_t isBlocking)
                 }
             } else {
                 setThreadState(sender, ThreadState_Running);
-                possibleSwitchTo(sender);
+                // Why no refill_ready like in sendIPC?
                 assert(sender->tcbSchedContext == NULL || refill_sufficient(sender->tcbSchedContext, 0));
+                if (isSchedulable(sender)) {
+                    possibleSwitchTo(sender);
```
Indanz, line 294: Same here. `isSchedulable` checks:
- isRunnable(thread) -> Always true because of the above `setThreadState(sender, ThreadState_Running);`
- thread->tcbSchedContext -> Checked by `possibleSwitchTo`
- !thread_state_get_tcbInReleaseQueue(thread->tcbState) -> Checked by `possibleSwitchTo`
- sc_active(thread->tcbSchedContext) -> Checked by `refill_sufficient`

src/object/endpoint.c
```diff
@@ -394,7 +404,11 @@ static inline void restart_thread_if_no_fault(tcb_t *thread)
                 refill_unblock_check(thread->tcbSchedContext);
             }
         }
-        possibleSwitchTo(thread);
+        // TODO: What if not schedulable?
+        if (isSchedulable(thread)) {
+            possibleSwitchTo(thread);
```
Indanz, line 440: I don't think should be a `possibleSwitchTo` at all, just normal enqueue.

Not sure about the `isSchedulable` check, this is called on each thread via `cancelAllIPC/cancelBadgedSends`.

src/object/notification.c
```diff
@@ -264,7 +264,10 @@ void cancelAllSignals(notification_t *ntfnPtr)
                     refill_unblock_check(thread->tcbSchedContext);
                 }
             }
-            possibleSwitchTo(thread);
+            // TODO: WHat happens if not schedulable?
```
Indanz: Nothing much. I honestly think all this MCS logic would be greatly simplified if treads are always schedulable when used, and weird stuff just doesn't work, like non-donating calls to passive servers.

src/object/schedcontrol.c
```diff
@@ -61,6 +61,8 @@ static exception_t invokeSchedControl_ConfigureFlags(sched_context_t *target, wo
         schedContext_resume(target);
         if (SMP_TERNARY(core == CURRENT_CPU_INDEX(), true)) {
             if (isRunnable(thread) && thread != NODE_STATE(ksCurThread)) {
+                // XX: guaranteed by assert sc_active(thread) and isRunnable.
+                assert(isSchedulable(thread));
```
Indanz, line 66: Then why add this?

src/object/tcb.c
```diff
@@ -441,6 +442,7 @@ word_t copyMRs(tcb_t *sender, word_t *sendBuf, tcb_t *receiver,
 #ifdef ENABLE_SMP_SUPPORT
 /* This checks if the current updated to scheduler queue is changing the previous scheduling
  * decision made by the scheduler. If its a case, an `irq_reschedule_ipi` is sent */
+// XXX: Does this only need to happen if isSchedulable()?
```
Indanz, line 446: Well, I don't think we want non-schedulable threads in the scheduling queue...

src/object/tcb.c
```diff
@@ -468,6 +470,7 @@ void remoteTCBStall(tcb_t *tcb)
 
     if (
 #ifdef CONFIG_KERNEL_MCS
+// XX: Why is this correct?
         tcb->tcbSchedContext &&
```
Indanz, line 475: Well, if it's the `ksCurThread`, it will have an SC. No idea why you want to check that here though.

src/fastpath/fastpath.c
```diff
@@ -670,6 +670,15 @@ void NORETURN fastpath_signal(word_t cptr, word_t msgInfo)
         crossnode = true;
     }
 
+#if defined(ENABLE_SMP_SUPPORT) || defined(CONFIG_KERNEL_MCS)
+    /* If we would cause our current thread to be migrated between cores,
+     * or between domains, go to the slowpath, as this would result in
+     * scheduling on our core/domain. */
+    if (crossnode && dest == NODE_STATE(ksCurThread)) {
+        slowpath(SysSend);
+    }
```
Indanz, line 679: You can't signal your own task cross-core or cross-domain.

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
 #endif /* CONFIG_HAVE_FPU */
-    tcb->tcbAffinity = new_core;
+    if (tcb == NODE_STATE(ksCurThread)) {
+        /**
+         * Switch the current thread to the idle thread. This is necessary to
+         * preserve the invariant the the current thread (NODE_STATE(ksCurThread))
+         * variable is always running on the current core.
+         *
+         */
+        switchToIdleThread();
```
Indanz, line 35: If we can guarantee that the scheduler picked another task (`rescheduleRequired()`?), we can just set `NODE_STATE(ksCurThread) = NODE_STATE(ksIdleThread)` here and trigger a `switchToIdleThread()` call by the scheduler if there is nothing else to run. `switchToIdleThread()` has no check whether the current task is already the idle task, so this would work I think.

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
+
+        /* Like rescheduleRequired but specific.
+         * This is broadly similar to how invokeTCB_SetAffinity() behaves.
```
Indanz, line 38: But `invokeTCB_SetAffinity` calls `migrateTCB`, so you have weird interaction going on here.

Ether all callers of `migrateTCB` need to do this, or it should only be done by `migrateTCB`.

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
+         */
+        tcb_t *action = NODE_STATE(ksSchedulerAction);
+        if (action == tcb || action == SchedulerAction_ResumeCurrentThread) {
+            NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
+        }
```
Indanz, line 43: Did I mention that I hate the `ksSchedulerAction` mechanism with a passion yet?

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
+
+        tcb->tcbAffinity = new_core;
+        if (isSchedulable(tcb)) {
+            SCHED_ENQUEUE(tcb);
+        }
+    } else {
+        tcb->tcbAffinity = new_core;
```
Indanz, line 50: Why can't you keep this at the beginning instead of duplicating it?

Indanz:
As far as I can tell, the actual fix is limited to `migrateTCB()`, correct?

midnightveil:
> As far as I can tell, the actual fix is limited to migrateTCB(), correct?

Yes, that ended up being the result of my testing, it covered all cases.

src/model/smp.c
midnightveil: The idea was that setting the affinity of the current thread at the start would cause the "current thread always on current core" invariant to be violated, albeit temporarily, and whilst the proofs probably could work around this it's easier to just not.

Then again, SMP will likely never have prooofs...

src/model/smp.c
midnightveil: Yes, that's why I said broadly: it omits this part.

> Ether all callers of migrateTCB need to do this, or it should only be done by migrateTCB.

Please see the message in the PR description: specifically note 2.

src/model/smp.c
midnightveil: iirc the issue with rescheduleRequired is that it requires the setting of the affinity to be done first, which temporarily violates the current-thread-on-current-core invariant.

src/fastpath/fastpath.c
midnightveil: Why not? (cross-domain but cross-core should be possible with an nonsense schedcontext setup iirc)

src/object/schedcontrol.c
midnightveil: To make it obvious that if I was incorrect here that the precondition assert of isSchedulable failing was because of this and so easier to track down. Admittedly most of the isSchedulable changes are unnecessary and could be dropped, I just don't like possibly-noop functions.

src/object/endpoint.c
midnightveil: Yes it's not a required change, but I prefer it because I don't like possibly no-op functions: because it becomes non-obvious whether or not the noop happened and if you should do something else then.

In this case the if should be converted to an assert or removed entirely if it is always true.

src/fastpath/fastpath.c
Indanz: How though? Your own task is always on the same core, because that's where it's currently running.

Did you had a situation in mind where there is a bound NF with an cross-core SC which may be donated? That can only happen with `seL4_NBSendRecv`, which isn't handled by the fastpath.

src/object/endpoint.c
midnightveil: (It's also true this would mess with verification, so it's likely worth dropping it).

However, the "no-op" case for possibleSwitchTo should be handled somehow I would think, which doing this movement allows you to see. However that should probably be separate PR?

lsf37:
So, what's the plan with this PR? It adds a lot of asserts to what for me are already way too many asserts on MCS. These asserts are not checked by verification and can fail any time if they are wrong, and if they are wrong they will also be quite misleading. I am absolutely not sure about many of the existing ones and I was actually planning on removing a lot of them now that the C verification has gotten there.

There are also a lot of what I would call dangerous  comments of the form "X is only used in Y" -- that may be true now, but anyone changing Y will not see X and definitely not update the comment, very likely also not when adding another use of X. They make sense for figuring something out about what the code does now (and that's fine for documenting this PR), but I don't think they should stay longer term.

Documenting expected invariants for SMP is good, though, that's exactly the opposite of "X only used this way". Same for preconditions. No problem adding those. 

For some of the "is this enough" questions for single-core MCS there should now be answers on the verification side. If there are important ones, I could go digging if you want me to.

midnightveil:
I was waiting on an agreement on whether or not the proposed invariant: that is, the current thread is running on the current core, seems like (a) a good idea, and (b) whether the change in migrateTCB (in smp.c) is an acceptable way to implement it.

There's also the semi-additional invariant of whether the current SC should also be on the current core, which I don't think is true because when the thread changes the SC won't necessarily be updated immediately.


---

> There are also a lot of what I would call dangerous comments of the form "X is only used in Y" -- that may be true now, but anyone changing Y will not see X and definitely not update the comment, very likely also not when adding another use of X. They make sense for figuring something out about what the code does now (and that's fine for documenting this PR), but I don't think they should stay longer term.

Yes, these definitely should not stay, nor should a lot of other comments.

>  If there are important ones, I could go digging if you want me to.

Yes, there are a few I'd like to check, but I need to context switch back to see which are most important.

> Documenting expected invariants for SMP is good, though, that's exactly the opposite of "X only used this way". Same for preconditions. No problem adding those.

Is it OK to assert those invariants in relevant places? Or should we remove those completely.

Alternatively, if it fails the invariant I could make the code still work by doing a manual migrateTCB call.

> It adds a lot of asserts to what for me are already way too many asserts on MCS. T

A lot of the ones around isSchedulable/possibleSwitchTo are... entirely optional and probably annoying to do. One thing I did when I was experimenting with this PR was that I removed the if-check from inside possibleSwitchTo that made it maybe do nothing; one issue with that (which is one of what the previous fix PR did) was it handled that else-case for one particular case; as if it wasn't true something else needed to happen. This can be undone, but I would like to follow-up on a few of the else-cases where I couldn't immediately see what happened/why it was correct to do nothing.


Just the change to migrateTCB would make this much smaller; but a lot of the comments are my uh, "execution trace" so to say of how I figured out that just doing that would be fine in the end; and that I never needed to update anywhere else.

midnightveil:
I suppose what I could do tomorrow is put up a PR containing the minimal diff.

midnightveil:
I'm closing this in favour of https://github.com/seL4/seL4/pull/1707.

PR #1663, opened 2026-05-15
workflows: use ${{ variable }} not invalid $${{

midnightveil:
I'm not entirely sure how this worked... but I guess GitHub ignores the extra $ sign.

### DIFF
--- a/.github/workflows/sel4test-deploy.yml
+++ b/.github/workflows/sel4test-deploy.yml
@@ -105,7 +105,7 @@ jobs:
           platform: ${{ matrix.platform }}
           compiler: ${{ matrix.compiler }}
           mode: ${{ matrix.mode }}
-          index: $${{ strategy.job-index }}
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

lsf37:
The "maintainers can modify" flag isn't set, so I can't update the branch to merge from my side. Otherwise good to go, though.

midnightveil:
> The "maintainers can modify" flag isn't set, so I can't update the branch to merge from my side. Otherwise good to go, though.

It's because it's part of an org repo: it's impossible to set this unfortunately :(

lsf37:
> > The "maintainers can modify" flag isn't set, so I can't update the branch to merge from my side. Otherwise good to go, though.
> 
> It's because it's part of an org repo: it's impossible to set this unfortunately :(

Ah, that's why that keeps popping up. Not much we can do about it, I guess.

PR #1664, opened 2026-05-20
x86_64/traps.S: fix infinite exception when using UBSAN

dreamliner787-9:
The System V AMD64 ABI requires the stack pointer (%rsp) to be 16-byte 
aligned immediately before a `call` instruction. In `kernel_exception`, 
pushing three 8-byte control registers (%cr2, %cr3, %cr4) after 
`INT_SAVE_STATE` misaligned the stack prior to calling 
`handleKernelException`. 

This commit adds an 8-byte padding before pushing the control registers 
to maintain 16-byte alignment.

Previously, compiling the kernel with LLVM and UBSAN enabled resulted in 
a recursive exception loop if a UB was encountered elsewhere in kernel
code. The misaligned stack would trigger UBSAN during exception
printing, causing an exception within the exception handler that 
looped infinitely.

### DIFF
--- a/src/arch/x86/64/traps.S
+++ b/src/arch/x86/64/traps.S
@@ -568,6 +568,11 @@ BEGIN_FUNC(kernel_exception)
     movq    24(%r8), %r8     # RFLAGS
     # handleKernelException(vector, errorcode, RIP, RSP, RFLAGS, CR0, CR2, CR3, CR4)
     movq    %cr0, %r9
+
+    # At this point the stack would be 16-bytes aligned, but we are only pushing 3x 8-byte
+    # values below, so we need to maintain stack alignment.
+    subq    $8, %rsp
+
     movq    %cr4, %r11
     push    %r11
     movq    %cr3, %r11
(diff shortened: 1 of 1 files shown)


Indanz (approved):
Do you have an architectural reference for this requirement?

This is kernel code, so the System V AMD64 ABI does not apply directly.

dreamliner787-9:
Hi Indan, this is what I was referring to: https://cs61.seas.harvard.edu/site/2025/pdf/x86-64-abi-20210928.pdf

> 3.2.2 The Stack Frame
> ...
> The end of the input argument area shall be aligned on a 16 (32 or 64, if __m256 or __m512 is passed on stack) byte boundary. In other words, the stack needs to be 16 (32 or 64) byte aligned immediately before the call instruction is executed.

It's not an architectural requirement in and of itself. But once we enter `handleKernelException`, and that C code calls other C code for printing, etc. That's where UBSAN wasn't happy because it would assume that the standard convention applies.

PR #1665, opened 2026-05-20
pc99/acpi: fix interrupt source override padding

### LINKED ISSUE #396: Unexpected behaviour on x86-64 when building SDK using LLVM
This issue does not affect any released versions of Microkit since the SDKs available for download are using GCC.

This also does not seem to affect anyone using LLVM for their program images (at TS that is what we do by default for our projects using Microkit), we've only had issues when the SDK itself is built with LLVM.

More investigation required.

dreamliner787-9:
`acpi_madt_iso_t` has an actual size of 10 bytes and 1 byte alignment.
But the compiler would sometimes pad it to 12 bytes and
require a 4 bytes alignment.

Previously there is no compile time assert to catch this issue. So
depending on how the BIOS laid out the MADT overall, the dereferences
under `case MADT_ISO:` of `acpi_madt_scan()` would be unaligned.

If you build the kernel with LLVM and UBSAN enabled then it would cause
the kernel to triple fault and endlessly reboot your machine. Since at
this point the kernel have not installed it's exception handlers.

Splitted the u32 in `acpi_madt_iso_t` into 2x u16 to ensure it's not
padded.

### DIFF
--- a/src/plat/pc99/machine/acpi.c
+++ b/src/plat/pc99/machine/acpi.c
@@ -143,14 +143,13 @@ typedef struct acpi_madt_iso {
     acpi_madt_header_t header;
     uint8_t            bus; /* always 0 (ISA) */
     uint8_t            source;
-    uint32_t           gsi;
+    /* Split the u32 into 2x u16 to prevent the compiler from padding the struct. */
+    uint16_t           gsi_low;
+    uint16_t           gsi_high;
     uint16_t           flags;
 } acpi_madt_iso_t;
-/* We can't assert on the sizeof acpi_madt_iso because it contains trailing
- * padding.
- */
-unverified_compile_assert(acpi_madt_iso_packed,
-                          OFFSETOF(acpi_madt_iso_t, flags) == sizeof(acpi_madt_header_t) + 6)
+compile_assert(acpi_madt_iso_packed,
+               sizeof(acpi_madt_iso_t) == sizeof(acpi_madt_header_t) + 8)
 
 /* workaround because string literals are not supported by C parser */
 const char acpi_str_rsd[]  = {'R', 'S', 'D', ' ', 'P', 'T', 'R', ' ', 0};
@@ -358,13 +357,12 @@ BOOT_CODE uint32_t acpi_madt_scan(
                         (*num_ioapic)++;
                     }
                     break;
-                case MADT_ISO:
-                    printf("ACPI: MADT_ISO bus=%d source=%d gsi=%d flags=0x%x\n",
-                           ((acpi_madt_iso_t *)acpi_madt_header)->bus,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->source,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->gsi,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->flags);
+                case MADT_ISO: {
+                    UNUSED acpi_madt_iso_t *iso = (acpi_madt_iso_t *)acpi_madt_header;
+                    printf("ACPI: MADT_ISO bus=%d source=%d gsi=0x%x%02x flags=0x%x\n",
+                           iso->bus, iso->source, iso->gsi_high, iso->gsi_low, iso->flags);
                     break;
+                }
                 default:
                     break;
                 }


src/plat/pc99/machine/acpi.c
```diff
@@ -139,13 +139,17 @@ typedef struct acpi_madt_ioapic {
...
 
-typedef struct acpi_madt_iso {
+struct acpi_madt_iso {
     acpi_madt_header_t header;
     uint8_t            bus; /* always 0 (ISA) */
     uint8_t            source;
     uint32_t           gsi;
     uint16_t           flags;
-} acpi_madt_iso_t;
+} __attribute__((packed));
```
lsf37: That attribute on structs is not in the verification subset, so this unfortunately does not work.

lsf37: We could explicitly add two bytes for padding to get a word-multiple size, though. At least that is what we do in other parts of the kernel.

dreamliner787-9: I see that things like `acpi_header_t` in here https://github.com/seL4/seL4/blob/3527859e72e524a819bb44102372e045836f7c28/include/plat/pc99/plat/machine/acpi.h#L29 uses `__attribute__((packed))` so I'm wondering why it can't be applied to `acpi_madt_iso_t` in this PR as well?

midnightveil: Note that the C parser doesn't like this at the moment, per the CI run. Perhaps it needs to look slightly different:

```
  /github/workspace/kernel/src/plat/pc99/machine/acpi.c:148.2-148.24: Extraneous crap after struct declaration
```

The C parser does have tests that use `((packed))`, so I'm not sure why it would be different here though. https://github.com/seL4/l4v/blob/master/tools/c-parser/testfiles/parse_struct.c#L100-L106

lsf37: hm, that does seem to work. I'm reasonably sure we are not modelling the effect of `packed`, though, so this might be interesting when it is ever used in a proof. Should be safe, though, in the sense that the C parser will not assume that there is no padding.

midnightveil: Per `StrictC.grm`:

```
attribute_specifier
: GCC_ATTRIBUTE LPAREN LPAREN attribute_list RPAREN RPAREN
                (wrap(attribute_list, GCC_ATTRIBUTEleft, RPAREN2right))
| OWNED_BY ID SPEC_BLOCKEND
                (wrap([OWNED_BY ID], OWNED_BYleft, SPEC_BLOCKENDright))

attribute_list
:               ([])
| attribute     ([attribute])
| attribute YCOMMA attribute_list
                (attribute :: attribute_list)

attribute
: ID            (let val idstr = if String.isPrefix "__" ID andalso
                                    String.sub(ID,size ID - 1) = #"_" andalso
                                    String.sub(ID,size ID - 2) = #"_" andalso
                                    size ID > 4
                                 then
                                   String.substring(ID,2,size ID - 4)
                                 else ID
                 in
                   GCC_AttribID idstr
                 end)
| CONST         (GCC_AttribID "const")
| ID LPAREN RPAREN
                (GCC_AttribFn (ID, []))
| ID LPAREN attribute_parameter_list1 RPAREN
                (GCC_AttribFn (ID, attribute_parameter_list1))

attribute_parameter_list1
: rexpression   ([rexpression])
| rexpression YCOMMA attribute_parameter_list1
                (rexpression :: attribute_parameter_list1)

```

So the only one specially handled is `GCC_AttribID "const"` and there's not really any other uses of `GCC_AttribID` aside from for "Noreturn". I can't see anything extra around, so yeah, I think this is mostly just ignored by the parser and the extracted layout would be wrong.

lsf37: It would assume that the compiler might insert padding, which usually means fewer things can be proved about it. I haven't really thought through all the implications, though. There might still be cases where this could lead to a wrong result.

Indanz (approved):

src/plat/pc99/machine/acpi.c
lsf37: Actually, Julia is right, this will just be a wrong model. I was mostly thinking about alignment and field access, but offsets and overall size will just be plain wrong. We need to be careful that this is not used for structs the proofs actually reason about, but I don't think it is a problem for this specific one, so I'd be fine with merging this.

lsf37 (approved):

src/plat/pc99/machine/acpi.c
lsf37: Have added an issue on l4v for this https://github.com/seL4/l4v/issues/994

Indanz:
In this case a simple work-around is to split `gsi` into two 16-bit values and combine them manually, then no packing is necessary.

Edit: Actually, the code uses `acpi_madt_header->length` to jump to the next entry, it is harmless that the sizeof doesn't match the actual size.

If it's just UBSAN that triggers this because it assumes that 32-bit words are aligned, then you need to add PACKED to all structures containing `uint32_t`, not just `acpi_madt_iso`. Any entry can be unaligned because of a previous entry.

lsf37:
Why are we now adding PACKED everywhere? We should avoid using it when we can. We may have to forbid it entirely at some point. Currently it is a soundness hazard for verification.

For C proofs, it's fine for structs behind the machine interface that the proofs do not interact with, but it will make binary verification impossible for functions that use it.

Edit: right, I had missed Indan's message. I'm still not happy with it.

src/plat/pc99/machine/acpi.c
```diff
@@ -49,7 +49,7 @@ typedef struct acpi_dmar_drhd {
     uint8_t            reserved;
     uint16_t           segment;
     uint32_t           reg_base[2];
-} acpi_dmar_drhd_t;
+} PACKED acpi_dmar_drhd_t;
 compile_assert(acpi_dmar_drhd_packed,
                sizeof(acpi_dmar_drhd_t) == sizeof(acpi_dmar_header_t) + 12)
```
lsf37: Why is the PACKED necessary if we already have a compile time assert that the structure is naturally packed? Are we not just forcing the check to be true and making it redundant?

lsf37 (changes requested):

src/plat/pc99/machine/acpi.c
dreamliner787-9: It's to workaround this compile error:
```
/home/dreamliner7879/TS/microkit-capdl-dev/seL4_ts/src/plat/pc99/machine/acpi.c:487:47: error: taking address of packed member 'devscope_0' of class or structure 'acpi_dmar_rmrr' may result in an unaligned pointer value [-Werror,-Waddress-of-packed-member]
  487 |                         acpi_dmar_devscope = &acpi_dmar_rmrr->devscope_0 + i;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

lsf37: How does adding PACKED change that? I would think it could only make it worse. Or are we specifically making something smaller so that something later becomes aligned? It'd still be a member of a packed struct, though, I don't see how it changes the error.

I assume it is gone when you add it, but I don't understand yet how and if there is another way to achieve the same thing. Is the packed memory layout prescribed by the hardware?

midnightveil:
> Edit: right, I had missed Indan's message. I'm still not happy with it.

I think it makes more sense to use Indan's other suggestion: that is, split the 32-bit in two and rearrange the struct layout to be fine (and then we can assert that the alignment/size matches what it should be). Especially given we probably should stay within the verification subset if we can.

Alternatively, we could just `memcpy` into a local struct that *is* aligned.

src/plat/pc99/machine/acpi.c
Indanz: Declaring a structure packed also changes what alignment the compiler assumes. If `acpi_dmar_devscope_t` is packed, the compiler won't assume natural alignment of pointers of that type.

If you want to pass unaligned pointers, you have to tell the compiler that some way if you want to avoid the warning. Doing it via PACKED is just the easiest way.

(I don't know how making `acpi_dmar_drhd_t` packed helps with the above compile error though, probably a mixup. But all pointers have this issue.)

Indanz:
> > Edit: right, I had missed Indan's message. I'm still not happy with it.

Well, some things are just unaligned. But in this case I don't understand why they haven't kept all structures 32-bit aligned, it would have been trivial and would have simplified parsing. Wouldn't surprise me if real `acpi_madt_iso` in the field report being 2 bytes larger than expected to make it modulo 32-bit.

> I think it makes more sense to use Indan's other suggestion: that is, split the 32-bit in two and rearrange the struct layout to be fine (and then we can assert that the alignment/size matches what it should be). Especially given we probably should stay within the verification subset if we can.

You have to do this for all 32-bit fields in all ACPI structures though.

> Alternatively, we could just `memcpy` into a local struct that _is_ aligned.

Probably the easiest work-around for verification. They're small enough to copy on stack.

Indanz (approved):
Safest would be to add PACKED to all structures. For instance, I don't know if 16-bit alignment is guaranteed by ACPI. I wouldn't count on it.

Indanz:
BTW, the parsing code is currently trusting the ACPI tables quite a bit: It only does a `header.length` check, but does not check that `header.length` is big enough to cover the structure size the parser code expects. Now the parser can look beyond the end of `acpi_madt_mapped`.

src/plat/pc99/machine/acpi.c
lsf37: Ok, so the point of the change is to tell the compiler that `acpi_dmar_devscope` is already potentially unaligned (because `acpi_dmar_devscope_t` is packed), and therefore it is fine if `acpi_dmar_rmrr->devscope_0` is at an unaligned address? Compilers are nuts.

Would using the attribute `aligned(1)` on the pointer do the same thing? This is probably rejected by the C parser at the moment, but we could add it. Not sure this is really much better, though. It'll still lead to wrong (wrong = not describing reality) behaviour of the C memory model in verification, so I'm not sure it's worth the work to add to the C parser.

But It would at least describe what we are trying to achieve. Currently, what we are doing looks nonsensical -- we're declaring PACKED on structs that we have asserts for being packed. At the very least this needs comments that we are interested in struct pointer alignment, not in the struct layout. And probably removal of the asserts, because there is no point to them any more.

If that's the best way to do this correctly for the compiler, I'm Ok with going forward for now. We do really need to be careful with this, though. If this leaks anywhere into verified code, the verification no longer applies.

lsf37:
> > Alternatively, we could just `memcpy` into a local struct that _is_ aligned.
> 
> Probably the easiest work-around for verification. They're small enough to copy on stack.

How much of that would we need to do? If it's a small amount only, that would be preferable, but if it just leads to a big mess, we should probably go the other way.

src/plat/pc99/machine/acpi.c
midnightveil: the aligned attribute can only increase alignment, not decrease it unfortunately.

midnightveil:
I see 2 or 3 switch statements with a few cases each, from a quick count about 10 instances of this pattern. I don't think it would be particularly onerous.

dreamliner787-9:
I've applied the u32 split approach to `acpi_madt_iso_t` so that it's verification friendly while still keeping the struct size equals to the true size. This fixed the kernel bootlooping on QEMU when built with LLVM+UBSAN for me.

I've also change the MADT ISO printing step so that the structure get copied into a stack allocated memory to satisfy any alignment requirements from the compiler. If everyone is happy with this then I can apply this copy to stack step for all the other structures, perhaps in a separate PR.

dreamliner787-9:
Oops

src/plat/pc99/machine/acpi.c
```diff
@@ -358,13 +357,16 @@ BOOT_CODE uint32_t acpi_madt_scan(
                         (*num_ioapic)++;
                     }
                     break;
-                case MADT_ISO:
+                case MADT_ISO: {
+                    acpi_madt_iso_t iso;
+                    memcpy(&iso, acpi_madt_header, sizeof(acpi_madt_iso_t));
```
Indanz: You could do one memcopy into a buffer at the top and use the header length for copying. Then cast the buffer pointer everywhere. It would avoid a whole bunch of memcopy calls all over the place.

You really need to add a check for specified length being at least equal to the expected structure size. This whole commit is about UBSAN complaining about theoretical problems, let's at least fix the real ones.

dreamliner787-9: I think we can merge this as is for now which fixes the UBSAN crash. Then I will implement your feedback in another PR.

Edit: done in https://github.com/seL4/seL4/pull/1684 and https://github.com/seL4/seL4/pull/1683

src/plat/pc99/machine/acpi.c
```diff
@@ -358,13 +357,18 @@ BOOT_CODE uint32_t acpi_madt_scan(
...
+                    uint16_t gsi_lo = ((acpi_madt_iso_t *)acpi_madt_header)->gsi_lo;
+                    uint16_t gsi_hi = ((acpi_madt_iso_t *)acpi_madt_header)->gsi_hi;
                     printf("ACPI: MADT_ISO bus=%d source=%d gsi=%d flags=0x%x\n",
                            ((acpi_madt_iso_t *)acpi_madt_header)->bus,
                            ((acpi_madt_iso_t *)acpi_madt_header)->source,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->gsi,
+                           (uint32_t) gsi_lo | ((uint32_t) gsi_hi) << 16,
                            ((acpi_madt_iso_t *)acpi_madt_header)->flags);
+                    (void)gsi_lo;
+                    (void)gsi_hi;
```
Indanz: Don't we have UNUSED to avoid these?

You could also just print them out as hex and concatenate them in the printf with: `0x%x%0.2x`

Also, why not add a local var and do the horrible cast once instead of 5 times?

src/plat/pc99/machine/acpi.c
```diff
@@ -358,13 +357,16 @@ BOOT_CODE uint32_t acpi_madt_scan(
...
-                           ((acpi_madt_iso_t *)acpi_madt_header)->gsi,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->flags);
+                case MADT_ISO: {
+                    UNUSED acpi_madt_iso_t *iso = (acpi_madt_iso_t *)acpi_madt_header;
+                    printf("ACPI: MADT_ISO bus=%d source=%d gsi=0x%x%02x flags=0x%x\n",
+                           iso->bus,
+                           iso->source,
+                           iso->gsi_hi,
+                           iso->gsi_lo,
+                           iso->flags);
```
Indanz:
```suggestion
                    printf("ACPI: MADT_ISO bus=%d source=%d gsi=0x%x%02x flags=0x%x\n",
                           iso->bus, iso->source, iso->gsi_high, iso->gsi_low, iso->flags);
```

Indanz (approved):
Gerwin, can you have another look? This should be fine for verification now.

lsf37 (approved):
Yes, this looks good now.

PR #1666, opened 2026-05-20
x86_64/vspace: made init_syscall_msrs not bootcode

dreamliner787-9:
`init_syscall_msrs()` is also called by `vcpu_restore_host_msrs()` so should not be BOOT_CODE.


lsf37 (approved):

Indanz (approved):
There's also a 32-bit static inline `init_syscall_msrs`, but `vcpu_restore_host_msrs` is called from 64-bit code, so uses the boot one.

PR #1669, opened 2026-05-20
x86/c_traps.c: VMX clobbering fixes

dreamliner787-9:
Please see commit messages for motivation, fixes guest hanging on x86 when the kernel is compiled with LLVM.

Tested with CAmkES VMM (minimal example on both 32 and 64 bits kernels) and our x86_64 VMM.

### DIFF
--- a/src/arch/x86/32/c_traps.c
+++ b/src/arch/x86/32/c_traps.c
@@ -36,11 +36,12 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
     /* Do not support breakpoints in VMs, so just disable all breakpoints */
     loadAllDisabledBreakpointState(cur_thread);
 #endif
+    word_t *guest_regs_ptr = vcpu->gp_registers;
     if (vcpu->launched) {
         /* attempt to do a vmresume */
         asm volatile(
             // Set our stack pointer to the top of the tcb so we can efficiently pop
-            "movl %0, %%esp\n"
+            "movl %[guest_regs], %%esp\n"
             "popl %%eax\n"
             "popl %%ebx\n"
             "popl %%ecx\n"
@@ -57,18 +58,17 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
             "leal kernel_stack_alloc + %c1, %%esp\n"
 #endif
             "call vmlaunch_failed\n"
-            :
-            : "r"(&vcpu->gp_registers[VCPU_EAX]),
-            "i"(BIT(CONFIG_KERNEL_STACK_BITS) - sizeof(word_t))
+            : [guest_regs] "+a"(guest_regs_ptr)
+            : "i"(BIT(CONFIG_KERNEL_STACK_BITS) - sizeof(word_t))
             // Clobber memory so the compiler is forced to complete all stores
             // before running this assembler
-            : "memory"
+            : "ebx", "ecx", "edx", "esi", "edi", "ebp", "memory"
         );
     } else {
         /* attempt to do a vmlaunch */
         asm volatile(
             // Set our stack pointer to the top of the tcb so we can efficiently pop
-            "movl %0, %%esp\n"
+            "movl %[guest_regs], %%esp\n"
             "popl %%eax\n"
             "popl %%ebx\n"
             "popl %%ecx\n"
@@ -85,12 +85,11 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
             "leal kernel_stack_alloc + %c1, %%esp\n"
 #endif
             "call vmlaunch_failed\n"
-            :
-            : "r"(&vcpu->gp_registers[VCPU_EAX]),
-            "i"(BIT(CONFIG_KERNEL_STACK_BITS) - sizeof(word_t))
+            : [guest_regs] "+a"(guest_regs_ptr)
+            : "i"(BIT(CONFIG_KERNEL_STACK_BITS) - sizeof(word_t))
             // Clobber memory so the compiler is forced to complete all stores
             // before running this assembler
-            : "memory"
+            : "ebx", "ecx", "edx", "esi", "edi", "ebp", "memory"
         );
     }
     UNREACHABLE();
--- a/src/arch/x86/64/c_traps.c
+++ b/src/arch/x86/64/c_traps.c
@@ -41,16 +41,10 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
     vcpu_restore_guest_msrs(vcpu);
 #endif /* CONFIG_X86_64_VTX_64BIT_GUESTS */
+
     /* attempt to do a vmlaunch/vmresume */
     asm volatile(
-        // Arguments are getting stored in general purpose registers that need to be used.
-        // Copy them to unused general purpose registers
-        "movq %[launched], %%rbx\n"
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
-        "movq %[host_msr], %%r8\n"
-        "movq %[guest_msr], %%r9\n"
-        "movq %[reg], %%r10\n"
-
         // Save host's GS, Shadow GS, and FS
         "mov $0xC0000101, %%ecx\n"
         "rdmsr\n"
@@ -67,7 +61,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
         "shl $0x20,%%rdx\n"
         "or %%rdx, %%rax\n"
         "mov %%rax, %%r13\n" // R13 has FS
-        "movq %%r8, %%rsp\n" // host_gs
+        "movq %[host_msr], %%rsp\n" // host_gs
         "pushq %%r13\n"
         "pushq %%r12\n"
         "pushq %%r11\n"
@@ -75,7 +69,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
         // Restore guest's GS and Shadow GS
         "mov $0xC0000101, %%ecx\n"
         "swapgs\n"
-        "movq %%r9, %%rsp\n" // guest_gs
+        "movq %[guest_msr], %%rsp\n" // guest_gs
         "popq %%rax\n"       // GS
         "mov %%rax,%%rdx\n"
         "shr $0x20,%%rdx\n"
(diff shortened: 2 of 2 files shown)


src/arch/x86/32/c_traps.c
```diff
@@ -62,17 +62,17 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #endif
         "call vmlaunch_failed\n"
         :
-        : "r"(&vcpu->gp_registers[VCPU_EAX]),
+        : [reg] "a"(&vcpu->gp_registers[VCPU_EAX]),
```
Indanz: It seems you force it to `rax` by specifying "a". Why not name it `rax` or `eax` instead of `reg` to make this more clear?

Also, you pop eax, so rax is clobbered too, I think you should use "+a".

src/arch/x86/32/c_traps.c
```diff
@@ -90,11 +90,11 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #endif
             "call vmlaunch_failed\n"
             :
-            : "r"(&vcpu->gp_registers[VCPU_EAX]),
+            : [reg] "a"(&vcpu->gp_registers[VCPU_EAX]),
```
Indanz: Same.

src/arch/x86/64/c_traps.c
```diff
@@ -199,7 +192,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #endif /* CONFIG_X86_64_VTX_64BIT_GUESTS */
         // Clobber memory so the compiler is forced to complete all stores
         // before running this assembler
-        : "memory"
+        : "memory", "rax", "rcx", "rdx", "r11", "r12", "r13"
```
Indanz: How is this supposed to work?

Before vmresume/vmlaunch, all registers are replaced with the VM's saved values. If the instruction fails, we end up at `done:` and try to use passed parameters. At that point the only safe ones to use are immediate, constant ones. So the `[failed]"r"(vmlaunch_failed),` is wrong.

I suppose it's safe to use `guest_msr` and `host_msr` earlier, if the clobber list is correct and lists all used registers before that point. As far as I can tell, you do that correctly. However, it would be better to explicitly specify which registers to use for those inputs, as that is more robust than clobbering all other registers used.

dreamliner787-9: > Before vmresume/vmlaunch, all registers are replaced with the VM's saved values. If the instruction fails, we end up at done: and try to use passed parameters. At that point the only safe ones to use are immediate, constant ones. So the [failed]"r"(vmlaunch_failed), is wrong.

Yes that's right, I fixed this problem in https://github.com/seL4/seL4/pull/1667.

> I suppose it's safe to use guest_msr and host_msr earlier, if the clobber list is correct and lists all used registers before that point. As far as I can tell, you do that correctly. However, it would be better to explicitly specify which registers to use for those inputs, as that is more robust than clobbering all other registers used.

Ok, I did it like that originally because I thought it looks clean, but this works too.

src/arch/x86/32/c_traps.c
dreamliner787-9: I don't think we should use "+a" because `vmlaunch_failed` is `NORETURN` anyways.

Indanz: That's not a very convincing argument to have incorrect code. Not listing all registers being clobbered because of the NORETURN is one thing, but knowingly saying a register is read-only while it is read-write is altogether a different thing.

dreamliner787-9: Yes I understand what you mean now, I tried "+a" before but got a compile error because I didn't do it correctly before. Will fix.

src/arch/x86/32/c_traps.c
```diff
@@ -44,7 +44,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
     /* attempt to do a vmresume */
     asm volatile(
         // Set our stack pointer to the top of the tcb so we can efficiently pop
-        "movl %0, %%esp\n"
+        "movl %[guest_regs], %%esp\n"
```
Indanz: `guest_stack` would be a more appropriate name.

src/arch/x86/32/c_traps.c
```diff
@@ -41,38 +41,39 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
...
-        "popl %%edx\n"
-        "popl %%esi\n"
-        "popl %%edi\n"
-        "popl %%ebp\n"
-        // Now do the vmresume
-        "vmresume\n"
-        // if we get here we failed
+    word_t *guest_state_ptr = &vcpu->gp_registers[VCPU_EAX];
+        /* attempt to do a vmresume */
+        asm volatile(
```
Indanz: What happened to indentation? This whole block is shifted to the right from here on.

src/arch/x86/32/c_traps.c
```diff
@@ -41,38 +41,39 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
...
-        "popl %%ebx\n"
-        "popl %%ecx\n"
-        "popl %%edx\n"
-        "popl %%esi\n"
-        "popl %%edi\n"
-        "popl %%ebp\n"
-        // Now do the vmresume
-        "vmresume\n"
-        // if we get here we failed
+    word_t *guest_state_ptr = &vcpu->gp_registers[VCPU_EAX];
```
Indanz:
```suggestion
    word_t *guest_state_ptr = vcpu->gp_registers;
```
If VCPU_EAX wasn't 0 your version would be better of course. But because it is, it's more confusing.

Also, guest_regs_ptr would be more consistent, there is a lot of other guest state elsewhere.

src/arch/x86/32/c_traps.c
```diff
@@ -41,38 +41,39 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
...
-        "call vmlaunch_failed\n"
-        :
-        : "r"(&vcpu->gp_registers[VCPU_EAX]),
-        "i"(BIT(CONFIG_KERNEL_STACK_BITS) - sizeof(word_t))
-        // Clobber memory so the compiler is forced to complete all stores
-        // before running this assembler
-        : "memory"
-    );
+            "call vmlaunch_failed\n"
+            : [guest_regs] "+a"(guest_state_ptr)
```
Indanz: You can keep this in the input list, that would make more sense (even though it looks neater here).

src/arch/x86/64/c_traps.c
```diff
@@ -183,12 +187,12 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
...
-        : [reg]"r"(&vcpu->gp_registers[VCPU_EAX]),
-        [launched]"r"(&vcpu->launched),
+        : [reg]"r"(r_reg),
+        [launched]"r"(r_launched),
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
         [stack_size]"i"(BIT(CONFIG_KERNEL_STACK_BITS)),
-        [guest_msr]"r"(&vcpu->guest_msr_registers[VCPU_GS]),
-        [host_msr]"r"(&vcpu->host_msr_registers[n_vcpu_msr_register])
+        [guest_msr]"r"(r_guest_msr),
+        [host_msr]"r"(r_host_msr)
```
Indanz: I find all the `r_` prefixes quite ugly, but whatever you prefer.

Also consistent naming between the 32-bit and 64-bit version would make sense.

src/arch/x86/32/c_traps.c
dreamliner787-9: I placed it in the output list because I was running into this problem:
```
/home/dreamliner7879/TS/camkes_vmm/kernel/src/arch/x86/32/c_traps.c: In function ‘restore_vmx’:
/home/dreamliner7879/TS/camkes_vmm/kernel/src/arch/x86/32/c_traps.c:70:5: error: input operand constraint contains ‘+’
   70 |     );
      |     ^
/home/dreamliner7879/TS/camkes_vmm/kernel/src/arch/x86/32/c_traps.c:98:9: error: input operand constraint contains ‘+’
   98 |         );
      |         ^
```

src/arch/x86/32/c_traps.c
dreamliner787-9: It was to appease the C style checker, but I've worked around the problem by placing this line outside the if statement instead as it is used in all paths anyways:
```c
word_t *guest_state_ptr = &vcpu->gp_registers[VCPU_EAX];
```

dreamliner787-9: Ok I'm not quite sure what's going on here:
```
Run seL4/ci-actions/style@master
/home/runner/work/_actions/seL4/ci-actions/master/js/../style/steps.sh
Setting up

Checking the following files:
src/arch/x86/32/c_traps.c
src/arch/x86/64/c_traps.c

Formatted  src/arch/x86/32/c_traps.c
Unchanged  src/arch/x86/64/c_traps.c
diff --git a/src/arch/x86/32/c_traps.c b/src/arch/x86/32/c_traps.c
index 36a5d27c..1306861b 100644
--- a/src/arch/x86/32/c_traps.c
+++ b/src/arch/x86/32/c_traps.c
@@ -40,7 +40,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
     NODE_STATE(vcpu->kernelSP = ((word_t)kernel_stack_alloc[getCurrentCPUIndex()]) + BIT(
                                     CONFIG_KERNEL_STACK_BITS) - 4;
-    word_t *guest_regs_ptr = vcpu->gp_registers;
+               word_t *guest_regs_ptr = vcpu->gp_registers;
     if (vcpu->launched) {
     /* attempt to do a vmresume */
     asm volatile(
```

Indanz: Maybe it's confused by the previous wrong NODE_STATE?

`NODE_STATE(vcpu->kernelSP = ((word_t)kernel_stack_alloc[getCurrentCPUIndex()]) + BIT(CONFIG_KERNEL_STACK_BITS) - 4;`

is plain wrong, it should be:

`vcpu->kernelSP = ((word_t)kernel_stack_alloc[getCurrentCPUIndex()]) + BIT(CONFIG_KERNEL_STACK_BITS) - 4;`

I introduced the problem with commit 1bd0e3b78851, oops.

src/arch/x86/64/c_traps.c
dreamliner787-9: I've improved this and removed all the `r_` prefixes.

src/arch/x86/32/c_traps.c
```diff
@@ -36,11 +36,12 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
     /* Do not support breakpoints in VMs, so just disable all breakpoints */
     loadAllDisabledBreakpointState(cur_thread);
 #endif
+    word_t *guest_regs_ptr = vcpu->gp_registers;
     if (vcpu->launched) {
         /* attempt to do a vmresume */
         asm volatile(
             // Set our stack pointer to the top of the tcb so we can efficiently pop
-            "movl %0, %%esp\n"
+            "movl %[guest_stack], %%esp\n"
```
Indanz: I changed my mind about the naming of this one: It's only very temporarily the guest stack: `vmresume` will switch to the configured stack pointer. So it's only used to restore and save the guest registers, so calling it `guest_regs` makes more sense. Same for the 64-bit version.

Indanz (approved):
Looks good otherwise.

PR #1672, opened 2026-05-21
x86,SMP: Fix compile errors

Indanz:
Introduced by the previous commit of PR #1641:

x86,SMP: Fix VMCheckBoundNotification IPI handling

### DIFF
--- a/src/arch/x86/object/vcpu.c
+++ b/src/arch/x86/object/vcpu.c
@@ -1329,7 +1329,7 @@ exception_t handleVmexit(void)
     /* the basic exit reason is the bottom 16 bits of the exit reason field */
     reason = vmread(VMX_DATA_EXIT_REASON) & MASK(16);
     if (reason == EXTERNAL_INTERRUPT) {
-        NODE_LOCK_IRQ();
+        NODE_LOCK_IRQ;
         if (vmx_feature_ack_on_exit) {
             interrupt = vmread(VMX_DATA_EXIT_INTERRUPT_INFO) & 0xff;
             ARCH_NODE_STATE(x86KScurInterrupt) = interrupt;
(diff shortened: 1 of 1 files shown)


Indanz:
I thought the PR passed the CI tests before, but apparently not. Oops.

(And I didn't test it locally because I don't have an x86 VM setup.)

midnightveil (approved):
Ooops...

(The checks *did* pass though. Maybe this isn't in the default set of checks?)

Indanz:
> (The checks _did_ pass though. Maybe this isn't in the default set of checks?)

That would explain it. But I'd expect that we would at least compile all supported configs? Maybe not for each board, but at least one combination for each arch.

lsf37:
> > (The checks _did_ pass though. Maybe this isn't in the default set of checks?)

Yes, most combinations of SMP, MCS, HYP are missing.

> That would explain it. But I'd expect that we would at least compile all supported configs? Maybe not for each board, but at least one combination for each arch.

The way to check that currently is to add the `hw-build` label on the PR.

I agree that it would be nice to catch these by default, but even just the main config combinations are a lot. The matrix for this would be `{release, debug} x {SMP on/off} x {MCS on/off} x {HYP on/off} x {mode 32/64} x {arm, riscv, x86}` = 48 jobs, and that would just be a single default board for each combination. 48 jobs is already over the limit of concurrent jobs (40) we have for the entire org, so a single PR would block all other jobs in the org until those are done. (Ok, it's a few less, since we don't have HYP for RISC-V yet, but still too many).

I guess we could run multiple combinations in a single job which would then take a bit longer, but consume fewer runners. If we only build kernel.elf, that is relatively quick (will not catch problems in libsel4, though). The "compile" job is a bit of an odd one out in the current CI setup, because it was one of Luke's first experiments for GitHub, predating the rest by quite a bit. Maybe it's time to overhaul that one.

PR #1674, opened 2026-05-22
x86_64/c_traps.c: fix compile error

dreamliner787-9:
Fix compile error from https://github.com/seL4/seL4/pull/1667.

### DIFF
--- a/src/arch/x86/64/c_traps.c
+++ b/src/arch/x86/64/c_traps.c
@@ -14,7 +14,7 @@
 #include <api/syscall.h>
 
 #ifdef CONFIG_VTX
-static void NORETURN vmlaunch_failed(word_t failInvalid, word_t failValid)
+static void NORETURN USED vmlaunch_failed(word_t failInvalid, word_t failValid)
 {
     NODE_LOCK_SYS;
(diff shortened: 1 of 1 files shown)


src/arch/x86/64/c_traps.c
```diff
@@ -14,7 +14,7 @@
 #include <api/syscall.h>
 
 #ifdef CONFIG_VTX
-static void NORETURN vmlaunch_failed(word_t failInvalid, word_t failValid)
+static void NORETURN USED vmlaunch_failed(word_t failInvalid, word_t failValid)
```
midnightveil, line 17: Just confirming - I don't know if it's wrong - but what are the semantics of referencing static functions from within inline assembly?

I guess it's fine since the kernel is a single translation unit, but I don't know if this is strictly correct.

Indanz: As it's still the same compilation unit, it shouldn't matter whether the symbol is global or not. In the past, without USED you would make it non-static in the hope the compiler doesn't optimise it away.

Indanz (approved):

PR #1675, opened 2026-05-22
fixes for IPC queue reduction in MCS

lsf37:
- Apply the changes from commit https://github.com/seL4/seL4/commit/771c9e43cef98a5ef7922a37014d4dd0a3de20c3 to the signal fast path, which was missed there.

- The IPC queue changes for MCS in commit https://github.com/seL4/seL4/commit/771c9e43cef98a5ef7922a37014d4dd0a3de20c3 has decreased the TCB size slightly, which brings TCB_SIZE_BITS down for config combinations that have MCS + SMP + HYP + benchmarking enabled from 12 to 11.


Ivan-Velickovic:
This seems to have fixed the compile error in Microkit I saw, thanks.

Indanz (approved):
Hard to review, but yes, after following all the old and new code paths, this looks correct.

PR #1676, opened 2026-05-22
x86,SMP: fix VMCheckBoundNotification call

Ivan-Velickovic:
Missed from https://github.com/seL4/seL4/pull/1641.

I think the issue is that x86-64 is not built with VT-X in any of the CI runs, so we should fix that.

### DIFF
--- a/src/arch/x86/smp/ipi.c
+++ b/src/arch/x86/smp/ipi.c
@@ -50,7 +50,7 @@ void handleRemoteCall(IpiRemoteCall_t call, word_t arg0, word_t arg1, word_t arg
 
             /* For a running VM notifications are already checked by handleVmexit */
             if (tcb != NODE_STATE(ksCurThread)) {
-                VMCheckBoundNotification();
+                VMCheckBoundNotification(tcb);
             }
             break;
 #endif


Ivan-Velickovic:
> I think the issue is that x86-64 is not built with VT-X in any of the CI runs, so we should fix that.

Actually, it looks like I'm wrong, but maybe it's just x86-64 SMP + HYP that's missing.

Ivan-Velickovic:
> Actually, it looks like I'm wrong, but maybe it's just x86-64 SMP + HYP that's missing.

More details here https://github.com/seL4/ci-actions/issues/476

Indanz (approved):

Indanz:
Okay, apparently I shouldn't count on CI catching stupid compile errors. I thought Alessandro Legnani tested my patch, but either they silently fixed the compile errors, or there was miscommunication.

@alegnani, do you remember what happened here? Reference: https://github.com/seL4/seL4/issues/1148#issuecomment-4151154055

alegnani:
@Indanz Looking back at my commits from the fork and it appears I applied the patch and fixed the compile issue without realizing. My apologies, this is on me.

Indanz:
> @Indanz Looking back at my commits from the fork and it appears I applied the patch and fixed the compile issue without realizing. My apologies, this is on me.

No worries, just trying to figure out what went wrong to avoid it happening again in the future. I should have at least compiled it myself once instead of counting on CI and you.

PR #1678, opened 2026-05-28
x86, 32bit, SMP, HYP: Remove unused vcpu.kernelSP

Indanz:
User context already has a kernelSP, which is used. Also, the 64-bit also doesn't has this, so probably a leftover.

GCC 10.2.1 doesn't like declarations inside a case statement without a new context.

Related:
- https://github.com/seL4/ci-actions/pull/482
- https://github.com/seL4/seL4/pull/1669#discussion_r3297712666


midnightveil:
> GCC 10.2.1 doesn't like declarations inside a case statement without a new context.

Until C23 this was invalid C and only allowed as a GCC extension.

midnightveil (approved):

PR #1680, opened 2026-06-01
mcs: fix missing cap fault info in `sendFaultIPC`

### LINKED ISSUE #864: Fault handling behavior question
I have a few question about the fault handling behavior inside the kernel:

- Why is `tptr->tcbLookupFailure` no populated in MCS? `sendFaultIPC()` seems to just set `tptr->tcbFault`?
- Why is a thread fault not stored unconditionally in `tptr->tcbFault` and `tptr->tcbLookupFailure`? This happens only if the check succeeded that a valid fault handler cap is set. MCS just checks it's an endpoint cap, non-MCS also check the endpoint rights.
- Besides the fault IPC message, is there a way in userspace to get the last fault of a thread (ie. the content of `tptr->tcbFault` and `tptr->tcbLookupFailure`) given one has access to the TCB?

michaelmcinerney:
- refactor `validFaultHandler` for readability
- fix missing cap fault info in `sendFaultIPC` (`tptr->tcbLookupFailure` was not being set)

### DIFF
--- a/src/kernel/faulthandler.c
+++ b/src/kernel/faulthandler.c
@@ -31,11 +31,12 @@ void handleTimeout(tcb_t *tptr)
 bool_t sendFaultIPC(tcb_t *tptr, cap_t handlerCap, bool_t can_donate)
 {
     if (cap_get_capType(handlerCap) == cap_endpoint_cap) {
-        assert(cap_endpoint_cap_get_capCanSend(handlerCap));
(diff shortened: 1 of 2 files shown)


michaelmcinerney:
Force pushed to add a body to the first commit.

lsf37 (approved):

Indanz (approved):

PR #1681, opened 2026-06-01
mcs: fix missing cap fault info in lookupReply

michaelmcinerney:
Set missing cap fault info (lookup_fault_missing_capability_new(0)) in lookupReply and move handleFault to caller, so that lookupReply is an actual lookup function without random side effects.


lsf37 (approved):
GitHub should probably not let me approve this since I am a co-author on the commit, but I do think it is correct :-)

midnightveil (approved):

Indanz (approved):

PR #1682, opened 2026-06-01
mcs: add check for preemption to `SysReplyRecv`

michaelmcinerney:
To ease verification, in the SysReplyRecv case of handleSyscall, add the explicit check for preemption, even though the call to handleInvocation should not be preemptible.

### DIFF
--- a/src/api/syscall.c
+++ b/src/api/syscall.c
@@ -594,8 +594,12 @@ exception_t handleSyscall(syscall_t syscall)
         case SysReplyRecv: {
             cptr_t reply = getRegister(NODE_STATE(ksCurThread), replyRegister);
             ret = handleInvocation(false, false, true, true, reply);
-            /* reply cannot error and is not preemptible */
-            assert(ret == EXCEPTION_NONE);
+            /* reply is not preemptible, but to ease verification we check explicitly */
+            if (unlikely(ret != EXCEPTION_NONE)) {
+                mcsPreemptionPoint();
+                checkInterrupt(/* was_interrupt_entry */ false);
+                break;
+            }
             handleRecv(true, true);
             break;
         }


lsf37 (approved):
This would be theoretically provable, but it's a lot of work for something we don't do in other syscalls either, and performance impact should be extremely minimal.

midnightveil (approved):
Just to clarify: is this a temporary state, or is it intended to be permanent? I'd prefer if the comment in the code was a bit more explicit (or if there is something tracking this?)

lsf37:
This is likely permanent unless we think it does actually cost us performance (doesn't affect the fast path, though)

src/api/syscall.c
```diff
@@ -594,8 +594,12 @@ exception_t handleSyscall(syscall_t syscall)
...
             cptr_t reply = getRegister(NODE_STATE(ksCurThread), replyRegister);
             ret = handleInvocation(false, false, true, true, reply);
-            /* reply cannot error and is not preemptible */
-            assert(ret == EXCEPTION_NONE);
+            /* reply is not preemptible, but to ease verification we check explicitly */
+            if (unlikely(ret != EXCEPTION_NONE)) {
+                mcsPreemptionPoint();
+                checkInterrupt(/* was_interrupt_entry */ false);
+                break;
+            }
```
Indanz, line 604:
```suggestion
            /* reply is not preemptible, but to ease verification we check explicitly */
            if (unlikely(ret != EXCEPTION_NONE)) {
                break;
            }
```
Considering it can't happen, no need to do the actual calls.

But the compiler should be smart enough to combine it with the other similar cases above, so code-size wise it shouldn't really matter.

Indanz (approved):

Indanz:
I think these were the checks I wanted to get rid off, but haven't gotten to yet. Need to show with benchmarks that the kernel exit+entry overhead of my approach isn't too bad compared to what we have now.

src/api/syscall.c
lsf37: That doesn’t work: the point of the change is that we currently have *not* proved the fact that preemption doesn’t occur, which means the current state of the proof thinks that it is possible and the calls in that case do have to happen when it occurs.

Indanz: Ah, to make the code equal to the kernel entry path. Okay, never mind then.

lsf37:
> I think these were the checks I wanted to get rid off, but haven't gotten to yet. Need to show with benchmarks that the kernel exit+entry overhead of my approach isn't too bad compared to what we have now.

The exit-kernel-on-preemption approach would indeed get rid of all of these. There's still a bunch to do on getting the MCS proofs in sync with the rest, but with this set of PRs, the main functional correctness proof for MCS RISC-V is now finished. We'll be looking at AArch64 next, but it's not entirely impossible that we might come back to this condition and prove that it's not necessary, but it's currently not the highest priority.

PR #1683, opened 2026-06-01
pc99/acpi: improve MADT parsing

### LINKED ISSUE #396: Unexpected behaviour on x86-64 when building SDK using LLVM

dreamliner787-9:
Depends on https://github.com/seL4/seL4/pull/1665.

- pc99/acpi: validate BIOS provided MADT entry size
Previously the kernel trusted the ACPI MADT entries length provided by
the BIOS. This commit added checks to make sure that they are correct
before any parsing steps happen.

- pc99/acpi: align access for each MADT entry
The struct types for the MADT entries have alignment requirements set by
the compiler. But since these entries are tightly packed by the BIOS,
entry `n-1` can cause entry `n` to be unaligned. So if you compile the
kernel with LLVM and UBSAN on, the kernel can crash depending on how
your BIOS packed the MADT.
This commit updated the code so that for each entry the kernel copies
it into a stack allocated buffer to respect the compiler alignment
requirements.

### DIFF
--- a/src/plat/pc99/machine/acpi.c
+++ b/src/plat/pc99/machine/acpi.c
@@ -151,6 +151,12 @@ typedef struct acpi_madt_iso {
 compile_assert(acpi_madt_iso_packed,
                sizeof(acpi_madt_iso_t) == sizeof(acpi_madt_header_t) + 8)
 
+#define MADT_MAX_ENTRY_SIZE sizeof(acpi_madt_x2apic_t)
+
+compile_assert(acpi_madt_apic_entry_size, sizeof(acpi_madt_apic_t) <= MADT_MAX_ENTRY_SIZE)
+compile_assert(acpi_madt_ioapic_entry_size, sizeof(acpi_madt_ioapic_t) <= MADT_MAX_ENTRY_SIZE)
+compile_assert(acpi_madt_iso_entry_size, sizeof(acpi_madt_iso_t) <= MADT_MAX_ENTRY_SIZE)
+
 /* workaround because string literals are not supported by C parser */
 const char acpi_str_rsd[]  = {'R', 'S', 'D', ' ', 'P', 'T', 'R', ' ', 0};
 const char acpi_str_fadt[] = {'F', 'A', 'C', 'P', 0};
@@ -160,6 +166,10 @@ const char madt_str_apic[] = {'A', 'P', 'I', 'C', 0};
 const char madt_str_x2apic[] = {'x', '2', 'A', 'P', 'I', 'C', 0};
 const char madt_str_ioapic[] = {'I', 'O', 'A', 'P', 'I', 'C', 0};
 const char madt_str_iso[] = {'I', 'S', 'O', 0};
+
+/* aligned(4) so that we can copy the BIOS provided MADT entries into this buffer
+ * then cast it to it's proper type then dereference it in an aligned manner. */
+BOOT_DATA char madt_entry[MADT_MAX_ENTRY_SIZE] ALIGN(4);
 
 BOOT_CODE static uint8_t acpi_calc_checksum(char *start, uint32_t length)
 {
@@ -291,7 +301,8 @@ BOOT_CODE uint32_t acpi_madt_scan(
     paddr_t     *ioapic_paddrs
 )
 {
-    unsigned int entries;
+    unsigned int        entries;
+    unsigned int        entry_size;
     uint32_t            num_cpu;
     uint32_t            count;
     acpi_madt_t        *acpi_madt;
@@ -299,6 +310,7 @@ BOOT_CODE uint32_t acpi_madt_scan(
 
     acpi_rsdt_t *acpi_rsdt_mapped;
     acpi_madt_t *acpi_madt_mapped;
+
     acpi_rsdt_mapped = (acpi_rsdt_t *)acpi_table_init((acpi_rsdt_t *)(word_t)acpi_rsdp->rsdt_address, ACPI_RSDT);
 
     num_cpu = 0;
@@ -320,6 +332,9 @@ BOOT_CODE uint32_t acpi_madt_scan(
             acpi_madt_header = (acpi_madt_header_t *)(acpi_madt_mapped + 1);
 
             while ((char *)acpi_madt_header < (char *)acpi_madt_mapped + acpi_madt_mapped->header.length) {
+                entry_size = MIN(acpi_madt_header->length, MADT_MAX_ENTRY_SIZE);
+                memcpy(madt_entry, acpi_madt_header, entry_size);
+
                 switch (acpi_madt_header->type) {
                 /* ACPI specifies the following rules when listing APIC IDs:
                  *  - Boot processor is listed first
@@ -332,9 +347,11 @@ BOOT_CODE uint32_t acpi_madt_scan(
                     if (!acpi_madt_valid(acpi_madt_header, madt_str_apic, sizeof(acpi_madt_apic_t))) {
                         break;
                     }
+                    acpi_madt_apic_t *madt_apic = (acpi_madt_apic_t *) madt_entry;
+
                     /* what Intel calls apic_id is what is called cpu_id in seL4! */
-                    uint8_t  cpu_id = ((acpi_madt_apic_t *)acpi_madt_header)->apic_id;
-                    uint32_t flags  = ((acpi_madt_apic_t *)acpi_madt_header)->flags;
+                    uint8_t  cpu_id = madt_apic->apic_id;
+                    uint32_t flags  = madt_apic->flags;
                     if (flags == 1) {
                         printf("ACPI: MADT_APIC apic_id=0x%x\n", cpu_id);
                         if (num_cpu == CONFIG_MAX_NUM_NODES) {
@@ -350,8 +367,10 @@ BOOT_CODE uint32_t acpi_madt_scan(
                     if (!acpi_madt_valid(acpi_madt_header, madt_str_x2apic, sizeof(acpi_madt_x2apic_t))) {
                         break;
                     }
-                    uint32_t cpu_id = ((acpi_madt_x2apic_t *)acpi_madt_header)->x2apic_id;
-                    uint32_t flags  = ((acpi_madt_x2apic_t *)acpi_madt_header)->flags;
+                    acpi_madt_x2apic_t *madt_x2apic = (acpi_madt_x2apic_t *) madt_entry;
+
+                    uint32_t cpu_id = madt_x2apic->x2apic_id;
+                    uint32_t flags  = madt_x2apic->flags;
                     if (flags == 1) {
                         printf("ACPI: MADT_x2APIC apic_id=0x%x\n", cpu_id);
                         if (num_cpu == CONFIG_MAX_NUM_NODES) {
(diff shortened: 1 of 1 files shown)


src/plat/pc99/machine/acpi.c
```diff
@@ -315,6 +315,12 @@ BOOT_CODE uint32_t acpi_madt_scan(
                  *  - APIC IDs < 0xFF should be listed in APIC subtable, APIC IDs >= 0xFF
                  *    should be listed in X2APIC subtable */
                 case MADT_APIC: {
+                    if (acpi_madt_header->length != sizeof(acpi_madt_apic_t)) {
```
Indanz: This should be a `acpi_madt_header->length >= sizeof(acpi_madt_apic_t)` check. It's perfectly sane to pad the whole to 4-byte alignment or to extend structures in the future. You have to check your assumptions here, without being too strict as you have no idea what you will encounter in the field.

src/plat/pc99/machine/acpi.c
```diff
@@ -315,6 +315,12 @@ BOOT_CODE uint32_t acpi_madt_scan(
...
+                        printf("ACPI: BIOS bug, MADT APIC entry size %u != %zu expected bytes. Skipping.\n", acpi_madt_header->length,
```
Indanz: Once it doesn't fit on one line, It's conventional to move all parameters to the next line so it's easier to see what's what. (The exception being if that would make the next line too long.)

src/plat/pc99/machine/acpi.c
dreamliner787-9: I think you meant `acpi_madt_header->length < sizeof(acpi_madt_apic_t)`?

dreamliner787-9:
I've folded https://github.com/seL4/seL4/pull/1684 's changes into this PR.

src/plat/pc99/machine/acpi.c
Indanz: Yes, sorry, got confused by the negation.

src/plat/pc99/machine/acpi.c
```diff
@@ -315,6 +315,13 @@ BOOT_CODE uint32_t acpi_madt_scan(
                  *  - APIC IDs < 0xFF should be listed in APIC subtable, APIC IDs >= 0xFF
                  *    should be listed in X2APIC subtable */
                 case MADT_APIC: {
+                    if (acpi_madt_header->length < sizeof(acpi_madt_apic_t)) {
+                        printf("ACPI: BIOS bug, MADT APIC entry size %u < %zu expected bytes. Skipping.\n",
+                               acpi_madt_header->length,
+                               sizeof(acpi_madt_apic_t));
```
Indanz: Why the arguments on separate lines? That looks weird. Please put them next to each other.

src/plat/pc99/machine/acpi.c
```diff
@@ -322,9 +337,11 @@ BOOT_CODE uint32_t acpi_madt_scan(
                         break;
                     }
 
+                    acpi_madt_apic_t *madt_apic_entry = (acpi_madt_apic_t *) madt_entry;
```
Indanz: Style checker probably disagrees with me, but I would do:

`                    acpi_madt_apic_t *madt_apic_entry = (acpi_madt_apic_t*)madt_entry;`

All the extra spaces look weird to me, makes it harder to spot that it's a cast.

src/plat/pc99/machine/acpi.c
```diff
@@ -358,6 +379,13 @@ BOOT_CODE uint32_t acpi_madt_scan(
                     }
                     break;
                 case MADT_ISO: {
+                    if (acpi_madt_header->length < sizeof(acpi_madt_iso_t)) {
+                        printf("ACPI: BIOS bug, MADT ISO entry size %u < %zu expected bytes. Skipping.\n",
+                               acpi_madt_header->length,
+                               sizeof(acpi_madt_iso_t));
+                        break;
+                    }
```
Indanz: Personally, I would add a small helper function and call it everywhere:

```c
static bool_t acpi_madt_valid(acpi_madt_t* header, const char* name, size_t expected)
{
    if (header->length < expected) {
        printf("ACPI: BIOS bug, MADT %s entry size %u < %zu expected bytes.\n",
               name, header->length, expected);
        return false;
    }
    return true;
}


if (!acpi_madt_valid(acpi_madt_header, "ISO", sizeof(acpi_madt_iso_t))) {
    break;
}
```

Indanz (approved):
About empty lines: No need to add random new lines, only add them if they have purpose. E.g. a line with just a closing brace on it really doesn't need another empty line after it. You do want to add an empty line after variable declarations though.

About naming: No need to name everything `acpi_madt_*_entry`, it adds nothing and calling it just `acpi_madt_*` is sufficient.

Other than nitpicking, it's fine.

Indanz:
(The style checker doesn't like my personal `*` style. That's an on-going battle between me, C and everyone else. I think it should obviously be part of the type, hence not stuck to the name, but oh well.)

src/plat/pc99/machine/acpi.c
```diff
@@ -345,26 +379,34 @@ BOOT_CODE uint32_t acpi_madt_scan(
...
-                    printf("ACPI: MADT_ISO bus=%d source=%d gsi=%d flags=0x%x\n",
-                           ((acpi_madt_iso_t *)acpi_madt_header)->bus,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->source,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->gsi,
-                           ((acpi_madt_iso_t *)acpi_madt_header)->flags);
+                case MADT_ISO: {
+                    if (!acpi_madt_valid(acpi_madt_header, "ISO", sizeof(acpi_madt_iso_t))) {
+                        break;
+                    }
+                    UNUSED acpi_madt_iso_t *iso = (acpi_madt_iso_t *) madt_entry;
```
Indanz, line 408: I think this UNUSED isn't needed any more, as it's not needed for the others.

dreamliner787-9: It is needed, because I think in the verified configuration, the printf statements aren't compiled so in that case the compiler will complains about unused variables.

midnightveil: If the only reason we parse it is to print it why do it at all?

dreamliner787-9: I don't want to cast a char * to a acpi_madt_iso_t * 5 times in the print statement

midnightveil: No, why do the print statement at all?

Indanz: That's not quite true: `printf()` is defined as:
```
#define printf(...)             ((void)(0))
```
Which does consume the input parameters, so should suppress unused warnings.

(If it really is needed, you should also add it to the other places. Now it's inconsistent.)

> No, why do the print statement at all?

Or more broadly, why parse it at all if we don't use it? Does it help with debugging anything or something?

dreamliner787-9: It is useful for development and debugging since it tells you the quirks about how your platform wired up the interrupts. From the ACPI spec:

> Interrupt Source Overrides are necessary to describe variances between the IA-PC standard dual 8259 interrupt definition and the platform’s implementation.
It is assumed that the ISA interrupts will be identity-mapped into the first I/O APIC sources. Most existing APIC
designs, however, will contain at least one exception to this assumption. The Interrupt Source Override Structure is
provided in order to describe these exceptions. It is not necessary to provide an Interrupt Source Override for every
ISA interrupt. Only those that are not identity-mapped onto the APIC interrupt inputs need be described.
• This specification only supports overriding ISA interrupt sources.
For example, if your machine has the ISA Programmable Interrupt Timer (PIT) connected to ISA IRQ 0, but in APIC
mode, it is connected to I/O APIC interrupt input 2, then you would need an Interrupt Source Override where the
source entry is ‘0’ and the Global System Interrupt is ‘2.’

While userspace can map then parses the MADT themselves, it is useful for the kernel to print it out to the user since the parsing work had been done.

midnightveil:
Was there anything still blocking for this?

dreamliner787-9:
> Was there anything still blocking for this?

None on my side

lsf37 (approved):
Looks good from my side, and I don't expect any proof breakage, but I've started a proof check to be sure. If that passed, this is ready to merge.

midnightveil:
X86 proofs failed with:

```
CKernel FAILED (see also "isabelle build_log -H Error CKernel")
*** exception Interrupt_Breakdown raised
*** /home/test-runner/ver/seL4/src/plat/pc99/machine/acpi.c:332.23-332.32: Attempting to cause decay to pointer on array without an address
*** 
*** At command "install_C_file" (line 94 of "~/ver/l4v/spec/cspec/X64/Kernel_C.thy")
Unfinished session(s): CKernel

```

At a guess; would this be because ` char madt_entry[MADT_MAX_ENTRY_SIZE] ALIGN(4);` is a stack variable, and the verification subset does not permit that? Or is it because it is missing `&madt_entry` in the memcpy?

lsf37:
> ```
> *** /home/test-runner/ver/seL4/src/plat/pc99/machine/acpi.c:332.23-332.32: Attempting to cause decay to pointer on array without an address
> ```
> 
> At a guess; would this be because ` char madt_entry[MADT_MAX_ENTRY_SIZE] ALIGN(4);` is a stack variable, and the verification subset does not permit that? Or is it because it is missing `&madt_entry` in the memcpy?

It is because it is a stack variable, yes. The verification subset does not allow pointers into the stack, and that's exactly what the `memcpy` line is trying to do (passing a pointer to the array on the stack).

Would there be interference if we made that array global instead? It's a bit weird to do that in terms of C, but that would work if the function is not recursive or otherwise multiple live versions of it would exist.

midnightveil:
I think it should be fine; I don't believe it's recursive.

Indanz:
> Would there be interference if we made that array global instead? It's a bit weird to do that in terms of C, but that would work if the function is not recursive or otherwise multiple live versions of it would exist.

It's also fine on SMP, as this code is used to figure out how many CPU cores there are, so only one caller on one CPU core.

Indanz:
Alternative would be to put the buffer in a struct and use a struct assignment instead of an explicit memcpy.

dreamliner787-9:
I've changed the buffer to be a global: https://github.com/seL4/seL4/compare/7006c3c5e862056d54efcb6af46aa4625d44f5df..e2f87779f0fed133d5be7a425186a9d3bbd7d3e3?diff=unified&w

src/plat/pc99/machine/acpi.c
```diff
@@ -315,9 +340,14 @@ BOOT_CODE uint32_t acpi_madt_scan(
                  *  - APIC IDs < 0xFF should be listed in APIC subtable, APIC IDs >= 0xFF
                  *    should be listed in X2APIC subtable */
                 case MADT_APIC: {
+                    if (!acpi_madt_valid(acpi_madt_header, "APIC", sizeof(acpi_madt_apic_t))) {
```
lsf37: The proofs are complaining about the "APIC":

> *** /home/test-runner/ver/seL4/src/plat/pc99/machine/acpi.c:343.59-343.65: Don't want to handle string literals!

I need to check how we do string literals elsewhere. printf is compiled out for verification, so they don't happen a lot, but it's possible that the CParser is fine with string literals assigned to globals.

lsf37: Ok, I checked, they are not supported at all. Best you can do is a global explicit `char[]` with the single letters, null terminated.

dreamliner787-9: Ok sounds good I'll change it.

dreamliner787-9: Could you please try again?

midnightveil: > I need to check how we do string literals elsewhere

Yeah, it's done via a global char[] array elsewhere.

> Could you please try again?

Proofs are rerunning on every push because of the proof-test label. You can look at the x64 job to see if it passes.

midnightveil: Question for @lsf37: is there a particular reason why the preprocess jobs don't catch this? I guess the preprocess check just does the AST, and not the whole post-processing /cleanup of the AST which a theory file will do?

lsf37: Yes, the preprocess test is just supposed to flag if something changed. It did that and I looked manually, missing all of these :-)

The errors here only seem to happen at the stage where it produces semantics, so the "C parser" job that just checks syntax doesn't fail.

midnightveil (approved):
It looks like the X64 proofs have passed now, so this should be good to go.

lsf37:
Yup, this is fine now. We really should add support for at least string literal globals at some point..

PR #1684, opened 2026-06-01
pc99/acpi: align access for each MADT entry

dreamliner787-9:
The struct types for the MADT entries have alignment requirements set by the compiler. But since these entries are tightly packed by the BIOS, entry n-1 can cause the n entry to be unaligned. So if you compile the kernel with LLVM and UBSAN on, the kernel can crash depending on how your BIOS packed the MADT.

I was able to reliably reproduce UBSAN crash from this issue on real x86 hardware we have in the lab.

This commit changed the code so that for each MADT entry the kernel copies it into a stack allocated variable to respect the compiler alignment requirements.

Relates to, but does not depends on https://github.com/seL4/seL4/pull/1665.

### DIFF
--- a/src/plat/pc99/machine/acpi.c
+++ b/src/plat/pc99/machine/acpi.c
@@ -152,6 +152,15 @@ typedef struct acpi_madt_iso {
 unverified_compile_assert(acpi_madt_iso_packed,
                           OFFSETOF(acpi_madt_iso_t, flags) == sizeof(acpi_madt_header_t) + 6)
 
+#define MAX_MADT_ENTRY_SIZE sizeof(acpi_madt_x2apic_t)
+
+compile_assert(acpi_madt_apic_within_max_entry_size,
+               sizeof(acpi_madt_apic_t) <= MAX_MADT_ENTRY_SIZE)
+compile_assert(acpi_madt_ioapic_within_max_entry_size,
+               sizeof(acpi_madt_ioapic_t) <= MAX_MADT_ENTRY_SIZE)
+compile_assert(acpi_madt_iso_within_max_entry_size,
+               sizeof(acpi_madt_iso_t) <= MAX_MADT_ENTRY_SIZE)
+
 /* workaround because string literals are not supported by C parser */
 const char acpi_str_rsd[]  = {'R', 'S', 'D', ' ', 'P', 'T', 'R', ' ', 0};
 const char acpi_str_fadt[] = {'F', 'A', 'C', 'P', 0};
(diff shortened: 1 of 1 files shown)


src/plat/pc99/machine/acpi.c
```diff
@@ -152,6 +152,15 @@ typedef struct acpi_madt_iso {
...
                           OFFSETOF(acpi_madt_iso_t, flags) == sizeof(acpi_madt_header_t) + 6)
 
+#define MAX_MADT_ENTRY_SIZE sizeof(acpi_madt_x2apic_t)
+
+compile_assert(acpi_madt_apic_within_max_entry_size,
+               sizeof(acpi_madt_apic_t) <= MAX_MADT_ENTRY_SIZE)
+compile_assert(acpi_madt_ioapic_within_max_entry_size,
+               sizeof(acpi_madt_ioapic_t) <= MAX_MADT_ENTRY_SIZE)
+compile_assert(acpi_madt_iso_within_max_entry_size,
+               sizeof(acpi_madt_iso_t) <= MAX_MADT_ENTRY_SIZE)
```
Indanz, line 162: These can be on one line if you give them less verbose names...

src/plat/pc99/machine/acpi.c
```diff
@@ -286,6 +295,11 @@ BOOT_CODE uint32_t acpi_madt_scan(
 
     acpi_rsdt_t *acpi_rsdt_mapped;
     acpi_madt_t *acpi_madt_mapped;
+
+    /* aligned(4) so that we can copy the BIOS provided MADT entries into this buffer
+     * then cast it to it's proper type then dereference it in an aligned manner. */
+    char madt_entry[MAX_MADT_ENTRY_SIZE] __attribute__((aligned(4)));
```
Indanz, line 301: We have an `ALIGN()` macro.

src/plat/pc99/machine/acpi.c
```diff
@@ -304,9 +318,21 @@ BOOT_CODE uint32_t acpi_madt_scan(
             printf("ACPI: MADT apic_addr=0x%x\n", acpi_madt_mapped->apic_addr);
             printf("ACPI: MADT flags=0x%x\n", acpi_madt_mapped->flags);
 
-            acpi_madt_header = (acpi_madt_header_t *)(acpi_madt_mapped + 1);
+            for (
+                acpi_madt_header = (acpi_madt_header_t *)(acpi_madt_mapped + 1);
+                (char *)acpi_madt_header < (char *)acpi_madt_mapped + acpi_madt_mapped->header.length;
+                acpi_madt_header = (acpi_madt_header_t *)((char *)acpi_madt_header + acpi_madt_header->length)
```
Indanz, line 324: This is horrible, please don't try to stuff all that into a for-loop.

src/plat/pc99/machine/acpi.c
```diff
@@ -304,9 +318,21 @@ BOOT_CODE uint32_t acpi_madt_scan(
...
+            ) {
+                if (acpi_madt_header->length <= MAX_MADT_ENTRY_SIZE) {
+                    memcpy(madt_entry, acpi_madt_header, acpi_madt_header->length);
+                } else {
+                    printf("ACPI: Skipping MADT entry of type 0x%x, length %u > max %zu bytes\n",
+                           acpi_madt_header->type,
+                           acpi_madt_header->length,
+                           MAX_MADT_ENTRY_SIZE);
+                    continue;
+                }
```
Indanz, line 334: If you use `MAX(acpi_madt_header->length, madt_entry)`, then you don't need this awkward extra check. And if a known entry is bigger than expected, you can still parse the fields you know and care about.

Indanz (commented):
This really should be combined with #1683, as it would avoid a horrible merge.

dreamliner787-9:
> This really should be combined with https://github.com/seL4/seL4/pull/1683, as it would avoid a horrible merge.

Sure I will fold this PR changes into the other PR and apply your suggestions.

PR #1687, opened 2026-06-05
arm,hyp: default KernelArmVtimerUpdateVOffset OFF

lsf37:
Set the default for `KernelArmVtimerUpdateVOffset` to `OFF`. This is the setting supported in verification builds, and it is also the setting used in most projects.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,8 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 * change default for `KernelArmHypEnableVCPUCP14SaveAndRestore` to `OFF` to provide equal defaults for
   verification and release builds. This affects AArch32/hyp only.
+* change default for `KernelArmVtimerUpdateVOffset` to `OFF`, which is the setting supported in
+  verification builds and used in most projects.
 
 ### Upgrade Notes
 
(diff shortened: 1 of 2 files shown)


lsf37:
This still needs a corresponding l4v PR before it can be merged.

Indanz (approved):
Great, this would restore some sanity on the VM side too, making it practically useful by default.

Only question I have is when is `DEFAULT_DISABLED` needed and do we need it here too?

lsf37:
> Great, this would restore some sanity on the VM side too, making it practically useful by default.
> 
> Only question I have is when is `DEFAULT_DISABLED` needed and do we need it here too?

`DEFAULT_DISABLED` is the value that is assigned when unsatisfied dependencies disable the option. If not set explicitly, `DEFAULT_DISABLED = DEFAULT`. In this case, if it's not a hyp build it doesn't matter. If it's disabled, because it is a verification build, OFF is the correct value.

Indanz:
> `DEFAULT_DISABLED` is the value that is assigned when unsatisfied dependencies disable the option. If not set explicitly, `DEFAULT_DISABLED = DEFAULT`. In this case, if it's not a hyp build it doesn't matter. If it's disabled, because it is a verification build, OFF is the correct value.

Okay, that's what I expected, but actual usage in the config files is so inconsistent that I wasn't sure any more.

Issue #1590, opened 2026-02-09
Please remove dependency on future for `sel4-deps`


micwoj92:
Hello,
python-future is python2/3 compatibility library. sel4-deps uses `python_requires='>=3'` , so it does not make sense to use it.

https://github.com/seL4/seL4/blob/c55f50b6e5e5c753ec9295d85ed5dff91a1e1562/tools/python-deps/setup.py#L29

lsf37:
You're right, that's a left-over from python2 times.

Ivan-Velickovic:
I think the same is true for the `six` dependency. I also don't see any use of `guardonce`?

midnightveil:
guardonce is used in style-c.sh from seL4_tools.

same with future; it's not used in seL4 itself but the other projects under the seL4 org.

Ivan-Velickovic:
Right, forgot that `sel4-deps` includes dependencies for other projects.

midnightveil:
Once #1688 is merged, future cannot be removed until we remove it from these are three uses: https://github.com/search?q=org%3AseL4+%22from+future%22+language%3APython&type=code. I don't see any other occurrences of `import future`, `import past`, or `from past import`.

PR #1688, opened 2026-06-10
tools/reciprocal.py: Remove dependency on past module (#1590)

timretout:
This script uses python3, so the 'range' function is memory-efficient.


Indanz (approved):

Indanz:
Gitlint fails with:
> 1: T1 Title exceeds max length (61>50): "tools/reciprocal.py: Remove dependency on past module (#1590)"

If you fix that, it's ready to go.

PR #1689, opened 2026-06-11
mcs: add cmake file for verified AARCH64 MCS

michaelmcinerney:
To build the verified configuration of AARCH64 MCS.

### DIFF
--- a/configs/AARCH64_MCS_verified.cmake
+++ b/configs/AARCH64_MCS_verified.cmake
@@ -1,0 +1,12 @@
+#!/usr/bin/env -S cmake -P
+#
+# Copyright 2024, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: GPL-2.0-only
+#
+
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Looks good. Can you please add a newline at the last line so the style checker doesn't complain?

michaelmcinerney:
Force pushed to add newline at the end of the file.

Indanz (approved):
Not for this PR, but can we please change the default `KernelStaticMaxPeriodUs` to what verification assumes? And also change the define of `MAX_RELEASE_TIME` to not depend on `MAX_PERIOD_US`?

lsf37:
> Not for this PR, but can we please change the default `KernelStaticMaxPeriodUs` to what verification assumes? And also change the define of `MAX_RELEASE_TIME` to not depend on `MAX_PERIOD_US`?

Yes, I'd be fine with both of these. I do want to make the verification-assumed value more flexible/configurable so that it doesn't matter too much what it is, but that's still on the todo list for now.

PR #1690, opened 2026-06-12
cmake: set policy CMP0174 when it exists

### LINKED ISSUE #1686: CMake policy CMP0174 warnings
CMake >= 3.31 warns about uses of `config_option` and `config_string` where `DEFAULT` or `DEFAULT_DISABLED` get passed an empty string.

The warning is because the empty string leads to the variable being unset instead of being set to empty. Our uses inside `config_option` and `config_string` are all safe (they do string comparison on `"${CONFIG_DEFAUL}"` etc.).

Should we check for CMake version and set that policy for the versions where it exists to suppress the warning, or should we change the functions?

We could change `cmake_parse_arguments(PARSE_ARGV 3 "CONFIG" "" "DEPENDS;DEFAULT_DISABLED;DEFAULT" "")` to take multi-value args instead of single-value args, which makes empty strings fine. The downside is that we then should check the multi-argument lists for > 1 arguments to catch errors where people have forgotten a keyword.

I don't really like all that version checking everywhere, but the additional argument checking is probably a bit noisy.

Any votes/opinions either way?

CMake file:
- https://github.com/seL4/seL4/blob/93f73a49ad2bafd8769f21ede552860e4a6c80aa/tools/helpers.cmake#L272-L279

CMake manual:
- https://cmake.org/cmake/help/v4.3/command/cmake_parse_arguments.html
- https://cmake.org/cmake/help/v4.3/policy/CMP0174.html#policy:CMP0174

lsf37:
Avoid warning about empty string arguments to functions `config_option()` and `config_string()` in newer cmake versions.

Both behaviour versions (with and without policy) are safe for how we are using those arguments. Setting the policy when it exists removes the warning.

Implements #1686 

CMake docs:
https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#transition-schedule
https://cmake.org/cmake/help/v4.3/policy/CMP0174.html#policy:CMP0174

### DIFF
--- a/tools/helpers.cmake
+++ b/tools/helpers.cmake
@@ -10,22 +10,19 @@ cmake_minimum_required(VERSION 3.16.0)
 # (with and without policy) are safe for our uses. If no policy is set, we get a warning for
 # function config_option below when it is called with empty strings.
 if(POLICY CMP0174)
-    cmake_policy(SET CMP0174 NEW)
+  cmake_policy(SET CMP0174 NEW)
 endif()
 
 # Wrapper function around find_file that generates a fatal error if it isn't found
 # Is equivalent to find_file except that it adds CMAKE_CURRENT_SOURCE_DIR as a path and sets
 # CMAKE_FIND_ROOT_PATH_BOTH
 function(RequireFile config_name file_name)
-    find_file(
-        ${config_name} "${file_name}"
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

Indanz (commented):
Annoying that I can't approve stuff after it's merged. Oh, well.

PR #1692, opened 2026-06-14
github: use new build matrix

lsf37:
Use the new build matrix from ci-actions that supports build subgroups for armv8a. This should speed up the hw-build from >50min to about 12:30min, now bounded by the riscv group, not armv8a.

See also https://github.com/seL4/ci-actions/pull/494


lsf37:
(added the hw-test label to check the central sel4test workflow as well)

Indanz (approved):

PR #1695, opened 2026-06-24
libsel4: add new seL4_UserVSpaceTop constant

### LINKED ISSUE #1693: seL4_UserTop inclusive vs exclusive is inconsistent across platforms
I swear I've made an issue for this before, but I can't find it. I have this [old GIST of mine which surveys all the uses and finds it is mostly exclusive](https://gist.github.com/midnightveil/866b8a4e22f6201f7d18a98743f59b20).

Most platforms, `seL4_UserTop` is consistently an exclusive value. Except for on AArch64, where it is inclusive, that is, it is either `0x00ffffffffff` or `0x0fffffffffff` or `0x7fffffffffff`.

This means you can't use a consistent `if (vaddr > UserTop)` or `if (vaddr >= UserTop)` check, but need to make it different cross-architecture and make assumptions about them.

It would be useful for this to be consistent.

midnightveil:
This is a consistent, inclusive-top value that is conveying similar information to that of seL4_UserTop. We are adding this so as to not break existing userspace, but it will make it easier for userspace to deal with other platforms.

Fixes #1693.

Test with: https://github.com/seL4/seL4_libs/pull/112

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,7 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
   the highest available address available to user level. The existing constant
   `seL4_UserTop` is a mix of inclusive and exclusive across existing platforms, however,
   changing this to be consistent would break userspace and require proof updates.
+  The existing `seL4_UserTop` define is deprecated.
 
 ### Platforms
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,10 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 * Add `seL4_DebugGetThreadAffinity()` syscall to return the current affinity of a thread
   in SMP configurations.
+* Add new libsel4 constant `seL4_UserVSpaceTop` which is a *inclusive* value representing
+  the highest available address available to user level. The existing constant
+  `seL4_UserTop` is a mix of inclusive and exclusive across existing platforms, however,
+  changing this to be consistent would break userspace and require proof updates.
 
 ### Platforms
 
--- a/libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
@@ -128,6 +128,9 @@ typedef enum {
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0x0000003ffffff000
 
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop (seL4_UserTop - 1)
+
 #ifdef CONFIG_ENABLE_BENCHMARKS
 /* size of kernel log buffer in bytes */
 #define seL4_LogBufferSize (LIBSEL4_BIT(20))
--- a/libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
@@ -125,7 +125,7 @@ typedef enum {
 #endif
 #endif /* __ASSEMBLER__ */
 
-/* First address in the virtual address space that is not accessible to user level */
+/* (Deprecated) First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0x0000003ffffff000
 
 /* Last address in the virtual address space that is accessible to user level */
--- a/libsel4/sel4_plat_include/allwinnerA20/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/allwinnerA20/sel4/plat/api/constants.h
@@ -9,7 +9,7 @@
 #include <sel4/config.h>
 #include <sel4/arch/constants_cortex_a7.h>
 
-/* First address in the virtual address space that is not accessible to user level */
+/* (Deprecated) First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0xa0000000
 
 /* Last address in the virtual address space that is accessible to user level */
--- a/libsel4/sel4_plat_include/allwinnerA20/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/allwinnerA20/sel4/plat/api/constants.h
@@ -11,3 +11,6 @@
 
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0xa0000000
+
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop (seL4_UserTop - 1)
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
@@ -243,7 +243,7 @@ SEL4_SIZE_SANITY(seL4_VSpaceEntryBits, seL4_VSpaceIndexBits, seL4_VSpaceBits);
  * Anything address above the range above triggers an
  * address size fault.
  */
-/* First address in the virtual address space that is not accessible to user level */
+/* Last address in the virtual address space that is accessible to user level */
 #if defined(CONFIG_ARM_PA_SIZE_BITS_44)
 #define seL4_UserTop 0x00000fffffffffff
 #elif defined(CONFIG_ARM_PA_SIZE_BITS_40)
@@ -253,6 +253,9 @@ SEL4_SIZE_SANITY(seL4_VSpaceEntryBits, seL4_VSpaceIndexBits, seL4_VSpaceBits);
 #endif
 
 #else
-/* First address in the virtual address space that is not accessible to user level */
+/* Last address in the virtual address space that is accessible to user level */
 #define seL4_UserTop 0x00007fffffffffff
 #endif
+
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop seL4_UserTop
--- a/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h
@@ -243,7 +243,7 @@ SEL4_SIZE_SANITY(seL4_VSpaceEntryBits, seL4_VSpaceIndexBits, seL4_VSpaceBits);
  * Anything address above the range above triggers an
  * address size fault.
  */
-/* Last address in the virtual address space that is accessible to user level */
+/* (Deprecated) Last address in the virtual address space that is accessible to user level */
 #if defined(CONFIG_ARM_PA_SIZE_BITS_44)
 #define seL4_UserTop 0x00000fffffffffff
 #elif defined(CONFIG_ARM_PA_SIZE_BITS_40)
@@ -253,7 +253,7 @@ SEL4_SIZE_SANITY(seL4_VSpaceEntryBits, seL4_VSpaceIndexBits, seL4_VSpaceBits);
 #endif
 
 #else
-/* Last address in the virtual address space that is accessible to user level */
+/* (Deprecated) Last address in the virtual address space that is accessible to user level */
 #define seL4_UserTop 0x00007fffffffffff
 #endif
 
--- a/libsel4/sel4_arch_include/ia32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/ia32/sel4/sel4_arch/constants.h
@@ -145,3 +145,6 @@ typedef enum {
 
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0xe0000000
+
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop (seL4_UserTop - 1)
--- a/libsel4/sel4_arch_include/ia32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/ia32/sel4/sel4_arch/constants.h
@@ -143,7 +143,7 @@ typedef enum {
 /* IPC buffer is 512 bytes, giving size bits of 9 */
 #define seL4_IPCBufferSizeBits 9
 
-/* First address in the virtual address space that is not accessible to user level */
+/* (Deprecated) First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0xe0000000
 
 /* Last address in the virtual address space that is accessible to user level */
--- a/libsel4/sel4_arch_include/riscv32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/riscv32/sel4/sel4_arch/constants.h
@@ -127,6 +127,9 @@ typedef enum {
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0x80000000lu
 
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop (seL4_UserTop - 1)
+
 #ifdef CONFIG_ENABLE_BENCHMARKS
 /* size of kernel log buffer in bytes */
 #define seL4_LogBufferSize (LIBSEL4_BIT(20))
--- a/libsel4/sel4_arch_include/riscv32/sel4/sel4_arch/constants.h
+++ b/libsel4/sel4_arch_include/riscv32/sel4/sel4_arch/constants.h
@@ -124,7 +124,7 @@ typedef enum {
 #endif
 #endif /* __ASSEMBLER__ */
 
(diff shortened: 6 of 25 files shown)


midnightveil:
I'm not sure if it's worth casting these values to explicitly be a `seL4_Word`. There's this comment in the seL4 code that complained about them being inconsistent for UserTop.

https://github.com/seL4/seL4/blob/ae11cf1450d810b4ecac1c151dbfcfceb9d96d35/src/arch/arm/kernel/boot.c#L422-L425

libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
```diff
@@ -128,6 +128,9 @@ typedef enum {
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0x0000003ffffff000
 
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop 0x0000003fffffefff
```
lsf37: Is the `e` a typo?

```suggestion
#define seL4_UserVSpaceTop 0x0000003fffffffff
```

lsf37: Oh, right, it's not a typo if we are consistent with `0x0000003ffffff000` being the first address not accessible. Somehow that seems weird, though.

CHANGES.md
```diff
@@ -30,6 +30,10 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 * Add `seL4_DebugGetThreadAffinity()` syscall to return the current affinity of a thread
   in SMP configurations.
+* Add new libsel4 constant `seL4_UserVSpaceTop` which is a *inclusive* value representing
+  for the highest available address available to user level. The existing constant
```
lsf37:
```suggestion
  the highest available address available to user level. The existing constant
```

CHANGES.md
```diff
@@ -30,6 +30,10 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
...
+  `seL4_UserTop` is a mix of inclusive and exclusive across existing platforms, however,
+  changing this to be consistent would break userspace and require proof updates.
```
lsf37, line 36: Do we have a mechanism for marking `seL4_UserTop` as deprecated or should we add that to the change log only?

libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
midnightveil: Yes, it does... x86 had the TLB bitmaps in the top page, I don't think RISC-V does.

CHANGES.md
midnightveil: We have `SEL4_DEPRECATE_MACRO`, yes. 

The issue is that internally we make `USER_TOP` defined to `seL4_UserTop`, so we'd have to change the kernel internals for that, too.

midnightveil:
OK, somehow this has broken a few of the (hyp) AArch64 platforms (but not all of them ??).

```
  [8040000000..8040241000)
  [8040241000..804024be26)
  [804024c000..8040632000)

Warning: using printf before serial is set up. This only works as your
printf is backed by seL4_Debug_PutChar()
find_range@vspace.c:271 Out of virtual memory
init_env@main.c:110 Failed to provide virtual memory for allocator
seL4 root server abort()ed
Debug halt syscall from user thread 0x80fffd8400 "sel4test-driver"
halting...
Kernel entry via Unknown syscall, word: 147073
[[Timeout]]
None
```

(Well, specifically the change I made to seL4_libs)

@lsf37 the summary in the job gives the cause as "incomplete", we probably could make that detect "halting..."

lsf37:
> OK, somehow this has broken a few of the (hyp) AArch64 platforms (but not all of them ??).
> 
> ```
> reserved virt address space regions: 3
>   [8040000000..8040241000)
>   [8040241000..804024be26)
>   [804024c000..8040632000)
> Booting all finished, dropped to user space
> 
> Warning: using printf before serial is set up. This only works as your
> printf is backed by seL4_Debug_PutChar()

Smells like a failing assertion somewhere early. Although I thought this was an equivalence transformation, so not sure.

> @lsf37 the summary in the job gives the cause as "incomplete", we probably could make that detect "halting..."

Yup, it's on my list.

midnightveil:
> Smells like a failing assertion somewhere early. Although I thought this was an equivalence transformation, so not sure.

Well, I think strictly speaking it is slightly different for seL4_libs for ARM: previously it did `ROUND_DOWN(user top, page size)`. Which means it would have given `0x00000ffffffff000` instead of the correct `0x0000100000000000`. But...

libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
midnightveil: This was changed in this commit to be the current value: 5fe6d0bcef3c7475a7bc4e4d7b786b2df8be2dc6

Also intriguing: the proofs definition: https://github.com/seL4/l4v/blob/24a4db686899acf17c05141f5280dd94523db5e7/spec/machine/RISCV64/Platform.thy#L137-L145, which had this change: https://github.com/seL4/l4v/commit/23866cbae9610840647b2ef4c5139ff466910628

libsel4/sel4_plat_include/allwinnerA20/sel4/plat/api/constants.h
```diff
@@ -11,3 +11,6 @@
 
 /* First address in the virtual address space that is not accessible to user level */
 #define seL4_UserTop 0xa0000000
+
+/* Last address in the virtual address space that is accessible to user level */
+#define seL4_UserVSpaceTop 0xbfffffff
```
Indanz: That should be 0x9fffffff.

Indanz (commented):
Please don't add more magic constants than necessary. There is no reason not to do:

`#define seL4_UserVSpaceTop (seL4_UserTop - 1)`

We could add the following to libsel4/include/sel4/deprecated.h:

```
#define seL4_UserTop SEL4_DEPRECATE_MACRO(seL4_UserTop)
```

Indanz:
> OK, somehow this has broken a few of the (hyp) AArch64 platforms (but not all of them ??).

That's weird, it is related to how much memory the boards have or something?

midnightveil:
> We could add the following to libsel4/include/sel4/deprecated.h:

That would give redefined macro warnings...

libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
lsf37: Looking more into this, I think the RISCV value is just bogus. Probably the setup was copied from x64 and never changed, because it doesn't lead to any problems in the kernel or proofs if there is an entirely unused top-level slot. I should try change this separately and see if the proofs work (might need some adaptation).

lsf37:
> > OK, somehow this has broken a few of the (hyp) AArch64 platforms (but not all of them ??).
> 
> That's weird, it is related to how much memory the boards have or something?

Memory layout on AArch64 hyp is different from AArch64 EL1. Hyp uses a separate TLB for the kernel (which I think we should do for EL1 as well, because there is one available, but that's a different story), so the user should be able to the entire virtual address space. Maybe the + 1 is exactly the kind of overflow I was talking about :-). It would not have happened before if it was rounding down to page size.

midnightveil:
> Memory layout on AArch64 hyp is different from AArch64 EL1

Yes, but some of the AArch64 boards (TX1/TX2 mainly) work in EL2. The rest of them only work in EL1.

Saying that now, however, the TX1 and TX2 have `KernelArmPASizeBits44` and thus a different UserTop. Most other boards are `KernelArmPASizeBits40` with the exception of the RPI which is not tested in CI, I believe.

> Maybe the + 1 is exactly the kind of overflow I was talking about :-)

Maybe :) Though none of our UserTops are anywhere close to `2^64 - 1` so I was expected it to be just fine.

midnightveil:
OK, at least I can reproduce this in QEMU's default EL2 configuration (but not when I make the cpu a cortex-a57 like t1/tx2) Which, is kinda weird to me that we only test the EL1 AArch64 in QEMU simulation even though EL2 is the verified config.

midnightveil:
OK, for Cortex-a53 (PA_SIZE = 40)

```
KERNEL_RESERVED_START is 0x10000000000
seL4_UserVSpaceTop is 0xffffffffff
seL4_UserTop is 0xffffffffff
Old KERNEL_RESERVED_START is 0xfffffff000
Old TOP_LEVEL_INDEX(KERNEL_RESERVED_START) is 0x1ff
New TOP_LEVEL_INDEX(KERNEL_RESERVED_START) is 0x0
```

For Cortex-a57 (PA_SIZE = 44)

```
KERNEL_RESERVED_START is 0x100000000000
seL4_UserVSpaceTop is 0xfffffffffff
seL4_UserTop is 0xfffffffffff
Old KERNEL_RESERVED_START is 0xffffffff000
Old TOP_LEVEL_INDEX(KERNEL_RESERVED_START) is 0x1f
New TOP_LEVEL_INDEX(KERNEL_RESERVED_START) is 0x20
```

The code does:

```c
    /* reserve the kernel region, we do this by marking the
     * top level entry as RESERVED */
    if (!data->is_empty) {
        for (int i = TOP_LEVEL_INDEX(KERNEL_RESERVED_START);
             i < VSPACE_LEVEL_SIZE; i++) {
            data->top_level->table[i] = RESERVED;
        }
    }
```

When the `TOP_LEVEL_INDEX` is 0 this breaks things, a bit.

Note:

```c

#if defined(CONFIG_ARM_PA_SIZE_BITS_44)

#define VSPACE_NUM_LEVELS       4

#elif defined(CONFIG_ARM_PA_SIZE_BITS_40)

#define VSPACE_NUM_LEVELS 3

#else
#error Unspecified PA size bits
#endif
```

```c
#define TOP_LEVEL_INDEX(x) INDEX_FOR_LEVEL(x, VSPACE_NUM_LEVELS - 1)
```

lsf37:
> OK, at least I can reproduce this in QEMU's default EL2 configuration (but not when I make the cpu a cortex-a57 like t1/tx2) Which, is kinda weird to me that we only test the EL1 AArch64 in QEMU simulation even though EL2 is the verified config.

They are not meant to be exhaustive (that would be way too many), and arguably it is more important to test the non-verified ones, but covering the main verified configs would be good, yes.

midnightveil:
OK, so I'm confused. If `seL4_UserTop` when PA_SIZE=40 is `1 << 40 - 1`, but when `PA_SIZE = 40` we only have 3 levels of page tables, the VA range permitted by the 3-level of page tables is `0..2^39-1`. So the `seL4_UserTop` is higher than the actual maximum virtual address permitted, which seems wrong.

<details>
<summary>I get side-tracked</summary>

TBH I don't quite understand why the `AARCH64_VSPACE_S2_START_L1` option exists, what is wrong with a 4-level page table when it's 40 bits when we already have this UserTop check? Anyhow... But also, what is wrong with a 48-bit VA even with a 40-bit PA? The VA and PA sizes should have nothing to do with each other, and it looks like ARM64 unconditionally supports 48-bit VAs.?

I don't see why `seL4_UserTop` isn't just `2^48 - 1` like on non-hyp, and the comment in the code doesn't make sense to me:

```
/* The userspace occupies the range 0x0 to 0xfffffffffff.
 * The stage-1 translation is disabled, and the stage-2
 * translation input addree size is constrained by the
 * ID_AA64MMFR0_EL1.PARange which is 44 bits on TX1.
 * Anything address above the range above triggers an
 * address size fault.
 */
 ```
 
However, there might be some weird interaction with Stage 2 tables for *VMs* that I haven't though about, but that's not what makes sense here.

EDIT: Ok, after looking at the ARM pseudocode, I understand this now. For some reason the way arm has implemented this, when Stage 1 translation is off, `VA => IPA` directly, and since S2 translation only goes from `IPA => VA`, and they make the `IPA` sized precisely the same as `PA`, [this check in the S1DisabledOutput() pseudocode](https://arm.jonpalmisc.com/latest_aarch64/shared_pseudocode#AArch64.S1DisabledOutput.5) causes a fault:

```
    constant AddressSize addrtop = AArch64.AddrTop(walkparams.tbid, accdesc.acctype,
                                                   walkparams.tbi);
    constant AddressSize pamax = AArch64.PAMax();

    if !IsZero(va<addrtop:pamax>) then
        fault.statuscode = Fault_AddressSize;
```

</details>

**Comment by midnightveil** [comment]
I've worked around that quirk in the `seL4_libs` PR, so now all the tests should pass.

EDIT: and now they do

**Comment by midnightveil** [comment]
Was there any comments I missed on this PR?

**Comment by Indanz** [comment]
> Was there any comments I missed on this PR?

Only this bit:

Please don't add more magic constants than necessary. There is no reason not to do:

`#define seL4_UserVSpaceTop (seL4_UserTop - 1)`

File: CHANGES.md (hunk 3 above)
Comment (Indanz) [reply CHANGES.md:36]: Or in the user space header file, do:

```c
#define seL4_UserTop_deprecated seL4_UserTop
#undef seL4_UserTop
#define seL4_UserTop SEL4_DEPRECATE_MACRO(seL4_UserTop_deprecated)
```

midnightveil:
> Please don't add more magic constants than necessary.

I somewhat disagree in this case but I will make the change. (I think it makes it harder to look at the value and just figure out what it is, and if they're defined next to each other it's obvious they're related).

lsf37 (approved):

libsel4/sel4_arch_include/riscv64/sel4/sel4_arch/constants.h
midnightveil: I've opened an issue to keep track of this: https://github.com/seL4/seL4/issues/1697

CHANGES.md
midnightveil: I don't understand why, but if I do that it seems like it complains that it is not defined. (If I remove the `#undef` it complains that it is both undefined *and* that it is being redefined).

(Also, I don't know if this would cause issues with "VSpaceUserTop" now being defined in terms of UserTop)

midnightveil:
OK, I changed the definition as per @Indanz's recommendation.

The only thing maybe left unresolved is if we can actually add it to the `deprecated.h` header, I can't seem to make it work: https://github.com/seL4/seL4/pull/1695#discussion_r3464655314.

I've made a change that uses comments to mark as deprecated, for now, under the *impression* that we want to deprecate this constant based on both Indan's and Gerwin's implicit agreement [here](https://github.com/seL4/seL4/pull/1695#discussion_r3464655314) that we should mark it as such. If not, I can drop the second commit.

lsf37 (approved):
Good from my side, I'm fine with just the deprecation comment for now. We can still come back to this later if we want to.

Indanz:
> > Please don't add more magic constants than necessary.
> 
> I somewhat disagree in this case but I will make the change. (I think it makes it harder to look at the value and just figure out what it is, and if they're defined next to each other it's obvious they're related).

Sure, but you made a mistake when doing that. That kind of mistakes can't happen if you only define a magic constant in one place.

Not exposing kernel-internal details to user space and having a separate seL4_UserVSpaceTop constant in a libsel4 header would be a very good reason to not re-use `seL4_UserTop`. But then I would define `seL4_UserTop` in terms of `seL4_UserVSpaceTop` if possible.

CHANGES.md
Indanz: Yeah, just shows that I don't use macros enough. Seems like they get expanded at a different moment than I thought. Oh well, never mind.

CHANGES.md
```diff
@@ -30,6 +30,11 @@ description indicates whether it is SOURCE-COMPATIBLE, BINARY-COMPATIBLE, or BRE
 
 * Add `seL4_DebugGetThreadAffinity()` syscall to return the current affinity of a thread
   in SMP configurations.
+* Add new libsel4 constant `seL4_UserVSpaceTop` which is a *inclusive* value representing
+  the highest available address available to user level. The existing constant
+  `seL4_UserTop` is a mix of inclusive and exclusive across existing platforms, however,
+  changing this to be consistent would break userspace and require proof updates.
+  We deprecate the old `seL4_UserTop` value.
```
Indanz:
```suggestion
  The existing `seL4_UserTop` define is deprecated.
```
Or remove this part if we don't really plan on removing it in the future. Doesn't hurt to keep, so we don't need to deprecate it.

Indanz (approved):

CHANGES.md
lsf37: I think we should remove it in the future, it's not good to have the same concept twice. Maybe after 2-3 releases, though (e.g. 1 year), there is no particular hurry.

midnightveil:
OK, rebased on main & adjusted to Indan's comment.

I'll merge this after hw-build succeeds.

Indanz (approved):

PR #1698, opened 2026-07-02
configs: proofs now generic in RISC-V config; add stm32mp2 for AArch64

lsf37:
- add a verified config for stm32mp2 (AArch64); proofs are passing unchanged
- add verified configs for all supported non-simulation RISC-V platforms

Related to https://github.com/seL4/l4v/pull/1024, but can be merged independently.


lsf37:
(The style checker is probably going to complain about the executable bit for the configs; this is expected)

lsf37:
> (The style checker is probably going to complain about the executable bit for the configs; this is expected)

It doesn't even complain any more, I'm disappointed :-)

Indanz (approved):

PR #1699, opened 2026-07-03
aarch64: print out debug information for invalid vector entries

midnightveil:
Previously, this would just print a 'halting via unknown'
in debug mode when an SError occurred and SError ignore was
not turned on. This changes the trap code so that we print
out the known cause of the trap. (Note that ELR/FAR might be
useless or garbage depending on the trap).

    KERNEL INVALID VECTOR ENTRY!
    Vector: 0x580 (SError 64-bit EL0/EL1)
    Fault attributed to address: 0x221dc0
    ESR: 0xbf000000 FAR: 0x2881c98000824100
    halting...
    Kernel entry via Unknown (0)

This was encountered after some rust-seL4 changes to the
initialiser for Microkit related to untyped mappings broke
boot: https://github.com/seL4/microkit/issues/541

Unrelatedly, but the `traps.S` implementation for ventry is less efficient than necessary, incurring a useless branch — each vector entry has space for 32 instructions (128-bit alignment) and most of our assembly code would fit directly. Probably not a big issue, though.

### DIFF
--- a/include/arch/arm/arch/kernel/traps.h
+++ b/include/arch/arm/arch/kernel/traps.h
@@ -57,3 +57,6 @@ void c_handle_vcpu_fault(word_t hsr)
 VISIBLE SECTION(".vectors.text");
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
+#if defined(CONFIG_ARCH_AARCH64) && defined(CONFIG_DEBUG_BUILD)
+void VISIBLE c_handle_invalid_vector_entry(word_t vect_offset, word_t pc);
+#endif
--- a/src/arch/arm/64/c_traps.c
+++ b/src/arch/arm/64/c_traps.c
@@ -65,3 +65,54 @@ void VISIBLE NORETURN restore_user_context(void)
     );
     UNREACHABLE();
 }
+
+#if defined(CONFIG_DEBUG_BUILD)
+/* See 'arm_vector_table' for details on 'vect_offset' */
+static const char *vect_offset_to_name(word_t vect_offset)
+{
+    switch (vect_offset) {
+    case 0x000:
+        return "Synchronous EL1t/EL2t";
+    case 0x080:
+        return "IRQ EL1t/EL2t";
+    case 0x100:
+        return "FIQ EL1t/EL2t";
+    case 0x180:
+        return "SError EL1t/EL2t";
+    case 0x200:
+        return "Synchronous Current EL";
+    case 0x280:
+        return "IRQ Current EL";
+    case 0x300:
+        return "FIQ Current EL";
+    case 0x380:
+        return "SError Current EL";
+    case 0x400:
+        return "Synchronous 64-bit EL0/EL1";
+    case 0x480:
+        return "IRQ 64-bit EL0/EL1";
+    case 0x500:
+        return "FIQ 64-bit EL0/EL1";
+    case 0x580:
+        return "SError 64-bit EL0/EL1";
+    case 0x600:
+        return "Synchronous 32-bit EL0/EL1";
+    case 0x680:
+        return "IRQ 32-bit EL0/EL1";
+    case 0x700:
+        return "FIQ 32-bit EL0/EL1";
+    case 0x780:
+        return "SError 32-bit EL0/EL1";
+    default:
+        return "<Unknown>";
+    }
+}
+
+void VISIBLE c_handle_invalid_vector_entry(word_t vect_offset, word_t pc)
+{
+    printf("\n\nKERNEL INVALID VECTOR ENTRY!\n");
+    printf("Vector: 0x%"SEL4_PRIx_word" (%s)\n", vect_offset, vect_offset_to_name(vect_offset));
+    printf("Fault attributed to program counter: 0x%"SEL4_PRIx_word"\n", pc);
+    printf("ESR: 0x%"SEL4_PRIx_word" FAR: 0x%"SEL4_PRIx_word"\n", getESR(), getFAR());
+}
+#endif
--- a/src/arch/arm/64/traps.S
+++ b/src/arch/arm/64/traps.S
@@ -40,28 +40,58 @@
     b       \label
 .endm
 
+#ifdef CONFIG_DEBUG_BUILD
+.macro ventry_invalid
+.align 7
+    /* compute vect_offset argument */
+    adr     x0, .
+    adr     x19, arm_vector_table
+    sub     x0, x0, x19
+    b       invalid_vector_entry
+.endm
+#else
+.macro ventry_invalid
+    ventry invalid_vector_entry
+.endm
+#endif
+
+#ifdef CONFIG_AARCH64_SERROR_IGNORE
+.macro ventry_serror
+.align 7
+    eret
+.endm
+#else
+.macro ventry_serror
+    ventry_invalid
+.endm
+#endif
+
 .section .vectors, "ax"
 
+/* The layout of this table are specified in Table D1-7, 'Vector offsets from
+ * vector table base address' of ARM DDI 0487B.b ID092517.
+ * Each vector entry has 128-byte alignment, and so supports 32 instructions.
+ */
 BEGIN_FUNC(arm_vector_table)
-    ventry  invalid_vector_entry           // Synchronous EL1t/EL2t
-    ventry  invalid_vector_entry           // IRQ EL1t/EL2t
-    ventry  invalid_vector_entry           // FIQ EL1t/EL2t
-    ventry  invalid_vector_entry           // SError EL1t/EL2t
+    ventry_invalid                         // Synchronous EL1t/EL2t
+    ventry_invalid                         // IRQ EL1t/EL2t
+    ventry_invalid                         // FIQ EL1t/EL2t
+    ventry_invalid                         // SError EL1t/EL2t
 
     ventry  cur_el_sync                    // Current EL Synchronous (EL1/2)
     ventry  cur_el_irq                     // IRQ
-    ventry  invalid_vector_entry           // FIQ
-    ventry  cur_el_serr                    // SError
+    ventry_invalid                         // FIQ
+    ventry_serror                          // SError
 
     ventry  lower_el_sync                  // Synchronous 64-bit EL0/EL1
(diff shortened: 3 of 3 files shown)


include/arch/arm/arch/kernel/traps.h
```diff
@@ -57,3 +57,7 @@ void c_handle_vcpu_fault(word_t hsr)
 VISIBLE SECTION(".vectors.text");
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
+#if defined(CONFIG_ARCH_AARCH64) && defined(CONFIG_DEBUG_BUILD)
+void c_handle_invalid_vector_entry(word_t vect_offset, word_t pc)
+VISIBLE;
```
Indanz: This looks very weird, I'd put it on one line:
```suggestion
VISIBLE void c_handle_invalid_vector_entry(word_t vect_offset, word_t pc);
```

src/arch/arm/64/traps.S
```diff
@@ -43,28 +40,44 @@
...
+.endm
+#else
+.macro ventry_invalid vect_offset
+    ventry invalid_vector_entry
+.endm
+#endif
+
 .section .vectors, "ax"
 
+/* The offsets in and layout of this table are specified in Table D1-7,
```
Indanz:
```suggestion
/* The layout of this table is specified in Table D1-7,
```
Also, you now have the same reference to the Arm doc twice, both here and above `vect_offset_to_name()`. I think just having this one is enough, and having a shorter "See arm_vector_table" above the other one is sufficient.

src/arch/arm/64/traps.S
```diff
@@ -43,28 +40,44 @@
...
+   'Vector offsets from vector table base address' of ARM DDI 0487B.b ID092517.
+   Each vector entry has 128-bit alignment, and so supports 32 instructions.
```
Indanz:
```suggestion
   Each vector entry has 128-byte alignment, and so supports 32 instructions.
```

src/arch/arm/64/traps.S
```diff
@@ -43,28 +40,44 @@
     b       \label
 .endm
 
+#ifdef CONFIG_DEBUG_BUILD
+.macro ventry_invalid vect_offset
+.align 7
+    mov     x0, \vect_offset
+    b       invalid_vector_entry
+.endm
```
Indanz:
```suggestion
.macro ventry_invalid
.align 7
    adr     x0, .
    adr     x19, arm_vector_table
    sub     x0, x0, x19
    b       invalid_vector_entry
.endm
```
Then you don't need to manually fill in the error prone offsets.

src/arch/arm/64/traps.S
```diff
@@ -144,6 +164,7 @@ BEGIN_FUNC(cur_el_serr)
 #ifdef CONFIG_AARCH64_SERROR_IGNORE
     eret
 #else
+    mov     x0, #0x280
```
Indanz: This one we can easily avoid by defining the entry to either `ventry cur_el_serr` or `ventry_invalid`.

src/arch/arm/64/traps.S
```diff
@@ -233,6 +254,7 @@ BEGIN_FUNC(lower_el_serr)
 #ifdef CONFIG_AARCH64_SERROR_IGNORE
     eret
 #else
+    mov     x0, #0x580
```
Indanz: Same here.

src/arch/arm/64/traps.S
```diff
@@ -126,6 +145,7 @@ cur_el_ia:
     b       halt
 
 cur_el_inv:
+    mov     x0, #0x200
```
Indanz, line 162: Maybe we could make `ventry_invalid` more generic and give it a label to jump to instead of hard coding it, and then use `ventry_invalid cur_el_sync`, or something like that.

```
.macro ventry_invalid func
.align 7
    adr     x0, .
    adr     x19, arm_vector_table
    sub     x0, x0, x19
    b       \func
.endm
```
(Maybe better to add the above macro and use it to define `ventry_invalid` instead of replacing it.)

Indanz:
> Previously, this would just print a 'halting via unknown' in debug mode when an SError occurred and SError ignore was not turned on. This changes the trap code so that we print out the known cause of the trap.

Yes, this would help debugging such problems much easier.

> Unrelatedly, but the `traps.S` implementation for ventry is less efficient than necessary, incurring a useless branch — each vector entry has space for 32 instructions (128-bit alignment) and most of our assembly code would fit directly. Probably not a big issue, though.

True, but not really a problem as the paths where we care about performance currently don't fit anyway. You'd need to fit all the instructions up to the branch to `c_handle_fastpath_call` in 32 instructions. That's a lot of effort to avoid one jump. You probably get more performance gain by cache line aligning `lower_el_sync`.

include/arch/arm/arch/kernel/traps.h
midnightveil: This is consistent with the rest of the definitions in this file.

I somewhat dislike it too, but it is consistent.

src/arch/arm/64/traps.S
midnightveil: I think this wouldn't work, because `curl_el_sync` does an `eret`, which requires that the stack frame of the thread invoking us is still valid.

i.e. we'd need to save the registers as per `kernel_enter`, otherwise we overwrite the registers of the user.

src/arch/arm/64/traps.S
midnightveil: To be clear: I'm interpreting this as putting the ifdef inside the arm_vector_table and leaving this alone.

include/arch/arm/arch/kernel/traps.h
Indanz: The other ones only put it on the next line if they also add a `SECTION(".vectors.text")`. The only other example is `void VISIBLE NORETURN restore_user_context(void);`, which doesn't do that. But that one puts the void before the other stuff, which is also ugly (but consistent across arches).

Long story short, just do what you personally prefer.

src/arch/arm/64/c_traps.c
```diff
@@ -65,3 +65,55 @@ void VISIBLE NORETURN restore_user_context(void)
     );
     UNREACHABLE();
 }
+
+#if defined(CONFIG_DEBUG_BUILD)
+/* See 'arm_vector_table' for details on 'vect_offset' */
+static const char *vect_offset_to_name(word_t vect_offset)
+{
+    switch (vect_offset) {
```
Indanz, line 73: I probably would have written it as a lookup table instead of a switch statement, because it is all contiguous. But this is fine too, it's only debug code and probably compiles to the same anyway.

src/arch/arm/64/c_traps.c
```diff
@@ -65,3 +65,55 @@ void VISIBLE NORETURN restore_user_context(void)
...
+    }
+}
+
+void VISIBLE c_handle_invalid_vector_entry(word_t vect_offset, word_t pc)
+{
+    printf("\n\nKERNEL INVALID VECTOR ENTRY!\n");
+    printf("Vector: 0x%"SEL4_PRIx_word" (%s)\n", vect_offset, vect_offset_to_name(vect_offset));
+    printf("Fault attributed to program counter: 0x%"SEL4_PRIx_word"\n", pc);
+    printf("ESR: 0x%"SEL4_PRIx_word" FAR: 0x%"SEL4_PRIx_word"\n", getESR(), getFAR());
+    halt();
```
Indanz: You could return and let the caller call halt instead. Would save an `#else`.

src/arch/arm/64/traps.S
```diff
@@ -95,9 +119,15 @@ END_FUNC(arm_vector_table)
     stp     x22, x23, [sp, #PT_ELR_EL1]
 .endm
 
+/* Takes argument x0: vect_offset in debug mode */
```
Indanz:
```suggestion
/* For debug builds x0 contains the vector offset */
```

Edit: Or "For debug builds the vector offset is passed via argument x0" or something. Main thing is to be unambiguous about debug builds and not some other kind of debug.

src/arch/arm/64/traps.S
Indanz: It always calls halt, but sometimes via `kernelDataAbort` or `kernelPrefetchAbort`. Did you mix it up with `lower_el_sync`?

Indanz (approved):
We can always do the `cur_el_sync` change later if we want, I'm happy with what we have.

I would have avoided the ifdefs in the vector table itself by defining `ventry_serror` separately into either `ventry_invalid` or directly into an `eret`, and remove the existing `cur_el_serr` and `lower_el_err`. Because SError is asynchronous, it doesn't matter whether it arrives in the lower or upper EL level anyway.

src/arch/arm/64/traps.S
midnightveil: Well, that's what your comment here was about (seemingly attached to the wrong spot).

src/arch/arm/64/c_traps.c
midnightveil: This makes it easier to handle the not found path (IMO). And if the style check didn't force it it would be just as compact.

src/arch/arm/64/traps.S
Indanz: `cur_el_inv` is branched to by `cur_el_sync`, which checks whether it's a data abort or prefetch abort or something else. `cu_el_inv` handles the something else case. My comment is here because that's where you've added the constant.

But to fix it my proposed way above you'd need to change the vector entry for `cur_el_sync` to add the extra vector offset info.

However, looking at it again, I think it would make more sense to just have `ventry_invalid` instead of `cur_el_sync` and do the check for data abort or prefetch inside of `c_handle_invalid_vector_entry`. The only missing info is `ELR`, but that's added easily enough. Then we have all the info needed to call `kernelDataAbort` or `kernelPrefetchAbort`.

Indanz: Hm, but if putting it in `c_handle_invalid_vector_entry`, we're back to using the magic constant 0x200. Oh well, just leave it as it is for now.

Indanz (approved):

PR #1700, opened 2026-07-05
x86,c_traps: avoid PC corruption in kernel lock

lsf37:
When we back out of the lock via IPI stall in `ipiStallCoreCallback()`, the `FaultIP`/`NextIP` are not updated yet and the lock exit code will set the wrong program counter.

Moving the `FaultIP`/`NextIP` update before the lock makes sure that all exit paths from the lock will restore to the correct program counter.

Even though this write happens outside the lock, it does not introduce a race. The only possible interference would be a `TCBWriteRegisters` syscall from another core along the lines of the following: other core enters syscall, current core gets to lock, makes FaultIP adjustment, other core exists syscall, current core overwrites effect. This path and analogous ones are not possible, because the `TCBWriteRegisters` call on the other core first leads to a stall  on this core before it proceeds, which means this core will be on idle and cannot attempt to enter the lock.

The fix is observable in the sel4bench SMP benchmark. Before, when run over multiple full iterations (n < 10 on haswell is usually enough, e.g. https://github.com/seL4/sel4bench/actions/runs/28563144172/job/84686342372), the benchmark hangs, because the TCB setup/teardown race with IPC operations on other cores and trigger the PC corruptions, which via multiple cascading VM faults eventually hang the main benchmark thread. After this change, the benchmark is stable for > 400 iterations on the same machine.

### DIFF
--- a/src/arch/x86/c_traps.c
+++ b/src/arch/x86/c_traps.c
@@ -147,6 +147,16 @@ void VISIBLE NORETURN c_handle_syscall(word_t cptr, word_t msgInfo, syscall_t sy
         x86_enable_ibrs();
     }
 
+    /* Must happen before NODE_LOCK_SYS so that lock exit via IPI stall restores
+       the correct program counter. */
+    if (config_set(CONFIG_SYSENTER)) {
+        /* increment NextIP to skip sysenter */
+        NODE_STATE(ksCurThread)->tcbArch.tcbContext.registers[NextIP] += 2;
+    } else {
+        /* set FaultIP */
+        setRegister(NODE_STATE(ksCurThread), FaultIP, getRegister(NODE_STATE(ksCurThread), NextIP) - 2);
+    }
+
     NODE_LOCK_SYS;
 
     c_entry_hook();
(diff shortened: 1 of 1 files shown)


midnightveil:
Is there a reason why this is done in C for x86 as opposed to ASM like ARM(64) or ARM(32)?

e.g.

https://github.com/seL4/seL4/blob/b161c3fe98787df5ac2c7862c7f1ea89342b52e7/src/arch/arm/64/traps.S#L193-L194

Indanz (approved):

Indanz:
> Is there a reason why this is done in C for x86 as opposed to ASM like ARM(64) or ARM(32)?

Same reason why other registers are not stored in assembly, like TPIDR: It was added later in a sub-optimal way, probably because it seemed easier. See also https://github.com/seL4/seL4/pull/490.

lsf37:
I'd like to merge the change in C for now just as minimal change, but I'd have no problem with this being in assembly. I'd be very happy if somebody picked this up (or/and remaining bits of #490) afterwards.

midnightveil (approved):

lsf37:
All hw tests have passed, will merge once the build after the branch update has succeeded.

PR #1702, opened 2026-07-08
libsel4: export seL4_IOPageTable{Index,Entry}Bits

midnightveil:
This is useful for all x86 IOMMU work, which otherwise has to duplicate the definitions of VTD_PT_INDEX_BITS in userspace code. Follow the same pattern as many other existing code in the libsel4 headers where the kernel depends on it.

Note we can't define seL4_IOPageTableBits in terms of Index + Entry because the capDL tool does the c-preprocessor to make a YAML file containing the constant, and it won't do evaluation.

Cross-ref: https://github.com/seL4/rust-sel4/pull/358/changes#r3541477792

### DIFF
--- a/include/arch/x86/arch/object/structures.h
+++ b/include/arch/x86/arch/object/structures.h
@@ -16,6 +16,7 @@
 #include <arch/machine/hardware.h>
 #include <arch/machine/registerset.h>
 #include <sel4/arch/constants.h>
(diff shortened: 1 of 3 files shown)


lsf37 (approved):
Looks good to me, should be non-breaking change and should be useful.

PR #1703, opened 2026-07-08
libsel4: export seL4_X86_IOSpace_CapData type

midnightveil:
Previously, to create IOSpaceCap's one needed to replicate the layout of these structures in your own handwritten code.

Use the existing shared_types mechanism to expose helpers, and update the internal kernel implementation.

As far as I can tell this is not verified code.

Naming scheme is based on `seL4_CNode_CapData`.

Cross-ref: https://github.com/seL4/rust-sel4/pull/358#discussion_r3541708597

### DIFF
--- a/include/arch/x86/arch/32/mode/object/structures.bf
+++ b/include/arch/x86/arch/32/mode/object/structures.bf
@@ -98,17 +98,13 @@ block io_port_cap {
 #ifdef CONFIG_IOMMU
 
 -- IO Space Cap
+-- See also: seL4_X86_IOSpace_CapData in arch/shared_types.bf
 block io_space_cap {
     field       capDomainID     16
     field       capPCIDevice    16
 
     padding                     28
     field       capType         4
-}
-
-block io_space_capdata {
-    field domainID              16
-    field PCIDevice             16
 }
 
 -- IO Page Table Cap
--- a/include/arch/x86/arch/64/mode/object/structures.bf
+++ b/include/arch/x86/arch/64/mode/object/structures.bf
@@ -124,6 +124,7 @@ block io_port_cap {
 #ifdef CONFIG_IOMMU
 
 -- IO Space Cap
+-- See also: seL4_X86_IOSpace_CapData in arch/shared_types.bf
 block io_space_cap {
     padding 64
     field       capType         5
@@ -131,12 +132,6 @@ block io_space_cap {
     field       capPCIDevice    16
 
     padding                     27
-}
-
-block io_space_capdata {
-    padding 32
-    field domainID              16
-    field PCIDevice             16
 }
 
 -- IO Page Table Cap
--- a/libsel4/arch_include/x86/sel4/arch/shared_types.bf
+++ b/libsel4/arch_include/x86/sel4/arch/shared_types.bf
@@ -25,3 +25,15 @@ tagged_union seL4_Fault seL4_FaultType {
     tag VMFault 5
 #endif
 }
+
+#ifdef CONFIG_IOMMU
+
+block seL4_X86_IOSpace_CapData {
+#ifdef CONFIG_ARCH_X86_64
+    padding 32
+#endif
+    field domainID              16
+    field PCIDevice             16
+}
+
+#endif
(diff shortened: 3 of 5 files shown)


midnightveil:
I can't actually find any code in any of the seL4_libs libraries that even constructs these, I'd update them but can't find it.

Indanz (commented):
This is the cap data passed to the mint syscall? In that case, it's `block seL4_CNode_CapData` that's wrong really.

Why not add the new structures to the existing libsel4/arch_include/x86/sel4/arch/shared_types.bf file? Doing it there with conditional padding for 64-bits seems cleaner than adding another file named shared_types.bf.

Seems weird that libsel4/sel4_arch_include/*/sel4/sel4_arch/types.bf can contain stuff that's not shared with the kernel, the only difference seems to be the padding.

midnightveil:
> This is the cap data passed to the mint syscall? In that case, it's block seL4_CNode_CapData that's wrong really.

Why is it wrong? Are you saying the name doesn't make sense?  It does to me. (it also matches the previous `iospac_cap_data` I'm exposing.

> Why not add the new structures to the existing libsel4/arch_include/x86/sel4/arch/shared_types.bf file? Doing it there with conditional padding for 64-bits seems cleaner than adding another file named shared_types.bf.

Generally it seems like we avoid doing arch/mode specific ifdefs in the bitfield files. But yes in theory we could do that, I also don't exactly prefer it however.

> Seems weird that libsel4/sel4_arch_include/*/sel4/sel4_arch/types.bf can contain stuff that's not shared with the kernel, the only difference seems to be the padding.

Yes there are a bunch of duplicated definitions for internal vs external for faults and such that could have been structured better (or at least explained better), but too late for that now.

lsf37:
> I can't actually find any code in any of the seL4_libs libraries that even constructs these, I'd update them but can't find it.

The only reference in the existing code I can find is this:

https://github.com/seL4/camkes-tool/blob/e6addd3be38aa46b48d86ed9426ec3cface6fcf9/camkes/templates/component.simple.c#L332

which is then used here:

https://github.com/seL4/camkes-tool/blob/e6addd3be38aa46b48d86ed9426ec3cface6fcf9/camkes/templates/component.simple.c#L442

which is then available in seL4_libs somewhere (but seL4_libs itself doesn't do anything to domainID and deviceID, it just passes them to this function eventually).

midnightveil:
Yeah, so it's doing the packing manually as well.

lsf37:
> Yeah, so it's doing the packing manually as well.

Yes, and it looks like the C capDL loader also does, although it's a bit more roundabout. The capDL tool prints a raw badge, and it computes the badge like this (where `domainID`, `pciBus`, `pciDev`, `pciFun` are parts of the spec):

```
showPCI :: Word -> (Word, Word, Word) -> String
showPCI domainID (pciBus, pciDev, pciFun) =
    hex $ shift domainID 16 .|. shift pciBus 8 .|. shift pciDev 3 .|. pciFun
```

lsf37:
Doesn't mean we can't export the bf structure and use it, though. It would at least properly define where the `domainID` sits. The kernel doesn't really care about the semantics of the rest (`deviceID`), it just passes it through.

It is cap data so the name is fine with me.

midnightveil:
> The kernel doesn't really care about the semantics of the rest (deviceID), it just passes it through.

That's true. I suppose the PCI deviceID layout is determined by the underlying layout. I was expecting it to be used by the kernel, but I guess userspace must duplicate the bit layout the hardware wants itself.

lsf37 (approved):
From the verification side, this is fine, and I think the name is Ok.

For `#ifdef` vs new file, we do prefer less `#ifdef`, but of course a new file has the same effect, so from the verification side it doesn't really matter.

In this case, I'd find an `#ifdef` slightly preferable, because the only thing that changes is the padding field. Usually more numbers change between 32/64 and `#ifdef` gets hard to read, but here it'd still be fine.

midnightveil:
OK, I'll do this with an ifdef in the x86 shared_types.bf file then.

Indanz (approved):

PR #1704, opened 2026-07-09
IOMMU: Add large page support for IOSpaces

cazb2:
Previously the IOMMU on x86 only supported 4KiB pages. 
This commit adds support for 2MiB pages.

The approach taken is given seL4 simplifies each level to a single type IOPT, we add bit 7 to the IOPT bitfield. 
When mapping in a large page seL4 will set this large page bit iff the hardware has reported that the IOMMU supports large pages.

In future it would be nice to provide in the boot info what additional IOMMU pages are supported by the hardware e.g. provide roughly the equivalent of x86KSSecondStageLargePageSupport to user level. However it is not needed at this point since user level can infer it when the kernel rejects the attempt to map a large page.

Screenshot 2026-07-09 at 2 47 39 pm
Screenshot 2026-07-09 at 2 51 00 pm

### DIFF
--- a/include/arch/x86/arch/model/statedata.h
+++ b/include/arch/x86/arch/model/statedata.h
@@ -76,6 +76,7 @@ extern vtd_rte_t *x86KSvtdRootTable;
 extern uint32_t x86KSnumIOPTLevels;
 extern uint32_t x86KSnumIODomainIDBits;
 extern uint32_t x86KSFirstValidIODomain;
+extern uint32_t x86KSSecondStageLargePageSupport;
 #endif
 
 #ifdef CONFIG_PRINTING
--- a/include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
+++ b/include/plat/pc99/plat/32/plat_mode/machine/hardware.bf
@@ -40,7 +40,9 @@ block vtd_pte {
     padding                         32
 
     field_high  addr                20
-    padding                         10
+    padding                         4
+    field       large_page          1
+    padding                         5
     field       write               1
     field       read                1
 }
--- a/include/plat/pc99/plat/64/plat_mode/machine/hardware.bf
+++ b/include/plat/pc99/plat/64/plat_mode/machine/hardware.bf
@@ -37,7 +37,9 @@ block vtd_cte {
 block vtd_pte {
     --- Assume AVAIL and TM as Reserved
     field_high  addr                52
-    padding                         10
+    padding                         4
+    field       large_page          1
+    padding                         5
     field       write               1
     field       read                1
 }
--- a/src/arch/x86/model/statedata.c
+++ b/src/arch/x86/model/statedata.c
@@ -51,6 +51,7 @@ vtd_rte_t *x86KSvtdRootTable;
 uint32_t x86KSnumIOPTLevels;
 uint32_t x86KSnumIODomainIDBits;
 uint32_t x86KSFirstValidIODomain;
+uint32_t x86KSSecondStageLargePageSupport;
 #endif
 
 #ifdef CONFIG_VTX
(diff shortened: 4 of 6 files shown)


lsf37:
This is a straightforward enough extension, but API changes do require an RFC, and there are a lot of questions that come to mind immediately. Pretty much what the RFC template would force answers for.

E.g. why are we adding only large pages and not 1GB pages? Is this an ad-hoc addition to the API just because we need something right now or does this generalise nicely? Should things be refactored? Does this make verification harder or simpler in the future? What needs to be updated in docs and other user-facing code and material? (e.g. C capDL tools, are there any tests in sel4test, etc) Is anyone updating the rest of the existing libraries (seL4_libs etc)? Any other considerations?

cazb2:
I will defer to @dreamliner787-9 on this. I do have tests for this although the remaining requirements I believe don't have anyone working on them.

I don't know much about verification but I would expect adding to VT-d/VTx would make it less simple...

lsf37:
> I don't know much about verification but I would expect adding to VT-d/VTx would make it less simple...

At least on AArch64 and RISC-V we found that having one generic mechanism to deal with all page levels was much faster to verify than special cases for each level. I'm not sure what the state of the x64 IO page tables is in that, but if there is a choice, that is the direction we should go for.

lsf37:
> I will defer to @dreamliner787-9 on this. I do have tests for this although the remaining requirements I believe don't have anyone working on them.

I that's the case, there needs to be some plan for how we get there, otherwise we can't merge something like this.

Indanz (commented):
I haven't looked at this into detail, but my impression is that the page table structures for IOMMU are the same as for virtualisation, is that right?

If so, how come this PR only mentions IO spaces and not virtualisation?

cazb2:
> I haven't looked at this into detail, but my impression is that the page table structures for IOMMU are the same as for virtualisation, is that right?
> 
> If so, how come this PR only mentions IO spaces and not virtualisation?

Yes, the intel VT-D spec explicitly confirms this. imo this is one of the several flaws with the current VT-d/x design, which manages the two things separately.

cazb2:
> > I will defer to @dreamliner787-9 on this. I do have tests for this although the remaining requirements I believe don't have anyone working on them.
> 
> I that's the case, there needs to be some plan for how we get there, otherwise we can't merge something like this.

Yep, understandable. I believe an RFC will be made wrt to this. @dreamliner787-9

Indanz:
> > I haven't looked at this into detail, but my impression is that the page table structures for IOMMU are the same as for virtualisation, is that right?
> > If so, how come this PR only mentions IO spaces and not virtualisation?
> 
> Yes, the intel VT-D spec explicitly confirms this. imo this is one of the several flaws with the current VT-d/x design, which manages the two things separately.

You mean the seL4 implementation? Virtualisation code isn't verified, so should be easy to change. The API change can probably be papered over with backward compatible defines in libsel4. It worked for the AArch64 page table unification, it should work for this too.

No point in just adding it for just IO spaces and leaving virtualisation untouched. Unifying the page table code as much as possible should be done first before adding more features I think.

lsf37:
> > > I haven't looked at this into detail, but my impression is that the page table structures for IOMMU are the same as for virtualisation, is that right?
> > > If so, how come this PR only mentions IO spaces and not virtualisation?
> > 
> > 
> > Yes, the intel VT-D spec explicitly confirms this. imo this is one of the several flaws with the current VT-d/x design, which manages the two things separately.
> 
> You mean the seL4 implementation? Virtualisation code isn't verified, so should be easy to change. 

It should be changed only even more carefully, because mistakes there will not be caught. But yes, it will at least not incur any formal verification cost/effort for now.

> The API change can probably be papered over with backward compatible defines in libsel4. It worked for the AArch64 page table unification, it should work for this too.
> 
> No point in just adding it for just IO spaces and leaving virtualisation untouched. Unifying the page table code as much as possible should be done first before adding more features I think.

If it can be unified, yes, that would be great, I agree.

cazb2:
You mean the seL4 implementation? yup

I believe vt-x already has large page support...

PR #1705, opened 2026-07-10
libsel4: IOSpace capdata with PCI bus/dev/fun

midnightveil:
This generates a helper _new that looks like this:

    seL4_X86_IOSpace_CapData.words[0] = 0
        | (domainID & 0xffffull) << 16
        | (pciBus & 0xffull) << 8
        | (pciDev & 0x1full) << 3
        | (pciFun & 0x7ull) << 0;

, which matches the layout the kernel expects.

In contrast to what we thought when we merged commit b57b3de, seL4 does actually rely on the layout of the PCIDevice field for the 'lookup_vtd_context_slot' function. Thus, to make it easier for userspace to generate the capData for an IOSpace cap we have the capData have fields for bus/dev/fun in the bitfield generator.

In addition, the C capDL loader does pack these values itself with its own hardcoded layout in 'showPCI'; and furthermore the existing kernel API (for X86IRQIssueIRQHandlerMSI) takes 'pci_bus','pci_dev', and 'pci_func' arguments.

This is technically a breaking change over the last commit, but the previous commit was well, the previous commit, so I don't expect it to break anyone. And it doesn't break anyone doing the packing directly themselves.

### DIFF
--- a/libsel4/arch_include/x86/sel4/arch/shared_types.bf
+++ b/libsel4/arch_include/x86/sel4/arch/shared_types.bf
@@ -33,7 +33,9 @@ block seL4_X86_IOSpace_CapData {
     padding 32
 #endif
     field domainID              16
-    field PCIDevice             16
+    field pciBus                8
+    field pciDev                5
+    field pciFun                3
 }
 
 #endif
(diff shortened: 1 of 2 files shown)


lsf37 (approved):

lsf37:
Breaking the API of the previous commit is fine if the previous commit is not a release :)

PR #1707, opened 2026-07-14
smp+mcs: fix bug involving ksCurThread migration

### LINKED ISSUE #941: MCS+SMP: Missing remoteQueueUpdate somewhere?

midnightveil:
We identified a bug where passive-server migration involving the current thread would not cause remote IPIs to be sent.

We fix this by adding a new invariant for SMP that the current thread is always running on the current core, as well as that the current ksSchedulerAction candidate must be a thread on the current core.

Thus, anytime we change the affinity of a thread we need to make sure it is not the current thread, and if so we switch to the idle thread and reschedule ourselves.

This invariant means that the schedule() function will only ever talk about threads running on the current core, and allows several of the TCB_SCHED_ENQUEUE/append calls to be changed to the current-core-only tcbSchedEnqueue call.

This is #1662 with many of the unrelated/debug changes removed. Fixes #941.

### DIFF
--- a/src/kernel/thread.c
+++ b/src/kernel/thread.c
@@ -371,6 +371,15 @@ static void scheduleChooseNewThread(void)
 
 void schedule(void)
 {
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
+    /* These invariants mean we don't need any remoteQueueUpdate calls, and
+       thus can call tcbSchedEnqueue/tcbSchedAppend directly. */
+#endif
 #ifdef CONFIG_KERNEL_MCS
     awaken();
     checkDomainTime();
@@ -380,7 +389,7 @@ void schedule(void)
         bool_t was_runnable;
         if (isSchedulable(NODE_STATE(ksCurThread))) {
             was_runnable = true;
-            SCHED_ENQUEUE_CURRENT_TCB;
+            tcbSchedEnqueue(NODE_STATE(ksCurThread));
         } else {
             was_runnable = false;
         }
@@ -399,7 +408,7 @@ void schedule(void)
                 || (candidate->tcbPriority < NODE_STATE(ksCurThread)->tcbPriority);
             if (fastfail &&
                 !isHighestPrio(ksCurDomain, candidate->tcbPriority)) {
-                SCHED_ENQUEUE(candidate);
+                tcbSchedEnqueue(candidate);
                 /* we can't, need to reschedule */
                 NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
                 scheduleChooseNewThread();
@@ -407,7 +416,7 @@ void schedule(void)
                 /* We append the candidate at the end of the scheduling queue, that way the
                  * current thread, that was enqueued at the start of the scheduling queue
                  * will get picked during chooseNewThread */
-                SCHED_APPEND(candidate);
+                tcbSchedAppend(candidate);
                 NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
                 scheduleChooseNewThread();
             } else {
@@ -479,6 +488,11 @@ void switchToThread(tcb_t *thread)
 
     tcbSchedDequeue(thread);
     NODE_STATE(ksCurThread) = thread;
+
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
+#endif
 }
 
 void switchToIdleThread(void)
@@ -565,7 +579,8 @@ void possibleSwitchTo(tcb_t *target)
         } else if (NODE_STATE(ksSchedulerAction) != SchedulerAction_ResumeCurrentThread) {
             /* Too many threads want special treatment, use regular queues. */
             rescheduleRequired();
-            SCHED_ENQUEUE(target);
+            /* We know that this thread must be on the current core */
+            tcbSchedEnqueue(target);
         } else {
             NODE_STATE(ksSchedulerAction) = target;
         }
@@ -573,6 +588,11 @@ void possibleSwitchTo(tcb_t *target)
     }
 #endif
 
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
+#endif
 }
 
 void setThreadState(tcb_t *tptr, _thread_state_t ts)
@@ -702,8 +722,13 @@ void timerTick(void)
 
 void rescheduleRequired(void)
 {
-    if (NODE_STATE(ksSchedulerAction) != SchedulerAction_ResumeCurrentThread
-        && NODE_STATE(ksSchedulerAction) != SchedulerAction_ChooseNewThread
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: if a thread, ksSchedulerAction belongs to the current core */
+    assert(!SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction)) ||
+           NODE_STATE(ksSchedulerAction)->tcbAffinity == getCurrentCPUIndex());
+#endif
+
+    if (SchedulerAction_IsCandidateThread(NODE_STATE(ksSchedulerAction))
 #ifdef CONFIG_KERNEL_MCS
         && isSchedulable(NODE_STATE(ksSchedulerAction))
 #endif
@@ -712,7 +737,7 @@ void rescheduleRequired(void)
         assert(refill_sufficient(NODE_STATE(ksSchedulerAction)->tcbSchedContext, 0));
         assert(refill_ready(NODE_STATE(ksSchedulerAction)->tcbSchedContext));
 #endif
-        SCHED_ENQUEUE(NODE_STATE(ksSchedulerAction));
+        tcbSchedEnqueue(NODE_STATE(ksSchedulerAction));
     }
     NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
 }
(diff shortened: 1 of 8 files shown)


src/object/tcb.c
```diff
@@ -468,6 +469,7 @@ void remoteTCBStall(tcb_t *tcb)
 
     if (
 #ifdef CONFIG_KERNEL_MCS
+// XX: Why is this correct?
         tcb->tcbSchedContext &&
```
midnightveil, line 473: .

src/object/schedcontext.c
```diff
@@ -211,6 +211,14 @@ static exception_t invokeSchedContext_YieldTo(sched_context_t *sc, bool_t call)
...
             rescheduleRequired();
 
+            /* We know that both the TCB is on the current core per earlier check,
+             * and that ksCurThread is on the current core by invariant.
+             * So neither of these needed a remoteQueueUpdate().
+             * @TODO: Guarantee sc->scCore == tcb->tcbAffinity?
+             */
+            SMP_COND_STATEMENT(assert(tcb->tcbAffinity == getCurrentCPUIndex()));
+            SMP_COND_STATEMENT(assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex()));
+
```
midnightveil, line 221: I think that this TODO is not true because often `ksCurSC` can often lag behind `ksCurThread` and is only updated at the end of schedule.

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
+         */
+        tcb_t *action = NODE_STATE(ksSchedulerAction);
+        if (action == tcb || action == SchedulerAction_ResumeCurrentThread) {
+            NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread;
+        }
+
+        tcb->tcbAffinity = new_core;
+        if (isSchedulable(tcb)) {
+            SCHED_ENQUEUE(tcb);
+        }
```
midnightveil, line 48: I think that this is the right check to use.

src/model/smp.c
```diff
@@ -25,7 +25,30 @@ void migrateTCB(tcb_t *tcb, word_t new_core)
...
+    } else {
+        tcb->tcbAffinity = new_core;
+    }
```
midnightveil, line 51: Something that is not obvious: `migrateTCB` doesn't call `schedEnqueue` when the thread is not the current one, but it doesn't (maybe shouldn't?) need to.

If we compare this with `invokeTCB_SetAffinity` (non-mcs), it also does `isRunnable()` -> `schedAppend`, and `if thread == current`, reschedule.

Callers of `migrateTCB`:

- `schedContext_donate`: This is the primary case where this is useful. Unlike #986 which fixes an instance of this in `receiveIPC`, this change makes `reply_push`, which calls `schedContext_donate`, then perform an IPI if necessary. i.e. it links the reply push/pop handling handle migrating the current thread. This is in essence, the primary reason why this works.

- `schedContext_bindTCB`: This was mostly-fine before as it always does a `SCHED_ENQUEUE` call; this means we do technically call `SCHED_ENQUEUE` twice or the current thread; however it still results in a single IPI, and the enqueue short-circuits if already enqueued.

- `invokeSchedControl_ConfigureFlags`: This has some duplicated logic and does things depending on current thread or not. 

So this can add some duplicated scheduler enqueues, but to remove that would require reworks to verified unicore MCS code.

midnightveil:
The preprocess changes are just for minor AST changes due to placing a check in a macro. I could make this consistent by removing a set of brackets from the macro, I suppose...

lsf37:
> The preprocess changes are just for minor AST changes due to placing a check in a macro. I could make this consistent by removing a set of brackets from the macro, I suppose...

Might be fine as it is, I've started a proof run to see if it breaks.

src/object/schedcontext.c
Indanz: I think we should remove `ksCurSC` and use `ksCurThread->tcbSchedContext` directly instead, but it may be too late for that now. Maintaining the same state twice manually is cumbersome and saving one indirect read isn't worth it I think.

Also, I think `seL4_SchedContext_YieldTo` to cross-core SC should fail. YieldTo makes little sense already, but doing it cross-core really doesn't make any sense.

The call is obscure enough that these asserts really add very little, so I would leave them out. The call doesn't change the SC or current task, so why check this here?

src/object/schedcontext.c
```diff
@@ -332,6 +340,11 @@ void schedContext_bindTCB(sched_context_t *sc, tcb_t *tcb)
         // verification work, so the work around is to use rescheduleRequired()
         //possibleSwitchTo(tcb);
     }
+
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
```
Indanz, line 346: This syscall doesn't change `ksCurThread` and `tcb` can never be running because it had no SC, so this assert seems pointless here.

src/object/tcb.c
Indanz: Probably an optimisation because the check is cheap. It doesn't hurt to have, but I don't think it's necessary.

src/model/smp.c
Indanz: seL4 scheduling is a bit of a mess, and extremely messy on MCS, so I'll need some time to check if the above changes are correct or not. Ideally we would streamline the code instead of just adding more slightly different subtle code, but that's probably wishful thinking.

Below are all existing callers of `migrateTBC`, I'm looking at those first to get a feeling of whether your changes make sense or not:
https://github.com/seL4/seL4/blob/0fcbb71a2a0fb37dabef0e40c97afacc41f7cc71/src/object/tcb.c#L481-L495
This seems correct and fairly straightforward because it is non-MCS. The only argument is whether that code should be in `migrateTCB` or in `invokeTCB_SetAffinity`. But for non-MCS this is the only way to migrate cores, so it doesn't matter.

https://github.com/seL4/seL4/blob/0fcbb71a2a0fb37dabef0e40c97afacc41f7cc71/src/object/schedcontrol.c#L13-L78
Very complex and hard to check for correctness. I don't understand why the code is going through so much pain to maybe call `possibleSwitchTo`, it would be simpler to only do SCHED_ENQUEUE and always reschedule.

Maybe the `else if (isRunnable(thread))` block should be removed after the changes in migrateTCB, or else all callers need to do something equivalent (and then the `SCHED_ENQUEUE(tcb)` in migrateTCB can be removed). `schedContext_bindTCB` does, maybe all callers of `schedContext_donate` also do.

https://github.com/seL4/seL4/blob/0fcbb71a2a0fb37dabef0e40c97afacc41f7cc71/src/object/schedcontext.c#L312-L335
`schedContext_bindTCB` does just the SCHED_ENQUEUE + rescheduleRequired.

https://github.com/seL4/seL4/blob/0fcbb71a2a0fb37dabef0e40c97afacc41f7cc71/src/object/schedcontext.c#L362-L382
` schedContext_donate` seems weird. On one hand it seems to rely on the caller to schedule the `to` thread, and yet it has a `ksCurThread` + `ksSchedulerAction` check. It could have just set `NODE_STATE(ksSchedulerAction) = SchedulerAction_ChooseNewThread` directly. But this code is similar to what you added to migrateTCB, except the reason is SC removal instead of core migration. Ideally this would be in a shared helper function, though that would require a verification update.

For both I fail to see how the thread can be set to `ksSchedulerAction` at this point though. This would require a `possibleSwitchTo` call before the `migrateTCB` call. I checked all callers and it seems okay to me. I'd rather add an assertion that `ksSchedulerAction != tcb` if possible.

Looking at `schedContext_unbindTCB`, all it does is calling `rescheduleRequired()` if `tcb == NODE_STATE(ksCurThread)`. With that in mind, I think your `switchToIdleThread` is overly strict and unnecessary. If code is using `ksCurThread` after it was migrated in the wrong way, we should fix that instead.

All in all it looks like one of the callers `schedContext_donate` might have a missing SCHED_ENQUEUE + rescheduleRequired code path. If possible, we should fix that there instead of adding a big hammer work-around to `migrateTCB`. After checking all callers:

- `receiveIPC`: case `EPState_Send` does `reply_push` which may donate an SC and switch cores for the current thread. SCHED_ENQUEUE is missing for migrating tasks. This actually matches the location @JE-Archer found in #986.
- `finaliseCap` calls `reply_remove` when finalising a reply cap, which may return an SC and doesn't schedule the receiver. However, in this case that seems like the correct behaviour, as there is no actual reply sent and the receiver will stay `ThreadState_BlockedOnReply`.

One of the problem of #986 was the very indirect way of letting `schedule` call `SCHED_ENQUEUE`. That seems easily avoidable by calling SCHED_ENQUEUE + rescheduleRequired directly in `receiveIPC` when the core changed, so my tentative conclusion is that that would be the correct fix. This has my preference to changing `ksCurThread` to the idle thread, as that might break other assumptions elsewhere.

src/model/smp.c
Indanz: ~~This may not be safe in case the SC is postponed.~~

I got confused, `isSchedulable()` checks `!thread_state_get_tcbInReleaseQueue`, but not `thread_state_get_tcbQueued`. I mixed those two up for some reason.

lsf37:
> > The preprocess changes are just for minor AST changes due to placing a check in a macro. I could make this consistent by removing a set of brackets from the macro, I suppose...
> 
> Might be fine as it is, I've started a proof run to see if it breaks.

Proof are all passing (incl MCS C), so no problem on the preprocess checks.

src/kernel/thread.c
```diff
@@ -371,6 +371,15 @@ static void scheduleChooseNewThread(void)
 
 void schedule(void)
 {
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
```
Indanz, line 376: So this one is too strong I think.

src/kernel/thread.c
```diff
@@ -380,7 +389,7 @@ void schedule(void)
         bool_t was_runnable;
         if (isSchedulable(NODE_STATE(ksCurThread))) {
             was_runnable = true;
-            SCHED_ENQUEUE_CURRENT_TCB;
+            tcbSchedEnqueue(NODE_STATE(ksCurThread));
```
Indanz, line 392: And then this can't be done.

src/kernel/thread.c
```diff
@@ -479,6 +488,11 @@ void switchToThread(tcb_t *thread)
 
     tcbSchedDequeue(thread);
     NODE_STATE(ksCurThread) = thread;
+
+#ifdef ENABLE_SMP_SUPPORT
+    /* Invariant: the current thread always belongs to the current core. */
+    assert(NODE_STATE(ksCurThread)->tcbAffinity == getCurrentCPUIndex());
```
Indanz, line 494: This one can stay, as we only want to schedule threads on the current core here.

PR #1708, opened 2026-07-14
cmake: Fix timer frequency export to userspace

midnightveil:
Previously, this would only be exported if you set `-DKernelTimerFrequency=TRUE` in the CMake invocation; but it was intended to always set it. Instead do the conditional based on whether CONFIGURE_TIMER_FREQUENCY exists.


lsf37 (approved):
Looks good.

Indanz (commented):
The domain test in sel4test will fail for x86 if this doesn't work as expected.

I don't really understand how it could work before though, did we just use 1 GHz for all platforms and that worked out fine? Oops.

PR #1709, opened 2026-07-14
aarch32,riscv: exclusive KernelPaddrUserTop

### LINKED ISSUE #1706: CONFIG_PADDR_USER_DEVICE_TOP is inconsistent for 32-bit platforms
`CONFIG_PADDR_USER_DEVICE_TOP` is 2^32-1 on 32-bit platforms, but is supposed to be an exclusive value. This leads to the cascade of untypeds that go up to `(2^32-1) & ~mask untypedSizeBits` on Aarch32 and rv32. x86 seems to still remove the full last page, but I haven't checked that one closely.

This is related to #571 and partially the problems described in #720. In particular the python code in `memory.py` [here](https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/tools/hardware/utils/memory.py#L112-L118) is trying to model that kernel behaviour.

Contrary to the discussion in #571, there would be no problem with an Untyped that wraps, i.e. gets a region `[p, 0)`. The code in `create_untypeds_for_region` will do the correct thing when `end` is 0: https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/kernel/boot.c#L762

Unfortunately, we can't just set CONFIG_PADDR_USER_DEVICE_TOP to 2^32, because this comparison then fails (assuming this is evaluated in `word_t`, if it's not then it's actually fine anyway):

https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/kernel/boot.c#L811

We could change that to 

```c
  if (start <= CONFIG_PADDR_USER_DEVICE_TOP-1) {
```

which would be correct for any value of `CONFIG_PADDR_USER_DEVICE_TOP` (it basically converts it to inclusive end reasoning, which doesn't have these stupid overflow problems, but I'm sure I mentioned that before).

There are no other uses of `CONFIG_PADDR_USER_DEVICE_TOP` that I can see.

So, my proposal would be to set `CONFIG_PADDR_USER_DEVICE_TOP ` to `2^32` and kick out the `as_max` special treatment in python and add comments that we have thought about the overflow issue there.

Since this is not verified code, we should double-check all this, but it should be pretty easy to see if the untyped list that is being printed out is correct or not.

lsf37:
The treatment of `KernelPaddrUserTop` was inconsistent between 32 and 64 bit architectures, because the desired 2^32 is not expressible on 32 bit systems. However, with a small tweak is safe to wrap it to 0 for untyped creation, which already deals gracefully with that case.

- Deal with (paddr_t) CONFIG_PADDR_USER_DEVICE_TOP potentially being 0 in boot.c

- Remove the special treatment of the top of the address space in hardware/utils/memory.py

- Remove now unused config.get_smallest_kernel_object_alignment()

This implements #1706, but merging is blocked until we have figured out what the impact on the Rust side is when `platform_gen.yaml` can produce output like

```yml
devices:
[..] 
- end: 0x100000000
  start: 0xfd000000
```

for 32 bit platforms (end = 2^32 above)

Test with: https://github.com/seL4/seL4_libs/pull/115

### DIFF
--- a/src/arch/arm/config.cmake
+++ b/src/arch/arm/config.cmake
@@ -39,9 +39,8 @@ config_set(KernelArmPASizeBits44 ARM_PA_SIZE_BITS_44 "${KernelArmPASizeBits44}")
 config_set(KernelArmICacheVIPT ARM_ICACHE_VIPT "${KernelArmICacheVIPT}")
 
 if(KernelSel4ArchAarch32)
-  # 64-bit targets may be building in 32-bit mode,
-  # so make sure maximum paddr is 32-bit.
-  math(EXPR KernelPaddrUserTop "(1 << 32) - 1")
+  # Maximum paddr needs to be exclusive. Code is fine when it wraps to 0.
+  math(EXPR KernelPaddrUserTop "1 << 32")
 endif()
 
 include(src/arch/arm/armv/armv7-a/config.cmake)
--- a/src/arch/riscv/config.cmake
+++ b/src/arch/riscv/config.cmake
@@ -66,9 +66,8 @@ if(KernelSel4ArchRiscV32)
 endif()
 if(KernelPTLevels EQUAL 2)
   if(KernelSel4ArchRiscV32)
-    # seL4 on RISCV32 uses 32-bit ints for addresses,
-    # so limit the maximum paddr to 32-bits.
-    math(EXPR KernelPaddrUserTop "(1 << 32) - 1")
+    # Maximum paddr needs to be exclusive. Code is fine when it wraps to 0.
+    math(EXPR KernelPaddrUserTop "1 << 32")
   else()
     math(EXPR KernelPaddrUserTop "1 << 34")
   endif()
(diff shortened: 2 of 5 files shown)


kent-mcleod (approved):

Indanz (approved):

lsf37:
I found a problem after all: user space needs to be able to deal with the fact that an untyped can reach to the end of the address space, and of course the current libs are full of terrible assumptions.

Specifically, ZynqMP is not showing output in release mode, which I think is because the UART is now in that last big untyped and there is something going wrong there somewhere when it tries to map it from that untyped. I'll try to track it down, but it does demonstrate that this is a potentially more breaking change than I thought (only for arguably buggy user space code, but still).

x86 still has its old definition, so this is relevant for rv32 and aarch32 user code.

lsf37:
See https://github.com/seL4/seL4_libs/pull/115 for that. I should add that this is only relevant for user space that is tracking what physical addresses an untyped covers (which is a normal thing to want to do, of course).

lsf37:
The Rust side is updated as well now, so I'll merge this one the build has passed after branch update.

PR #1710, opened 2026-07-14
Add extra DSB to seL4_ARM_Page_Unify_Instruction

KurtWu10:
EDIT: this applies to aarch32 as well, and also targets functions `seL4_ARMPDUnify_Instruction()` and `seL4_ARMPageUnify_Instruction()`.

This PR fixes a missing dsb during I-cache invalidation for all addresses on aarch64. It is similar to the commit https://github.com/seL4/seL4/commit/21888e6a954d5f6853514cf4710f2115a773afb0.

I didn't try to reproduce the issue on real hardware.

## Description
The erroneous execution occurs when functions `seL4_ARM_Page_Unify_Instruction()` or `seL4_ARM_VSpace_Unify_Instruction()` are called, the kernel will perform the following cache maintenance operations (`dc cvau -> dmb -> dsb -> ic iallu -> isb -> isb`):
https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/arch/arm/64/kernel/vspace.c#L1128-L1139

these include calling the function `invalidateCacheRange_I()`:
https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/arch/arm/machine/cache.c#L135-L148

and for processors with VIPT I-cache (e.g. Cortex A35, A53 and A55) and the kernel is built as a hypervisor, it calls the function `invalidate_I_PoU` https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/include/arch/arm/arch/64/mode/machine.h#L305-L313

This is not sufficient to guarantee that by the time `isb` (or other context synchronization events, e.g. exception return) is executed, the effects of i-cache invalidation instructions have completed. Therefore, a userspace application calling the unification functions has to perform another `dsb` followed by an `isb`, before it can guarantee that the unification operation has completed for later instructions of itself.

## Justification
Instructions for I-cache invalidation can be checked by the [herd](https://developer.arm.com/herd7) memory model tool (7.58). With the following program adapted from [SM.B%2Bcachesync-isb.litmus](https://github.com/herd/herdtools7/blob/ArmARM-M.c/catalogue/aarch64-ifetch/tests/DIC0-IDC0/SM.B%2Bcachesync-isb.litmus),
```
AArch64 SM.B+cachesync-isb
variant=ifetch
{
  0:X0=NOP; 0:X1=P0:m0; 0:X9=0;
}
P0              ;
    STR W0,[X1] ;
    DC CVAU,X1  ;
    DSB ISH     ;
    IC IALLU    ;
    DSB ISH     ;
    ISB         ;
m0: B l0        ;
    MOV W9,#1   ;
l0:             ;
exists 0:X9=0
```
the herd tools should return a single consistent execution, indicating that the instruction written by the first line will be observed by the branch instruction at `m0`. If the `dsb` instruction after `ic iallu` is removed, the tool should return an additional consistent execution that observes the old instruction.

## Note
I found the use of `dmb` and `dsb` in the kernel to be messy. For example, the function `invalidate_I_PoU()` is used in other places
https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/arch/arm/machine/cache.c#L196-L208

and this PR adds an unnecessary (and costly) `dsb` call to these well-synchronised executions.

### DIFF
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -1788,8 +1788,10 @@ static void doFlush(int invLabel, vptr_t start, vptr_t end, paddr_t pstart)
         /* ...then invalidate the corresponding instruction lines
            to point of unification... */
         invalidateCacheRange_I(start, end, pstart);
-        /* ...then invalidate branch predictors. */
+        /* ...then invalidate branch predictors... */
         branchFlushRange(start, end, pstart);
+        /* ... then wait for the completion of invalidations. */
+        dsb();
         /* Ensure new instructions come from fresh cache lines. */
         isb();
         break;
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1134,6 +1134,8 @@ static void doFlush(word_t invLabel, vptr_t start, vptr_t end, paddr_t pstart)
         /* ...then invalidate the corresponding instruction lines
            to point of unification... */
         invalidateCacheRange_I(start, end, pstart);
+        /* ... then wait for the completion of invalidations... */
+        dsb();
         /* ... and ensure new instructions come from fresh cache lines. */
         isb();
         break;


lsf37:
Nice find.


> I found the use of `dmb` and `dsb` in the kernel to be messy. For example, the function `invalidate_I_PoU()` is used in other places 
> https://github.com/seL4/seL4/blob/df6cb8475392a5adbdbc2ba0882decf13ca66103/src/arch/arm/machine/cache.c#L196-L208
> 
> and this PR adds an unnecessary (and costly) `dsb` call to these well-synchronised executions.

Would it make sense to provide a boolean parameter that says whether `dsb` is necessary?

KurtWu10:
> Would it make sense to provide a boolean parameter that says whether dsb is necessary?

I think that’s a good idea.

KurtWu10:
I don't know whether this applies to 32-bit Arm as well. This breaks the compilation. I will do more research on it.

lsf37:
This will need a small (hopefully quick) verification update. Let me know if/when you know whether this applies to AArch32 as well.

KurtWu10:
It does apply to AArch32, as described in the latest commit.

For AArch32, in the current implementation, invalidations of the instruction cache and branch predictors are serialised instead of being executed in parallel. This is one (among many other) potential optimisations.

Indanz (commented):
Please choose between doing the dsb+isb inside or outside of `invalidate_I_PoU` and stick to that. There aren't that many callers.

KurtWu10:
I prefer doing dsb+isb outside of the function, and relying on the effect of an exception return for context synchronisation.

I also took this as an opportunity to remove some unused isb instructions, with the assumption that the kernel does not execute instructions from memory that has just been unified before a context switch.

I only consider the `invalidate_I_PoU` function and its call sites in the current implementation. I do not consider the `invalidateByVA_I` function, although we can perform some similar optimisations.

lsf37:
> I prefer doing dsb+isb outside of the function, and relying on the effect of an exception return for context synchronisation.

I'd be Ok with that. It's a bit more noise on the verification side, but I need to make one pass through this anyway, so it shouldn't matter too much.

 
> I also took this as an opportunity to remove some unused isb instructions, with the assumption that the kernel does not execute instructions from memory that has just been unified before a context switch.

Yes, kernel memory should never be able to be affected by any of these and should never be executing instructions from memory and mappings that user-level can affect in any way. As long as we are sure the user sees the right instructions when we return, I think this should be fine.


> I only consider the `invalidate_I_PoU` function and its call sites in the current implementation. I do not consider the `invalidateByVA_I` function, although we can perform some similar optimisations.

That's fine, we can do them some other time, but could you please open an issue for that, linking to this PR and discussion?

KurtWu10:
Sure.

lsf37:
So reading this now, this is a lot less obviously safe than it was before. Maybe we should split out fixing the issue the tool has identified from making optimisations. Adding a barrier is safe, removing is not, and we should have some validation for those optimisations, which is much more precise if we don't mix in a fix for a previous issue.

kent-mcleod:
When adding new dsb after `IC IALLU` on aarch64 code paths, note that the kernel's `dsb()` function is for a `dsb sy` rather than `dsb ish`.

KurtWu10:
Do we want to introduce a new `dsb_ish` function for that?

I agree that ISH is the most appropriate shareability domain for SMP and MK, and we can even use NSH for unicore.

For the litmus test above in cat, herd assumes inner-shareable and that's why the code uses `dsb ish`. To the best of my knowledge, the use of other shareability domains is orthogonal to the completion of IC effects, as long as the domain includes CPU cores of interest.

KurtWu10:
I prefer my latest approach over my previous approach, because the aarch32 and aarch64 implementations are consistent now, and the involved synchronisation instructions are more explicit.

Indanz (approved):

Indanz:
> Do we want to introduce a new `dsb_ish` function for that?

I think we do.

> For the litmus test above in cat, herd assumes inner-shareable and that's why the code uses `dsb ish`. To the best of my knowledge, the use of other shareability domains is orthogonal to the completion of IC effects, as long as the domain includes CPU cores of interest.

Yes, but there are other writes done before this code, `dbs sy` versus `dsb ish` may cause a significant timing difference. Because of the earlier `dsb sy` in the same function, there probably isn't any difference here though. Would be interesting to time.

KurtWu10:
That's reasonable. Should I implement `dsb_ish` in this PR or create another PR?

Indanz:
> That's reasonable. Should I implement `dsb_ish` in this PR or create another PR?

It could be part of an optimisation PR, I wouldn't do it here.

Indanz (commented):
You need to change the commit message now that you also update AArch32.

lsf37:
Looks also fine from the verification side (modulo commit message update), I just need to put through a small spec update to register the additional `dsb()`.

KurtWu10:
I've updated the commit message, is this OK now?

Indanz:
Commit title and message could be better still.

> #### aarch32,aarch64: add missing dsb

Proposal: "ARM: add missing DSB to seL4_ARM_Page_Unify_Instruction"

> This commit adds a missing dsb during i-cache invalidation.

and drop this line.

> According to the Arm manual (DDI 0487 M.c) section B2.3,

That's way too coarse of a reference, that covers 32 pages.

> an IC effect generated
> by i-cache invalidation instructions requires a dsb to guarantee that
> it has been completed before the execution of the following context
> synchronization events, like an isb or a return from an exception,

The whole point of context synchronisation/isb is to wait for completion,
so this sentence doesn't make sense.

But the issue here is that the instruction completion is not enough,
the memory effects of that instructions may also not be re-ordered
with later instruction fetch memory operations.

ISB serialises instructions, it is not a memory barrier. The DSB is needed
to avoid speculative memory accesses across the cache operation. The ISB
is needed to make sure that newer instructions are re-fetched from the point
of coherence/RAM instead of using old data.

(That's why there is a DSB between the data clean and I cache invalidate,
to avoid re-ordering those two. If that would happen a speculative fetch could
cache stale instructions. I suspect that that DSB could be replaced with a DMB.)

> such that later instructions can see the effects of i-cache
> invalidation. A similar pattern exists in section B2.7.4.2 of the
> manual. This applies to ARMv7 as well.

This is a good reference, I think I'd drop the previous one only keep
this one, except if you can find a more exact one in the general
section too.

KurtWu10:
> I suspect that that DSB could be replaced with a DMB.

I was referring to the 'Completed IC Effects' in the manual. A DMB is insufficient because roughly speaking, it only talks about the ordering, so by the time later instructions (e.g.. isb) are executed, the broadcast invalidation message sent by the previous i-cache invalidation message may not have reached the instruction cache; the DSB guarantees that the invalidation message is complete by the destination.

An implementation that uses DMB after IC will fail the litmus test that I mentioned at the top using the memory model tool.

There's an alternative (earlier) semantics available in [this](https://cambium.inria.fr/~maranget/papers/esop2020-extended.pdf) ESOP'20 paper.

KurtWu10:
What about the current one? (sorry there's a typo in the commit message above)

Indanz:
> I was referring to the 'Completed IC Effects' in the manual. 

Yes, I missed that, I thought you were talking about the invalidate instruction itself.

> > I suspect that that DSB could be replaced with a DMB.
> 
> A DMB is insufficient because roughly speaking, it only talks about the ordering, so by the time later instructions are executed, the broadcast invalidation message sent by the previous i-cache invalidation message may not have reached the instruction cache; the DSB guarantees that the invalidation message is complete by the destination. An implementation that uses DMB after IC will fail the litmus test that I mentioned at the top using the memory model tool.

Thanks for the reminder. I meant the first DSB, not the one you added. This seems to pass:
```
AArch64 SM.B+cachesync-isb
variant=ifetch
{
  0:X0=NOP; 0:X1=P0:m0; 0:X9=0;
}
P0              ;
    STR W0,[X1] ;
    DC CVAU,X1  ;
    DMB         ;
    IC IALLU    ;
    DSB ISH     ;
    ISB         ;
m0: B l0        ;
    MOV W9,#1   ;
l0:             ;
exists 0:X9=0
```

KurtWu10:
~~Right, it could be, should I change my implementation as well?~~

No, it returns
```
Test SM.B+cachesync-isb Allowed
States 2
0:X9=0;
0:X9=1;
Ok
Witnesses
Positive: 1 Negative: 1
Flag Shareability-argument-for-DMB-is-deprecated
Condition exists (0:X9=0)
Observation SM.B+cachesync-isb Sometimes 1 1
Time SM.B+cachesync-isb 0.18
Hash=d3b5c2263d89661bebb02ca4b1edfb55
```

for that program. I am using the web interface (and a local installation of herdtools 7.58).

Indanz:
> Right, it could be, should I change my implementation as well?

No, I'm not sure if I did it correctly, it seems to fail now. I think I have to specify the optional condition to DMB too for the tool to work, so I think it failed to parse and used the old version instead. (This was on the web version.)

Indanz:
By the way, adding that test and a link to the herd tool would be great documentation to have in a commit. :-)

Indanz:
Ready to go now, just waiting on verification.

Edit: Except title is too long. Maybe "Add missing DSB to seL4_ARM_Page_Unify_Instruction"?

Edit2: That's still one character too long, "Add extra DSB to seL4_ARM_Page_Unify_Instruction"?

midnightveil:
You'll need to rebase on latest master otherwise this can't be merged, FYI.

lsf37:
The verification PR is up. @KurtWu10 can you please rebase so we can include this in the release?

KurtWu10:
Done.

PR #1714, opened 2026-07-21
boot: fix potential unsigned underflow in x86 ELF loader

Brine28:
Fix ELF boot-module segment validation to prevent unsigned underflow and memory corruption.


Indanz:
Does this solve a practical problem you ran into or is this purely theoretical?

If you want to refactor code, having one ELF loading implementation for all archs would be a more effective way of spending your time.

lsf37:
The elf loader runs on trusted input (if you can change what the elf loader gets, you can just as well load something completely different), so doing things like preventing underflow in this path is mostly useful for debugging only. 

Given that, could the same be achieved more easily with asserts?

(Edit: to clarify, I'm not against more boot time checking, making debugging easier is good)

Brine28:
Thanks for the explanation. I agree that the normal boot path generally relies on trusted inputs, but the loader still has to handle faulty or unexpected data from firmware/bootloaders or accidental corruption. In that sense, it's more of a robustness and debugging improvement than a new security frontier. assert helps during development, but runtime checks are still useful because they prevent the propagation of invalid conditions and explicitly indicate the error in release-style builds as well. My apologies if I wrote this strangely, as English isn't my native language.

Brine28:
I found a robustness issue in the x86 boot time ELF loader  the loader didnt validate that each program header’s memory size was at least its file size before zero filling the tail of the segment. In malformed cases  this could lead to an unsigned underflow and an oversized zero fill operation. I changed the loader to validate segment sizes and address wraparound  and to reject invalid images safely instead of continuing with invalid state

src/arch/x86/32/kernel/elf.c
```diff
@@ -58,11 +74,20 @@ BOOT_CODE void elf_load(Elf32_Header_t *elfFile, seL4_Word offset)
...
         src = (paddr_t)elfFile + phdr[i].p_offset;
         dst = phdr[i].p_vaddr + offset;
         len = phdr[i].p_filesz;
         memcpy((void *)dst, (char *)src, len);
-        dst += len;
-        memset((void *)dst, 0, phdr[i].p_memsz - len);
+        if (phdr[i].p_memsz > len) {
+            dst += len;
+            memset((void *)dst, 0, phdr[i].p_memsz - len);
+        }
```
midnightveil, line 82: I think this change is unnecessary; because of the check in `elf_getMemoryBounds` *and* earlier in this function, it should be true that `assert(phdr[i].p_memsz <= len)`. I'd think it makes more sense to assert here; since this not being true would imply the boot code has corrupted the elf in the mean time. (very bad!)

src/arch/x86/32/kernel/elf.c
```diff
@@ -58,11 +74,20 @@ BOOT_CODE void elf_load(Elf32_Header_t *elfFile, seL4_Word offset)
 
     /* loop through all program headers (segments) and load them */
     for (i = 0; i < elfFile->e_phnum; i++) {
+        if (phdr[i].p_memsz < phdr[i].p_filesz) {
+            printf("ELF segment has file size larger than memory size\n");
+            return false;
+        }
```
midnightveil, line 73: This should also not ever be the case, similarly because of the getMemoryBounds? Then elf_load can't fail, and you don't need to make it `bool_t`.

src/arch/x86/64/kernel/elf.c
```diff
@@ -31,22 +31,38 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf64_Header_t *elf)
...
-                elf_reg.start = sect_start;
-            }
-            if (sect_end > elf_reg.end) {
-                elf_reg.end = sect_end;
-            }
+        if (phdr[i].p_memsz == 0) {
+            continue;
+        }
+
+        if (phdr[i].p_memsz < phdr[i].p_filesz) {
```
midnightveil: Same comments apply to 64-bit version as 32-bit version.

src/arch/x86/32/kernel/elf.c
```diff
@@ -33,22 +33,38 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf32_Header_t *elfFile)
...
-                elf_reg.start = sect_start;
-            }
-            if (sect_end > elf_reg.end) {
-                elf_reg.end = sect_end;
-            }
+        if (phdr[i].p_memsz == 0) {
+            continue;
+        }
+
+        if (phdr[i].p_memsz < phdr[i].p_filesz) {
```
midnightveil: I know I made the comments about `elf_load` not needing to be changed because of this, but now having read the whole thing I wonder why was `elf_getMemoryBounds()` changed at all? Given it never looked at `p_filesz` in the first place.

It might make sense to keep the if-check in `elf_load` instead of here for the `p_memsz < p_filesz` error condition.

midnightveil (commented):
Commit messages could be more explicit that some only apply to 32-bit and others are 64-bit. They also need a description; they'll fail gitlint otherwise.

Brine28:
Thanks for taking the time to review this in detail
I noticed your comments seem to go in two different directions on where the p_memsz < p_filesz check should live: the earlier comments suggest keeping it in elf_getMemoryBounds and using an assert in elf_load instead, while your last comment suggests the opposite keeping the check in elf_load and questioning why elf_getMemoryBounds was touched at all
Both make sense to me individually but Id rather not guess and end up going back and forth Could you confirm which one youd prefer?
My own leaning (happy to be corrected) keep the check in elf_getMemoryBounds since thats the function that decides whether the image is accepted or rejected in the first place Then elf_load can rely on that invariant already being enforced and just assert(phdr[i].p_memsz >= phdr[i].p_filesz) instead of checking and returning false again  that would also mean the return type doesnt need to change to bool_t.

src/arch/x86/32/kernel/elf.c
```diff
@@ -33,22 +33,38 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf32_Header_t *elfFile)
...
-            sect_start = phdr[i].p_vaddr;
-            sect_end = sect_start + phdr[i].p_memsz;
-            if (sect_start < elf_reg.start) {
-                elf_reg.start = sect_start;
-            }
-            if (sect_end > elf_reg.end) {
-                elf_reg.end = sect_end;
-            }
+        if (phdr[i].p_memsz == 0) {
+            continue;
```
midnightveil, line 37: I believe continues are in issue for the C verification toolchain (@lsf37 ?) (

midnightveil:
> My own leaning (happy to be corrected) keep the check in elf_getMemoryBounds since thats the function that decides whether the image is accepted or rejected in the first place Then elf_load can rely on that invariant already being enforced and just assert(phdr[i].p_memsz >= phdr[i].p_filesz) instead of checking and returning false again that would also mean the return type doesnt need to change to bool.

I was leaning that way initially, however, before your changes `elf_getMemoryBounds()` could only fail if there was no loadable segments. Also, by the name of the function, it's getting the memory bounds. So checking the filesz makes no sense there.

So I thought it would be nicer to keep the `bool_t elf_load` change, and make `elf_load` check the filesz/memsz thing directly. But I am not sure what is better :/

Brine28:
Now?

src/arch/x86/32/kernel/elf.c
lsf37: They are an issue for the binary verification pass, the C parser itself is fine with it. x64 has a few more issues that would need to be solved for binary verification, so I'd be Ok with this one for once (it also has another continue somewhere in the boot code).

PR #1715, opened 2026-07-22
aarch32,hyp: fix PAR save/restore asm

lsf37:
Fix asm block declaration for restoring the previously saved PAR. It previously used ("=r") instead of input ("r").

Also drop PURE attribute now that the function has side effects.

### DIFF
--- a/include/arch/arm/arch/32/mode/machine_pl2.h
+++ b/include/arch/arm/arch/32/mode/machine_pl2.h
@@ -118,7 +118,7 @@ static inline void invalidateHypTLB(void)
     isb();
 }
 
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

Indanz (approved):
The PURE can stay, function still has no side-effects from the compiler's point of view.

Good find on the other changes.

lsf37:
> The PURE can stay, function still has no side-effects from the compiler's point of view.

I'd rather be conservative on that one and keep it like on AArch64. You're right that end-to-end there is no side effect, but I don't think PURE is fine with save-then-restore global state in general. Most of it is happening on the stack, but the register we're saving/restoring global (hence the entire save/restore).

PR #1718, opened 2026-07-22
github: avoid strategy.job-index as key

lsf37:
Replace `strategy.job-index` by something else that identifies which board the job is going to run, because `strategy.job-index` can vary between workflow runs when GitHub expands the run matrix in a different order.


lsf37:
(For after the release)

.github/workflows/proof.yml
```diff
@@ -36,7 +36,9 @@ jobs:
             features: MCS
             session: CRefine
     # test only most recent push to PR:
-    concurrency: seL4-PR-C-proofs-pr-${{ github.event.number }}-idx-${{ strategy.job-index }}
+    concurrency:
+      group: seL4-PR-C-proofs-pr-${{ github.event.number }}-idx-${{ matrix.arch }}-${{ matrix.features }}
```
midnightveil: does idx make sense to still be here?

midnightveil:
btw, typo in the commit and pr name.

.github/workflows/proof.yml
lsf37: better to remove it, yes

Indanz (approved):

PR #1720, opened 2026-07-27
Fix/x86 elf load bounds check

Brine28:
Summary
elf_load() in the x86 boot code trusted the ELF program headers of
the boot module without validating them, and loaded every program
header regardless of its type. This PR adds proper validation before
loading, and merges the 32-bit/64-bit elf.c implementations that were
otherwise identical.

 Problem
- If a segment's `p_filesz` was greater than its `p_memsz`, the
  BSS-clearing step computed `p_memsz - p_filesz` as an unsigned
  subtraction, which underflows and wraps to a huge value. The
  subsequent `memset()` would then write far beyond the intended
  segment, corrupting arbitrary memory.
- Segments were copied to `p_vaddr + offset` without checking that
  they actually fall within the image region computed by
  `elf_getMemoryBounds()`.
- `elf_checkFile()` only checked the ELF magic and class, nothing
  else (e_machine, e_type, e_phentsize, whether the program header
  table and segment file ranges are even inside the boot module).
- Every program header was loaded regardless of `p_type`, instead of
  only `PT_LOAD` segments.

 Changes (2 commits)

1. `x86: elf.c: merge 32-bit and 64-bit into one file`**
Pure move, no behavioural change. `src/arch/x86/32/kernel/elf.c` and
`src/arch/x86/64/kernel/elf.c` were byte-for-byte identical except
for struct field ordering and one literal — merged into a single
`src/arch/x86/kernel/elf.c` using the mode-independent
`Elf_Header_t`/`Elf_Phdr_t` aliases already defined in `elf.h`.

2. `x86: elf: validate boot module before loading`**
- `elf_checkFile()` now also takes the boot module's size and
  validates: the ELF header fits within it, `e_machine`/`e_type` are
  sane for the build, `e_phentsize` matches `Elf_Phdr_t`, and the
  program header table plus every `PT_LOAD` segment's file range lie
  within the module. All checks use subtraction against the known
  size instead of add-then-check-overflow.
- `elf_getMemoryBounds()`/`elf_load()` now only consider `PT_LOAD`
  segments.
- `elf_load()` rejects a malformed segment (bad `p_filesz`/`p_memsz`,
  or out of the expected image region) and returns `bool_t`, so
  `load_boot_module()` can detect and abort instead of silently
  continuing to boot with a corrupted kernel window.


include/kernel/sporadic.h
```diff
@@ -46,7 +46,8 @@
  * than 2^62. */
 #define MAX_PERIOD_US (getMaxUsToTicks() / 8)
 #endif /* CONFIG_KERNEL_STATIC_MAX_PERIOD_US != 0 */
-#define MAX_RELEASE_TIME (UINT64_MAX - 5 * usToTicks(MAX_PERIOD_US))
```
midnightveil: This seems like some rebase issue.

src/arch/x86/kernel/boot_sys.c
```diff
@@ -136,7 +136,9 @@ BOOT_CODE static paddr_t load_boot_module(word_t boot_module_start, paddr_t load
         (void *)boot_state.ui_info.p_reg.start,
         boot_state.ui_info.p_reg.end - boot_state.ui_info.p_reg.start
     );
-    elf_load(elf_file, boot_state.ui_info.pv_offset);
+    if (!elf_load(elf_file, boot_state.ui_info.pv_offset, v_reg)) {
+        return 0;
```
midnightveil, line 142: I think it would be a good idea to add a `printf` here like in the other return cases of `load_boot_module` for failure.

src/arch/x86/64/kernel/elf.c
```diff
@@ -55,11 +56,42 @@ BOOT_CODE void elf_load(Elf64_Header_t *elf, seL4_Word offset)
...
+        /* Re-derive this segment's virtual range and check it against the
+         * region that elf_getMemoryBounds() computed for the whole image
+         * (image_region), guarding against vaddr/memsz overflow and against
+         * a segment landing outside the physical memory that was actually
+         * allocated for this image. */
+        vptr_t seg_start = phdr[i].p_vaddr;
+        vptr_t seg_end = seg_start + phdr[i].p_memsz;
+        if (seg_end < seg_start ||
+            seg_start < image_region.start || seg_end > image_region.end) {
+            printf("ELF load failed: segment %d lies outside of the expected "
```
midnightveil: Do you think you could print the segment / image bounds information here?

src/arch/x86/64/kernel/elf.c
```diff
@@ -55,11 +56,42 @@ BOOT_CODE void elf_load(Elf64_Header_t *elf, seL4_Word offset)
...
+        }
+
+        /* A corrupt/malformed ELF could claim more file bytes than memory
+         * bytes for a segment. p_memsz - p_filesz is computed below to zero
+         * the BSS portion of the segment; if p_filesz > p_memsz that
+         * subtraction underflows (unsigned) and turns into an enormous
+         * memset, corrupting memory well outside the intended segment.
+         * Reject such images instead of loading them. */
+        if (phdr[i].p_filesz > phdr[i].p_memsz) {
+            printf("ELF load failed: segment %d has p_filesz > p_memsz\n", i);
```
midnightveil: (ditto printing here)

midnightveil (commented):
That's nicer, just a few minor cleanup nits.

Context for this: https://github.com/seL4/seL4/pull/1714 <--- old PR here 

A PR description would have been nice, too.

include/kernel/sporadic.h
Brine28: I don't know where the sporadic.h and boot.c files came from I actually only modified four files. I clicked the sync fork button, maybe that's why. I did a force push to pass the gitlint test, so these 2 files were accidentally included.

Brine28:
@midnightveil Thanks for the review Addressed everything.
 Added a PR description.
 The `sporadic.h` / `boot.c` changes were accidental — they came in from
  clicking "Sync fork" before a force-push, not from this branch's actual
  work. Rebuilt the branch cleanly from `upstream/master` so only the 4
  intended files remain in the diff now.
 Added a `printf` in `boot_sys.c` for the `elf_load()` failure case,
  matching the other failure paths in `load_boot_module()`.
 Both `elf_load()` error messages now print the actual values
  (`p_filesz`/`p_memsz`, and the segment range vs. the expected image
  bounds) instead of just a generic message.

Let me know if anything else needs changing.

src/arch/x86/32/kernel/elf.c
```diff
@@ -58,11 +59,45 @@ BOOT_CODE void elf_load(Elf32_Header_t *elfFile, seL4_Word offset)
...
+        }
+
+        /* A corrupt/malformed ELF could claim more file bytes than memory
+         * bytes for a segment. p_memsz - p_filesz is computed below to zero
+         * the BSS portion of the segment; if p_filesz > p_memsz that
+         * subtraction underflows (unsigned) and turns into an enormous
+         * memset, corrupting memory well outside the intended segment.
+         * Reject such images instead of loading them. */
+        if (phdr[i].p_filesz > phdr[i].p_memsz) {
+            printf("ELF load failed: segment %d has p_filesz (0x%lx) > p_memsz (0x%lx)\n",
```
midnightveil: The kernel has like, `seL4_PRIxWord` macros to use so you don't need to do this variety of casting et al.

Brine28:
@midnightveil I hope everything is as it should be.

midnightveil (approved):

Brine28:
I need to fix the gitlint error.

Indanz (changes requested):
There already is an existing `elf_checkFile` function, please add all new checks there. 

No need for new printfs to explain in details what's wrong for every case separately. The exact corruption isn't interesting, you just want to detect bad files.

Please combine the 32 and 64 bit x86 elf.c files into one (in a separate commit) before adding more code.

The current code is extremely simplistic and simple and does everything in around 60 lines. You add 30 lines for basically just a `p_filesz > p_memsz` check and some indirect overflow checks, that doesn't seem very effective and somewhat arbitrary, compared to everything else that can be wrong. Examples:

- `e_phoff` could be outside the file.
- `e_phentsize` may not match our `Elf_Phdr_t` size.
- Heck, the ELF file might be for the wrong arch or be the type altogether.
- Ridiculous alignments. (But we ignore those anyway.)

The code doesn't even do the most basic things like only loading `PT_LOAD` segments.

I'm not saying we should add checks for everything and so on, but if you don't trust `p_filesz` and `p_memsz`, then you shouldn't trust all the information leading to those either.

Overflows should be avoided early instead of detected indirectly later.

Edit: Also, most current checks are done in `load_boot_module()`, you can also add more seL4 specific checks there. This is for stuff that could be valid ELF, but we don't support or that would break some kind of assumptions seL4 makes.

Brine28:
@Indanz thank you for the detailed feedback. I am currently in the UK and will return my country on the 29th of this month. I will address all your comments and restructure the PR as soon as I get back.

Indanz:
> I will address all your comments and restructure the PR as soon as I get back.

Please do a force push to this branch to update the PR instead of re-opening a new one, thanks.

Brine28:
Actually I previously mentioned Id deploy the patch on the 29th since I dont have my PC. However I managed to get it ready early and I'm pushing it now via Termux. @Indanz

src/arch/x86/config.cmake
```diff
@@ -342,6 +342,7 @@ add_sources(
          kernel/xapic.c
          kernel/x2apic.c
          kernel/boot_sys.c
+         kernel/elf.c
```
Indanz, line 345: Thanks for preserving git history by doing a `git mv` on one of the files.

In this case there isn't anything interesting happening in the other file, so it's not worth retaining. But
in future, if you want to preserve history of both source files before merging them, it is possible in git, just a bit fiddly:

- Create two branches, on for each file you want to merge.
- In each branch, do one `git mv`.
- Merge the two branches. Here you can solve any merge conflicts by keeping the code you want and removing the code you don't want to move (but some of this can also be done before the merge).

This trick can also be used to preserve history when you want to refactor code and move some code from one file to another while preserving history. In that case when merging the branch, you retain the source file, except for the part you want to move. Same for the destination file, you retain the original content, but just add the new code.

src/arch/x86/kernel/elf.c
```diff
@@ -9,26 +9,26 @@
...
-BOOT_CODE bool_t elf_checkFile(Elf32_Header_t *elfFile)
+BOOT_CODE bool_t elf_checkFile(Elf_Header_t *elfFile)
 {
     return (
                elfFile->e_ident[0] == '\177' &&
                elfFile->e_ident[1] == 'E'    &&
                elfFile->e_ident[2] == 'L'    &&
                elfFile->e_ident[3] == 'F'    &&
-               elfFile->e_ident[4] == 1
+               elfFile->e_ident[4] == (sizeof(word_t) == 8 ? 2 : 1)
```
Indanz: I'd add defines for `ELFCLASS32` and `ELFCLASS64` and use that instead of 1 and 2, then at least it's clear what the numbers mean.

src/arch/x86/kernel/elf.c
```diff
@@ -9,26 +9,26 @@
...
 {
-    Elf32_Phdr_t *phdr = (Elf32_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
+    Elf_Phdr_t *phdr = (Elf_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
     v_region_t elf_reg;
     vptr_t     sect_start;
     vptr_t     sect_end;
     uint32_t   i;
 
-    elf_reg.start = 0xffffffff;
+    elf_reg.start = (vptr_t)-1;
```
Indanz: Suggestion: `elf_reg.start = ~0UL;`

src/arch/x86/kernel/elf.c
```diff
@@ -48,13 +48,13 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf32_Header_t *elfFile)
...
-    Elf32_Phdr_t *phdr = (Elf32_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
-    paddr_t       src;
-    paddr_t       dst;
-    uint32_t      len;
-    uint32_t      i;
+    Elf_Phdr_t *phdr = (Elf_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
+    paddr_t    src;
+    paddr_t    dst;
+    word_t     len;
+    uint32_t   i;
```
Indanz, line 81: This now seems aligned on the `*` instead of the `p`, was that intentional?

Brine28:
@Indanz I forgot to press the re-request button.

src/arch/x86/kernel/elf.c
```diff
@@ -6,18 +6,64 @@
...
+    if (elfFile->e_phentsize != sizeof(Elf_Phdr_t) || elfFile->e_phnum == 0) {
+        return false;
+    }
+
+    /* The program header table must be entirely within the boot
+     * module. e_phnum is 16 bits and sizeof(Elf_Phdr_t) is small, so
+     * this multiplication cannot overflow word_t; the bounds check
+     * itself is written with subtraction so that it cannot overflow
+     * either. */
+    phdr_table_size = (word_t)elfFile->e_phnum * sizeof(Elf_Phdr_t);
```
Indanz: No need for the cast.

src/arch/x86/kernel/elf.c
```diff
@@ -6,18 +6,64 @@
...
+    if (elfFile->e_phoff > max_length || phdr_table_size > max_length - elfFile->e_phoff) {
+        return false;
+    }
+
+    /* Each segment's file contents must also be entirely within the
+     * boot module. We don't trust p_filesz/p_memsz later on unless we
+     * also don't trust what leads to them here. */
```
Indanz: I'd remove the last sentence, it doesn't add much and is confusing.

src/arch/x86/kernel/elf.c
```diff
@@ -28,12 +74,12 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile)
     vptr_t     sect_end;
     uint32_t   i;
 
-    elf_reg.start = (vptr_t)-1;
+    elf_reg.start = (vptr_t) -1;
```
Indanz: Stray space. (And better ~0UL.)

src/arch/x86/kernel/elf.c
```diff
@@ -48,21 +94,43 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile)
...
+         * subtraction rather than computing p_vaddr + p_memsz, so the
+         * checks themselves cannot overflow. elf_checkFile() already
+         * validated the file-offset side of things. */
+        if (phdr[i].p_filesz > phdr[i].p_memsz ||
+            phdr[i].p_vaddr < image_region.start ||
+            phdr[i].p_vaddr > image_region.end ||
+            phdr[i].p_memsz > image_region.end - phdr[i].p_vaddr) {
+            printf("ELF load failed: segment %d is malformed\n", i);
+            return false;
+        }
```
Indanz: Please move the `p_filesz > p_memsz` check to `elf_checkFile()`.

`p_memsz > image_region.end` should be impossible.

I don't understand why you compare virtual addresses with the physical addresses of the image region. `p_vaddr` is the virtual address the memory should be mapped at for the program, it has nothing to do with the address space of the kernel and could be anything. You've already checked that the segment content fits within the ELF file.

Indanz (commented):
Please add defines for all the fields values you check to the arch header files and just use the defines here instead of checking word_t size everywhere.

No need for all these empty new lines, and it would make sense to combine more if-checks together.

Brine28:
@Indanz Not intentional thats just what the projects Style check
(uncrustify) produces for this block I didnt hand align it. and yea youre right thanks for catching this. image_region was derived
from the same p_vaddr/p_memsz values I was then comparing it
against so the check was circular and added no real safety 
and comparing a virtual address (p_vaddr) against whats really
a physical memory sizing question was confused on my part

Removed the image_region comparison from elf_load() entirely
elf_load() is back to its original void signature and does no
validation of its own now  elf_checkFile() already guarantees
p_filesz <= p_memsz and that every PT_LOAD segments file range
fits in the boot module so theres nothing left to check before
the copy.

include/arch/x86/arch/32/mode/kernel/elf.h
```diff
@@ -39,3 +39,8 @@ typedef struct Elf32_Phdr {
     uint32_t      p_flags;     /* Flags: logical "or" of PF_ constants below */
     uint32_t      p_align;     /* Reqd alignment of segment in memory */
 } Elf32_Phdr_t, Elf_Phdr_t;
+
+/* the ELF class value a boot module must have to be loadable on this
+ * build */
```
Indanz: This fits on one line. But frankly the comment doesn't add anything, so I would remove them.

include/arch/x86/arch/kernel/boot_sys.h
```diff
@@ -22,6 +22,7 @@ typedef struct boot_state {
     acpi_rsdp_t  acpi_rsdp; /* copy of the rsdp */
     paddr_t      mods_end_paddr; /* physical address where boot modules end */
     paddr_t      boot_module_start; /* physical address of first boot module */
+    paddr_t      boot_module_end; /* physical address of end of first boot module */
```
Indanz:
```suggestion
    paddr_t      boot_module_end; /* Exclusive physical end address */
```

include/arch/x86/arch/kernel/elf.h
```diff
@@ -7,10 +7,35 @@
...
+ * word size. */
+#define ELFCLASS32 1
+#define ELFCLASS64 2
+#define PT_LOAD    1 /* loadable program segment */
+#define ET_EXEC    2 /* executable file */
+#define ET_DYN     3 /* shared object / position-independent executable */
+#define EM_386     3  /* Intel 80386 (32-bit x86) */
+#define EM_X86_64  62 /* AMD x86-64 */
+
 #include <mode/kernel/elf.h>
```
Indanz, line 19: Macros are expanded on use, not when they are defined, so there is no need for this, you can include elf.h first I think. Yes, that's a bit weird, so you don't have to change this order.

I would remove the comment, because it doesn't add any new information that can't be inferred from reading the code.

include/arch/x86/arch/kernel/elf.h
```diff
@@ -7,10 +7,35 @@
...
+ * every PT_LOAD segment's file range inside the module and
+ * p_filesz <= p_memsz. This only validates the ELF file itself; it
+ * does not check whether seL4 can actually load it (see
+ * elf_getMemoryBounds()/elf_load() and their callers for that). */
+bool_t elf_checkFile(Elf_Header_t *elfFile, word_t max_length);
+
 v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile);
-void elf_load(Elf_Header_t *elfFile, seL4_Word offset);
 
+/* Loads the PT_LOAD segments of the ELF image at the given offset.
```
Indanz:
```suggestion
/* Loads the ELF image to the given offset.
```
Offset is a very confusing parameter name...

src/arch/x86/kernel/boot_sys.c
```diff
@@ -75,13 +75,14 @@ BOOT_CODE static paddr_t find_load_paddr(paddr_t min_paddr, word_t image_size)
     return 0;
 }
 
-BOOT_CODE static paddr_t load_boot_module(word_t boot_module_start, paddr_t load_paddr)
+BOOT_CODE static paddr_t load_boot_module(word_t boot_module_start, word_t boot_module_end,
+                                          paddr_t load_paddr)
```
Indanz: I'd try putting this on one line and see if the style checker complains... Probably does.

src/arch/x86/kernel/elf.c
```diff
@@ -0,0 +1,111 @@
...
+     * elf_load()'s BSS-clearing step computes p_memsz - p_filesz,
+     * which would underflow (unsigned) into a huge value and
+     * memset() far past the segment otherwise. */
+    phdr = (Elf_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
+    for (i = 0; i < elfFile->e_phnum; i++) {
+        if (phdr[i].p_type != PT_LOAD) {
+            continue;
+        }
+        if (phdr[i].p_offset > max_length || phdr[i].p_filesz > max_length - phdr[i].p_offset ||
+            phdr[i].p_filesz > phdr[i].p_memsz) {
```
Indanz: For readability, swap the checks around:
```suggestion
        if (phdr[i].p_offset > max_length || phdr[i].p_filesz > phdr[i].p_memsz ||
            phdr[i].p_filesz > max_length - phdr[i].p_offset) {
```

Indanz (commented):
Please don't add pointless empty new lines. E.g. it rarely makes sense to add a new line next to a line with a bracket or before return.

src/arch/x86/kernel/boot_sys.c
Brine28: If the tests pass I suppose I can mark it as resolved

include/arch/x86/arch/kernel/elf.h
```diff
@@ -10,10 +10,27 @@
 
 #define ELFCLASS32 1
 #define ELFCLASS64 2
+#define PT_LOAD    1 /* loadable program segment */
+#define ET_EXEC    2 /* executable file */
+#define ET_DYN     3 /* shared object / position-independent executable */
+#define EM_386     3  /* Intel 80386 (32-bit x86) */
```
Indanz: Extra unnecessary space between the 3 and /*.

include/arch/x86/arch/kernel/elf.h
```diff
@@ -10,10 +10,27 @@
...
 /* minimal ELF functionality for loading GRUB boot module */
-bool_t elf_checkFile(Elf_Header_t *elfFile);
+
+/* Checks that 'elfFile' points to a well-formed ELF file for this
+ * architecture, entirely contained within the first 'max_length'
+ * bytes starting at 'elfFile' (i.e. within the boot module), with
+ * every PT_LOAD segment's file range inside the module and
+ * p_filesz <= p_memsz. This only validates the ELF file itself; it
+ * does not check whether seL4 can actually load it (see
+ * elf_getMemoryBounds()/elf_load() and their callers for that). */
```
Indanz: This is overly verbose and doesn't add enough useful information.

include/arch/x86/arch/kernel/elf.h
```diff
@@ -10,10 +10,27 @@
...
+bool_t elf_checkFile(Elf_Header_t *elfFile, word_t max_length);
+
 v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile);
+
+/* Loads the ELF image to the given offset.
+ * elf_checkFile() must have already validated this ELF file. */
```
Indanz: Again, not a very useful comment.

src/arch/x86/kernel/elf.c
```diff
@@ -9,15 +9,50 @@
...
-               elfFile->e_ident[2] == 'L'    &&
-               elfFile->e_ident[3] == 'F'    &&
-               elfFile->e_ident[4] == ELF_EXPECTED_CLASS
-           );
+    Elf_Phdr_t *phdr;
+    uint32_t    i;
+    word_t      phdr_table_size;
+
+    /* We can't safely read any field until we know the header itself
+     * fits inside the boot module. */
```
Indanz: Idem.

src/arch/x86/kernel/elf.c
```diff
@@ -9,15 +9,50 @@
...
+    if (max_length < sizeof(Elf_Header_t) ||
+        elfFile->e_ident[0] != '\177' || elfFile->e_ident[1] != 'E' ||
+        elfFile->e_ident[2] != 'L'    || elfFile->e_ident[3] != 'F' ||
+        elfFile->e_ident[4] != ELF_EXPECTED_CLASS ||
+        elfFile->e_machine != ELF_EXPECTED_MACHINE ||
+        (elfFile->e_type != ET_EXEC && elfFile->e_type != ET_DYN) ||
+        elfFile->e_phentsize != sizeof(Elf_Phdr_t) || elfFile->e_phnum == 0) {
+        return false;
+    }
+
```
Indanz: Please remove these kind of empty new lines. For bigger comments, use the following style:
```
/*
 * text...
 */
```

src/arch/x86/kernel/elf.c
```diff
@@ -9,15 +9,50 @@
...
+    /* The program header table must be entirely within the boot
+     * module. e_phnum is 16 bits and sizeof(Elf_Phdr_t) is small, so
+     * this multiplication cannot overflow word_t; the bounds check
+     * itself is written with subtraction so that it cannot overflow
+     * either. */
```
Indanz: Suggestion: "Overflow-safe check that the program table fits within the file."

src/arch/x86/kernel/elf.c
```diff
@@ -9,15 +9,50 @@
...
+    phdr_table_size = elfFile->e_phnum * sizeof(Elf_Phdr_t);
+    if (elfFile->e_phoff > max_length || phdr_table_size > max_length - elfFile->e_phoff) {
+        return false;
+    }
+
+    /* Each PT_LOAD segment's file contents must also be entirely
+     * within the boot module, and p_filesz must not exceed p_memsz:
+     * elf_load()'s BSS-clearing step computes p_memsz - p_filesz,
+     * which would underflow (unsigned) into a huge value and
+     * memset() far past the segment otherwise. */
```
Indanz: Suggestion: "Overflow-safe checks that the segments are contained within the file."

src/arch/x86/kernel/elf.c
```diff
@@ -31,9 +66,9 @@ BOOT_CODE v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile)
     elf_reg.start = ~0UL;
     elf_reg.end = 0;
 
-    /* loop through all program headers (segments) and record start/end address */
+    /* loop through all loadable program headers and record start/end address */
```
Indanz: "Loop through all loadable segments and record start/end address"

Or leave it as it was.

src/arch/x86/kernel/elf.c
```diff
@@ -56,8 +90,14 @@ BOOT_CODE void elf_load(Elf_Header_t *elfFile, seL4_Word offset)
     word_t     len;
     uint32_t   i;
 
-    /* loop through all program headers (segments) and load them */
+    /* loop through all program headers and load the loadable segments.
+     * elf_checkFile() already validated that p_offset/p_filesz stay
+     * within the boot module and that p_filesz <= p_memsz for every
+     * PT_LOAD segment, so no further validation is needed here. */
```
Indanz: "Loop through all segments and load them"

Or leave it as it was. Sometimes simpler is better. Over the top verbosity only lowers entropy.

Indanz (approved):
Almost there, last round of nitpicks.

For your commit "x86: elf: validate boot module before loading ", the first and last paragraph would be sufficient. The rest just describes the code you added: We can see that already, no need for such verbosity.

For your commit "x86: elf.c: merge 32-bit and 64-bit into one file", again, the first and last paragraphs would suffice, the inner two are unnecessary.

Otherwise everything looks fine now.

Brine28:
Ive sorted it out. You can take a look. Its just pending approval for the tests to start @Indanz

Indanz (approved):

Indanz:
This will require a proof-sync (even though boot code isn't verified, but to stop the CI preprocess error), which will take some time considering that Gerwin is away at the moment. Feel free to remind us in a few weeks if it's still not merged.

src/arch/x86/kernel/elf.c
```diff
@@ -0,0 +1,93 @@
...
+
+BOOT_CODE v_region_t elf_getMemoryBounds(Elf_Header_t *elfFile)
+{
+    Elf_Phdr_t *phdr = (Elf_Phdr_t *)((paddr_t)elfFile + elfFile->e_phoff);
+    v_region_t elf_reg;
+    vptr_t     sect_start;
+    vptr_t     sect_end;
+    uint32_t   i;
+
+    elf_reg.start = ~0UL;
```
lsf37, line 56: Should this be `~(word_t)0` instead since this is now shared between 32 and 64 bit? I guess it works out, but it feels strange to just use a fixed `UL`.

src/arch/x86/kernel/elf.c
```diff
@@ -0,0 +1,93 @@
...
+    for (i = 0; i < elfFile->e_phnum; i++) {
+        if (phdr[i].p_type != PT_LOAD || phdr[i].p_memsz == 0) {
+            continue;
+        }
+        src = (paddr_t)elfFile + phdr[i].p_offset;
+        dst = phdr[i].p_vaddr + offset;
+        len = phdr[i].p_filesz;
+        memcpy((void *)dst, (char *)src, len);
+        dst += len;
+        memset((void *)dst, 0, phdr[i].p_memsz - len);
```
lsf37: This is just moved, but since I'm staring at it anyway -- do we really need this memset? 

https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/src/arch/x86/kernel/boot_sys.c#L135-L138

seems to already clear that memory just before `elf_load()` is called. Or am I missing something?

lsf37:
From the PR description:

> - Segments were copied to p_vaddr + offset without checking that
> they actually fall within the image region computed by
> elf_getMemoryBounds().

It looks to me like this is still the case. Should we add a check for this? I.e. add a check for `p_memsz > max_vaddr - p_vaddr` in `elf_checkFile`?

I'm not exactly an expert in ELF loading, so I can only offer a fairly shallow review (and agreement that more checking is good, as is the merge of 32/64 for this one). If @Indanz and @midnightveil are fine with the rest, I'm happy from the verification side.

Brine28:
@lsf37  Wow I dont know how I missed that... youre absolutely right. Would you like me to apply the code changes as you suggested and force push again or should I wait?

src/arch/x86/kernel/elf.c
Indanz: `UL` is the right size for both, so I don't see why it would be wrong.

Kernel is full of unnecessary `word_t`'s, that's why a normal `UL` looks strange to you. :-)

Brine28:
Pushed the overflow check as suggested @lsf37

src/arch/x86/kernel/elf.c
Indanz: Yes, it does. It's because of gaps between loaded segments: The kernel will hand all memory to user space as caps, so it wants to zero those and can't count on `elf_load()` to do that.

But should `elf_load()` count on the caller zeroing the memory? I suppose we could, would simplify the code.

src/arch/x86/kernel/elf.c
Brine28: Youre right reverted to ~0UL lsf37 raised the same question
independently but since UL is exactly word_t's width here
anyway Ive kept it consistent (including in the new overflow
check I just added per lsf37's other comment). Pushed.

Indanz: Also, `word_t` is equally wrong, it's either an `uint32_t` or `uint64_t`. But `UL` is shorter and more readable and avoids a cast altogether.

src/arch/x86/kernel/elf.c
Brine28: Agreed since load_boot_module() already unconditionally zeroes
the whole region right before calling elf_load(), I'll drop the
memset() and document that elf_load() now relies on the caller
having pre zeroed the destination.

Indanz (approved):

Brine28:
@Indanz  New push.  I also implemented the new push notification settings you mentioned you can check if there are any mistake

Indanz:
> @Indanz New push. I also implemented the new push notification settings you mentioned you can check if there are any mistake

I don't see a difference and I don't understand what you mean.

Brine28:
@Indanz  I actually wrote that after you had already given your approval my apologies. I simply hadnt noticed at the time that you had approved it.

Brine28:
And... @Indanz  Sorry, that was a translation error on my end  ignore the
"push notification settings" part that doesn't make sense.

What I meant: I pushed a new version that addresses lsf37's
overflow check suggestion (added p_memsz > ~0UL - p_vaddr to
elf_checkFile), and kept ~0UL as you suggested rather than
~(word_t)0.

src/arch/x86/kernel/elf.c
lsf37: Ok, happy to bow to the majority on this one. It does work out.

src/arch/x86/kernel/elf.c
```diff
@@ -0,0 +1,95 @@
...
+     * does, via memzero()). */
+    for (i = 0; i < elfFile->e_phnum; i++) {
+        if (phdr[i].p_type != PT_LOAD || phdr[i].p_memsz == 0) {
+            continue;
+        }
+        src = (paddr_t)elfFile + phdr[i].p_offset;
+        dst = phdr[i].p_vaddr + offset;
+        len = phdr[i].p_filesz;
+        memcpy((void *)dst, (char *)src, len);
+    }
```
lsf37, line 94: There was a `dst += len` here after the `memcpy` (before the zeroing `memcpy`). I think that is still needed, no? Otherwise `dst` doesn't progress.

Indanz: `dst` is reset at every loop, it could have been `memset((void *)dst + len, 0, phdr[i].p_memsz - len);`

Indanz: But the removal of memset change should be in a separate commit.

Brine28: No bug there  dst is recomputed fresh from
phdr[i].p_vaddr + offset at the top of every loop iteration so
it never needed to carry over between segments Indanz's
explanation above is correct.

Brine28:
@Indanz  split into a separate commit
("x86: elf_load: drop redundant BSS-clearing memset"). pushed.

src/arch/x86/kernel/elf.c
lsf37: Indeed, now I see it. Apologies for the noise.

PR #1721, opened 2026-07-28
python-deps: add typing_extensions and protobuf

lsf37:
- Add typing_extensions to enable type annotations in older python 3.x versions.

- Add protobuf, because some Linux distros come with a too old protobuf and it is easier/nicer to reduce the distro package install list.

Needed for #1637 and related to https://github.com/seL4/docs/issues/317

After this is merged, we should

- [x] publish the pypi package
- [x] remove protobuf from the apt package install list
- [x] remove the protobuf apt install in docker

### DIFF
--- a/tools/python-deps/setup.py
+++ b/tools/python-deps/setup.py
@@ -24,6 +24,8 @@ twine upload -r pypi dist/*
 
 from setuptools import setup
 
+# six and future are for python 2 compatibility and will be removed in the future.
+# They are currently kept so current sel4-deps still works with older code bases.
 DEPS = [
     'six',
     'future',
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

midnightveil:
At some point I think we should aplit out the deps needed by just seL4 and by seL4-adjacent projects. Currently Microkit just suggests sel4-deps but doesn't care about most.

lsf37:
> At some point I think we should aplit out the deps needed by just seL4 and by seL4-adjacent projects. Currently Microkit just suggests sel4-deps but doesn't care about most.

I haven't done an actual audit of this, but there is already a separate camkes-deps. What sel4-deps contains should be what is needed for sel4test and sel4bench only, i.e. what you need to do a platform port. That is a bit more than what you need to do build the kernel, so maybe there should be a different package for that. Or just a requirements.txt in the seL4 repo maybe.

midnightveil:
It might be possible to do like `sel4-deps[kernel-only]` but it also doesn't matter too much.

Indanz (approved):

PR #1724, opened 2026-07-31
bitfield_gen: fix illegal call of umm.base_name()

lsf37:
A previous commit introduced asserts and type checks, which reveal that `umm.base_name()` was called on non-base_name fields.

Use shortcut `and` as intended to guard by `umm.is_base()` instead of bitwise `&`.

This previously worked fine, because the non-sensical return values from `base_name()` were then ignored.

### DIFF
--- a/tools/bitfield_gen.py
+++ b/tools/bitfield_gen.py
@@ -3101,8 +3101,8 @@ if __name__ == '__main__':
         print("end", file=out_file)
     elif options.hol_proofs:
         def is_bit_type(tp):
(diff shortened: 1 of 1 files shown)


midnightveil (approved):
Oops.

Xaphiosis (approved):
Hahaha nice

lsf37:
I guess this counts as success for the new checks, but maybe a reminder that I should run proof tests on things that are used on the proof side.

Issue #1723, opened 2026-07-31
IOMMU builds give 'WARNING: internal book keeping error. Less pagetables allocated than predicted'


midnightveil:
Noticed in this run on Microkit: https://github.com/seL4/microkit/actions/runs/30589966874/job/91035292461. That's on Skylake.

```

  IOMMU 0x0: 8-bit domain IDs supported
  IOMMU 0x1: 8-bit domain IDs supported
  IOMMU: Using 4 page-table levels (max. supported: 4)
  Available phys memory regions: 5
    [100000..d5610000)
    [d565b000..d56b2000)
    [d5eb3000..db4aa000)
    [dc4ff000..dc500000)
    [100000000..21e000000)
  Reserved virt address space regions: 1
    [ffffff8000100000..ffffff8000a7e000)
  IOMMU: Create VTD context table for PCI bus 0x0 (pptr=0xffffff8215c09000)
  IOMMU: Create VTD context table for PCI bus 0x1 (pptr=0xffffff8215c27000)
  IOMMU: Create VTD context table for PCI bus 0x2 (pptr=0xffffff8215c28000)
  IOMMU: Create VTD context table for PCI bus 0x3 (pptr=0xffffff8215c29000)
  IOMMU: Create VTD context table for PCI bus 0x4 (pptr=0xffffff8215c2a000)
  IOMMU: Create VTD context table for PCI bus 0x5 (pptr=0xffffff8215c2b000)
  IOMMU: Create VTD context table for PCI bus 0x6 (pptr=0xffffff8215c2c000)
  IOMMU: Create VTD context table for PCI bus 0x7 (pptr=0xffffff8215c2d000)
  IOMMU: Create VTD context table for PCI bus 0x8 (pptr=0xffffff8215c2e000)
  IOMMU: Create VTD context table for PCI bus 0x9 (pptr=0xffffff8215c2f000)
  IOMMU: Create VTD context table for PCI bus 0xa (pptr=0xffffff8215c30000)
  IOMMU: Create VTD context table for PCI bus 0xb (pptr=0xffffff8215c31000)
  IOMMU: Create VTD context table for PCI bus 0xc (pptr=0xffffff8215c32000)
  IOMMU: Create VTD context table for PCI bus 0xd (pptr=0xffffff8215c33000)
  IOMMU: Create VTD context table for PCI bus 0xe (pptr=0xffffff8215c34000)
[...]

PR #1726, opened 2026-08-04
VT-d: fix oversized RMRR paging reservation

terryzbai:
Before enabling Intel VT-d, the kernel parses RMRR entries in the ACPI DMAR table to maintain 1:1 identity mappings for reserved device regions. The kernel pre-allocates and assigns these paging structures to the root server so it can later manage or extend the mappings.

The previous calculation produced an excessively large paging count, causing the kernel to reserve significantly more memory for the root server than necessary.
```
ERROR: no free memory region is big enough for root server objects, need size/alignment of 2^22
ERROR: free memory management initialization failed
seL4 called fail at seL4/src/arch/x86/kernel/boot_sys.c:723 in function boot_sys, saying "boot_sys failed for some reason :(
"
halting...
Kernel entry via Unknown (0)
```

This corrects the calculation so the reserved memory footprint matches [the paging structures actually allocated for RMRR regions](https://github.com/seL4/seL4/blob/b5050de6daaf13fefbaf791cb28f92109598aa6b/src/plat/pc99/machine/intel-vtd.c#L316-L393).

The number of paging structures needed for RMRR regions includes:
- 1 for root table
- `N_VTD_CONTEXTS`: each for one context (PCIe bus)
- `rmrr_list->num`: each for one PML4 of a device
- `0 or 1` for PDPT of a device (the kernel assumes only the first entry of PDPT is used)
- `varied count` for PD and PT depending on size of RMRR regions

The original implementation filters the identical RMRR regions by bus number id, which does not make sense because every device has its own paging structures. And `size += get_n_paging(region, 32 - (VTD_PT_INDEX_BITS * i + seL4_PageBits));` does not produce the correct number of paging structures at the target level.

In theory, each device is allowed to have multiple RMRR regions and the update calculation adds two more redundant pages (one for PML4 and one for PDPT) in this case, but it should be fine because most of devices declare a single RMRR region and this does not waste much memory as it used to.

### DIFF
--- a/src/plat/pc99/machine/intel-vtd.c
+++ b/src/plat/pc99/machine/intel-vtd.c
@@ -79,6 +79,7 @@
 #define SAGAW_4_LEVEL 0x04
 #define SAGAW_5_LEVEL 0x08
 #define SAGAW_6_LEVEL 0x10
+#define MAX_IOPT_LEVELS 6
 
 #define CONTEXT_GLOBAL_INVALIDATE 0x1
 #define IOTLB_GLOBAL_INVALIDATE   0x1
@@ -87,6 +88,9 @@
 #define DMA_TLB_WRITE_DRAIN BIT(16)
 
 #define N_VTD_CONTEXTS 256
+
+#define ROUND_DOWN_PAGE_IDX(n, b) ((n) >> (b))
+#define ROUND_UP_PAGE_IDX(n, b)  (((n) >> (b)) + (((n) & MASK(b)) != 0))
 
 typedef uint32_t drhu_id_t;
 
@@ -274,39 +278,57 @@ BOOT_CODE word_t vtd_get_n_paging(acpi_rmrr_list_t *rmrr_list)
 
     word_t size = 1; /* one for the root table */
     size += N_VTD_CONTEXTS; /* one for each context */
-    size += rmrr_list->num; /* one for each device */
 
     if (rmrr_list->num == 0) {
         return size;
     }
 
-    /* filter the identical regions by pci bus id */
-    acpi_rmrr_list_t filtered;
-    filtered.entries[0] = rmrr_list->entries[0];
-    filtered.num = 1;
+    for (word_t i = 0; i < rmrr_list->num; i++) {
+        vptr_t paging_start_idx[MAX_IOPT_LEVELS];
+        vptr_t paging_end_idx[MAX_IOPT_LEVELS];
 
-    for (word_t i = 1; i < rmrr_list->num; i++) {
-        if (vtd_get_root_index(rmrr_list->entries[i].device) !=
-            vtd_get_root_index(filtered.entries[filtered.num - 1].device) &&
-            rmrr_list->entries[i].base != filtered.entries[filtered.num - 1].base &&
-            rmrr_list->entries[i].limit != filtered.entries[filtered.num - 1].limit) {
-            filtered.entries[filtered.num] = rmrr_list->entries[i];
-            filtered.num++;
+        for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
+            /* If we are still looking up bits beyond the 32bit of physical
+             * that we support then we select entry 0 in the current PT */
+            if ((VTD_PT_INDEX_BITS * k + seL4_PageBits) >= 32) {
+                paging_start_idx[k] = 0;
+                paging_end_idx[k] = 1;
+            } else {
+                paging_start_idx[k] = ROUND_DOWN_PAGE_IDX(rmrr_list->entries[i].base, VTD_PT_INDEX_BITS * k + seL4_PageBits);
+                paging_end_idx[k] = ROUND_UP_PAGE_IDX(rmrr_list->entries[i].limit, VTD_PT_INDEX_BITS * k + seL4_PageBits);
+            }
         }
-    }
 
-    for (word_t i = x86KSnumIOPTLevels - 1; i > 0; i--) {
-        /* If we are still looking up bits beyond the 32bit of physical
-         * that we support then we select entry 0 in the current PT */
-        if ((VTD_PT_INDEX_BITS * i + seL4_PageBits) >= 32) {
-            size++;
-        } else {
-            for (word_t j = 0; j < filtered.num; j++) {
-                v_region_t region = (v_region_t) {
-                    .start = filtered.entries[j].base,
-                    .end = filtered.entries[j].limit
-                };
-                size += get_n_paging(region, 32 - (VTD_PT_INDEX_BITS * i + seL4_PageBits));
+        for (word_t j = 0; j < i; j++) {
+            if (rmrr_list->entries[j].device == rmrr_list->entries[i].device) {
+                for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
+                    vptr_t start;
+                    vptr_t end;
+                    if ((VTD_PT_INDEX_BITS * k + seL4_PageBits) >= 32) {
+                        start = 0;
+                        end = 1;
+                    } else {
+                        start = ROUND_DOWN_PAGE_IDX(rmrr_list->entries[j].base, VTD_PT_INDEX_BITS * k + seL4_PageBits);
+                        end = ROUND_UP_PAGE_IDX(rmrr_list->entries[j].limit, VTD_PT_INDEX_BITS * k + seL4_PageBits);
+                    }
+
+                    if (paging_start_idx[k] >= start && paging_start_idx[k] < end) {
+                        paging_start_idx[k] = MAX(paging_start_idx[k], end);
+                    }
+                    if (paging_end_idx[k] > start && paging_end_idx[k] <= end) {
+                        paging_end_idx[k] = MIN(paging_end_idx[k], start);
+                    }
+                    if (paging_start_idx[k] >= paging_end_idx[k]) {
+                        paging_start_idx[k] = 0;
+                        paging_end_idx[k] = 0;
+                    }
+                }
+            }
+        }
+
+        for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
+            if (paging_start_idx[k] < paging_end_idx[k]) {
+                size += (paging_end_idx[k] - paging_start_idx[k]);
             }
         }
     }


midnightveil:
Does this happen to fix #1723, btw?

terryzbai:
That is caused by the same issue, but like what I mentioned above that the kernel reserves a few more redundant paging strcutures if there are two or more RMRR regions for same device.

```
ACPI: IOMMU host address width: 39
        ACPI: registering RMRR entry for region for device: bus=0x0 dev=0x14 fun=0x0
        ACPI: registering RMRR entry for region for device: bus=0x0 dev=0x14 fun=0x0
        ACPI: registering RMRR entry for region for device: bus=0x0 dev=0x2 fun=0x0
...
IOMMU 0x0: enabling... enabled
IOMMU 0x1: enabling... enabled
WARNING: internal book keeping error. Less pagetables allocated than predicted: 3 page tables allocated but not used.
```

The strictly correct calculation would be to book keep every paging structures for every device like what it does at the real allocation phase.

midnightveil:
> The strictly correct calculation would be to book keep every paging structures for every device like what it does at the real allocation phase.

It should be made strictly correct...? It's somewhat important that the kernel should boot properly.

Also, you have a typo in the commit and PR title.

terryzbai:
In quick tests, the warning messages on `skylake` and `haswell` have gone with updated paging reservation.

midnightveil:
It looks like the 64-bit skylake platforms currently report this message (not 64-bit haswell).

```
WARNING: internal book keeping error. Less pagetables allocated than predicted: 1 page tables allocated but not used.
```

I think in theory once this is fixed we could upgrade this warning to a boot failure (?).

terryzbai:
> It looks like the 64-bit skylake platforms currently report this message (not 64-bit haswell).

This has been fixed, and tested on skylake and haswell in both 64-bit and 32-bit mode. It would be good to try hw-test again.

> I think in theory once this is fixed we could upgrade this warning to a boot failure (?).

I think so

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -88,6 +89,9 @@
 
 #define N_VTD_CONTEXTS 256
 
+#define ROUND_DOWN_PAGE_IDX(n, b) ((n) >> (b))
+#define ROUND_UP_PAGE_IDX(n, b)  (((n) >> (b)) + (((n) & MASK(b)) != 0))
```
Indanz, line 93:
```suggestion
#define ROUND_UP_PAGE_IDX(n, b)  (ROUND_UP(n, b) >> (b))
```
Also misleading naming, as it doesn't take a page index as input.

But if you make `paging_end_idx` inclusive, you don't need this at all and can only use `ADDR_TO_PAGE_IDX()` or whatever other more sensible name you can think of.

Also, if you do add helper macros, you should be able to use them elsewhere too. Otherwise they really come out of the blue, while this code is supposed to match other code.

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -274,39 +278,57 @@ BOOT_CODE word_t vtd_get_n_paging(acpi_rmrr_list_t *rmrr_list)
...
-            vtd_get_root_index(filtered.entries[filtered.num - 1].device) &&
-            rmrr_list->entries[i].base != filtered.entries[filtered.num - 1].base &&
-            rmrr_list->entries[i].limit != filtered.entries[filtered.num - 1].limit) {
-            filtered.entries[filtered.num] = rmrr_list->entries[i];
-            filtered.num++;
+    for (word_t i = 0; i < rmrr_list->num; i++) {
+        vptr_t paging_start_idx[MAX_IOPT_LEVELS];
+        vptr_t paging_end_idx[MAX_IOPT_LEVELS];
+
+        for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
```
Indanz, line 290: `vtd_map_reserved_page()` does `i = x86KSnumIOPTLevels - 1`, is this deliberately different?

midnightveil:
Are you able to explain why this is actually necessary? This sort of just says what the kernel does, not why it needs to do it.

> Before enabling Intel VT-d, the kernel parses RMRR entries in the ACPI DMAR table to maintain 1:1 identity mappings for reserved device regions. 

Why does kernel need to do it at boot, and it not be a structure managed by user-level (and also the kernel have to reserve a certain number of domain IDs).

cc @cazb2

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -274,39 +278,57 @@ BOOT_CODE word_t vtd_get_n_paging(acpi_rmrr_list_t *rmrr_list)
...
+                    if (paging_start_idx[k] >= start && paging_start_idx[k] < end) {
+                        paging_start_idx[k] = MAX(paging_start_idx[k], end);
+                    }
+                    if (paging_end_idx[k] > start && paging_end_idx[k] <= end) {
+                        paging_end_idx[k] = MIN(paging_end_idx[k], start);
+                    }
+                    if (paging_start_idx[k] >= paging_end_idx[k]) {
+                        paging_start_idx[k] = 0;
+                        paging_end_idx[k] = 0;
+                    }
```
Indanz, line 324: A comment would be nice, as this doesn't seem to make any sense.

Also, please write: `start <= paging_start_idx[k] && paging_start_idx[k] < end` for readability.

src/plat/pc99/machine/intel-vtd.c
```diff
@@ -274,39 +278,57 @@ BOOT_CODE word_t vtd_get_n_paging(acpi_rmrr_list_t *rmrr_list)
...
+                }
+            }
+        }
+
+        for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
+            if (paging_start_idx[k] < paging_end_idx[k]) {
```
Indanz, line 330: `paging_start_idx[k] > paging_end_idx[k]` should be impossible.

Indanz (commented):
You really need to explain the algorithm and why it is correct.

The code does not match the allocation/mapping code, so I can't really verify if it is correct or not.

> Before enabling Intel VT-d, the kernel parses RMRR entries in the ACPI DMAR table to maintain 1:1 identity mappings for reserved device regions.

The main question is why?

I don't think IOMMU is an all or nothing thing, it's enabled per device, so you shouldn't need to create a 1:1 mapping for reserved regions.

Or is it creating all VTD context tables, and your comment is wrong?

Why is the kernel doing this instead of user space?

Why is the kernel doing any of this IOMMU stuff instead of user space? Can't we just export the minimum functionality necessary? (Like GetAddr on PTs and access to some system regs.)

cazb2:
> Are you able to explain why this is actually necessary? This sort of just says what the kernel does, not why it needs to do it.
> 
> > Before enabling Intel VT-d, the kernel parses RMRR entries in the ACPI DMAR table to maintain 1:1 identity mappings for reserved device regions.
> 
> Why does kernel need to do it at boot, and it not be a structure managed by user-level (and also the kernel have to reserve a certain number of domain IDs).
> 
> cc @cazb2
@midnightveil 

My understanding is that these RMRR are for devices such as integrated graphics, audio, storage, maybe integrated nics etc. Without mapping in these regions these devices will fault when the IOMMU is turned on, in theory this may crash the system, not 100% sure if that is always true. This is the intel vt-d specs RMRR section
IMG_7058


If the kernel does map in reserved memory region for these devices it has to reserve the domain IDs that it uses when setting up the VT-D translation tables. The domain IDs are the identifiers to IO address spaces, which can be shared by multiple devices. Hence if user level was able to use a reserved domain ID, then an untrusted device could DMA to any memory mapped into the address space such as the reserved memory regions from above.

The domain IDs themselves are used for caching by the IOMMU, which is how you get problems if you don't reserve them.

terryzbai:
> I don't think IOMMU is an all or nothing thing, it's enabled per device, so you shouldn't need to create a 1:1 mapping for reserved regions.

Take our test machines as an example, the USB and the graphics devices are used by the firmware at boot time, they keep sending DMA requests even if the IOMMU is turned on until they are explicitly disabled. The VT-d faults are triggered if the 1:1 mappings do not exist.

> Or is it creating all VTD context tables, and your comment is wrong?

It is creating all the VTD context tables, as well as 1:1 mapping for all the reserved regions.

src/plat/pc99/machine/intel-vtd.c
terryzbai: This is deliberate, and matches [iopt creation for each device](https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/src/plat/pc99/machine/intel-vtd.c#L327) if it does not exist.

src/plat/pc99/machine/intel-vtd.c
terryzbai: `paging_start_idx[k]` might equal to `paging_end_idx[k]` when no new paging structures need to be created.

src/plat/pc99/machine/intel-vtd.c
Indanz: It does not match https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/src/plat/pc99/machine/intel-vtd.c#L343

src/plat/pc99/machine/intel-vtd.c
Indanz: Yes, and?

That does not explain the guard check, `size += (paging_end_idx[k] - paging_start_idx[k]);` works perfectly fine if it adds zero.

Indanz: Also, again, you are off-by-one when indexing `paging_start_idx` because you go from high to low.

Indanz:
So there are multiple options:

1) Do the 1:1 mapping and enable IOMMU for all devices. (What we do now.)
2) Don't do the 1:1 mapping, enable IOMMU and let the boot devices fault. This should be harmless, other than the devices not working properly and possibly spamming faults.
3) Disable those devices at boot. Might not be easy, as it might be device specific how to do that.
4) Let user space decide and only enable IOMMU after the user configured it. Then the burden of doing all this is on user space instead of the kernel.

It looks like we currently have the most complicated solution (for the kernel).

midnightveil:
This is an interesting read: https://access.redhat.com/sites/default/files/attachments/rmrr-wp1.pdf

Note; deciding how to do this bstter will probably be difficult. This PR just fixed a kernel bookkeeping issue, which is likely worth doing.

src/plat/pc99/machine/intel-vtd.c
terryzbai: When allocating memory, it checks `vtd_cte_ptr_get_present(vtd_context_slot)` for top-level paging structures, or `vtd_pte_ptr_get_write(vtd_pte_slot)` for lower-level paging structures.

When calculating number of paging structures, it just needs to check if there are any previous reserved regions specified for the same device and removes the duplicated part. For simplicity, I just merged the top-level count into the for loop. I will leave more comments for this implementation. Does this implementation raise any other concerns that I did not realise?

src/plat/pc99/machine/intel-vtd.c
terryzbai: > paging_start_idx[k] > paging_end_idx[k] should be impossible.

Now I remember why I had this check. `paging_start_idx[k] > paging_end_idx[k]` is possible when the paging structures for another reserved region include what is needed for the current reserved region. E.g., [0, 1] is counted for a previous region, and [0, 1] is also needed for the current region. Then `paging_start_idx[k]` would be `MAX(paging_start_idx[k], end) = 1`, and `paging_end_idx[k]` would be `MIN(paging_end_idx[k], start) = 0`.

> Also, again, you are off-by-one when indexing paging_start_idx because you go from high to low.

For every level with address bits > 32, we always use only entry 0. This restricts the IOMMU vaddr below 4GB, and makes 32-bit mode works fine without other changes.

For top-level paging structure (i.e. k = x86KSnumIOPTLevels), it also checks if [0, 1) has been counted for previous regions. As mentioned in above example, `paging_start_idx[k] > paging_end_idx[k]` if there are regions for same device, so no duplicated paging structures need to be reserved for top-level.

src/plat/pc99/machine/intel-vtd.c
terryzbai: In 32bit mode, an overflow can occur at level-2. That was why I used page idx rather than `ROUND_UP(addr)` for duplication checks.

> But if you make paging_end_idx inclusive, you don't need this at all and can only use ADDR_TO_PAGE_IDX() or whatever other more sensible name you can think of.

uh, you are right. Then the algorithm would also be easier I think.

terryzbai:
> 3. Disable those devices at boot. Might not be easy, as it might be device specific how to do that.

Clearing `Bus Master Enable` bit for the PCIe device on ECAM disables the further DMA requests, so no spamming faults will be produced. But the kernel needs to parse the ACPI MCFG table for reading ECAM base address and bus range, which is easy.

src/plat/pc99/machine/intel-vtd.c
Indanz: None of what you said makes sense to me.

Is `MAX_IOPT_LEVELS` guaranteed to be at least one greater than `x86KSnumIOPTLevels`?

If not, how is `paging_start_idx[x86KSnumIOPTLevels]` not out of bound? It's declared as `paging_start_idx[MAX_IOPT_LEVELS]`.

src/plat/pc99/machine/intel-vtd.c
Indanz: And maybe have one virtual range first before you convert that to PT levels.

Doesn't look like you support large page mappings anyway (which would make counting this exactly even harder).

src/plat/pc99/machine/intel-vtd.c
Indanz: When you say top level count, do you mean vtd context page? But that one has no virtual range, so merging that with this loop doesn't make sense? I am confused.

src/plat/pc99/machine/intel-vtd.c
Indanz, line 324:
```suggestion
                    /* Adjust region so they don't overlap with previously counted regions: */
                    if (start <= paging_start_idx[k] && paging_start_idx[k] < end) {
                        paging_start_idx[k] = end;
                    }
                    if (start < paging_end_idx[k] && paging_end_idx[k] <= end) {
                        paging_end_idx[k] = start;
                    }
                    /* If region fully overlaps previous regions, ignore it: */
                    if (paging_start_idx[k] >= paging_end_idx[k]) {
                        paging_start_idx[k] = 0;
                        paging_end_idx[k] = 0;
                        /* No need to check other regions too */
                        break;
                    }
```
(The break currently doesn't work, but if you do everything on one virtual range first instead of per PT level, it does work.)

src/plat/pc99/machine/intel-vtd.c
Indanz: > > paging_start_idx[k] > paging_end_idx[k] should be impossible.
> 
> Now I remember why I had this check. `paging_start_idx[k] > paging_end_idx[k]` is possible when the paging structures for another reserved region include what is needed for the current reserved region.

Yes, but this is already handled by the other if-check which sets the region's start and end to zero.

Indanz:
> This is an interesting read: https://access.redhat.com/sites/default/files/attachments/rmrr-wp1.pdf

Interesting indeed. But the main takeaway is that they still do the 1:1 mapping, they just disallow such devices to be passed through to virtual machines because it causes problems and can break containment.

> Note; deciding how to do this bstter will probably be difficult. This PR just fixed a kernel bookkeeping issue, which is likely worth doing.

True, so all that is a bit off-topic, but still important to think about.

Indanz:
> > 3. Disable those devices at boot. Might not be easy, as it might be device specific how to do that.
> 
> Clearing `Bus Master Enable` bit for the PCIe device on ECAM disables the further DMA requests, so no spamming faults will be produced. But the kernel needs to parse the ACPI MCFG table for reading ECAM base address and bus range, which is easy.

Okay great. Only problem is that that only works for well-behaving devices like USB and graphics, and not for stupid system management stuff that could do anything in the background, so just disabling those devices doesn't seem like a safe option, sadly.

dreamliner787-9:
> > 3. Disable those devices at boot. Might not be easy, as it might be device specific how to do that.
> 
> Clearing `Bus Master Enable` bit for the PCIe device on ECAM disables the further DMA requests, so no spamming faults will be produced. But the kernel needs to parse the ACPI MCFG table for reading ECAM base address and bus range, which is easy.

You can clear the Bus Master Enable bit via the legacy 0xCF8 and 0xCFC PCI Access Mechanism type 1 I/O ports to avoid having to parse the MCFG table in-kernel, since that table won't always exist anyways.

midnightveil:
(Notably the kernel already has to turn some VTD-related feature on per-device, I'm not sure if it is that feature)

src/plat/pc99/machine/intel-vtd.c
terryzbai: The top-level table means the PML4 table in case of 4-level paging structures. Each entry in context table points to the PML4 table of corresponding device.
And now the current kernel creates the PML4 table for only the devices having reserved regions.

src/plat/pc99/machine/intel-vtd.c
terryzbai: > Is `MAX_IOPT_LEVELS` guaranteed to be at least one greater than `x86KSnumIOPTLevels`?
> 
> If not, how is `paging_start_idx[x86KSnumIOPTLevels]` not out of bound? It's declared as `paging_start_idx[MAX_IOPT_LEVELS]`.

Seems like that the latest VT-d does not support 2-level and 6-level page table according to Specs version 4.0 and 5.0, but it used to support these two (https://www.intel.la/content/dam/www/public/us/en/documents/datasheets/4th-gen-core-family-mobile-u-y-processor-lines-vol-2-datasheet.pdf#:~:text=0%3A%2030%2Dbit%20AGAW%20(2%2Dlevel%20page%20table).%201%3A,tables%20is%20one%20of%20the%20supported%20guest).

But yes, I should have set `MAX_IOPT_LEVELS` to at least 7 regarding to the existing implementation.

> Yes, but this is already handled by the other if-check which sets the region's start and end to zero.

uh, you are right. That's a redundant check.

Indanz: > But yes, I should have set `MAX_IOPT_LEVELS` to at least 7 regarding to the existing implementation.

No, it only mentions 6-level page table support, not 7-level, so the 6 is correct. What is incorrect is you accessing it out-of-bound because you're off-by-one. If you want to do that for other reasons like simpler loop or code, you need to be explicit about it and explain that in a comment, otherwise it looks like a bug.

(The link doesn't work in the sense that it only opens the document, it doesn't open it at the correct place. Probably an Adobe-only thing.)

src/plat/pc99/machine/intel-vtd.c
Indanz: But you don't want to count the top PT by counting on the 32-bit overflow if-check, do you? Because that's what you seem to be doing now. I'd rather you don't do that and bring back the `size += rmrr_list->num; /* one for each device */`. If we ever remove that artificial 32-bit limit, this will subtly break.

midnightveil: As far as I can tell, Indan is specifically commenting on the distinction between:

```c
for (word_t k = x86KSnumIOPTLevels; k > 0; k--) {
```

and

```c
for (i = x86KSnumIOPTLevels - 1; i >= 0; i--) {
```

This is used, e.g. here:

```c
if ((VTD_PT_INDEX_BITS * k + seL4_PageBits) >= 32) {
```

and

```c
if (VTD_PT_INDEX_BITS * i + seL4_PageBits >= 32) {
```

Note how in the existing allocation code that the value of `i` ranges from `[0, numLevels - 1]` whereas in this new code `k` ranges from `[1, numLevels]`. This would cause that if check to be different, and other code.

src/plat/pc99/machine/intel-vtd.c
midnightveil: > The link doesn't work in the sense that it only opens the document, it doesn't open it at the correct place. Probably an Adobe-only thing

It's a Google-chrome only thing.

PR #1727, opened 2026-08-05
Print TCB schedule information from all cores

midnightveil:
Currently it only prints the local core, which isn't very
helpful when you have multiple cores.

### DIFF
--- a/include/api/debug.h
+++ b/include/api/debug.h
@@ -154,8 +154,10 @@ static inline void debug_dumpScheduler(void)
            config_set(CONFIG_ENABLE_SMP_SUPPORT) ? "-------" : "",
            (CONFIG_NUM_DOMAINS > 1)              ? "-------" : "",
            config_set(CONFIG_KERNEL_MCS)         ? "-----------------" : "");
-    for (tcb_t *curr = NODE_STATE(ksDebugTCBs); curr != NULL; curr = TCB_PTR_DEBUG_PTR(curr)->tcbDebugNext) {
-        debug_printTCB(curr);
+    for (word_t core = 0; core < CONFIG_MAX_NUM_NODES; core++) {
+        for (tcb_t *curr = NODE_STATE_ON_CORE(ksDebugTCBs, core); curr != NULL; curr = TCB_PTR_DEBUG_PTR(curr)->tcbDebugNext) {
+            debug_printTCB(curr);
+        }
     }
 }
 #endif /* CONFIG_PRINTING */


include/api/debug.h
```diff
@@ -135,7 +135,7 @@ static inline void debug_printTCB(tcb_t *tcb)
 
 static inline void debug_dumpScheduler(void)
 {
-    printf("Dumping all tcbs!\n");
+    printf("Dumping all tcbs! (called from core %" SEL4_PRIu_word ")\n", CURRENT_CPU_INDEX());
```
Indanz: The user should know from which core they call it, and also, the task calling it is in the list, so I don't think this is needed.

include/api/debug.h
```diff
@@ -154,9 +154,36 @@ static inline void debug_dumpScheduler(void)
...
+        /* this can be NULL because you could call dumpScheduler (by modifying
+         * kernel code) before the other cores have started. this should never
+         * happen after that point. */
+        tcb_t *current = NODE_STATE_ON_CORE(ksCurThread, core);
+        char *current_name = current ? TCB_PTR_DEBUG_PTR(current)->tcbName : "<NUL unknown>";
+        printf("[CPU %" SEL4_PRIu_word "] current scheduler action: %s (cur: %s)\n", core,
+               action == SchedulerAction_ResumeCurrentThread ? "resume current"
+               : (action == SchedulerAction_ChooseNewThread ? "choose new"
+                  : (TCB_PTR_DEBUG_PTR(action)->tcbName)), current_name);
+    }
```
Indanz: This is pointless: You know what the scheduler action is, because we're doing this debug syscall. So it's always `SchedulerAction_ResumeCurrentThread`. Other cores can't be in the kernel because of the big kernel lock and `ksSchedulerAction` shows the last scheduler action, not the current one (there is none currently).

include/api/debug.h
```diff
@@ -154,9 +154,36 @@ static inline void debug_dumpScheduler(void)
...
+
+#ifdef CONFIG_KERNEL_MCS
+    printf("\nDumping MCS release queues:\n");
+    printf("%4s   %4s   %-40s   %-18s\n", "Core", "Index", "Name", "Ready Time");
+    printf("------------------------------------------------------------------------\n");
+    for (word_t core = 0; core < CONFIG_MAX_NUM_NODES; core++) {
+        word_t i = 0;
+        for (tcb_t *curr = NODE_STATE_ON_CORE(ksReleaseQueue.head, core); curr != NULL; curr = curr->tcbSchedNext) {
+            printf("%4lu   %-40s   %4lu   %-18llu", i, TCB_PTR_DEBUG_PTR(curr)->tcbName, core, refill_head(curr->tcbSchedContext)->rTime);
+        }
```
Indanz: Probably better to add a separate, MCS specific debug syscall that shows the refill state.

We don't print out all the tasks in the scheduling queue either, or all the tasks queued on endpoints, so why print this out? I wouldn't add this until you actually needed this for debugging.

Indanz (commented):
If you limit it to the first commit, it's ready to go. I'm not too fond of the other changes, they seem a bit unnecessary.

include/api/debug.h
midnightveil: So, one thing I was using this was to call it from kernel-space, where in that case you don't actually know that.

include/api/debug.h
midnightveil: Ditto regarding calling from kernel-land.

I can't disagree but there was a reason why I added it for #1662 that I don't know anymore.

Probably fine to remove though, idm.

include/api/debug.h
midnightveil: Fair. I just thought it might be useful.

include/api/debug.h
Indanz: In that case it's trivial to add such print yourself.

include/api/debug.h
Indanz: In that case you can also call another debug function. Splitting this code into its own debug function would make more sense to me than adding it here. I wouldn't do it always because then it prints out stale information. If you call it manually in kernel space, you know the context.

midnightveil:
Two removed patches:

[0002-debug-print-current-scheduler-action-for-all-cores.patch](https://github.com/user-attachments/files/30767482/0002-debug-print-current-scheduler-action-for-all-cores.patch)
[0001-debug-UNTESTED-print-release-queues.patch](https://github.com/user-attachments/files/30767483/0001-debug-UNTESTED-print-release-queues.patch)


This just prints schedule information of all cores now.

Indanz (approved):

PR #1731, opened 2026-08-14
gh-1730: arm,asid: Add new asid invocations

### LINKED ISSUE #1730: Pre-RFC: Feature request: Arm ASID invocations that take index parameters
Context: Across all architectures, the kernel assigns an address space ID (ASID) to address space objects. These IDs are then stored inside all caps associated with the address space. On some platforms, the assigned ASID is also used as a tag for the hardware TLB. Currently ASIDs are managed via the `ARMASIDControlMakePool` and `ARMASIDPoolAssign` invocations.

The Arm architecture has TLB maintenance instructions that can be broadcast to multiple Processing elements in the same inner shareable share-ability domain. When supporting SMP guests on multikernel seL4 kernel's, the VMID TLB tag needs to match across all PEs that are participating in virtualizing an SMP guest so that the guest's TLB maintenance instructions that are broadcast correctly lead to the correct entries being invalidated.

Problem: Because the kernel assigns ASIDs using internal policy, and then hardware VMIDs are derived from them using another internal policy, the only way a user space can ensure that the ASIDs of guest virtual machines are aligned is to allocate them in an exact ordering on all PEs. This doesn't scale when there are system architectures with a different number of address spaces on each PE.

Solution: Adding an additional invocation to the existing ASID management invocations that take an index parameter, can allow a system resource manager to more easily apply a different allocation policy that supports assigning ASIDs in an appropriate order.

Discussion: Looking at an example implementation that I will link here, the invocations being added aren't expected to cause any interesting changes to the protection model of the system.

Making the actual change will require an RFC.

kent-mcleod:
Add two new asid invocations that take an index parameter.

### DIFF
--- a/libsel4/arch_include/arm/interfaces/object-api-arch.xml
+++ b/libsel4/arch_include/arm/interfaces/object-api-arch.xml
@@ -489,6 +489,61 @@
                 </description>
             </error>
         </method>
+        <method id="ARMASIDControlMakePoolByIndex" name="MakePoolByIndex" manual_name="Make Pool By Index">
+            <brief>
+                Create an ASID Pool at the requested index.
+            </brief>
+            <description>
+                Together with a capability to <texttt text="Untyped Memory"/>, which is passed as an argument,
+                create an <texttt text="ASID Pool"/>. The untyped capability must represent a
+                4K memory object. This will create an ASID pool with enough space for 1024 VSpaces.
+            </description>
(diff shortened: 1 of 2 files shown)


Indanz:
Does this approach assume a 1:1 mapping between sw ASIDs and hw ASIDs? That part seems to be missing in this PR.

PR #1732, opened 2026-08-14
x86/vcpu: fix multiple VMs not working on single and multi cores

dreamliner787-9:
Please see the commit messages for problems and solutions. Fixes VMs not working or exhibiting erratic behaviours when sharing one physical core or placed on multiple cores.

### DIFF
--- a/include/arch/x86/arch/model/statedata.h
+++ b/include/arch/x86/arch/model/statedata.h
@@ -36,10 +36,6 @@ NODE_STATE_DECLARE(interrupt_t, x86KSPendingInterrupt);
 /* Bitmask of all cores should receive the reschedule IPI */
 NODE_STATE_DECLARE(word_t, ipiReschedulePending);
 
-#ifdef CONFIG_VTX
-NODE_STATE_DECLARE(vcpu_t *, x86KSCurrentVCPU);
-#endif
-
 NODE_STATE_DECLARE(word_t, x86KSCurrentFSBase);
 NODE_STATE_DECLARE(word_t, x86KSCurrentGSBase);
 
@@ -49,6 +45,12 @@ NODE_STATE_DECLARE(word_t, x86KSCurrentGSBase);
 NODE_STATE_DECLARE(word_t, x86KSGPExceptReturnTo);
 
 NODE_STATE_TYPE_DECLARE(modeNodeState, mode);
+
+#ifdef CONFIG_VTX
+NODE_STATE_DECLARE(vcpu_t *, x86KSCurrentVCPU);
+NODE_STATE_DECLARE(struct vmxon_region, x86KSVMXOnRegion);
+#endif
+
 NODE_STATE_END(archNodeState);
 
 /* this is per core state grouped into a separate struct as it needs to be available
--- a/include/arch/x86/arch/object/vcpu.h
+++ b/include/arch/x86/arch/object/vcpu.h
@@ -232,6 +232,13 @@ typedef uint16_t vpid_t;
 #ifdef CONFIG_VTX
 #define VTX_TERNARY(vtx, nonvtx) vtx
 
+#define VMXON_REGION_SIZE 4096
+
+struct vmxon_region {
+    uint32_t revision;
+    char data[VMXON_REGION_SIZE - sizeof(uint32_t)];
+} ALIGN(VMXON_REGION_SIZE);
+
 enum vcpu_gp_register {
     VCPU_EAX = 0,
     VCPU_EBX,
--- a/src/arch/x86/64/c_traps.c
+++ b/src/arch/x86/64/c_traps.c
@@ -182,7 +182,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
         [stack_size]"i"(BIT(CONFIG_KERNEL_STACK_BITS)),
         [guest_msr]"r"(vcpu->guest_msr_registers),
-        [host_msr]"r"(vcpu->host_msr_registers)
+        [host_msr]"r"(&vcpu->host_msr_registers[n_vcpu_msr_register])
 #else /* not CONFIG_X86_64_VTX_64BIT_GUESTS */
         [stack_size]"i"(BIT(CONFIG_KERNEL_STACK_BITS))
 #ifdef ENABLE_SMP_SUPPORT
--- a/src/arch/x86/object/vcpu.c
+++ b/src/arch/x86/object/vcpu.c
@@ -54,7 +54,7 @@ static struct PACKED {
 
 static msr_bitmaps_t msr_bitmap_region ALIGN(BIT(seL4_PageBits));
 
-static char null_ept_space[seL4_PageBits] ALIGN(BIT(seL4_PageBits));
+static char null_ept_space[BIT(seL4_PageBits)] ALIGN(BIT(seL4_PageBits));
 
 /* Cached value of the hardware defined vmcs revision */
 static uint32_t vmcs_revision;
--- a/src/arch/x86/object/vcpu.c
+++ b/src/arch/x86/object/vcpu.c
@@ -25,8 +25,6 @@
 #define VMX_EXIT_QUAL_TYPE_CLTS 2
 #define VMX_EXIT_QUAL_TYPE_LMSW 3
 
-#define VMXON_REGION_SIZE 4096
-
 const vcpu_gp_register_t crExitRegs[] = {
     VCPU_EAX, VCPU_ECX, VCPU_EDX, VCPU_EBX, VCPU_ESP, VCPU_EBP, VCPU_ESI, VCPU_EDI,
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
@@ -46,11 +44,6 @@ typedef struct msr_bitmaps {
     msr_bitmap_t low_msr_write;
     msr_bitmap_t high_msr_write;
 } msr_bitmaps_t;
-
-static struct PACKED {
-    uint32_t revision;
-    char data[VMXON_REGION_SIZE - sizeof(uint32_t)];
-} vmxon_region ALIGN(VMXON_REGION_SIZE);
 
 static msr_bitmaps_t msr_bitmap_region ALIGN(BIT(seL4_PageBits));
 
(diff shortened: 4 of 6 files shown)


src/arch/x86/object/vcpu.c
```diff
@@ -1217,13 +1214,18 @@ BOOT_CODE bool_t vtx_init(void)
     }
     write_cr4(read_cr4() | CR4_VMXE);
     /* we are required to set the VMCS region in the VMXON region */
-    vmxon_region.revision = vmcs_revision;
+#ifdef ENABLE_SMP_SUPPORT
+    struct vmxon_region *vmxon_kptr = &ARCH_NODE_STATE(x86KSVMXOnRegion);
+#else
+    struct vmxon_region *vmxon_kptr = &vmxon_region;
+#endif
```
midnightveil: This shouldn't be necessary, `ARCH_NODE_STATE` should work on both SMP and not-SMP. That's why `NODE_STATE_DECLARE` exists.

dreamliner787-9: Oh I see, will fix

include/arch/x86/arch/model/statedata.h
```diff
@@ -37,6 +37,7 @@ NODE_STATE_DECLARE(interrupt_t, x86KSPendingInterrupt);
 NODE_STATE_DECLARE(word_t, ipiReschedulePending);
 
 #ifdef CONFIG_VTX
+NODE_STATE_DECLARE(struct vmxon_region, x86KSVMXOnRegion ALIGN(VMXON_REGION_SIZE));
```
Indanz: For effective use of caching, please put this at the end, so it's away from the other arch node states.

midnightveil (approved):

midnightveil:
We really need some basic VM tests in sel4test tbh...

Indanz:
> We really need some basic VM tests in sel4test tbh...

It's very high on my todo list, but I keep getting distracted by other things. :-/

src/arch/x86/64/c_traps.c
```diff
@@ -182,7 +182,7 @@ static void NORETURN restore_vmx(tcb_t *cur_thread, vcpu_t *vcpu)
 #ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
         [stack_size]"i"(BIT(CONFIG_KERNEL_STACK_BITS)),
         [guest_msr]"r"(vcpu->guest_msr_registers),
-        [host_msr]"r"(vcpu->host_msr_registers)
+        [host_msr]"r"(&vcpu->host_msr_registers[n_vcpu_msr_register])
```
Indanz, line 185: The old code had `&vcpu->guest_msr_registers[VCPU_GS]`, but that one was actually wrong, because we use `guest_msr` to pop from the stack. So basically you moved the bug from guest to host and I missed it when reviewing.

include/arch/x86/arch/object/vcpu.h
```diff
@@ -232,6 +232,13 @@ typedef uint16_t vpid_t;
 #ifdef CONFIG_VTX
 #define VTX_TERNARY(vtx, nonvtx) vtx
 
+#define VMXON_REGION_SIZE 4096
+
+struct vmxon_region {
+    uint32_t revision;
+    char data[VMXON_REGION_SIZE - sizeof(uint32_t)];
+} PACKED;
```
Indanz: There is no reason for the `PACKED`, if you're changing things anyway, you can as well remove it.

And I think you actually want to add `ALIGN(VMXON_REGION_SIZE)` here.

src/model/statedata.c
```diff
@@ -14,8 +14,22 @@
...
+ * So let's account for this manually. */
+#if defined(__clang__) && defined(CONFIG_VTX)
+#define KSSMP_TABLE_ALIGN MAX(L1_CACHE_LINE_SIZE, _Alignof(smpStatedata_t))
+#else
+#define KSSMP_TABLE_ALIGN L1_CACHE_LINE_SIZE
+#endif
+
 /* Collective cpu states, including both pre core architecture dependant and independent data */
-SMP_STATE_DEFINE(smpStatedata_t, ksSMP[CONFIG_MAX_NUM_NODES] ALIGN(L1_CACHE_LINE_SIZE));
+SMP_STATE_DEFINE(smpStatedata_t, ksSMP[CONFIG_MAX_NUM_NODES] ALIGN(KSSMP_TABLE_ALIGN));
```
Indanz: Once you add the proper alignment requirement to `struct vmxon_region`, there should be no need for this.

Indanz (commented):
Other than the alignment fixes not fixing the source of the problem, all looks good.

src/model/statedata.c
midnightveil: Not true. The alignment of the smpStateData_t is already correct from the alignment specified in x86KSVMXOnRegion; we tested with _Alignof and it was fine (because in NODE_STATE that alignment ends up as part of a struct anyway)

This does seem to be an issue where clang treats align on the variable overriding the alignment of the type of the variable, contrary to what gcc documents it.

Indanz: Ah, yes, there is alignment specified with `NODE_STATE_DECLARE` already. Behaviour might be different if it's on the type though.

This seems like a serious Clang bug. Does the same happen if you change the `#define ALIGN(n)     __attribute__((__aligned__(n)))` in include/util.h with `#define ALIGN(n)     _Alignas(n)`?

In that case, I'd much rather remove the `ALIGN` from `ksSMP` if possible But how are we supposed to find all other places where alignment is silently wrong?

Indanz: It's also possible that it all works as-is once you remove the PACKED, looks like Clang had issues with that in the past.

midnightveil: Hrm. https://godbolt.org/z/4943T6c9r

The assembly output looks about the same for both. This could be a linker difference, or an assembler difference. Interestingly the alignment seems to match the Clang case.

Indanz: So we just got lucky with gcc then. At least `_Alignas(64)` on `global_test` gives a compile error.

Indanz:
Maybe a better approach is to make `vmxon_region` part of the VCPU itself, then any state that leaks into it is contained to one VM too.

dreamliner787-9:
> Maybe a better approach is to make vmxon_region part of the VCPU itself, then any state that leaks into it is contained to one VM too.

No that won't work, the vmxon region is per logical processor not per VCPU, that is what the VMCS region is for

include/arch/x86/arch/object/vcpu.h
dreamliner787-9: > There is no reason for the PACKED, if you're changing things anyway, you can as well remove it.

The previous code had `PACKED` so I left it there, and it is necessary though because we want the vmxon_region to always be 4K sized, rather than leaving it to the compiler and having to check it with a static assert.

src/model/statedata.c
dreamliner787-9: So I assume no changes are needed @Indanz ?

Indanz:
> > Maybe a better approach is to make vmxon_region part of the VCPU itself, then any state that leaks into it is contained to one VM too.
> 
> No that won't work, the vmxon region is per logical processor not per VCPU, that is what the VMCS region is for

Sure, it would "waste" 4k per VCPU, but other than that, why would it not work?

vmxon is only used while a VM is active, so it can be anywhere. There is no requirement that the same memory is used for different vmx sessions.

include/arch/x86/arch/object/vcpu.h
Indanz: Compiler isn't allowed to just add random padding.

src/model/statedata.c
Indanz: You just proved that having alignment specified on `ksSMP` silently overrules alignment requirements on inner elements, so of course some change is needed.

I think we should remove manual alignment on `ksSMP` altogether, to prevent this from happening again in the future.

And check for nested conflicting ALIGN attributes elsewhere somehow...

Indanz: And I would still prefer vmxon region to be inside the VCPU object itself. Sure, that wastes a little bit of memory per VCPU if you run more than one per CPU, but considering there are megabytes of memory wasted on running a VM, those few kilobytes don't make the difference. It also saves 4k per cpu for use cases that don't run VMs.

(If you do this, then no need to change `ksSMP` alignment yet, that should be done in a separate PR anyway. I'll have a look today.)

midnightveil:
> Sure, it would "waste" 4k per VCPU, but other than that, why would it not work?
> vmxon is only used while a VM is active, so it can be anywhere. There is no requirement that the same memory is used for different vmx sessions.

Not true. Per Intel SDM:

> Before executing VMXON, software allocates a region of memory (called the VMXON region) that the logical processor uses to support VMX operation..... Software should use a separate region for each logical processor and should not access or modify the VMXON region of a logical processor between execution of VMXON and VMXOFF on that logical processor. Doing otherwise may lead to unpredictable behaviors
>
> Before entering VMX operation, the host VMM allocates a VMXON region. A VMM can host several virtual machines and have many VMCSs active under its management. A unique VMCS region is required for each virtual machine; a VMXON region is required for the VMM itself.

Note: we execute VXMON only once during boot. I suspect doing VMXON -> VMXOFF i -> VMXON constantly during context switch I highly doubt is the intended way one is supposed to do VCPUs on x86.

Indanz:
> Note: we execute VXMON only once during boot.

Ah, never mind then, I thought it was similar to vmenter.

> I suspect doing VMXON -> VMXOFF i -> VMXON constantly during context switch I highly doubt is the intended way one is supposed to do VCPUs on x86.

Yeah, this seems some weird x86 quirk.

src/model/statedata.c
dreamliner787-9: > I think we should remove manual alignment on ksSMP altogether, to prevent this from happening again in the future.

I've removed the manual alignment on `ksSMP` and everything still worked OK.

Indanz (approved):
Code-wise all is fine now, but could you put the removal of  ALIGN(L1_CACHE_LINE_SIZE)) in a separate commit with explanation of why we removed it? Something like:

"Outer ALIGN statements override inner ALIGN statements, leading to wrong alignment of the final structure. Remove the one on `ksSMP` to avoid nested ALIGNs."

Indanz:
By the way, if we want cache-line size alignment, then we should assure that `smpStatedata_t` has at least 64 byte alignment, not `ksSMP`, as the latter is an array and putting it there would only guarantee it for the first entry. So this shouldn't change anything for non-VTX build as far as alignment goes.

dreamliner787-9:
> Code-wise all is fine now, but could you put the removal of ALIGN(L1_CACHE_LINE_SIZE)) in a separate commit with explanation of why we removed it? Something like:
> 
> "Outer ALIGN statements override inner ALIGN statements, leading to wrong alignment of the final structure. Remove the one on `ksSMP` to avoid nested ALIGNs."

Yep all done.

lsf37:
@dreamliner787-9 can you please rebase? I picked the wrong merge order and didn't see that I can't update the branch.

dreamliner787-9:
> @dreamliner787-9 can you please rebase? I picked the wrong merge order and didn't see that I can't update the branch.

All good, done.

PR #1733, opened 2026-08-14
Optimized `merge_regions` function.

quasar32:
Current `merge_regions` function copies all regions after merged region causing the function to take quadratic time.
I wrote an optimized version that can operate in only linear time by have the `ndks_boot.reserved` overwrite itself.

### DIFF
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -47,21 +47,18 @@ BOOT_CODE p_region_t get_p_reg_kernel_img(void)
 BOOT_CODE static void merge_regions(void)
 {
     /* Walk through reserved regions and see if any can be merged */
-    for (word_t i = 1; i < ndks_boot.resv_count;) {
-        if (ndks_boot.reserved[i - 1].end == ndks_boot.reserved[i].start) {
+    word_t top = 0;
+    for (word_t i = 1; i < ndks_boot.resv_count; i++) {
+        if (ndks_boot.reserved[top].end == ndks_boot.reserved[i].start) {
             /* extend earlier region */
-            ndks_boot.reserved[i - 1].end = ndks_boot.reserved[i].end;
-            /* move everything else down */
-            for (word_t j = i + 1; j < ndks_boot.resv_count; j++) {
-                ndks_boot.reserved[j - 1] = ndks_boot.reserved[j];
-            }
-
-            ndks_boot.resv_count--;
-            /* don't increment i in case there are multiple adjacent regions */
+            ndks_boot.reserved[top].end = ndks_boot.reserved[i].end;
         } else {
-            i++;
+            /* add region */
+            top++;
+            ndks_boot.reserved[top] = ndks_boot.reserved[i];
         }
     }
+    ndks_boot.resv_count = ndks_boot.resv_count > 0 ? top + 1 : 0;
 }
 
 BOOT_CODE bool_t reserve_region(p_region_t reg)


Indanz (commented):
Good improvement, except `resv_count` ends up wrong if it was initially zero.

Please remove the stray empty new line.

quasar32:
Fixed issues with amended commit.

Indanz (approved):
Okay, I actually looked at the caller, and how it's called, `recv_count` is always greater than zero, but let's not count on that here.

You might want to look at `reserve_region()` region next, and the caller of that and the other boot code. It all seems extremely inefficient... But if you go down that road, it would be good to create some test code and run both the original and the changed code on different inputs and see if the results stay the same. This should speed up development too. We only test on qemu and a few systems, and many kind of errors would be undetected.

Indanz:
The C Parser error I would ignore for now (as boot code isn't verified yet), except if it just doesn't understand pre-increment or something, and putting the `++top` on a separate line fixes it. But please let the HW tests finish first before doing any further pushes.

quasar32:
Despite the nested loop and call to `merge_regions`,  `reserve_region` runs in linear time as it returns after executing the function/inner loop.

Indanz:
> Despite the nested loop and call to `merge_regions`, `reserve_region` runs in linear time as it returns after executing the function/inner loop.

Yes, but it's called once for each region added, so over all the complexity is still $\mathrm{O(n^2)}$.

quasar32:
Still an improvement over cubic time it was running.
`insert_region` also runs in linear time as it only calls `reserve_region` once.
The only way to improve the asymptotic times further of these function is to replace `reserved` and `freemem` with binary trees.
That isn't good idea since not only would greatly increase the size of code and the resulting kernel but also, by my understanding these arrays are very small (double digits number of elements), so it will in practice give worse performance.
I could look into full rewrite `init_freemem` through that would require practically rewriting `boot.c`.
Could be worth it if the resulting `boot.c` ends up be fewer lines of code like `merge_regions` did.

quasar32:
Got the C Parser check to pass, you were right that it was the pre-increment.

src/kernel/boot.c
```diff
@@ -47,21 +47,18 @@ BOOT_CODE p_region_t get_p_reg_kernel_img(void)
...
-            }
-
-            ndks_boot.resv_count--;
-            /* don't increment i in case there are multiple adjacent regions */
+            ndks_boot.reserved[top].end = ndks_boot.reserved[i].end;
         } else {
-            i++;
+            /* add region */
+            top++;
+            ndks_boot.reserved[top] = ndks_boot.reserved[i];
```
axel-h, line 58: Could avoid doing a copy-to-self. It's Mostly for stylistic reasons here and if this ever use for something bigger.. Practically it should not matter as the object is quite small. 
```suggestion
           /* add region, but don't do a self-copy */
           top++;
           if (top < i) {
               ndks_boot.reserved[top] = ndks_boot.reserved[i];     
           }
```

axel-h: We could even wipe `ndks_boot.reserved[i]` then to avoid leaving stale data.

Indanz: `top` is always less than `i`, as both increment at most once per iteration, but `top` starts at 0 and `i` at 1.

Wiping `ndks_boot.reserved[i]` would be tidier, but it's all boot memory that is going to disappear soon anyway.

axel-h: > top is always less than i, as both increment at most once per iteration, but top starts at 0 and i at 1.

No. after `top++` it can be equal to `i` and in this case (ie we don't merge regions) we do a self-copy.

Indanz: Ah yes, of course. (My brain was thinking we were in the other if-case for some reason.)

I think the branch would be slower than an unconditional copy though, and it's more code, so I think this version is fine.

Indanz:
This is ready to be merged I think, but it will require a proof-sync (even though boot code isn't verified, but to stop the CI preprocess errors), which will take some time considering that Gerwin is away at the moment. Feel free to remind us in a few weeks if it's still not merged.

src/kernel/boot.c
quasar32: I agree especially since the `reserved` can easily fit in L1 cache.

lsf37:
I would usually be against optimising functions that are called on the order of 10 regions, because there is a) no point and b) danger of introducing bugs by violating assumptions of other functions later that might depend on quirks.

In this case, if the proof check passes, I'm fine with it, because it makes the code simpler and as far as I can see easier to verify when we get there. Also, as far as I can tell via just review without a proof, it is fully functionally equivalent, so downstream functions will see the same behaviour.

Let's not try to optimise anything in reserve_region, though, please. 

What would be worth looking at for reserve_region and the other region code in general would be a review for correctness, because it has been more than a decade that we did that and code changed a lot since then without proofs. Similarly, a good set of tests as Indan mentioned. That would be great.

For instance, reserve_region returns bool_t, but there seem to be functions that ignore that return value. That seems fishy to me. There should at least be comments why that is Ok.

quasar32:
The `reserved_region` function can only fail if there too many reserved regions.
This is most likely a bug as `MAX_NUM_RESV_REG >= MAX_NUM_FREEMEM_REG` as indicated by using the command:
`grep -R "define MAX_NUM_RESV_REG"`

lsf37 (approved):
This is good to merge from my side now.

PR #1734, opened 2026-08-23
arch64,hyp: Set SWIO bit of HCR_EL2 reg always

kent-mcleod:
According to the architecture specification the effect of this bit being set is:
"Causes Non-secure EL1 execution of the data cache invalidate by set/way instructions to be treated as data cache clean and invalidate by set/way"

Without this bit set, the set/way cache operations can be used by one VM to invalidate the dirty lines of data belonging to another protection domain if they share a cache which breaks isolation.

Some implementations define this bit as RES1 which already effictively implements this change.

Supported implementations that don't define as RES1:
- Cortex-A57
- Cortex-A72

Supported implementations that do define as RES1:
- Cortex-A35
- Cortex-A53
- Cortex-A55
- Cortex-A76

### DIFF
--- a/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
+++ b/include/arch/arm/armv/armv8-a/64/armv/vcpu.h
@@ -15,11 +15,11 @@
 
 /* Note that the HCR_DC for ARMv8 disables S1 translation if enabled */
 /* Trap SMC and override CPSR.AIF */
-#define HCR_COMMON ( HCR_VM | HCR_RW | HCR_AMO | HCR_IMO | HCR_FMO | HCR_TSC)
+#define HCR_COMMON ( HCR_VM | HCR_SWIO | HCR_RW | HCR_AMO | HCR_IMO | HCR_FMO | HCR_TSC)
 
 /* Allow native tasks to run at EL0, but restrict access */
 #define HCR_NATIVE ( HCR_COMMON | HCR_TGE | HCR_TVM | HCR_TTLB | HCR_DC \
-                   | HCR_TAC | HCR_SWIO |  HCR_TSC )
+                   | HCR_TAC | HCR_TSC )
 
 #ifdef CONFIG_DISABLE_WFI_WFE_TRAPS
 #define HCR_VCPU   ( HCR_COMMON)


Indanz (approved):
Good idea.

can you do the same change for 32-bit Arm too:

https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/include/arch/arm/armv/armv7-a/armv/vcpu.h#L17-L22

midnightveil (changes requested):
I don't think this is actually necessary. Maybe it's a good idea to set anyway, but it's already the case that when `HCR_EL2.VM` is set then `DC ISW` acts as `DC CISW` (same with `DC IVAC` -> `DC CIVAC`).

Then `DC ISW` does `AArch64_DC(X{64}(t), CacheType_Data, CacheOp_Invalidate, CacheOpScope_SetWay);` [`AArch64_DC`](https://arm.jonpalmisc.com/latest_aarch64/shared_pseudocode#aarch64.functions.dc.AArch64_DC) will downgrade the operation to a clean-invalidate if *either* SWIO is 1, or `HCR_EL2().[DC,VM] != 00`.

```
        if (cacheop == CacheOp_Invalidate) && PSTATE.EL == EL1 && EL2Enabled() &&
          (HCR_EL2().SWIO == '1' || HCR_EL2().[DC,VM] != '00')) then
            cache.cacheop = CacheOp_CleanInvalidate;
        end;
```

That latter condition is already the case, as `HCR_VM` is always 1 per `HCR_COMMON`.


Also, from body text in DD0487L.b on p. D7-6522: (also, as far back as DD0487D.b, so it's not some recent feature)

> When executed at EL1, a DC ISW instruction performs a clean and invalidate, meaning it performs the same maintenance as a DC CISW instruction, if all of the following apply:
> - EL2 is implemented and enabled in the current Security state.
> - Either:
>   - The value of HCR_EL2.SWIO is 1, forcing a cache clean to perform a clean and invalidate.
>   - The value of HCR_EL2.VM is 1, meaning EL1&0 stage two address translation is enabled.

Indanz:
Okay, that's reassuring, but any reason not to set SWIO explicitly anyway?

Also, the [documentation](https://arm.jonpalmisc.com/latest_sysreg/AArch64-dc-isw) seems inconsistent:
```
if !IsFeatureImplemented(FEAT_AA64) then
    Undefined();
elsif PSTATE.EL == EL0 then
    Undefined();
elsif PSTATE.EL == EL1 then
    if EL2Enabled() && HCR_EL2().TSW == '1' then
        AArch64_SystemAccessTrap(EL2, 0x18);
    elsif EL2Enabled() && IsFeatureImplemented(FEAT_FGT) && (!HaveEL(EL3) || SCR_EL3().FGTEn == '1') && HFGITR_EL2().DCISW == '1' then
        AArch64_SystemAccessTrap(EL2, 0x18);
    else
        AArch64_DC(X{64}(t), CacheType_Data, CacheOp_Invalidate, CacheOpScope_SetWay);
    end;
elsif PSTATE.EL == EL2 then
    AArch64_DC(X{64}(t), CacheType_Data, CacheOp_Invalidate, CacheOpScope_SetWay);
elsif PSTATE.EL == EL3 then
    AArch64_DC(X{64}(t), CacheType_Data, CacheOp_Invalidate, CacheOpScope_SetWay);
end;
```

midnightveil:
> Also, the documentation seems inconsistent:

No, I linked the behaviour of `AArch64_DC` which silently changes invalidated to clean & invalidate.

In version D.d of the manual it was explicitly inlined that the change happens then the op, but I guess they got tired of repeating that check everywhere.

> Okay, that's reassuring, but any reason not to set SWIO explicitly anyway?

Not sure... I'd have to skim the pseudocode again but I don't think there's anything that checks for it specifically. I don't know why it exists.

Indanz:
> > Also, the documentation seems inconsistent:
> 
> No, I linked the behaviour of `AArch64_DC` which silently changes invalidated to clean & invalidate.

I linked to specifically DC ISW, which doesn't mention that behaviour, while your link says:

```
    if opscope == CacheOpScope_SetWay then
        let ss : SecurityState = SecurityStateAtEL(PSTATE.EL);
        cache.cpas = CPASAtSecurityState(ss);
        cache.shareability = Shareability_NSH;
        cache.(setnum, waynum, level) = DecodeSW(regval, cachetype);
        if (cacheop == CacheOp_Invalidate && PSTATE.EL == EL1 && EL2Enabled() &&
          (HCR_EL2().SWIO == '1' || HCR_EL2().[DC,VM] != '00')) then
            cache.cacheop = CacheOp_CleanInvalidate;
        end;

        CACHE_OP(cache);
        return;
    end;
```
So yes, that's being inconsistent.

> > Okay, that's reassuring, but any reason not to set SWIO explicitly anyway?
> 
> Not sure... I'd have to skim the pseudocode again but I don't think there's anything that checks for it specifically. I don't know why it exists.

Sure, but why block this change just because it's (probably) unnecessary? Enabling the bits won't do any harm.

midnightveil:
> I linked to specifically DC ISW, which doesn't mention that behaviour, while your link says:

Yes, mine is the primitive called by the "DC ISW" function, AArch64_DC. Note your call to `AArch64_DC(X{64}(t), CacheType_Data, CacheOp_Invalidate, CacheOpScope_SetWay);`, which is what has this check inside.

> Sure, but why block this change just because it's (probably) unnecessary? Enabling the bits won't do any harm.

Because, at the very least, the PR/commit description implies there is an issue.

I also don't know why this bit exists, so I'm wary of setting it. Aa far as I can tell, for seL4 it will change exactly zero behaviour. Which makes me think I've missed something, because then it seems pointless.

It's possible that the new ARM virtualisation extensions that seL4 does not use the TCR_EL2.VM bit (which is for the EL1&0 regime; which I *think* is not used for that?)

Indanz:
Okay, I'm convinced now that setting SWIO is unnecessary in our case. However, setting it would document the actual behaviour we are getting and want, so I still see no harm in doing this change.

SWIO affects behaviour when `HCR_EL2().[DC,VM] == '00'`. That we don't use it doesn't mean no one else does.

midnightveil:
Actually, I suppose the reason for setting it would be that the various CPU cores (as Kent mentioned) set it as RES1, which I think means you're probably supposed to keep it as 1. Though I suppose given what we have now works it must just be ignored.

> SWIO affects behaviour when HCR_EL2().[DC,VM] == '00'. That we don't use it doesn't mean no one else does.

I'm just having a hard time figuring out what kind of system would actually use it. Why would you want guest OS that doesn't use stage 2 translations (?). But it is true it is architecturally allowed.

kent-mcleod:
> I don't think this is actually necessary. Maybe it's a good idea to set anyway, but it's already the case that when `HCR_EL2.VM` is set then `DC ISW` acts as `DC CISW` (same with `DC IVAC` -> `DC CIVAC`).

You're right. In the 8.0 manual that I was using it had:
```
The architecture does not require cache cleaning when switching between virtual machines. Cache
invalidation by set/way must not present an opportunity for one virtual machine to corrupt state associated
with a second virtual machine. To ensure this requirement is met, Non-secure clean by set/way operations
can be upgraded to clean and invalidate by set/way.
```
And then right after:
```
The AArch64 Data cache invalidate by set/way instruction DC ISW, at EL1 and EL0, and the AArch32 Data
cache invalidate by set/way instruction DCISW, at EL1, perform a cache clean as well as a cache invalidation
if both of the following apply:
—
The value of HCR_EL2.SWIO is 1.
—
The instruction is executed in Non-secure state, or EL3 is not implemented.
```

I went back and checked, and higher on the same page it has:
```
The AArch64 Data cache invalidate instructions, DC IVAC and DC ISW, at EL1 and EL0, and the AArch32
Data cache invalidate instructions DCIMVAC and DCISW, at EL1, perform a cache clean as well as a cache
invalidation if both of the following apply:
—
The value of HCR.VM is 1.
—
The instruction is executed in Non-secure state, or EL3 is not implemented.
```

kent-mcleod:
> I'm just having a hard time figuring out what kind of system would actually use it. Why would you want guest OS that doesn't use stage 2 translations (?). But it is true it is architecturally allowed.

When the EL1 kernel isn't a guest OS, but is more of a DOM0 OS, or isn't returning to EL2 i guess.  The elfloader for EL1 kernel sets up HCR_EL2 with VM and DC = {0, 0}.

midnightveil:
Yeah, though I expect in that situation you'd want access to cache invalidate.

Similarly for e.g. EL1 seL4, it would probably want access to the invalidate instructions, since there's no EL2 involved. I wouldn't expect a need for SWIO because then EL1 would be the highest normal-world (non-secure) authority.

PR #1738, opened 2026-08-24
aarch64,hyp: Add compile_assert for doFlush ranges

kent-mcleod:
When the kernel executes in EL2, the kernel invocations for performing cache maintenance operations first translate the physical address into the kernel window and perform the maintenance operation via the kernel's own mapping.

To ensure that the mapping exists, the largest device UT address that's given to user level needs to be mapped in the kernel window, otherwise a translation fault can be generated.

Add a static_assert to capture this requirement.

(This compile_assert will fail on all platforms currently because the kernel window is smaller than 2^39, and all definitions of `CONFIG_PADDR_USER_DEVICE_TOP` are larger than 2^39).

### DIFF
--- a/include/arch/arm/arch/64/mode/hardware.h
+++ b/include/arch/arm/arch/64/mode/hardware.h
@@ -303,5 +303,6 @@
 compile_assert(ut_max_less_than_canonical, CONFIG_PADDR_USER_DEVICE_TOP <= BIT(47));
 #ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
 compile_assert(ut_max_is_canonical, (PPTR_BASE + CONFIG_PADDR_USER_DEVICE_TOP) <= BIT(48));
+compile_assert(ut_max_is_in_kernel_window, (PPTR_BASE + CONFIG_PADDR_USER_DEVICE_TOP) <= PPTR_TOP);
 #endif
 #endif


kent-mcleod:
- The kernel currently allows cache maintenance invocations to be performed on mapped address ranges constructed from device frames, but device frames aren't guaranteed to cover address ranges that won't cause a hardware exception when cache maintenance operations are performed on them.
- Ultimately, the aarch64 kernel should require all cache maintenance operations to be performed from user space which the hardware supports, except for a way to perform a full instruction cache invalidation to point of unification for CPUs that have a VIPT instruction cache where `IC ALL*` are the only architecture provided way of invalidating all aliases for a particular address range.

midnightveil:
We had a related discussion in this PR: https://github.com/seL4/seL4/pull/1619, which would have caused non-RAM regions in general to not be mapped into the kernel physical window.

Specifically I think one way we could solve this (without removing it entirely) would be to only allow these operations on "normal" memory (i.e. non-device): https://github.com/seL4/seL4/pull/1619#issuecomment-5138608615.

I agree with both your points, if userspace can do them — or if userspace can do the technically-slower (but likely not actually slower) clean&invalidate instead of the invalidate — they should, if possible, and the ability for the kernel should do them should be removed.

midnightveil:
re:

> but device frames aren't guaranteed to cover address ranges that won't cause a hardware exception when cache

specifically: at least for ARM, the cache operations do not care about device memory. writes to device memory themselves as a result of the cache operations could theoretically cause SError, but that is no different to userspace doing it directly without cached mappings.

> To ensure that the mapping exists, the largest device UT address that's given to user level needs to be mapped in the kernel window, otherwise a translation fault can be generated.

This is also false, there's a check that the paddr lies between PADDR_START AND PADDR_END which I believe corresponds to mapped addresses (currently)

Indanz:
> > To ensure that the mapping exists, the largest device UT address that's given to user level needs to be mapped in the kernel window, otherwise a translation fault can be generated.
> 
> This is also false, there's a check that the paddr lies between PADDR_START AND PADDR_END which I believe corresponds to mapped addresses (currently)

There's a `if (pstart < PADDR_BASE || ((end - start) + pstart) > PADDR_TOP` check in `decodeARMFrameInvocation`, but this is missing from `decodeARMVSpaceRootInvocation`.

The other exception is that I think it should be the user space's responsibility to flush page tables when they are used for non-coherent SMMU, so that we can remove the clean to RAM code and speed everything up in the general code. For that PT objects also needs to support some cache ops, but at least in that case we know it's normal memory mapped by the kernel. (Alternative is to force users to declare SMMU use with a special flag when mapping, so the kernel knows whether to clean to RAM or not.)

lsf37:
For decodeARMVSpaceRootInvocation we do check that we are in the kernel window here

https://github.com/seL4/seL4/blob/c2b8ff5a83d3c327eea8ab61cb99664484b22656/src/arch/arm/64/kernel/vspace.c#L1439-L1440

and for decodeARMFrameInvocation it is here:

https://github.com/seL4/seL4/blob/c2b8ff5a83d3c327eea8ab61cb99664484b22656/src/arch/arm/64/kernel/vspace.c#L1765

so I think we are at least safe in the current kernel window settings. If we do what is proposed in #1619, we will need something more sophisticated, but that is for then.

Of course this does not mean that we're not in a device area that is not actual RAM. That is still a problem, but it's also not covered by the compile time assert proposed here.

I think this is one we can actually skip.

(Edit: well, I should double check that the USER_TOP check is actually sufficient)

Indanz:
> For decodeARMVSpaceRootInvocation we do check that we are in the kernel window here
> 
> https://github.com/seL4/seL4/blob/c2b8ff5a83d3c327eea8ab61cb99664484b22656/src/arch/arm/64/kernel/vspace.c#L1439-L1440

That only checks the virtual address, not the physical address.

lsf37:
> That only checks the virtual address, not the physical address.

I (incorrectly) assumed that this would tell us something about the window, but it's a user virtual address, not kernel virtual address. So the VSpace path is in fact missing the check that `decodeARMFrameInvocation` has and we should add one. (And also check AArch32 hyp).

I think I'd prefer an explicit check in the decode path over a compile time assert, though. We're already computing the physical start/end addresses, so we can compare against PADDR_BASE/PADDR_TOP as we do on the frame path.

It'd be nice to not constrain things by kernel window statically if we don't have to.

midnightveil:
Would it be nicer instead to enforce that it should only be non-device memory? Without checking I believe all normal memory should be in the kernel window (same as for e.g. untyped reset).

It might be easier for verification purposes to check the window directly, IDK, but if we want to do #1619 to fix the speculation issue we probably should do based on if-is-device. 

Alternatively we could remove everything to do with in-kernel cache maintenance, though admittedly that does prevent some technically-allowable (but not actually useful (?)) operations and likely needs some more thought.

---

Wondering if we should just fix this missing one check using the same pattern, or if we should fix a few other issues along the way.

lsf37:
> Would it be nicer instead to enforce that it should only be non-device memory? Without checking I believe all normal memory should be in the kernel window (same as for e.g. untyped reset).
>
> It might be easier for verification purposes to check the window directly, IDK, but if we want to do #1619 to fix the speculation issue we probably should do based on if-is-device.

I guess we should really be checking both, but there is plenty of normal RAM that will be device frames/untyped, especially in multikernel. It would still usually be in the kernel window I assume, although I'm not sure there is anything that would forbid hardware to put RAM into exotic places in the physical address space.

> Alternatively we could remove everything to do with in-kernel cache maintenance, though admittedly that does prevent some technically-allowable (but not actually useful (?)) operations and likely needs some more thought.

I would not mind removing all of the cache operations when they can safely be delegated to user space anyway. (Apart from the instruction cache invalidation that Kent mentions, but it would at least reduce the problem space).

> Wondering if we should just fix this missing one check using the same pattern, or if we should fix a few other issues along the way.

My preference would be to make it safe first for the current state and then keep investigating.

midnightveil:
> I guess we should really be checking both, but there is plenty of normal RAM that will be device frames/untyped, especially in multikernel. It would still usually be in the kernel window I assume, although I'm not sure there is anything that would forbid hardware to put RAM into exotic places in the physical address space.

I suppose my thinking is that it might be fine sometimes (multikernel) but I don't know if there's a way to make it so always. However maybe it's not worth worrying about.

kent-mcleod:
I think the aarch32 and aarch64 contexts are different enough to treat them differently:
### Under aarch64: 
1. The ISA allows all useful cache maintenance to be performed from EL0 except Virtually indexed-Physically tagged (VIPT) instruction caches which needs an operation  which can only be done in EL1. 
2. The kernel cannot cause a hardware exception when performing decode or invoke parts of an invocation.
3. The kernel only accesses memory for objects created from regular untypeds where it is safe to perform cache maintenance ops without causing a fault. The kernel needs to do things like clean frames before they are mapped uncached into user space. it will always need to do this in the kernel.

In the aarch64 case, the kernel only needs to support an invocation for invalidating the icache for VIPT cache systems and otherwise the explicit operations should be done at user space.

### Under aarch32:
The hardware doesn't provide ways to do cache maintenance in user mode and as such the kernel needs invocations.
When the vspace invocations were added, device UT didn't exist.
In the same way that the kernel can't use device UT for things like IPC buffer, it can't safely perform CMO inside frames constructed from device UT. 
To do safe CMO, the kernel needs to know that the memory range is regular mem and derived from a regular untyped.
For frame based invocations this is in the frame cap.
For vspace based invocations we'd need to store the frame type as a bit in the hardware PTE in a software controlled bit (which hopefully exists on aarch32).
Otherwise we'd need to remove the vspace invocations because I don't think there's a way to make them safe and still allow device frames to be constructed and mapped into vspace and not have the kernel have to handle synchronous exceptions from CMOs performed in kernel mode.

midnightveil:
Don't the VSpace-based flush operations only support flushes of a single page? In essence it's a shortcut, so it's easy enough to check the frame that gets flushed in the end is not device.

If it wasn't we could always check each page separately, though that might be slow.

kent-mcleod:
> In essence it's a shortcut, so it's easy enough to check the frame that gets flushed in the end is not device.

How would this be checked? The invocation doesn't take a frame cap, so the only information that kernel can get is from the hardware page table entry. There isn't a way to go from the PTE contents to the frame cap at the moment.

kent-mcleod:
> It might be easier for verification purposes to check the window directly, IDK, but if we want to do #1619 to fix the speculation issue we probably should do based on if-is-device.

@midnightveil: Were you meaning that if the kernel window was changed to only map non-device memory then the vspace flush ops could check those page table entries?

midnightveil:
Right, I misread what `decodeARMPageDirectoryInvocation` is doing, of course it's walking the actual page tables...

midnightveil:
> > It might be easier for verification purposes to check the window directly, IDK, but if we want to do #1619 to fix the speculation issue we probably should do based on if-is-device.
> 
> Were you meaning that if the kernel window was changed to only map non-device memory then the vspace flush ops could check those page table entries?

No, I was meaning that it might be easier for verification to do the same check that we do for the frame-based invocations for the vaddr ones. Since we can still get the paddr from the PTE, and thus compare to the PPTR window, which is what we do for frame-based ones. (i.e. I was thinking it was easiest to do the same check rather than a new mechanism for the check).


However, doing an `AT` instruction could probably work too, and was somewhat in my mind as an alternative to checking the device. It would be one way for instance of solving #1619 without using the device-bit (which, as you correctly point out, is not possible for the vspace invocations) that doesn't require keeping around and iterating over the ram regions array.

lsf37:
> > > It might be easier for verification purposes to check the window directly, IDK, but if we want to do #1619 to fix the speculation issue we probably should do based on if-is-device.
> > 
> > Were you meaning that if the kernel window was changed to only map non-device memory then the vspace flush ops could check those page table entries?
> 
> No, I was meaning that it might be easier for verification to do the same check that we do for the frame-based invocations for the vaddr ones. Since we can still get the paddr from the PTE, and thus compare to the PPTR window, which is what we do for frame-based ones. (i.e. I was thinking it was easiest to do the same check rather than a new mechanism for the check).

That I could imagine doing, yes.


> However, doing an `AT` instruction could probably work too, and was somewhat in my mind as an alternative to checking the device. It would be one way for instance of solving #1619 without using the device-bit (which, as you correctly point out, is not possible for the vspace invocations) that doesn't require keeping around and iterating over the ram regions array.

I would want to avoid `AT` instructions. We'd just have to model what they do on the verification side and the likelihood of getting that wrong is high, because the instruction works on a different abstraction level than what we need to use. We're already using an address translate in the fault handling path, and that is only Ok, because we don't make any claims about what it returns (we just pass it on to user level). This was already painful for infoflow, because we had to introduce axioms on what this instruction can read. Using it for a check that the kernel depends on would not be good.

lsf37:
> > No, I was meaning that it might be easier for verification to do the same check that we do for the frame-based invocations for the vaddr ones. Since we can still get the paddr from the PTE, and thus compare to the PPTR window, which is what we do for frame-based ones. (i.e. I was thinking it was easiest to do the same check rather than a new mechanism for the check).
> 
> That I could imagine doing, yes.

Ah, right, but we don't have a frame cap. We could walk the user page table, but that is less fun. Still better than an AT instruction from the verification perspective, though. At least we'd be talking about semantics we can see.

midnightveil:
For the Vspace flush issue: We're already walking the user page table for vspace flush to get an address, so we can just use this to fix the immediate issue by checking against PPTR window in the same way as we do for frame-based flush.

The rest of discussion is about future hypothetical fixed for #1619; we can't determine if device  (instead of bu PPTT window) because we don't get sel4 frame cap, just the physical address.

Similarly, I was suggesting AT on kernel PPTR mapping to determine if mapped in PPTR window: could do the same thing via walking kernel page tables or by walking the normal_regions (avail_reg) array, or some secret third thing. But this is only relevant if we plan to do #1619 and keep these operations around.

lsf37:
Sorry, yes, I should have read that in detail first.

Ok, proposed order of things: we do the PTE base address check now to fix the immediate issue. I will try to do the proof update for that soon (maybe adding a precondition to the flush ops on AArch64 hyp that the address has to be in the kernel window, so that we don't miss any others). After that or concurrently, we write up an RFC to remove the VSpace flush ops for AArch64, because they can be done from user level.

We still need to figure out what to do for AArch32 hyp. I agree with Kent there: if there is a bit in page/section PTEs that we can use for "is from device cap", then we can do the check. Otherwise I don't see an option really other than to remove the VSpace flush ops and only leave the frame cap flush ops.

midnightveil:
I've created an sel4test which demonstrates this: https://github.com/seL4/sel4test/pull/158

midnightveil:
@kent-mcleod Can I push to this the PR fixes for the page address checks? That way we keep the discussion together?

(The fixes are also quite trivial just copying a check from the page-base flushed ones, so if you want to add them that's also OK)

PR #1739, opened 2026-08-24
aarch64,tlb: avoid using invalid PT address

lsf37:
While the PT address as such is irrelevant for the flush operation, it should still point to a valid PT and avoid the possibility of adding stale TLB mappings via any potential speculative execution.


lsf37:
Not sure what's up with the machine queue..

Indanz (approved):

PR #1741, opened 2026-08-25
aarch64,types: make hw_asid_t 16 bit for non-hyp

### LINKED ISSUE #1735: aarch64: invalidateTranslationASID is called with wrong argument type
[`invalidateTranslationASID`](https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/include/arch/arm/arch/machine/tlb.h#L44) takes `hw_asid_t`, which is unconditionally [declared `uint8_t`](https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/include/arch/arm/arch/types.h#L26), but on AArch64 in non-hyp mode [it is called with `asid_t`](https://github.com/seL4/seL4/blob/1326364bc9135d9445d936ebc01e38a402c1f4c6/src/arch/arm/64/kernel/vspace.c#L1079), which is 16 bit. Which means for seL4 ASIDs > 255 we're flushing the wrong hw ASID.

(On the verified hyp kernel, the types do match)

We already have a [draft PR](https://github.com/seL4/seL4/pull/1628) for changes to switch on compiler warnings for this kind of thing, but we need to schedule time to do the necessary updates for it (incl proof changes, but they should be minor).

lsf37:
Currently all supported AArch64 platforms have 8 bit VMIDs. When new platforms are added that have larger VMIDs (e.g. Orin), they either need to be added to the #ifdef or we need to refactor this further.

Previously the AArch64 non-hyp code expected 16 bit hardware ASIDs, but used an 8-bit type, with the consequence of flushing the wrong TLB entries when there are more than 256 threads.

Fixes #1735

### DIFF
--- a/include/arch/arm/arch/64/mode/machine.h
+++ b/include/arch/arm/arch/64/mode/machine.h
@@ -161,6 +161,26 @@ static inline bool_t checkTCR_EL2(void)
     return (tcr_el2 == TCR_EL2_DEFAULT);
 }
 
+/* Supported ASID size field of ID_AA64MMFR0_EL1 */
+#define ID_AA64MMFR0_EL1_ASIDBITS(x)  (((x) >> 4) & 0xf)
+#define ID_AA64MMFR0_EL1_ASIDBITS_16  2
+/* Selected ASID size in TCR_EL1 */
+#define TCR_EL1_AS                    BIT(36)
+
+/* For non-hyp AArch64 configs, check that 16-bit ASIDs are supported
+ * and enabled. */
+static inline bool_t check16BitASID(void)
+{
+    word_t id_aa64mmfr0_el1 = 0;
+    word_t tcr_el1 = 0;
+
+    MRS("id_aa64mmfr0_el1", id_aa64mmfr0_el1);
+    MRS("tcr_el1", tcr_el1);
+
+    return ID_AA64MMFR0_EL1_ASIDBITS(id_aa64mmfr0_el1) == ID_AA64MMFR0_EL1_ASIDBITS_16 &&
+           (tcr_el1 & TCR_EL1_AS);
+}
+
 static inline void setCurrentKernelVSpaceRoot(ttbr_t ttbr)
 {
     dsb();
--- a/include/arch/arm/arch/32/mode/types.h
+++ b/include/arch/arm/arch/32/mode/types.h
@@ -14,3 +14,5 @@ compile_assert(long_is_32bits, sizeof(unsigned long) == 4)
 #define wordBits (1 << wordRadix)
 
 typedef uint32_t timestamp_t;
+
+typedef uint8_t  hw_asid_t;
--- a/include/arch/arm/arch/64/mode/types.h
+++ b/include/arch/arm/arch/64/mode/types.h
@@ -14,3 +14,9 @@ compile_assert(long_is_64bits, sizeof(unsigned long) == 8)
 #define wordBits (1 << wordRadix)
 
 typedef uint64_t timestamp_t;
+
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+typedef uint8_t  hw_asid_t;
+#else
+typedef uint16_t hw_asid_t;
+#endif
--- a/include/arch/arm/arch/types.h
+++ b/include/arch/arm/arch/types.h
@@ -23,8 +23,6 @@ typedef word_t node_id_t;
 typedef word_t cpu_id_t;
 typedef word_t dom_t;
 
-typedef uint8_t  hw_asid_t;
-
 enum hwASIDConstants {
     hwASIDMax = 255,
     hwASIDBits = 8
--- a/src/arch/arm/kernel/boot.c
+++ b/src/arch/arm/kernel/boot.c
@@ -194,6 +194,14 @@ BOOT_CODE static bool_t init_cpu(void)
         vcpu_boot_init();
     }
 
+#ifdef CONFIG_ARCH_AARCH64
+    /* After activate_kernel_vspace(), so printf() is available. */
+    if (!config_set(CONFIG_ARM_HYPERVISOR_SUPPORT) && !check16BitASID()) {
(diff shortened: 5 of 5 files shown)


lsf37:
We should really also check that the hardware we're running on actually has 16 bit ASIDs. It looks to me that the hardware can report [8 or 16](https://support.arm.com/documentation/111107/2026-06/AArch64-Registers/ID-AA64MMFR0-EL1--AArch64-Memory-Model-Feature-Register-0?lang=en).

Maybe an assert in the boot code somewhere? Debug would be enough. This would mostly be to trigger for hardware ports that don't think about this, because for 8 bit ASIDs everything would likely just work, but we'd have the same kind of failure we have now.

KurtWu10:
> Currently all supported AArch64 platforms have 8 bit VMIDs.

According to the manual, both [Cortex-A55](https://support.arm.com/documentation/100442/0200/Register-descriptions/AArch64-system-registers/ID-AA64MMFR1-EL1--AArch64-Memory-Model-Feature-Register-1--EL1) and [Cortex-A76](https://support.arm.com/documentation/100798/0401/AArch64-System-registers/ID-AA64MMFR1-EL1--AArch64-Memory-Model-Feature-Register-1--EL1) support 16-bit ASID/VMID.

Indanz (approved):

lsf37:
> > Currently all supported AArch64 platforms have 8 bit VMIDs.
> 
> According to the manual, both [Cortex-A55](https://support.arm.com/documentation/100442/0200/Register-descriptions/AArch64-system-registers/ID-AA64MMFR1-EL1--AArch64-Memory-Model-Feature-Register-1--EL1) and [Cortex-A76](https://support.arm.com/documentation/100798/0401/AArch64-System-registers/ID-AA64MMFR1-EL1--AArch64-Memory-Model-Feature-Register-1--EL1) support 16-bit ASID/VMID.

Right, we just only use 8 bits of them. I'll update the commit message to avoid future confusion.

lsf37:
Ok, updated the commit message and added another commit with the check for 16-bit ASIDs. @Indanz do can you have another short look before I merge?

midnightveil:
Hm, was there actually an issue for non-hyp?

findFreeHwAsids uses -1 here, so the max would be constrained by this value anyway?

https://github.com/seL4/seL4/blob/654cc889f02a35252d95ec11d0b06fec59d3f0ee/src/arch/arm/64/kernel/vspace.c#L968

Maybe we should instead define MAX_HW_ASID somewhere as a bit more explicit version that doesn't depend on precise type sizes.

(I think it's safe afterwards and before?)

Indanz:
> Hm, was there actually an issue for non-hyp?

Yes, non-hyp doesn't use `findFreeHWASID`, but instead uses the `asid_t` value directly. Truncating that to 8 bits is wrong.

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -161,6 +161,26 @@ static inline bool_t checkTCR_EL2(void)
     return (tcr_el2 == TCR_EL2_DEFAULT);
 }
 
+/* Supported ASID size field of ID_AA64MMFR0_EL1 */
+#define ID_AA64MMFR0_EL1_ASIDBITS(x)  (((x) >> 4) & 0xf)
+#define ID_AA64MMFR0_EL1_ASIDBITS_16  0b0010
```
Indanz: I think standard C-support for this bitwise notation is fairly new, and we don't use it much elsewhere, only capdl.c. Maybe just use 2?

include/arch/arm/arch/64/mode/machine.h
```diff
@@ -161,6 +161,26 @@ static inline bool_t checkTCR_EL2(void)
...
+static inline bool_t check16BitASID(void)
+{
+    word_t id_aa64mmfr0_el1 = 0;
+    word_t tcr_el1 = 0;
+
+    MRS("id_aa64mmfr0_el1", id_aa64mmfr0_el1);
+    MRS("tcr_el1", tcr_el1);
+
+    return ID_AA64MMFR0_EL1_ASIDBITS(id_aa64mmfr0_el1) == ID_AA64MMFR0_EL1_ASIDBITS_16 &&
+           (tcr_el1 & TCR_EL1_AS) != 0;
```
Indanz: Strictly speaking, just checking TCR_EL1 would be sufficient.

```suggestion
    return ID_AA64MMFR0_EL1_ASIDBITS(id_aa64mmfr0_el1) == ID_AA64MMFR0_EL1_ASIDBITS_16 &&
           (tcr_el1 & TCR_EL1_AS);
```

Indanz (approved):
Just nitpicks. Alternative is to do this check on ASID assignment instead, but we can always do that later.

include/arch/arm/arch/64/mode/machine.h
lsf37: Happy to, yes. I just took it to show the same bit pattern as in the spec, but 2 is fine.

lsf37: turns out the C parser already didn't like the 0b syntax, which I should have known..

lsf37:
> > Hm, was there actually an issue for non-hyp?
> 
> Yes, non-hyp doesn't use `findFreeHWASID`, but instead uses the `asid_t` value directly. Truncating that to 8 bits is wrong.

So, together with Kurt's point that we already do support platforms with 16-bit VMID, we should probably make that 1-1 mapping available there as well. Maybe not highest priority at the moment, but I'll add an issue for it.

Indanz:
> So, together with Kurt's point that we already do support platforms with 16-bit VMID, we should probably make that 1-1 mapping available there as well. Maybe not highest priority at the moment, but I'll add an issue for it.

Ideally, we would support both and do runtime selection between the two. But for that to work, the hw asid must either be 16-bit, or always 8-bit and only used by the asid lookup code (this is currently almost the case, with `invalidateTranslationASID*` the only exception).

PR #1743, opened 2026-08-26
Replace all unreachable defaults with assert(0)

0aids:
Better code convention so that whenever a new cap type is added, all related cap data functions obviously fail.

For `Arch_isCapRevocable` and `isCapRevocable`, should this be changed to exhaustive switches? 
(marked by my comments in this PR). 

Also are there any other places where this should be done?

Thanks!

### DIFF
--- a/include/arch/arm/arch/32/mode/object/structures.h
+++ b/include/arch/arm/arch/32/mode/object/structures.h
@@ -304,6 +304,7 @@ static inline word_t CONST cap_get_archCapSizeBits(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return 0;
     }
 }
@@ -350,6 +351,7 @@ static inline bool_t CONST cap_get_archCapIsPhysical(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return false;
     }
 }
@@ -394,6 +396,7 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return NULL;
     }
 }
--- a/include/arch/x86/arch/64/mode/object/structures.h
+++ b/include/arch/x86/arch/64/mode/object/structures.h
@@ -150,6 +150,7 @@ static inline word_t CONST cap_get_modeCapSizeBits(cap_t cap)
         return seL4_PDPTBits;
 
     default:
+        assert(!"Unknown cap type");
         return 0;
     }
 }
@@ -169,6 +170,7 @@ static inline bool_t CONST cap_get_modeCapIsPhysical(cap_t cap)
         return true;
 
     default:
+        assert(!"Unknown cap type");
         return false;
     }
 }
@@ -187,6 +189,7 @@ static inline void *CONST cap_get_modeCapPtr(cap_t cap)
         return PDPT_PTR(cap_pdpt_cap_get_capPDPTBasePtr(cap));
 
     default:
+        assert(!"Unknown cap type");
         return NULL;
     }
 }
--- a/include/arch/x86/arch/object/structures.h
+++ b/include/arch/x86/arch/object/structures.h
@@ -310,6 +310,7 @@ static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
         return cap_get_capType(srcCap) == cap_io_port_control_cap;
 
     default:
+        // @0aids: Intentionally non-exhaustive?
         return false;
     }
 }
--- a/include/arch/arm/arch/64/mode/object/structures.h
+++ b/include/arch/arm/arch/64/mode/object/structures.h
@@ -113,6 +113,7 @@ static inline word_t CONST cap_get_archCapSizeBits(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return 0;
     }
 }
@@ -155,6 +156,7 @@ static inline bool_t CONST cap_get_archCapIsPhysical(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return false;
     }
 }
@@ -196,6 +198,7 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
         return NULL;
     }
 }
--- a/include/arch/arm/arch/object/structures.h
+++ b/include/arch/arm/arch/object/structures.h
@@ -26,6 +26,7 @@ static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 #endif
 
     default:
+        // @0aids: intentionally non-exhaustive?
         return false;
     }
 }
--- a/include/arch/riscv/arch/object/structures.h
+++ b/include/arch/riscv/arch/object/structures.h
@@ -145,6 +145,7 @@ static inline void *CONST cap_get_archCapPtr(cap_t cap)
 
 static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
 {
+    // @0aids: intentionally non-exhaustive?
     return false;
 }
 
--- a/src/object/objecttype.c
+++ b/src/object/objecttype.c
@@ -1019,6 +1019,7 @@ bool_t CONST isCapRevocable(cap_t derivedCap, cap_t srcCap)
         return true;
 
     default:
+        // @0aids: intentionally non-exhaustive?
         return false;
     }
 }


include/arch/arm/arch/32/mode/object/structures.h
```diff
@@ -304,6 +304,7 @@ static inline word_t CONST cap_get_archCapSizeBits(cap_t cap)
 
     default:
         /* Unreachable, but GCC can't figure that out */
+        assert(!"Unknown cap type");
```
lsf37, line 307: If at all, we should use `assert(0)` and a comment, not a negated pointer to a string literal.

I'm not a fan, though. Adding asserts adds failure to the debug kernel that the proofs don't see, and if you get an assert wrong it will be highly misleading. 

If there is no proof impact, I'm ok with adding it to the places that already have an "unreachable" comment and are obvious.

For all others, you should check the C proof preconditions for the corresponding function first and link to it for evidence (find the `ccorres` or `spec` lemma for that function, and check if the precondition precludes all branches that are not in the switch statement).

include/arch/x86/arch/64/mode/object/structures.h
```diff
@@ -150,6 +150,7 @@ static inline word_t CONST cap_get_modeCapSizeBits(cap_t cap)
         return seL4_PDPTBits;
 
     default:
+        assert(!"Unknown cap type");
```
lsf37, line 153: how do you know? the function could be called on some other cap type and expect 0

include/arch/x86/arch/object/structures.h
```diff
@@ -310,6 +310,7 @@ static inline bool_t CONST Arch_isCapRevocable(cap_t derivedCap, cap_t srcCap)
         return cap_get_capType(srcCap) == cap_io_port_control_cap;
 
     default:
+        // @0aids: Intentionally non-exhaustive?
```
lsf37, line 313: yes, it is. By default they are not exhaustive, that's what the `default:` statement is for.

include/arch/arm/arch/32/mode/object/structures.h
midnightveil: > If at all, we should use assert(0) and a comment, not a negated pointer to a string literal.

this is already done for RISC-V arches, which is why I suggested to Aidan duplicating this.

> If there is no proof impact, I'm ok with adding it to the places that already have an "unreachable" comment and are obvious.

The idea is that this should be attached to the ones that are unreachable supposedly.

lsf37: > > If at all, we should use assert(0) and a comment, not a negated pointer to a string literal.
> 
> this is already done for RISC-V arches, which is why I suggested to Aidan duplicating this.

ah, that's where this is coming from. I swear everyone introduces their own conventions in every single port. I guess the proofs don't see it, so it doesn't really matter that much, but something bugs me about this pattern. If others are keen on it, I can live with it, though.

> > If there is no proof impact, I'm ok with adding it to the places that already have an "unreachable" comment and are obvious.
> 
> The idea is that this should be attached to the ones that are unreachable supposedly.

It should only be the ones that are unreachable definitely. If they are right, asserts are good for debugging, but if they are wrong, they are a problem. I'm also uncomfortable with all these asserts in MCS. We have no idea if they are actually right, and the proofs don't say anything about them directly.

include/arch/x86/arch/object/structures.h
midnightveil: > For Arch_isCapRevocable and isCapRevocable, should this be changed to exhaustive switches?
(marked by my comments in this PR).

From the PR description. This one would obviously need a proof update, unfortunately.

My argument for this is that it would make development easier, and if we annotate the default case with the UNREACHABLE macro following the assert a compiler should be easily able to generate the same code in both cases (for release mode).

But this is a bigger change, and maybe contentious: which is why I suggested putting it here for discussion.

include/arch/arm/arch/32/mode/object/structures.h
midnightveil: Additional context: for PMU features Krishnan Winter developed last year, part of this was experimentation with adding a new vPMU cap¹. As part of this, adding a case for this new cap type to isPhysical was missed. I believe at least a few days of time was wasted for this. (There was another issue where the object sizes didn't match either... but that can't really be fixed generically, aside from adding the same static asserts everything else has).

Aidan's been working on another project (Record & Replay debugging) and as part of this needs performance conters context switched. Somehow Krishnan's bug fixes for isPhysical et al aren't in any Git branches we had, and Aidan has run into the same issue and wasted a few days - and it turns out it was the same issue again.

One could probably say skill issue... and maybe that's true, but it has hard to debug symptoms (memory corruption).

¹ Honestly, like FPU, would be easier to just make it embedded in the TCB, the downside being bigger memory usage. But then again we don't have proof funding for it, so. So the proof catching these cases won't happen.

midnightveil: > t doesn't really matter that much, but something bugs me about this pattern. 

Another thing we could do is use UNREACHABLE, and define it in release builds to do `fail("Unknown cap type")`. Since we do have that function call. Although not sure if that has line numbers.

Indanz: I think we had the same missing isPhysical bug for VCPUs or something in the past.

include/arch/x86/arch/object/structures.h
lsf37: Exhaustive matches don't really help in C. The compiler doesn't check they are exhaustive, and you're relying on a runtime assert that you're very likely to not hit. But the cost is that you have to enumerate a lot of cases and you have to update that every time you add a cap or an object. People are almost guaranteed to forget some of these (we certainly have before). If there is a safe default like here, that is much better at preventing mistakes on average.

Of course it's not sure, but that is what the proofs are for, they *will* catch these cases (in either method).

So the strategy so far has been: if the case distinction is dense (enumerates almost all cases already anyway), make it exhaustive. And if the case distinction is sparse with a good default, use the default.

include/arch/arm/arch/32/mode/object/structures.h
lsf37: That is true, we did. That one at least would be likely to be hit by an assert, because it's called often, so I guess that one would be fine to do. We should not do this randomly, though, but with intent.

There are plenty of small functions that are called only during revocation for instance, and people never test revoking the new objects they build. So the assert never fires in any test. Maybe we could make a generic test that forces revocation for all object types, but I'm not sure if that is possible. If it's not generic people will of course not update the tests.

include/arch/x86/arch/object/structures.h
midnightveil: > People are almost guaranteed to forget some of these (we certainly have before)

I suppose then there's no point. My hope was making it exhaustive would make it easier to not miss the cases where it should be different from the default, but I guess maybe not.

lsf37: For the cases that are not covered by the proof. Could we add a comment marker that people can search for when they add new caps? It would make it much easier to find where potential updates are needed and could maybe even be turned into a todo list one can work through.

midnightveil: That would be a good idea. It would also make it easier to review.

> So the strategy so far has been: if the case distinction is dense (enumerates almost all cases already anyway), make it exhaustive. And if the case distinction is sparse with a good default, use the default.

This makes sense.

include/arch/arm/arch/32/mode/object/structures.h
midnightveil: Prior art also: https://github.com/seL4/seL4/blob/28b8f4c40d4a48a206bedc7875c6695f6106f34b/src/arch/arm/32/object/objecttype.c#L577

include/arch/x86/arch/64/mode/object/structures.h
lsf37: I should add that some of these functions are explicitly written like this in the proofs as well (but not always). There is this technique where instead of making a function partial, you make the function total by returning a value that still works for most of the properties you want. That way you often can get away with tracking fewer preconditions.

I'm not sure that is going on here (in fact I'm fairly sure it is not), I just want to point out that there is a lot more subtlety here and people have not just forgotten to do things. A lot of this is on purpose. Apart of course from the cases where people did indeed forget :-)

include/arch/arm/arch/32/mode/object/structures.h
lsf37: Note that `fail(..)` is `halt()`, which generates an explicit proof obligation, which is different to `assert(..)` which disappears in the proof.

The thing that bugs me is mostly the `!"string"` in the `assert`. I know it's legal and works, I just don't like it :-)

midnightveil: Right. We could add an assert variant that just takes a string and always fail. I'm used to all the horrible things C allows now :)

Idea is that you see a message, not just assert(false), but I suppose it doesn't matter much.

(`assert(0)` to me is worse than `!"string"`)

lsf37: You're not wrong that seeing a message is better. I don't know, maybe I just need to get over myself on that one.

lsf37: But picking up your proposal: we could call it `debug_fail` for instance to make clear that it only fires for debug configs.

0aids:
Ok so generic summary I assume is the following:
1. Replace `assert(!"reason")` with a `debug_fail("reason")` macro in reasonable theoretically unreachable cases.
2. `include/arch/x86/arch/64/mode/object/structures.h`'s `cap_get_mode*` functions, check exhaustiveness.
3. Possible documentation or comments for places to check when adding new caps. 

For 1, I'll put `debug_fail` (an alternative to `fail` that doesn't create a proof obligation) in `include/assert.h`. I'll also replace all relevant areas that do the same thing already with asserts (and `fail` for 32bit x86?)

For 2, the reason I believed that `cap_get_mode*` functions were exhaustive is because the equivalent functions for 32bit x86 just failed on the call. 
IE `include/arch/x86/arch/32/mode/object/structures.h`
```c
static inline word_t CONST cap_get_modeCapSizeBits(cap_t cap)
{
    fail("Invalid mode cap type");
}
```
And that the 2 caps that are x64-specific are<sup>?</sup> both checked (the `pml4` cap and `pdpt` cap). If this reasoning is not satisfactory, we can just ignore x64 for now?

<sup>?</sup> the seL4 manual specifies that `vcpu` and `pdpt` cap are x64 specific, but the x64 specific code checks `pml4` and `pdpt` (`vcpu` is checked in generic x64/x86 code).

For 3, if it's just small comments where places need to be updated when new caps are added, then I assume it's relevant enough to go with this PR?

Issue #1613, opened 2026-03-11
Feature request: Generic Aarch64 kernel platform


kent-mcleod:
Goal: It seems like it would be possible to add a generic aarch64 kernel platform, e.g. `-DKernelPlatform=ArmGeneric`. This configuration would produce a kernel.elf that could be run on compliant hardware targets provided the bootloader (eg elfloader) could provide the right system layout for the kernel.

This issue aims to track the feasibility of this hypothesis.

kent-mcleod:
Regarding interrupt controller: We currently distinguish GICv2 and GICv3 drivers because they have different CPU interfaces. System register interface on GICv3 vs MMIO register interface on GICv2.

Assuming GICv2 is phasing out and not present on newer platforms, I'd suggest starting with a GICv3-only assumption. It'd probably also be possible to support a build config with both GICv2 and GICv3 support with a runtime selection mechanism, but this is probably not an initial goal to investigate.

kent-mcleod:
Regarding: `declare_default_headers()` that each platform.cmake file declares:

### MAX_IRQ: 

This config is used to size the intStateIRQTable array and intStateIRQNode CNode to a static size at compile time to a platform specific value. This could still be a config option for the ArmGeneric platform, set to an upper bound of the interrupt number to support across all target platforms. Alternatively, it could be later refactored to be a boot time allocated structure based on boot arguments.
```
irq_state_t intStateIRQTable[INT_STATE_ARRAY_SIZE];
/* CNode containing interrupt handler endpoints - like all seL4 objects, this CNode needs to be
 * of a size that is a power of 2 and aligned to its size. */
cte_t intStateIRQNode[BIT(IRQ_CNODE_SLOT_BITS)] ALIGN(BIT(IRQ_CNODE_SLOT_BITS + seL4_SlotBits));
```

### INTERRUPT_CONTROLLER and NUM_PPI

Fixing the interrupt controller to GICv3 causes these to have constant values for all platforms.

### TIMER

This variable declares a timer driver via a header file path in the build system. For the ArmGeneric platform, this uses the architecture specified GenericTimer.

### TIMER_FREQUENCY

Used in kernel conversions between ticks and time (in us or ms). With proposed RFC to convert kernel time accounting to ticks only basis, the frequency constant is no longer required. 

### CLK_MAGIC and CLK_SHIFT

These values are unused on aarch64 currently. (They'd also be removed from aarch64 with the ticks-based timer RFC).

### KERNEL_WCET

[...]

PR #1745, opened 2026-09-11
Pre-RFC: Multikernel Kernel Boot Code / Loader Changes

midnightveil:
This is a pre-RFC containing the code changes to the seL4 boot protocol done in Trustworthy System's [multikernel-manifest](https://github.com/au-ts/multikernel-manifest) repository around this time last year. This was previously shared [here for comments in our private fork](https://github.com/au-ts/seL4/pull/2)[^1]. The code has been pulled out from the kernel branch and almost certainly does not build and contains "dead" changes.

The code within this PR can be roughly categorised into three parts, and where attention should be focused:

1. `libsel4/include/sel4/loader_info.h`. This contains *an* implementation of a loader protocol. I'm not convinced this is the ideal solution, perhaps something with a set of TLV (tag-length-value) records would be better, or otherwise.

2. `ksKernelElfPaddrBase` in the statedata. This allows the same kernel image to be loaded at different physical addresses (since #1516 has been implemented) without recompiling. Thus the associated changes to `common_arm.lds` (note that this is not strictly necessary).

3. `src/arch/arm/kernel/boot.c` and specifically `try_init_kernel` which does the initialisation based on the loader info.

An additional, fourth change in `tools/hardware/outputs/json.py` allows the loader to map the UART for the kernel prior to the kernel switching virtual address spaces: this might be worth it standalone.

This doesn't include other changes necessary for the Multikernel, *just* the changes necessary to load the kernel memory, and userspace untyped regions (normal/device) at runtime with a single build of the kernel.

I'm opening a PR to prompt asynchronous discussion. The idea here is to use this to help write up a proper RFC for multikernel changes to work.

---

Ping of @dasch8-neutrality as I believe you did a summit talk on this (although I have not attended, so I don't know the contents of the talk).

Related: #1613.

[^1]: I will replicate the discussion from there.

### DIFF
--- a/config.cmake
+++ b/config.cmake
@@ -150,12 +150,25 @@ config_string(KernelNumPriorities NUM_PRIORITIES
 config_string(KernelMaxNumNodes MAX_NUM_NODES "Max number of CPU cores to boot" DEFAULT 1
               DEPENDS "${KernelNumDomains} EQUAL 1" UNQUOTE)
 
-# Set CONFIG_ENABLE_SMP_SUPPORT as an alias of CONFIG_MAX_NUM_NODES > 1
 if(KernelMaxNumNodes GREATER 1)
-  config_set(KernelEnableSMPSupport ENABLE_SMP_SUPPORT ON)
+    config_set(KernelEnableSMPSupport ENABLE_SMP_SUPPORT ON)
 else()
-  config_set(KernelEnableSMPSupport ENABLE_SMP_SUPPORT OFF)
+    config_set(KernelEnableSMPSupport ENABLE_SMP_SUPPORT OFF)
 endif()
+
+config_option(
+    KernelEnableMultikernelSupport ENABLE_MULTIKERNEL_SUPPORT
+    "Multikernel support"
+    DEFAULT OFF
+)
+
+# XXX: HACK.
+config_string(
+    KernelMultikernelNumCPUs MULTIKERNEL_NUM_CPUS
+    "Multikernel number of CPUs"
+    DEFAULT 1
+    UNQUOTE
+)
 
 config_string(
   KernelStackBits
--- a/include/api/types.h
+++ b/include/api/types.h
@@ -138,7 +138,7 @@ extern struct debug_syscall_error current_debug_error;
     do {                                                                       \
         out_error(ANSI_BOLD "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
                 ANSI_BOLD " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n", \
-                CURRENT_CPU_INDEX(),                                           \
+                NODE_STATE(boot_cpu_id),                                       \
                 __func__, __LINE__, NODE_STATE(ksCurThread),                   \
                 THREAD_NAME,                                                   \
                 (word_t)getRestartPC(NODE_STATE(ksCurThread)),                 \
--- a/src/kernel/boot.c
+++ b/src/kernel/boot.c
@@ -741,7 +741,7 @@ BOOT_CODE static bool_t provide_untyped_cap(
  * @param first_untyped_slot First available untyped boot info slot.
  * @return true on success, false on failure.
  */
-BOOT_CODE static bool_t create_untypeds_for_region(
+BOOT_CODE bool_t create_untypeds_for_region(
     cap_t      root_cnode_cap,
     bool_t     device_memory,
     region_t   reg,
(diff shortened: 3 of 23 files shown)


config.cmake
```diff
@@ -150,13 +150,26 @@ config_string(KernelNumPriorities NUM_PRIORITIES
...
+)
+
+# XXX: HACK.
+config_string(
+    KernelMultikernelNumCPUs MULTIKERNEL_NUM_CPUS
+    "Multikernel number of CPUs"
+    DEFAULT 1
+    UNQUOTE
+)
+
```
midnightveil, line 172: https://github.com/au-ts/seL4/pull/2#discussion_r3509261316

---

@lsf37: What does this one do?

---

@midnightveil: I might have reverted the changes. It was a hack because there is a bunch if code that creates structure based on CONFIG_MAX_NUM_NODES so I qant that to be 0, but then for a few other places I still wanted the number of actual cpus. I don't know if those places still exist in this PR.

---

@lsf37: Ok, that is the bit I was worried about :-)

When the multikernel has to know the number of available CPUs then it's an indication that multikernel separation is broken. Needing to know its own CPU number is fine, it has to write to the things that depend on it, but needing to know things about other CPUs is concerning. It could still be fine, depending on how exactly it occurs.

---

@midnightveil iirc (I might be wrong, it's been a while), the primary reason was the MPIDR map, for SGIs on GICv3.

I can't remember what we settled upon as there was upstream changes to this - iirc the user-level API expects MPIDRs? In which case that's not necessary; however:

In that case the other reason was just to be able to pass the list of MPIDRs to the root task from the loader - as otherwise different root tasks cannot communicate. There is a few other ways to do this, though.

---

@lsf37 Just passing through information would be fine. And I guess a MPIDR map would also be fine, that's more a map of what other devices are available in the machine.

---

@Indanz I think the kernel only needs to communicate its own MPIDR for the current core.

The whole list the kernel doesn't know itself either and user space can extract that from DTB itself if it needs to, or share the info via shared memory.

---

@midnightveil The DTB doesn't provide the MPIDRs, it provides the target_cpu value for PSCI. Which is supposed to be MPIDR, but technically it can be something else. On the ODROIDC4 for instance, DTB tells you 0,1,2,3 but MPIDR is 0, 0x100, 0x200, 0x300.

However, as I mentioned: this could be passed to the root task as part of the initial root task image by the loader, so it's not strictly necessary for the kernel to do it.

(And if there was some scheme where the loader boots the primary instance of seL4 and then userspace 'hotplugs' the rest of seL4 out of boot via SMC PSCI, then it couldn't be given to seL4 as it wouldn't be known).

(So yes, this could be removed, and is probably the least important of these changes).

---

@Indanz The Linux [documentation](https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/cpus.yaml) is very clear about this:

```
      On ARM v8 64-bit systems this property is required and matches the
      MPIDR_EL1 register affinity bits.

        * If cpus node's #address-cells property is set to 2

          The first reg cell bits [7:0] must be set to bits [39:32] of
          MPIDR_EL1.

          The second reg cell bits [23:0] must be set to bits [23:0] of
          MPIDR_EL1.

        * If cpus node's #address-cells property is set to 1

          The reg cell bits [23:0] must be set to bits [23:0] of MPIDR_EL1.
```
ODROIDC4 has address-cells set to 2, so bits [7:0] must be set to bits [39:32] of MPIDR_EL1, which corresponds to Aff3.

So either the dts for Odroid-C4 is wrong in using Aff2 instead of Aff3, or there is a mistake in the Linux documentation.

Whatever the case, the MPIDR is encoded in the DTB, one way or the other.

---

@midnightveil Well, for certain the ODROIDC4 PSCI firmware only works for on with 0,1,2,3 and that's how the elfloader works for the dtb values for booting odroid cores, and so does linux. Notably using 0x100/0x200 etc for odroid
PSCI_CPU_ON does not work.

I think the ARM specification of the argument meaning changed at some point.

---

@Indanz Last time we checked this, the ARM specification also mentioned MPIDR. I think it probably does the same bits shuffle, and that was probably added later when they introduced Aff3. Would have to check other DTS files to see what's going on exactly, but I can't be bothered right now.

midnightveil: Comment now: I believe this is probably unnecessary now and can be removed.

src/kernel/boot.c
```diff
@@ -960,27 +960,21 @@ BOOT_CODE static word_t init_avail_reg(word_t n_available,
 
 
 BOOT_CODE static bool_t check_reserved_memory(word_t n_reserved,
-                                              const region_t *reserved)
+                                              region_t *reserved)
 {
-    printf("Reserved virt address space regions: %"SEL4_PRIu_word"\n",
-           n_reserved);
+    bool_t needs_sort = false;
```
midnightveil, line 965: Somehow I think this got missed when splitting out this code: something that this did is sort the regions of reserved / available memory, as there was no "easy" way to ensure that they came in order.

Specifically, (On AArch64) previous the kernel region was always the first inserted in the reserved regions, followed by any others; this implicitly assumes the kernel is at the lowest region of memory. This is no longer the case when we have multiple kernels.

It might however still be possible to not have to do a sort, either by inserting in sorted order (is this really different to a sort?) or by structuring the loader info in such a way that there is only one set of reserved regions.

Currently, the current-kernel-memory and reserved arrays in the loader info are two separate arrays. This is because we needed to know the current kernel memory. It might however be possible to enforce that current-kernel-memory is in the reserved array *anyway*.

include/api/types.h
```diff
@@ -138,7 +138,7 @@ extern struct debug_syscall_error current_debug_error;
     do {                                                                       \
         out_error(ANSI_BOLD "<<" ANSI_GREEN "seL4(CPU %" SEL4_PRIu_word ")"    \
                 ANSI_BOLD " [%s/%d T%p \"%s\" @%lx]: " M ">>" ANSI_RESET "\n", \
-                CURRENT_CPU_INDEX(),                                           \
+                NODE_STATE(boot_cpu_id),                                       \
```
midnightveil, line 141: Alternative: print the MPIDR. In the BootInfo (passed to userspace) there's a currentNodeID and numNodes fields, whether we want to keep that being 0/1 on multikernel I am uncertain.

Indanz:
Neutrality's summit [talk ](https://events.linuxfoundation.org/sel4-summit/program/schedule/)slides can be found [here](https://hosted-files.sched.co/sel4summit2026/cc/sel4_policy_free_boot.pdf).

An interesting extra addition they propose are named resources, which seems like something Microkit could also use. Of course there are always other ways of passing info to user space, but having something somewhat standardised wouldn't be bad I think.

As far as boot protocol, our quick conclusion was that all kernels should be passed the same information, so that all kernels have full overview of all resources and can check for overlap and do other sanity checks. The resources are partitioned by CPU. Maybe have a logical CPU cores to HW CPUID map, and tag all resources with the logical core number. Then, if things are sorted by address, it's easy to check for overlap and conflicts.

I think the main challenge with the boot protocol is to keep it simple, but flexible enough. If it's too complicated, then we can as well use a restricted subset of an existing format like DTB.

Now that there is a multikernel mailing list (contact Gerwin if you want to subscribe!), I think it's best to have high-level discussion over there and focus on implementation specific details in PRs like this.

PR #1747, opened 2026-09-14
x64: remove unused get_current_esp()

lsf37:
The function is unused and contains inline an assembly block that returns a pointer, which the C parser currently does not translate correctly (will fail if used), and will in the future flag as error.

See also https://github.com/seL4/l4v/pull/1059


midnightveil (approved):

Indanz (approved):
It probably was a debug function.

PR #1749, opened 2026-09-15
aarch64,vspace: force XN for VMKernelOnly

lsf37:
On AArch64, read access is not required to run code (as opposed to AArch32). Executing code essentially confers read capabilities to page content, so we need to force NX=1 for pages that do not have read rights set. In seL4 these are `VMKernelOnly` pages, i.e. those with neither `VMReadOnly` or `VMReadWrite`.


src/arch/arm/64/kernel/vspace.c
```diff
@@ -788,7 +788,8 @@ static lookupPTSlot_ret_t lookupPTSlot(vspace_root_t *vspace, vptr_t vptr)
  */
 static pte_t makeUserPagePTE(paddr_t paddr, vm_rights_t vm_rights, vm_attributes_t attributes, vm_page_size_t page_size)
 {
-    bool_t nonexecutable = vm_attributes_get_armExecuteNever(attributes);
+    bool_t nonexecutable = vm_attributes_get_armExecuteNever(attributes) ||
```
midnightveil: If this is easier for verification, that's OK, but I think I would prefer that the code in `decodeARMFrameInvocation` in `arm/64/kernel/vspace.c` reports an error if you try to map a page that has no permissions. I don't see why you woiuld ever want to do that.

(Hmm, I guess a user *could* want to map something as execute-only but no read or anything; but I suppose that's not permitted by seL4's cap permissions at this time).

Indanz: Yes, agreed. It should be fine to map a page execute-only, but it should only be allowed if the page cap has read permission.

This is okay as a stop-gap fix if the proper fix needs more time, but this would be a breaking change as it is.

midnightveil: > Yes, agreed. It should be fine to map a page execute-only, but it should only be allowed if the page cap has read permissioms.

That is the case with this PR already fwiw, I wanted something slightly different, that it errors if you don't have any permissions.

Indanz:
What about something like:

```diff
diff --git a/src/arch/arm/64/kernel/vspace.c b/src/arch/arm/64/kernel/vspace.c
index 7334c1d77..6b54f3977 100644
--- a/src/arch/arm/64/kernel/vspace.c
+++ b/src/arch/arm/64/kernel/vspace.c
@@ -1515,7 +1515,7 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         cap_t vspaceRootCap;
         vspace_root_t *vspaceRoot;
         asid_t asid, frame_asid;
-        vm_rights_t vmRights;
+        vm_rights_t vmRights, capRights;
         vm_page_size_t frameSize;
         vm_attributes_t attributes;
         findVSpaceForASID_ret_t find_ret;
@@ -1530,8 +1530,8 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         vspaceRootCap = current_extra_caps.excaprefs[0]->cap;
 
         frameSize = cap_frame_cap_get_capFSize(cap);
-        vmRights = maskVMRights(cap_frame_cap_get_capFVMRights(cap),
-                                rightsFromWord(getSyscallArg(1, buffer)));
+        capRights = cap_frame_cap_get_capFVMRights(cap);
+        vmRights = maskVMRights(capRights, rightsFromWord(getSyscallArg(1, buffer)));
 
         if (unlikely(!isValidNativeRoot(vspaceRootCap))) {
             current_syscall_error.type = seL4_InvalidCapability;
@@ -1597,6 +1597,12 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+        if (unlikely(!seL4_CapRights_get_capAllowRead(capRights) &&
+            !vm_attributes_get_armExecuteNever(attributes))) {
+            userError("ARMPageMap: Frame cap needs read rights to map frame executable");
+            current_syscall_error.type = seL4_IllegalOperation;
+            return EXCEPTION_SYSCALL_ERROR;
+        }
         setThreadState(NODE_STATE(ksCurThread), ThreadState_Restart);
         return performPageInvocationMap(asid, cap, cte,
                                         makeUserPagePTE(base, vmRights, attributes, frameSize), lu_ret.ptSlot);
```
(It would make sense to turn the debug error print in `maskVMRights` into an explicit error too instead.)

src/arch/arm/64/kernel/vspace.c
Indanz: I don't think so. `maskVMRights` returns `VMKernelOnly` for a cap with no read or write rights and then this new check will disable execution, because it only sees the end result, not the original cap rights.

lsf37: We could do that, but it would be inconsistent with other current behaviour. Write-only mappings are also silently downgraded to VMKernelOnly, and I want to keep this to a fix not to general API improvement.

lsf37:
> What about something like:

I would only want to do that if we change how this works for other mappings (including the x86 EPT rights for instance). Otherwise we have error reporting for this one obscure case on one architecture and different (downgrade) behaviour for the rest.

I do agree that an explicit error could be a nicer API, so I'm not against a separate PR, but that would need an RFC and a long enough comment period to check if anyone is relying on the current behaviour. There are other downgrade behaviours in the kernel, though, so this would need a bit more survey and thinking.

src/arch/arm/64/kernel/vspace.c
Indanz: The difference is that write-only isn't supported by hardware, while execute-only is.

The current API allows mapping a page execute-only, while this change would always disallow that, so it is an API break.

lsf37: It is true that it already is an API break, yes, but it's unavoidable. Returning error instead of downgrading is inconsistent and a further API break and is avoidable.

Indanz: It is avoidable though, by still allowing execute-only mappings if the cap has read rights. Only downside of this is more verification work.

Returning an error for something that shouldn't have been allowed is not an API break. We could silently downgrade instead of returning an error if you prefer, while still allowing execute-only mappings, but that seems more effort, as we would need to pass the original cap rights to `makeUserPagePTE`.

lsf37:
On Indan's idea, I pushed a separate commit that allows execute-only pages when the cap has Read right. (To be squashed into the previous commit if people are happy with it)

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1631,7 +1632,8 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
         vspaceRootCap = current_extra_caps.excaprefs[0]->cap;
 
         frameSize = cap_frame_cap_get_capFSize(cap);
-        vmRights = maskVMRights(cap_frame_cap_get_capFVMRights(cap),
+        capRights = cap_frame_cap_get_capFVMRights(cap);
+        vmRights = maskVMRights(capRights,
                                 rightsFromWord(getSyscallArg(1, buffer)));
```
Indanz:
```suggestion
        vmRights = maskVMRights(capRights, rightsFromWord(getSyscallArg(1, buffer)));
```

src/arch/arm/64/kernel/vspace.c
```diff
@@ -1698,9 +1700,11 @@ static exception_t decodeARMFrameInvocation(word_t invLabel, word_t length,
             return EXCEPTION_SYSCALL_ERROR;
         }
 
+        bool_t cap_read = capRights == VMReadOnly || capRights == VMReadWrite;
```
Indanz: Although not necessary, this is conventional for C code:
```suggestion
        bool_t cap_read = (capRights == VMReadOnly || capRights == VMReadWrite);
```

Indanz (approved):

src/arch/arm/64/kernel/vspace.c
lsf37: Yes, fair. The new commit (about to be squashed) addresses this and allows execute-only if Read is present on the cap. We're still downgrading silently when that is not there.

Indanz (approved):
I'd expect it to be easier for verification to pass `capRights` to `makeUserPagePTE` and do the same check as you had before there instead of introducing `cap_read` here. But whatever works for verification is fine.

lsf37:
> I'd expect it to be easier for verification to pass `capRights` to `makeUserPagePTE` and do the same check as you had before there instead of introducing `cap_read` here. But whatever works for verification is fine.

Turned out there was not much difference (only one "execute this symbolically" step), and this is a bit nicer to read, I think.

midnightveil (approved):

PR #1750, opened 2026-09-15
aarch32,vspace: reserve HW ASID 0 for empty PD

lsf37:
Previously, the empty/global PD used the currently running hw ASID/VMID, leading to stale TLB mappings being available in the nominally empty address space. Instead, reserve hw ASID 0 and use it for the empty/global PD.


src/arch/arm/32/kernel/vspace.c
```diff
@@ -997,6 +997,24 @@ bool_t CONST isIOSpaceFrameCap(cap_t cap)
...
+    /* dsb/isb analogous to armv_contextSwitch_HWASID() */
+    dsb();
+    /* First switch to global PD on old ASID. Stale TLB entries may exist
+       under the old ASID, but no new stale mappings can be added any more. */
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    setCurrentPD(addrFromKPPtr(armUSGlobalPD));
+#else
+    setCurrentPD(addrFromKPPtr(armKSGlobalPD));
+#endif
+    isb();
```
lsf37, line 1014: I have that `isb()` in here, because `armv_contextSwitch_HWASID` has. I think it can be left out, but would like a second opinion on it.

KurtWu10: The call to `setCurrentPD()` above has already included an `isb()`:
- under hyp, `setCurrentPD()` calls `setCurrentPDPL2()`, which has an `isb()` before return
- under non-hyp, the `setCurrentPD()` implementation has an `isb()` before return

so the `isb()` here is redundant.

Indanz: Same is true for the `dsb()`, so that can be removed too.

Indanz (commented):
I'm not sure if this code is correct, I think we should follow the logic in [armv_contextSwitch_HWASID](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/armv/armv7-a/armv/context_switch.h#L22), which calls `writeContextIDAndPD` for HYP.

At this point it might be a good idea to just remove `setCurrentPD`, the only remaining callers are `activate_kernel_vspace` and `setGlobalPD`. The former can just call the latter, and the latter is clearer if it open-codes what it wants to do directly.

Indanz:
Maybe something like (on top of your branch):

```diff
diff --git a/include/arch/arm/arch/32/mode/machine.h b/include/arch/arm/arch/32/mode/machine.h
index 37215e704..38e41bf6d 100644
--- a/include/arch/arm/arch/32/mode/machine.h
+++ b/include/arch/arm/arch/32/mode/machine.h
@@ -324,19 +324,6 @@ static inline word_t readDACR(void)
     return reg;
 }
 
-static inline void setCurrentPD(paddr_t addr)
-{
-    /* Before changing the PD ensure all memory stores have completed */
-    if (config_set(CONFIG_ARM_HYPERVISOR_SUPPORT)) {
-        setCurrentPDPL2(addr);
-    } else {
-        dsb();
-        writeTTBR0Ptr(addr);
-        /* Ensure the PD switch completes before we do anything else */
-        isb();
-    }
-}
-
 static inline void setKernelStack(word_t stack_address)
 {
     /* Setup kernel stack pointer.
diff --git a/include/arch/arm/armv/armv7-a/armv/context_switch.h b/include/arch/arm/armv/armv7-a/armv/context_switch.h
index 73dee245e..521e89fc5 100644
--- a/include/arch/arm/armv/armv7-a/armv/context_switch.h
+++ b/include/arch/arm/armv/armv7-a/armv/context_switch.h
@@ -33,9 +33,8 @@ static inline void armv_contextSwitch_HWASID(pde_t *cap_pd, hw_asid_t hw_asid)
      * The reserved Page directory method has shown to perform better
      * than the reserved ASID method.
      *
-     * We do not call setCurrentPD here as we want to perform a
-     * minimal number of DSB and ISBs and the second PD switch we
-     * do does not need a DSB
+     * We want to perform a minimal number of DSB and ISBs and the
+     * second PD switch does not need a DSB.
      */
     dsb();
     writeTTBR0Ptr(addrFromKPPtr(armKSGlobalPD));
diff --git a/src/arch/arm/32/kernel/vspace.c b/src/arch/arm/32/kernel/vspace.c
index 43e1dedb3..0af4655fc 100644
--- a/src/arch/arm/32/kernel/vspace.c
+++ b/src/arch/arm/32/kernel/vspace.c
@@ -53,6 +53,8 @@
 #define ATTRINDX_NONCACHEABLE 0x0  /* strongly ordered or device memory */
 #endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
+static void setGlobalPD(void);
+
 struct resolve_ret {
     paddr_t frameBase;
     vm_page_size_t frameSize;
@@ -548,7 +550,7 @@ BOOT_CODE void activate_kernel_vspace(void)
        that everything we've written (particularly the kernel page tables)
        is committed. */
     cleanInvalidateL1Caches();
-    setCurrentPD(addrFromKPPtr(armKSGlobalPD));
+    setGlobalPD();
     invalidateLocalTLB();
     lockTLBEntry(PPTR_BASE);
     lockTLBEntry(PPTR_VECTOR_TABLE);
@@ -1000,19 +1002,20 @@ bool_t CONST isIOSpaceFrameCap(cap_t cap)
 /** Switch to the empty PD on the reserved HW ASID */
 static void setGlobalPD(void)
 {
-    /* dsb/isb analogous to armv_contextSwitch_HWASID() */
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    writeContextIDAndPD(hwASIDReserved, addrFromKPPtr(armUSGlobalPD));
+#else
+    /* Before changing the PD ensure all memory accesses have completed */
     dsb();
     /* First switch to global PD on old ASID. Stale TLB entries may exist
        under the old ASID, but no new stale mappings can be added any more. */
-#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
-    setCurrentPD(addrFromKPPtr(armUSGlobalPD));
-#else
-    setCurrentPD(addrFromKPPtr(armKSGlobalPD));
-#endif
+    writeTTBR0Ptr(addrFromKPPtr(armKSGlobalPD));
+    /* Ensure the PD switch completes before we do anything else */
     isb();
     /* Switch to reserved HW ASID. Only empty/global kernel mappings are
        now available from the TLB. */
     setHardwareASID(hwASIDReserved);
+#endif
 }
 
 void setVMRoot(tcb_t *tcb)
```

lsf37:
> Maybe something like (on top of your branch):

Yes, I think that makes sense. Will try it out.

Indanz:
@KurtWu10, what do you think, is there currently a `DSB` missing in [writeContextIDAndPD()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/arch/32/mode/machine_pl2.h#L32)? My feeling is that there should be a DSB before writing the new context and page table info.

lsf37:
I've pushed the proposed diff as a second commit (to be squashed later). Verification update is in progress.

KurtWu10:
> is there currently a DSB missing in [writeContextIDAndPD()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/arch/32/mode/machine_pl2.h#L32)?

I also feel that a `dsb` is missing here, as in the other branch of [armv_contextSwitch_HWASID()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/armv/armv7-a/armv/context_switch.h#L22-L25).

The only (complicated) case where a `dsb` is not required is as follows: [writeContextIDAndPD()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/arch/32/mode/machine_pl2.h#L32) is called in [armv_contextSwitch_HWASID()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/armv/armv7-a/armv/context_switch.h#L22-L25), which in turn is called in [switchToThread_fp()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/arch/32/mode/fastpath/fastpath.h#L43-L51). after [vcpu_switch()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/src/arch/arm/object/vcpu.c#L236-L259), and if both the old vcpu and the new vcpu are active, then there's a `dsb` when using [vcpu_save()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/src/arch/arm/object/vcpu.c#L32-L38) to save the old vcpu.

Indanz:
> > is there currently a DSB missing in [writeContextIDAndPD()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/arch/32/mode/machine_pl2.h#L32)?
> 
> I also feel that a `dsb` is missing here, as in the other branch of [armv_contextSwitch_HWASID()](https://github.com/seL4/seL4/blob/6bd2385582caf2caaef5d36c3c968013b63cc6d7/include/arch/arm/armv/armv7-a/armv/context_switch.h#L22-L25).

That's why I proposed to add it to `writeContextIDAndPD()`, so that all callers have it if needed. There's an ISB there already.

> The only (complicated) case where a `dsb` is not required is as follows: [...]

That's just coincidence and doesn't affect our reasoning.

So having thought about this a bit more, I think both a DSB as well as the ISB are not necessary:
- We are in hypervisor mode and VTTBR does not apply to us, it is out of context and currently not used for address translations, so there are no pending or future memory accesses to synchronise. An ISB is only necessary if we want to use the new table immediately in kernel mode somehow (with address translation instructions that explicitly use the user page table and reading the result in PAR. I have a local branch which removes ISBs for normal context switches).
- Even ignoring the above, the DSB is never necessary, because the table and VMID are updated atomically. The kernel only accesses global entries, so even if it was using the page tables at VTTBR, speculative accesses would use either the old or the new table. I'd argue that having no DSB and letting speculative accesses use the old table (assuming that happens) would be safer than letting them use the new table.

Indanz (approved):

src/arch/arm/32/kernel/vspace.c
```diff
@@ -1126,23 +1139,24 @@ hw_asid_t findFreeHWASID(void)
...
     invalidateTranslationASID(hw_asid);
     armKSHWASIDTable[hw_asid] = asidInvalid;
 
-    /* Increment the NextASID index */
+    /* Increment the NextASID index, skipping the reserved ASID 0 on wrap */
     armKSNextASID++;
-
+    if (armKSNextASID == hwASIDReserved) {
+        armKSNextASID = hwASIDMin;
+    }
```
midnightveil, line 1159: I'm not entirely sure this is necessary.  Wouldn't `findFreeHWASID` skip it anyway? Maybe it is easier if `armKSNextASID` is kept valid between iterations though, so I'm fine to keep this as is. (plus you'd probably have to reverify..)

midnightveil (approved):
Can't see anything else on ARM32 where HWASID0 is used.

src/arch/arm/32/kernel/vspace.c
lsf37: I did need it as an invariant for `findFreeHWASID` to be correct, so I think it is actually needed.

midnightveil: Hm, weird. I don't understand but I guess the verification is more trustworthy than me, I didn't want to block this PR either way.

KurtWu10:
> I think both a DSB as well as the ISB are not necessary: [...]

I'm not sure whether TLB conflict aborts (sections B3.9.2 and B3.9.5 of ARM DDI 0406C.d) could be generated, with or without synchronisation. I'm not familiar with the kernel and architecture details enough.

src/arch/arm/32/kernel/vspace.c
lsf37: The reason might be more on the proof side, tbh. We have a slightly different formulation of the behaviour of that wrapping loop, and they are only equivalent if `armKSNextASID` never points to `hwASIDReserved`. But that does still mean we need the C do this as well.

lsf37:
Given we're not entirely sure yet about the isb/dsb situation and it is safer to leave them in, I'll merge this later as is when I have the proof side ready (with isb+dsb still there). 

Happy to keep discussing here and tweak later in another PR if/when we are sure.

src/arch/arm/32/kernel/vspace.c
midnightveil: Right, yeah. I was wondering if you had some kind of thing that assumed that it was valid after exit / between kernel invocations

Indanz: If we reclaim a hw ASID 256 times, `armKSNextASID` will wrap back to zero. So we need a check either at the beginning (and then we don't need to init `armKSNextASID` to something valid), or here. I think avoiding `armKSNextASID` ever being reserved is nicer.

Like the comment says: "skipping the reserved ASID 0 on wrap".

Indanz:
> > I think both a DSB as well as the ISB are not necessary: [...]

To be clear, I am talking about the hypervisor case only here! Non-hyp does need both the DSB and the ISB, as that page table is actively being used and the ASID change is for the current ASID.

> I'm not sure whether TLB conflict aborts (sections B3.9.2 and B3.9.5 of ARM DDI 0406C.d) could be generated, with or without synchronisation. I'm not familiar with the kernel and architecture details enough.

For HYP, VMID and tables are changed atomically, so B3..9.2 does not apply. The combination of VMID+page tables stays consistent and TLB conflict aborts can't happen. (Also those registers aren't actively being used at the moment, as the kernel has its own page table in HYP mode.)

> Given we're not entirely sure yet about the isb/dsb situation

I think there is agreement that the current version is correct, but I think I confused people by giving the impression I meant that the DSB+ISB could be removed for non-HYP too. I was only talking about `writeContextIDAndPD()`, but didn't make that clear enough.

Indanz (approved):

PR #1751, opened 2026-09-16
fix(build): don't break when GCC_COLORS is set with specific options

MaddyGuthridge:
GCC can be configured to use specific colours for specific kinds of output. This is done by setting `GCC_COLORS=error=01;31:warning=01;35:note=01;36:caret=01;32:locus=01:quote=01` or similar.

Problematically, this script blindly evaluated the variable, rather than checking if it was defined, and did something similar for `FORCE_COLORED_OUTPUT`. This meant that the semicolons in the variable were considered to be statement delimiters, causing cmake to think that multiple args had been given to `if`, and giving an error.

This commit updates the configuration to prevent this issue, and hopefully reduce the risk of random environment variables getting evaluated as cmake code.

I don't have much familiarity with cmake, so I apologise if I have done this in a questionable way. The issue may also be present elsewhere in the cmake files. I only fixed in the place where I personally encountered it, and sadly do not have the time to dive through the rest of the build system to try to find more instances of the bug.


gcc.cmake
```diff
@@ -170,14 +170,14 @@ mark_as_advanced(CCACHE)
 # We upgrade this to -fdiagnostics-color=always if FORCE_COLORED_OUTPUT is set
 # We default FORCE_COLORED_OUTPUT=ON if GCC_COLORS is set in the environment
 # otherwise FORCE_COLORED_OUTPUT is left off.
-if($ENV{GCC_COLORS})
+if(DEFINED ENV{GCC_COLORS})
```
lsf37: That was indeed just wrong, thanks for finding it. If I read the docs correctly, GCC would interpret an env variable GCC_COLORS that is set but empty still as "no colours", so if we want to properly match that we could do 

`if(NOT "$ENV{GCC_COLORS}" STREQUAL "")`

instead.

gcc.cmake
```diff
@@ -170,14 +170,14 @@ mark_as_advanced(CCACHE)
...
     set(coloured_output ON)
 else()
     set(coloured_output OFF)
 endif()
 option(FORCE_COLORED_OUTPUT "Always produce ANSI-colored output." ${coloured_output})
 mark_as_advanced(FORCE_COLORED_OUTPUT)
-if(${FORCE_COLORED_OUTPUT})
+if(FORCE_COLORED_OUTPUT)
```
lsf37, line 143: Your version is slightly nicer style and more direct, so I'm fine with it, but at least this one is not a problem. The reason is that the only values this variable can have are `ON` and `OFF`, which is fine when we evaluate them.

lsf37:
It's fine to ignore the style checker (we've recently adjusted the cmake style and this file is not yet updated, I'll push a separate fixup commit when everything is ready), but the gitlint check is complaining about line length (title max 50, body max 72). Could you please reformat for that?

We don't usually use `fix(build)` tags, but `gcc.cmake:` or just `cmake:` would be useful.

I did a grep for `if($ENV` and `if(${` and the remaining cases all look benign. This one is the only one that uses ENV this way, and the other evaluations with `${..}` are on variables where the values are controlled.

MaddyGuthridge:
Thanks for the review! I believe I have resolved the issue (and added editor guides for commit message line lengths). Let me know if any further adjustments are needed :)

Indanz:
Signed-of-by is missing, you can add it with `git commit -s --amend`.

MaddyGuthridge:
My bad (didn't get preserved when I reformatted the message). Should be fixed now :3

lsf37:
Thank you. I've tweaked the commit message slightly so that we're not confused about the FORCE_COLORED_OUTPUT change in the future and added a style commit. 

This is now ready to merge when the tests have passed.

lsf37 (approved):

PR #1753, opened 2026-09-18
stm32mp2: Add missing timer to elfloader-devices

Indanz:
Otherwise CNTVOFF_EL2 is not initialised to zero and the clocks will be out of sync between cores and start at an arbitrary (sometimes too high) value.

~~This really should be done by the kernel init when possible.~~


bruelc:
> Otherwise CNTVOFF_EL2 is not initialised to zero and the clocks will be out of sync between cores and start at an arbitrary (sometimes too high) value.
> 
> This really should be done by the kernel init when possible.

thanks !!
So it works by luck with the kernel booted in EL2 ?  I don't know why CNTVOFF_EL2 was so different in EL1
 
edit:  in EL2 we read cntpct_el0, cntvoff_el2 is only used in EL1, which uses cntvct_el0  (quite confusing :()). 
that explain the discrepancy, it didn't work by luck in EL2

lsf37 (approved):

Indanz:
Ah, yes, if the kernel is in EL1 then it can't do this. I'll update the commit message, but wait with pushing till the hw tests are done.

midnightveil (approved):

Indanz:
There is still some issue with SMP though.

Maybe we're running into the speculative access problem that we had with Elfloader, but this time it's the kernel that triggers speculative reads to reserved memory? Would that just hang the kernel, or give some kind of error print somewhere?

bruelc:
> There is still some issue with SMP though.
> 

Arg, which configuration ? 

> 
> Maybe we're running into the speculative access problem that we had with Elfloader, but this time it's the kernel that triggers speculative reads to reserved memory? Would that just hang the kernel, or give some kind of error print somewhere?

Are there some indications from optee ? Like Iac or serc ? Or is it just a random crash somewhere in secure space which indeed could imply a invalid speculative access ?

Indanz:
> Arg, which configuration ?

All SMP configs fail, see https://github.com/seL4/seL4/actions/runs/35339432442.

> Are there some indications from optee ? Like Iac or serc ? Or is it just a random crash somewhere in secure space which indeed could imply a invalid speculative access ?

No, output just hangs at seemingly random places during early boot. Example:

```
Fri, 18 Sep 2026 12:39:54 GMT   Enabling MMU and jumping to entry point...
Fri, 18 Sep 2026 12:39:54 GMT   
Fri, 18 Sep 2026 12:39:54 GMT   Bootstrapping kernel
Fri, 18 Sep 2026 12:39:54 GMT   Available phys memory regions: 1
Fri, 18 Sep 2026 12:39:54 GMT     [84000000..180000000)
Fri, 18 Sep 2026 12:39:54 GMT   Reserved virt address space regions: 3
Fri, 18 Sep 2026 12:39:54 GMT     [ffffff8084000000..ffffff8084251000)
Fri, 18 Sep 2026 12:39:54 GMT     [ffffff8084251000..ffffff808425d876)
Fri, 18 Sep 2026 12:39:54 GMT     [ffffff808425e000..ffffff8084644000)
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1   Panic 'Watchdog' at /usr/src/debug/optee-os-stm32mp/4.0.0-stm32mp-r3/core/drivers/stm32_iwdg.c:218 <stm32_iwdg_it_handler>
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1   TEE load address @ 0x82000000
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1   Call stack:
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1    0x82007f30
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1    0x82043b88
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1    0x8201c5dc
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1    0x820416b4
Fri, 18 Sep 2026 12:40:20 GMT   E/TC:1    0x82014368
Fri, 18 Sep 2026 12:40:20 GMT   I/TC: Halting CPU 0
```
Edit: Not random, but MCS configs hang after the clock sync test output.

(I currently don't have much time to track this down, but if it's deterministic then adding printfs to boot code should pin it down.)

Indanz:
This PR can be merged, the SMP issue is independent.

bruelc:
Those SMP configs are ok at my end. Let's wait Monday We ll need to find out what are the differences now in our setup. 

In your log:   E/TC:1   Panic 'Watchdog' at /usr/src/debug/optee-os-stm32mp/4.0.0-stm32mp-r3/core/drivers/stm32_iwdg.c:218 <stm32_iwdg_it_handler>

Are you on top of the smc watchdog disabled  patches ?

midnightveil:
My guess is that bringing up the secondary cores takes too long.

This is before userspace has started and so the watchdog disable can't have happened yet.

Maybe it needs to be run in the elfloader?

bruelc:
> My guess is that bringing up the secondary cores takes too long.
> 

30 secs is too long and it is immediate on my setup. The watchdog is a good guard here


> 
> 
> This is before userspace has started and so the watchdog disable can't have happened yet.
> 
> 
> 
> Maybe it needs to be run in the elfloader?

Maybe not, it is useful here imho

midnightveil:
It's 30 seconds from the last time the watchdog was kicked though, right? That could have been as early as 29 seconds; though I would expect the reset period to be much lower: U-Boot's default is 4s or so.

But yeah I would expect it to be faster too.

bruelc:
> It's 30 seconds from the last time the watchdog was kicked though, right? That could have been as early as 29 seconds; though I would expect the reset period to be much lower: U-Boot's default is 4s or so.
> 
> But yeah I would expect it to be faster too.

yes 32 secs in fact since the iwdg initial init or last kick. 

we can print in the elfloader in mode the remaining time to rule out a iwdg init bug (with stm32_iwdg_get_timeleft()) ?

lsf37:
I'll merge this for now, because the SMP issue seems unrelated. I have temporarily switched off SMP tests for this platform, because there are a bunch of pull requests to merge that are otherwise blocked. We can switch it back on when we think we have a solution.

Indanz:
> My guess is that bringing up the secondary cores takes too long.

Yes, 26 seconds of no progress during kernel boot, see timestamps. I'd classify that as a hang.
Disabling the watchdog earlier won't help here.

bruelc:
I just realized that I configure with -DNUM_NODES=2. that could be the cause of the hang if not forcing it.
I forgot to update the plat/config.cmake to override the default which is 4

midnightveil:
Yes, I can also reproduce that... I see it's also defaulting to 4. That would definitely do it.

The i.MX93 also has the same thing, https://github.com/seL4/sel4bench/pull/50 and https://github.com/seL4/ci-actions/commit/dda28c57b388906c5d0f149aacfecaa833b518e0 were done to fix this. There might be a better way to do it, but it's how done in past.

---

What would others think of:

```diff
diff --git a/config.cmake b/config.cmake
index 750ab9b19..471829420 100644
--- a/config.cmake
+++ b/config.cmake
@@ -157,6 +157,10 @@ else()
   config_set(KernelEnableSMPSupport ENABLE_SMP_SUPPORT OFF)
 endif()

+if (DEFINED KernelPlatformMaxNumNodes AND KernelMaxNumNodes GREATER KernelPlatformMaxNumNodes)
+  message(FATAL_ERROR "KernelMaxNumNodes ('${KernelMaxNumNodes}') must be less than KernelPlatformMaxNumNodes ('${KernelPlatformMaxNumNodes}')")
+endif()
+
 config_string(
   KernelStackBits
   KERNEL_STACK_BITS
diff --git a/src/plat/imx93/config.cmake b/src/plat/imx93/config.cmake
index f41b1d5a9..1a82d7321 100644
--- a/src/plat/imx93/config.cmake
+++ b/src/plat/imx93/config.cmake
@@ -22,6 +22,7 @@ if(KernelPlatformIMX93)
         INTERRUPT_CONTROLLER arch/machine/gic_v3.h
         KERNEL_WCET 10u
     )
+    set(KernelPlatformMaxNumNodes 2)
 endif()

 add_sources(
diff --git a/src/plat/stm32mp2/config.cmake b/src/plat/stm32mp2/config.cmake
index f35757059..8db0b4b5d 100644
--- a/src/plat/stm32mp2/config.cmake
+++ b/src/plat/stm32mp2/config.cmake
@@ -40,6 +40,8 @@ if(KernelPlatformMP2)
   declare_default_headers(
     TIMER_FREQUENCY 40000000 NUM_PPI 32 MAX_IRQ 415 INTERRUPT_CONTROLLER arch/machine/gic_v2.h
     TIMER drivers/timer/arm_generic.h KERNEL_WCET 10u)
+
+  set(KernelPlatformMaxNumNodes 2)
 endif()

 add_sources(DEP "KernelPlatformMP2" CFILES src/arch/arm/machine/gic_v2.c
 ```

**Comment by midnightveil** [comment]
Alternatively we could add this check to the elfloader: 

```diff
project kernel/
diff --git a/tools/hardware/outputs/elfloader.py b/tools/hardware/outputs/elfloader.py
index 24af655d4..790506fb8 100644
--- a/tools/hardware/outputs/elfloader.py
+++ b/tools/hardware/outputs/elfloader.py
@@ -144,6 +144,11 @@ def run(tree: fdt.FdtParser, hardware: rule.HardwareYaml, config: config.Config,
     devices = tree.get_elfloader_devices()
     cpu_info = get_elfloader_cpus(tree, devices)

+    # On ARM/RISC-V platforms this is how many CPUs the kernel expects to boot.
+    kernel_num_nodes = int(kernel_config_dict["CONFIG_MAX_NUM_NODES"])
+    if len(cpu_info) < kernel_num_nodes:
+        raise ValueError("The kernel is configured with CONFIG_MAX_NUM_NODES ({}) but we only found {} cpus".format(kernel_num_nodes, len(cpu_info)))
+
     max_reg = 1
     device_info = []
     for dev in devices:

project tools/seL4/
diff --git a/elfloader-tool/CMakeLists.txt b/elfloader-tool/CMakeLists.txt
index 5ef1dcc..36ca641 100644
--- a/elfloader-tool/CMakeLists.txt
+++ b/elfloader-tool/CMakeLists.txt
@@ -375,6 +375,7 @@ if(DEFINED KernelDTBPath)
             --hardware-schema "${schema_file}"
             --dtb "${KernelDTBPath}"
             --sel4arch "${KernelSel4Arch}"
+            --kernel-config-flags "CONFIG_MAX_NUM_NODES=${KernelMaxNumNodes}"
         VERBATIM
         DEPENDS ${KernelDTBPath} ${config_file} ${schema_file}
     )
```

Both of these would catch the issue.  This one has the advantage that we don't need to maintain even more information in the build system, although it wouldn't help for people not using the elfloader.

bruelc:
I confirm that adding _set(KernelPlatformMaxNumNodes 2)_ in kernel/src/plat/stm32mp2 fixes the SMP run.

I let you to finalizing the config.cmake fixups and the elfloader checks?

Thank you!

bruelc:
there is a function get_cpus() function that infers this from the device tree. Couldn't this be used ?

midnightveil:
Yeah, that's what get_elfloader_cpus uses in seL4.

Indanz:
I forgot how annoying this issue is when you run into this issue.

> What would others think of:

No, that doesn't solve the problem, too many SoCs have variable amount of cores, and for users the chance to run into this issue is greatest when that is the case. Also, we haven't ran into this yet, but if more cores than expected are booted, things go wrong too, and that wouldn't be caught with this check.

Your second proposal of a check in Elfloader is much better.

midnightveil:
Please see: https://github.com/seL4/seL4_tools/pull/269, https://github.com/seL4/seL4/pull/1767, https://github.com/seL4/ci-actions/pull/529#issuecomment-5769822097.

Running SMP sel4bench:

```
smp Benchmarks (iteration 0)
============================

Failed to alloc at: 1

Failed to find a way to map address 0x44200704
Failed to init stm32 timeout timer
Shutting down stm32mp257f-ev1-1
```

It does work in sel4test, so I'm not sure what goes wrong here...
It's not because of the unaligned address.

bruelc:
> Please see: [seL4/seL4_tools#269](https://github.com/seL4/seL4_tools/pull/269), #1767, [seL4/ci-actions#529 (comment)](https://github.com/seL4/ci-actions/pull/529#issuecomment-5769822097).
> 
> Running SMP sel4bench:
> 
> ```
> smp Benchmarks (iteration 0)
> ============================
> 
> Failed to alloc at: 1
> 
> Failed to find a way to map address 0x44200704
> Failed to init stm32 timeout timer
> Shutting down stm32mp257f-ev1-1
> ```
> 
> It does work in sel4test, so I'm not sure what goes wrong here... It's not because of the unaligned address.

I don't see this. Can you share your gen_config.h

Edit: sorry . Ok I can reproduce on smp

Indanz:
Where is RCC_TIM2 defined? I can't find it.

midnightveil:
MAP_IF_NULL appends _PADDR and _SIZE to the name.

bruelc:
Tim2 rcc is not properly destroyed in libplatsupport, I ll push a pr

bruelc:
please see https://github.com/seL4/util_libs/pull/207
Refactor the destroy code to unmap the RCC

PR #1755, opened 2026-09-18
arm,config.cmake: set SErrorIgnore default to ON 

lsf37:
Change the default of `KernelAArch64SErrorIgnore` from OFF to ON, and allow it for verification builds. In addition require it in the provided AARCH64_verified cmake configs.

The previous rationale for keeping the default OFF was that user-level systems should be set up such that they do not trigger SErrors, and that having the default OFF means they will not be noticed.

However, with KernelAArch64SErrorIgnore OFF, unprivileged user code when given access to e.g. device memory can halt the machine instead of just producing a fault.

Similarly, speculative execution in the kernel that touches holes in the kernel window can also halt the machine. (The kernel never touches those holes, even if they are mapped, but speculative execution might.)

Overall, it is better to have the default on the safe choice for production systems even if it might mean the issue is masked during development if people don't know about this option.

Longer-term we should add a fault handler for SError or something similar, despite the information flow leak that this introduces.


Indanz (commented):
You also need to remove `KernelAArch64SErrorIgnore` from configs/seL4Config.cmake:150, basically undoing commit 7b0602c5f724d888ba20c6ee9a61cd87e27c9906 (which I never understood, because I added the option to make it user configurable, but that commit made it ignore what users configured).

lsf37:
> You also need to remove `KernelAArch64SErrorIgnore` from configs/seL4Config.cmake:150, basically undoing commit [7b0602c](https://github.com/seL4/seL4/commit/7b0602c5f724d888ba20c6ee9a61cd87e27c9906) (which I never understood, because I added the option to make it user configurable, but that commit made it ignore what users configured).

As our favourite token machines say: you are absolutely right :-)

This now basically fully reverts 7b0602c5f724d888ba20c6ee9a61cd87e27c9906, because it no longer needs to be set by `tx2` either, because it's now the global default. This means it can now be handled like any other user config.

The reason 7b0602c5f724d888ba20c6ee9a61cd87e27c9906 was necessary was evaluation order, I think. The config setting declaration comes later than the platform settings. Not sure any more what would have been overwriting what, but some action was necessary. That's now no longer the case.

Indanz (approved):
Looks fine now.

lsf37:
This is good to merge from my side whenever it fits in the queue.

Indanz:
I can't merge it when it's not up-to-date, hence me updating it. But then someone else merged something, so I'll have to keep doing this till I get lucky.

midnightveil (approved):

PR #1756, opened 2026-09-18
x86,vcpu: clean up VPID in finalise

lsf37:
Previously the `vcpu_finalise()` function only cleaned up the VCPU/TCB association, but not the VPID/VCPU association.

Factor out the `invalidateVPID()` call + table update in `findFreeVPID` and reuse the resulting `releaseVPID()` in `vcpu_finalise()`.


src/arch/x86/object/vcpu.c
```diff
@@ -1489,6 +1495,17 @@ static void invalidateVPID(vpid_t vpid)
     }
 }
 
+/** Disassociate a VPID from its VCPU */
+static void releaseVPID(vpid_t vpid)
+{
+    vcpu_t *vcpu = x86KSVPIDTable[vpid];
```
Indanz, line 1500: Empty new line after declarations please.

Indanz (approved):
All looks correct.

Not sure what the added value of the asserts is though, I think they can be removed. The assumptions they check can be checked by reviewing this file instead (which I did), all code is fairly self-contained.

lsf37:
> Not sure what the added value of the asserts is though, I think they can be removed. The assumptions they check can be checked by reviewing this file instead (which I did), all code is fairly self-contained.

Fair, and they are pretty standard properties, too. I mostly followed other functions that do the same, but there is no point, really. Will remove.

PR #1757, opened 2026-09-18
arm+x86,iospace: final capability check for unmap

lsf37:
ARMIOPageTableUnmap and X86IOPageTableUnmap did not check for isFinalCapability like other page table unmap operations which leaves the mapping data stored in the other caps inconsistent.

Add the missing check and document the new error conditions.


Indanz (approved):
Okay, this makes behaviour consistent with normal page table unmaps.

But I don't understand how all this can be safe: If we allow cap copies, then we can have mapped and unmapped caps. What assures that the final cap is a mapped one and that the page table gets unmapped when it gets deleted?

Indanz:
Ah, that is solved by allowing only page table cap copies of mapped PT caps.

PR #1758, opened 2026-09-18
arm,iospace: clear only valid, owned slots

lsf37:
The check in deleteIOPageTable was wrong: we can only clear slots that we have valid authority to, which means the PDE must be valid, of the right size, and of the right address.

The new check now mirrors the code in all other delete page table operations.


Indanz (approved):

src/arch/arm/object/iospace.c
```diff
@@ -388,8 +388,8 @@ void deleteIOPageTable(cap_t io_pt_cap)
             return;
         }
 
-        if (isIOPDEValid(lu_ret.iopdSlot) &&
-            iopde_ptr_get_page_size(lu_ret.iopdSlot) == iopde_iopde_pt &&
+        if (!isIOPDEValid(lu_ret.iopdSlot) ||
+            iopde_ptr_get_page_size(lu_ret.iopdSlot) != iopde_iopde_pt ||
             iopde_iopde_pt_ptr_get_address(lu_ret.iopdSlot) != (pptr_to_paddr((void *)cap_io_page_table_cap_get_capIOPTBasePtr(
                                                                                   io_pt_cap)))) {
```
Indanz, line 399: Two observations:
- Here it uses `iopde_iopde_pt_ptr_get_address(lu_ret.iopdSlot)`, while just below it does `addrFromPPtr(lu_ret.iopdSlot)`. To me it feels like they can't both be correct.
- Below it does `*lu_ret.iopdSlot = iopde_iopde_pt_new(0, 0, 0, 0);` before taking the address, to me it seems that would always result in address zero.

lsf37: I think these are correct:

- `lu_ret.iopdSlot` is a pointer to a PDE.  `iopde_iopde_pt_ptr_get_address` follows the pointer (`_ptr`), and reads the address stored in the PDE, i.e. the content of the slot, which is the table address the slot points to. That must also be what the cap points to.

- The `*lu_ret.iopdSlot = iopde_iopde_pt_new(0, 0, 0, 0)` zeroes the content of the slot, and `addrFromPPtr(lu_ret.iopdSlot)` is the physical address of the slot itself, which is what you need if you want to invalidate the cache for it.

Issue #1018, opened 2023-04-20
Check aarch64 system registers are properly isolated between guest VMs


kent-mcleod:
For each of the following registers we should confirm:

- What the access rights are for EL0 and EL1, 
- What the kernel does in EL1 or EL2 for access control:
    - Saved and restored in object state on context switch,
    - Access trapped and forwared as a fault
    - Emulated (eg vtimer registers)
    - Access allowed but RO register
- Then there should ideally be tests in sel4test for trying to break isolation between two guests or native threads for each one.

## 0b00 These encodings provide:
> - Instructions with an immediate field for accessing PSTATE, the current PE state.
> - The architectural hint instructions.
> - Barriers and the CLREX instruction.

There don't seem to be any system registers within these encodings (other than PSTATE which is handled elsewhere)

## 0b01 These encodings provide the cache maintenance, TLB maintenance, and address translation instructions.

There don't seem to be any system registers within these encodings

## 0b10 These encodings provide moves to and from:
> - Legacy AArch32 System registers for execution environments, to provide access to these registers from higher exception levels that are using AArch64.
> - Debug and trace registers.

**Debug registers**

OSDTRRX_EL1
MDCCINT_EL1
MDSCR_EL1
OSDTRTX_EL1
OSECCR_EL1
DBGBVR<n>_EL1 0-15
DBGBCR<n>_EL1 0-15
DBGWVR<n>_EL1 0-15
DBGWCR<n>_EL1 0-15
MDRAR_EL1
OSLAR_EL1
OSLSR_EL1
OSDLR_EL1
DBGPRCR_EL1
DBGCLAIMCLR_EL1
DBGAUTHSTATUS_EL1
DBGDTR_EL0
DBGDTRRX_EL0
DBGCLAIMSET_EL1
MDCCSR_EL0
DBGDTRTX_EL0
DBGVCR32_EL2

**Trace registers**

*TODO: The Arch reference manual states these are defined elsewhere.*

## 0b11 These encodings provide:
> - Moves to and from Non-debug System registers. The accessed registers provide system control, and system status information.
> - Access to Special-purpose registers.

MIDR_EL1
MPIDR_EL1
REVIDR_EL1
ID_PFR0_EL1
ID_PFR1_EL1
ID_DFR0_EL1
ID_AFR0_EL1
ID_MMFR0_EL1
ID_MMFR1_EL1
ID_MMFR2_EL1
ID_MMFR3_EL1
ID_ISAR0_EL1
[...]

PR #1761, opened 2026-09-19
aarch64,hyp: switch VDISR_EL2 and CSSELR_EL1

lsf37:
Context switch the registers `VDISR_EL2` and `CSSELR_EL1`, and trap accesses to RAS error records.

VDISR_EL2 is relevant for cores with FEAT_RAS. Under our current config, guest DISR_EL1 accesses are redirected to VDISR_EL2, which is currently not context switched. VDISR_EL2 is undefined without FEAT_RAS. In this case return 0 for read and ignore writes.

CSSELR_EL1 is the Cache Size Selection Register and was just missing.

Add both registers to the VCPU save/restore range.

Also set HCR_TERR in HCR_VCPU so that accesses to the RAS error records (ERRIDR_EL1, ERRSELR_EL1, ERX*_EL1) trap when FEAT_RAS is present.


include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -429,6 +435,46 @@ static inline void writeVMPIDR_EL2(word_t reg)
     MSR(REG_VMPIDR_EL2, reg);
 }
 
+/** MODIFIES: */
+/** DONT_TRANSLATE */
```
lsf37: Needed, because it uses 32-bit inline asm on 64-bit arch.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -429,6 +435,46 @@ static inline void writeVMPIDR_EL2(word_t reg)
...
+    word_t reg;
+    MRS(REG_ID_AA64PFR0_EL1, reg);
+    return ((reg >> ID_AA64PFR0_EL1_RAS) & MASK(4)) != 0;
+}
+
+/* VDISR_EL2 is undefined when FEAT_RAS is not supported (e.g. Cortex-A53/A57) */
+static inline word_t readVDISR_EL2(void)
+{
+    word_t reg = 0;
+    if (cpu_has_ras()) {
```
lsf37, line 461: I wasn't sure if `cpu_has_ras()` should read only once at boot and store a value instead of reading every time. Unclear to me how fast/slow those reads are or if there is a different way entirely.

midnightveil:
Related: https://github.com/seL4/seL4/issues/1018
I suspect we have several others: that list is only Armv8.0, not anything following.

lsf37:
> Related: #1018 I suspect we have several others: that list is only Armv8.0, not anything following.

I'm mostly going through the ones we definitely know are missing for now. I have a few more coming up for AArch32 as well. After that we still need to go through that full list and confirm what is missing.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
lsf37: Maybe we should do this statically by board instead. But then there is a risk that people will be missing it when they do a board port.

midnightveil: I think we'd want to avoid doing that. I think we're all in support of #1613 and making new board-specific configs would be the opposite of that.

(I mean, in theory it is a performance degradation and it would be faster to have no code dealing with this for systems without FEAT_RAS, but unfortunately that's something we have to deal with. As long as we stored all actively used feature bits in the same cache line we should be ok...)

lsf37: Ok, and Google claims that reading CPU configuration registers should be way faster (1-3 cycles) than reading from memory, so querying the register every time is faster than caching the value somewhere.

I think this is fine as is then.

Indanz: I wouldn't count on those numbers to be true for all registers though... But should still be fast enough.

include/arch/arm/armv/armv8-a/64/armv/vcpu.h
```diff
@@ -429,6 +435,49 @@ static inline void writeVMPIDR_EL2(word_t reg)
...
+        MRS(REG_VDISR_EL2, reg);
+    }
+    return reg;
+}
+
+/* VDISR_EL2 is undefined when FEAT_RAS is not supported (e.g. Cortex-A53/A57).
+   In this case, ignore writes. */
```
Indanz:
```suggestion
/* VDISR_EL2 is undefined when FEAT_RAS is not supported. In this case, ignore writes. */
```
Too many don't have it, no point in naming a few. From the CPUs we support, currently only Cortex-A55 and Cortex-A76 might have it.

Indanz (commented):
Please read and write the whole 64-bits values. Top half is currently RES0, but don't count on it staying zero or not being writeable.

midnightveil:
Side-comment: we have boards which have non-heterogeneous CPUs, so some cores may have certain feature bits and others not. This might cause some funky issues on SMP, though admittedly it's not verified...

midnightveil (approved):

lsf37:
> Side-comment: we have boards which have non-heterogeneous CPUs, so some cores may have certain feature bits and others not. This might cause some funky issues on SMP, though admittedly it's not verified...

Ok, so it is definitely better to read at runtime. This should always produce the right result, even if you migrate cores, you'll then get this register zeroed, which I think is what one would want.

Indanz:
Also, doesn't 32-bit Arm have the same issue with `CSSELR` and `VDISR`? (The latter only on ARMv8 hardware running 32-bit code.)

Indanz:
> Side-comment: we have boards which have non-heterogeneous CPUs, so some cores may have certain feature bits and others not. This might cause some funky issues on SMP, though admittedly it's not verified...

Which boards are that? Do we actually support SMP across both sets of incompatible cores? That seems bad and waiting for subtle problems.

KurtWu10:
> Which boards are that? Do we actually support SMP across both sets of incompatible cores? That seems bad and waiting for subtle problems.

`rockpro64b`, which has four A53 cores in one cluster and two A72 cores in another cluster. I tested booting the SMP kernel a month ago but haven't started using all cores for userland applications.

Indanz:
> `rockpro64b`, which has four A53 cores in one cluster and two A72 cores in another cluster. I tested booting the SMP kernel a month ago but haven't started using all cores for userland applications.

Those should be compatible, that's like Arm's second or third big-little attempt, so they learned their lessons by then surely. Main difference seems to be 44-bit versus 40-bit physical address space size, but it should work if you configure it as 40-bit.

midnightveil:
It looks like Exynos5422 also has Cortex-A7 and Cortex-A15 cores. I don't see anything else. (Arguably NVIDIA's cores are somewhat distinct - iirc the TX2 has NVIDIA Denver cores and also normal ARM cores).

I'm surprised we don't have more, TBH, but I guess no-one has tried running seL4 on a bunch of Android phones or newer Laptops.

Indanz:
> It looks like Exynos5422 also has Cortex-A7 and Cortex-A15 cores.

Yeah, I wouldn't run those in SMP mode together.

lsf37:
> Also, doesn't 32-bit Arm have the same issue with `CSSELR` and `VDISR`? (The latter only on ARMv8 hardware running 32-bit code.)

Yes, there are more in the list of things to check, but I'm doing AArch32 hyp separately, otherwise the proof updates get very noisy.

lsf37:
> Please read and write the whole 64-bits values. Top half is currently RES0, but don't count on it staying zero or not being writeable.

Done. Also doesn't need to be hidden from the verification side, should have thought of that in the first place..

PR #1762, opened 2026-09-19
riscv/x64,fastpath: add ASID validity check

lsf37:
Add the ASID validity check that all of the other architectures perform also to RISC-V and x64. Leaving it out not only makes the fast path incorrect (not equivalent to slow path), but it is also an actual behaviour bug.

The shape is now the same for all architectures. For RISC-V I introduced a fast path variant of findVSpaceForASID to avoid the exception handling the original function is doing (same as other fast path setups). 

This will decrease performance numbers somewhat, but hopefully not by much. Also, there is no choice really, the old behaviour was just wrong.

Longer term, we should refactor this a bit. Maybe have an ASID check function with a standard interface that defined per architecture and then called from the generic fast path code. Also, using that Arm 32 `stored_hw_asid` of a completely wrong type with this horrible hack is terrible. But both of these should be done later, first I want to have correct behaviour.


Indanz (commented):
You've forgotten to update `fastpath_vm_fault`.

But at this point it may make more sense to do it the same way as `setVMRoot` does it, and to set it to the global mapping in `switchToThread_fp` instead of falling back to the slow path.

Because then all the code is contained to one spot and automatically correct (and because all ASID code is in one spot, may be faster than the current code too). Because it's closer to the slowpath code, it may be easier to proof for equality too, although it does handle an extra path.

If not doing the above, we probably want at least a helper function that avoids duplicating the same code three times, but maybe that complicates the proofs.

src/fastpath/fastpath.c
```diff
@@ -87,11 +87,25 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
...
 #endif
 
 #ifdef CONFIG_ARCH_IA32
+    /* Need to test that the ASID is still valid */
+    asid_t asid = cap_page_directory_cap_get_capPDMappedASID(newVTable);
+    asid_map_t asid_map = findMapForASID(asid);
+    if (unlikely(asid_map_get_type(asid_map) != asid_map_asid_map_vspace ||
+                 (vspace_root_t *)asid_map_asid_map_vspace_get_vspace_root(asid_map) != cap_pd)) {
+        slowpath(SysCall);
+    }
```
Indanz: Considering this is exactly the same as the 64-bit version, what about introducing a common wrapper for `cap_pml4_cap_get_capPML4MappedASID_fp` and `cap_page_directory_cap_get_capPDMappedASID` so the rest of the code doesn't need to be duplicated?

Indanz (commented):
Code looks correct, just the missing signal fastpath cases.

One discrepancy I found between the 32-bit and 64-bit versions of x86's `setVMRoot` is that the 32-bit one does more tlb bitmap updates compared to the 64-bit one. But I think it's just a matter of optimisation, not correctness, and that the 64-bit version just doesn't bother optimising rare corner cases.

src/fastpath/fastpath.c
lsf37: Yes, I think that makes sense.

lsf37:
> You've forgotten to update `fastpath_vm_fault`.

Right, good point, I will have a look.
 
> But at this point it may make more sense to do it the same way as `setVMRoot` does it, and to set it to the global mapping in `switchToThread_fp` instead of falling back to the slow path.
>
> Because then all the code is contained to one spot and automatically correct (and because all ASID code is in one spot, may be faster than the current code too). Because it's closer to the slowpath code, it may be easier to proof for equality too, although it does handle an extra path.

I'd rather keep the existing pattern. If we change it for the new ones, we should change it in the verified versions and that is too much proof disturbance for me now. I'm happy to add an issue for it, though, it's definitely worth looking into, especially if it might also end up being faster.

> If not doing the above, we probably want at least a helper function that avoids duplicating the same code three times, but maybe that complicates the proofs.

Yes, like I said in the PR description, for the future we should do something like that (or what you suggested above), but not directly for the fix.

lsf37:
> > You've forgotten to update `fastpath_vm_fault`.
> 
> Right, good point, I will have a look.

Turns out I forgot, but it was also fine :-). `KernelExceptionFastpath` (=`CONFIG_EXCEPTION_FASTPATH`, which guards `fastpath_vm_fault`) depends on `KernelSel4ArchAarch64`, and the AARCH64 path in `fastpath_vm_fault` does the ASID check + vmid check. There is no implementation for other architectures in that fast path yet overall.

The signal fast path doesn't switch to any new thread (it bails out to slow path for that case). It only applies to the case where it can return to the current thread.

So this should be complete.

src/fastpath/fastpath.c
lsf37: Ok, I've tried a few versions, this new one gives me the same `kernel.elf` as before and I think it's nice enough. There is still an `#ifdef` for the `hw_asid` that is passed out (ia32 completely ignores it, so it is 0), but that one is actually different and it doesn't really seem worth trying to wrap.

lsf37: Although I should probably move it into the header, I haven't done that bit yet. One sec.

src/fastpath/fastpath.c
```diff
@@ -12,6 +12,12 @@
 #endif
 #include <benchmark/benchmark_utilisation.h>
 
+#if defined(CONFIG_ARCH_X86_64)
+#define MAPPED_ASID(newVTable) cap_pml4_cap_get_capPML4MappedASID_fp(newVTable)
+#elif defined(CONFIG_ARCH_IA32)
+#define MAPPED_ASID(newVTable) cap_page_directory_cap_get_capPDMappedASID(newVTable)
+#endif
```
Indanz: This should be in `include/arch/x86/arch/[32,64]/mode/fastpath/fastpath.h`.

There is already a commonly named `cap_vtable_cap_get_vspace_root_fp` there, so following that, maybe rename `cap_pml4_cap_get_capPML4MappedASID_fp` to `cap_vtable_cap_get_MappedAsid_fp` and add a static inline function to 32-bit x86 that returns `cap_page_directory_cap_get_capPDMappedASID`?

src/fastpath/fastpath.c
lsf37: Ok, now :-)

src/fastpath/fastpath.c
```diff
@@ -86,15 +86,23 @@ void NORETURN fastpath_call(word_t cptr, word_t msgInfo)
...
 #endif
-
-#ifdef CONFIG_ARCH_IA32
+#if defined(CONFIG_ARCH_X86_64)
+    /* borrow the stored_hw_asid for PCID */
+    stored_hw_asid.words[0] = asid;
+#elif defined(CONFIG_ARCH_IA32)
     /* stored_hw_asid is unused on ia32 fastpath, but gets passed into a function below. */
     stored_hw_asid.words[0] = 0;
 #endif
```
Indanz: I think it's nicer if both use the same code, even if the asid isn't used on 32-bit x86. That will get optimised away anyway.

The reason ASID is unused on 32-bit is because the hardware has no PCID support. So the seL4 ASID validity check is only needed to make sure that the vspace is still valid.

Indanz (commented):
Github may have eaten one of my comments, or it may pop up again...

src/fastpath/fastpath.c
lsf37: I wasn't sure it would be optimised, but you're right, if it's not used it will be. Will update.

src/fastpath/fastpath.c
lsf37: We overlapped on that one, no worries.

src/fastpath/fastpath.c
lsf37: Ok, now also done. This is indeed nicer.

Indanz (approved):
Last nitpick: Commit message only mentions x64, not ia32. Suggestion:

Fastpath, RISC-V/x86: Add ASID Validity Checks

Add the ASID validity check that all of the other architectures
perform also to RISC-V and x86. Leaving it out not only makes the
fastpath incorrect (not equivalent to slowpath), but it is also an
actual behaviour bug.

src/fastpath/fastpath.c
```diff
@@ -394,7 +409,13 @@ void NORETURN fastpath_reply_recv(word_t cptr, word_t msgInfo)
...
 
 #ifdef CONFIG_ARCH_RISCV
-    stored_hw_asid.words[0] = cap_page_table_cap_get_capPTMappedASID(newVTable);
+    /* Need to test that the ASID is still valid */
+    asid_t asid = cap_page_table_cap_get_capPTMappedASID(newVTable);
```
midnightveil, line 413: side note here: the ASID on RISC-V isn't really used; it's set in the register but we actually do a full `sfence.vma` so we essentially force the architecture to flush the TLB and so we don't have any stale ASIDs in the TLB.

include/arch/x86/arch/64/mode/fastpath/fastpath.h
```diff
@@ -29,7 +29,7 @@ static inline vspace_root_t *cap_vtable_cap_get_vspace_root_fp(cap_t vtable_cap)
     return PML4E_PTR(vtable_cap.words[1]);
 }
 
-static inline word_t cap_pml4_cap_get_capPML4MappedASID_fp(cap_t vtable_cap)
+static inline word_t vs_root_getMappedASID_fp(cap_t vtable_cap)
```
midnightveil: what is `vs_root`? Is that `vspace_root`? I don't think we've done `vs_root` before.

src/fastpath/fastpath.c
lsf37: The check here is not so much to get the hardware ASID (which indeed we do not use on RISC-V), but to check that the cap is not stale and we are allowed to switch to this VSpace.

include/arch/x86/arch/64/mode/fastpath/fastpath.h
lsf37: yes, `vspace_root`. It looks like we indeed only use that abbreviation on the verification side. Happy to expand it.

src/fastpath/fastpath.c
lsf37: Maybe we should explicitly assign 0 like we do on ia32, but I have this unfounded hope that at some point we will have actual hardware ASIDs in RISC-V, so I left that part in.

PR #1765, opened 2026-09-20
arm,machine: prevent TEEHBR access

lsf37:
TEEHBR (ThumbEE Handler Base Register) is not context switched, but accessible to user space.

- trap access to it from user mode
- trap access to the configuration register in hyp mode so that guests cannot re-enable it.


lsf37:
This one should have no verification impact, just making the verification assumptions more true. Running the proof test to make sure. I've also switch on hw-test, but the machine queue seems to be sad this weekend.

src/arch/arm/armv/armv7-a/user_access.c
```diff
@@ -20,6 +20,26 @@
...
+
+static void disable_thumbee_user_access(void)
+{
+    uint32_t v;
+    /* When TEECR.XED == 0, TEEHBR is readable and writable for user space.
+     * Clear, disable, and trap access. */
+    MCR(TEEHBR, 0);
+    MRC(TEECR, v);
+    v |= TEECR_XED;
+    MCR(TEECR, v);
```
Indanz:
```suggestion
    MCR(TEECR, TEECR_XED);
```
I suspect you found [this](https://lists.infradead.org/pipermail/linux-arm-kernel/2012-March/091593.html) patch and implemented something equivalent, but the other bits are SBZ (Should Be Zero), so writing zero is fine.

src/arch/arm/armv/armv7-a/user_access.c
```diff
@@ -63,6 +83,11 @@ static void check_export_arch_timer(void)
 void armv_init_user_access(void)
 {
     uint32_t v;
+    /* ThumbEE */
+    MRC(ID_PFR0, v);
```
Indanz:
```suggestion

    /* Disable Jazelle RCT */
    MRC(ID_PFR0, v);
```
(with extra new line.)

src/arch/arm/armv/armv7-a/user_access.c
```diff
@@ -20,6 +20,26 @@
...
+    uint32_t v;
+    /* When TEECR.XED == 0, TEEHBR is readable and writable for user space.
+     * Clear, disable, and trap access. */
+    MCR(TEEHBR, 0);
+    MRC(TEECR, v);
+    v |= TEECR_XED;
+    MCR(TEECR, v);
+#ifdef CONFIG_ARM_HYPERVISOR_SUPPORT
+    /* Trap access to config register so guest cannot re-enable TEEHBR. */
+    setHSTR(getHSTR() | HSTR_TTEE);
```
Indanz: We might want to disable Jazelle too while we're at it, TJDBX, bit[17].

Indanz (approved):
That's a legacy Thumb-2EE or Jazelle RCT thing, looks like it's relevant for am335x, omap3, imx7 and other older platforms. Other than some nitpicks, code looks fine.

Used DDI0406C_C_arm_architecture_reference_manual.pdf to review code, newer manual doesn't document the bits any more.

src/arch/arm/armv/armv7-a/user_access.c
lsf37: Yes, that's nicer.

lsf37:
> That's a legacy Thumb-2EE or Jazelle RCT thing, looks like it's relevant for am335x, omap3, imx7 and other older platforms. 

Yes, definitely not something we want to support in any way. This is also how it slipped past the machine model infoflow validation for AArch32 from back in the day. The low level model (L3 in those days, not SAIL yet), didn't have Thumb mode.
 
> Used DDI0406C_C_arm_architecture_reference_manual.pdf to review code, newer manual doesn't document the bits any more.

Yes, it's just great :-)

Indanz:
> > That's a legacy Thumb-2EE or Jazelle RCT thing, looks like it's relevant for am335x, omap3, imx7 and other older platforms.
> 
> Yes, definitely not something we want to support in any way. This is also how it slipped past the machine model infoflow validation for AArch32 from back in the day. The low level model (L3 in those days, not SAIL yet), didn't have Thumb mode.

This is the kind of concrete information that was missing in the CAVEATS PR. :-)

Say there that AArch32 infoflow was validated with the low level L3 model, but that it misses Thumb and other modes.

> > Used DDI0406C_C_arm_architecture_reference_manual.pdf to review code, newer manual doesn't document the bits any more.
> 
> Yes, it's just great :-)

Even if we have an automated check against Arm supplied system description/model, then I still want to have a hw test that does a brute force search to find anything not mentioned there. Neither will be waterproof, but together it should give decent assurance.

src/arch/arm/armv/armv7-a/user_access.c
lsf37: I've done that in a second commit (fine to squash if we want), and added the same hyp procedure as well so it can't be re-enabled there. Please have a look if these make sense.

Indanz (approved):

PR #1766, opened 2026-09-20
caveats: describe missed-register risk

lsf37:
Explicitly point out that register set completeness is not covered by the verification, and that specifically AArch64 validation is still in progress.


CAVEATS.md
```diff
@@ -226,6 +227,18 @@ verified configurations in seL4 yet. For static systems, on hardware where
...
+
+Note that the verification assumptions include that there are no more registers
+in the machine than the formal model contains and seL4 knows about. If one is
+missed that will usually mean it is not context-switched or otherwise protected.
+Especially the hypervisor configurations have large sets of system registers and
+newer architecture versions, e.g. Armv8.3 vs Armv8.1, often introduce new system
+registers. The strength and completeness of validation of these register set
+assumptions is not equal between all architectures seL4 supports, and a deeper
+survey for AArch64, which is the architecture that currently sees the most
+change, is in progress.
```
Indanz, line 240: I'm inclined to leave this part out, as it doesn't give the user concrete information that they can use or rely on.

Having stronger guarantees based on Arm's architecture models would be nice, but that is a big project that will take a while to get done. Even when we have that, the actual hardware may behave in unexpected ways which are not documented, or fall outside of the CPU model (e.g. system registers for peripherals like GIC or coherency controllers or who knows what), so the above text will stay true no matter what we do.

If you do keep it, make it a separate paragraph.

Indanz (approved):
To give an indication how far behind things we are, the newest profile is Armv9.7-A.

We really need an automated way to keep up-to-date, especially if Arm keeps breaking backward compatibility.

CAVEATS.md
lsf37: We did do that kind of project on AArch32 non-hyp, and it's holding up pretty well. The main problem on AArch64 is a) it's hyp, so much more state to consider, and b) it's changing a lot (new versions all the time).

Of course you can never be sure even if you have a full formal connection to the SAIL spec, but there is currently a much higher risk of registers being missing on AArch64 hyp than on AArch32 non-hyp or RISC-V.

So I do want to say something in there, but maybe it's not coming across what I am trying to say :-) Maybe @Xaphiosis or @corlewis have ideas from the verification side.

lsf37:
> To give an indication how far behind things we are, the newest profile is Armv9.7-A.

We do only claim support for Armv8, at least. But yes, new things all the time.

> We really need an automated way to keep up-to-date, especially if Arm keeps breaking backward compatibility.

That would be very nice.

CAVEATS.md
Indanz: Well, AArch32 did miss TEEHBR, so I'm not entirely convinced. AArch32 also has HYP, so that's no excuse. I think Arm was just better at not breaking backward compatibility back then, now they're being sloppy and hasty.

Indanz: But if the point you're trying to make is that AArch32 was more thoroughly reviewed or check, then just say that directly.

KurtWu10: > Well, AArch32 did miss TEEHBR, so I'm not entirely convinced.

The TEEHBR one is for ARMv7 only, which is before Alastair Reid's work on cleaning up the pseudocode for Armv8 ([blog](https://alastairreid.github.io/ARM-v8a-xml-release/)). I hope the situation for Armv8 is better?

lsf37: > The TEEHBR one is for ARMv7 only, which is before Alastair Reid's work on cleaning up the pseudocode for Armv8 ([blog](https://alastairreid.github.io/ARM-v8a-xml-release/)). I hope the situation for Armv8 is better?

The problem is that we have a missing register for AArch64 hyp that also doesn't seem to be in Alastair's SAIl model. But maybe we just haven't found it because the name is different or something like that.

In any case Alastair's model is very good and we should either use it directly or use the XML that he is deriving it from (which I think Indan wants to do). I'm fine with either.

> But if the point you're trying to make is that AArch32 was more thoroughly reviewed or check, then just say that directly.

Fair point, but I do want to say that we're working on AArch64, and not just leave people hanging with "stuff might be missing". That sounds like we don't care, which is the opposite of what is going on.

